PR #367, opened 2022-06-02
Snow 539401

sfc-gh-yuwang:

   Add function that allow user to load local file into dataframe

   - [*] I am adding a new automated test(s) to verify correctness of my new code

   Check if user is using a local file. If yes, upload it to a temporary stage.

### DIFF
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -1,7 +1,7 @@
 #
 # Copyright (c) 2012-2022 Snowflake Computing Inc. All rights reserved.
 #
-
+import os.path
 from typing import Dict, Iterable, Union
 
 import snowflake.snowpark
@@ -208,7 +208,23 @@ class DataFrameReader:
             |</test>            |
             ---------------------
             <BLANKLINE>
+    Example 10:
+        Loading a JSON file from local:
+        >>> df = session.read.json("/users/user1/local.json")
+        >>> # Load the data from a local file or directory.
+        >>> df.show()
+        ------------------------
+        |"$1"                  |
+        ------------------------
+        |{                     |
+        |  "color": "yellow",  |
+        |  "fruit": "banana",  |
+        |  "taste": "sweet"    |
+        |}                     |
+        ------------------------
 
+    Loading other file type from local are the same. Instead of uploading manually, you can put your local path in the
+    function.
     """
 
     def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
@@ -249,10 +265,23 @@ class DataFrameReader:
 
         Args:
             path: The stage location of a CSV file, or a stage location that has CSV files.
+            Location of a local CSV file, or a location that has CSV files are also acceptable.
 
         Returns:
             a :class:`DataFrame` that is set up to load data from the specified CSV file(s) in a Snowflake stage.
         """
+        _ = self._session.sql("CREATE or REPLACE temp STAGE mystage").collect()
+        if os.path.exists(path) and os.path.isfile(path):
+            _ = self._session.file.put(path, "@mystage", auto_compress=True)
+            path = "@mystage" + "/" + path.split("//")[-1]
+        if os.path.exists(path) and os.path.isdir(path):
+            filelist = os.listdir(path)
+            for file in filelist:
+                if file.split(".")[-1].lower() == "csv":
+                    _ = self._session.file.put(
+                        os.path.join(path, file), "@mystage", auto_compress=True
+                    )
+            path = "@mystage/"
         if not self._user_schema:
             raise SnowparkClientExceptionMessages.DF_MUST_PROVIDE_SCHEMA_FOR_READING_FILE()
 
@@ -276,6 +305,7 @@ class DataFrameReader:
 
         Args:
             path: The stage location of a JSON file, or a stage location that has JSON files.
+            Location of a local JSON file, or a location that has JSON files are also acceptable.
 
         Returns:
             a :class:`DataFrame` that is set up to load data from the specified JSON file(s) in a Snowflake stage.
@@ -287,6 +317,7 @@ class DataFrameReader:
 
         Args:
             path: The stage location of an AVRO file, or a stage location that has AVRO files.
+            Location of a local AVRO file, or a location that has AVRO files are also acceptable.
 
         Returns:
             a :class:`DataFrame` that is set up to load data from the specified AVRO file(s) in a Snowflake stage.
@@ -298,6 +329,7 @@ class DataFrameReader:
 
         Args:
             path: The stage location of a PARQUET file, or a stage location that has PARQUET files.
+            Location of a local PARQUET file, or a location that has PARQUET files are also acceptable.
 
         Returns:
             a :class:`DataFrame` that is set up to load data from the specified PARQUET file(s) in a Snowflake stage.
@@ -309,6 +341,7 @@ class DataFrameReader:
 
         Args:
             path: The stage location of a ORC file, or a stage location that has ORC files.
+            Location of a local ORC file, or a location that has ORC files are also acceptable.
 
         Returns:
             a :class:`DataFrame` that is set up to load data from the specified ORC file(s) in a Snowflake stage.
@@ -320,6 +353,7 @@ class DataFrameReader:
 
         Args:
             path: The stage location of an XML file, or a stage location that has XML files.
+            Location of a local XML file, or a location that has XML files are also acceptable.
 
         Returns:
             a :class:`DataFrame` that is set up to load data from the specified XML file(s) in a Snowflake stage.
@@ -357,6 +391,19 @@ class DataFrameReader:
         return self
 
     def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
+        # upload file if it is local
+        _ = self._session.sql("CREATE or REPLACE temp STAGE mystage").collect()
+        if os.path.exists(path) and os.path.isfile(path):
+            _ = self._session.file.put(path, "@mystage", auto_compress=True)
+            path = "@mystage" + "/" + path.split("//")[-1]
+        if os.path.exists(path) and os.path.isdir(path):
+            filelist = os.listdir(path)
+            for file in filelist:
+                if file.split(".")[-1].lower() == format.lower():
+                    _ = self._session.file.put(
+                        os.path.join(path, file), "@mystage", auto_compress=True
+                    )
+            path = "@mystage/"
         if self._user_schema:
             raise ValueError(f"Read {format} does not support user schema")
         self._file_path = path
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -208,20 +208,21 @@ class DataFrameReader:
             |</test>            |
             ---------------------
             <BLANKLINE>
+
     Example 10:
         Loading a JSON file from local:
-        >>> df = session.read.json("/users/user1/local.json")
+        >>> df = session.read.json("tests/resources/testJson.json")
         >>> # Load the data from a local file or directory.
         >>> df.show()
-        ------------------------
-        |"$1"                  |
-        ------------------------
-        |{                     |
-        |  "color": "yellow",  |
-        |  "fruit": "banana",  |
-        |  "taste": "sweet"    |
-        |}                     |
-        ------------------------
+        -----------------------
+        |"$1"                 |
+        -----------------------
+        |{                    |
+        |  "color": "Red",    |
+        |  "fruit": "Apple",  |
+        |  "size": "Large"    |
+        |}                    |
+        -----------------------
 
     Loading other file type from local are the same. Instead of uploading manually, you can put your local path in the
     function.
@@ -270,18 +271,7 @@ class DataFrameReader:
         Returns:
             a :class:`DataFrame` that is set up to load data from the specified CSV file(s) in a Snowflake stage.
         """
-        _ = self._session.sql("CREATE or REPLACE temp STAGE mystage").collect()
-        if os.path.exists(path) and os.path.isfile(path):
-            _ = self._session.file.put(path, "@mystage", auto_compress=True)
-            path = "@mystage" + "/" + path.split("//")[-1]
-        if os.path.exists(path) and os.path.isdir(path):
-            filelist = os.listdir(path)
-            for file in filelist:
-                if file.split(".")[-1].lower() == "csv":
-                    _ = self._session.file.put(
-                        os.path.join(path, file), "@mystage", auto_compress=True
-                    )
-            path = "@mystage/"
+        path = self._upload_local_file_2_stage(path)
         if not self._user_schema:
             raise SnowparkClientExceptionMessages.DF_MUST_PROVIDE_SCHEMA_FOR_READING_FILE()
 
@@ -392,18 +382,7 @@ class DataFrameReader:
 
     def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         # upload file if it is local
-        _ = self._session.sql("CREATE or REPLACE temp STAGE mystage").collect()
-        if os.path.exists(path) and os.path.isfile(path):
-            _ = self._session.file.put(path, "@mystage", auto_compress=True)
-            path = "@mystage" + "/" + path.split("//")[-1]
-        if os.path.exists(path) and os.path.isdir(path):
-            filelist = os.listdir(path)
-            for file in filelist:
-                if file.split(".")[-1].lower() == format.lower():
-                    _ = self._session.file.put(
-                        os.path.join(path, file), "@mystage", auto_compress=True
-                    )
-            path = "@mystage/"
+        path = self._upload_local_file_2_stage(path, format)
         if self._user_schema:
             raise ValueError(f"Read {format} does not support user schema")
         self._file_path = path
@@ -499,3 +478,19 @@ class DataFrameReader:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_2_stage(self, path: str, format: str = "csv") -> str:
+        temp_stage = self._session.get_session_stage()
+        if os.path.exists(path) and os.path.isfile(path):
+            _ = self._session.file.put(path, temp_stage, auto_compress=False)
+            _, filename = os.path.split(path)
+            path = os.path.join(temp_stage, filename)
+        if os.path.exists(path) and os.path.isdir(path):
+            filelist = os.listdir(path)
+            for file in filelist:
+                if os.path.splitext(file)[-1][1:].lower() == format.lower():
+                    _ = self._session.file.put(
+                        os.path.join(path, file), temp_stage, auto_compress=False
+                    )
+            path = temp_stage
+        return path
--- a/tests/integ/scala/test_dataframe_reader_suite.py
+++ b/tests/integ/scala/test_dataframe_reader_suite.py
@@ -332,6 +332,30 @@ def test_to_read_files_from_stage(session, resources_path, mode):
         session.sql(f"DROP STAGE IF EXISTS {data_files_stage}")
 
 
+@pytest.mark.parametrize("mode", ["select", "copy"])
+def test_to_read_file_from_local(session, resources_path, mode):
+
+    test_files = TestFiles(resources_path)
+
+    reader = get_reader(session, mode)
+
+    try:
+        df = (
+            reader.schema(user_schema)
+            .option("compression", "auto")
+            .csv(f"{test_files.test_file_csv}")
+        )
+        res = df.collect()
+        res.sort(key=lambda x: x[0])
+        assert res == [
+            Row(1, "one", 1.2),
+            Row(2, "two", 2.2),
+        ]
+    finally:
+        pass
+    #     session.sql(f"DROP STAGE IF EXISTS {data_files_stage}")
+
+
 @pytest.mark.xfail(reason="SNOW-575700 flaky test", strict=False)
 @pytest.mark.parametrize("mode", ["select", "copy"])
 def test_for_all_csv_compression_keywords(session, temp_schema, mode):
@@ -439,6 +463,36 @@ def test_read_json_with_no_schema(session, mode):
 
 
 @pytest.mark.parametrize("mode", ["select", "copy"])
+def test_to_read_json_from_local(session, resources_path, mode):
+
+    test_files = TestFiles(resources_path)
+    reader = get_reader(session, mode)
+
+    df = reader.json(f"{test_files.test_file_json}")
+    res = df.collect()
+    assert res == [
+        Row('{\n  "color": "Red",\n  "fruit": "Apple",\n  "size": "Large"\n}')
+    ]
+    res = df.where(sql_expr("$1:color") == "Red").collect()
+    assert res == [
+        Row('{\n  "color": "Red",\n  "fruit": "Apple",\n  "size": "Large"\n}')
+    ]
+    # assert user cannot input a schema to read json
+    with pytest.raises(ValueError):
+        get_reader(session, mode).schema(user_schema).json(test_files.test_file_json)
+
+    # user can input customized formatTypeOptions
+    df2 = (
+        get_reader(session, mode)
+        .option("FILE_EXTENSION", "json")
+        .json(test_files.test_file_json)
+    )
+    assert df2.collect() == [
+        Row('{\n  "color": "Red",\n  "fruit": "Apple",\n  "size": "Large"\n}')
+    ]
+
+
+@pytest.mark.parametrize("mode", ["select", "copy"])
 def test_read_avro_with_no_schema(session, mode):
     avro_path = f"@{tmp_stage_name1}/{test_file_avro}"
 
@@ -459,6 +513,39 @@ def test_read_avro_with_no_schema(session, mode):
 
     # user can input customized formatTypeOptions
     df2 = get_reader(session, mode).option("COMPRESSION", "NONE").avro(avro_path)
+    res = df2.collect()
+    assert res == [
+        Row(str="str1", num=1),
+        Row(str="str2", num=2),
+    ]
+
+
+@pytest.mark.parametrize("mode", ["select", "copy"])
+def test_to_read_avro_from_local(session, resources_path, mode):
+
+    test_files = TestFiles(resources_path)
+    reader = get_reader(session, mode)
+
+    df = reader.avro(f"{test_files.test_file_avro}")
+    res = df.collect()
+    assert res == [
+        Row(str="str1", num=1),
+        Row(str="str2", num=2),
+    ]
+    # query_test
+    res = df.where(sql_expr('"num"') > 1).collect()
+    assert res == [Row(str="str2", num=2)]
+
+    # assert user cannot input a schema to read avro
+    with pytest.raises(ValueError):
+        get_reader(session, mode).schema(user_schema).avro(test_files.test_file_avro)
+
+    # user can input customized formatTypeOptions
+    df2 = (
+        get_reader(session, mode)
+        .option("COMPRESSION", "NONE")
+        .avro(test_files.test_file_avro)
+    )
     res = df2.collect()
     assert res == [
         Row(str="str1", num=1),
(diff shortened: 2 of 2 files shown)


sfc-gh-yuwang:

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -249,10 +266,12 @@ def csv(self, path: str) -> DataFrame:
 
         Args:
             path: The stage location of a CSV file, or a stage location that has CSV files.
```
sfc-gh-jdu, line 270:
```suggestion
            path: A path pointing to a local CSV file, a local directory that has CSV files, a CSV file on a stage, or a stage location that has CSV files.
```
Same below, but others don't support multiple files.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +475,19 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_2_stage(self, path: str, format: str = "csv") -> str:
```
sfc-gh-jdu, line 479:
```suggestion
    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
```

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +332,30 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+
+    reader = get_reader(session, mode)
+
+    try:
+        df = (
+            reader.schema(user_schema)
+            .option("compression", "auto")
+            .csv(f"{test_files.test_file_csv}")
+        )
+        res = df.collect()
```
sfc-gh-jdu, line 348: I know previously we wrote tests like this and we should fix it. We already have `Utils.check_answer` and you can use this for assertion (see how other tests use this method)

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +332,30 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+
+    test_files = TestFiles(resources_path)
+
+    reader = get_reader(session, mode)
+
+    try:
+        df = (
+            reader.schema(user_schema)
+            .option("compression", "auto")
+            .csv(f"{test_files.test_file_csv}")
```
sfc-gh-jdu, line 346:
```suggestion
            .csv(test_files.test_file_csv)
```
same below

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +332,30 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
 
 
+@pytest.mark.parametrize("mode", ["select", "copy"])
+def test_to_read_file_from_local(session, resources_path, mode):
+
+    test_files = TestFiles(resources_path)
+
+    reader = get_reader(session, mode)
+
+    try:
```
sfc-gh-jdu, line 342: you don't need try/finally here, right?

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +332,30 @@ def test_to_read_files_from_stage(session, resources_path, mode):
         session.sql(f"DROP STAGE IF EXISTS {data_files_stage}")
 
 
+@pytest.mark.parametrize("mode", ["select", "copy"])
+def test_to_read_file_from_local(session, resources_path, mode):
```
sfc-gh-jdu, line 336:
```suggestion
def test_to_read_csv_from_local(session, resources_path, mode):
```
to be consistent

sfc-gh-jdu: Also, for csv, we should test whether we can read from a directory, right?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -209,6 +209,26 @@ class DataFrameReader:
             ---------------------
             <BLANKLINE>
 
+    Loading other file type from local are the same. Instead of uploading manually, you can put your local path in the function.
```
sfc-gh-jdu, line 212: hmm I meant this line is not needed anymore : )

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -209,6 +209,26 @@ class DataFrameReader:
...
+            |{                    |
+            |  "color": "Red",    |
+            |  "fruit": "Apple",  |
+            |  "size": "Large"    |
+            |}                    |
+            -----------------------
+            <BLANKLINE>
+
+
+
```
sfc-gh-jdu, line 231: nit: you don't need these blank line?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -248,11 +268,13 @@ def csv(self, path: str) -> DataFrame:
         """Specify the path of the CSV file(s) to load.
 
         Args:
-            path: The stage location of a CSV file, or a stage location that has CSV files.
+            path: A path pointing to a local CSV file, a local directory that has CSV files, a CSV file on a stage, or
+            a stage location that has CSV files.
```
sfc-gh-jdu, line 272: I remember we need two indents here if an arg has multiple lines. But if the generated doc is ok then it's ok.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -275,7 +297,7 @@ def json(self, path: str) -> DataFrame:
         """Specify the path of the JSON file(s) to load.
 
         Args:
-            path: The stage location of a JSON file, or a stage location that has JSON files.
+            path: The stage location of a JSON file, or location of a JSON file.
```
sfc-gh-jdu, line 300: hmm I meant that similar to the sentence for csv, you should have
```
path: A path pointing to a local JSON file, or a JSON file on a stage.
```
Same below. Sorry for not being clear.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +476,20 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+
+    def _upload_local_file_to_stage(self, path: str, format: str = "csv") -> str:
+        temp_stage = self._session.get_session_stage()
+        if os.path.exists(path) and os.path.isfile(path):
+            _ = self._session.file.put(path, temp_stage, auto_compress=False)
+            _, filename = os.path.split(path)
+            path = os.path.join(temp_stage, filename)
+        if os.path.exists(path) and os.path.isdir(path):
+            filelist = sorted(os.listdir(path))
+            print(filelist)
```
sfc-gh-jdu, line 488: remove print

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +476,20 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str = "csv") -> str:
+        temp_stage = self._session.get_session_stage()
+        if os.path.exists(path) and os.path.isfile(path):
+            _ = self._session.file.put(path, temp_stage, auto_compress=False)
+            _, filename = os.path.split(path)
+            path = os.path.join(temp_stage, filename)
+        if os.path.exists(path) and os.path.isdir(path):
+            filelist = sorted(os.listdir(path))
```
sfc-gh-jdu, line 487: I don't think we need sorting?

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +332,51 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+        res.sort(key=lambda x: x[0])
+        assert res == [
+            Row(1, "one", 1.2),
+            Row(2, "two", 2.2),
+        ]
+    finally:
+        pass
+
+
+# @pytest.mark.parametrize("mode", ["select", "copy"])
```
sfc-gh-jdu, line 358: I think you need this one instead of the one above, right?

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +332,51 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+#     reader = get_reader(session, mode)
+#
+#     try:
+#         df = (
+#             reader.schema(user_schema)
+#                 .option("compression", "auto")
+#                 .csv(test_files)
+#         )
+#         res = df.collect()
+#         res.sort(key=lambda x: x[0])
```
sfc-gh-jdu, line 371: You don't need sorting here, right? If you need, `Utils.check_answer` provides such an argument.

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +332,51 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+#
+#     reader = get_reader(session, mode)
+#
+#     try:
+#         df = (
+#             reader.schema(user_schema)
+#                 .option("compression", "auto")
+#                 .csv(test_files)
+#         )
+#         res = df.collect()
```
sfc-gh-jdu, line 370: You also don't need `collect()`, you can directly do `Utils.check_answer(df, ...)`. Same below.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +476,20 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str = "csv") -> str:
+        temp_stage = self._session.get_session_stage()
+        if os.path.exists(path) and os.path.isfile(path):
+            _ = self._session.file.put(path, temp_stage, auto_compress=False)
+            _, filename = os.path.split(path)
+            path = os.path.join(temp_stage, filename)
+        if os.path.exists(path) and os.path.isdir(path):
```
sfc-gh-jdu, line 486: hmm I just realized, we only support csv file here, right? So it should be 
```suggestion
        if os.path.exists(path) and os.path.isdir(path):
            if format != "csv":
                raise ValueError(f"Only support a local directory for reading CSV files, but got {format}")
```

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +476,20 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str = "csv") -> str:
```
sfc-gh-jdu, line 480:
```suggestion
    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
```

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -248,11 +268,13 @@ def csv(self, path: str) -> DataFrame:
...
 
         Args:
-            path: The stage location of a CSV file, or a stage location that has CSV files.
+            path: A path pointing to a local CSV file, a local directory that has CSV files, a CSV file on a stage, or
+            a stage location that has CSV files.
 
         Returns:
             a :class:`DataFrame` that is set up to load data from the specified CSV file(s) in a Snowflake stage.
         """
+        path = self._upload_local_file_to_stage(path)
```
sfc-gh-jdu, line 277:
```suggestion
        path = self._upload_local_file_to_stage(path, "csv")
```

tests/integ/scala/test_dataframe_reader_suite.py
sfc-gh-jdu: still^

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +333,83 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+                Row(2, "two", 2.2),
+            ],
+        )
+    finally:
+        pass
+
+
+@pytest.mark.parametrize("mode", ["select", "copy"])
+def test_to_read_csv_from_local_directory(session, resources_path, mode, tmpdir):
+    # test_files = TestFiles(resources_path)
```
sfc-gh-jdu, line 363: remove this

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +333,83 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+                Row(1, "one", 1.2),
+                Row(2, "two", 2.2),
+            ],
+        )
+    finally:
+        pass
+
+
+@pytest.mark.parametrize("mode", ["select", "copy"])
+def test_to_read_csv_from_local_directory(session, resources_path, mode, tmpdir):
```
sfc-gh-jdu, line 362: you don't need `resources_path` anymore

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +333,83 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+            StructField("b", StringType()),
+            StructField("c", StringType()),
+        ]
+    )
+    reader = get_reader(session, mode)
+
+    try:
+        df = reader.schema(schema).option("compression", "auto").csv(tmpdir.strpath)
+        res = df.collect()
+        # because result of read from a dir is random, we check the length of the result here
```
sfc-gh-jdu, line 382:
```suggestion
        # because the result of reading from a dir is nondeterministic, we check the length of the result here
```

sfc-gh-jdu (approved):
looks good! only some minor comments

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -248,11 +264,12 @@ def csv(self, path: str) -> DataFrame:
         """Specify the path of the CSV file(s) to load.
 
         Args:
-            path: The stage location of a CSV file, or a stage location that has CSV files.
+            path: A path pointing to a local CSV file, a local directory that has CSV files, a CSV file on a stage, or a stage location that has CSV files.
```
sfc-gh-sfan, line 267: Given that this now has four possibilities, I'm wondering if we should write it like:

```
path: A path pointing to one of the following:
    - a local CSV file
    - a local directory that has CSV files
    - a CSV file on a stage
    - a stage location that has CSV files
```

cc @sfc-gh-mabrennan

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +471,33 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
+        ostype = os.name
+        temp_stage = self._session.get_session_stage()
+        if os.path.exists(path) and os.path.isfile(path):
+            _ = self._session.file.put(path, temp_stage, auto_compress=False)
```
sfc-gh-sfan, line 479: nit: We don't need `_ =` I think?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +471,33 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+                )
+            filelist = os.listdir(path)
+            for file in filelist:
+                if os.path.splitext(file)[-1][1:].lower() == format.lower():
+                    filepath = os.path.join(path, file)
+                    if ostype == "nt":
+                        filepath = filepath.replace("/", "\\")
+                    else:
+                        filepath = filepath.replace("\\", "/")
+                    _ = self._session.file.put(
```
sfc-gh-sfan, line 499: ditto

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +471,33 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
+        ostype = os.name
+        temp_stage = self._session.get_session_stage()
+        if os.path.exists(path) and os.path.isfile(path):
```
sfc-gh-sfan, line 478: Should we also check if the file is the specified format?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +471,33 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
```
sfc-gh-sfan, line 475: I'm thinking if the following flow makes sense:

```
if path does not exist:
    do we want to do something?
filelist = []
if path is file:
    filelist.append(path)
    stage_path = stage join path
elif path is directory:
    filelist = listdir(path)
    stage_path = stage

for file in filelist:
    do path transformation and put

return stage_path
```

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +471,33 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+        ostype = os.name
+        temp_stage = self._session.get_session_stage()
+        if os.path.exists(path) and os.path.isfile(path):
+            _ = self._session.file.put(path, temp_stage, auto_compress=False)
+            _, filename = os.path.split(path)
+            path = os.path.join(temp_stage, filename)
```
sfc-gh-sfan, line 481: We should probably create a `stage_path` variable for this. Similar to directory one

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mabrennan: I like it either way so I'll let you decide.

sfc-gh-mabrennan (commented):
Doc strings look great. Thank you.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: already deleted, this is just trying to fix a bug before

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: OK, I will create a stage_path variable at the beginning of the fucntion

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: ok, i will add a valueerror exception if the path does not exist

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: I think its ok to not check the format? The format should have been checked in other functions

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: I think its ok to remain the same? Cause in the docs the length of this line looks just fine

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -438,6 +539,40 @@ def test_read_json_with_no_schema(session, mode):
...
 
 
+@pytest.mark.parametrize("mode", ["select", "copy"])
+def test_to_read_json_from_local(session, resources_path, mode):
+
+    test_files = TestFiles(resources_path)
+    reader = get_reader(session, mode)
+
+    df = reader.json(Utils.escape_path(test_files.test_file_json))
+    res = df.collect()
```
sfc-gh-jdu, line 549: no need to `collect()`. same below. (Just to make the code clean)

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: I think we do ignore and do nothing if the path does not exist because when you want to access a file on stage, whose path does not exist in local machine, you dont need to do anything else then the previous code do

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -332,6 +332,107 @@ def test_to_read_files_from_stage(session, resources_path, mode):
...
+    schema = StructType(
+        [
+            StructField("a", StringType()),
+            StructField("b", StringType()),
+            StructField("c", StringType()),
+        ]
+    )
+    reader = get_reader(session, mode)
+
+    df = (
```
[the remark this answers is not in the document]
sfc-gh-yuwang: fixed

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +471,30 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
+        temp_stage = self._session.get_session_stage()
+        stage_path = path
+        if os.path.exists(path):
+            raise ValueError(f"Path of local {format} file dose not exist")
```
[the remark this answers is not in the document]
sfc-gh-yuwang: I removed this because it conflicts previous tests

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-jdu: hmm I think it'd better to raise a ValueError here if the local file doesn't exist. Otherwise the error message the user will get is misleading saying stage file doesn't exist, but actually, the local file doesn't exist. 

You can just check whether it's remote file by:
```
from snowflake.snowpark._internal.utils import STAGE_PREFIX
if path.startswith(STAGE_PREFIX):
    ...
```

Also, you should add a negative test to test when
1) the input is a local file and the file doesn't exist
2) the input is a local directory, and we want to read parquet files.

sfc-gh-sfan (approved):
LGTM

sfc-gh-jdu (approved):

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +472,30 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
+        temp_stage = self._session.get_session_stage()
+        stage_path = path
+        if not path.startswith(STAGE_PREFIX) and not os.path.exists(path):
+            raise ValueError(f"The local file {path} does not exist")
+        else:
```
sfc-gh-yixie, line 481: Nit: no need to have `else` since the `if` is to validate input.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +472,30 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
+        temp_stage = self._session.get_session_stage()
+        stage_path = path
```
sfc-gh-yixie, line 478: Nit: better move these two lines to after input validation.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +472,30 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+        if not path.startswith(STAGE_PREFIX) and not os.path.exists(path):
+            raise ValueError(f"The local file {path} does not exist")
+        else:
+            if os.path.isfile(path):
+                self._session.file.put(
+                    path, temp_stage, auto_compress=False, overwrite=True
+                )
+                _, filename = os.path.split(path)
+                stage_path = f"{temp_stage}/{filename}"
+            if os.path.isdir(path):
```
sfc-gh-yixie, line 488: What will happen if it's a symlink on linux?

sfc-gh-yixie: `if/esif/else` instead of all `if`s is slightly better.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +472,30 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+                    raise ValueError(
+                        f"Only support a local directory for reading CSV files, but got {format}"
+                    )
+                filelist = os.listdir(path)
+                for file in filelist:
+                    if os.path.splitext(file)[-1][1:].lower() == format.lower():
+                        filepath = os.path.join(path, file)
+                        self._session.file.put(
+                            filepath, temp_stage, auto_compress=False, overwrite=True
+                        )
```
sfc-gh-yixie, line 499: `file.put` supports pattern. Can we use that? I guess the performance is better than uploading files one by one.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +472,30 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+            if os.path.isfile(path):
+                self._session.file.put(
+                    path, temp_stage, auto_compress=False, overwrite=True
+                )
+                _, filename = os.path.split(path)
+                stage_path = f"{temp_stage}/{filename}"
+            if os.path.isdir(path):
+                if format != "csv":
+                    raise ValueError(
+                        f"Only support a local directory for reading CSV files, but got {format}"
```
sfc-gh-yixie, line 491: Is it verified that a stage location that has multiple non-CSV files can't be loaded to a table?
cc @sfc-gh-jdu

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: I will revise this

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: I have a misunderstanding about previous docs, it actually support non-CSV files

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: Added a exception for symlink

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: Revised

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: revised

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -452,3 +472,27 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
         )
         df._reader = self
         return df
+
+    def _upload_local_file_to_stage(self, path: str, format: str) -> str:
+
+        if not path.startswith(STAGE_PREFIX) and not os.path.exists(path):
+            raise ValueError(f"The local file {path} does not exist")
+        temp_stage = self._session.get_session_stage()
```
sfc-gh-yixie, line 480: Once a file is uploaded and read into a table, it becomes unuseful (trash in the session stage). 
If `session.read` read data from the same stage path the next time, will it read the trash files so the result data could be wrong? 
In a production env, a session may be alive for long time. All these trash will be kept. So they should be dropped after they're used. To avoid this, I suggest:
1. Use a random prefix after the session stage each time when this API is called.
2. Remove the trashes by the end of read operations.

sfc-gh-jdu: No, if no option is specifies in `sesison.read`, we will use select query from the file directly, e.g.:
```
SELECT $1::INT AS "A", $2::STRING AS "B", $3::DOUBLE AS "C" FROM @SNOWPARK_TEMP_STAGE_BF5CP972WY/testCSV.csv( FILE_FORMAT  => '"TESTDB_SNOWPARK_PYTHON"."TESTSCHEMA_SNOWPARK_PYTHON".SNOWPARK_TEMP_FILE_FORMAT_KW11239MUH')
```
If we don't cache the result first, we will always need this file on the stage. 

Also, if the users want to remove this file after uploading, they can set `PURGE` to true in options. https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#copy-options-copyoptions

sfc-gh-yixie: For user's own files on stage, users can decide to use "PURGE" or not. For these auto-uploaded files for loading into a table, I feel it's better that they're automatically purged by default without user intervention.

PR #374, opened 2022-06-09
SNOW-606808: Move blocklist of SP regression test to snowpark 

sfc-gh-jdu:
In this way, we can easily fix/disable the test without committing to snowflake repo. Also, I enabled several test which is working in SP regression test now:
- test_add_import_stage_file (SNOW-539576 is done)
- test_limit_on_order_by (fixture is added, but this test will be skipped since sample data is not available)
- test_add_import_package (this doesn't not require the specific package on the server)

Please also review other disabled tests (which probably can be enabled)!

Regression test:
https://buwa.c1.us-west-2.aws-dev.app.snowflake.com/builds/precommit/ci73.int.snowflakecomputing.com/8080 (the only failed test is caused by the new commit https://github.com/snowflakedb/snowpark-python/commit/1604b6573da36445489c05dac2bbf10bf0c5b4b4)

### DIFF
--- a/tests/integ/scala/test_result_attributes_suite.py
+++ b/tests/integ/scala/test_result_attributes_suite.py
@@ -4,6 +4,8 @@
 #
 from typing import List
 
+import pytest
+
 from snowflake.snowpark import Session
 from snowflake.snowpark._internal.analyzer.expression import Attribute
 from snowflake.snowpark._internal.utils import TempObjectType
@@ -20,7 +22,7 @@ from snowflake.snowpark.types import (
     TimeType,
     VariantType,
 )
-from tests.utils import Utils as utils
+from tests.utils import IS_IN_STORED_PROC, Utils
 
 
 def get_table_attributes(session: Session, name: str) -> List[Attribute]:
@@ -32,16 +34,16 @@ def get_attributes_with_types(
 ) -> List[Attribute]:
     try:
         schema = ",".join([f"col_{idx} {tp}" for idx, tp in enumerate(types)])
-        utils.create_table(session, name, schema)
+        Utils.create_table(session, name, schema)
         attributes = get_table_attributes(session, name)
     finally:
-        utils.drop_table(session, name)
+        Utils.drop_table(session, name)
     return attributes
 
 
 def test_integer_type(session):
     integers = ["number", "decimal", "numeric", "bigint", "int", "integer", "smallint"]
-    table_name = utils.random_name_for_temp_object(TempObjectType.TABLE)
+    table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
     attributes = get_attributes_with_types(session, table_name, integers)
     for attribute in attributes:
         assert type(attribute.datatype) == LongType
@@ -49,7 +51,7 @@ def test_integer_type(session):
 
 def test_float_type(session):
     floats = ["float", "float4", "double", "real"]
-    table_name = utils.random_name_for_temp_object(TempObjectType.TABLE)
+    table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
     attributes = get_attributes_with_types(session, table_name, floats)
     for attribute in attributes:
         assert type(attribute.datatype) == DoubleType
@@ -57,7 +59,7 @@ def test_float_type(session):
 
 def test_string_type(session):
     strings = ["varchar", "char", "character", "string", "text"]
-    table_name = utils.random_name_for_temp_object(TempObjectType.TABLE)
+    table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
     attributes = get_attributes_with_types(session, table_name, strings)
     for attribute in attributes:
         assert type(attribute.datatype) == StringType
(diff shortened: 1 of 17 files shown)


tests/integ/test_udf.py
```diff
@@ -545,6 +548,9 @@ def plus_one_month(x):
     session.clear_imports()
 
 
+@pytest.mark.skipif(
+    IS_IN_STORED_PROC, reason="SNOW-609328: support caplog in SP regression test"
```
sfc-gh-jdu, line 552: caplog should work since pytest has been working, but somehow it's still failing. I guess it's related to python udf logging. So I created https://snowflakecomputing.atlassian.net/browse/SNOW-609328 for tracking.

tests/integ/scala/test_sql_suite.py
```diff
@@ -10,9 +10,12 @@
 from snowflake.snowpark.exceptions import SnowparkSQLException
 from snowflake.snowpark.functions import col
 from snowflake.snowpark.types import LongType, StructField, StructType
-from tests.utils import TestFiles, Utils
+from tests.utils import IS_IN_STORED_PROC, TestFiles, Utils
 
 
+@pytest.mark.skipif(
+    IS_IN_STORED_PROC, reason="SNOW-533647: Support statement level parameter"
```
sfc-gh-jdu, line 17: I guess this is already done and we just need to change the tests to avoid using `alter session` to pass in statement level parameter?

sfc-gh-sfan: Yeah we might have already supported this, as long as the statement level parameter used here is allowed in owner's mode. Let's add a TODO and we could look into it later.

tests/integ/test_df_to_pandas.py
```diff
@@ -117,8 +115,7 @@ def check_fetch_data_exception(query: str) -> None:
 
 
 @pytest.mark.skipif(
-    version.parse(snowflake.connector.__version__) < version.parse("2.7.5"),
-    reason="Python connector >= v2.7.5 ignores the index for the dataframe from fetch_pandas_all()",
+    IS_IN_STORED_PROC, reason="SNOW-507565: Need localaws for large result"
```
sfc-gh-sfan, line 118: We might be able to use `IS_IN_STORED_PROC_LOCALFS` here, again, orthogonal to your change and can be looked into separately

tests/integ/test_query_history.py
```diff
@@ -63,6 +65,9 @@ def test_query_history_no_actions(session):
     assert len(query_listener.queries) == 0
 
 
+@pytest.mark.skipif(
+    IS_IN_STORED_PROC, reason="SNOW-533647: Support statement level parameters"
```
sfc-gh-sfan, line 69: If we don't do `alter session` then this might be supported. Maybe we should wait for `executemany` supporting statement level params.

tests/integ/test_udf.py
```diff
@@ -733,6 +739,7 @@ def return_dict(v: dict) -> Dict[str, str]:
     )
 
 
+@pytest.mark.skipif(IS_IN_STORED_PROC, reason="need to support permanent udf")
```
sfc-gh-sfan, line 742: Let's change this to `Cannot create session in SP`. I think the original reason I put there is wrong 🤦

tests/utils.py
```diff
@@ -42,6 +42,9 @@
 IS_MACOS = platform.system() == "Darwin"
 IS_LINUX = platform.system() == "Linux"
 IS_UNIX = IS_LINUX or IS_MACOS
+IS_IN_STORED_PROC = is_in_stored_procedure()
+# this env variable is set in regression test
+IS_IN_STORED_PROC_LOCALFS = IS_IN_STORED_PROC and os.getenv("IS_LOCAL_FS")
```
sfc-gh-sfan, line 47: This is neat :)

sfc-gh-sfan (approved):

tests/integ/scala/test_sql_suite.py
sfc-gh-jdu: Hmm I have tried to enable it, but it failed because `alter session` is not allowed in SP. The correct way is to use https://github.com/snowflakedb/snowflake-connector-python/blob/2ccd3c811958170854ff4cac159e7c125bf8b4f3/src/snowflake/connector/cursor.py#L568. (but there is no such API in snowpark). Maybe our `collect()`, `show()` should support this argument. cc @sfc-gh-yixie

tests/integ/scala/test_result_attributes_suite.py
```diff
@@ -89,27 +91,30 @@ def test_date_and_time_type(session):
...
 def test_array_type(session):
     semi_structures = ["array"]
-    table_name = utils.random_name_for_temp_object(TempObjectType.TABLE)
+    table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
     attributes = get_attributes_with_types(session, table_name, semi_structures)
     assert type(attributes[0].datatype) == ArrayType
 
 
+@pytest.mark.skipif(
+    IS_IN_STORED_PROC, reason="SNOW-507565: fix local_aws reg test environment"
```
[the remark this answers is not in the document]
sfc-gh-jdu: SNOW-507565 shows it's backlog 🤦 And I tried it too but still failed.

tests/integ/test_query_history.py
sfc-gh-jdu: ok

tests/integ/scala/test_sql_suite.py
sfc-gh-jdu: I created https://snowflakecomputing.atlassian.net/browse/SNOW-609875 to track that whether we can enable more tests.

tests/integ/test_udf.py
sfc-gh-jdu: changed

PR #376, opened 2022-06-14
SNOW-607224: "Create or replace" temp table for `create_dataframe` with large local data and reading files from a stage

sfc-gh-jdu:

   Fixes SNOW-607224 by replacing the created temp table. We should avoid using `if not exist` when creating a temp table in case the underlying table is changed, though both should be fine as our dataframe is immutable. Actually not only `create_dataframe`, this PR also fixes this issue when reading files from a stage. 

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -578,7 +578,7 @@ def create_table_statement(
     temp: bool = False,
 ) -> str:
     return (
-        f"{CREATE}{(OR + REPLACE) if replace else EMPTY_STRING} "
+        f"{CREATE}{(OR + REPLACE) if replace else EMPTY_STRING}"
         f"{TEMPORARY if temp else EMPTY_STRING}"
         f"{TABLE}{table_name}{(IF + NOT + EXISTS) if not replace and not error else EMPTY_STRING}"
         f"{LEFT_PARENTHESIS}{schema}{RIGHT_PARENTHESIS}"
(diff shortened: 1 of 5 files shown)


sfc-gh-sfan (commented):
To make sure I understand correctly, we use `replace` to make sure even if we create a temp table twice, it still works, is it correct? The title feels like we are fixing so we only create the table once.

sfc-gh-jdu:
> To make sure I understand correctly, we use `replace` to make sure even if we create a temp table twice, it still works, is it correct? The title feels like we are fixing so we only create the table once.

Yes, fixed the title.

sfc-gh-yixie:
Can we change the query plan so a single `.collect()` call won't create the same table twice?

sfc-gh-yixie:
Can we change the query plan so a single `.collect()` call won't create the same table twice?

sfc-gh-jdu:
> Can we change the query plan so a single `.collect()` call won't create the same table twice?

I've thought about this, but feel it'd a little bit difficult unless we change how the current query plan works. There are two options to create the table only once:
1. Use `if not exist` when creating the table. However, it will cause the insertion happens twice and give rise to wrong results. 
2. Check if the table exists or not before creating a new table and inserting data. This requires the queries in the plan can "conditionally" executed, which is not achievable currently (only sequentially now). It can't be determined when generating query plan because no table will be created at that time.   

So I feel "create or replace" is easiest and most secure way to fix this bug for now. Do you have any other insights?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -578,7 +578,7 @@ def create_table_statement(
     temp: bool = False,
 ) -> str:
     return (
-        f"{CREATE}{(OR + REPLACE) if replace else EMPTY_STRING} "
```
sfc-gh-kwagner, line 581: Is this a functional change or is it just because there is an extra space here?

sfc-gh-jdu: Just because there is an extra space here, which is not needed.

sfc-gh-yixie (approved):

PR #378, opened 2022-06-15
feat: Allow any Python minor release >= 3.8

### LINKED ISSUE #377: SNOW-612394: Support for Python >= 3.9
Install fails on python 3.9

That it works on python 3.9. Our use-case is installing snowpark in the [jupyterhub/docker-stacks](https://github.com/jupyter/docker-stacks) scipy image, which is in fact already on python 3.10.

More flexibility. Is there really such a breaking change in python 3.9 that warrants this oddly specific version pin on 3.8? I suspect just [releasing the pin](https://github.com/snowflakedb/snowpark-python/blob/5ed94f4a2fd99ed510c5792e4f6b0981a174e619/setup.py#L45) would probably be enough, as I'm not aware of any non-compatibilities between 3.8 and >=3.9

dingobar:

   Fixes #377

   We should not require the Python minor version to match exactly `3.8`. Instead, we could state in the docs that the connector is tested on `3.8` only and that it is not guaranteed to work on other versions?

I have tested that the package installs and works just fine on Python `3.10.4` locally.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
         "Issues": "https://github.com/snowflakedb/snowpark-python/issues",
         "Changelog": "https://github.com/snowflakedb/snowpark-python/blob/main/CHANGELOG.md",
     },
-    python_requires="==3.8.*",
+    python_requires=">=3.8",
     install_requires=[
         "setuptools>=40.6.0",
         "wheel",


dingobar:

willingc:
Thanks for the PR @dingobar. I'm another user who would love to see the pin relaxed so 3.9 and 3.10 can be used with snowpark.

PR #389, opened 2022-06-24
SNOW-607203: ml transformers(StandardScaler, MinmaxScaler, OnehotEncoder, OrdinalEncoder, Binarizer)

sfc-gh-yuwang:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -97,6 +97,7 @@ from snowflake.snowpark.functions import (
     row_number,
     sql_expr,
     stddev,
+    stddev_pop,
     to_char,
 )
 from snowflake.snowpark.row import Row
@@ -1101,9 +1102,15 @@ class DataFrame:
 
         This is equivalent to performing a SELECT DISTINCT in SQL.
         """
-        return self.group_by(
-            [self.col(quote_name(f.name)) for f in self.schema.fields]
-        ).agg([])
+        return self._distinct(self.columns)
+
+    def _distinct(self, columns: List[str]) -> "DataFrame":
+        """Internal function that return a new DataFrame that contains only the rows with distinct values
+        from the current DataFrame.
+
+        This is equivalent to performing a SELECT DISTINCT in SQL.
+        """
+        return self.group_by([self.col(quote_name(c)) for c in columns]).agg([])
 
     def drop_duplicates(self, *subset: Union[str, Iterable[str]]) -> "DataFrame":
         """Creates a new DataFrame by removing duplicated rows on given subset of columns.
@@ -2352,6 +2359,13 @@ class DataFrame:
         Args:
             cols: The names of columns whose basic statistics are computed.
         """
+        # These are five stats that pyspark's describe() outputs
+        return self._describe(cols, stats=["count", "mean", "stddev", "min", "max"])
+
+    def _describe(
+        self, *cols: Union[str, List[str]], stats: Optional[List[str]] = None
+    ) -> "DataFrame":
+
         cols = parse_positional_args_to_list(*cols)
         df = self.select(cols) if len(cols) > 0 else self
 
@@ -2362,13 +2376,20 @@ class DataFrame:
             if isinstance(field.datatype, (StringType, _NumericType))
         }
 
-        stat_func_dict = {
+        all_stat_func_dict = {
             "count": count,
             "mean": mean,
             "stddev": stddev,
+            "stddev_pop": stddev_pop,
             "min": min_,
             "max": max_,
         }
+
+        stat_func_dict = (
+            {k: v for k, v in all_stat_func_dict.items() if k in stats}
+            if stats
+            else all_stat_func_dict
+        )
 
         # if no columns should be selected, just return stat names
         if len(numerical_string_col_type_dict) == 0:
--- a/src/snowflake/snowpark/ml/transformer.py
+++ b/src/snowflake/snowpark/ml/transformer.py
@@ -50,12 +50,12 @@ class Transformer:
     def output_cols(self, value: List[str]) -> None:
         self._output_cols = value
 
-    def fit(self, df: DataFrame) -> "Transformer":
+    def fit(self, df: DataFrame) -> "None":
         self._fitted = True
         if not self._input_cols:
-            raise ValueError("Input columns can not be None")
+            raise ValueError("Input column can not be empty")
 
-    def transform(self, df: DataFrame) -> DataFrame:
+    def transform(self, df: DataFrame) -> None:
         if not self.input_cols:
             raise ValueError("Input column can not be empty")
         elif not self.output_cols:
@@ -133,7 +133,7 @@ class MinMaxScaler(Transformer):
         super().__init__(input_cols, output_cols)
         if feature_range[0] >= feature_range[1]:
             raise ValueError(
-                "Maximum  can not be equal or greater than minimum in feature range"
+                "Maximum can not be equal or greater than minimum in feature range"
             )
         self._states_table_cols = (
             [f"states_{input_col}" for input_col in self._input_cols]
@@ -161,15 +161,16 @@ class MinMaxScaler(Transformer):
     def transform(self, df: DataFrame) -> DataFrame:
         """
         The transformation is given by:
-        X_std = (X - X.min(axis=0)) / (X.max(axis=0) - X.min(axis=0))
+        X_std = (X - min(X)) / (max(X) - min(X))
         X_scaled = X_std * (max - min) + min
-        where min, max are decided by feature_range
+        where min, max are provided by feature_range
         """
         super().transform(df)
         states_table = df._session.table(self._states_table_name)
 
         res_column = df.columns
 
+        # if max(X)-min(X) equals to 0, we will set it as 1 to avoid divided by 0 error
         for input_col, states_col, output_col in zip(
             self.input_cols, self._states_table_cols, self.output_cols
         ):
@@ -249,6 +250,10 @@ class OneHotEncoder(Transformer):
         else:
             raise ValueError(
                 "The number of output column and input column does not match"
+            )
+        if not self._fitted:
+            raise ValueError(
+                "The transformer is not fitted yet, call fit() function before transform"
             )
 
         original_columns = df.columns
@@ -429,7 +434,7 @@ class OrdinalEncoder(Transformer):
         df: DataFrame,
     ) -> DataFrame:
         """
-        The ordinal encoder will be "null" if there are new types in dataframe fed into transform compared to dataframe
+        The encoder value will be null if there are unknown types in this dataframe compared to that used in fit()
         fed into fit() function.
         """
         if not self.input_cols:
@@ -444,7 +449,10 @@ class OrdinalEncoder(Transformer):
             raise ValueError(
                 "The number of output column and input column does not match"
             )
-
+        if not self._fitted:
+            raise ValueError(
+                "The transformer is not fitted yet, call fit() function before transform"
+            )
         original_columns = df.columns
         encoder_structure = []
 
@@ -493,7 +501,12 @@ class Binarizer(Transformer):
         super().transform(df)
         for input_col, output_col in zip(self._input_cols, self._output_cols):
             df = df.with_column(
-                output_col, iff(col(input_col) > self._threshold, lit(1), lit(0))
+                output_col,
+                iff(
+                    col(input_col).is_null() | (col(input_col) == float("nan")),
+                    None,
+                    iff(col(input_col) > self._threshold, lit(1), lit(0)),
+                ),
             )
 
         return df
--- a/src/snowflake/snowpark/ml/transformer.py
+++ b/src/snowflake/snowpark/ml/transformer.py
@@ -21,7 +21,11 @@ from snowflake.snowpark.functions import (
     iff,
     lit,
 )
-from snowflake.snowpark.ml.utils import encoder_fit, scaler_fit
+from snowflake.snowpark.ml.utils import (
+    check_if_input_output_match,
+    encoder_fit,
+    scaler_fit,
+)
 
 
 class Transformer:
@@ -60,8 +64,6 @@ class Transformer:
             raise ValueError("Input column can not be empty")
         elif not self.output_cols:
             raise ValueError("Output column can not be empty")
-        elif len(self.input_cols) != len(self.output_cols):
-            raise ValueError("Input column and output column does not match")
         elif not self._fitted:
             raise ValueError(
                 "The transformer is not fitted yet, call fit() function before transform"
@@ -102,6 +104,7 @@ class StandardScaler(Transformer):
         z = (x - mean) / stddev
         """
         super().transform(df)
+        check_if_input_output_match(self)
         states_table = df._session.table(self._states_table_name)
         res_column = df.columns
         for input_col, states_col, output_col in zip(
@@ -166,6 +169,7 @@ class MinMaxScaler(Transformer):
         where min, max are provided by feature_range
         """
         super().transform(df)
+        check_if_input_output_match(self)
         states_table = df._session.table(self._states_table_name)
 
         res_column = df.columns
@@ -239,21 +243,14 @@ class OneHotEncoder(Transformer):
         df: DataFrame,
         sparse: bool = True,
     ) -> DataFrame:
-        if not self.input_cols:
-            raise ValueError("Input column can not be empty")
-        elif not self.output_cols:
-            raise ValueError("Output column can not be empty")
-        elif len(self._output_cols) == 1:
+        super().transform(df)
+        if len(self._output_cols) == 1:
             self._merge = True
         elif len(self._input_cols) == len(self._output_cols):
             self._merge = False
         else:
             raise ValueError(
                 "The number of output column and input column does not match"
-            )
-        if not self._fitted:
-            raise ValueError(
-                "The transformer is not fitted yet, call fit() function before transform"
             )
 
         original_columns = df.columns
@@ -434,24 +431,17 @@ class OrdinalEncoder(Transformer):
         df: DataFrame,
     ) -> DataFrame:
         """
-        The encoder value will be null if there are unknown types in this dataframe compared to that used in fit()
-        fed into fit() function.
+        The encoded value will be null if there are unknown types in this dataframe compared to that dataframe used in
+        fit() function.
         """
-        if not self.input_cols:
-            raise ValueError("Input column can not be empty")
-        elif not self.output_cols:
-            raise ValueError("Output column can not be empty")
-        elif len(self._output_cols) == 1:
+        super().transform(df)
+        if len(self._output_cols) == 1:
             self._merge = True
         elif len(self._input_cols) == len(self._output_cols):
             self._merge = False
         else:
             raise ValueError(
                 "The number of output column and input column does not match"
-            )
-        if not self._fitted:
-            raise ValueError(
-                "The transformer is not fitted yet, call fit() function before transform"
             )
         original_columns = df.columns
         encoder_structure = []
@@ -499,6 +489,7 @@ class Binarizer(Transformer):
 
     def transform(self, df: DataFrame) -> DataFrame:
         super().transform(df)
+        check_if_input_output_match(self)
         for input_col, output_col in zip(self._input_cols, self._output_cols):
             df = df.with_column(
                 output_col,
--- a/src/snowflake/snowpark/ml/transformer.py
+++ b/src/snowflake/snowpark/ml/transformer.py
@@ -1,0 +1,499 @@
+#
+# Copyright (c) 2012-2022 Snowflake Computing Inc. All rights reserved.
+#
+
+
+from typing import List, Optional, Tuple, Union
+
+from snowflake.snowpark._internal.utils import (
+    TempObjectType,
+    random_name_for_temp_object,
+)
+from snowflake.snowpark.dataframe import DataFrame
+from snowflake.snowpark.functions import (
+    array_agg,
+    array_cat,
+    array_construct,
+    array_construct_compact,
+    array_insert,
+    array_slice,
+    col,
+    iff,
+    lit,
+)
+from snowflake.snowpark.ml.utils import encoder_fit, scaler_fit
+
+
+class Transformer:
+    def __init__(
+        self,
+        input_cols: Optional[List[str]] = None,
+        output_cols: Optional[List[str]] = None,
+    ) -> None:
+        self._input_cols = input_cols
+        self._output_cols = output_cols
+        self._fitted = False
+
+    @property
+    def input_cols(self) -> List[str]:
+        return self._input_cols
+
+    @input_cols.setter
+    def input_cols(self, value: List[str]) -> None:
+        self._input_cols = value
+
+    @property
+    def output_cols(self) -> List[str]:
+        return self._output_cols
+
+    @output_cols.setter
+    def output_cols(self, value: List[str]) -> None:
+        self._output_cols = value
+
+    def fit(self, df: DataFrame) -> "Transformer":
+        self._fitted = True
+        if not self._input_cols:
+            raise ValueError("Input columns can not be None")
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        if not self.input_cols:
+            raise ValueError("Input column can not be empty")
+        elif not self.output_cols:
+            raise ValueError("Output column can not be empty")
+        elif len(self.input_cols) != len(self.output_cols):
+            raise ValueError("Input column and output column does not match")
+        elif not self._fitted:
+            raise ValueError(
+                "The transformer is not fitted yet, call fit() function before transform"
+            )
+
+
+class StandardScaler(Transformer):
+    def __init__(
+        self,
+        input_cols: Optional[List[str]] = None,
+        output_cols: Optional[List[str]] = None,
+    ) -> None:
+        super().__init__(input_cols, output_cols)
+        self._states_table_cols = (
+            [f"states_{input_col}" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+        self._states_table_name = random_name_for_temp_object(TempObjectType.TABLE)
+
+    @Transformer.input_cols.setter
+    def input_cols(self, value: List[str]) -> None:
+        self._input_cols = value
+        self._states_table_cols = (
+            [f"states_{input_col}" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+
+    def fit(self, df: DataFrame) -> "StandardScaler":
+        super().fit(df)
+        scaler_fit(self, df, ["mean", "stddev_pop"])
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """
+        The standard score of a sample x is calculated as:
+        z = (x - mean) / stddev
+        """
+        super().transform(df)
+        states_table = df._session.table(self._states_table_name)
+        res_column = df.columns
+        for input_col, states_col, output_col in zip(
+            self.input_cols, self._states_table_cols, self.output_cols
+        ):
+            # if stddev equals to 0, we will set it to 1 to avoid divided by 0 error
+            res_column.append(
+                (
+                    (df[input_col] - col(states_col)["mean"])
+                    / iff(
+                        col(states_col)["stddev_pop"] == 0,
+                        1,
+                        col(states_col)["stddev_pop"],
+                    )
+                ).as_(output_col)
+            )
+
+        df_transform = states_table.join(df).select(res_column)
+        return df_transform
+
+
+class MinMaxScaler(Transformer):
+    def __init__(
+        self,
+        feature_range: Tuple[float, float] = (0, 1),
+        input_cols: Optional[List[str]] = None,
+        output_cols: Optional[List[str]] = None,
+    ) -> None:
+        super().__init__(input_cols, output_cols)
+        if feature_range[0] >= feature_range[1]:
+            raise ValueError(
+                "Maximum  can not be equal or greater than minimum in feature range"
+            )
+        self._states_table_cols = (
+            [f"states_{input_col}" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+        self.minimum, self.maximum = feature_range
+        self._states_table_name = random_name_for_temp_object(TempObjectType.TABLE)
+
+    @Transformer.input_cols.setter
+    def input_cols(self, value: List[str]) -> None:
+        self._input_cols = value
+        self._states_table_cols = (
+            [f"states_{input_col}" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+
+    def fit(self, df: DataFrame) -> "MinMaxScaler":
+        super().fit(df)
+
+        scaler_fit(self, df, ["min", "max"])
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """
+        The transformation is given by:
+        X_std = (X - X.min(axis=0)) / (X.max(axis=0) - X.min(axis=0))
+        X_scaled = X_std * (max - min) + min
+        where min, max are decided by feature_range
+        """
+        super().transform(df)
+        states_table = df._session.table(self._states_table_name)
+
+        res_column = df.columns
+
+        for input_col, states_col, output_col in zip(
+            self.input_cols, self._states_table_cols, self.output_cols
+        ):
+            res_column.append(
+                (
+                    (df[input_col] - col(states_col)["min"])
+                    / iff(
+                        (col(states_col)["max"] - col(states_col)["min"]) == 0,
+                        1,
+                        col(states_col)["max"] - col(states_col)["min"],
+                    )
+                    * (self.maximum - self.minimum)
+                    + self.minimum
+                ).as_(output_col)
+            )
+
+        df_transform = states_table.join(df).select(res_column)
+        return df_transform
+
+
+class OneHotEncoder(Transformer):
+    def __init__(
+        self,
+        input_cols: Optional[List[str]] = None,
+        output_cols: Optional[List[str]] = None,
+        category: Union[str, List[List[str]]] = "auto",
+    ) -> None:
+        super().__init__(input_cols, output_cols)
+        self._category = category
+        self._merge = False
+        self._states_table_cols = (
+            [f"states_{input_col}" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+        self._encoder_table_cols = (
+            [f"states_{input_col}_encoder" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+        self._category_table_name = {}
+        self._encoder_count_table = random_name_for_temp_object(TempObjectType.TABLE)
+
+    @Transformer.input_cols.setter
+    def input_cols(self, value: List[str]) -> None:
+        self._input_cols = value
+        self._states_table_cols = (
+            [f"states_{input_col}" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+        self._encoder_table_cols = (
+            [f"states_{input_col}_encoder" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+
+    def fit(self, df: DataFrame) -> "OneHotEncoder":
+        super().fit(df)
+
+        encoder_fit(self, df, "onehot")
+        return self
+
+    def transform(
+        self,
+        df: DataFrame,
+        sparse: bool = True,
+    ) -> DataFrame:
+        if not self.input_cols:
+            raise ValueError("Input column can not be empty")
+        elif not self.output_cols:
+            raise ValueError("Output column can not be empty")
+        elif len(self._output_cols) == 1:
+            self._merge = True
+        elif len(self._input_cols) == len(self._output_cols):
+            self._merge = False
+        else:
+            raise ValueError(
+                "The number of output column and input column does not match"
+            )
+
+        original_columns = df.columns
+        encoder_structure = [] if sparse else array_construct()
+        encoder_offset = lit(0)
+        df_count_table = df._session.table(self._encoder_count_table)
+        df = df.join(df_count_table)
+
+        for states_col, encoder_col, (input_col, table_name) in zip(
+            self._states_table_cols,
+            self._encoder_table_cols,
+            self._category_table_name.items(),
+        ):
+            category_table = df._session.table(table_name)
+            # if sparse is set to true, convert category_table so that it stores onehot array
+            if not sparse:
+                # create a [0,0...0,0] array as the base of sparse encoder array
+                df_dense_array = category_table.with_column("dense", lit(0)).select(
+                    [array_agg(col("dense")).as_(f"{states_col}_dense")]
+                )
+                # join tables to get the length of each encoder array, STATES_X_DENSE looks like [0,...,0], length of
+                # this array depends on how many categories in this column
+                df_dense_array = category_table.join(df_dense_array).join(
+                    df_count_table
+                )
+                """
+                df_dense_array looks like:
+                -------------------------------------------------------------------------------------------------
+                |"STATES_A"  |"STATES_A_ENCODER"  |"STATES_A_DENSE"  |"STATES_A_COUNT"  |... |"ENCODER_COUNT"  |
+                -------------------------------------------------------------------------------------------------
+                |1           |0                   |[                  |2                 |... |6                |
+                |            |                    |  0,               |                  |    |                 |
+                |            |                    |  0                |                  |    |                 |
+                |            |                    |]                  |                  |    |                 |
+                |2           |1                   |[                  |2                 |... |6                |
+                |            |                    |  0,               |                  |    |                 |
+                |            |                    |  0                |                  |    |                 |
+                |            |                    |]                  |                  |    |                 |
+                -------------------------------------------------------------------------------------------------
+                """
+                # create sparse encoder array for each type in each category
+                category_table = df_dense_array.select(
+                    [
+                        col(states_col),
+                        array_slice(
+                            array_insert(
+                                col(f"{states_col}_dense"), col(encoder_col), lit(1)
+                            ),
+                            lit(0),
+                            col(f"{states_col}_count"),
+                        ).as_(encoder_col),
+                    ]
+                )
+                """
+                category_table looks like:
+                -----------------------------------
+                |"STATES_A"  |"STATES_A_ENCODER"  |
+                -----------------------------------
+                |1           |[                   |
+                |            |  1,                |
+                |            |  0                 |
+                |            |]                   |
+                |2           |[                   |
+                |            |  0,                |
+                |            |  1                 |
+                |            |]                   |
+                -----------------------------------
+                """
+                df = category_table.join(
+                    df, df[input_col] == category_table[states_col], join_type="right"
+                ).join(df_dense_array.select([col(f"{states_col}_dense")]).limit(1))
+            else:
+                df = category_table.join(
+                    df, df[input_col] == category_table[states_col], join_type="right"
+                )
+            if self._merge:
+                if sparse:
+                    # encoder_offset is used to shift from one category to another category
+                    encoder_structure.append(
+                        iff(
+                            col(encoder_col).is_null(),
+                            col(encoder_col),
+                            col(encoder_col) + encoder_offset,
+                        )
+                    )
+                    encoder_offset += col(f"{states_col}_count")
+                else:
+                    encoder_structure = array_cat(
+                        encoder_structure,
+                        iff(
+                            col(encoder_col).is_null(),
+                            df[f"{states_col}_dense"],
+                            col(encoder_col),
+                        ),
+                    )
+        # depending on self._merge and sparse, prepare the final output
+        if self._merge:
+            if sparse:
+                # null value will be ignored here
+                encoder_structure = array_construct_compact(*encoder_structure)
+                onehot_array = array_construct(
+                    df["encoder_count"], encoder_structure, lit(1)
+                )
+            else:
+                onehot_array = encoder_structure
+
+            original_columns.append(self._output_cols[0])
+            df = df.with_column(self._output_cols[0], onehot_array)
+            df = df.select([col(input_col) for input_col in original_columns])
+        else:
+            onehot_arrays = []
+            for states_col, output_col in zip(
+                self._states_table_cols, self._output_cols
+            ):
+                if sparse:
+                    onehot_array = array_construct_compact(
+                        df[f"{states_col}_count"], df[f"{states_col}_encoder"], lit(1)
+                    )
+                else:
+                    # df[f"{states_col}_encoder"] is null if df fed into transform has unknown category compare
+                    # to category_table because of right join
+                    onehot_array = iff(
+                        df[f"{states_col}_encoder"].is_null(),
+                        df[f"{states_col}_dense"],
+                        df[f"{states_col}_encoder"],
+                    )
+                onehot_arrays.append(onehot_array)
+                original_columns.append(output_col)
+            df = df.with_columns(self._output_cols, onehot_arrays)
+            df = df.select([col(input_col) for input_col in original_columns])
+        return df
+
+
+class OrdinalEncoder(Transformer):
+    def __init__(
+        self,
+        input_cols: Optional[List[str]] = None,
+        output_cols: Optional[List[str]] = None,
+        category: Union[str, List[List[str]]] = "auto",
+    ) -> None:
+        super().__init__(input_cols, output_cols)
+        self._category = category
+        self._merge = False
+        self._states_table_cols = (
+            [f"states_{input_col}" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+        self._encoder_table_cols = (
+            [f"states_{input_col}_encoder" for input_col in self._input_cols]
+            if self._input_cols
+            else None
+        )
+        self._category_table_name = {}
+
+    @Transformer.input_cols.setter
+    def input_cols(self, value: List[str]) -> None:
+        self._input_cols = value
+        self._states_table_cols = (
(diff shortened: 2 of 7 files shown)


src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,360 @@
...
+        originalC = df.columns
+        encoder_structure = array_construct()
+        for fc, ec, (c, tabname) in zip(
+            self._states_table_cols,
+            self._encoder_table_cols,
+            self._category_table_name.items(),
+        ):
+            cattable = df._session.table(tabname)
+            df = cattable.join(df, df[c] == cattable[fc])
+            encoder_structure = array_append(encoder_structure, col(ec))
```
sfc-gh-jdu, line 349: encoder_structure is only used when `self.merge` is true, right? So you should only do it when `self.merge` is true.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,360 @@
...
+            df = cattable.join(df, df[c] == cattable[fc])
+            if not sparse:
+                encoder_structure = array_append(
+                    encoder_structure, col(ec) + encoder_offset
+                )
+                encoder_offset += col(f"{fc}_count")
+            else:
+                encoder_structure = array_cat(encoder_structure, col(ec))
+        if not sparse:
+            onehotarray = array_construct(
```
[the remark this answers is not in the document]
sfc-gh-jdu: If self.merge is False, you don't need to compute onehotarray, right? onehotarray is re-computed below without using `encoder_structure`.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,360 @@
...
+                    ]
+                )
+            df = cattable.join(df, df[c] == cattable[fc])
+            if not sparse:
+                encoder_structure = array_append(
+                    encoder_structure, col(ec) + encoder_offset
+                )
+                encoder_offset += col(f"{fc}_count")
+            else:
+                encoder_structure = array_cat(encoder_structure, col(ec))
```
sfc-gh-jdu, line 255: Why do you use `array_cat` here?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,360 @@
...
+                        array_slice(
+                            array_insert(col(f"{fc}_sparse"), col(ec), lit(1)),
+                            lit(0),
+                            col(f"{fc}_count"),
+                        ).as_(ec),
+                    ]
+                )
+            df = cattable.join(df, df[c] == cattable[fc])
+            if not sparse:
+                encoder_structure = array_append(
```
sfc-gh-jdu, line 250: I would avoid using array_append to append every column, you can have append these columns to a python list first, then use `array_construct`?

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,286 @@
...
+    minmaxscaler.output_cols = ["res"]
+    model = minmaxscaler.fit(df)
+    res = model.transform(df)
+    Utils.check_answer(
+        res,
+        [Row(0.0), Row(0.3333333333333333), Row(1.0)],
+    )
+
+
+def test_merge_onehotencoder(session):
```
sfc-gh-jdu, line 164: You should have a test to test `sparse` argument.

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -123,7 +123,7 @@ def copy_into_location(
         file_format_type: Optional[str] = None,
         format_type_options: Optional[Dict[str, str]] = None,
         header: bool = False,
-        **copy_options: Optional[str],
+        **copy_options: Optional[Union[str, bool]],
```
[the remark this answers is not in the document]
sfc-gh-yuwang: this argument is also used in following code on line 181, should I remove that too?

src/snowflake/snowpark/ml/transformer.py
sfc-gh-yuwang: I'd love to do that but array_construct does not accept python list as an arg. I'll see if I can do it in other ways

src/snowflake/snowpark/ml/transformer.py
sfc-gh-yuwang: if customer want the sparse form, col(ec) here would be a one hot array of the corresponding input col and encoder_structure is only used when you want to merge it.

src/snowflake/snowpark/ml/transformer.py
sfc-gh-jdu: you can do `array_construct(*l)`, where l is a python list.

sfc-gh-yuwang: got it!

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,88 @@
...
+            transformer._category,
+        ):
+            transformer._category_table_name[c] = random_name_for_temp_object(
+                TempObjectType.TABLE
+            )
+            df_save = session.create_dataframe(cat, schema=[fc])
+            if encoder_type == "onehot":
+                df_count[f"{fc}_count"] = len(cat)
+                df_encoder_length += len(cat)
+            df_save.with_column(f"{fc}_encoder", builtin("seq8")()).write.save_as_table(
```
sfc-gh-jdu, line 79: add a comment about what this table looks like

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,88 @@
...
+                transformer._category_table_name[c], create_temp_table=True
+            )
+        df_count = session.create_dataframe([df_count])
+        if encoder_type == "onehot":
+            df_count.with_column(
+                "encoder_count", lit(df_encoder_length)
+            ).write.save_as_table(
```
sfc-gh-jdu, line 86: add a comment about what this table looks like

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,88 @@
...
+    transformer: "snowflake.snowpark.ml.Transformer", df: DataFrame, stats: [str]
+):
+    if len(transformer.input_cols) != len(transformer.output_cols):
+        raise ValueError("The number of input column and output column does not match")
+    df._describe(transformer.input_cols, stats=stats).select(
+        [
+            object_agg("summary", c).as_(fc)
+            for c, fc in zip(transformer.input_cols, transformer._states_table_cols)
+        ]
+    ).write.save_as_table(transformer._states_table_name, create_temp_table=True)
```
sfc-gh-jdu, line 24: add a comment about what this table looks like

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,356 @@
...
+        scaler_fit(self, df, ["mean", "stddev_pop"])
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        states_table = df._session.table(self._states_table_name)
+        return (
+            states_table.join(df)
+            .select(
+                [
+                    (df[c] - col(fc)["mean"]) / col(fc)["stddev_pop"]
```
sfc-gh-jdu, line 86: let's write down the math in comment here?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,356 @@
...
+        scaler_fit(self, df, ["min", "max"])
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        states_table = df._session.table(self._states_table_name)
+        return (
+            states_table.join(df)
+            .select(
+                [
+                    (df[c] - col(fc)["min"])
```
sfc-gh-jdu, line 125: let's write down the math in comment here?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,356 @@
...
+            [f"states_{c}" for c in self._input_cols] if self._input_cols else None
+        )
+        self._encoder_table_cols = (
+            [f"states_{c}_encoder" for c in self._input_cols]
+            if self._input_cols
+            else None
+        )
+
+    def fit(self, df: DataFrame) -> "OneHotEncoder":
+        if not self._input_cols:
```
sfc-gh-jdu, line 170: If you check this for encoder, you can also check it for scaler, right?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,356 @@
...
+        elif len(self._output_cols) == 1:
+            self._merge = True
+        elif len(self._input_cols) == len(self._output_cols):
+            self._merge = False
+        else:
+            raise ValueError(
+                "The number of output colum and input column does not match"
+            )
+
+        original_cloumns = df.columns
```
sfc-gh-jdu, line 193: typo

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,290 @@
...
+    MinMaxScaler,
+    OneHotEncoder,
+    OrdinalEncoder,
+    StandardScaler,
+)
+from snowflake.snowpark.types import IntegerType, StructField, StructType
+from tests.utils import Utils
+
+
+# In the tests below, we test both scenarios: SELECT & COPY
```
sfc-gh-jdu, line 19: remove these that are not used in the test (line 19-51)

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    This table store the value required by standard scaler or minmax scaler. stddev, mean and minimum, maximum to be
```
sfc-gh-jdu, line 27:
```suggestion
    This table stores the value required by standard scaler or minmax scaler. stddev, mean and minimum, maximum to be
```

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,140 @@
...
+):
+    if len(transformer.input_cols) != len(transformer.output_cols):
+        raise ValueError("The number of input column and output column does not match")
+    df._describe(transformer.input_cols, stats=stats).select(
+        [
+            object_agg("summary", c).as_(fc)
+            for c, fc in zip(transformer.input_cols, transformer._states_table_cols)
+        ]
+    ).write.save_as_table(transformer._states_table_name, create_temp_table=True)
+    """
```
sfc-gh-jdu, line 26: Move this docstring before the implementation? line 18 here

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        if encoder_type == "onehot":
+            df_count = session.create_dataframe([df_count])
+            df_count.with_column(
+                "encoder_count", lit(df_encoder_length)
+            ).write.save_as_table(
+                transformer._encoder_count_table, create_temp_table=True
+            )
+        """
+        When using fit() function of onehot encoder. A table which record the number of types in each input column will
```
sfc-gh-jdu, line 125:
```suggestion
        When using fit() function of onehot encoder. A table which records the number of types in each input column will
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+        df_count_table = df._session.table(self._encoder_count_table)
+        df = df.join(df_count_table)
+
+        for fc, ec, (c, table_name) in zip(
+            self._states_table_cols,
+            self._encoder_table_cols,
+            self._category_table_name.items(),
+        ):
+            category_table = df._session.table(table_name)
+            # if sparse is set to true, covert category_table so that it store onehot array
```
sfc-gh-jdu, line 230:
```suggestion
            # if sparse is set to true, convert category_table so that it stores onehot array
```

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,140 @@
...
+            )
+        if encoder_type == "onehot":
+            temp = lit(0)
+            for c in transformer._states_table_cols:
+                temp += df_encoder_length[f"{c}_count"]
+
+            df_encoder_length.with_column("encoder_count", temp).write.save_as_table(
+                transformer._encoder_count_table,
+                create_temp_table=True,
+                mode="overwrite",
```
sfc-gh-jdu, line 81: I don't think we need to overwrite?

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,140 @@
...
+                TempObjectType.TABLE
+            )
+            df_save = session.create_dataframe(cat, schema=[fc])
+            if encoder_type == "onehot":
+                df_count[f"{fc}_count"] = len(cat)
+                df_encoder_length += len(cat)
+            df_save.with_column(f"{fc}_encoder", builtin("seq8")()).write.save_as_table(
+                transformer._category_table_name[c], create_temp_table=True
+            )
+        """
```
sfc-gh-jdu, line 102: Move up this docstring and the following string? Similar to what you did in scaler_fit?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+                df = category_table.join(
+                    df, df[c] == category_table[fc], join_type="right"
+                ).join(df_sparse_array.select([col(f"{fc}_sparse")]).limit(1))
+            else:
+                df = category_table.join(
+                    df, df[c] == category_table[fc], join_type="right"
+                )
+            if self._merge:
+                if not sparse:
+
```
sfc-gh-jdu, line 257: remove empty line

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+            category_table = df._session.table(table_name)
+            # if sparse is set to true, covert category_table so that it store onehot array
+            if sparse:
+                df_sparse_array = category_table.with_column("sparse", lit(0)).select(
+                    [array_agg(col("sparse")).as_(f"{fc}_sparse")]
+                )
+                df_sparse_array = category_table.join(df_sparse_array).join(
+                    df_count_table
+                )
+                category_table = df_sparse_array.select(
```
sfc-gh-jdu, line 238: add comment for this operation?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+                else:
+                    encoder_structure = array_cat(
+                        encoder_structure,
+                        iff(col(ec).is_null(), df[f"{fc}_sparse"], col(ec)),
+                    )
+        # depending on self._merge and self.sparse, prepare the final output
+        if self._merge:
+            if not sparse:
+
+                encoder_structure = array_construct_compact(*encoder_structure)
```
sfc-gh-jdu, line 271: add a comment saying null will be ignored?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+                ).join(df_sparse_array.select([col(f"{fc}_sparse")]).limit(1))
+            else:
+                df = category_table.join(
+                    df, df[c] == category_table[fc], join_type="right"
+                )
+            if self._merge:
+                if not sparse:
+
+                    encoder_structure.append(
+                        iff(col(ec).is_null(), col(ec), col(ec) + encoder_offset)
```
sfc-gh-jdu, line 259: add a comment saying offset is used to shift from one category to another category?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+        for fc, ec, (c, table_name) in zip(
+            self._states_table_cols,
+            self._encoder_table_cols,
+            self._category_table_name.items(),
+        ):
+            category_table = df._session.table(table_name)
+            df = category_table.join(df, df[c] == category_table[fc], join_type="right")
+            if self._merge:
+                encoder_structure.append(col(ec))
+        if self._merge:
```
sfc-gh-jdu, line 361: This can be included into the following if clause.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+        df: DataFrame,
+    ) -> DataFrame:
+        if self._output_cols is None:
+            raise ValueError("Output column can not be empty")
+        elif len(self._output_cols) == 1:
+            self._merge = True
+        elif len(self._input_cols) == len(self._output_cols):
+            self._merge = False
+        else:
+            raise ValueError("number of output colum and input column does not match")
```
sfc-gh-jdu, line 349:
```suggestion
            raise ValueError("The number of output colum and input column does not match")
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+            self._states_table_cols,
+            self._encoder_table_cols,
+            self._category_table_name.items(),
+        ):
+            category_table = df._session.table(table_name)
+            df = category_table.join(df, df[c] == category_table[fc], join_type="right")
+            if self._merge:
+                encoder_structure.append(col(ec))
+        if self._merge:
+            encoder_structure = array_construct(*encoder_structure)
```
sfc-gh-jdu, line 362: Add a comment saying unknown value will be set to NULL

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+                    onehot_array = array_construct_compact(
+                        df[f"{fc}_count"], df[f"{fc}_encoder"], lit(1)
+                    )
+                else:
+                    onehot_array = iff(
+                        df[f"{fc}_encoder"].is_null(),
+                        df[f"{fc}_sparse"],
+                        df[f"{fc}_encoder"],
+                    )
+                df = df.with_column(oc, onehot_array)
```
sfc-gh-jdu, line 293: Could you use [with_columns](https://github.com/snowflakedb/snowpark-python/blob/e0cc59f81bc48e027223528d112825edc9dd0b65/src/snowflake/snowpark/dataframe.py#L1764) here to add multiple columns? It will optimize the query under the hood. The same for ordinal encoder and binarizer.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+        """
+        The standard score of a sample x is calculated as:
+        z = (x - mean) / stddev
+        """
+        if not self._output_cols or len(self._input_cols) != len(self._output_cols):
+            raise ValueError(
+                "The number of input column and output column does not match or output column is empty"
+            )
+        states_table = df._session.table(self._states_table_name)
+        return (
```
sfc-gh-jdu, line 95: My bad, actually we should also retain the original columns for scaler. Could you try to use with_columns to add these output cols, instead of using to_df?

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,346 @@
...
+
+def test_input_cols_output_cols_not_match(session):
+    df = session.createDataFrame(
+        [[1, 4, 5], [2, 4, 5], [2, 3, 6]], schema=["a", "b", "c"]
+    )
+    onehotencoder = OneHotEncoder()
+    onehotencoder.input_cols = ["A", "B", "C"]
+    onehotencoder.output_cols = ["B", "C"]
+    model = onehotencoder.fit(df)
+    with pytest.raises(ValueError):
```
sfc-gh-jdu, line 277: assert the content of error message

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,346 @@
...
+        model.transform(df)
+
+
+def test_preset_category_not_match_real_category(session):
```
sfc-gh-jdu, line 281: This edge case doesn't only happen when we have preset category, right? For example, 
```
df = session.createDataFrame(
        [[1, 4, 5], [2, 4, 5], [2, 3, 6]], schema=["a", "b", "c"]
    )
```
If we get 7 in transform() call, it should be handled correctly. Also, let's check this edge case for 
1. onehot encoder 
   1.a sparse
   1.b not sparse 
   (you can also test merge, but not very necessary)
   1.c preset category
   1.d auto mode
2. ordinal encoder
   1.a preset category
   1.b auto mode

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,395 @@
...
+            df = df.select([col(c) for c in original_columns])
+        else:
+            for fc, oc in zip(self._states_table_cols, self._output_cols):
+                df = df.with_column(oc, df[f"{fc}_encoder"])
+                original_columns.append(oc)
+            df = df.select([col(c) for c in original_columns])
+        return df
+
+
+class Binarizer(Transformer):
```
sfc-gh-jdu, line 375: Add test for binarizer

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,422 @@
...
+        """
+        if not self._output_cols or len(self._input_cols) != len(self._output_cols):
+            raise ValueError(
+                "The number of input column and output column does not match or output column is empty"
+            )
+        states_table = df._session.table(self._states_table_name)
+        res_column = []
+        for c in self.input_cols:
+            res_column.append(df[c])
+        for c, fc, oc in zip(
```
sfc-gh-sfan, line 98: nit: let's give a more descriptive name than `c`, `fc`, `oc`. For example: `input_col`, `state_table_col`, `output_col`

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,422 @@
...
+        if self._output_cols is None:
+            raise ValueError("Output column can not be empty")
+        elif len(self._output_cols) == 1:
+            self._merge = True
+        elif len(self._input_cols) == len(self._output_cols):
+            self._merge = False
+        else:
+            raise ValueError(
+                "The number of output colum and input column does not match"
+            )
```
sfc-gh-sfan, line 372: I'm curious if it makes sense to extract this check into base class given it is called in different sub classes. Ditto for other `transform` checks.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,422 @@
...
+                encoder_structure.append(col(ec))
+
+        if self._merge:
+            encoder_structure = array_construct(*encoder_structure)
+            original_columns.append(self._output_cols[0])
+            df = df.with_column(self._output_cols[0], encoder_structure)
+            df = df.select([col(c) for c in original_columns])
+        else:
+            encoder_columns = []
+            for fc, oc in zip(self._states_table_cols, self._output_cols):
```
sfc-gh-sfan, line 394: ditto: more descriptive name

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,422 @@
...
+        self._threshold = threshold
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        if not self._input_cols:
+            raise ValueError("Input columns can not be empty")
+        elif not self._output_cols or len(self._input_cols) != len(self._output_cols):
+            raise ValueError(
+                "The number of input column and output column does not match or output column is empty"
+            )
+        for c, oc in zip(self._input_cols, self._output_cols):
```
sfc-gh-sfan, line 419: ditto: more descriptive name

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,422 @@
...
+    ):
+        super().__init__(input_cols, output_cols)
+        self._states_table_cols = (
+            [f"states_{c}" for c in self._input_cols] if self._input_cols else None
+        )
+        self._states_table_name = random_name_for_temp_object(TempObjectType.TABLE)
+
+    @Transformer.input_cols.setter
+    def input_cols(self, value: List[str]) -> None:
+        self._input_cols = value
```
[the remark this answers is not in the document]
sfc-gh-jdu: It has to be checked in fit() method, because we have two ways to set input cols (another one is to use constructor of transformer).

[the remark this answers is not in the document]
sfc-gh-sfan: Right. I was thinking that it does not make sense for `value` to be `None` in `setter`. But after a second thought, maybe we don't want to duplicate the check.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,422 @@
...
+
+        scaler_fit(self, df, ["mean", "stddev_pop"])
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """
+        The standard score of a sample x is calculated as:
+        z = (x - mean) / stddev
+        """
+        if not self._output_cols or len(self._input_cols) != len(self._output_cols):
```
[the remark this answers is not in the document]
sfc-gh-jdu: It has been checked in fit() method (we have to call transform() after fit())

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,505 @@
...
+    Utils.check_answer(
+        res,
+        [
+            Row(A=1, B=4, C=5, ONEHOTARRAY="[\n  0,\n  1,\n  0,\n  1,\n  0\n]"),
+            Row(A=2, B=3, C=6, ONEHOTARRAY="[\n  1,\n  0,\n  1,\n  0,\n  1\n]"),
+            Row(A=2, B=4, C=5, ONEHOTARRAY="[\n  1,\n  1,\n  0,\n  1,\n  0\n]"),
+        ],
+    )
+
+    # merge and sparse is true
```
sfc-gh-jdu, line 338:
```suggestion
    # merge is false and sparse is true
```
Also fix the previous comment

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,505 @@
...
+                A=2,
+                B=4,
+                C=5,
+                RES_A="[\n  1,\n  0,\n  1\n]",
+                RES_B="[\n  2,\n  0,\n  1\n]",
+                RES_C="[\n  2,\n  0,\n  1\n]",
+            ),
+        ],
+    )
+    # merge and sparse is false
```
sfc-gh-jdu, line 370:
```suggestion
    # merge is false and sparse is false
```

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,505 @@
...
+        [Row(STATES_A=1, STATES_A_ENCODER=0), Row(STATES_A=2, STATES_A_ENCODER=1)],
+    )
+    df_unknown_value = session.createDataFrame(
+        [[6, 4, 5], [2, 4, 5], [2, 3, 6]], schema=["a", "b", "c"]
+    )
+    res = model.transform(df_unknown_value)
+    Utils.check_answer(
+        res,
+        [
+            Row(
```
sfc-gh-jdu, line 424: It looks like the row order is not preserved?

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,505 @@
...
+        res,
+        [
+            Row(A=1, B=4, C=5, ORDINALARRAY="[\n  0,\n  0,\n  0\n]"),
+            Row(A=2, B=3, C=6, ORDINALARRAY="[\n  1,\n  1,\n  1\n]"),
+            Row(A=2, B=4, C=5, ORDINALARRAY="[\n  1,\n  0,\n  0\n]"),
+        ],
+    )
+
+
+def test_preset_category_ordinal_encoder(session):
```
sfc-gh-jdu, line 468: Also add a test for ordinal encoder when there is unknown value?

tests/integ/ml/test_transformer.py
sfc-gh-yuwang: row order is preserved, but util.check_answer accept random order of answer

src/snowflake/snowpark/dataframe.py
```diff
@@ -1092,9 +1093,15 @@ def distinct(self) -> "DataFrame":
...
-        return self.group_by(
-            [self.col(quote_name(f.name)) for f in self.schema.fields]
-        ).agg([])
+        return self._distinct(self.columns)
+
+    def _distinct(self, columns: List[str]) -> "DataFrame":
+        """Internal function that return a new DataFrame that contains only the rows with distinct values
+        from the current DataFrame.
+
+        This is equivalent to performing a SELECT DISTINCT in SQL.
```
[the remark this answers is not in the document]
sfc-gh-jdu: I'm also not sure 🤦 we just follow scala's implementation here

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+
+class Transformer:
+    def __init__(
+        self,
+        input_cols: Optional[List[str]] = None,
+        output_cols: Optional[List[str]] = None,
+    ):
+        self._input_cols = input_cols
+        self._output_cols = output_cols
+        self._fit_was_called = False
```
sfc-gh-jdu, line 35: maybe just use `self._fitted`, which is more concise?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+            raise ValueError("Input columns can not be None")
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        if not self._input_cols:
+            raise ValueError("Input column can not be empty")
+        elif not self._output_cols:
+            raise ValueError("Output column can not be empty")
+        elif not self._fit_was_called:
+            raise Exception(
+                "The transformer was not fitted yet, call fit() function before transform"
```
sfc-gh-jdu, line 65:
```suggestion
                "The transformer is not fitted yet, call fit() before transform()"
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+            )
+        elif len(self._output_cols) == 1:
+            self._merge = True
```
sfc-gh-jdu, line 68: move this check to encoder, as scaler doesn't have `merge` option.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+        if not self._input_cols:
+            raise ValueError("Input columns can not be None")
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        if not self._input_cols:
+            raise ValueError("Input column can not be empty")
+        elif not self._output_cols:
+            raise ValueError("Output column can not be empty")
+        elif not self._fit_was_called:
+            raise Exception(
```
sfc-gh-jdu, line 64: Let's also use ValueError

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+        super().fit(df)
+        scaler_fit(self, df, ["mean", "stddev_pop"])
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """
+        The standard score of a sample x is calculated as:
+        z = (x - mean) / stddev
+        """
+        if not self._fit_was_called:
```
sfc-gh-jdu, line 110: you don't need these checks here anymore, right?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+    def __init__(
+        self,
+        feature_range: Tuple[float, float] = (0, 1),
+        input_cols: Optional[List[str]] = None,
+        output_cols: Optional[List[str]] = None,
+    ):
+        super().__init__(input_cols, output_cols)
+        if feature_range[0] >= feature_range[1]:
+            raise ValueError(
+                "Maximum in feature range can not be equal or greater than minimum in feature range"
```
sfc-gh-jdu, line 149:
```suggestion
                "Minimum cannot be equal or greater than maximum in feature range"
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+    def fit(self, df: DataFrame) -> "OneHotEncoder":
+        super().fit(df)
+
+        encoder_fit(self, df, "onehot")
+        return self
+
+    def transform(
+        self,
+        df: DataFrame,
+        dense: bool = True,
```
sfc-gh-jdu, line 259: sorry I meant just changing the variable and column name in the code, but not changing the argument name. We should keep it as `sparse`.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+        for states_col, encoder_col, (input_col, table_name) in zip(
+            self._states_table_cols,
+            self._encoder_table_cols,
+            self._category_table_name.items(),
+        ):
+            category_table = df._session.table(table_name)
+            # if dense is set to true, convert category_table so that it stores onehot array
+            if not dense:
+                # create a [0,0...0,0] array as the base of dense encoder array
+                df_dense_array = category_table.with_column("dense", lit(0)).select(
```
sfc-gh-jdu, line 278: Yes, this should be changed to dense

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+                    [array_agg(col("dense")).as_(f"{states_col}_dense")]
+                )
+                # join tables to get the length of each encoder array, STATES_X_SPARSE looks like [0,...,0], length of
```
sfc-gh-jdu, line 281:
```suggestion
                # join tables to get the length of each encoder array, STATES_X_DENSE looks like [0,...,0], length of
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+                # this array depending on how many category in this column
```
sfc-gh-jdu, line 282:
```suggestion
                # this array depends on how many categories in this column
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+                else:
+                    encoder_structure = array_cat(
+                        encoder_structure,
+                        iff(
+                            col(encoder_col).is_null(),
+                            df[f"{states_col}_dense"],
+                            col(encoder_col),
+                        ),
+                    )
+        # depending on self._merge and self.dense, prepare the final output
```
sfc-gh-jdu, line 356:
```suggestion
        # depending on self._merge and sparse, prepare the final output
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+            for states_col, output_col in zip(
+                self._states_table_cols, self._output_cols
+            ):
+                if dense:
+                    onehot_array = array_construct_compact(
+                        df[f"{states_col}_count"], df[f"{states_col}_encoder"], lit(1)
+                    )
+                else:
+                    # df[f"{states_col}_encoder"] is null if df fed into transform has unknown category compare
+                    # to category_table
```
sfc-gh-jdu, line 381:
```suggestion
                    # to category_table because of right join. 
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -2353,14 +2367,21 @@ def describe(self, *cols: Union[str, List[str]]) -> "DataFrame":
...
+        all_stat_func_dict = {
             "count": count,
             "mean": mean,
             "stddev": stddev,
+            "stddev_pop": stddev_pop,
             "min": min_,
             "max": max_,
         }
 
+        stat_func_dict = (
```
sfc-gh-yixie, line 2379: Are we using a dict like a tuple? I don't see a key lookup.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+    @output_cols.setter
+    def output_cols(self, value: List[str]) -> None:
+        self._output_cols = value
+
+    def fit(self, df: DataFrame) -> "Transformer":
+        self._fit_was_called = True
+        if not self._input_cols:
+            raise ValueError("Input columns can not be None")
+
+    def transform(self, df: DataFrame) -> DataFrame:
```
sfc-gh-yixie, line 58: Some subclasses override `transform` without calling the `super().transform`. How about removing the implementation here and making it an abstract method?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+    array_insert,
+    array_slice,
+    col,
+    iff,
+    lit,
+)
+from snowflake.snowpark.ml.utils import encoder_fit, scaler_fit
+
+
+class Transformer:
```
sfc-gh-yixie, line 27: Is it possible to make it an abstract class.

src/snowflake/snowpark/ml/transformer.py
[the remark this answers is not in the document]
sfc-gh-yixie: The parent class already have this property setter. Why do we need it here again?

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,501 @@
...
+        self._output_cols = output_cols
+        self._fit_was_called = False
+
+    @property
+    def input_cols(self) -> List[str]:
+        return self._input_cols
+
+    @input_cols.setter
+    def input_cols(self, value: List[str]) -> None:
+        self._input_cols = value
```
sfc-gh-yixie, line 43: Subclasses override it and doesn't call the super. So the implementation isn't useful at all.

src/snowflake/snowpark/ml/transformer.py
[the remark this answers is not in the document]
sfc-gh-yuwang: > The parent class already have this property setter. Why do we need it here again?

The implementation of input_cols.setter is different from that base class, so I actually rewrite it here

src/snowflake/snowpark/ml/transformer.py
sfc-gh-yuwang: i think it should be dense here?

sfc-gh-yuwang: I see your comment above, got it

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,499 @@
...
+
+        scaler_fit(self, df, ["min", "max"])
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """
+        The transformation is given by:
+        X_std = (X - X.min(axis=0)) / (X.max(axis=0) - X.min(axis=0))
+        X_scaled = X_std * (max - min) + min
+        where min, max are decided by feature_range
```
sfc-gh-jdu, line 166:
```suggestion
        where min, max are provided by feature_range
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,499 @@
...
+    def fit(self, df: DataFrame) -> "MinMaxScaler":
+        super().fit(df)
+
+        scaler_fit(self, df, ["min", "max"])
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """
+        The transformation is given by:
+        X_std = (X - X.min(axis=0)) / (X.max(axis=0) - X.min(axis=0))
```
sfc-gh-jdu, line 164:
```suggestion
        X_std = (X - min(X)) / (max(X) - min(X))
```
Let's avoid numpy expression since we don't have axis concept.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,499 @@
...
+    def __init__(
+        self,
+        feature_range: Tuple[float, float] = (0, 1),
+        input_cols: Optional[List[str]] = None,
+        output_cols: Optional[List[str]] = None,
+    ) -> None:
+        super().__init__(input_cols, output_cols)
+        if feature_range[0] >= feature_range[1]:
+            raise ValueError(
+                "Maximum  can not be equal or greater than minimum in feature range"
```
sfc-gh-jdu, line 136:
```suggestion
                "Maximum can not be equal or greater than minimum in feature range"
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,499 @@
...
+        res_column = df.columns
+
+        for input_col, states_col, output_col in zip(
+            self.input_cols, self._states_table_cols, self.output_cols
+        ):
+            res_column.append(
+                (
+                    (df[input_col] - col(states_col)["min"])
+                    / iff(
+                        (col(states_col)["max"] - col(states_col)["min"]) == 0,
```
sfc-gh-jdu, line 180: also add a comment like ` # if stddev equals to 0, we will set it to 1 to avoid divided by 0 error`

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,499 @@
...
+
+        encoder_fit(self, df, "ordinal")
+        return self
+
+    def transform(
+        self,
+        df: DataFrame,
+    ) -> DataFrame:
+        """
+        The ordinal encoder will be "null" if there are new types in dataframe fed into transform compared to dataframe
```
sfc-gh-jdu, line 432:
```suggestion
        The encoded value will be null if there are unknown types in this dataframe, compared to the dataframe
```

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,499 @@
...
+
+        encoder_fit(self, df, "onehot")
+        return self
+
+    def transform(
+        self,
+        df: DataFrame,
+        sparse: bool = True,
+    ) -> DataFrame:
+        if not self.input_cols:
```
sfc-gh-jdu, line 241: You should also call super().transform() and remove the duplicate code, right? The same for ordinal encoder.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -0,0 +1,499 @@
...
+
+    @property
+    def output_cols(self) -> List[str]:
+        return self._output_cols
+
+    @output_cols.setter
+    def output_cols(self, value: List[str]) -> None:
+        self._output_cols = value
+
+    def fit(self, df: DataFrame) -> "Transformer":
```
sfc-gh-jdu, line 53:
```suggestion
    def fit(self, df: DataFrame) -> None:
```
In this base class, we return nothing so the type hint should be fixed. The same for transform().

src/snowflake/snowpark/ml/transformer.py
sfc-gh-jdu:
```suggestion
    def transform(self, df: DataFrame) -> None:
```

src/snowflake/snowpark/ml/utils.py
```diff
@@ -0,0 +1,139 @@
...
+    -------------------------------------------------------------------------------------------------------------------------
+    |"STATES_A"                             |"STATES_B"                             |"STATES_C"                             |
+    -------------------------------------------------------------------------------------------------------------------------
+    |{                                      |{                                      |{                                      |
+    |  "mean": 1.666667000000000e+00,       |  "mean": 3.666667000000000e+00,       |  "mean": 5.333333000000000e+00,       |
+    |  "stddev_pop": 4.714042850887123e-01  |  "stddev_pop": 4.714042850887123e-01  |  "stddev_pop": 4.714042850887123e-01  |
+    |}                                      |}                                      |}                                      |
+    -------------------------------------------------------------------------------------------------------------------------
+    """
+    if len(transformer.input_cols) != len(transformer.output_cols):
```
sfc-gh-jdu, line 37: We don't need this here, right?

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,524 @@
...
+    Binarizer,
+    MinMaxScaler,
+    OneHotEncoder,
+    OrdinalEncoder,
+    StandardScaler,
+)
+from tests.utils import Utils
+
+
+def test_standardscaler(session):
```
sfc-gh-jdu, line 19:
```suggestion
def test_standard_scaler(session):
```

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,524 @@
...
+        res,
+        [
+            Row(A=1, B=2, C=3, RES=-1.069044547667817),
+            Row(A=2, B=3, C=4, RES=-0.26726093647105154),
+            Row(A=4, B=5, C=6, RES=1.3363062859224795),
+        ],
+    )
+
+
+def test_minmaxscaler(session):
```
sfc-gh-jdu, line 49:
```suggestion
def test_minmax_scaler(session):
```

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,524 @@
...
+        res,
+        [
+            Row(A=1, B=2, C=3, RES=0.0),
+            Row(A=2, B=3, C=4, RES=0.3333333333333333),
+            Row(A=4, B=5, C=6, RES=1.0),
+        ],
+    )
+
+
+def test_merge_onehotencoder(session):
```
sfc-gh-jdu, line 73:
```suggestion
def test_merge_onehot_encoder(session):
```
and the same for all other places in this file.

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,524 @@
...
+        res,
+        [
+            Row(A=1, B=4, C=5, ONEHOTARRAY="[\n  1,\n  0,\n  1,\n  0,\n  1,\n  0\n]"),
+            Row(A=2, B=4, C=5, ONEHOTARRAY="[\n  0,\n  1,\n  1,\n  0,\n  1,\n  0\n]"),
+            Row(A=2, B=3, C=6, ONEHOTARRAY="[\n  0,\n  1,\n  0,\n  1,\n  0,\n  1\n]"),
+        ],
+    )
+
+
+def test_input_cols_output_cols_not_match(session):
```
sfc-gh-jdu, line 278: Let's rename this test as `test_tranformer_input_cols_output_cols_negative` and have all negative cases here? Like no input columns, no output columns, etc.

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,524 @@
...
+                A=2,
+                B=3,
+                C=6,
+                ONEHOTARRAY="[\n  6,\n  [\n    1,\n    3,\n    5\n  ],\n  1\n]",
+            ),
+        ],
+    )
+
+
+def test_merge_ordinalcoder(session):
```
sfc-gh-jdu, line 446:
```suggestion
def test_merge_ordinal_encoder(session):
```

tests/integ/ml/test_transformer.py
```diff
@@ -0,0 +1,524 @@
...
+        res,
+        [
+            Row(A=3, B=4, C=5, ONEHOTARRAY="[\n  undefined,\n  0,\n  0\n]"),
+            Row(A=2, B=3, C=6, ONEHOTARRAY="[\n  1,\n  1,\n  1\n]"),
+            Row(A=2, B=4, C=5, ONEHOTARRAY="[\n  1,\n  0,\n  0\n]"),
+        ],
+    )
+
+
+def test_binarizer(session):
```
sfc-gh-jdu, line 514: Probably add one more test case about float data and float threshold (you can also try to use nan as threshold)

src/snowflake/snowpark/ml/transformer.py
sfc-gh-yuwang: No, I have to customize it here. If length of output_col is 1 and not equals to input_col, super().transform() will raise a ValueError directly.

sfc-gh-jdu: I see, then super().transform() shouldn't check this. The method in the base class should work for all subclasses.

src/snowflake/snowpark/ml/transformer.py
```diff
@@ -429,7 +434,7 @@ def transform(
         df: DataFrame,
     ) -> DataFrame:
         """
-        The ordinal encoder will be "null" if there are new types in dataframe fed into transform compared to dataframe
+        The encoder value will be null if there are unknown types in this dataframe compared to that used in fit()
```
sfc-gh-jdu, line 437:
```suggestion
        The encoded value will be null if there are unknown types in this dataframe compared to that dataframe used in fit() function.
```
and remove the following line

src/snowflake/snowpark/ml/utils.py
```diff
@@ -80,26 +80,32 @@ def encoder_fit(
     """
     if transformer._category == "auto":
         df_encoder_length = None
-        for c, fc in zip(transformer.input_cols, transformer._states_table_cols):
-            transformer._category_table_name[c] = random_name_for_temp_object(
+        for input_cols, states_col in zip(
```
sfc-gh-jdu, line 83:
```suggestion
        for input_col, states_col in zip(
```

src/snowflake/snowpark/ml/utils.py
```diff
@@ -114,20 +120,22 @@ def encoder_fit(
         session = df._session
         df_count = {}
         df_encoder_length = 0
-        for c, fc, cat in zip(
+        for input_cols, states_col, category in zip(
```
sfc-gh-jdu, line 123:
```suggestion
        for input_col, states_col, category in zip(
```

tests/integ/ml/test_transformer.py
```diff
@@ -275,19 +275,36 @@ def test_sparse_merge_onehotencoder(session):
...
-def test_input_cols_output_cols_not_match(session):
+def test_input_cols_output_cols_negative(session):
     df = session.create_dataframe(
         [[1, 4, 5], [2, 4, 5], [2, 3, 6]], schema=["a", "b", "c"]
     )
     onehotencoder = OneHotEncoder()
+    # test no input column
+    with pytest.raises(ValueError) as ex_info:
+        onehotencoder.transform(df)
+    assert "Input column can not be empty" == str(ex_info.value)
```
sfc-gh-jdu, line 286:
```suggestion
    assert "Input column can not be empty" in str(ex_info.value)
```
The same for other assertion of strings

sfc-gh-jdu (approved):
🚀

sfc-gh-yixie (approved):

PR #390, opened 2022-06-27
SNOW-616199: updated _resolve_packages to only query for added packages

sfc-gh-aalam:

   No existing github issue. Only addressing JIRA SNOW-616199

   - Updated code to only query for packages and versions that are requested in `add_packages` method.
   - Updated code to check if requested versions are available in Snowflake. Raise error if they are not.
   - Updated code to warn when local version of python package does not match with any supported packages on Snowflake server

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -7,7 +7,6 @@ import decimal
 import json
 import logging
 import os
-import warnings
 from array import array
 from functools import reduce
 from logging import getLogger
@@ -59,7 +58,7 @@ from snowflake.snowpark._internal.utils import (
     PythonObjJSONEncoder,
     TempObjectType,
     calculate_checksum,
-    deprecate,
+    deprecated,
     get_connector_version,
     get_os_name,
     get_python_version,
@@ -73,6 +72,7 @@ from snowflake.snowpark._internal.utils import (
     unwrap_single_quote,
     unwrap_stage_location_single_quote,
     validate_object_name,
+    warning,
     zip_file_or_directory_to_stream,
 )
 from snowflake.snowpark.column import Column
@@ -806,7 +806,7 @@ class Session:
                             package_name
                         ).version
                         if package_client_version not in valid_packages[package_name]:
-                            logging.warning(
+                            _logger.warning(
                                 "The version of package %s in the local environment is %s, "
                                 "which does not fit the criteria for the requirement %s. "
                                 "Your UDF might not work when the package version is different "
@@ -816,7 +816,7 @@ class Session:
                                 package,
                             )
                     except pkg_resources.DistributionNotFound:
-                        logging.warning(
+                        _logger.warning(
                             "package %s is not installed in the local environment"
                             "Your UDF might not work when the package is installed "
                             "on the server but not on your local environment.",
@@ -977,7 +977,7 @@ class Session:
                 self,
                 TableFunctionRelation(func_expr),
             )
-        set_api_call_source(d, f"Session.table_function[{func_expr.func_name}]")
+        set_api_call_source(d, "Session.table_function")
         return d
 
     def sql(self, query: str) -> DataFrame:
@@ -1136,12 +1136,10 @@ class Session:
             exception will be raised.
         """
         if create_temp_table:
-            warnings.warn(
+            warning(
+                "write_pandas.create_temp_table",
                 "create_temp_table is deprecated. We still respect this parameter when it is True but "
                 'please consider using `table_type="temporary"` instead.',
-                DeprecationWarning,
-                # warnings.warn -> write_pandas
-                stacklevel=2,
             )
             table_type = "temporary"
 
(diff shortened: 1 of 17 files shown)


sfc-gh-aalam:

src/snowflake/snowpark/session.py
```diff
@@ -680,14 +681,30 @@ def _resolve_packages(
...
+        package_names = list()
+        for package in packages:
+            if isinstance(package, ModuleType):
+                package_name = MODULE_NAME_TO_PACKAGE_NAME_MAP.get(
+                    package.__name__, package.__name__
+                )
+                package = f"{package_name}=={pkg_resources.get_distribution(package_name).version}"
+            else:
+                package = package.strip().lower()
+            package_names.append(pkg_resources.Requirement.parse(package).key)
```
sfc-gh-jdu, line 693: You can save (package name -> (package requirement identifier, is_from_local)) as a dict, which can be used later in this function. Then you can avoid duplication here https://github.com/snowflakedb/snowpark-python/blob/6f0616627e53780ceb0ab897d92e60fad1365c91/src/snowflake/snowpark/session.py#L698-L709.

src/snowflake/snowpark/session.py
```diff
@@ -717,11 +734,23 @@ def _resolve_packages(
                         f"and add it via session.add_import()."
                     )
                 elif not use_local_version:
+                    # check if user as requested for a specific version of package 'pkg==version'
+                    if package_req.specs:
+                        package_version_req = package_req.specs[0][1]
+                        if package_version_req not in valid_packages[package_name]:
+                            raise ValueError(
+                                f"Cannot add package {package_name} {package_version_req} "
```
sfc-gh-sfan, line 742: I'm wondering if we should have a helper function for this error message as this looks very similar to the one above and the only difference is `{package_version_req}`. This way if we want to modify the message, we run a lower risk of forgetting one.

src/snowflake/snowpark/session.py
```diff
@@ -717,11 +734,23 @@ def _resolve_packages(
                         f"and add it via session.add_import()."
                     )
                 elif not use_local_version:
+                    # check if user as requested for a specific version of package 'pkg==version'
```
sfc-gh-sfan, line 737: Why is this check only under `use_local_version`?

tests/integ/test_udf.py
```diff
@@ -1211,14 +1211,18 @@ def test_add_packages_negative(session, caplog):
 
     with pytest.raises(ValueError) as ex_info:
         with caplog.at_level(logging.WARNING):
-            session.add_packages("numpy", "numpy==0.1.0")
+            session.add_packages("numpy", "numpy==1.16.6")
```
sfc-gh-jdu, line 1214: here 1.16.6 is the numpy version that is supported on the server side and not installed on the client side, right? I feel it's not good since it might not be supported on the server side in the future, and we can't control the version installed on the client side, but I couldn't think of a better solution and since it's just a test, maybe add a comment here to explain why do you use 1.16.6?

src/snowflake/snowpark/session.py
sfc-gh-aalam: Do we have a specific file where we put these helper functions or should I create a function locally?

tests/integ/test_udf.py
sfc-gh-aalam: handled in the latest commit

src/snowflake/snowpark/session.py
sfc-gh-aalam: Updated to cover all cases

sfc-gh-jdu (approved):

src/snowflake/snowpark/session.py
```diff
@@ -680,48 +681,77 @@ def _resolve_packages(
...
-                use_local_version = True
-            else:
-                package = package.strip().lower()
-                use_local_version = False
+        for package in package_dict:
             package_req = pkg_resources.Requirement.parse(package)
-            # get the standard package name
-            package_name = package_req.key
+            package_name = package_dict[package][0]
+            use_local_version = package_dict[package][1]
```
sfc-gh-sfan, line 721: nit: `package_name, use_local_version = package_dict[package]`, or replacing `package_dict[package]` after using `items` as suggested by Jianzhun.

sfc-gh-sfan (approved):
Looks good!

tests/integ/test_udf.py
```diff
@@ -1185,12 +1185,10 @@ def is_numpy_available() -> bool:
     # add module objects
     # but we can't register a udf with these versions
     # because the server might not have them
-    session.add_packages(numpy, pandas, dateutil)
-    assert session.get_packages() == {
-        "numpy": f"numpy=={numpy.__version__}",
-        "pandas": f"pandas=={pandas.__version__}",
-        "python-dateutil": f"python-dateutil=={dateutil.__version__}",
-    }
```
sfc-gh-yixie, line 1193: This part uses the public API `session.get_packages`. From test perspective, it's different from testing `session._resolve_packages`. I suggest you keep this part and add new tests for `session._resolve_packages`

sfc-gh-yixie: So there is a behavior change here. We need to check the behavior change process.

sfc-gh-yixie (approved):

sfc-gh-jdu:
Just dump the release notes here 

When adding a package using `session.add_packages()` with a specific version to UDFs or stored procedures, if the version is not supported by Snowflake, an exception will be raised immediately instead of when registering the UDF or stored procedure.

PR #393, opened 2022-07-01
SNOW-620677 Using Builtin sum Function Causes Hanging

sfc-gh-kwagner:
Description
When we use the builtin sum function instead of the Snowpark defined sum function, then the library hangs and uses up a lot of memory. This PR addresses that and tests for other builtin functions that shadow Snowpark function names

Testing
Tested builtin sum function as well as other builtin functions

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -304,7 +304,14 @@ class Column:
         )
 
     def __iter__(self) -> None:
-        raise TypeError("Column is not iterable")
+        raise TypeError(
+            "Column is not iterable. This error can occur when you use the Python built-ins for sum, min and max. Please make sure you use the corresponding function from snowflake.snowpark.functions."
+        )
+
+    def __round__(self, n=None):
+        raise TypeError(
+            "Column cannot be rounded. This error can occur when you use the Python built-in round. Please make sure you use the snowflake.snowpark.functions.round function intead."
+        )
 
     def in_(
         self,
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -303,6 +303,9 @@ class Column:
             "'~' for 'not' if you're building DataFrame filter expressions. For example, use df.filter((col1 > 1) & (col2 > 2)) instead of df.filter(col1 > 1 and col2 > 2)."
         )
 
+    def __iter__(self) -> None:
+        raise TypeError("Column is not iterable")
+
     def in_(
         self,
         *vals: Union[
--- a/tests/integ/test_column.py
+++ b/tests/integ/test_column.py
@@ -151,3 +151,36 @@ def test_when_accept_sql_expr(session):
     assert TestData.null_data1(session).select(
         when("a is NULL", 5).when("a = 1", 6).as_("a")
     ).collect() == [Row(5), Row(None), Row(6), Row(None), Row(5)]
+
+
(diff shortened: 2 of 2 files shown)


tests/integ/test_column.py
```diff
@@ -151,3 +151,36 @@ def test_when_accept_sql_expr(session):
...
+    with pytest.raises(TypeError) as ex_info:
+        TestData.double1(session).select(min(col("a"), 25)).collect()
+    assert "Cannot convert a Column object into bool" in str(ex_info)
+
+    with pytest.raises(TypeError) as ex_info:
+        TestData.double1(session).select(min(col("a"))).collect()
+    assert "Column is not iterable" in str(ex_info)
+
+    # This works because we explicitly implement the __pow__ and __rpow__ methods
+    assert TestData.double1(session).select(pow(col("a"), 2)).collect() == [
```
sfc-gh-sfan, line 174: Should we use `check_answer` given it is comparing floats?

sfc-gh-sfan (approved):

sfc-gh-jdu (approved):

src/snowflake/snowpark/column.py
```diff
@@ -303,6 +303,9 @@ def __bool__(self) -> bool:
             "'~' for 'not' if you're building DataFrame filter expressions. For example, use df.filter((col1 > 1) & (col2 > 2)) instead of df.filter(col1 > 1 and col2 > 2)."
         )
 
+    def __iter__(self) -> None:
+        raise TypeError("Column is not iterable")
```
sfc-gh-yixie, line 307: How about adding more text like if you use `sum`, `count`, please make sure you use `snowflake.snowpark.functions.sum` and `count` instead of the `sum`/`count` of python or other libraries (numpy also has sum...).

sfc-gh-yixie (commented):
Do you know why it consumes so much memory when `__iter__` is not overriden?

sfc-gh-yixie:
Please rebase to pass the failed test. Afroz has fixed the doctest in a separate PR.

sfc-gh-kwagner:
> Do you know why it consumes so much memory when `__iter__` is not overriden?

So when I tested this, it just started hanging, but didn't run out of memory. My theory is that the `sum`  built-in function tries to get the actual data from the column (similar to how the `pow` built-in function works) and in the case of Sonra's performance test, the data was so large that it sucked up all the memory on the machine and become unresponsive

src/snowflake/snowpark/column.py
```diff
@@ -303,6 +303,16 @@ def __bool__(self) -> bool:
             "'~' for 'not' if you're building DataFrame filter expressions. For example, use df.filter((col1 > 1) & (col2 > 2)) instead of df.filter(col1 > 1 and col2 > 2)."
         )
 
+    def __iter__(self) -> None:
+        raise TypeError(
+            "Column is not iterable. This error can occur when you use the Python built-ins for sum, min and max. Please make sure you use the corresponding function from snowflake.snowpark.functions."
+        )
+
+    def __round__(self, n=None):
```
sfc-gh-yixie, line 311: We override `__pow__` on `Column`. Maybe we should support `__round__` too. It's not an aggregate function.
Let's discuss tmr.

sfc-gh-yixie: Discussed that we'll only support `__pow__`.

sfc-gh-yixie (approved):

PR #394, opened 2022-07-01
SNOW-620212: Execute and get the result query id of a dataframe

sfc-gh-jdu:

   This internal method is needed for table stored proc that executes the dataframe and get the result query ID, which will be returned back to GS and used for getting the result back with result scan. 

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -7,7 +7,7 @@ import os
 import sys
 import time
 from logging import getLogger
-from typing import IO, Any, Dict, Iterator, List, Optional, Set, Union
+from typing import IO, Any, Dict, Iterator, List, Optional, Set, Tuple, Union
 
 import snowflake.connector
 from snowflake.connector import SnowflakeConnection, connect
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -450,6 +450,14 @@ def _internal_collect_with_tag(self) -> List["Row"]:
             else None,
         )
 
+    def _execute_and_get_query_id(self) -> str:
```
sfc-gh-yixie, line 453: Nit: Better add a comment here it is for Table SP. It's internal but not called by other APIs so people may wonder why we need it.

sfc-gh-yixie (approved):

src/snowflake/snowpark/dataframe.py
sfc-gh-jdu: added

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -426,16 +432,20 @@ def get_result_set(
...
-    ) -> Union[List[Row], List[Attribute]]:
+    ) -> Tuple[List[Row], List[Attribute]]:
         result_set, result_meta = self.get_result_set(plan, **kwargs)
-        result = result_set_to_rows(result_set)
+        result = result_set_to_rows(result_set["data"])
         meta = convert_result_meta_to_attribute(result_meta)
         return result, meta
 
+    def get_result_query_id(self, plan: SnowflakePlan, **kwargs) -> str:
+        result_set, _ = self.get_result_set(plan, **kwargs)
```
sfc-gh-yixie, line 446: @sfc-gh-jdu I just realized whether this would fetch the whole result set to the client. If so this may use a lot of memory. So does it fetch data?

sfc-gh-jdu: Yeah it does, I will put out another PR that this method only gets the iterator but doesn't fetch it.

PR #413, opened 2022-07-25
Pin sphinx==5.0.2 instead of latest version

sfc-gh-yixie:
Some new versions of Sphinx may introduce break changes. For instance, 5.1.0 reports `Extension error (sphinx.ext.napoleon)`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ setup(
             "pytest",
             "pytest-cov",
             "coverage",
-            "sphinx",
+            "sphinx==5.0.2",
             "cachetools",  # used in UDF doctest
         ],
     },


setup.py
```diff
@@ -74,7 +74,7 @@
             "pytest",
             "pytest-cov",
             "coverage",
-            "sphinx",
+            "sphinx==5.0.2",
```
sfc-gh-sfan, line 77: Should we be more flexible on the micro version, e.g. `5.0.x`?

sfc-gh-sfan (approved):

setup.py
sfc-gh-yixie: This is tooling. It's not released as a part of the library so I prefer a fixed issue to avoid any possible breaks.

sfc-gh-yixie: I'm going to merge now to unblock other PRs. We can change it in the future if we change mind.

PR #416, opened 2022-07-28
SNOW-629121: create async job api

sfc-gh-yuwang:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/__init__.py
+++ b/src/snowflake/snowpark/__init__.py
@@ -34,6 +34,7 @@ __all__ = [
     "WhenNotMatchedClause",
     "QueryRecord",
     "QueryHistory",
+    "AsyncJob",
 ]
 
 
@@ -42,6 +43,7 @@ from snowflake.snowpark.version import VERSION
 __version__ = ".".join(str(x) for x in VERSION if x is not None)
 
 
+from snowflake.snowpark.async_job import AsyncJob
 from snowflake.snowpark.column import CaseExpr, Column
 from snowflake.snowpark.dataframe import DataFrame
 from snowflake.snowpark.dataframe_na_functions import DataFrameNaFunctions
--- a/src/snowflake/snowpark/__init__.py
+++ b/src/snowflake/snowpark/__init__.py
@@ -34,6 +34,8 @@ __all__ = [
     "WhenNotMatchedClause",
     "QueryRecord",
     "QueryHistory",
+    "AsyncJob",
+    "_AsyncDataType",
 ]
 
 
@@ -42,6 +44,7 @@ from snowflake.snowpark.version import VERSION
 __version__ = ".".join(str(x) for x in VERSION if x is not None)
 
 
+from snowflake.snowpark.async_job import AsyncJob, _AsyncDataType
 from snowflake.snowpark.column import CaseExpr, Column
 from snowflake.snowpark.dataframe import DataFrame
 from snowflake.snowpark.dataframe_na_functions import DataFrameNaFunctions
--- a/src/snowflake/snowpark/__init__.py
+++ b/src/snowflake/snowpark/__init__.py
@@ -34,8 +34,6 @@ __all__ = [
     "WhenNotMatchedClause",
     "QueryRecord",
     "QueryHistory",
-    "AsyncJob",
-    "_AsyncDataType",
 ]
 
 
@@ -44,7 +42,6 @@ from snowflake.snowpark.version import VERSION
 __version__ = ".".join(str(x) for x in VERSION if x is not None)
 
 
-from snowflake.snowpark.async_job import AsyncJob, _AsyncDataType
 from snowflake.snowpark.column import CaseExpr, Column
 from snowflake.snowpark.dataframe import DataFrame
 from snowflake.snowpark.dataframe_na_functions import DataFrameNaFunctions
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -42,7 +42,7 @@ from snowflake.snowpark._internal.utils import (
     result_set_to_rows,
     unwrap_stage_location_single_quote,
 )
-from snowflake.snowpark.async_job import AsyncDataType, AsyncJob
+from snowflake.snowpark.async_job import AsyncJob, _AsyncDataType
 from snowflake.snowpark.query_history import QueryHistory, QueryRecord
 from snowflake.snowpark.row import Row
 
@@ -309,7 +309,7 @@ class ServerConnection:
         to_iter: bool = False,
         is_ddl_on_temp_object: bool = False,
         block: bool = True,
-        data_type: AsyncDataType = AsyncDataType.ROW,
+        data_type: _AsyncDataType = _AsyncDataType.ROW,
         **kwargs,
     ) -> Union[Dict[str, Any], AsyncJob]:
         try:
@@ -394,13 +394,13 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
         block: bool = True,
-        data_type: AsyncDataType = AsyncDataType.ROW,
+        data_type: _AsyncDataType = _AsyncDataType.ROW,
         **kwargs,
     ) -> Union[
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
     ]:
         if is_in_stored_procedure() and not block:
-            raise ValueError("can not use async and stored procedure at the same time")
+            raise ValueError("async query isn't yet supported in stored procedure")
         result_set, result_meta = self.get_result_set(
             plan, to_pandas, to_iter, **kwargs, block=block, data_type=data_type
         )
@@ -421,7 +421,7 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
         block: bool = True,
-        data_type: AsyncDataType = AsyncDataType.ROW,
+        data_type: _AsyncDataType = _AsyncDataType.ROW,
         **kwargs,
     ) -> Tuple[
         Dict[
@@ -457,10 +457,9 @@ class ServerConnection:
                         data_type=data_type,
                         **kwargs,
                     )
-                    if block:
-                        placeholders[query.query_id_place_holder] = result["sfqid"]
-                    else:
-                        placeholders[query.query_id_place_holder] = result.query_id
+                    placeholders[query.query_id_place_holder] = (
+                        result["sfqid"] if block else result.query_id
+                    )
                     result_meta = self._cursor.description
                 if action_id < plan.session._last_canceled_id:
                     raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -42,7 +42,7 @@ from snowflake.snowpark._internal.utils import (
     result_set_to_rows,
     unwrap_stage_location_single_quote,
 )
-from snowflake.snowpark.async_job import AsyncJob
+from snowflake.snowpark.async_job import AsyncDataType, AsyncJob
 from snowflake.snowpark.query_history import QueryHistory, QueryRecord
 from snowflake.snowpark.row import Row
 
@@ -309,7 +309,7 @@ class ServerConnection:
         to_iter: bool = False,
         is_ddl_on_temp_object: bool = False,
         block: bool = True,
-        data_type: str = "row",
+        data_type: AsyncDataType = AsyncDataType.ROW,
         **kwargs,
     ) -> Union[Dict[str, Any], AsyncJob]:
         try:
@@ -383,7 +383,7 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
         block: bool = True,
-        data_type: str = "row",
+        data_type: AsyncDataType = AsyncDataType.ROW,
         **kwargs,
     ) -> Union[
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
@@ -408,7 +408,7 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
         block: bool = True,
-        data_type: str = "row",
+        data_type: AsyncDataType = AsyncDataType.ROW,
         **kwargs,
     ) -> Tuple[
         Dict[
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -441,6 +441,29 @@ class ServerConnection:
         result, result_meta = None, None
         try:
             placeholders = {}
+            if len(plan.queries) > 1 and not block:
+                sql = "begin "
+                for q in plan.queries:
+                    sql += q.sql + ";"
+                sql += " end;"
+                final_query = sql
+                for holder, id_ in placeholders.items():
+                    final_query = final_query.replace(holder, id_)
+                result = self.run_query(
+                    sql,
+                    to_pandas,
+                    to_iter,
+                    is_ddl_on_temp_object=plan.queries[0].is_ddl_on_temp_object,
+                    block=block,
+                    data_type=data_type,
+                    **kwargs,
+                )
+                placeholders[plan.queries[0].query_id_place_holder] = (
+                    result["sfqid"] if block else result.query_id
+                )
+                result_meta = self._cursor.description
+                if action_id < plan.session._last_canceled_id:
+                    raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
             for i, query in enumerate(plan.queries):
                 if isinstance(query, BatchInsertQuery):
                     self.run_batch_insert(query.sql, query.rows, **kwargs)
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -309,6 +309,7 @@ class ServerConnection:
         to_iter: bool = False,
         is_ddl_on_temp_object: bool = False,
         block: bool = True,
+        data_type: str = "row",
         **kwargs,
     ) -> Union[Dict[str, Any], AsyncJob]:
         try:
@@ -325,9 +326,11 @@ class ServerConnection:
                 logger.debug(f"Execute query [queryID: {results_cursor.sfqid}] {query}")
             else:
                 results_cursor = self._cursor.execute_async(query, **kwargs)
-                self.notify_query_listeners(QueryRecord(results_cursor.sfqid, query))
+                self.notify_query_listeners(
+                    QueryRecord(results_cursor["queryId"], query)
+                )
                 logger.debug(
-                    f"Execute query [queryID: {results_cursor['queryID']}] {query}"
+                    f"Execute query [queryID: {results_cursor['queryId']}] {query}"
                 )
         except Exception as ex:
             query_id_log = f" [queryID: {ex.sfqid}]" if hasattr(ex, "sfqid") else ""
@@ -340,7 +343,13 @@ class ServerConnection:
         # have non-select statements, and it shouldn't fail if the user
         # calls to_pandas() to execute the query.
         if not block:
-            return AsyncJob(results_cursor["queryId"], query, self._conn)
+            return AsyncJob(
+                results_cursor["queryId"],
+                query,
+                self._conn,
+                self._cursor.describe(query, **kwargs),
+                data_type,
+            )
         if to_pandas:
             try:
                 data_or_iter = (
@@ -374,16 +383,17 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
         block: bool = True,
+        data_type: str = "row",
         **kwargs,
     ) -> Union[
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
     ]:
         result_set, result_meta = self.get_result_set(
-            plan, to_pandas, to_iter, **kwargs
+            plan, to_pandas, to_iter, **kwargs, block=block, data_type=data_type
         )
         if not block:
             return result_set
-        if to_pandas:
+        elif to_pandas:
             return result_set["data"]
         else:
             if to_iter:
@@ -398,6 +408,7 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
         block: bool = True,
+        data_type: str = "row",
         **kwargs,
     ) -> Tuple[
         Dict[
@@ -430,11 +441,13 @@ class ServerConnection:
                         to_iter and (i == len(plan.queries) - 1),
                         is_ddl_on_temp_object=query.is_ddl_on_temp_object,
                         block=block,
+                        data_type=data_type,
                         **kwargs,
                     )
-                    placeholders[query.query_id_place_holder] = (
-                        result["sfqid"] if not block else result["queryID"]
-                    )
+                    if block:
+                        placeholders[query.query_id_place_holder] = result["sfqid"]
+                    else:
+                        placeholders[query.query_id_place_holder] = result.query_id
                     result_meta = self._cursor.description
                 if action_id < plan.session._last_canceled_id:
                     raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -346,10 +346,9 @@ class ServerConnection:
             return AsyncJob(
                 results_cursor["queryId"],
                 query,
-                self._conn,
                 self,
                 data_type,
-                kwargs,
+                **kwargs,
             )
         else:
             return self._to_data_or_iter(
@@ -452,16 +451,15 @@ class ServerConnection:
                     is_batch_insert = True
                     break
             if len(plan.queries) > 1 and not block and not is_batch_insert:
-                sql = f"""execute immediate $$
-declare
+                final_query = f"""EXECUTE IMMEDIATE $$
+DECLARE
     res resultset;
-begin
+BEGIN
 {";".join(q.sql for q in plan.queries[:-1])};
 res := ({plan.queries[-1].sql});
 return table(res);
-end;
+END;
 $$"""
-                final_query = sql
                 for holder, id_ in placeholders.items():
                     final_query = final_query.replace(holder, id_)
                 result = self.run_query(
@@ -478,7 +476,7 @@ $$"""
                     placeholders[q.query_id_place_holder] = (
                         result["sfqid"] if block else result.query_id
                     )
-                result_meta = self._cursor.description
+                result_meta = None
                 if action_id < plan.session._last_canceled_id:
                     raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
             else:
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -366,12 +366,12 @@ class ServerConnection:
                 data_or_iter = (
                     map(
                         functools.partial(
-                            self._fix_pandas_df_integer, results_cursor=results_cursor
+                            _fix_pandas_df_integer, results_cursor=results_cursor
                         ),
                         results_cursor.fetch_pandas_batches(),
                     )
                     if to_iter
-                    else self._fix_pandas_df_integer(
+                    else _fix_pandas_df_integer(
                         results_cursor.fetch_pandas_all(), results_cursor
                     )
                 )
@@ -404,7 +404,9 @@ class ServerConnection:
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
     ]:
         if is_in_stored_procedure() and not block:
-            raise ValueError("Async query is not supported in stored procedure yet")
+            raise NotImplementedError(
+                "Async query is not supported in stored procedure yet"
+            )
         result_set, result_meta = self.get_result_set(
             plan, to_pandas, to_iter, **kwargs, block=block, data_type=data_type
         )
@@ -450,14 +452,15 @@ class ServerConnection:
                 if isinstance(q, BatchInsertQuery):
                     is_batch_insert = True
                     break
+            # since batch insert does not support async execution (? in the query), we handle it separately here
             if len(plan.queries) > 1 and not block and not is_batch_insert:
                 final_query = f"""EXECUTE IMMEDIATE $$
 DECLARE
     res resultset;
 BEGIN
-{";".join(q.sql for q in plan.queries[:-1])};
-res := ({plan.queries[-1].sql});
-return table(res);
+    {";".join(q.sql for q in plan.queries[:-1])};
+    res := ({plan.queries[-1].sql});
+    return table(res);
 END;
 $$"""
                 for holder, id_ in placeholders.items():
@@ -476,6 +479,8 @@ $$"""
                     placeholders[q.query_id_place_holder] = (
                         result["sfqid"] if block else result.query_id
                     )
+                # since we will return a AsyncJob instance, result_meta is not needed, we will create reuslt_meta in
+                # AsyncJob instance when needed
                 result_meta = None
                 if action_id < plan.session._last_canceled_id:
                     raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
@@ -566,19 +571,20 @@ $$"""
             )
         logger.debug("Execute batch insertion query %s", query)
 
-    def _fix_pandas_df_integer(
-        self, pd_df: "pandas.DataFrame", results_cursor: SnowflakeCursor
-    ) -> "pandas.DataFrame":
-        for column_metadata, pandas_dtype, pandas_col_name in zip(
-            results_cursor.description, pd_df.dtypes, pd_df.columns
+
+def _fix_pandas_df_integer(
+    pd_df: "pandas.DataFrame", results_cursor: SnowflakeCursor
+) -> "pandas.DataFrame":
+    for column_metadata, pandas_dtype, pandas_col_name in zip(
+        results_cursor.description, pd_df.dtypes, pd_df.columns
+    ):
+        if (
+            FIELD_ID_TO_NAME.get(column_metadata.type_code) == "FIXED"
+            and column_metadata.precision is not None
+            and column_metadata.scale == 0
+            and not str(pandas_dtype).startswith("int")
         ):
-            if (
-                FIELD_ID_TO_NAME.get(column_metadata.type_code) == "FIXED"
-                and column_metadata.precision is not None
-                and column_metadata.scale == 0
-                and not str(pandas_dtype).startswith("int")
-            ):
-                pd_df[pandas_col_name] = pandas.to_numeric(
-                    pd_df[pandas_col_name], downcast="integer"
-                )
-        return pd_df
+            pd_df[pandas_col_name] = pandas.to_numeric(
+                pd_df[pandas_col_name], downcast="integer"
+            )
+    return pd_df
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -42,6 +42,7 @@ from snowflake.snowpark._internal.utils import (
     result_set_to_rows,
     unwrap_stage_location_single_quote,
 )
+from snowflake.snowpark.async_job import AsyncJob
 from snowflake.snowpark.query_history import QueryHistory, QueryRecord
 from snowflake.snowpark.row import Row
 
@@ -309,7 +310,7 @@ class ServerConnection:
         is_ddl_on_temp_object: bool = False,
         async_: bool = False,
         **kwargs,
-    ) -> Dict[str, Any]:
+    ) -> Union[Dict[str, Any], AsyncJob]:
         try:
             # Set SNOWPARK_SKIP_TXN_COMMIT_IN_DDL to True to avoid DDL commands to commit the open transaction
             if is_ddl_on_temp_object:
@@ -335,7 +336,7 @@ class ServerConnection:
         # have non-select statements, and it shouldn't fail if the user
         # calls to_pandas() to execute the query.
         if async_:
-            return results_cursor
+            return AsyncJob(results_cursor["queryId"], query, self._conn)
         if to_pandas:
             try:
                 data_or_iter = (
@@ -427,7 +428,9 @@ class ServerConnection:
                         async_=async_,
                         **kwargs,
                     )
-                    placeholders[query.query_id_place_holder] = result["sfqid"]
+                    placeholders[query.query_id_place_holder] = (
+                        result["sfqid"] if not async_ else None
+                    )
                     result_meta = self._cursor.description
                 if action_id < plan.session._last_canceled_id:
                     raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -463,8 +463,13 @@ BEGIN
     return table(res);
 END;
 $$"""
-                for holder, id_ in placeholders.items():
-                    final_query = final_query.replace(holder, id_)
+                # In multiple queries scenario, we are unable to get the query id of former query, so we replace
+                # place holder with fucntion last_query_id() here
+                for q in plan.queries:
+                    final_query = final_query.replace(
+                        f"'{q.query_id_place_holder}'", "LAST_QUERY_ID()"
+                    )
+
                 result = self.run_query(
                     final_query,
                     to_pandas,
@@ -475,10 +480,7 @@ $$"""
                     **kwargs,
                 )
                 result.query = final_query
-                for q in plan.queries:
-                    placeholders[q.query_id_place_holder] = (
-                        result["sfqid"] if block else result.query_id
-                    )
+
                 # since we will return a AsyncJob instance, result_meta is not needed, we will create reuslt_meta in
                 # AsyncJob instance when needed
                 result_meta = None
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -400,7 +400,7 @@ class ServerConnection:
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
     ]:
         if is_in_stored_procedure() and not block:
-            raise ValueError("async query isn't yet supported in stored procedure")
+            raise ValueError("Async query is not supported in stored procedure yet")
         result_set, result_meta = self.get_result_set(
             plan, to_pandas, to_iter, **kwargs, block=block, data_type=data_type
         )
@@ -450,7 +450,7 @@ class ServerConnection:
                 for holder, id_ in placeholders.items():
                     final_query = final_query.replace(holder, id_)
                 result = self.run_query(
-                    sql,
+                    final_query,
                     to_pandas,
                     to_iter,
                     is_ddl_on_temp_object=plan.queries[0].is_ddl_on_temp_object,
@@ -458,34 +458,36 @@ class ServerConnection:
                     data_type=data_type,
                     **kwargs,
                 )
-                placeholders[plan.queries[0].query_id_place_holder] = (
-                    result["sfqid"] if block else result.query_id
-                )
+                for q in plan.queries:
+                    placeholders[q.query_id_place_holder] = (
+                        result["sfqid"] if block else result.query_id
+                    )
                 result_meta = self._cursor.description
                 if action_id < plan.session._last_canceled_id:
                     raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
-            for i, query in enumerate(plan.queries):
-                if isinstance(query, BatchInsertQuery):
-                    self.run_batch_insert(query.sql, query.rows, **kwargs)
-                else:
-                    final_query = query.sql
-                    for holder, id_ in placeholders.items():
-                        final_query = final_query.replace(holder, id_)
-                    result = self.run_query(
-                        final_query,
-                        to_pandas,
-                        to_iter and (i == len(plan.queries) - 1),
-                        is_ddl_on_temp_object=query.is_ddl_on_temp_object,
-                        block=block,
-                        data_type=data_type,
-                        **kwargs,
-                    )
-                    placeholders[query.query_id_place_holder] = (
-                        result["sfqid"] if block else result.query_id
-                    )
-                    result_meta = self._cursor.description
-                if action_id < plan.session._last_canceled_id:
-                    raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
+            else:
+                for i, query in enumerate(plan.queries):
+                    if isinstance(query, BatchInsertQuery):
+                        self.run_batch_insert(query.sql, query.rows, **kwargs)
+                    else:
+                        final_query = query.sql
+                        for holder, id_ in placeholders.items():
+                            final_query = final_query.replace(holder, id_)
+                        result = self.run_query(
+                            final_query,
+                            to_pandas,
+                            to_iter and (i == len(plan.queries) - 1),
+                            is_ddl_on_temp_object=query.is_ddl_on_temp_object,
+                            block=block,
+                            data_type=data_type,
+                            **kwargs,
+                        )
+                        placeholders[query.query_id_place_holder] = (
+                            result["sfqid"] if block else result.query_id
+                        )
+                        result_meta = self._cursor.description
+                    if action_id < plan.session._last_canceled_id:
+                        raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
         finally:
             # delete created tmp object
             for action in plan.post_actions:
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -348,8 +348,19 @@ class ServerConnection:
                 query,
                 self._conn,
                 self._cursor.describe(query, **kwargs),
+                self,
                 data_type,
             )
+        return self._to_data_or_iter(
+            results_cursor=results_cursor, to_pandas=to_pandas, to_iter=to_iter
+        )
+
+    def _to_data_or_iter(
+        self,
+        results_cursor: SnowflakeCursor,
+        to_pandas: bool = False,
+        to_iter: bool = False,
+    ) -> Dict[str, Any]:
         if to_pandas:
             try:
                 data_or_iter = (
@@ -388,6 +399,8 @@ class ServerConnection:
     ) -> Union[
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
     ]:
+        if is_in_stored_procedure() and not block:
+            raise ValueError("can not use async and stored procedure at the same time")
         result_set, result_meta = self.get_result_set(
             plan, to_pandas, to_iter, **kwargs, block=block, data_type=data_type
         )
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -308,7 +308,7 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
         is_ddl_on_temp_object: bool = False,
-        async_: bool = False,
+        block: bool = True,
         **kwargs,
     ) -> Union[Dict[str, Any], AsyncJob]:
         try:
@@ -317,7 +317,7 @@ class ServerConnection:
                 if not kwargs.get("_statement_params"):
                     kwargs["_statement_params"] = {}
                 kwargs["_statement_params"]["SNOWPARK_SKIP_TXN_COMMIT_IN_DDL"] = True
-            if not async_:
+            if block:
                 results_cursor = self._cursor.execute(query, **kwargs)
                 self.notify_query_listeners(
                     QueryRecord(results_cursor.sfqid, results_cursor.query)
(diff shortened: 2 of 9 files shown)


src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -414,9 +425,12 @@ def get_result_set(
                         to_pandas,
                         to_iter and (i == len(plan.queries) - 1),
                         is_ddl_on_temp_object=query.is_ddl_on_temp_object,
+                        async_=async_,
                         **kwargs,
                     )
-                    placeholders[query.query_id_place_holder] = result["sfqid"]
+                    placeholders[query.query_id_place_holder] = (
+                        result["sfqid"] if not async_ else None
```
sfc-gh-jdu, line 432: why can't async job also have query id?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -362,13 +369,16 @@ def execute(
...
         **kwargs,
     ) -> Union[
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
     ]:
         result_set, result_meta = self.get_result_set(
             plan, to_pandas, to_iter, **kwargs
         )
+        if async_:
+            return result_set
         if to_pandas:
```
sfc-gh-jdu, line 382: nit: probably use `elif` here for consistency?

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2012-2022 Snowflake Computing Inc. All rights reserved.
+#
+from snowflake.snowpark._internal.server_connection import SnowflakeConnection
```
sfc-gh-jdu, line 5:
```suggestion
from snowflake.connector import SnowflakeConnection
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class AsyncJob:
+    def __init__(
+        self,
+        query_id: str = None,
```
sfc-gh-jdu, line 11: can it be none? If not, remove default value. And the same for query and conn, so you don't need to any check in the function.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,45 @@
...
+        self._conn = conn
+        self._cursor = self._conn.cursor()
+
+        if not conn:
+            raise ValueError("connection cannot be empty")
+        return
+
+    # iterator
+    def __iter__(self):
+        pass
```
sfc-gh-jdu, line 26: you will implement it, right?

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,45 @@
...
+        return not is_running
+
+    def cancel(self) -> None:
+        # stop and cancel current query id
+        self._conn._cancel_query(self.query, self.query_id)
+
+    def result(self):
+        # return result of the query, in the form of a list of Row object
+        self._cursor.get_results_from_sfqid(self.query_id)
+        return self._cursor.fetchall()
```
sfc-gh-jdu, line 45: fetchall won't give you a list rows. Also, I think we should maintain a private property in this class, which is used for indicating where async job comes from: if it's from collect(), return a list of rows; if it's from to_pandas, return a pandas df; if it's from to_local_iterator, return an iterator, etc. Then in this way, probably we don't need to implement `__iter__` anymore.

src/snowflake/snowpark/dataframe.py
```diff
@@ -451,14 +451,28 @@ def collect(
             statement_params=statement_params
         )
 
-    def _internal_collect_with_tag_no_telemetry(
+    @df_collect_api_telemetry
+    def collect_nowait(
         self, *, statement_params: Optional[Dict[str, str]] = None
+    ) -> List["Row"]:
```
sfc-gh-jdu, line 457:
```suggestion
    ) -> AsyncJob:
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -451,14 +451,28 @@ def collect(
...
+        """Executes the query representing this DataFrame asynchronously and returns query id(asyncjob instance).
```
sfc-gh-jdu, line 458:
```suggestion
        """Executes the query representing this DataFrame asynchronously and returns :class:`AsyncJob`.
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,53 @@
...
+        result_meta: List[ResultMetadata],
+        data_type: str = "row",
+    ) -> None:
+        self.query_id = query_id
+        self.query = query
+        self._conn = conn
+        self._cursor = self._conn.cursor()
+        self._data_type = data_type
+        self._result_meta = result_meta
+        if not conn:
```
sfc-gh-jdu, line 27: you don't need this check

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,53 @@
...
+            raise ValueError("connection cannot be empty")
+        return
```
sfc-gh-jdu, line 29: remove

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -414,9 +440,14 @@ def get_result_set(
                         to_pandas,
                         to_iter and (i == len(plan.queries) - 1),
                         is_ddl_on_temp_object=query.is_ddl_on_temp_object,
+                        block=block,
+                        data_type=data_type,
                         **kwargs,
                     )
-                    placeholders[query.query_id_place_holder] = result["sfqid"]
+                    if block:
```
sfc-gh-jdu, line 447: maybe 
```
placeholders[query.query_id_place_holder] = result["sfqid"] if block else result.query_id
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -451,14 +452,37 @@ def collect(
             statement_params=statement_params
         )
 
+    @df_collect_api_telemetry
+    def collect_nowait(
+        self,
+        *,
+        statement_params: Optional[Dict[str, str]] = None,
+        data_type: str = "row",
```
sfc-gh-jdu, line 460: this argument is not provided by the user: collect() will return rows and to_pandas() will return a pandas dataframe.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def cancel(self) -> None:
+        # stop and cancel current query id
+        self._conn._cancel_query(self.query, self.query_id)
+
+    def result(self):
+        # return result of the query, in the form of a list of Row object
+        self._cursor.get_results_from_sfqid(self.query_id)
+        result_data = self._cursor.fetchall()
+        if self._data_type == "row":
```
sfc-gh-jdu, line 46: You can try to use Enum to have these types. Similar to https://github.com/snowflakedb/snowpark-python/blob/34a28a5f541bb9635d769381d477c2f2dea56873/src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py#L47

sfc-gh-jdu (commented):
You need to some tests for your APIs.

sfc-gh-sfan (commented):
Let's also add an explicit check that this API should not be invoked in stored proc mode?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -307,19 +308,30 @@ def run_query(
...
+                    QueryRecord(results_cursor.sfqid, results_cursor.query)
+                )
+                logger.debug(f"Execute query [queryID: {results_cursor.sfqid}] {query}")
+            else:
+                results_cursor = self._cursor.execute_async(query, **kwargs)
+                self.notify_query_listeners(
+                    QueryRecord(results_cursor["queryId"], query)
+                )
+                logger.debug(
+                    f"Execute query [queryID: {results_cursor['queryId']}] {query}"
```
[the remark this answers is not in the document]
sfc-gh-yuwang: of course

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-yuwang: I tried this one. But if return value is a asyncjob instance . will raise error that 'asyncjob class is not subscribable'

src/snowflake/snowpark/dataframe.py
sfc-gh-yuwang: yes, i am think about passing this arg to asyncjob instance inside the corresponding function like collect() that calls it. In this way, users will not see or operate on this arg

sfc-gh-yuwang:
> Let's also add an explicit check that this API should not be invoked in stored proc mode?

Sure. will do!

src/snowflake/snowpark/async_job.py
sfc-gh-yuwang: sure, will do

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-jdu: no this comments means you can simplify your code, no functional change. So if you get an error, it means your original code is not correct.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -307,19 +308,30 @@ def run_query(
         to_pandas: bool = False,
         to_iter: bool = False,
         is_ddl_on_temp_object: bool = False,
+        block: bool = True,
+        data_type: AsyncDataType = AsyncDataType.ROW,
```
sfc-gh-yixie, line 312: Better put them into kwargs

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -362,14 +393,20 @@ def execute(
...
         to_pandas: bool = False,
         to_iter: bool = False,
+        block: bool = True,
+        data_type: AsyncDataType = AsyncDataType.ROW,
         **kwargs,
     ) -> Union[
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
     ]:
+        if is_in_stored_procedure() and not block:
+            raise ValueError("can not use async and stored procedure at the same time")
```
sfc-gh-yixie, line 403: How about "async query isn't yet supported in stored procedure."

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,68 @@
...
+from snowflake.connector import SnowflakeConnection
+from snowflake.connector.cursor import ResultMetadata
+from snowflake.snowpark._internal.utils import result_set_to_iter, result_set_to_rows
+
+
+class AsyncDataType(Enum):
+    PANDAS = "pandas"
+    ROW = "row"
+    ITERATOR = "iterator"
+    PANDAS_BATCH = "pandas_batch"
```
sfc-gh-yixie, line 17: Do users need to use this Enum? If no we better hide it.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-yuwang: got it. turns out I got a wrong if last time.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-yuwang: sure, will do

src/snowflake/snowpark/async_job.py
sfc-gh-yuwang: Do you mean that I should put this in other place like _internal.utils ?

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-yuwang: I think we should keep it this way because kwargs will be passed into cursor.execute() in the end while these args are not needed in it.

src/snowflake/snowpark/table.py
```diff
@@ -519,9 +531,18 @@ def merge(
...
         add_api_call(new_df, "Table.update")
-        return _get_merge_result(
-            new_df._internal_collect_with_tag(statement_params=statement_params),
-            inserted=inserted,
-            updated=updated,
-            deleted=deleted,
+        result = new_df._internal_collect_with_tag(
+            statement_params=statement_params,
+            block=block,
+            data_type=f"{str(snowflake.snowpark._AsyncDataType.MERGE.value)}_{int(inserted)}{int(updated)}{int(deleted)}",
```
sfc-gh-jdu, line 537: do not pass in the data in this way (alternatively add more enum types, or add one more argument to _internal_collect_with_tag, etc.)

tests/integ/scala/test_async_job_suite.py
```diff
@@ -0,0 +1,267 @@
...
+    Utils.upload_to_stage(
+        session, "@" + tmp_stage_name1, test_files.test_file_csv, compress=False
+    )
+    df = session.read.schema(user_schema).csv(f"@{tmp_stage_name1}/{test_file_csv}")
+    assert len(df.queries) > 1
+    res = df.collect_nowait()
+    Utils.check_answer(res.result(), df.collect())
+
+
+def test_async_is_running_and_cancel(session):
```
sfc-gh-jdu, line 266: you will add the implementation later, right?

tests/integ/scala/test_async_job_suite.py
```diff
@@ -0,0 +1,267 @@
...
+    pass
```
sfc-gh-jdu, line 267: You need a test to test batch insert case (though scala might not have it, we should have). You can refer to https://github.com/snowflakedb/snowpark-python/blob/ef0e8320feb374aab1949ce1b40fb69235fc4c24/tests/integ/test_dataframe.py#L1756-L1762

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -402,6 +441,29 @@ def get_result_set(
         result, result_meta = None, None
         try:
             placeholders = {}
+            if len(plan.queries) > 1 and not block:
+                sql = "begin "
```
sfc-gh-jdu, line 445: I would prefer to build a query string in this way:
```
sql = f"""BEGIN
{";\n".join(q.sql for q in plan.queries)};
END;
"""

```

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -402,6 +441,29 @@ def get_result_set(
...
+                for q in plan.queries:
+                    sql += q.sql + ";"
+                sql += " end;"
+                final_query = sql
+                for holder, id_ in placeholders.items():
+                    final_query = final_query.replace(holder, id_)
+                result = self.run_query(
+                    sql,
```
sfc-gh-jdu, line 453:
```suggestion
                    final_query,
```

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -402,6 +441,29 @@ def get_result_set(
...
+                    to_pandas,
+                    to_iter,
+                    is_ddl_on_temp_object=plan.queries[0].is_ddl_on_temp_object,
+                    block=block,
+                    data_type=data_type,
+                    **kwargs,
+                )
+                placeholders[plan.queries[0].query_id_place_holder] = (
```
sfc-gh-jdu, line 461: you should do 
```suggestion
                for q in plan.queries:
                    placeholders[q.query_id_place_holder] = (
```

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -402,6 +441,29 @@ def get_result_set(
         result, result_meta = None, None
         try:
             placeholders = {}
+            if len(plan.queries) > 1 and not block:
```
sfc-gh-jdu, line 444: You should consider batch insertion case

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -402,6 +441,29 @@ def get_result_set(
...
+                    data_type=data_type,
+                    **kwargs,
+                )
+                placeholders[plan.queries[0].query_id_place_holder] = (
+                    result["sfqid"] if block else result.query_id
+                )
+                result_meta = self._cursor.description
+                if action_id < plan.session._last_canceled_id:
+                    raise SnowparkClientExceptionMessages.SERVER_QUERY_IS_CANCELLED()
             for i, query in enumerate(plan.queries):
```
sfc-gh-jdu, line 467: shouldn't you have an else block here? Otherwise for async multiple queries, this is also executed?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -426,6 +492,7 @@ def get_result_set(
                 self.run_query(
```
sfc-gh-jdu, line 492: hmm I just realized that plan.post_actions should be executed after the async query is done and should be executed here. We can discuss more tomorrow.

src/snowflake/snowpark/__init__.py
```diff
@@ -34,8 +34,6 @@
     "WhenNotMatchedClause",
     "QueryRecord",
     "QueryHistory",
-    "AsyncJob",
```
sfc-gh-jdu: AsyncJob is still needed, but _AsyncDataType is not needed. Any module prefixed with _ is considered for internal usage only.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,132 @@
...
+        self._data_type = data_type
+        self._parameters = parameters
+        self._result_meta = None
+        self._server_connection = server_connection
+        self._inserted = False
+        self._updated = False
+        self._deleted = False
+        self._plan = None
+
+        return
```
sfc-gh-jdu, line 53: remove

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,132 @@
...
+            if self._updated:
+                rows_updated = int(result_data[0][idx])
+                idx += 1
+            if self._deleted:
+                rows_deleted = int(result_data[0][idx])
+            return snowflake.snowpark.MergeResult(
+                rows_inserted, rows_updated, rows_deleted
+            )
+
+    def result(self) -> Union[List[Row], pandas.DataFrame, Iterator[Row], int]:
```
sfc-gh-jdu, line 94:
```suggestion
    def result(self) -> Union[List[Row], "pandas.DataFrame", Iterator[Row], int, None]:
```

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -330,15 +342,39 @@ def run_query(
         # because when the query plan has multiple queries, it will
         # have non-select statements, and it shouldn't fail if the user
         # calls to_pandas() to execute the query.
+        if not block:
+            return AsyncJob(
+                results_cursor["queryId"],
+                query,
+                self._conn,
+                kwargs,
```
sfc-gh-jdu, line 350: you can do 
```
return AsyncJob(
                results_cursor["queryId"],
                query,
                self._conn,
                self,
                data_type,
                **kwargs,
```
and change the constructor of AsyncJob as well.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -402,32 +446,77 @@ def get_result_set(
...
+                if isinstance(q, BatchInsertQuery):
+                    is_batch_insert = True
+                    break
+            if len(plan.queries) > 1 and not block and not is_batch_insert:
+                sql = f"""execute immediate $$
+declare
+    res resultset;
+begin
+{";".join(q.sql for q in plan.queries[:-1])};
+res:=({plan.queries[-1].sql});
```
sfc-gh-jdu, line 460:
```suggestion
res := ({plan.queries[-1].sql});
```

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -402,32 +446,77 @@ def get_result_set(
...
+                result = self.run_query(
+                    final_query,
+                    to_pandas,
+                    to_iter,
+                    is_ddl_on_temp_object=plan.queries[0].is_ddl_on_temp_object,
+                    block=block,
+                    data_type=data_type,
+                    **kwargs,
+                )
+                result.query = plan.queries[-1].sql
```
sfc-gh-jdu, line 476: Shouldn't it be 
```
result.query = final_query
```
?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -402,32 +446,77 @@ def get_result_set(
...
+                for q in plan.queries:
+                    placeholders[q.query_id_place_holder] = (
+                        result["sfqid"] if block else result.query_id
                     )
-                    placeholders[query.query_id_place_holder] = result["sfqid"]
-                    result_meta = self._cursor.description
+                result_meta = self._cursor.description
```
sfc-gh-jdu, line 481: You don't need result_meta here for async query, right? If not, set it to None.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,132 @@
...
+            elif self._data_type == _AsyncDataType.COUNT:
+                result = result_data[0][0]
+            elif self._data_type in [
+                _AsyncDataType.UPDATE,
+                _AsyncDataType.DELETE,
+                _AsyncDataType.MERGE,
+            ]:
+                result = self._table_result(result_data)
+            else:
+                raise ValueError(f"{self._data_type} is not a supported data type")
```
sfc-gh-jdu, line 125:
```suggestion
                raise ValueError(f"{self._data_type} is not supported")
```

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -522,3 +523,14 @@ def func_call_wrapper(*args, **kwargs):
         return func_call_wrapper
 
     return deprecate_wrapper
+
+
+def check_is_pandas_dataframe(result):
```
sfc-gh-jdu, line 528:
```suggestion
def check_is_pandas_dataframe_in_to_pandas(result):
```
more explicit

src/snowflake/snowpark/dataframe.py
```diff
@@ -563,7 +588,7 @@ def to_pandas_batches(
         *,
         statement_params: Optional[Dict[str, str]] = None,
         **kwargs: Dict[str, Any],
-    ) -> Iterator["pandas.DataFrame"]:
+    ) -> Union[Iterator["pandas.DataFrame"], AsyncJob]:
```
sfc-gh-jdu, line 591: since we'll not support it right now, remove all changes in this method

tests/integ/scala/test_async_job_suite.py
```diff
@@ -0,0 +1,297 @@
...
+        [
+            StructField("a", DoubleType()),
+            StructField("b", IntegerType()),
+            StructField("c", IntegerType()),
+        ]
+    )
+    # check if copy is successful
+    async_job = df.write.copy_into_location(remote_location, block=False)
+    res = async_job.result()
+    Utils.check_answer(res, [Row(rows_unloaded=4, input_bytes=27, output_bytes=47)])
```
sfc-gh-jdu, line 242: let's only check rows_unloaded because other two might be changed in the future and make the test flaky.

tests/integ/scala/test_async_job_suite.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+    test_files = TestFiles(resources_path)
+    Utils.create_stage(session, tmp_stage_name1, is_temporary=True)
+    Utils.upload_to_stage(
+        session, "@" + tmp_stage_name1, test_files.test_file_csv, compress=False
+    )
+    df = session.read.schema(user_schema).csv(f"@{tmp_stage_name1}/{test_file_csv}")
+    assert len(df._plan.queries) > 1
+    res = df.collect_nowait()
+    Utils.check_answer(res.result(), df.collect())
```
sfc-gh-jdu, line 265:
```suggestion
    Utils.check_answer(res.result(), df)
```

tests/integ/scala/test_async_job_suite.py
```diff
@@ -0,0 +1,297 @@
...
+        sleep(1.0)
+    assert async_job.is_done()
+
+    async_job2 = session.sql("select SYSTEM$WAIT(10)").collect_nowait()
+    assert not async_job2.is_done()
+    async_job2.cancel()
+    start = time()
+    while not async_job2.is_done():
+        sleep(1.0)
+    assert (time() - start) / 1000 < 60
```
sfc-gh-jdu, line 296: will it be flaky? like cancellation takes more than 1s

src/snowflake/snowpark/__init__.py
sfc-gh-jdu: @sfc-gh-yuwang ^ this one

sfc-gh-yuwang: got it

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -330,15 +342,39 @@ def run_query(
         # because when the query plan has multiple queries, it will
         # have non-select statements, and it shouldn't fail if the user
         # calls to_pandas() to execute the query.
+        if not block:
+            return AsyncJob(
+                results_cursor["queryId"],
+                query,
+                self._conn,
```
sfc-gh-jdu, line 349: If you pass in the server connection object, you don't need to pass in self._conn.

src/snowflake/snowpark/dataframe.py
```diff
@@ -439,26 +441,48 @@ def stat(self) -> DataFrameStatFunctions:
...
+        )
+
+    @df_collect_api_telemetry
+    def collect_nowait(
+        self,
+        *,
+        statement_params: Optional[Dict[str, str]] = None,
+    ) -> AsyncJob:
+        """Executes the query representing this DataFrame asynchronously and returns: class:'AsyncJob'.
+
```
sfc-gh-jdu, line 464: Let's add some examples of using async apis in this method

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,130 @@
...
+    ITERATOR = "iterator"
+    PANDAS_BATCH = "pandas_batch"
+    COUNT = "count"
+    NONE_TYPE = "no_type"
+    UPDATE = "update"
+    DELETE = "delete"
+    MERGE = "merge"
+
+
+class AsyncJob:
```
sfc-gh-jdu, line 30: Let's add docstrings for this class, including the introduction, examples, caveats and notes. You can refer to other files for examples.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+        cancel this asynchronously executed query
+        retrieve the query id and perform other actions on this query id manually
+
+    AsyncJob is strongly related to dataframe functions, to use AsyncJob, you need to create a dataframe first
+
+    create dataframe:
+        >>> df = session.createDataFrame([[float(4), 3, 5], [2.0, -4, 7], [3.0, 5, 6],[4.0,6,8]], schema=["a", "b", "c"])
+
+    Example 1
+        df.collect() can be performed asynchronously::
```
sfc-gh-jdu, line 44:
```suggestion
        :meth:`DataFrame.collect` can be performed asynchronously::
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+
+            >>> async_job = df.collect_nowait()
+            >>> async_job.result()
+            [Row(A=4.0, B=3, C=5), Row(A=2.0, B=-4, C=7), Row(A=3.0, B=5, C=6), Row(A=4.0, B=6, C=8)]
+
+        you can also do::
```
sfc-gh-jdu, line 50:
```suggestion
        You can also do::
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+            >>> async_job = df.collect(block=False)
+            >>> async_job.result()
+            [Row(A=4.0, B=3, C=5), Row(A=2.0, B=-4, C=7), Row(A=3.0, B=5, C=6), Row(A=4.0, B=6, C=8)]
+
+    Example 2
+        df.to_pandas can be performed asynchronously::
```
sfc-gh-jdu, line 56:
```suggestion
        :meth:`DataFrame.to_pandas` can be performed asynchronously::
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+            >>> async_job = df.to_pandas(block=False)
+            >>> async_job.result()
+                 A  B  C
+            0  4.0  3  5
+            1  2.0 -4  7
+            2  3.0  5  6
+            3  4.0  6  8
+
+    Example 3
+        df.first() can be performed asynchronously::
```
sfc-gh-jdu, line 67:
```suggestion
        :meth:`DataFrame.first` can be performed asynchronously::
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+
+            >>> async_job = df.first(block=False)
+            >>> async_job.result()
+            [Row(A=4.0, B=3, C=5)]
+
+    Example 4
+        df.count() can be performed asynchronously::
```
sfc-gh-jdu, line 74:
```suggestion
        :meth:`DataFrame.count` can be performed asynchronously::
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+
+            >>> async_job = df.count(block=False)
+            >>> async_job.result()
+            4
+
+    Example 5
+        save dataframe to table or copy it into other locations can also be performed asynchronously::
```
sfc-gh-jdu, line 81:
```suggestion
        Save a dataframe to a table or copy it into a stage file asynchronously::
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+
+            >>> table_name = "name"
+            >>> async_job = df.write.save_as_table(table_name, block=False)
+            >>> # copy into other location
```
sfc-gh-jdu, line 85:
```suggestion
            >>> # copy into a stage file
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+            >>> remote_location = f"{session.get_session_stage()}/name.csv"
+            >>> async_job = df.write.copy_into_location(remote_location, block=False)
```
sfc-gh-jdu, line 87: show the result

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Example 7
+        table operations like merge, update and delete can also be perfromed asynchronously::
```
sfc-gh-jdu, line 90:
```suggestion
        :meth:`Table.merge`, :meth:`Table.update` and :meth:`Table.delete` can be performed asynchronously::
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+
+            >>> target_df = session.create_dataframe([(10, "old"), (10, "too_old"), (11, "old")], schema=["key", "value"])
+            >>> target_df.write.save_as_table("my_table", mode="overwrite", table_type="temporary")
+            >>> target = session.table("my_table")
+            >>> source = session.create_dataframe([(10, "new"), (12, "new"), (13, "old")], schema=["key", "value"])
+            >>> target.merge(source,target["key"] == source["key"],[when_matched().update({"value": source["value"]}),when_not_matched().insert({"key": source["key"]}),],block=False,)
```
sfc-gh-jdu, line 96:
```suggestion
            >>> target.merge(source,target["key"] == source["key"],[when_matched().update({"value": source["value"]}),when_not_matched().insert({"key": source["key"]})],block=False)
```
You need to import when_matched and when_not_matched, right? Also it returns an AsyncJob and you need to show the results.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Example 8
+        AsyncJob also allows to cancel the query before it is finished::
```
sfc-gh-jdu, line 99:
```suggestion
        Cancel the running query associated with the dataframe::
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+        self._parameters = kwargs
+        self._result_meta = None
+        self._server_connection = server_connection
+        self._inserted = False
+        self._updated = False
+        self._deleted = False
+        self._plan = None
+
+    def is_done(self) -> bool:
+        # return a bool value to indicate whether the query is finished
```
sfc-gh-jdu, line 128: You need docstrings for all public methods in this class.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,204 @@
...
+            >>> source = session.create_dataframe([(10, "new"), (12, "new"), (13, "old")], schema=["key", "value"])
+            >>> target.merge(source,target["key"] == source["key"],[when_matched().update({"value": source["value"]}),when_not_matched().insert({"key": source["key"]}),],block=False,)
+
+    Example 8
+        AsyncJob also allows to cancel the query before it is finished::
+
+            >>> df = session.sql("select SYSTEM$WAIT(3)")
+            >>> async_job = df.collect_nowait()
+            >>> async_job.cancel()
+    """
```
sfc-gh-jdu, line 104: You need a note section to mention the caveats:
```
Note:
    - If a dataframe is associated with multiple queries,
        + if you use `:meth:Session.create_dataframe` to create a dataframe from a large amount of local data and evaluate this dataframe asynchronously, data will still be loaded into Snowflake synchronously, and only fetching data from Snowflake again will be performed asynchronously. 
        + otherwise, multiple queries will be wrapped into a `Snowflake Anonymous Block <https://docs.snowflake.com/en/developer-guide/snowflake-scripting/blocks.html#using-an-anonymous-block>`_ and executed asynchronously as one query. 
    - Temporary objects (e.g., tables) might be created when evaluating dataframes and they will be dropped automatically after all queries finish when calling a synchronous API. When you evaluating dataframes asynchronously, temporary objects will only be dropped after calling :meth:`result`. 
```
Or you can move the second bullet point into `result` method.

src/snowflake/snowpark/dataframe.py
```diff
@@ -406,6 +408,25 @@ class DataFrame:
             >>> df = session.create_dataframe([[1, 2], [3, 4], [5, -1]], schema=["a", "b"])
             >>> df.stat.corr("a", "b")
             -0.5960395606792697
+
+    Example 16
+        Performing a query and returning an array of Rows asynchronously::
+
+            >>> df = session.createDataFrame([[float(4), 3, 5], [2.0, -4, 7], [3.0, 5, 6],[4.0,6,8]], schema=["a", "b", "c"])
```
sfc-gh-jdu, line 415:
```suggestion
            >>> df = session.create_dataframe([[float(4), 3, 5], [2.0, -4, 7], [3.0, 5, 6], [4.0, 6, 8]], schema=["a", "b", "c"])
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -406,6 +408,25 @@ class DataFrame:
             >>> df = session.create_dataframe([[1, 2], [3, 4], [5, -1]], schema=["a", "b"])
             >>> df.stat.corr("a", "b")
             -0.5960395606792697
+
+    Example 16
+        Performing a query and returning an array of Rows asynchronously::
```
sfc-gh-jdu, line 413:
```suggestion
        Performing a query asynchronously and returning a list of :class:`Row` objects::
```

sfc-gh-jdu (commented):
You need to build the doc to see whether the generated doc is correct and expected.

tests/integ/scala/test_async_job_suite.py
```diff
@@ -0,0 +1,297 @@
...
+        sleep(1.0)
+    assert async_job.is_done()
+
+    async_job2 = session.sql("select SYSTEM$WAIT(10)").collect_nowait()
+    assert not async_job2.is_done()
+    async_job2.cancel()
+    start = time()
+    while not async_job2.is_done():
+        sleep(1.0)
+    assert (time() - start) < 10
```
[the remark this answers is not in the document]
sfc-gh-yuwang: means if you cancel it, it takes less time then the original running time

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+    PANDAS_BATCH = "pandas_batch"
+    COUNT = "count"
+    NONE_TYPE = "no_type"
+    UPDATE = "update"
+    DELETE = "delete"
+    MERGE = "merge"
+
+
+class AsyncJob:
+    """AsyncJob is strongly related to dataframe functions, to use AsyncJob, you need to create a dataframe first
```
sfc-gh-jdu, line 30: Move this sentence after introducing usage.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+    Provides a way to track an asynchronous query in Snowflake. A :class:`DataFrame` object can be evaluated asynchronously and an :class:`AsyncJob` object will be returned. With this instance, you can do:
+    - retrieve results;
+    - check the query status (still running or done);
+    - cancel the running query;
+    - retrieve the query ID and perform other operations on this query ID manually.
+    We demonstrate how to evaluate a :class:`DataFrame` asynchronously and use returned :class:`AsyncJob` object:
+
+
+    First, we create a dataframe:
+        >>> from snowflake.snowpark import Session
```
sfc-gh-jdu, line 40: I don't think this import is needed?

src/snowflake/snowpark/__init__.py
```diff
@@ -34,6 +34,7 @@
     "WhenNotMatchedClause",
     "QueryRecord",
     "QueryHistory",
+    "AsyncJob",
```
sfc-gh-jdu, line 37: You will need to add this class here to make doc working https://github.com/snowflakedb/snowpark-python/blob/main/docs/source/_templates/module.rst

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+        self._parameters = kwargs
+        self._result_meta = None
+        self._server_connection = server_connection
+        self._inserted = False
+        self._updated = False
+        self._deleted = False
+        self._plan = None
+
+    def is_done(self) -> bool:
+        """Check the status of query associated with this instance, return a bool value indicate whether the query is
```
sfc-gh-jdu, line 141:
```suggestion
        """Checks the status of they query associated with this instance and returns a bool value indicating whether the query has
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+        finished.
+
+        """
+        # return a bool value to indicate whether the query is finished
```
sfc-gh-jdu, line 145: this is not needed anymore

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+        status = self._conn.get_query_status(self.query_id)
+        is_running = self._conn.is_still_running(status)
+
+        return not is_running
+
+    def cancel(self) -> None:
+        """Cancel the query associated with this instance."""
```
sfc-gh-jdu, line 152:
```suggestion
        """Cancels the query associated with this instance."""
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+                rows_updated = int(result_data[0][idx])
+                idx += 1
+            if self._deleted:
+                rows_deleted = int(result_data[0][idx])
+            return snowflake.snowpark.MergeResult(
+                rows_inserted, rows_updated, rows_deleted
+            )
+
+    def result(self) -> Union[List[Row], "pandas.DataFrame", Iterator[Row], int, None]:
+        """Block and wait until the query associated with this instance is finished and return query results, this
```
sfc-gh-jdu, line 185:
```suggestion
        """Blocks and waits until the query associated with this instance finishes, then returns query results. 
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+        functions acts like execute query in a synchronous way. Return data type is decided by the function that create
```
sfc-gh-jdu, line 186:
```suggestion
        The data type of returned results is determined by how you create this :class:`AsyncJob` instance. For example, if this instance is returned by :meth:`DataFrame.collect_nowait`, you will get a list of :class:`Row`s from this method. 
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+        this instance
+
```
sfc-gh-jdu, line 188: remove empty line

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        # return result of the query
```
sfc-gh-jdu, line 190: not needed

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+        self._cursor.get_results_from_sfqid(self.query_id)
+        if self._data_type == _AsyncDataType.NONE_TYPE:
+            self._cursor.fetchall()
+            result = None
+        elif self._data_type == _AsyncDataType.PANDAS:
+            result = self._server_connection._to_data_or_iter(
+                self._cursor, to_pandas=True, to_iter=False
+            )["data"]
+            check_is_pandas_dataframe_in_to_pandas(result)
+        elif self._data_type == _AsyncDataType.PANDAS_BATCH:
```
sfc-gh-jdu, line 200: add a comment here:
```
TODO: SNOW-642562 support to_pandas_batches once the connector side bug gets fixed
```

tests/integ/scala/test_async_job_suite.py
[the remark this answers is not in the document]
sfc-gh-jdu: ok add a comment here to explain

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,227 @@
...
+    Example 5
+        Save a dataframe to table or copy it into a stage file can also be performed asynchronously::
+
+            >>> table_name = "name"
+            >>> async_job = df.write.save_as_table(table_name, block=False)
+            >>> # copy into a stage file
+            >>> remote_location = f"{session.get_session_stage()}/name.csv"
+            >>> async_job = df.write.copy_into_location(remote_location, block=False)
+            >>> async_job.result()
+            [Row(rows_unloaded=4, input_bytes=27, output_bytes=47)]
```
sfc-gh-jdu, line 90: Let's only check rows_unloaded here 
```
async_job.result()['rows_unloaded']
```
other data varies across different os platform.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,226 @@
...
+    MERGE = "merge"
+
+
+class AsyncJob:
+    """Provides a way to track an asynchronous query in Snowflake. A :class:`DataFrame` object can be evaluated asynchronously and an :class:`AsyncJob` object will be returned. With this instance, you can do:
+    - retrieve results;
+    - check the query status (still running or done);
+    - cancel the running query;
+    - retrieve the query ID and perform other operations on this query ID manually.
+    AsyncJob is strongly related to dataframe functions, to use AsyncJob, you need to create a dataframe first
```
sfc-gh-jdu, line 35:
```suggestion
    :class:`AsyncJob` is created by action methods in :class:`DataFrame`. Therefore, to use it, you need to create a dataframe first. Here we demonstrate how to do that: 
```
combine with the line below

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,226 @@
...
+            if self._deleted:
+                rows_deleted = int(result_data[0][idx])
+            return snowflake.snowpark.MergeResult(
+                rows_inserted, rows_updated, rows_deleted
+            )
+
+    def result(self) -> Union[List[Row], "pandas.DataFrame", Iterator[Row], int, None]:
+        """Blocks and waits until the query associated with this instance finishes, then returns query results, this
+        functions acts like execute query in a synchronous way. The data type of returned results is determined by how
+        you create this :class:`AsyncJob` instance.For example, if this instance is returned by
```
sfc-gh-jdu, line 185:
```suggestion
        you create this :class:`AsyncJob` instance. For example, if this instance is returned by
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,226 @@
...
+            if self._updated:
+                rows_updated = int(result_data[0][idx])
+                idx += 1
+            if self._deleted:
+                rows_deleted = int(result_data[0][idx])
+            return snowflake.snowpark.MergeResult(
+                rows_inserted, rows_updated, rows_deleted
+            )
+
+    def result(self) -> Union[List[Row], "pandas.DataFrame", Iterator[Row], int, None]:
```
sfc-gh-sfan, line 182: I'm wondering, can this method or part of the code be shared with some existing function? I imagine most of the process will be similar to how we process the result today, so we could avoid some code duplication.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,226 @@
...
+    Example 8
+        Cancel the running query associated with the dataframe::
+
+            >>> df = session.sql("select SYSTEM$WAIT(3)")
+            >>> async_job = df.collect_nowait()
+            >>> async_job.cancel()
+
+    Note:
+        - If a dataframe is associated with multiple queries,
+            + if you use `:meth:Session.create_dataframe` to create a dataframe from a large amount of local data and evaluate this dataframe asynchronously, data will still be loaded into Snowflake synchronously, and only fetching data from Snowflake again will be performed asynchronously.
```
sfc-gh-jdu, line 111:
```suggestion
            + if you use :meth:`Session.create_dataframe` to create a dataframe from a large amount of local data and evaluate this dataframe asynchronously, data will still be loaded into Snowflake synchronously, and only fetching data from Snowflake again will be performed asynchronously.
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,226 @@
...
+            + otherwise, multiple queries will be wrapped into a `Snowflake Anonymous Block <https://docs.snowflake.com/en/developer-guide/snowflake-scripting/blocks.html#using-an-anonymous-block>`_ and executed asynchronously as one query.
+        - Temporary objects (e.g., tables) might be created when evaluating dataframes and they will be dropped automatically after all queries finish when calling a synchronous API. When you evaluating dataframes asynchronously, temporary objects will only be dropped after calling :meth:`result`.
```
sfc-gh-jdu, line 113:
```suggestion
        - Temporary objects (e.g., tables) might be created when evaluating dataframes and they will be dropped automatically after all queries finish when calling a synchronous API. When you evaluate dataframes asynchronously, temporary objects will only be dropped after calling :meth:`result`.
```

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -330,15 +342,38 @@ def run_query(
         # because when the query plan has multiple queries, it will
         # have non-select statements, and it shouldn't fail if the user
         # calls to_pandas() to execute the query.
+        if not block:
```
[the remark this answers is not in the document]
sfc-gh-yuwang: Block means if this function block other functions. so block is true means this is a synchronous function while block is false means this is a asynchronous function

src/snowflake/snowpark/async_job.py
sfc-gh-yuwang: So actually this function is a bit different from what we have now. Most of other code is synchronously executed, so you dont have to use the query id to fetch results. Instead, you get a cursor instantly and processing the result is based on the cursor

sfc-gh-sfan: I see. I think my worry is mainly about if we need to change how we process the result, we might have to change two places. But it seems most of the actual processing is pushed down to connector, so we are probably okay.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,226 @@
...
+    PANDAS_BATCH = "pandas_batch"
+    COUNT = "count"
+    NONE_TYPE = "no_type"
+    UPDATE = "update"
+    DELETE = "delete"
+    MERGE = "merge"
+
+
+class AsyncJob:
+    """Provides a way to track an asynchronous query in Snowflake. A :class:`DataFrame` object can be evaluated asynchronously and an :class:`AsyncJob` object will be returned. With this instance, you can do:
```
sfc-gh-mabrennan, line 30: remove "do" at the end of this sentence.

sfc-gh-mabrennan (commented):
Nice job on the docstrings. I had two small suggestions. Thanks!

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,226 @@
...
+    - retrieve results;
+    - check the query status (still running or done);
+    - cancel the running query;
+    - retrieve the query ID and perform other operations on this query ID manually.
+    :class:`AsyncJob` is created by action methods in :class:`DataFrame`. Therefore, to use it, you need to create a
+    dataframe first. Here we demonstrate how to do that:
```
sfc-gh-yixie, line 36: Nit: I suggest explaining that all the `..._nowait` methods use async query, and explaining that it's equivalent to the counterpart methods that has no `nowait` but has `block=False`.

src/snowflake/snowpark/dataframe.py
```diff
@@ -439,26 +460,55 @@ def stat(self) -> DataFrameStatFunctions:
...
     def collect(
-        self, *, statement_params: Optional[Dict[str, str]] = None
+        self, *, statement_params: Optional[Dict[str, str]] = None, block: bool = True
     ) -> List["Row"]:
         """Executes the query representing this DataFrame and returns the result as a
         list of :class:`Row` objects.
 
         Args:
             statement_params: Dictionary of statement level parameters to be set while executing this action.
+            block: A bool value indicating whether blocking this function until the result is available. When it is ``False``,  this function executes the underlying queries of the dataframe asynchronously and returns a :class:`AsyncJob`.
```
sfc-gh-yixie, line 470: I prefer "whether this function will wait..."

src/snowflake/snowpark/async_job.py
sfc-gh-yixie: Plus, could we add an example that shows the user that issuing two async queries then getting the results would be faster than doing them one by one if the virtual warehouse has enough resources?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -307,19 +308,30 @@ def run_query(
...
+            if block:
+                results_cursor = self._cursor.execute(query, **kwargs)
+                self.notify_query_listeners(
+                    QueryRecord(results_cursor.sfqid, results_cursor.query)
+                )
+                logger.debug(f"Execute query [queryID: {results_cursor.sfqid}] {query}")
+            else:
+                results_cursor = self._cursor.execute_async(query, **kwargs)
+                self.notify_query_listeners(
+                    QueryRecord(results_cursor["queryId"], query)
```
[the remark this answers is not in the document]
sfc-gh-yuwang: execute_async() return a different result form execute(), which does not contain query

src/snowflake/snowpark/async_job.py
sfc-gh-jdu: It is already noted in `collect_nowait()`. So I don't think we need to mention it here.

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,247 @@
...
+    MERGE = "merge"
+
+
+class AsyncJob:
+    """Provides a way to track an asynchronous query in Snowflake. A :class:`DataFrame` object can be evaluated asynchronously and an :class:`AsyncJob` object will be returned. With this instance, you can:
+    - retrieve results;
+    - check the query status (still running or done);
+    - cancel the running query;
+    - retrieve the query ID and perform other operations on this query ID manually.
+    :class:`AsyncJob` is created by action methods in :class:`DataFrame`. All functions in :class:`DataFrame` with a subfix of '_nowait' execute
```
sfc-gh-jdu, line 35:
```suggestion
    :class:`AsyncJob` is created by action methods in :class:`DataFrame`. All methods in :class:`DataFrame` with a sudfix of ``_nowait`` execute
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,247 @@
...
+    asynchronously and create a :class:`AsyncJob` instance. They are also equivalent to functions in :class:`DataFrame` that
```
sfc-gh-jdu, line 36:
```suggestion
    asynchronously and create an :class:`AsyncJob` instance. They are also equivalent to corresponding methods in :class:`DataFrame` that
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,247 @@
...
+    set `block=False`. Therefore, to use it, you need to create a
```
sfc-gh-jdu, line 37:
```suggestion
    set ``block=False``. Therefore, to use it, you need to create a
```

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,247 @@
...
+    dataframe first. Here we demonstrate how to do that:
```
sfc-gh-jdu, line 38: move this line up along with the previous line

src/snowflake/snowpark/async_job.py
```diff
@@ -0,0 +1,247 @@
...
+            return snowflake.snowpark.MergeResult(
+                rows_inserted, rows_updated, rows_deleted
+            )
+
+    def result(self) -> Union[List[Row], "pandas.DataFrame", Iterator[Row], int, None]:
+        """Blocks and waits until the query associated with this instance finishes, then returns query results, this
+        functions acts like execute query in a synchronous way. The data type of returned results is determined by how
+        you create this :class:`AsyncJob` instance. For example, if this instance is returned by
+        :meth:`DataFrame.collect_nowait`, you will get a list of :class:`Row` s from this method.
+
```
sfc-gh-jdu, line 208: remove empty line

sfc-gh-jdu (approved):

sfc-gh-yixie (approved):

PR #422, opened 2022-08-05
SNOW-584202: Support upload stream in file API

sfc-gh-aalam:

   Fixes #SNOW-584202

   SNOW-584202 is created to add upload and download stream in file API. This PR covers the upload part by using the python connector to handle the upload

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -122,6 +122,7 @@ class ServerConnection:
                             func.__name__, duration, sfqid
                         )
                     logger.debug(f"Finished in {duration:.4f} secs")
+                    return result
 
                 return wrap
 
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -291,7 +291,7 @@ class ServerConnection:
         # https://docs.python.org/3/library/io.html#io.IOBase.close
         except ValueError as ex:
             if input_stream.closed:
-                raise SnowparkClientExceptionMessages.SERVER_UDF_UPLOAD_FILE_STREAM_CLOSED(
+                raise SnowparkClientExceptionMessages.SERVER_UPLOAD_FILE_STREAM_CLOSED(
                     dest_filename
                 )
             else:
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -257,6 +257,7 @@ class ServerConnection:
         compress_data: bool = True,
         source_compression: str = "AUTO_DETECT",
         overwrite: bool = False,
+        is_in_udf: bool = False,
     ) -> Optional[Dict[str, Any]]:
         uri = normalize_local_file(f"/tmp/placeholder/{dest_filename}")
         try:
@@ -291,9 +292,14 @@ class ServerConnection:
         # https://docs.python.org/3/library/io.html#io.IOBase.close
         except ValueError as ex:
             if input_stream.closed:
-                raise SnowparkClientExceptionMessages.SERVER_UPLOAD_FILE_STREAM_CLOSED(
-                    dest_filename
-                )
+                if is_in_udf:
+                    raise SnowparkClientExceptionMessages.SERVER_UDF_UPLOAD_FILE_STREAM_CLOSED(
+                        dest_filename
+                    )
+                else:
+                    raise SnowparkClientExceptionMessages.SERVER_UPLOAD_FILE_STREAM_CLOSED(
+                        dest_filename
+                    )
             else:
                 raise ex
 
--- a/src/snowflake/snowpark/file_operation.py
+++ b/src/snowflake/snowpark/file_operation.py
@@ -212,7 +212,7 @@ class FileOperation:
         self,
         input_stream: IO[bytes],
         stage_location: str,
-        dest_filename: str,
+        dest_file_name: str,
         *,
         parallel: int = 4,
         auto_compress: bool = True,
@@ -224,7 +224,7 @@ class FileOperation:
         Args:
             input_stream: The input stream from which the data will be uploaded.
             stage_location: The stage and prefix where you want to upload the files.
-            dest_filename: The filename used to store the file stream in the stage.
+            dest_file_name: The filename used to store the file stream in the stage.
             parallel: Specifies the number of threads to use for uploading files. The upload process separates batches of data files by size:
 
                   - Small files (< 64 MB compressed or uncompressed) are staged in parallel as individual files.
@@ -244,7 +244,7 @@ class FileOperation:
         put_result = self._session._conn.upload_stream(
             input_stream=input_stream,
             stage_location=stage_location,
-            dest_filename=dest_filename,
+            dest_filename=dest_file_name,
             parallel=parallel,
             compress_data=auto_compress,
             source_compression=source_compression,
--- a/src/snowflake/snowpark/file_operation.py
+++ b/src/snowflake/snowpark/file_operation.py
@@ -212,7 +212,6 @@ class FileOperation:
         self,
         input_stream: IO[bytes],
         stage_location: str,
-        dest_file_name: str,
         *,
         parallel: int = 4,
         auto_compress: bool = True,
@@ -223,8 +222,7 @@ class FileOperation:
 
         Args:
             input_stream: The input stream from which the data will be uploaded.
-            stage_location: The stage and prefix where you want to upload the files.
-            dest_file_name: The filename used to store the file stream in the stage.
+            stage_location: The full stage path with prefix and file name where you want the file to be uploaded.
             parallel: Specifies the number of threads to use for uploading files. The upload process separates batches of data files by size:
 
                   - Small files (< 64 MB compressed or uncompressed) are staged in parallel as individual files.
@@ -246,10 +244,23 @@ class FileOperation:
                 "Stream uploads for stored procedure are not supported yet"
             )
         else:
+
+            def parse_stage_file_location(stage_location: str):
+                if not stage_location:
+                    raise ValueError("stage_location cannot be empty")
+                elif stage_location[-1] == "/":
+                    raise ValueError("stage_location should end with target filename")
+                else:
+                    stage_location_parts = stage_location.split("/")
+                    filename = stage_location_parts[-1]
+                    stage_with_prefix = "/".join(stage_location_parts[:-1])
+                    return stage_with_prefix, filename
+
+            stage_with_prefix, dest_filename = parse_stage_file_location(stage_location)
             put_result = self._session._conn.upload_stream(
                 input_stream=input_stream,
-                stage_location=stage_location,
-                dest_filename=dest_file_name,
+                stage_location=stage_with_prefix,
+                dest_filename=dest_filename,
                 parallel=parallel,
                 compress_data=auto_compress,
                 source_compression=source_compression,
--- a/src/snowflake/snowpark/file_operation.py
+++ b/src/snowflake/snowpark/file_operation.py
@@ -219,7 +219,7 @@ class FileOperation:
         source_compression: str = "AUTO_DETECT",
         overwrite: bool = False,
     ) -> PutResult:
-        """Uploads local files to the stage via file stream.
+        """Uploads local files to the stage via a file stream.
 
         Args:
             input_stream: The input stream from which the data will be uploaded.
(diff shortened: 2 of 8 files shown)


src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -122,6 +122,7 @@ def wrap(*args, **kwargs):
                             func.__name__, duration, sfqid
                         )
                     logger.debug(f"Finished in {duration:.4f} secs")
+                    return result
```
sfc-gh-aalam, line 125: added this return so that caller of `upload_stream` gets back the result from put.

https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/server_connection.py#L248-L259

tests/integ/scala/test_file_operation_suite.py
```diff
@@ -219,6 +222,55 @@ def test_put_negative(session, temp_stage, temp_source_directory, path1):
...
+def test_put_stream_with_one_file(session, temp_stage, path1, path2, path3):
+    stage_prefix = f"prefix_{random_alphanumeric_name()}"
+    stage_with_prefix = f"@{temp_stage}/{stage_prefix}"
+    file_name = os.path.basename(path1)
+    with open(path1, "rb") as fd:
+        first_result = session.file.put_stream(fd, stage_with_prefix, file_name)
+    assert first_result.source == os.path.basename(path1)
+    assert first_result.target == os.path.basename(path1) + ".gz"
+    assert first_result.source_size in (10, 11)
+    assert first_result.target_size in (32, 64)
```
sfc-gh-aalam, line 234: I'm unsure of how to set correct values for asserting source and target size. They passes on my mac but could change later

src/snowflake/snowpark/file_operation.py
```diff
@@ -207,3 +207,47 @@ def get(
...
+                Defaults to 4.
+            auto_compress: Specifies whether Snowflake uses gzip to compress files during upload. Defaults to True.
+            source_compression: Specifies the method of compression used on already-compressed files that are being staged.
+                Values can be 'AUTO_DETECT', 'GZIP', 'BZ2', 'BROTLI', 'ZSTD', 'DEFLATE', 'RAW_DEFLATE', 'NONE'. Defaults to "AUTO_DETECT".
+            overwrite: Specifies whether Snowflake will overwrite an existing file with the same name during upload. Defaults to False.
+
+        Returns:
+            An object of :class:`PutResult` which represents the results of an uploaded file.
+        """
+        put_result = self._session._conn.upload_stream(
```
sfc-gh-sfan, line 244: This is probably a bit out of the scope for this PR, but `upload_stream` does not really handle options like `compress_data`, `overwrite` well in stored proc. We might need to add stream support for `_upload` and use that https://github.com/snowflakedb/Stored-Proc-Python-Connector/blob/main/src/snowflake/connector/cursor.py#L543

src/snowflake/snowpark/file_operation.py
```diff
@@ -207,3 +207,47 @@ def get(
...
+        parallel: int = 4,
+        auto_compress: bool = True,
+        source_compression: str = "AUTO_DETECT",
+        overwrite: bool = False,
+    ) -> PutResult:
+        """Uploads local files to the stage via file stream.
+
+        Args:
+            input_stream: The input stream from which the data will be uploaded.
+            stage_locatio: The stage and prefix where you want to upload the files.
```
sfc-gh-jdu, line 226:
```suggestion
            stage_location: The stage and prefix where you want to upload the files.
```

src/snowflake/snowpark/file_operation.py
```diff
@@ -207,3 +207,47 @@ def get(
...
+            dest_filenam: The filename used to store the file stream in the stage.
```
sfc-gh-jdu, line 227:
```suggestion
            dest_filename: The filename used to store the file stream in the stage.
```

src/snowflake/snowpark/file_operation.py
sfc-gh-jdu: can't we use this https://github.com/snowflakedb/Stored-Proc-Python-Connector/blob/1b3d7f8d5e54a6aaa8852eb08d4b6150d777e266/src/snowflake/connector/cursor.py#L581?

tests/integ/scala/test_file_operation_suite.py
sfc-gh-jdu: hmm it would be flaky. Probably just check it's not None.

tests/integ/scala/test_file_operation_suite.py
```diff
@@ -219,6 +222,55 @@ def test_put_negative(session, temp_stage, temp_source_directory, path1):
...
+
+
+def test_put_stream_negative(session, temp_stage, temp_source_directory, path1):
+    stage_prefix = f"prefix_{random_alphanumeric_name()}"
+    stage_with_prefix = f"@{temp_stage}/{stage_prefix}"
+    file_name = os.path.basename(path1)
+    fd = open(path1, "rb")
+    fd.close()
+
+    with pytest.raises(SnowparkUploadUdfFileException) as ex_info:
```
sfc-gh-jdu, line 269: we should have a new exception because now this API is not used for udf only.

src/snowflake/snowpark/file_operation.py
sfc-gh-sfan: We might be able to. It's actually interesting that the java side does not provide an `overwrite` check at all. The only option is `compress`.

https://github.com/snowflakedb/thundersnow/blob/master/src/main/java/com/snowflake/snowpark_java/FileOperation.java#L158-L160

It depends on how we want our API to look like. If it is the same as Java then this API should just work (even though it does not handle compression today). If we want to have a uniform API just like `file.put`, then it might be easier to go through `file_transfer_agent` where we have overwrite check etc.

src/snowflake/snowpark/file_operation.py
```diff
@@ -207,3 +207,47 @@ def get(
...
+        input_stream: IO[bytes],
+        stage_location: str,
+        dest_filename: str,
+        *,
+        parallel: int = 4,
+        auto_compress: bool = True,
+        source_compression: str = "AUTO_DETECT",
+        overwrite: bool = False,
+    ) -> PutResult:
+        """Uploads local files to the stage via file stream.
```
sfc-gh-mabrennan, line 222: Maybe add "a" before "file stream".

sfc-gh-mabrennan (commented):
Nice job on the docstrings. I just had one optional suggestions. Thanks!

src/snowflake/snowpark/file_operation.py
sfc-gh-sfan: As discussed offline, let's fail if it is stored proc for now, until we fix the `upload_stream` in sp connector

tests/integ/scala/test_file_operation_suite.py
```diff
@@ -219,6 +222,55 @@ def test_put_negative(session, temp_stage, temp_source_directory, path1):
     assert "does not exist or not authorized." in str(stage_not_exist_info)
 
 
+def test_put_stream_with_one_file(session, temp_stage, path1, path2, path3):
```
sfc-gh-sfan, line 225: Let's skip these tests for stored proc. One example could be found in https://github.com/snowflakedb/snowpark-python/blob/c838f17c3fb373f7f3cf2192c6a5bf2d74a0dc19/tests/integ/test_session.py#L36-L39

tests/integ/scala/test_file_operation_suite.py
sfc-gh-aalam: I talked to Jianzhun about this offline and it seems like a new exception for this could mean another behavior change. Should I proceed with BCR?

src/snowflake/snowpark/file_operation.py
sfc-gh-jdu: I think you can just add a check here: if the user is in SP and the user calls upload_stream, raise an NotImplementedError immediately?

tests/integ/scala/test_file_operation_suite.py
sfc-gh-jdu: To avoid BCR, probably we can check whether it’s used in UDF, if it is, then throw the original error; if not, throw a new error. @sfc-gh-yixie wdyt?

src/snowflake/snowpark/file_operation.py
sfc-gh-sfan: Yeah I agree with @sfc-gh-jdu 's suggestion.

tests/integ/scala/test_file_operation_suite.py
sfc-gh-yixie: Yeah different platforms may have different size. It almost killed me.

tests/integ/scala/test_file_operation_suite.py
sfc-gh-yixie: We can create a new upload exception class for `put_stream`. To avoid break change, UDF still throws a SnowparkUploadUdfFileException, which can be changed to be a subclass of the new upload exception.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -290,7 +291,7 @@ def upload_stream(
         # https://docs.python.org/3/library/io.html#io.IOBase.close
         except ValueError as ex:
             if input_stream.closed:
-                raise SnowparkClientExceptionMessages.SERVER_UDF_UPLOAD_FILE_STREAM_CLOSED(
+                raise SnowparkClientExceptionMessages.SERVER_UPLOAD_FILE_STREAM_CLOSED(
```
sfc-gh-aalam, line 294: Is there a way to check weather this function was called by udf or not?

tests/unit/scala/test_error_message.py
```diff
@@ -380,6 +380,14 @@ def test_server_udf_upload_file_stream_closed():
     assert "A file stream was closed when uploading UDF files" in ex.message
 
 
+def test_server_upload_file_stream_closed():
+    dest_filename = "file"
+    ex = SnowparkClientExceptionMessages.SERVER_UPLOAD_FILE_STREAM_CLOSED(dest_filename)
+    assert isinstance(ex, SnowparkUploadUdfFileException)
```
sfc-gh-aalam, line 386: oops, I just noticed this error. Line 386 need to check for instance `SnowparkUploadUdfFileException`

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-jdu: we can add an`is_in_udf` argument to the internal upload_stream method, and set it accordingly based on where it's called.

sfc-gh-sfan (approved):
LGTM

src/snowflake/snowpark/file_operation.py
```diff
@@ -207,3 +207,52 @@ def get(
         # connector raises IndexError when no file is downloaded from python connector.
         except IndexError:
             return []
+
+    def put_stream(
```
sfc-gh-jdu, line 211: I just took a look at the counterpart in scala and actually there are two subtle differences: https://github.com/snowflakedb/thundersnow/blob/4731ff818927154f66d90fec0f2eaf91b1a17090/src/main/scala/com/snowflake/snowpark/FileOperation.scala#L146-L160
1. Should it be put_stream or upload_stream? (I slightly prefer put_stream)
2. Should we support a full stage path with `stage_location`, or support the separate `dest_file_name`? The user can specify `stage_location='@/xxx/yyy.csv'` and doesn't need `dest_file_name`. I feel it looks neater (also aligns with scala). 

@sfc-gh-sfan @sfc-gh-yixie thoughts?

sfc-gh-sfan: 1. I also prefer `put_stream`. 
2. Yeah combining into one parameter seems more concise.

src/snowflake/snowpark/file_operation.py
```diff
@@ -207,3 +207,63 @@ def get(
...
+            An object of :class:`PutResult` which represents the results of an uploaded file.
+        """
+        if is_in_stored_procedure():
+            raise NotImplementedError(
+                "Stream uploads for stored procedure are not supported yet"
+            )
+        else:
+
+            def parse_stage_file_location(stage_location: str):
+                if not stage_location:
```
sfc-gh-jdu, line 249: it will be safer if you do `stage_location = stage_location.strip()` first

src/snowflake/snowpark/file_operation.py
```diff
@@ -207,3 +207,63 @@ def get(
...
+                    raise ValueError("stage_location cannot be empty")
+                elif stage_location[-1] == "/":
+                    raise ValueError("stage_location should end with target filename")
+                else:
+                    stage_location_parts = stage_location.split("/")
```
sfc-gh-jdu, line 254: looks like https://docs.python.org/3/library/stdtypes.html#str.rsplit with maxsplit=1 is better here

src/snowflake/snowpark/file_operation.py
sfc-gh-yixie: I agree with `put_stream` and one parameter.

sfc-gh-jdu (approved):
👍

PR #428, opened 2022-08-10
SNOW-641391 Add more join types in doc

sfc-gh-sfan:
Description

We apparantly support anti/cross/semi joins, so we might as
well add that to the doc

Testing

N/A

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -1600,7 +1600,14 @@ class DataFrame:
             right: The other :class:`Dataframe` to join.
             using_columns: A list of names of the columns, or the column objects, to
                 use for the join.
-            join_type: The type of join ("inner", "full", "left", "right", "semi", "anti", "cross").
+            join_type: We support the following join types:
+              - Inner join: "inner"
+              - Left outer join: "left", "leftouter"
+              - Right outer join: "right", "rightouter"
+              - Full outer join: "full", "outer", "fullouter"
+              - Left semi join: "semi", "leftsemi"
+              - Left anti join: "anti", "leftanti"
+              - Cross join: "cross"
 
         Note:
             When performing chained operations, this method will not work if there are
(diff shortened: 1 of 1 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -1600,7 +1600,7 @@ def join(
             right: The other :class:`Dataframe` to join.
             using_columns: A list of names of the columns, or the column objects, to
                 use for the join.
-            join_type: The type of join ("inner", "full", "left", "right").
+            join_type: The type of join ("inner", "full", "left", "right", "semi", "anti", "cross").
```
sfc-gh-jdu, line 1603: all allowed join type strs are here https://github.com/snowflakedb/snowpark-python/blob/2fb6672b9c5851f6098a01e55e64b1d7ffc2fa84/src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py#L10-L24

sfc-gh-sfan: Yeah I thought about it but I'm not sure if we want to add duplicate like `full` and `fullouter`, so I just followed the existing pattern where each category only contains the shorter one. But if this is preferred I could do a full copy for sure.

sfc-gh-mabrennan (approved):

src/snowflake/snowpark/dataframe.py
sfc-gh-jdu: probably we can mentioned them all (like fullouter and rightouter are both supported by snowflake https://docs.snowflake.com/en/sql-reference/constructs/join.html#syntax), and I believe some like leftsemi and semi are added to be compatible with pyspark. But it's also worthwhile mentioning that some strs are just aliases of another one, e.g., `leftanti` is an alias of `anti` (you can find all of them here https://github.com/snowflakedb/snowpark-python/blob/2fb6672b9c5851f6098a01e55e64b1d7ffc2fa84/src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py#L30-L48)

sfc-gh-sfan: Yeah let me make it more organized, such as:

```
We support the following join types:

Left outer join: "left" or "leftouter"
...
```

What do you think?

sfc-gh-jdu: yes, looks perfect

PR #429, opened 2022-08-11
SNOW-641398 Skip create table if exists in append mode

sfc-gh-sfan:
Description

Create table requires a different privilege from insert, and
appending to an existing table should be allowed for user without
create privilege. To achieve that, we skip creating table if
it already exists.

Testing

integ test

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -63,7 +63,6 @@ from snowflake.snowpark._internal.utils import (
     generate_random_alphanumeric,
     random_name_for_temp_object,
 )
-from snowflake.snowpark.exceptions import SnowparkSQLException
 from snowflake.snowpark.row import Row
 from snowflake.snowpark.types import StructType
 
@@ -491,16 +490,6 @@ class SnowflakePlanBuilder:
             source_plan,
         )
 
-    def _table_exists(self, table_name):
-        try:
-            # DESC is used here because SHOW does not work for qualified table name
-            self.session.sql(f"DESC TABLE {table_name}").collect()
-            return True
-        except SnowparkSQLException:
-            # We could verify if this is a compile error and throw otherwise, but directly return False gives no
-            # behavior change as the old behavior assumes table does not exist
-            return False
-
     def save_as_table(
         self,
         table_name: str,
@@ -510,7 +499,7 @@ class SnowflakePlanBuilder:
         child: SnowflakePlan,
     ) -> SnowflakePlan:
         if mode == SaveMode.APPEND:
-            if self._table_exists(table_name):
+            if self.session._table_exists(table_name):
                 return self.build(
                     lambda x: insert_into_statement(
                         table_name=table_name,
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -63,6 +63,7 @@ from snowflake.snowpark._internal.utils import (
     generate_random_alphanumeric,
     random_name_for_temp_object,
 )
+from snowflake.snowpark.exceptions import SnowparkSQLException
 from snowflake.snowpark.row import Row
 from snowflake.snowpark.types import StructType
 
@@ -490,6 +491,16 @@ class SnowflakePlanBuilder:
             source_plan,
         )
 
+    def _table_exists(self, table_name):
+        try:
+            # DESC is used here because SHOW does not work for qualified table name
+            self.session.sql(f"DESC TABLE {table_name}").collect()
+            return True
+        except SnowparkSQLException:
+            # We could verify if this is a compile error and throw otherwise, but directly return False gives no
+            # behavior change as the old behavior assumes table does not exist
+            return False
+
     def save_as_table(
         self,
         table_name: str,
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -490,6 +491,16 @@ def join(
...
 
+    def _table_exists(self, table_name):
+        try:
+            # DESC is used here because SHOW does not work for qualified table name
+            self.session.sql(f"DESC TABLE {table_name}").collect()
+            return True
+        except SnowparkSQLException:
+            # We could verify if this is a compile error and throw otherwise, but directly return False gives no
+            # behavior change as the old behavior assumes table does not exist
+            return False
```
sfc-gh-sfan, line 502: To be honest I'm not sure if using `DESC` is a good idea or not as it requires a try-catch. Using `SHOW` is not straightforward as we need to extract database and schema from the table_name. It could easily get messy when there are dots in the double quoted identifier

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -490,6 +491,16 @@ def join(
             source_plan,
         )
 
+    def _table_exists(self, table_name):
```
sfc-gh-jdu, line 494: actually you can use https://github.com/snowflakedb/snowpark-python/blob/2fb6672b9c5851f6098a01e55e64b1d7ffc2fa84/src/snowflake/snowpark/session.py#L1565
(and please add a return type hint for it)

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -499,32 +510,43 @@ def save_as_table(
...
-                    *child.queries[0:-1],
-                    Query(create_table),
-                    Query(
-                        insert_into_statement(
-                            table_name=table_name,
-                            child=child.queries[-1].sql,
-                            column_names=column_names,
-                        )
+            if self._table_exists(table_name):
+                return self.build(
```
sfc-gh-jdu, line 514: in this case, why is `child.queries` are not considered anymore?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-sfan: Hmmm I would think the way this function works is wrong because it does not support qualified table name (e.g. `test_db.test_schema.test_table`). The database and schema need to be explicitly specified as shown in https://docs.snowflake.com/en/sql-reference/sql/show-tables.html. This might be a bug? I could change this to using `DESC` as well.

A simple repro:

```
session = Session.builder.configs(CONNECTION_PARAMETERS).create()
session.sql("create or replace table sfan_schema.test_table (id int)").collect()
print(session._table_exists('sfan_schema.test_table'))

False
```

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-sfan: Nice catch. It should also have `child, None` as other modes.

sfc-gh-sfan: Nice catch. It should also have `child, None` as other modes.

sfc-gh-sfan: Wait it already has `child, None`, just not very obvious in the diff. I assume `child` should have `child.queries` handled?

sfc-gh-jdu: oops bad github!

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: yea you're right. But I'm also not sure whether using desc is appropriate because we might send a failed query for the user... A hacky way to avoid this is to use self.session._conn._cursor.describe('DESC TABLE xxx'), which will not show up in user's query history.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-sfan: Yeah totally...I also did not see that in github, and stared at my pycharm questioning the meaning of life :/

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-sfan: That is actually a good idea :)

src/snowflake/snowpark/session.py
```diff
@@ -1623,8 +1623,14 @@ def query_history(self) -> QueryHistory:
         return query_listener
 
     def _table_exists(self, table_name: str):
-        tables = self._run_query(f"show tables like '{table_name}'")
-        return tables is not None and len(tables) > 0
+        try:
+            # DESC is used here because SHOW does not work for qualified table name
```
sfc-gh-jdu, line 1627: add one more comment saying we use `cursor.describe` so the failed query won't show up in query history?

src/snowflake/snowpark/session.py
```diff
@@ -1623,8 +1623,14 @@ def query_history(self) -> QueryHistory:
         return query_listener
 
     def _table_exists(self, table_name: str):
```
sfc-gh-jdu, line 1625:
```suggestion
    def _table_exists(self, table_name: str) -> bool:
```

src/snowflake/snowpark/session.py
```diff
@@ -1623,8 +1623,14 @@ def query_history(self) -> QueryHistory:
...
-        tables = self._run_query(f"show tables like '{table_name}'")
-        return tables is not None and len(tables) > 0
+        try:
+            # DESC is used here because SHOW does not work for qualified table name
+            self._conn._cursor.describe(f"DESC TABLE {table_name}")
+            return True
+        except ProgrammingError:
+            # We could verify if this is a compile error and throw otherwise, but directly return False gives no
```
sfc-gh-jdu, line 1631: Ideally we can check whether `does not exist or not authorized` is in error message then return false but it might be fragile and unnecessary.

tests/integ/test_dataframe.py
```diff
@@ -1572,6 +1572,21 @@ def test_write_invalid_table_type(session):
         df.write.save_as_table(table_name, table_type="invalid")
 
 
+def test_append_existing_table(session):
+    table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
+    session.sql(f"create table {table_name} (a int, b int)").collect()
```
sfc-gh-jdu, line 1577:
```suggestion
    Utils.create_table(session, table_name, "a int, b int", is_temporary=True)
```

sfc-gh-jdu (approved):

src/snowflake/snowpark/session.py
sfc-gh-sfan: Damn I forgot....Thanks for catching it (again).

PR #432, opened 2022-08-12
SNOW-631099: display unpickled python udf body registered by snowpark python

sfc-gh-aling:

   Fixes SNOW-631099

The PR is to generate source code for the python udf and append it as **comment** to the original code.
there is no api change required as this is an enhancement to the current approach.
The execution still uses function pickled by cloudpickle for source of truth.

Inspired by cloudpickle, we do source code re-generation by analyzing target python UDF and its references objects.


src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+    have to be inferred manually. Consider the following example:
+
+    import a1.a2.a3.a4
+    def func():
+        a1.a2.a3.a4.foo()
+
+    func.__code__.co_names only contains ("a1", "a2", "a3", "a4", "foo") which does not include the
+    complete import path information.
+
+    To reconstruct "a1.a2.a3.a4", the current strategy is to import each prefix import of the import chains.
```
sfc-gh-jdu, line 379: so it will do 
```
import a1
import a1.a2
import a1.a2.a3
import a1.a2.a3.a4
```
?
Is `co_names` guaranteed ordered?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+            else f"{func_module_name}.{func.__name__}"
+        )
+
+    # 6. handle function assignment
+    complete_source_code = f"""\
+{complete_source_code}
+{f"func = {func_assignment}"}\
+""".strip()
+
+    # 7. if code as comment is true, prefix each line with '#'
```
sfc-gh-jdu, line 297: I just feel use `"""` is better?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+                f"#{f' {line}' if line else ''}"
+                for line in complete_source_code.splitlines()
+            ]
+        )
+    )
+
+    return complete_source_code.strip()
+
+
+def is_lambda(func: FunctionType):
```
sfc-gh-jdu, line 312:
```suggestion
def is_lambda(func: FunctionType) -> bool:
```

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+    return func.__name__ == "<lambda>"
+
+
+def get_lambda_code_text(code_text: str):
```
sfc-gh-jdu, line 316:
```suggestion
def get_lambda_code_text(code_text: str) -> str:
```

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+    # such that ast could parse the expression tree:
+    #     session.udf.register(
+    #         lambda x, y: x + y, ...
+    #     )
+    try:
+        source_ast = ast.parse(code_text)
+    except SyntaxError as exc:
+        if "cannot assign to lambda" in str(exc):
+            # handle case like:
+            # session.udf.register(
```
sfc-gh-jdu, line 327: is 
```
 session.udf.register(
            lambda x, y: x + y, ...
            # some comments
)
```
also handled here?

src/snowflake/snowpark/functions.py
```diff
@@ -2913,6 +2913,7 @@ def udf(
     parallel: int = 4,
     max_batch_size: Optional[int] = None,
     statement_params: Optional[Dict[str, str]] = None,
+    source_code_generation: bool = False,
```
sfc-gh-jdu, line 2916: true?

tests/integ/test_udf.py
```diff
@@ -1584,3 +1584,24 @@ def test_udf_pickle_failure(session):
         "local environment first, add it to the UDF with session.add_import(), and read it from "
         "the UDF." in str(ex_info)
     )
+
+
+def test_comment_in_udf_description(session):
```
sfc-gh-jdu, line 1589: Is it possible it run all current udf tests with source_code mode? Though it will be time-consuming...

src/snowflake/snowpark/functions.py
sfc-gh-aling: nice catch, updated to True

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,393 @@
...
+) -> str:
+    if code_as_comment:
+        return udf_source_code
+    # remove the udf/pandas_udf annotation to avoid re-registration
+    res = re.search(r"@(udf|pandas_udf)\(", udf_source_code)
+    if res is None:
+        return udf_source_code
+    udf_anno_begin = res.start()
+    udf_anno_end = res.end()
+    cnt = 1
```
sfc-gh-sfan, line 129: nit: `parenthesis_count`?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,393 @@
...
+    return out_names
+
+
+def remove_function_udf_annotation(
+    udf_source_code: str, code_as_comment: bool = True
+) -> str:
+    if code_as_comment:
+        return udf_source_code
+    # remove the udf/pandas_udf annotation to avoid re-registration
+    res = re.search(r"@(udf|pandas_udf)\(", udf_source_code)
```
sfc-gh-sfan, line 124: Should we assert that there is only one occurrence of such decorator? I understand that having two does not work anyway because you cannot execute `@udf` inside a UDF, but just curious if we want to throw this earlier.

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,393 @@
...
+    param func: The target function that has to been generated
+    param code_as_comment: All lines of the generated source code will be prefixed with '#' in which case code
+     are turned in comments. This is the default behavior.
+    """
+
+    try:
+        check_func_type(func)
+    except TypeError:
+        if code_as_comment:
+            # if it is an unsupported type, then no code generation and return emtpy string
```
sfc-gh-sfan, line 175: typo: `emtpy` -> `empty`

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,393 @@
...
+    if (
+        isinstance(func, classmethod)
+        or inspect.ismethod(func)
+        or not (isinstance(func, (FunctionType, BuiltinFunctionType)))
+    ):
+        logger.debug(f"Code generation for {type(func)} is not supported yet.")
+        raise TypeError(f"Code generation for {type(func)} is not supported yet.")
+
+
+def generate_source_code(
```
sfc-gh-sfan, line 152: For each step in this function, can we extract them to a function so it is easier to reason?

I think the major confusion for me is that I don't understand if each step has dependencies on previous steps, or they just happen to be in this sequence.

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,393 @@
...
+                (obj.__name__, name)
+            )  # name could be an alias
+        else:
+            # function/class/variables defined in the same module
+            if inspect.isfunction(obj):
+                func_text += textwrap.dedent(inspect.getsource(obj))
+            elif inspect.isclass(obj):
+                # dynamic class parsing will be handled separately
+                pass
+            else:
```
sfc-gh-sfan, line 236: Just sanity check: Is `else` all about global variables?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+    "The following comment contains the UDF source code"
+    " generated by snowpark-python for explanatory purposes.\n"
+)
+CODE_HEADER = """\
+from __future__ import annotations
+import pickle
+"""
+
+
+def get_func_references(func: FunctionType, ref_objects: dict):
```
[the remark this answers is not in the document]
sfc-gh-aling: it's a bit difficult.. because it could basically be `Any` as a global variable could be of Any object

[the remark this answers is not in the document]
sfc-gh-aling: but I have added bunch of type hint to the variable used, hope that would help

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+    classes_to_generate.insert(0, cls)
+    # get base classes first
+    inferred_clses = [v for v in reversed(cls.__bases__) if v.__qualname__ != "object"]
+
+    # __dict__ contains function, classmethod, classes attributes within a given class which
+    # needs to be further analyzed
+    clsdict = dict(cls.__dict__)
+    for v in clsdict.values():
+        if isinstance(v, type) and v.__module__ == func_module_name:
+            # v is a class defined in the same module as UDF func's, need to dynamically parse the class
```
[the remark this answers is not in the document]
sfc-gh-aling: nice catch, this is a bug in my implementation, if not defined in the same module we should import it from the other module.

I have updated the code

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+):
+    """
+    To get the referenced objects of a dynamically defined class.
+    A class could have methods, classes referencing objects.
+
+    """
+    # order matters to classes_to_generate, when constructing source code, referenced classes need to be defined first.
+    classes_to_generate.insert(0, cls)
+    # get base classes first
+    inferred_clses = [v for v in reversed(cls.__bases__) if v.__qualname__ != "object"]
```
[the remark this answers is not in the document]
sfc-gh-aling: the code is originally inspired by cloudpickle, the `reversed` here is due to Python's interpreter sequences of execution of the inherited methods when there're multiple inheritance and cloudpickle needs the sequence for pickling different classes.

I don't think it's necessary for us here, I can't think about a case that would impact the code generation..
let's remove this `reversed` first

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+                out_names.update(extract_func_global_refs(const))
+
+    return out_names
+
+
+def remove_function_udf_annotation(udf_source_code, code_as_comment=True):
+    if code_as_comment:
+        return udf_source_code
+    # remove the udf/pandas_udf annotation to avoid re-registration
+    res = re.search(r"@(udf|pandas_udf)\(", udf_source_code)
```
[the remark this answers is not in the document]
sfc-gh-aling: updated

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: fixed

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: yes, it's like a pipeline.

it's a piece of good suggestion to split them, I have updated the code to make them separate function calls.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: updated

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,393 @@
...
+    # 1. resolve function global references
+    code_object = func.__code__
+    globals_ref = extract_func_global_refs(
+        func.__code__
+    )  # get the names of the objects which func references
+    globals = {
+        k: func.__globals__[k]
+        for k in globals_ref
+        if k in func.__globals__  # retrieve the objects by names
+        and k not in ref_objects
```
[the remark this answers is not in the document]
sfc-gh-aling: local variables are not included, `ref_objects` are modules, classes, methods, nonlocal/global variables referenced by the function.

tests/integ/test_udf.py
sfc-gh-aling: yes, since we're now launching the beta release plan for experimental features, I think that would a better place for real execution.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: yes, I have updated the unit test for it

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: I initially want to use `"""`, however, the source code itself code could contain `"""` comments which will mess up the generated code.

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,391 @@
...
+            inspect.isclass(obj) or inspect.isfunction(obj)
+        ) and obj.__module__ != func_module_name:
+            # b) classes or functions imported from other modules
+            to_import_from_module[obj.__module__].add(
+                (obj.__name__, name)
+            )  # name could be an alias
+        else:
+            # function/class/variables defined in the same module
+            if inspect.isfunction(obj):
+                func_text += textwrap.dedent(inspect.getsource(obj))
```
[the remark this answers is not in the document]
sfc-gh-aling: updated

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: yes!

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,393 @@
...
+    """
+    # order matters to classes_to_generate, when constructing source code, referenced classes need to be defined first.
+    classes_to_generate.insert(0, cls)
+    # get base classes first
+    inferred_clses = [v for v in reversed(cls.__bases__) if v.__qualname__ != "object"]
+
+    # __dict__ contains function, classmethod, classes attributes within a given class which
+    # needs to be further analyzed
+    clsdict = dict(cls.__dict__)
+    for v in clsdict.values():
```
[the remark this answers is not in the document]
sfc-gh-aling: right, others are ignored as they won't be counted for code generation.

I have updated comments for the else statement, we may want to revisit the class variables but that would be a separate stories.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: I honestly feel it's not the duty of the code generation -- it's more like udf source code validation.

and if 2 `@udf` is designed not to work, we probably should throw at an even earlier stage priori to source code generation.

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,393 @@
...
+            k: v
+            for k, v in zip(
+                code_object.co_freevars,
+                list(map(lambda x: x.cell_contents, func.__closure__)),
+            )
+        }
+        for k, v in closures.items():
+            ref_objects[k] = v
+            # if the closure item is a function and is not itself (recursive) and has not been visited
+            if isinstance(v, FunctionType) and k not in ref_objects and v != func:
```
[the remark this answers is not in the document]
sfc-gh-aling: nice catch, removed!

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: well, yes and no (which is then probably no).

I have configured the following two weird modules for testing:

- a4.a3.a2.a1.SomeClass
- a1.a2.a3.a4.SomeClass

and it turns out `conames` would just come with `(a4, a3, a2, a1, SomeClass)`, so from the behavior, it's like a first-in set.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: had some more thoughts on this after testing the current behavior.

`@udf` twice today would give us error:
`TypeError: <snowflake.snowpark.udf.UserDefinedFunction object at 0x121cff820> is not a module, class, method, or function.`

if code is generated as comments, we don't check and remove annotations, leaving the decorators as they're in the comments.

however I do feel checking multiple `@udf`s exist provide value for code being generated for execution (in the future).

I have updated the implementation to check and raise.

how do you feel now?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+    """
+    imports = set()
+    for module in top_level_modules:
+        module_prefix = f"{module.__name__}."
+        for name in [m for m in sys.modules if m.startswith(module_prefix)]:
+            tokens = set(name[len(module_prefix) :].split("."))
+            if not tokens - set(
+                func.__code__.co_names
+            ):  # only add imports that co_names contains
+                imports.add((name, name))
```
sfc-gh-jdu, line 404: the docstring says the second item of the tuple is the alias, but from here actually they are the same?

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-jdu: probably it's fine since it's unusual. But can we handle something like this:
```
import pandas.io
import io
```
?
Will it be generated like 
```
import pandas
import pandas.io
```
?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+            or submodules.
+
+    Returns:
+        A set of tuple with each tuple composed of two string, the first one is actual name for the imported object,
+        and the second one is alias.
+
+    """
+    imports = set()
+    for module in top_level_modules:
+        module_prefix = f"{module.__name__}."
```
sfc-gh-jdu, line 398: so top level modules are not added here? If yes, should we add a comment?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+    inferred_classes: List[Tuple[Any, bool]] = []
+
+    for base_class in cls.__bases__:
+        base_class_name = base_class.__qualname__
+        top_level_class = base_class_name.split(".")[0]
+        if base_class_name == "object":
+            continue
+        elif base_class.__module__ == func_module_name:
+            # if base class is from the same module, we need to parse the class as well as generate code for the class
+            # False in the tuple means this is not a nested class
```
sfc-gh-jdu, line 112: move this comment (explanation for the bool value to the line before inferred_classes definition)?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+
+    # find the pairing ')' of the leading 'udf('
+    while parenthesis_count != 0:
+        if udf_source_code[udf_anno_end] == "(":
+            parenthesis_count += 1
+        elif udf_source_code[udf_anno_end] == ")":
+            parenthesis_count -= 1
+        udf_anno_end += 1
+
+    # check if there are still @udf annotations, then it's a
```
sfc-gh-jdu, line 200: typo?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+            ref_objects[top_level_class] = func.__globals__[top_level_class]
+
+    # __dict__ contains function, classmethod, classes attributes within a given class which
+    # needs to be further analyzed
+    for v in dict(cls.__dict__).values():
+        if inspect.isclass(v):
+            top_level_cls_name = v.__qualname__.split(".")[0]
+            if v.__module__ == func_module_name:
+                # v is a class defined in the same module as UDF func's, need to dynamically parse the class
+                # if v is class defined in cls, then we should not re-generate code for the nested class
```
sfc-gh-sfan, line 125: Maybe we should move this comment to `for inferred_class, is_nested_class in inferred_classes:`? Here we only process whether a class is nested or not.

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+                # v is a class defined in another module, import the top level class from another module
+                ref_objects[top_level_cls_name] = v
+        elif inspect.isfunction(v) or isinstance(v, classmethod):
+            # v is a function/classmethod, get the references objects of the Function object
+            get_func_references(
+                v if not isinstance(v, classmethod) else v.__func__, ref_objects
+            )
+        else:
+            # cls.__dict__ would also return __module__, __doc__, __weakref__ which are not required
+            # for code generation, however, class variables is also included in __dict__, we don't do value evaluation
```
sfc-gh-sfan, line 139: Will it cause the generated code not working?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+
+
+def extract_func_global_refs(code: CodeType) -> Dict[str, None]:
+    # inspired by cloudpickle to recursively extract all the global references used by the target func's code object
+    co_names = code.co_names
+    out_names = {}
+    for instr in dis.get_instructions(code):
+        op = instr.opcode
+        if op in GLOBAL_OPS:
+            out_names[co_names[instr.arg]] = None
```
sfc-gh-sfan, line 163: Is `instr.arg` guaranteed to be part of `co_names`? Also, why do we set it to `None`?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+def remove_function_udf_annotation(udf_source_code: str) -> str:
+    """
+    Remove the udf/pandas_udf annotation to avoid re-registration.
+    """
+    res = re.search(r"@(pandas_)?udf", udf_source_code)
+    if res is None:
+        return udf_source_code
+    udf_anno_begin = res.start()
+    udf_anno_end = res.end()
+    if udf_source_code[udf_anno_end] == "\n":
```
sfc-gh-sfan, line 182: I'm just curious, is comment/spaces stripped in `udf_source_code`, so that we are sure it could only be followed by `\n` or `(`?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+            # handle case like:
+            # session.udf.register(
+            #    lambda x, y: x + y, ...
+            # )
+            code_text = f"wrapper({code_text})"
+        elif "unmatched ')'" in str(exc):
+            # handle case like:
+            # session.udf.register(
+            #    lambda x, y: x + y, ...)
+            code_text = f"wrapper({code_text}"
```
sfc-gh-sfan, line 339: I'm wondering if we could add `code_text` to the cases as well. This makes it easier to understand these handlings :)

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+        to_import.update(
+            extract_submodule_imports(
+                func, [v for v in ref_objects.values() if isinstance(v, ModuleType)]
+            )
+        )
+        for v in [
+            v
+            for v in ref_objects.values()
+            if inspect.isclass(v) and v.__module__ == func_module_name
+        ]:
```
sfc-gh-sfan, line 437: Let's extract the list to a variable so it is easier to read?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+            get_class_references(v, func, ref_objects, classes_to_generate)
+    elif isinstance(func, BuiltinFunctionType):
```
sfc-gh-sfan, line 439: Do we expect more types? Are we ignoring them or should we raise?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+        code_as_comment: Whether the code will be generated as comment.
+
+    Returns:
+        A tuple of two strings, the first one is the source code of referenced functions defined in the same module
+        as the target function's, and the second is the source code of referenced variables.
+    """
+    func_module_name = func.__module__
+    func_text = ""
+    global_vars_text = ""
+    for name, obj in ref_objects.items():
```
sfc-gh-sfan, line 472: It seems like we are doing different actions based on different predicates and there are very few if any shared code, how about do early return style for all:

```
for ...:
  if ...:
    ...
    continue
  if ...:
    ...
    continue
...
```

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+    Returns:
+        A string of generated imports.
+    """
+    imports = [
+        f"import {name + ' as ' if name != alias else ''}{alias}"
+        for name, alias in sorted(to_import)
+    ]
+    for module, name_alias_pairs in sorted(to_import_from_module.items()):
+        classes = ", ".join(
+            f"{name + ' as ' if name != alias else ''}{alias}"
```
sfc-gh-sfan, line 537: If we explicitly check for `name != alias`, then we might as well just set `alias` to `None`, and write a small helper function to generate this?

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+        func_source_code = textwrap.dedent(inspect.getsource(func))
+        if not code_as_comment:
+            func_source_code = remove_function_udf_annotation(func_source_code)
+        if not is_lambda(func):
+            complete_source_code = f"{complete_source_code}\n{func_source_code}"
+        func_assignment = (
+            get_lambda_code_text(func_source_code) if is_lambda(func) else func.__name__
+        )
+    else:
+        # BuiltinFunctionType
```
sfc-gh-sfan, line 572: We might want to claim this in `isinstance` and raise for `else`?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -376,11 +378,20 @@ def generate_python_code(
         pickled_func = pickle_function(func)
     args = ",".join(arg_names)
 
+    try:
+        source_code_comment = (
+            code_generation.generate_source_code(func) if source_code_display else ""
+        )
+    except Exception as exc:
+        logger.debug(f"Source code comment could not be generated due to error {exc!r}")
```
sfc-gh-sfan, line 386: I'm wondering if this could be part of telemetry that is sent over to snowhouse? Even some success/failure counter might be some useful signal.

tests/unit/test_code_generation.py
```diff
@@ -0,0 +1,427 @@
...
+    var = 123
+
+    def udf():
+        return var + 1
+
+    assert (
+        generate_source_code(udf)
+        == """\
+# The following comment contains the UDF source code generated by snowpark-python for explanatory purposes.
+# var  # variable of type <class 'int'>
```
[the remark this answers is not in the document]
sfc-gh-aling: for code generated as comments, I have removed the part of serializing objects -- cloudpickle will serialize the objects. I feel there is less value in serialization again by us if code are comments.

but for code generated for execution, we will exactly be having the `var = pickle.loads(<hexcode>)`.

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+DELETE_GLOBAL = opcode.opmap["DELETE_GLOBAL"]
+LOAD_GLOBAL = opcode.opmap["LOAD_GLOBAL"]
+GLOBAL_OPS = (STORE_GLOBAL, DELETE_GLOBAL, LOAD_GLOBAL)
+
+CODE_AS_COMMENT_HINT = (
+    "The following comment contains the UDF source code"
+    " generated by snowpark-python for explanatory purposes.\n"
+)
+CODE_HEADER = """\
+from __future__ import annotations
```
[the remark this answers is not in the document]
sfc-gh-aling: it's to postpone evaluation of annotations, type hint could contain classes which are not necessarily required to be imported.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: it is already checked at the beginning of `generate_source_code` via `check_func_type`. I could add a comment in this func.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aling: I have created a separate JIRA for this: https://snowflakecomputing.atlassian.net/browse/SNOW-651381 and would like to make a separate PR for it.

trying to restrict the scope of this PR.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: I personally prefer `if-elif-else` style.

my personal experience of early return is that the executions results are chained sequentially in different clauses while here we are dealing with completely different predicates.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: I have introduced a named tuple for the ImportNameAliasPair and defaults both name and alias None.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: func type check is done in `check_func_type` which will be called at the very beginning when generating code.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: yes, instr.arg is 1-1 mapping parameter position info for the co_names.

as for none in dict, it is copied from cloudpickle, cloudpickle uses dict for deterministic ordering of pickled bytes. we don't need it. I've changed to use set.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: so far it won't break execution in my test code.

I would say we need to write more test to ensure there's no impact. probably more language reference to understand how python uses them.

we can do more test when I put this feature in the beta branch for source code execution.
it doesn't impact the source code generation.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -376,11 +378,20 @@ def generate_python_code(
...
+        source_code_comment = ""
```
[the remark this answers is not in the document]
sfc-gh-aling: I've added the comment with the exception and created a JIRA for telemetry

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-sfan: Right. I guess I'm just a bit paranoid and feels it is always good to explicitly check our assumption lol, so that if one day there is a new type, we won't have something goes wrong silently :)

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: good catch, I think it's worthwhile to add strip the string first in this function

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: you mean adding some test cases? yes, I added some

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: make sense to me, I've updated the code

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: I have added more comments, I feel I should write a essay for it (no joke)

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: I have added test for it, it will be 
```python
import io
import pandas
import pandas.io
```

it's not easy to locate the exact the submodule that is imported. the current implementation imports level by level.

I have created a JIRA for it: https://snowflakecomputing.atlassian.net/browse/SNOW-651634

tests/integ/test_stored_procedure.py
```diff
@@ -820,7 +820,12 @@ def return1(session_: Session) -> str:
     for row in describe_res:
         if row[0] == "packages":
             assert "snowflake-snowpark-python" in row[1]
-            break
+        if row[0] == "body":
+            # This line shall be removed after SP source code display is supported
```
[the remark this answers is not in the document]
sfc-gh-aling: updated

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+        closures = {
+            k: v
+            for k, v in zip(
+                code_object.co_freevars,
+                list(map(lambda x: x.cell_contents, func.__closure__)),
+            )
+        }
+        for k, v in closures.items():
+            ref_objects[k] = v
+            # if the closure item is a function and is not itself (recursive) and has not been visited
```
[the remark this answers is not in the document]
sfc-gh-aling: updated

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+    # 2. resolve function closure references
+    if func.__closure__ is not None:
+        closures = {
+            k: v
+            for k, v in zip(
+                code_object.co_freevars,
+                list(map(lambda x: x.cell_contents, func.__closure__)),
+            )
+        }
+        for k, v in closures.items():
```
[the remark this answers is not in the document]
sfc-gh-aling: updated

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+        and the second one is alias.
+
+    """
+    imports = set()
+    for module in top_level_modules:
+        module_prefix = f"{module.__name__}."
+        for name in [m for m in sys.modules if m.startswith(module_prefix)]:
+            tokens = set(name[len(module_prefix) :].split("."))
+            if not tokens - set(
+                func.__code__.co_names
```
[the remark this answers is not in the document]
sfc-gh-aling: updated and added comments

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: it could be an alias, I've taken Shixuan's idea to use named tuple with default values. it should now look more intuitive.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: this function is just to handle sub module, top level modules are already taken care by `get_func_references`.

I've updated the comments.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: updated, added more comments to make it clear

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-sfan: Ah I just meant what the `code_text` looks like in the example showed in comment :)

src/snowflake/snowpark/_internal/code_generation.py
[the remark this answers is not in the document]
sfc-gh-aling: you're right, a set should be enough for us, this part is inspired by cloudpickle, they use dict to keep pickle bytes ordered which we don't need.

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,594 @@
...
+
+    for base_class in cls.__bases__:
+        base_class_name = base_class.__qualname__
+        top_level_class = base_class_name.split(".")[0]
+        if base_class_name == "object":
+            continue
+        elif base_class.__module__ == func_module_name:
+            # if base class is from the same module, we need to parse the class as well as generate code for the class
+            # False in the tuple means this is not a nested class
+            inferred_classes.append((base_class, False))
```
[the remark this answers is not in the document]
sfc-gh-aling: updated

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: updated

CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 # Release History
+## 0.9.0 (Unreleased)
+
+### New Features:
+- Added support for displaying source code as comments in the generated scripts when registering UDFs.
+This feature is turned on by default and could be turned off by passing the new keyword argument `source_code_display` as `False` when calling `register()` or `@udf()`.
```
sfc-gh-mabrennan, line 6: I suggest changing:
"This feature is turned on by default and could be turned off by passing the... "
to
"This feature is turned on by default. To turn it off, pass the... "

sfc-gh-mabrennan (commented):
Docstrings look good. I made one suggestion.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-jdu: got it

sfc-gh-jdu (approved):

CHANGELOG.md
sfc-gh-aling: thanks, updated!

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -0,0 +1,628 @@
...
+        for name in [m for m in sys.modules if m.startswith(module_prefix)]:
+            tokens = set(name[len(module_prefix) :].split("."))
+            if (
+                not tokens - func_co_names
+            ):  # only add imports that co_names contains all the tokens
+                # if any token does not show up in func_co_names, it means the module is not used by
+                # func, thus there is no need to import the module
+                # if all tokens showing up in the func_co_names, it means the module *might* be used by
+                # func, import the token
+                imports.add(ImportNameAliasPair(name=name))
```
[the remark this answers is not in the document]
sfc-gh-aling: right, module alias would be detected by `get_func_references`, inferred submodule imports don't have alias.
I have added comment to explain it.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: I've updated the comment to include `code_text`

tests/unit/test_code_generation.py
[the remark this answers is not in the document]
sfc-gh-aling: I was originally thinking of to add it when making the PR for code execution. but definitely no harm if we want to add one now -- added one test case for variable serialization.

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: updated to raise error for other unsupported types

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -351,11 +349,13 @@ def get_lambda_code_text(code_text: str) -> str:
             # session.udf.register(
             #    lambda x, y: x + y, ...
             # )
+            # code_text in this case is "lambda x, y: x + y, ..."
             code_text = f"wrapper({code_text})"
         elif "unmatched ')'" in str(exc):
             # handle case like:
             # session.udf.register(
             #    lambda x, y: x + y, ...)
+            # code_text in this case is "lambda x, y: x + y, ...)"
```
sfc-gh-sfan, line 358: This final closing parenthesis is very interesting :)

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/code_generation.py
sfc-gh-aling: yes, I was surprised when seeing this -- "Python what are u doing here?!"

PR #438, opened 2022-08-16
SNOW-637329 Add in telemetry for the Python packages in a Snowpark user's environment

sfc-gh-kwagner:
Description

Testing
Added in testing to make sure this is safe if exceptions or empty lists show up

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -22,6 +22,8 @@ from json import JSONEncoder
 from random import choice
 from typing import IO, Any, Dict, Iterator, List, Optional, Type
 
+import pkg_resources
+
 import snowflake.snowpark
 from snowflake.connector.cursor import ResultMetadata, SnowflakeCursor
 from snowflake.connector.description import OPERATING_SYSTEM, PLATFORM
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -522,3 +524,9 @@ def func_call_wrapper(*args, **kwargs):
         return func_call_wrapper
 
     return deprecate_wrapper
+
+
+def get_local_packages():
```
sfc-gh-jdu, line 529: This is to get all installed packages in the local environment, instead of the current imported packages? Which one do we want? I assumed the latter is more precise because people will only import packages when they need them in the current environment, but they can install many packages before.

sfc-gh-kwagner: Yes I can rename the function to `get_installed_packages()`. From what I understand, we want to get all the packages in a user's environment because we not only want to know the packages that they are currently using but all of the packages that they could use so that we can get those packages in Anaconda before a customer requests it. Correct me if I'm wrong @sfc-gh-achandrasekaran

sfc-gh-sfan: I think the information about imported packages should be queryable from snowhouse because that is part of the generated UDF/sprocs. Installed packages seem useful to provide some package popularity info.

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/utils.py
sfc-gh-achandrasekaran: Can we get both? If not, the priority is to get the packages that are being imported, not just installed.

PR #448, opened 2022-08-23
SNOW-650679: Add alias support to TableFunctionCall

sfc-gh-aalam:

   Fixes #SNOW-650679

   Add support to alias the columns names that come from the output of table function joins. The support can be seen for functions `select`, `join_table_function`, `with_column` and `with_columns`

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -1781,6 +1781,22 @@ class DataFrame:
             ----------------------------------------------------------------------------------------
             <BLANKLINE>
 
+        Example 4
+            Lateral join a table function with aliasing the output column names:
+
+            >>> from snowflake.snowpark.functions import table_function
+            >>> split_to_table = table_function("split_to_table")
+            >>> df = session.sql("select 'James' as name, 'address1 address2 address3' as addresses")
+            >>> df.join_table_function(split_to_table(col("addresses"), lit(" ")).alias("seq", "idx", "val")).show()
+            ------------------------------------------------------------------
+            |"NAME"  |"ADDRESSES"                 |"SEQ"  |"IDX"  |"VAL"     |
+            ------------------------------------------------------------------
+            |James   |address1 address2 address3  |1      |1      |address1  |
+            |James   |address1 address2 address3  |1      |2      |address2  |
+            |James   |address1 address2 address3  |1      |3      |address3  |
+            ------------------------------------------------------------------
+            <BLANKLINE>
+
         Args:
 
             func_name: The SQL function name.
@@ -1797,6 +1813,14 @@ class DataFrame:
         func_expr = _create_table_function_expression(
             func, *func_arguments, **func_named_arguments
         )
+        if func_expr.aliases:
+            join_plan = self._session._analyzer.resolve(
+                TableFunctionJoin(self._plan, func_expr)
+            )
+            old_cols, new_cols = _get_cols_after_join_table(
+                func_expr, self._plan, join_plan
+            )
+            return self._with_plan(Project([*old_cols, *new_cols], join_plan))
         return DataFrame(self._session, TableFunctionJoin(self._plan, func_expr))
 
     @df_api_usage
@@ -1873,7 +1897,9 @@ class DataFrame:
         )
 
     @df_api_usage
-    def with_column(self, col_name: str, col: Column) -> "DataFrame":
+    def with_column(
+        self, col_name: str, col: Union[Column, TableFunctionCall]
+    ) -> "DataFrame":
         """
         Returns a DataFrame with an additional column with the specified name
         ``col_name``. The column is computed by using the specified expression ``col``.
@@ -1881,7 +1907,7 @@ class DataFrame:
         If a column with the same name already exists in the DataFrame, that column is
         replaced by the new column.
 
-        Example::
+        Example 1::
 
             >>> df = session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"])
             >>> df.with_column("mean", (df["a"] + df["b"]) / 2).show()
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -1893,14 +1919,33 @@ def with_column(self, col_name: str, col: Column) -> "DataFrame":
...
+            ...         yield (a + b, )
+            >>> df = session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"])
+            >>> df.with_column("total", sum_udtf(df.a, df.b)).show()
+            -----------------------
+            |"A"  |"B"  |"TOTAL"  |
+            -----------------------
+            |3    |4    |7        |
+            |1    |2    |3        |
+            -----------------------
+            <BLANKLINE>
```
sfc-gh-aalam, line 1937: The row order in output for `show()` is not the same all the time causing doctests to fail. How should I structure this to pass them?

sfc-gh-sfan: Let's maybe add a `sort(col("A"))`? Maybe there are better ways that I'm not aware of.

src/snowflake/snowpark/table_function.py
```diff
@@ -106,6 +108,22 @@ def over(
             new_table_function._order_by = order_spec
         return new_table_function
 
+    def alias(self, *aliases: str) -> "TableFunctionCall":
```
sfc-gh-sfan, line 111: Can table function itself has alias like table? If that is the case, this name might be confusing I think? Should we name it as `col_alias` for example?

src/snowflake/snowpark/dataframe.py
```diff
@@ -1893,14 +1919,33 @@ def with_column(self, col_name: str, col: Column) -> "DataFrame":
...
+
         Args:
             col_name: The name of the column to add or replace.
-            col: The :class:`Column` to add or replace.
+            col: The :class:`Column` or :class:`table_function.TableFunctionCall` with single column output to add or replace.
```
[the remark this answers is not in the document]
sfc-gh-aalam: It should fail at 
```
    if aliases:
        if len(new_cols) != len(aliases):
            raise ValueError(
                f"The number of aliases should be same as the number of cols added by table function. "
                f"Columns added by table function are {new_cols} and aliases given are {aliases}"
            )
```

Let me add a test to verify this.

src/snowflake/snowpark/table_function.py
sfc-gh-aalam: I don't think we can. The code at present does not support the table function to be renamed. I also don't see a need to rename a table function because we cal always assign a new variable to point to an existing table function, right?

sfc-gh-jdu: Should we also consider adding `as_` as what we did in column.py?

tests/integ/test_dataframe.py
```diff
@@ -387,6 +406,96 @@ def process(self, n: int) -> int:
...
+    with pytest.raises(ValueError) as ex_info:
+        df.select(multiplier_udtf(df.a).alias("double"))
+    assert (
+        "The number of aliases should be same as the number of cols added by table function"
+        in str(ex_info)
+    )
+
+
+def test_with_column(session):
+    df = session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"])
```
sfc-gh-jdu, line 427: Probably also worth adding tests for built-in table functions

sfc-gh-jdu (approved):

tests/integ/test_dataframe.py
sfc-gh-aalam: Are there any built-in functions that have a single column output?

sfc-gh-sfan (approved):

tests/integ/test_dataframe.py
sfc-gh-jdu: Probably this one? https://docs.snowflake.com/en/sql-reference/functions/generator.html (or you can search in all snowflake table functions https://docs.snowflake.com/en/sql-reference/functions-table.html)

sfc-gh-jdu: Then is it possible to support `df.with_columns(["a", "b", "c"], [col("x"), my_udtf("y", "z")])` where `my_udtf` returns 2 columns?

sfc-gh-aalam: the current version of the code does not support that. I could try to make it work

src/snowflake/snowpark/table_function.py
```diff
@@ -106,6 +108,33 @@ def over(
             new_table_function._order_by = order_spec
         return new_table_function
 
+    def as_(self, *aliases: str) -> "TableFunctionCall":
+        """Alias the output columns from the output of this table function call.
+
+        Args:
+            aliases: An iterable of unique column names that do not collide with columns names after join with the main table.
```
sfc-gh-mabrennan, line 115: Change "columns" to "column".

src/snowflake/snowpark/table_function.py
```diff
@@ -106,6 +108,33 @@ def over(
...
+        Raises:
+            ValueError: Raises error when the aliases are not unique after being canonicalized.
+        """
+        return self.alias(*aliases)
+
+    def alias(self, *aliases: str) -> "TableFunctionCall":
+        """Alias the output columns from the output of this table function call.
+
+        Args:
+            aliases: An iterable of unique column names that do not collide with columns names after join with the main table.
```
sfc-gh-mabrennan, line 126: "columns" -> "column"

sfc-gh-mabrennan (commented):
Nice job on the docstrings. Thank you!

tests/integ/test_dataframe.py
sfc-gh-aalam: updated code to make it work like you suggested. also added tests that check this

src/snowflake/snowpark/table_function.py
```diff
@@ -106,6 +108,33 @@ def over(
             new_table_function._order_by = order_spec
         return new_table_function
 
+    def as_(self, *aliases: str) -> "TableFunctionCall":
```
sfc-gh-jdu, line 111: actually you can just do `as_ = alias` (we should also do that in column.py actually 🤦 )

tests/integ/test_dataframe.py
sfc-gh-jdu: nice

sfc-gh-jdu (approved):

PR #459, opened 2022-08-30
SNOW-654033 add a randomized sleep to handle rmtree error

sfc-gh-aalam:

   Fixes #SNOW-654033

   Fix flaky test for windows environment caused by `rmtree`

### DIFF
--- a/tests/integ/scala/test_file_operation_suite.py
+++ b/tests/integ/scala/test_file_operation_suite.py
@@ -6,7 +6,6 @@ import os
 import random
 import shutil
 import string
-from time import sleep
 
 import pytest
 
@@ -25,24 +24,12 @@ def random_alphanumeric_name():
     )
 
 
-def onerror(func, path, exc_info):
-    retries = 3
-    for retry in range(retries):
-        try:
-            sleep(1)
-            func(path)
-            return
-        except Exception as ex:
-            if retry == retries - 1:
-                raise ex
-
-
 @pytest.fixture(scope="module")
 def temp_source_directory(tmpdir_factory):
     directory = tmpdir_factory.mktemp("snowpark_test_source")
     yield directory
     if not is_in_stored_procedure():
-        shutil.rmtree(str(directory), onerror=onerror)
+        shutil.rmtree(str(directory))
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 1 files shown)


tests/integ/scala/test_file_operation_suite.py
```diff
@@ -24,20 +25,25 @@ def random_alphanumeric_name():
     )
 
 
+def onerror(func, path, exc_info):
+    sleep(1 + random.random() * 4)
```
sfc-gh-yixie, line 29: does it retry automatically after onerror is called?

sfc-gh-aalam: `onerror` is called after the first error is encountered. So in a way we are giving two tries for `rmtree` to work

sfc-gh-jdu: so if the second try still fails, it will fail?

sfc-gh-aalam: yes

sfc-gh-jdu (approved):

tests/integ/scala/test_file_operation_suite.py
sfc-gh-yixie: I prefer removing the random seconds. If it fails again we have no clue how long time is good.
If we want to wait more time, retry every 1 second and retry 3 or 5 times is better.

sfc-gh-aalam: how many times should we retry?

sfc-gh-aalam: I misread your comment. You meant try every 1 sec with 3-5 retries.

sfc-gh-yixie (approved):

sfc-gh-sfan (approved):

tests/integ/scala/test_file_operation_suite.py
```diff
@@ -286,6 +286,7 @@ def test_put_stream_with_one_file_twice(session, temp_stage, path1):
     second_result = session.file.put_stream(
         fd, f"{stage_with_prefix}/{file_name}", overwrite=False
     )
+    fd.close()
```
sfc-gh-aalam, line 289: precommit failed again after the earlier "fix" so I check code for all file `open()`s and found one without a `close()`

sfc-gh-sfan: Should we use `with` in the test, or maybe not using file stream but just `BytesIO`?

sfc-gh-aalam: We are using `with` is other tests except for two. First, to check if we support two uploads from the same `fd`, and second, to check if we raise error on closed `fd`.

PR #464, opened 2022-08-31
SNOW-630799 Support table_type in session#write_pandas

sfc-gh-sfan:
Description

Now that connector's latest change has been released, we could
use this from the connector

Testing

integ test

   Fixes SNOW-630799

   This change adds table_type in write_pandas, which make uses of `write_pandas` from connector.

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -7,12 +7,13 @@ import decimal
 import json
 import logging
 import os
+import warnings
 from array import array
 from functools import reduce
 from logging import getLogger
 from threading import RLock
 from types import ModuleType
-from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, Union
+from typing import Any, Dict, Iterable, List, Literal, Optional, Set, Tuple, Union
 
 import cloudpickle
 import pkg_resources
@@ -988,6 +989,7 @@ class Session:
         auto_create_table: bool = False,
         create_temp_table: bool = False,
         overwrite: bool = False,
+        table_type: Literal["", "temp", "temporary", "transient"] = "",
     ) -> Table:
         """Writes a pandas DataFrame to a table in Snowflake and returns a
         Snowpark :class:`DataFrame` object referring to the table where the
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -1043,12 +1047,33 @@ def write_pandas(
...
         """
+        if create_temp_table:
+            warnings.warn(
+                "create_temp_table is deprecated. We still respect this parameter when it is True but "
+                'please consider using `table_type="temporary"` instead.',
+                DeprecationWarning,
+                # warnings.warn -> write_pandas
+                stacklevel=2,
+            )
+            table_type = "temporary"
```
sfc-gh-aalam, line 1070: should we take care of case when someone sets `create_temp_table=True` and `table_type='transient'`. In this case `table_type` will be overwritten to `temporary` (which won't be a problem if it means the same thing as `transient`)

tests/integ/test_pandas_to_df.py
```diff
@@ -248,9 +309,7 @@ def test_write_pandas_temp_table_and_irregular_column_names(session):
     )
     table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
     try:
-        session.write_pandas(
-            pd, table_name, auto_create_table=True, create_temp_table=True
-        )
+        session.write_pandas(pd, table_name, auto_create_table=True, table_type="temp")
```
sfc-gh-aalam, line 312: same point. does making `table_type="transient"` work?

src/snowflake/snowpark/session.py
sfc-gh-jdu: I think create_temp_table will just overwrite table_type. Maybe worth mentioning it in docstring and a test?

sfc-gh-sfan: Good point. I'll add this to doc string as well as making the "no breaking change" test to cover this overwrite

tests/integ/test_pandas_to_df.py
sfc-gh-sfan: Oh I overlooked this and thought the usage of temp table is just for test purpose. I'll make it parametrized and cover transient as well.

src/snowflake/snowpark/session.py
```diff
@@ -1013,17 +1015,21 @@ def write_pandas(
...
-            create_temp_table: The to-be-created table will be temporary if this is set to ``True``.
+            create_temp_table: (Deprecated) The to-be-created table will be temporary if this is set to ``True``. Note
+                that to avoid breaking changes, currently when this is set to True, it overrides ``table_type``.
             overwrite: Default value is ``False`` and the Pandas DataFrame data is appended to the existing table. If set to ``True`` and if auto_create_table is also set to ``True``,
                 then it drops the table. If set to ``True`` and if auto_create_table is set to ``False``,
                 then it trunctates the table. Note that in both cases (when overwrite is set to ``True``) it will replace the existing
                 contents of the table with that of the passed in Pandas DataFrame.
+            table_type: The table type of table to be created. The supported values are: ``temp``, ``temporary``,
+                        and ``transient``. An empty string means to create a permanent table. Learn more about table
+                        types in https://docs.snowflake.com/en/user-guide/tables-temp-transient.html.
```
sfc-gh-jdu, line 1026: let's use a hyperlink in docstring instead of the raw link.

src/snowflake/snowpark/session.py
```diff
@@ -1043,12 +1049,33 @@ def write_pandas(
...
+            warnings.warn(
+                "create_temp_table is deprecated. We still respect this parameter when it is True but "
+                'please consider using `table_type="temporary"` instead.',
+                DeprecationWarning,
+                # warnings.warn -> write_pandas
+                stacklevel=2,
+            )
+            table_type = "temporary"
+
+        if table_type and table_type.lower() not in ["temp", "temporary", "transient"]:
```
sfc-gh-jdu, line 1074: can we make these three valid values as a constant set?

sfc-gh-jdu (approved):

PR #467, opened 2022-09-02
SNOW-622058: Use job-scoped temp object in snowpark API

sfc-gh-aling:

   Fixes SNOW-622058

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -528,9 +528,9 @@ def deprecate(*, deprecate_version, extra_warning_text="", extra_doc_string=""):
     return deprecate_wrapper
 
 
-def get_temp_type_for_object(use_scoped_temp_objects: bool, is_generated: bool):
+def get_temp_type_for_object(use_scoped_temp_objects: bool, is_generated: bool) -> str:
     return (
-        TEMPORARY_STRING
-        if not use_scoped_temp_objects
-        else (SCOPED_TEMPORARY_STRING if is_generated else TEMPORARY_STRING)
+        SCOPED_TEMPORARY_STRING
+        if use_scoped_temp_objects and is_generated
+        else TEMPORARY_STRING
     )
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -117,6 +117,8 @@ SKIP_LEVELS_TWO = (
 SKIP_LEVELS_THREE = (
     3  # limit traceback to return up to 3 stack trace entries from traceback object tb
 )
+TEMPORARY_STRING = "TEMPORARY"
+SCOPED_TEMPORARY_STRING = "SCOPED TEMPORARY"
 
 SUPPORTED_TABLE_TYPES = ["temp", "temporary", "transient"]
 
@@ -524,3 +526,11 @@ def deprecate(*, deprecate_version, extra_warning_text="", extra_doc_string=""):
         return func_call_wrapper
 
     return deprecate_wrapper
+
+
+def get_temp_type_for_object(use_scoped_temp_objects: bool, is_generated: bool):
+    return (
+        TEMPORARY_STRING
+        if not use_scoped_temp_objects
+        else (SCOPED_TEMPORARY_STRING if is_generated else TEMPORARY_STRING)
+    )
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -75,6 +75,7 @@ from snowflake.snowpark._internal.utils import (
     zip_file_or_directory_to_stream,
 )
 from snowflake.snowpark.column import Column
+from snowflake.snowpark.context import _use_scoped_temp_objects
 from snowflake.snowpark.dataframe import DataFrame
 from snowflake.snowpark.dataframe_reader import DataFrameReader
 from snowflake.snowpark.file_operation import FileOperation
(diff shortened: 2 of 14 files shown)


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -524,3 +526,11 @@ def func_call_wrapper(*args, **kwargs):
         return func_call_wrapper
 
     return deprecate_wrapper
+
+
+def get_temp_type_for_object(use_scoped_temp_objects: bool, is_generated: bool):
+    return (
```
sfc-gh-jdu, line 532: IIRC, it can be 
```
return SCOPED_TEMPORARY_STRING if use_scoped_temp_objects and is_generated else TEMPORARY_STRING
```
?

src/snowflake/snowpark/session.py
```diff
@@ -113,6 +114,10 @@
 
 _session_management_lock = RLock()
 _active_sessions: Set["Session"] = set()
+_use_scoped_temp_objects = True
```
sfc-gh-jdu, line 117: let's put it in context.py, which is similar to the opt-out option of @sfc-gh-yixie's simplifier? Also, do we need prefix underscore?

src/snowflake/snowpark/session.py
```diff
@@ -248,6 +253,16 @@ def __init__(self, conn: ServerConnection) -> None:
         self._plan_builder = SnowflakePlanBuilder(self)
         self._last_action_id = 0
         self._last_canceled_id = 0
+        self._use_scoped_temp_objects = bool(
+            _use_scoped_temp_objects
+            and (
+                conn._conn._session_parameters.get(
+                    _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, True
```
sfc-gh-jdu, line 260: Does it handle the scenario that this parameter in connection parameters is in lower case? You can just check `conn._lower_case_parameters`. Also, to make code more clear:
```
self._use_scoped_temp_objects = _use_scoped_temp_objects and conn._lower_case_parameters.get(_PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, True)
```
?

tests/unit/test_analyzer_util_suite.py
```diff
@@ -0,0 +1,138 @@
...
+
+    assert (
+        analyzer_utils.create_table_statement(
+            temp_table_name,
+            temp_schema_name,
+            table_type="",
+            use_scoped_temp_objects=True,
+            is_generated=True,
+        )
+        == f" CREATE    TABLE {temp_table_name}({temp_schema_name})"
```
sfc-gh-jdu, line 137: I'm a little surprised that it has so many spaces 🤦

sfc-gh-sfan (commented):
I think we could only merge this once the parameter is available on the server and is populated through sp connector?

src/snowflake/snowpark/session.py
sfc-gh-jdu: I think it does because the connection is created from lower case parameters? https://github.com/snowflakedb/snowpark-python/blob/952c9b31efc35b72b7c06f44fb4094a6c778dc81/src/snowflake/snowpark/_internal/server_connection.py#L138. 

I think connection parameters will be set by users. So the user can potentially pass in the lower case parameter (but I'm not sure whether it'll be converted to upper case in the connector).

src/snowflake/snowpark/_internal/utils.py
sfc-gh-aling: good catch

src/snowflake/snowpark/session.py
sfc-gh-aling: you're right @sfc-gh-jdu , the `_lower_case_parameters` is client-side parameters which contains `schema`, `database` etc.

but it doesn't have the server side session parameter so I'm afraid it can't be used to get the `_PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING`

tests/unit/test_analyzer_util_suite.py
sfc-gh-aling: yeah.. there're many spaces in our string constants.

src/snowflake/snowpark/session.py
sfc-gh-aling: moving this into `context.py` would result in circular imports.. I don't have a good solution for this (I do not want to do import while in the run time..) since I want to keep this parameter private, I think it's fine to leave it in the session.py?

also for the underscore, this change should be transparent to users, however, we do want to have a flag on client side to turn it off in case something went wrong.

and ultimately, we can just remove this config.

sfc-gh-jdu: can you just reference `snowflake.snowpark.context._use` here? I'd prefer to keep these options in one place and consistent.

src/snowflake/snowpark/session.py
sfc-gh-jdu: I see, I misread it as a client param. Then it should be good

tests/unit/test_analyzer_util_suite.py
sfc-gh-jdu: sure

src/snowflake/snowpark/session.py
sfc-gh-aling: based upon our OOB discussion, removed the parameter to context. thanks for the suggestion!

sfc-gh-jdu (approved):

PR #468, opened 2022-09-03
Implement a cleanup step for cached DataFrames

### LINKED ISSUE #463: Drop cached temporary tables that are no longer being referened
`cache_result` does not clean up a temporary table when it is no longer being referenced. This results in unnecessary storage credits throughout the entirety of the session whenever a result is re-cached, which is common thing to do in an interactive session.

```python
>>> dff = session.create_dataframe([1,2,3])

>>> dff.cache_result().explain()

---------DATAFRAME EXECUTION PLAN----------
Query List:
1.
SELECT  *  FROM (SNOWPARK_TEMP_TABLE_X3FCJ1U38A)
...

--------------------------------------------

>>> dff.cache_result().explain()

---------DATAFRAME EXECUTION PLAN----------
Query List:
1.
SELECT  *  FROM (SNOWPARK_TEMP_TABLE_Z9H68STVDH)
....
```

Ideal behavior would be a temporary table cleanup by defining a  `__del__` method for `snowflake.snowpark.dataframe.DataFrame`. I'll admit that I don't have a deep understanding of snowpark's innerworkings, but on the surface it seems like all cached results are essentialy a `select * from (temp_table_name)` which seems generalizable. In pseudo-ish code, it could be something like:
```python
# Module: snowflake.snowpark.dataframe.DataFrame

class DataFrame:
...

    def __del__(self):
        if self.is_cached:
            temp_table = # extract name from self.queries string or some other method
            self._session.sql(f'drop table {temp_table').collect()

 ```
Avoid overcharging on storage credits :)

### PR DESCRIPTION
**Pull request by mbkupfer**
Add an explicit SQL drop command that drops the underlying temporary table for cached DataFrames.

   Fixes #463 

   Add a `__del__` finalizer method that drops the temporary table that gets created for cached results.

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -414,10 +414,12 @@ class DataFrame:
         session: Optional["snowflake.snowpark.Session"] = None,
         plan: Optional[LogicalPlan] = None,
         is_cached: bool = False,
+        cached_temp_table_name: str = None,
     ) -> None:
         self._session = session
         self._plan = session._analyzer.resolve(plan)
         self.is_cached: bool = is_cached  #: Whether it is a cached dataframe
+        self._temp_table_name: str = cached_temp_table_name  # For explicitly dropping temp table at garbage collection
 
         self._reader: Optional["snowflake.snowpark.DataFrameReader"] = None
         self._writer = DataFrameWriter(self)
@@ -433,6 +435,11 @@ class DataFrame:
         self.dropna = self._na.drop
         self.fillna = self._na.fill
         self.replace = self._na.replace
+
+    def __del__(self):
+        """Clean up step that explicitly drops the temporary table if DataFrame is cached"""
+        if self.is_cached:
+            self._session.sql(f"DROP TABLE IF EXISTS  {self._temp_table_name}").collect()
 
     @property
     def stat(self) -> DataFrameStatFunctions:
@@ -2786,7 +2793,7 @@ class DataFrame:
             ),
         )
         new_plan = self._session.table(temp_table_name)._plan
-        return DataFrame(session=self._session, plan=new_plan, is_cached=True)
+        return DataFrame(session=self._session, plan=new_plan, is_cached=True, cached_temp_table_name=temp_table_name)
 
     @df_collect_api_telemetry
     def random_split(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mbkupfer** [comment]

**Comment by mbkupfer** [comment]
recheck

**Comment by sfc-gh-yixie** [comment]
Hi @mbkupfer thank you for putting out the PR. Agree that it's important to clean up cached tables. We'll discuss the solution in the next week. Will keep you updated. Our questions are:
1. Will the clean-up cause any problems after a DataFrame is destructed. The underlying query plan may be referenced by other DataFrames.
2. Connecting to a network resource from `__del__` may impact the Python garbage collection.
3. Aside from auto-cleanup, do we give users an API to clean up whenever they want. Do you think this API will be helpful?

**Comment by mbkupfer** [comment]
@sfc-gh-yixie those are some valid points. Gave it some thought and here are my replies:

> Will the clean-up cause any problems after a DataFrame is destructed. The underlying query plan may be referenced by other DataFrames.

Yes, this is problematic indeed. IMHO I'd prefer to get an error as opposed to the program quietly using a previous table.  
Moreover, this is actually more systemic to the overall package functionality and not just cached results. For instance, there would be no "safeguards" when end users don't use `cache_result` such as in the case below. 

```python
df = session.create_dataframe([1,2], schema=['col'])
df2 = df.where(F.col('col') > 1)
df = session.create_dataframe([3,4,5], schema=['col'])
df2.show()
```
Perhaps a way around this would be to insert additional clean up steps in the `__del__` function that checks all the referents to the deleted dataframe, regardless of being cached or not, and ensure that those dataframes get deleted as well. It's a bit clunky, but it ensures that users avoid any pitfalls. What do you think?
Also want to note that this is out of scope for the temp table cleanup issue.

> Connecting to a network resource from `__del__` may impact the Python garbage collection.

True. I guess we can just call it a best attempt convenience feature with no guarantees. 😏 

> Aside from auto-cleanup, do we give users an API to clean up whenever they want. Do you think this API will be helpful?

I appreciate you seeking my opinion on this, and of course I defer the final decision to your team, but I don't really think this is necessary. I'd rather have the library use optimized intelligence to do any cleanup on its own as opposed to users blindly sending unnecessary clean up commands. Maybe if it comes with use-cases and best practices then sure, but just handing over the option may introduce confusion.

PR #471, opened 2022-09-06
SNOW-645479: Support qualified table name in session._table_exists

sfc-gh-aling:

   Fixes SNOW-645479

using `show table like xxx in yyy` to check whether table exists in schema/database.

I'm following the spec https://docs.snowflake.com/en/sql-reference/name-resolution.html to implement object parsing.

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -1663,56 +1663,47 @@ class Session:
         # note: object name could have dots, e.g, a table could be created via: create table "abc.abc" (id int)
         # currently validate_object_name does not allow it, but if in the future we want to support the case, we need to
         # update the implementation accordingly in this method
-        try:
-            qualified_table_name = table_name.split(".")
-            if len(qualified_table_name) == 1:
-                # name in the form of "table"
-                tables = self._run_query(f"show tables like '{table_name}'")
-            elif len(qualified_table_name) == 2:
-                # name in the form of "schema.table" omitting database
-                # schema: qualified_table_name[0]
-                # table: qualified_table_name[1]
-                tables = self._run_query(
-                    f"show tables like '{qualified_table_name[1]}' in schema {qualified_table_name[0]}"
-                )
-            elif len(qualified_table_name) == 3:
-                # name in the form of "database.schema.table"
-                # database: qualified_table_name[0]
-                # schema: qualified_table_name[1]
-                # table: qualified_table_name[2]
-                if qualified_table_name[1] == "":
-                    # Double-Dot Notation: "database..table" in which the schema is empty
-                    tables = self._run_query(
-                        f"show tables like '{qualified_table_name[2]}' in database {qualified_table_name[0]}"
-                    )
-                else:
-                    # "database.schema.table", schema is not empty
-                    tables = self._run_query(
-                        f"show tables like '{qualified_table_name[2]}' in database {qualified_table_name[0]}"
-                    )
-                    # table is a tuple and the 3rd col(0 based idx) is schema_name
-                    return any(
-                        [
-                            table
-                            for table in tables
-                            if table[3].lower() == qualified_table_name[1].lower()
-                        ]
-                    )
-            else:
-                # we do not support len(qualified_table_name) > 3 for now
-                raise SnowparkClientExceptionMessages.GENERAL_INVALID_OBJECT_NAME(
-                    table_name
-                )
-
-            return tables is not None and len(tables) > 0
-        except ProgrammingError as exc:
-            if exc.errno == 2043:
-                # SQL compilation error: Object does not exist, or operation cannot be performed
-                return False
-            _logger.debug(
-                f"Error occurred when checking table existence for {table_name}, error: {exc!r}"
+        qualified_table_name = table_name.split(".")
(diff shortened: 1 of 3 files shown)


tests/integ/test_session.py
```diff
@@ -210,6 +211,43 @@ def test_table_exists(session):
...
+    session.sql(f'create temp table "{table_name}"(col_a varchar)').collect()
+    assert session._table_exists(qualified_table_name) is True
+
+    # negative cases
+    with pytest.raises(SnowparkClientException):
+        # invalid qualified name
+        session._table_exists("a.b.c.d")
+
+    random_database = Utils.random_temp_database()
+    random_schema = Utils.random_temp_database()
```
sfc-gh-aling, line 243:
```suggestion
    random_schema = Utils.random_temp_schema()
```

src/snowflake/snowpark/session.py
```diff
@@ -1658,9 +1658,53 @@ def query_history(self) -> QueryHistory:
...
+                            for table in tables
+                            if table[3].lower() == qualified_table_name[1].lower()
+                        ]
+                    )
+
+            return tables is not None and len(tables) > 0
+        except ProgrammingError as exc:
+            if exc.errno == 2043:
+                # SQL compilation error: Object does not exist, or operation cannot be performed
+                return False
```
sfc-gh-sfan, line 1703: When does this happen? Is it because the database/schema does not exist?

It would be nice to make 2043 a static constant. Hopefully the same as the variable name in GS (if there is one).

sfc-gh-aling: yes, this happens when database/schema does not exist.

actually after reconsidering the case, I feel this case we shall just raise the error instead of returning false..
returning `False` is kinda misleading because it's telling users that the `table` doesn't exist, not the schema or database.

If such error happens, users should go and check their databse/schema.

thoughts?

src/snowflake/snowpark/session.py
```diff
@@ -1658,9 +1658,53 @@ def query_history(self) -> QueryHistory:
         return query_listener
 
     def _table_exists(self, table_name: str):
-        # TODO: Support qualified table names
-        tables = self._run_query(f"show tables like '{table_name}'")
-        return tables is not None and len(tables) > 0
+        # implementation based upon: https://docs.snowflake.com/en/sql-reference/name-resolution.html
+        validate_object_name(table_name)
+        try:
+            qualified_table_name = table_name.split(".")
```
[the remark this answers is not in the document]
sfc-gh-aling: wow, thanks for the knowledge! I will add a comment to this part so that we have context in the future if we want to make the change.

src/snowflake/snowpark/session.py
```diff
@@ -1658,9 +1658,53 @@ def query_history(self) -> QueryHistory:
...
+                # name in the form of "table"
+                tables = self._run_query(f"show tables like '{table_name}'")
+            elif len(qualified_table_name) == 2:
+                # name in the form of "schema.table" omitting database
+                # schema: qualified_table_name[0]
+                # table: qualified_table_name[1]
+                tables = self._run_query(
+                    f"show tables like '{qualified_table_name[1]}' in schema {qualified_table_name[0]}"
+                )
+            else:  # len(qualified_table_name) == 3, there is no other case, validate_object_name checked for us
```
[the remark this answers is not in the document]
sfc-gh-aling: updated to throw if len > 3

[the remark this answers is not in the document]
sfc-gh-aling: updated to throw if len > 3

src/snowflake/snowpark/session.py
```diff
@@ -1658,9 +1658,61 @@ def query_history(self) -> QueryHistory:
...
+                # table: qualified_table_name[2]
+                if qualified_table_name[1] == "":
+                    # Double-Dot Notation: "database..table" in which the schema is empty
+                    tables = self._run_query(
+                        f"show tables like '{qualified_table_name[2]}' in database {qualified_table_name[0]}"
+                    )
+                else:
+                    # "database.schema.table", schema is not empty
+                    tables = self._run_query(
+                        f"show tables like '{qualified_table_name[2]}' in database {qualified_table_name[0]}"
```
sfc-gh-jdu, line 1691: I think `show tables in schema database.schema` will also work, which is neater?

src/snowflake/snowpark/session.py
sfc-gh-jdu: I also think we should just raise an exception here

sfc-gh-sfan: When a database/schema does not exist, I think the table creation should fail, right? If that's the case, then it seems legit to just raise an exception because the following table creation will fail on the server anyways.

sfc-gh-sfan (approved):

src/snowflake/snowpark/session.py
sfc-gh-aling: it works! thanks for the suggestion!

src/snowflake/snowpark/session.py
sfc-gh-aling: yes, if no db/schema, then we would get schema/db not exist error.

PR #475, opened 2022-09-08
[SNOW-647398] show details when printing a snowpark session

sfc-gh-stan:

   Fixes [SNOW-647398](https://snowflakecomputing.atlassian.net/browse/SNOW-647398)

It implements a `__repr__` method for Snowpark.Session so `print(session)` will display some helpful information about the session.
Before: `<snowflake.snowpark.session.Session object at 0x11c681700>`
After: `<snowflake.snowpark.session.Session: role='ROLE', database='DB', schema='SCHEMA', warehouse='WH'>`

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -265,8 +265,8 @@ class Session:
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
-    def __repr__(self):
-        return f"<{self.__class__.__module__}.{self.__class__.__name__}: account={repr(self._conn._conn.account)}, role={repr(self._conn._conn.role)}, database={repr(self._conn._conn.database)}, schema={repr(self._conn._conn.schema)}, warehouse={repr(self._conn._conn.warehouse)}>"
+    def __str__(self):
+        return f"<{self.__class__.__module__}.{self.__class__.__name__}: account={self.get_current_account()}, role={self.get_current_role()}, database={self.get_current_database()}, schema={self.get_current_schema()}, warehouse={self.get_current_warehouse()}>"
 
     def _generate_new_action_id(self) -> int:
         self._last_action_id += 1
@@ -1420,6 +1420,13 @@ class Session:
         df = DataFrame(self, range_plan)
         set_api_call_source(df, "Session.range")
         return df
+
+    def get_current_account(self) -> Optional[str]:
+        """
+        Returns the name of the current account for the Python connector session attached
+        to this session.
+        """
+        return self._conn._get_current_parameter("account")
 
     def get_current_database(self) -> Optional[str]:
         """
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -266,7 +266,7 @@ class Session:
         self.close()
 
     def __repr__(self):
-        return f"<{self.__class__.__module__}.{self.__class__.__name__}: role={repr(self._conn._conn.role)}, database={repr(self._conn._conn.database)}, schema={repr(self._conn._conn.schema)}, warehouse={repr(self._conn._conn.warehouse)}>"
+        return f"<{self.__class__.__module__}.{self.__class__.__name__}: account={repr(self._conn._conn.account)}, role={repr(self._conn._conn.role)}, database={repr(self._conn._conn.database)}, schema={repr(self._conn._conn.schema)}, warehouse={repr(self._conn._conn.warehouse)}>"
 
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -265,6 +265,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
+    def __repr__(self):
+        return f"<{self.__class__.__module__}.{self.__class__.__name__}: role={repr(self._conn._conn.role)}, database={repr(self._conn._conn.database)}, schema={repr(self._conn._conn.schema)}, warehouse={repr(self._conn._conn.warehouse)}>"
```
sfc-gh-kbregula, line 269: Should we also add an account?

sfc-gh-stan: Makes sense to me, I will add account in front of role.

src/snowflake/snowpark/session.py
```diff
@@ -265,6 +265,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
+    def __repr__(self):
```
sfc-gh-jdu, line 268: I will prefer to implement `__str__` instead of `__repr__`. If `__repr__` gets implemented, we have to make sure `eval(repr(x)) == x`. https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr

src/snowflake/snowpark/session.py
```diff
@@ -265,6 +265,9 @@ def __enter__(self):
...
+        return f"<{self.__class__.__module__}.{self.__class__.__name__}: account={repr(self._conn._conn.account)}, role={repr(self._conn._conn.role)}, database={repr(self._conn._conn.database)}, schema={repr(self._conn._conn.schema)}, warehouse={repr(self._conn._conn.warehouse)}>"
```
sfc-gh-jdu, line 269: 1. `self._conn._conn.role` and other similar properties can be None. You should use `get_current_*` methods in session class
2. I don't think you need `repr(x)` then.
3. Add tests when these properties are None (`get_current_*` methods  can also return None).

sfc-gh-kbregula: When can roles and other properties be empty? It seems to me that the `snowflake-connector-python` maintains the contents of these variables while logging in and executing SQL queries.
https://github.com/snowflakedb/snowflake-connector-python/blob/d0efd0f8333db210a3c482b6f631fe4ba01eabdc/src/snowflake/connector/connection.py#L1013-L1026

sfc-gh-stan: 1+2. I will use change the implementation to use `get_current_*` for better consistency. 
3. I agree with @sfc-gh-kbregula that for our use case, `get_current_*` will always short circuit at `getattr(self._conn, param) ` and return default values that the Python connector connection store.

sfc-gh-jdu: When we create a session in Snowflake, database/schema/warehouse is optional. If we don't set them in connection parameters of python connector, then these properties will be empty.

sfc-gh-jdu: nope, it's possible that the user doesn't have a default database/schema/warehouse. (For example, if you log into snowflake at the first time and don't create them before, how can we have defaults?)

sfc-gh-stan: I see, that makes sense 👍

sfc-gh-jdu (approved):

src/snowflake/snowpark/session.py
```diff
@@ -265,6 +265,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
+    def __str__(self):
+        return f"<{self.__class__.__module__}.{self.__class__.__name__}: account={self.get_current_account()}, role={self.get_current_role()}, database={self.get_current_database()}, schema={self.get_current_schema()}, warehouse={self.get_current_warehouse()}>"
```
sfc-gh-sfan, line 269: Just curious if we could line break this in code? e.g.

```
return (f"<{self.__class__.__module__}.{self.__class__.__name__}: "
        f"account={self.get_current_account()}, "
        ...)

```

sfc-gh-sfan (approved):

src/snowflake/snowpark/session.py
sfc-gh-stan: True, that looks a lot better, I wonder why flake8 doesn't catch this.

PR #477, opened 2022-09-08
SNOW-624881: Wrap df.join() result with the flattenable SelectStatement

sfc-gh-yixie:

   Fixes SNOW-624881

   Warp dataframe.join result with SelectStatement so the sql simplifier can continue to flatten the SQLs after df.join().
But df.join itself doesn't flatten the SQL yet.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -268,6 +268,7 @@ class SelectSnowflakePlan(Selectable):
             if isinstance(snowflake_plan, SnowflakePlan)
             else analyzer.resolve(snowflake_plan)
         )
+        self.expr_to_alias.update(self._snowflake_plan.expr_to_alias)
         self.pre_actions = self._snowflake_plan.queries[:-1]
         self.post_actions = self._snowflake_plan.post_actions
 
@@ -430,6 +431,7 @@ class SelectStatement(Selectable):
             len(cols) == 1
             and isinstance(cols[0], UnresolvedAlias)
             and isinstance(cols[0].child, Star)
+            and not cols[0].child.expressions
         ):
             return self
         final_projection = []
@@ -442,6 +444,7 @@ class SelectStatement(Selectable):
             # There must be duplicate columns in the projection.
             # We don't flatten when there are duplicate columns.
             can_be_flattened = False
+            disable_next_level_flatten = True
         elif self.flatten_disabled or self.has_clause_using_columns:
             can_be_flattened = False
         else:
@@ -737,6 +740,7 @@ def initiate_column_states(column_names: List[str]) -> ColumnStateDict:
             referenced_by_same_level_columns=COLUMN_DEPENDENCY_EMPTY,
             state_dict=column_states,
         )
+    column_states.projection = column_names
     return column_states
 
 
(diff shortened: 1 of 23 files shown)


tests/integ/test_simplifier_suite.py
```diff
@@ -501,3 +501,41 @@ def test_use_sql_simplifier(session, simplifier_table):
...
+        == f'SELECT "B" FROM ( SELECT *, "A" FROM ( SELECT  *  FROM {simplifier_table}))'
+    )
+
+    with pytest.raises(SnowparkSQLException, match="ambiguous column name 'A'"):
+        df3.select("a").collect()
+
+    with pytest.raises(SnowparkSQLException, match="ambiguous column name 'A'"):
+        df4.select("a").collect()
+
+    # tests/integ/scala/test_dataframe_join_suite.py has tests of join and select *. We don't repeat those tests here.
```
sfc-gh-aalam, line 541: are there test to verify the join simplifier also flattens for example joining two tables and making sure dropped columns are not flattened and multiple `.select('a').select('a')` is flattened?

sfc-gh-sfan (changes requested):
If this PR does two things...let's do them in two PRs?

tests/integ/test_simplifier_suite.py
sfc-gh-yixie: Good point. Added the test

sfc-gh-yixie:
> If this PR does two things...let's do them in two PRs?

They're actually related. join + select star is a more complex version of select star. 
`test_dataframe_join_suite.py` has these tests related to both join and select star.
- `test_select_all_on_join_result()`
- `test_select_left_right_combination_on_join_result()`
- `test_select_columns_on_join_result_with_conflict_name()`

Refer to https://github.com/snowflakedb/snowpark-python/blob/19f1dc924feced4958a72764bfe0d7231f027cd8/tests/integ/scala/test_dataframe_join_suite.py#L968

sfc-gh-sfan:
I personally would prefer `select *` being fixed first before introducing another change. As a rule of thumb I think a PR title should not have two verbs, which is a strong hint that something can be broken down.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -436,6 +437,7 @@ def select(self, cols: List[Expression]) -> "SelectStatement":
             len(cols) == 1
             and isinstance(cols[0], UnresolvedAlias)
             and isinstance(cols[0].child, Star)
+            and not cols[0].child.expressions
```
sfc-gh-jdu, line 440: when will the child has expressions?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -810,14 +814,14 @@ def derive_column_states_from_subquery(
     column_states = ColumnStateDict()
     for c in cols:
         if isinstance(c, UnresolvedAlias) and isinstance(c.child, Star):
-            column_states.projection.extend(from_.column_states.projection)
+            if c.child.expressions:
+                columns_from_star = map(analyzer.analyze, c.child.expressions)
```
sfc-gh-jdu, line 818: If it has children, then we don't need to consider from_.column_states?

sfc-gh-jdu (commented):
I also feel it should be split to 2 PRs, and I'm a little bit confused about how you fix select *. Could you add more comments?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yixie: A simple example is that `df.select(df["*"])` has child expressions, which are attribute list of all columns from `df`.
When joining two dataframes, you can do the following. 
```python
    df_left = session.create_dataframe([[1, 2]]).to_df("a", "b")
    df_right = session.create_dataframe([[3, 4]]).to_df("c", "d")
    df = df_left.join(df_right)
    df = df.select(df_right["*"], df_left["*"])  # the two stars have child expressions.
    # the final df has columns C D B A instead of A B C D
```
`df.select("*")` doesn't have the child expressions.

sfc-gh-yixie:
> I personally would prefer `select *` being fixed first before introducing another change. As a rule of thumb I think a PR title should not have two verbs, which is a strong hint that something can be broken down.

@sfc-gh-jdu @sfc-gh-sfan Please review #485 that fixes `df.select(df["*"])`.

sfc-gh-jdu (approved):

sfc-gh-sfan (approved):
LGTM

PR #520, opened 2022-09-23
SNOW-658225 add telemetry to udf, udtf and stored procs

sfc-gh-aalam:

   Fixes #SNOW-658225

   Adding telemetry message on creation and invocation of sprocs, udfs and udtfs.

### DIFF
--- a/src/snowflake/snowpark/stored_procedure.py
+++ b/src/snowflake/snowpark/stored_procedure.py
@@ -62,7 +62,7 @@ class StoredProcedure:
         self._input_types = input_types
         self._execute_as = execute_as
 
-        api_call_source = api_call_source or "sproc_create"
+        api_call_source = api_call_source or "StoredProcedure.__init__"
         session = session or snowflake.snowpark.session._get_active_session()
         session._conn._telemetry_client.send_function_usage_telemetry(
             api_call_source, TelemetryField.FUNC_CAT_CREATE.value
@@ -80,7 +80,7 @@ class StoredProcedure:
             )
 
         session._conn._telemetry_client.send_function_usage_telemetry(
-            "sproc_invoke", TelemetryField.FUNC_CAT_USAGE.value
+            "StoredProcedure.__call__", TelemetryField.FUNC_CAT_USAGE.value
         )
         return session.call(self.name, *args)
 
--- a/src/snowflake/snowpark/stored_procedure.py
+++ b/src/snowflake/snowpark/stored_procedure.py
@@ -11,6 +11,7 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 import snowflake.snowpark
 from snowflake.connector import ProgrammingError
 from snowflake.snowpark._internal.error_message import SnowparkClientExceptionMessages
+from snowflake.snowpark._internal.telemetry import TelemetryField
 from snowflake.snowpark._internal.type_utils import convert_sp_to_sf_type
 from snowflake.snowpark._internal.udf_utils import (
     UDFColumn,
@@ -50,6 +51,7 @@ class StoredProcedure:
         name: str,
         execute_as: typing.Literal["caller", "owner"] = "owner",
         session: Optional["snowflake.snowpark.session.Session"] = None,
+        api_call_source: Optional[str] = None,
     ) -> None:
         #: The Python function.
         self.func: Callable = func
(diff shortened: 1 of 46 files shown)


src/snowflake/snowpark/stored_procedure.py
```diff
@@ -59,6 +62,12 @@ def __init__(
         self._input_types = input_types
         self._execute_as = execute_as
 
+        api_call_source = api_call_source or "StoredProcedure.__init__"
+        session = session or snowflake.snowpark.session._get_active_session()
+        session._conn._telemetry_client.send_function_usage_telemetry(
+            api_call_source, TelemetryField.FUNC_CAT_CREATE.value
```
sfc-gh-sfan, line 68: This might not be correct. Creation of this class does not mean a stored proc is created.

The actually sp creation is in `StoredProcedureRegistration`. I'm wondering if we should add this `api_call_source` to `udf_utils#create_python_udf_or_sp`? This way we could get both udf and sp covered, though of course we need to pass the `api_call_source`.

src/snowflake/snowpark/udf.py
```diff
@@ -63,6 +66,12 @@ def __init__(
         self._input_types = input_types
         self._is_return_nullable = is_return_nullable
 
+        api_call_source = api_call_source or "UserDefinedFunction.__init__"
+        session = session or snowflake.snowpark.session._get_active_session()
+        session._conn._telemetry_client.send_function_usage_telemetry(
+            api_call_source, TelemetryField.FUNC_CAT_CREATE.value
```
sfc-gh-sfan, line 72: Similarly, we should add telemetry to the actual creation of UDF.

sfc-gh-aalam: is `__call__` the right place to send the usage telemetry API?

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -69,6 +70,10 @@ def __call__(
             raise ValueError(
                 f"Incorrect number of arguments passed to the stored procedure. Expected: {len(self._input_types)}, Found: {len(args)}"
             )
+
+        session._conn._telemetry_client.send_function_usage_telemetry(
+            "StoredProcedure.__call__", TelemetryField.FUNC_CAT_USAGE.value
```
sfc-gh-jdu, line 75: call this sp or udf will also under `TelemetryField.FUNC_CAT_USAGE`?

src/snowflake/snowpark/udf.py
```diff
@@ -550,9 +557,11 @@ def register(
             replace,
             parallel,
             max_batch_size,
-            kwargs.get("_from_pandas_udf_function", False),
+            _from_pandas,
             statement_params=statement_params,
             source_code_display=source_code_display,
+            api_call_source="UDFRegistration.register"
+            + ("[pandas]" if _from_pandas else ""),
```
sfc-gh-jdu, line 564:
```suggestion
            + ("[pandas_udf]" if _from_pandas else ""),
```

src/snowflake/snowpark/stored_procedure.py
sfc-gh-aalam: It will be. Would you recommend a more appropriate category (or a new category?). I am using `create` for udf/sp creation and `usage` for invocations.

src/snowflake/snowpark/udf.py
sfc-gh-jdu: yes

src/snowflake/snowpark/stored_procedure.py
sfc-gh-jdu: yea, once a udf is created, it can be called many times. So we should separate them into two different categories

sfc-gh-aalam: Sorry, I don't understand. Is the current implementation ok, or should it be changed to show multiple invocations separately. Currently, create and invoke send different telemetry messages

sfc-gh-jdu: ah I see. I'm not sure how important the category is in telemetry? Because the category is "create", though the message is different, the actual udf invocation has nothing to do with "create"?

sfc-gh-aalam: that's right, invocation has nothing to do with create. We only call create once the stored procedure is registered

sfc-gh-jdu (approved):

src/snowflake/snowpark/udf.py
```diff
@@ -85,6 +86,10 @@ def __call__(
                     f"The input of UDF {self.name} must be Column, column name, or a list of them"
                 )
 
+        session = snowflake.snowpark.context.get_active_session()
```
sfc-gh-yixie, line 89: @sfc-gh-aalam This line may raise an exception when there are multiple active sessions.

PR #521, opened 2022-09-23
SNOW-665404: Improve async job related APIs' annotations

### LINKED ISSUE #505: SNOW-665294: Pandas DataFrame types not working in VS Code
It looks like the recent change of return types here: 

```
 def to_pandas(
        self,
        *,
        block: bool = True,
        **kwargs: Dict[str, Any],
    ) -> Union["pandas.DataFrame", AsyncJob]:
```

https://github.com/snowflakedb/snowpark-python/blob/96e96a0fb431a32278d8a9463cd4ba0f1c28b91d/src/snowflake/snowpark/dataframe.py#L608 

causes some issues with VS Code autocomplete which sees this as returning (Any | AsyncJob) which turned my code into a lot of red squiggly lines.

I imagine the double quote string literal of type here is the issue, and previously I expect it was assuming it was type Any so wouldn't auto-complete, but wouldn't red squiggle. The code actually runs fine, but editors are confused by the type being returned.

sfc-gh-jdu:

   Fixes #505 

   Now you can get autocomplete after to_pandas()! Also, using @overload can improve the API doc readability:
image

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -18,10 +18,11 @@ from typing import (
     Optional,
     Tuple,
     Union,
+    overload,
 )
 
 import snowflake.snowpark
-from snowflake.connector.options import pandas
+from snowflake.connector.options import installed_pandas
 from snowflake.snowpark._internal.analyzer.analyzer_utils import quote_name
 from snowflake.snowpark._internal.analyzer.binary_plan_node import (
     Cross,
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -487,10 +488,22 @@ def __init__(
     def stat(self) -> DataFrameStatFunctions:
         return self._stat
 
+    @overload
+    def collect(
+        self, *, statement_params: Optional[Dict[str, str]] = None, block: bool = True
```
sfc-gh-sfan, line 493: I'm curious, given that the default value of `block` is different though the function name is the same, does this mean `block` effectively becomes a required argument?

sfc-gh-sfan (approved):

sfc-gh-yixie:
Did you try with jupyter notebook and VS code?

sfc-gh-jdu:
> Did you try with jupyter notebook and VS code?

Yes

src/snowflake/snowpark/dataframe.py
```diff
@@ -611,6 +636,29 @@ def to_local_iterator(
     def __copy__(self) -> "DataFrame":
         return DataFrame(self._session, copy.copy(self._plan))
 
+    if installed_pandas:
```
sfc-gh-yixie, line 639: What will happen to the IDE/notebook when pandas is not installed?

sfc-gh-jdu: It's the same as before. Again @overload is for docstring and type checking tools so this PR doesn't have any functional change. When pandas is not installed, only this definition will be skipped, and users are not able to get autocomplete for pandas methods.

sfc-gh-yixie (approved):

PR #528, opened 2022-09-27
SNOW-623555: Allow ntile to accept int value

sfc-gh-jdu:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -2814,12 +2814,15 @@ def lead(
     )
 
 
-def ntile(e: ColumnOrName) -> Column:
+def ntile(e: Union[int, ColumnOrName]) -> Column:
     """
     Divides an ordered data set equally into the number of buckets specified by n.
     Buckets are sequentially numbered 1 through n.
+
+    Args:
+        e: The desired number of buckets; must be a positive integer value.
     """
-    c = _to_col_if_str(e, "ntile")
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/functions.py
```diff
@@ -2814,12 +2814,15 @@ def lead(
     )
 
 
-def ntile(e: ColumnOrName) -> Column:
+def ntile(e: Union[int, ColumnOrName]) -> Column:
```
sfc-gh-jdu, line 2817: I just add `int` to avoid breaking change. Should we deprecate using column? @snowflakedb/snowpark-python-api

sfc-gh-sfan: Yeah we probably should. Let's add a warning first? Then in two releases we could use a BCR to remove it.

sfc-gh-sfan (approved):
LGTM

src/snowflake/snowpark/functions.py
sfc-gh-yixie: It's a sql documentation problem since sql function ntile actually works when a column expression is used.
So I suggest we shouldn't deprecate it.

sfc-gh-yixie (approved):

src/snowflake/snowpark/functions.py
sfc-gh-sfan: Oh I'm not aware of that. If column works then we should just keep the current way.

PR #536, opened 2022-09-30
SNOW-664571: Add "on" as an alias for "using_columns" and "how" as an alias for "join_type" in DataFrame.join

sfc-gh-jdu:

   Fixes #NNNN

   To align with [pyspark](https://spark.apache.org/docs/3.1.2/api/python/reference/api/pyspark.sql.DataFrame.join.html) and [pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.join.html).

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -1813,10 +1813,23 @@ class DataFrame:
 
     @df_api_usage
     def natural_join(
-        self, right: "DataFrame", join_type: Optional[str] = None
+        self, right: "DataFrame", how: Optional[str] = None, **kwargs
     ) -> "DataFrame":
-        """Performs a natural join of the specified type (``joinType``) with the
+        """Performs a natural join of the specified type (``how``) with the
         current DataFrame and another DataFrame (``right``).
+
+        Args:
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -1841,16 +1854,12 @@ def natural_join(
             |5    |6    |NULL  |
             --------------------
             <BLANKLINE>
-
-        Args:
-            right: the other :class:`DataFrame` to join
-            join_type: The type of join ("inner", "full", "left", "right"). The default value is "left".
         """
-        join_type = join_type or "inner"
+        join_type = kwargs.get("join_type") or how
```
sfc-gh-sfan, line 1858: Do we plan to eventually deprecate `join_type`?

sfc-gh-sfan (approved):
LGTM. This does not seem like a behavior change because both keyword arg and positional arg are not impacted.

src/snowflake/snowpark/dataframe.py
sfc-gh-jdu: I don't think we need to do it currently, because join_type also makes sense. And we just use `how` to align with pyspark and pandas

PR #546, opened 2022-10-05
SNOW-664210: Fix create_dataframe when schema has special characters

sfc-gh-yixie:

   Fixes #SNOW-664210

   Use `SELECT $1 AS "a b" FROM ( VALUES (1 :: INT))` instead of `SELECT  *  FROM ( VALUES (1 :: INT) AS SNOWPARK_TEMP_TABLE_EVQQHTZLRR("a b"))`. The latter has column `"""a b"""` instead of `"a b"`. The former has column `"a b"`.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -452,10 +452,8 @@ def values_statement(output: List[Attribute], data: List[Row]) -> str:
         SELECT
         + COMMA.join([f"{DOLLAR}{i+1}{AS}{c}" for i, c in enumerate(names)])
         + FROM
-        + LEFT_PARENTHESIS
         + VALUES
         + COMMA.join(rows)
-        + RIGHT_PARENTHESIS
     )
     return query
 
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -438,7 +438,6 @@ def range_statement(start: int, end: int, step: int, column_name: str) -> str:
 
 
 def values_statement(output: List[Attribute], data: List[Row]) -> str:
-    table_name = random_name_for_temp_object(TempObjectType.TABLE)
     data_types = [attr.datatype for attr in output]
     names = [quote_name(attr.name) for attr in output]
     rows = []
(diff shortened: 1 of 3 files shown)


tests/integ/scala/test_session_suite.py
```diff
@@ -96,6 +96,12 @@ class P1(NamedTuple):
     assert [field.name for field in df.schema.fields] == ["A", "B", "C"]
 
 
+def test_create_dataframe_special_char_column_name(session):
```
sfc-gh-jdu, line 99: Can we also add a test case for other special chars?

tests/integ/scala/test_session_suite.py
```diff
@@ -96,6 +96,12 @@ class P1(NamedTuple):
     assert [field.name for field in df.schema.fields] == ["A", "B", "C"]
```
sfc-gh-jdu, line 98: Can we also have a test case for the second issue described here https://snowflake.slack.com/archives/C02EMSN23HD/p1663287588067719? (also let's move these tests to test_dataframe.py? I usually created `create_dataframe` related tests there)

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -448,16 +447,15 @@ def values_statement(output: List[Attribute], data: List[Row]) -> str:
...
+    query = (
+        SELECT
+        + COMMA.join([f"{DOLLAR}{i+1}{AS}{c}" for i, c in enumerate(names)])
+        + FROM
+        + VALUES
         + COMMA.join(rows)
-        + AS
-        + table_name
-        + LEFT_PARENTHESIS
-        + COMMA.join(names)
```
sfc-gh-jdu, line 457: I'm curious why we chose the previous way.

sfc-gh-sfan (approved):
LGTM. Please address @sfc-gh-jdu 's comment about testing.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-yixie: Not sure if it's a randomized choice, or a purposeful one. The [VALUES documentation](https://docs.snowflake.com/en/sql-reference/constructs/values.html) doesn't say they have difference except that the one we chose to use can have alias.
I think the column aliases have a bug for quotes but it's hard to change (would-be break change).

tests/integ/scala/test_session_suite.py
sfc-gh-yixie: Added tests for other special chars and reusing the column names with double quotes.
Both problems are because of the wrong actions on double quotes.

sfc-gh-jdu (approved):

sfc-gh-yixie:
@sfc-gh-bli @sfc-gh-mrui FYI, we fixed a bug about `session.create_data_frame` that doesn't accept special chars in schema.

PR #549, opened 2022-10-06
SNOW-654710: Add drop_table() to class Table

sfc-gh-yixie:

   Fixes #SNOW-654710

   Instead of purging the cached table automatically, we added an API and context manager to drop the table manually.

### DIFF
--- a/src/snowflake/snowpark/table.py
+++ b/src/snowflake/snowpark/table.py
@@ -668,4 +668,6 @@ class Table(DataFrame):
         Note that subsequent operations such as :meth:`DataFrame.select`, :meth:`DataFrame.collect` on this ``Table`` instance and the derived DataFrame will raise errors because the underlying
         table in the Snowflake database no longer exists.
         """
-        self._session.sql(f"drop table {self.table_name}").collect()
+        self._session.sql(
+            f"drop table {self.table_name}"
+        )._internal_collect_with_tag_no_telemetry()
--- a/src/snowflake/snowpark/table.py
+++ b/src/snowflake/snowpark/table.py
@@ -668,6 +668,4 @@ class Table(DataFrame):
         Note that subsequent operations such as :meth:`DataFrame.select`, :meth:`DataFrame.collect` on this ``Table`` instance and the derived DataFrame will raise errors because the underlying
         table in the Snowflake database no longer exists.
         """
-        self._session.sql(
-            f"drop table if exists {self.table_name}"
-        )._internal_collect_with_tag_no_telemetry()
+        self._session.sql(f"drop table {self.table_name}").collect()
--- a/src/snowflake/snowpark/table.py
+++ b/src/snowflake/snowpark/table.py
@@ -274,6 +274,12 @@ class Table(DataFrame):
     def __copy__(self) -> "Table":
         return Table(self.table_name, self._session)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.drop_table()
+
     def sample(
         self,
         frac: Optional[float] = None,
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/table.py
```diff
@@ -655,3 +661,11 @@ def merge(
             if block
             else result
         )
+
+    def drop_table(self):
```
sfc-gh-jdu, line 665:
```suggestion
    def drop_table(self) -> None:
```

src/snowflake/snowpark/table.py
```diff
@@ -655,3 +661,11 @@ def merge(
...
+        """Drops the table from the Snowflake database.
+
+        Note that subsequent API such as ``select()``, ``collect()`` on this ``Table`` instance and the derived DataFrame will raise errors because the underlying
```
sfc-gh-jdu, line 668:
```suggestion
        Note that subsequent operations such as :meth:`DataFrame.select`, :meth:`DataFrame.collect` on this ``Table`` instance and the derived DataFrame will raise errors because the underlying
```

src/snowflake/snowpark/table.py
```diff
@@ -655,3 +661,11 @@ def merge(
...
+        table in the Snowflake database no longer exists.
+        """
+        self._session.sql(f"drop table {self.table_name}").collect()
```
sfc-gh-jdu, line 671: I suggest to use `self._session._run_query`, which are usually used to run some internal queries in our codebase.

src/snowflake/snowpark/table.py
```diff
@@ -274,6 +274,12 @@ def __init__(
     def __copy__(self) -> "Table":
         return Table(self.table_name, self._session)
 
+    def __enter__(self):
```
sfc-gh-jdu, line 277: since we add them, let's have an example in docstring of `Table` class?

src/snowflake/snowpark/table.py
sfc-gh-yixie: I remember `_run_query` still throws a ProgrammingError?
We better change this first then update here. Now `session.sql` raises the correct error.

src/snowflake/snowpark/table.py
sfc-gh-yixie: Added text and samples in  the `cache_result` docstring because it's the main use case.

src/snowflake/snowpark/table.py
```diff
@@ -655,3 +661,11 @@ def merge(
             if block
             else result
         )
+
+    def drop_table(self) -> None:
```
sfc-gh-stan, line 665: What about adding an optional parameter `if_exist`?

src/snowflake/snowpark/table.py
sfc-gh-jdu: Yea you're right. And here you might want to use internal_collect() (and for all other internal session.sql...)

sfc-gh-yixie: I changed to `_internal_collect_with_tag_no_telemetry`

src/snowflake/snowpark/table.py
sfc-gh-yixie: Very good suggestion. Updated.

sfc-gh-sfan (approved):
LGTM

src/snowflake/snowpark/table.py
sfc-gh-stan: I was thinking of having an optional parameter in the method `drop_table` and if that is set to 
 true, we append `if exists` to the end of the sql. Maybe the user wants an operation to fail if the table has been dropped/not created properly before.

sfc-gh-jdu (approved):

src/snowflake/snowpark/table.py
sfc-gh-yixie: As offline discussion, we'll not add the param and add it later when we think about other Snowflake objects.

PR #557, opened 2022-10-11
SNOW-664934 Support schema for pandas DF in create_dataframe()

### LINKED ISSUE #504: SNOW-664934: Dataframe schemas don't persist between Pandas and Snowpark

Python 3.8

Snowflake Python UDF

Only snowflake-snowpark-python

4. What did you do?

Snowpark has a problem / bug that it does not maintain the types between Pandas and Snowpark, nor does it allow to manually set its schema.

For instance,
```py
df1 = session.sql(sql).to_pandas()
df2 = session.create_dataframe(df)
```

The timestamp field on `df1` with `TimestampType` has become a `LongType`.

I've also tried to store the schema and use it, but same results.

```py
df1 = session.sql(sql)
df1_schema = df1.schema
df1 = df1.to_pandas()
df2 = session.create_dataframe(df, df1_schema)
```

This stops me from being able to write the `DataFrame` back to the table as it needs to be of `TimestampType` rather than `LongType`.

The schemas persist among the conversions.

sfc-gh-sfan:
Description

Prior to this change, schema is ignored if the input data is a pandas DF in create_dataframe(). This can be confusing as user would expect schema to take effect.

Note that this is a behavior change.

Testing

integ test

   Fixes #504 

   Prior to this change, schema is ignored if the input data is a pandas DF in create_dataframe(). This can be confusing as user would expect schema to take effect.

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -22,6 +22,7 @@ from snowflake.connector.options import installed_pandas, pandas
 from snowflake.connector.pandas_tools import write_pandas
 from snowflake.snowpark._internal.analyzer.analyzer import Analyzer
 from snowflake.snowpark._internal.analyzer.analyzer_utils import (
+    attribute_to_schema_string,
     escape_quotes,
     quote_name,
 )
@@ -1217,7 +1218,8 @@ class Session:
                 be a :class:`list`, :class:`tuple` or pandas DataFrame. Every element in
                 ``data`` will constitute a row in the DataFrame.
             schema: A :class:`~snowflake.snowpark.types.StructType` containing names and
-                data types of columns, or a list of column names, or ``None``.
+                data types of columns, or a list of column names, or ``None``. If ``data``
+                is pandas DataFrame, ``schema`` has to be ``StructType`` if specified.
                 When ``schema`` is a list of column names or ``None``, the schema of the
                 DataFrame will be inferred from the data across all rows. To improve
                 performance, provide a schema. This avoids the need to infer data types
@@ -1274,14 +1276,24 @@ class Session:
             sf_database = self._conn._get_current_parameter("database", quoted=False)
             sf_schema = self._conn._get_current_parameter("schema", quoted=False)
 
+            if schema:
+                if not isinstance(schema, StructType):
+                    raise ValueError(
+                        f"Expects StructType for schema if specified when creating snowpark dataframe from pandas dataframe. Got {type(schema)}"
+                    )
+
+                self._run_query(
+                    f"create or replace temporary table {table_name} ({attribute_to_schema_string(schema._to_attributes())})"
+                )
+
             t = self.write_pandas(
                 data,
                 table_name,
                 database=sf_database,
                 schema=sf_schema,
                 quote_identifiers=True,
-                auto_create_table=True,
-                create_temp_table=True,
+                auto_create_table=schema is None,
+                table_type="temporary",
             )
             set_api_call_source(t, "Session.create_dataframe[pandas]")
             return t
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -1290,20 +1290,27 @@ class Session:
                     f"create or replace temporary table {table_name} ({attribute_to_schema_string(schema._to_attributes())})"
                 )
 
-            try:
+                try:
+                    t = self.write_pandas(
+                        data,
+                        table_name,
+                        database=sf_database,
+                        schema=sf_schema,
+                        quote_identifiers=True,
+                    )
+                except Exception as e:
+                    self._run_query(f"drop table if exists {table_name}")
+                    raise e
+            else:
                 t = self.write_pandas(
                     data,
                     table_name,
                     database=sf_database,
                     schema=sf_schema,
                     quote_identifiers=True,
-                    auto_create_table=schema is None,
-                    # auto_create_table=True,
+                    auto_create_table=True,
                     table_type="temporary",
                 )
-            except Exception as e:
-                self._run_query(f"drop table if exists {table_name}")
-                raise e
 
             set_api_call_source(t, "Session.create_dataframe[pandas]")
             return t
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -1217,7 +1218,8 @@ def create_dataframe(
                 be a :class:`list`, :class:`tuple` or pandas DataFrame. Every element in
                 ``data`` will constitute a row in the DataFrame.
             schema: A :class:`~snowflake.snowpark.types.StructType` containing names and
-                data types of columns, or a list of column names, or ``None``.
+                data types of columns, or a list of column names, or ``None``. If ``data``
+                is pandas DataFrame, ``schema`` has to be ``StructType`` if specified.
```
sfc-gh-jdu, line 1222:
```suggestion
                is a ``pandas.DataFrame``, ``schema`` has to be a :class:`~snowflake.snowpark.types.StructType` if specified.
```

src/snowflake/snowpark/session.py
```diff
@@ -1274,14 +1276,24 @@ def create_dataframe(
             sf_database = self._conn._get_current_parameter("database", quoted=False)
             sf_schema = self._conn._get_current_parameter("schema", quoted=False)
 
+            if schema:
+                if not isinstance(schema, StructType):
+                    raise ValueError(
+                        f"Expects StructType for schema if specified when creating snowpark dataframe from pandas dataframe. Got {type(schema)}"
+                    )
+
+                self._run_query(
```
sfc-gh-jdu, line 1285: If write_pandas fails, then this table is left without being cleaned? Should we use try/catch to remove this table if write_pandas failed?

tests/integ/test_pandas_to_df.py
```diff
@@ -295,6 +303,88 @@ def test_create_dataframe_from_pandas(session):
     # assert_frame_equal(results, pd, check_dtype=False)
 
 
+def test_create_dataframe_from_pandas_with_schema(session):
```
sfc-gh-jdu, line 306: can we also try time-related types and some structured types?

src/snowflake/snowpark/session.py
sfc-gh-sfan: Nice catch. Let me do that.

tests/integ/test_pandas_to_df.py
sfc-gh-sfan: Time-related types requires enabling parameter that controls parquet timestamp behavior, otherwise I'm hitting `E               snowflake.snowpark.exceptions.SnowparkFetchDataException: (1406): Failed to fetch a Pandas Dataframe. The error is: Casting from timestamp[us] to timestamp[ns] would result in out of bounds timestamp: 5325833340940354560
`. Looks like the similar test is also disabled in `test_create_dataframe_from_pandas`. 

Should we wait until new logical type is rolled out in production?

sfc-gh-jdu: hmm sure, let's have a jira for that? (and when will new logical type be in production? )

sfc-gh-sfan: I asked Yucan but haven't heard back yet. Will create a JIRA that covers both this test and `test_create_dataframe_from_pandas`.

sfc-gh-sfan: Unfortunately we might not be able to test for semi-structured types as well due to connector returning json strings :/ Will create another JIRA to track this test TODO as well.

src/snowflake/snowpark/session.py
```diff
@@ -1286,15 +1290,21 @@ def create_dataframe(
...
-            )
+            try:
+                t = self.write_pandas(
+                    data,
+                    table_name,
+                    database=sf_database,
+                    schema=sf_schema,
+                    quote_identifiers=True,
+                    auto_create_table=schema is None,
+                    # auto_create_table=True,
```
sfc-gh-jdu, line 1301: let's explicitly set it to False, and don't set `table_type`?

tests/integ/test_pandas_to_df.py
```diff
@@ -295,6 +304,90 @@ def test_create_dataframe_from_pandas(session):
...
```
sfc-gh-jdu, line 307: sorry I should mention it in my previous comments, but let's try binary and decimal types? We have such tests on these two types for create_dataframe from native python values.

sfc-gh-jdu (approved):

src/snowflake/snowpark/session.py
sfc-gh-sfan: Sorry this is some leftover from my local testing. But yeah I think this is a good idea.

tests/integ/test_pandas_to_df.py
sfc-gh-sfan: I added test for decimal types. 

Today `write_pandas` does not work great with `binary` type as I will hit the following if I created pandas DF with `b"t1"`:

`Expression type does not match column data type, expecting BINARY(8388608) but got VARIANT for column SHOE_MODEL_BINARY`

Looks like a bug and we should use `TO_BINARY` for our generated `copy_into_sql` [here](https://github.com/snowflakedb/snowflake-connector-python/blob/0afbf902b020c99d2db47e24aa181ad2085c421c/src/snowflake/connector/pandas_tools.py#L268-L273)? cc @sfc-gh-mkeller  If that is the case then I could create a ticket so we could work on that separately, and in the meanwhile create another TODO to add test for binary type.

sfc-gh-sfan:
This is a behavior change and the next scheduled behavior change release is going to be next year, so this PR will be shelved until we are ready for next BCR.

tests/integ/test_pandas_to_df.py
sfc-gh-sfan: Apparently there was a long discussion about binary not working since python connector 2.3.7. Currently this is blocked by snowflake internal feature tracked in https://snowflakecomputing.atlassian.net/browse/SNOW-204246

PR #575, opened 2022-10-25
SNOW-681372: Use session parameter PYTHON_SNOWPARK_USE_SQL_SIMPLIFIER

sfc-gh-yixie:

   Fixes #SNOW-681372

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -263,27 +263,18 @@ class Session:
         self._plan_builder = SnowflakePlanBuilder(self)
         self._last_action_id = 0
         self._last_canceled_id = 0
-        # consider making _session_parameters check a helpful function
-        self._use_scoped_temp_objects = bool(
+        self._use_scoped_temp_objects: bool = (
             _use_scoped_temp_objects
-            and (
-                conn._conn._session_parameters.get(
-                    _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, True
-                )
-                if conn._conn._session_parameters
-                else True
+            and self._get_client_side_session_parameter(
+                _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, True
             )
         )
 
         self._file = FileOperation(self)
 
         self._analyzer = Analyzer(self)
-        self._sql_simplifier_enabled: bool = (
-            conn._conn._session_parameters.get(
-                _PYTHON_SNOWPARK_USE_SQL_SIMPLIFIER_STRING, False
-            )
-            if conn._conn._session_parameters
-            else False
+        self._sql_simplifier_enabled: bool = self._get_client_side_session_parameter(
+            _PYTHON_SNOWPARK_USE_SQL_SIMPLIFIER_STRING, False
         )
         _logger.info("Snowpark Session information: %s", self._session_info)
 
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -277,7 +278,13 @@ def __init__(self, conn: ServerConnection) -> None:
         self._file = FileOperation(self)
 
         self._analyzer = Analyzer(self)
-        self._sql_simplifier_enabled: bool = False
+        self._sql_simplifier_enabled: bool = (
+            conn._conn._session_parameters.get(
+                _PYTHON_SNOWPARK_USE_SQL_SIMPLIFIER_STRING, False
+            )
```
sfc-gh-yixie, line 284: @sfc-gh-mkeller Is this the right way to get a session parameter from a SnowflakeConnection if the parameter exists in the Snowflake database?

src/snowflake/snowpark/session.py
```diff
@@ -277,7 +278,13 @@ def __init__(self, conn: ServerConnection) -> None:
         self._file = FileOperation(self)
 
         self._analyzer = Analyzer(self)
-        self._sql_simplifier_enabled: bool = False
+        self._sql_simplifier_enabled: bool = (
```
sfc-gh-sfan, line 281: Since this is the second usage of the same pattern like `_use_scoped_temp_objects`, I think we should extract this into a helper/util function to avoid code duplication.

src/snowflake/snowpark/session.py
```diff
@@ -330,6 +337,12 @@ def sql_simplifier_enabled(self, value: bool) -> None:
         self._conn._telemetry_client.send_sql_simplifier_telemetry(
             self._session_id, value
         )
+        try:
+            self._conn.run_query(
+                f"alter session set {_PYTHON_SNOWPARK_USE_SQL_SIMPLIFIER_STRING} = {value}"
```
sfc-gh-sfan, line 342: Just a quick note that this won't work inside owner's right stored proc. Should be okay since this is just a setter and not on the critical path, and for stored proc they could `alter session` before calling the stored proc.

tests/integ/test_session.py
```diff
@@ -371,3 +376,19 @@ def test_use_secondary_roles(session):
 
     # after stripping the quotes actually it works - but it's not documented in Snowflake
     session.use_secondary_roles(current_role[1:-1])
+
+
+def test_sql_simplifier_enabled_on_session(db_parameters):
```
sfc-gh-sfan, line 381: Let's skip stored proc test.

src/snowflake/snowpark/session.py
sfc-gh-jdu: You can use something similar to `self._use_scoped_temp_objects`

src/snowflake/snowpark/session.py
sfc-gh-jdu: this will also not work on our merge gate (in production), because the role doesn't have the privilege to alter the parameter, so does for customers. So I guess here just changing the local attribute is ok.

src/snowflake/snowpark/session.py
sfc-gh-yixie: It's aligned with use scoped temp objects

tests/integ/test_session.py
sfc-gh-yixie: Added skip

src/snowflake/snowpark/session.py
sfc-gh-yixie: Manually tested with my local regression environment. Confirmed that this works.

src/snowflake/snowpark/session.py
sfc-gh-yixie: Added an internal function `_get_client_side_session_parameter` on class Session.

sfc-gh-sfan (approved):

sfc-gh-jdu (approved):

PR #582, opened 2022-10-27
SNOW-680030 Organize the API reference in terms of functionality

sfc-gh-stan:

   Fixes #SNOW-680030

   Change the API organization.


docs/source/_templates/autosummary/class.rst
```diff
@@ -1,15 +1,15 @@
...
 
-{% set methods = methods|reject("equalto", "__init__") %}
+{% set methods =(methods| reject("equalto", "__init__") |list) %}
 
-   {% block methods %}
+{% block methods %}
 
-   {% if methods %}
-   .. rubric:: {{ _('Methods') }}
+{% if methods and methods|length > 0 %}
```
sfc-gh-stan, line 7: This is redundant, I will remove this.

docs/source/conf.py
```diff
@@ -47,7 +47,7 @@
 autosummary_generate = True
 autoclass_content="class"
 autodoc_default_options = {
-    "members": True,  # Include all methods.
+    #"members": True,  # Include all methods.
```
sfc-gh-stan, line 50: I removed this because we are explicitly listing the methods now.

docs/source/conf.py
```diff
@@ -90,6 +90,4 @@
 html_show_sourcelink = False  # Hide "view page source" link
 
 # Disable footer message "Built with Sphinx using a theme provided by Read the Docs."
-html_show_sphinx = False
-
-add_module_names = False
+html_show_sphinx = False
```
sfc-gh-stan, line 93: Will fix.

sfc-gh-stan:
I will categorize `snowflake.snowpark.functions` in a different ticket.

docs/source/_templates/autosummary/class.rst
```diff
@@ -1,15 +1,15 @@
 {% extends "!autosummary/class.rst" %}
 
-{% set methods = methods|reject("equalto", "__init__") %}
+{% set methods =(methods| reject("equalto", "__init__") |list) %}
```
sfc-gh-sfan, line 3: Just curious what this `list` does

docs/source/conf.py
sfc-gh-sfan, line 50: Probably set it to False or just remove it?

sfc-gh-sfan (commented):
Curious: does this mean we need to add to the docs for new APIs that we introduce?

docs/source/column.rst
```diff
@@ -0,0 +1,63 @@
...
+    Column.name
+    Column.over
+    Column.regexp
+    Column.rlike
+    Column.startswith
+    Column.substr
+    Column.substring
+    Column.try_cast
+    Column.within_group
+
```
sfc-gh-jdu, line 59: These blank lines are not necessary?

docs/source/IO.rst
```diff
@@ -0,0 +1,63 @@
+=========================
```
sfc-gh-jdu, line 1: nit: should we use io.rst as the file name to keep consistent with other files?

docs/source/table.rst
```diff
@@ -0,0 +1,135 @@
...
+    :toctree: api/
+
+    Table
+
+.. rubric:: Methods
+
+.. autosummary::
+    :toctree: api/
+
+    Table.agg
```
sfc-gh-jdu, line 19: Table subclasses DataFrame so it inherits many methods. But actually it only have several methods specific to Table class, which are more important. Table class on the original API doc only has these methods, which is more clear I believe.

sfc-gh-jdu (commented):
I have the same question with Shixuan: how should we maintain this API doc with incoming changes in the code in the future? It shouldn't become a manual process.

docs/source/index.rst
```diff
@@ -13,13 +13,22 @@ information, see the `Snowpark Developer Guide for Python <https://docs.snowflak
...
+   grouping 
    table_function
+   table
    stored_procedures
-   types
    udf
    udtf
    exceptions
+   observability
+   async_job
```
sfc-gh-aalam, line 32: should we do them in sorted order?

docs/source/_templates/autosummary/class.rst
sfc-gh-stan: This is because `| reject` creates a generator, which defies the purpose of `if methods` below, turning it back into a list will ensure that when the list is empty, we don't print the section header "Methods"

docs/source/conf.py
sfc-gh-stan: Will remove.

docs/source/IO.rst
sfc-gh-stan: Good point, will change.

docs/source/table.rst
sfc-gh-stan: I see, I copy/pasted the contents of the generate file, lemme remove the inherited methods then.

sfc-gh-stan:
> I have the same question with Shixuan: how should we maintain this API doc with incoming changes in the code in the future? It shouldn't become a manual process.

Good question, for now I am copy/pasting the contents from the generated `rst` files manually, in the future we could probably write a script to do this.

CC: @sfc-gh-sfan

docs/source/index.rst
sfc-gh-stan: I am roughly following PySpark's order here. But I'm also open to using sorted order.

sfc-gh-sfan:
If they come from different files/classes, can we make the doc smart enough to auto-generate based on files/classes so we only need to maintain it at a coarser level? I don't have much context so this might not be possible.

docs/source/index.rst
sfc-gh-yixie: Good question. I was also thinking about ordering. How about moving `async_job` next after `table` and `observability` just before `exceptions`?

docs/source/IO.rst
```diff
@@ -0,0 +1,63 @@
...
+
+.. currentmodule:: snowflake.snowpark
+
+.. rubric:: Classes
+    
+.. autosummary::
+    :toctree: api/
+
+    DataFrameReader
```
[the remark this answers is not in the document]
sfc-gh-stan: This is manually edited.

docs/source/index.rst
sfc-gh-stan: @sfc-gh-yixie , sounds good

sfc-gh-stan:
> If they come from different files/classes, can we make the doc smart enough to auto-generate based on files/classes so we only need to maintain it at a coarser level? I don't have much context so this might not be possible.

We can definitely automate by writing [a Sphinx extension](https://www.sphinx-doc.org/en/master/extdev/appapi.html#extension-setup) to customize the `autosummary` execution, something like what Pandas did [here](https://github.com/pandas-dev/pandas/blob/3020b8df58927ef0246091d325debce7097ace60/doc/source/conf.py#L786) I will follow up with the documentation team to discuss whether we should invest more effort in Sphinx, since the new platform might get rid of this dependency?

Ticket [SNOW-685279](https://snowflakecomputing.atlassian.net/browse/SNOW-685279) is created to track this.

docs/source/udf.rst
```diff
@@ -1,23 +1,32 @@
-snowflake.snowpark.udf 
+UDF
```
sfc-gh-jdu, line 1: Probably it's better to use `User-defined Functions`, the same for udtf

sfc-gh-jdu (approved):

sfc-gh-sfan (approved):

sfc-gh-yixie (approved):

PR #594, opened 2022-10-31
SNOW-683059 Review test coverage for column.py, dataframe.py, etc

sfc-gh-stan:

   Fixes #683059


src/snowflake/snowpark/column.py
```diff
@@ -605,7 +605,7 @@ def __str__(self):
         return f"Column[{self._expression}]"
 
     def __repr__(self):
-        return f"Column({self._expression})"
+        return f"Column({self._expression})"  # pragma: no cover
```
sfc-gh-stan, line 608: The onlt test I could think of for `__repr__` is `assert eval(repr(x)) == x`, which does not apply for our case.

src/snowflake/snowpark/file_operation.py
```diff
@@ -219,8 +219,7 @@ def get(
                     self._session, plan
                 )._internal_collect_with_tag(statement_params=statement_params)
             return [GetResult(**file_result.asDict()) for file_result in get_result]
-        # connector raises IndexError when no file is downloaded from python connector.
-        except IndexError:
+        except OperationalError:
```
sfc-gh-stan, line 222: The second test case in `test_get_negative_test` fails with OperationalError instead of IndexError. I looked into Python connector's code and couldn't find where it raises `IndexError` in file transfer related code.

sfc-gh-stan: Also:  Looks like `test_get_negative_test` is skipped to avoid Azure/GCP-specific issues, we should still run it for AWS. Let's follow Python connector and use markers `@pytest.mark.aws` to select the corresponding tests in our tox config.

src/snowflake/snowpark/column.py
```diff
@@ -81,7 +81,7 @@ def _to_col_if_lit(
         return col
     elif isinstance(col, VALID_PYTHON_TYPES_FOR_LITERAL_VALUE):
         return Column(Literal(col))
-    else:
+    else:  # pragma: no cover
```
sfc-gh-sfan, line 84: Frankly I think these `else` cases should be fairly simple to test in a unit test, though given they are "private" functions, it is probably okay to skip.

src/snowflake/snowpark/column.py
sfc-gh-sfan, line 608: Why is `__repr__` different from `__str__`? Also this might be testable?

src/snowflake/snowpark/dataframe.py
```diff
@@ -813,7 +809,7 @@ def to_df(self, *names: Union[str, Iterable[str]]) -> "DataFrame":
         """
         col_names = parse_positional_args_to_list(*names)
         if not all(isinstance(n, str) for n in col_names):
-            raise TypeError(
+            raise TypeError(  # pragma: no cover
```
sfc-gh-sfan, line 812: This is an API call. Should we have a negative test?

src/snowflake/snowpark/dataframe.py
```diff
@@ -840,12 +836,12 @@ def __getitem__(self, item):
         elif isinstance(item, int):
             return self.__getitem__(self.columns[item])
         else:
-            raise TypeError(f"Unexpected item type: {type(item)}")
+            raise TypeError(f"Unexpected item type: {type(item)}")  # pragma: no cover
```
sfc-gh-sfan, line 839: I feel this should be tested?

src/snowflake/snowpark/dataframe.py
```diff
@@ -840,12 +836,12 @@ def __getitem__(self, item):
...
 
     def __getattr__(self, name):
         # Snowflake DB ignores cases when there is no quotes.
         if name.lower() not in [c.lower() for c in self.columns]:
-            raise AttributeError(
+            raise AttributeError(  # pragma: no cover
```
sfc-gh-sfan, line 844: Is this testable?

src/snowflake/snowpark/dataframe.py
```diff
@@ -956,7 +952,7 @@ def select(
                 )
                 names.extend(new_cols)
             else:
-                raise TypeError(
+                raise TypeError(  # pragma: no cover
```
sfc-gh-sfan, line 955: This is an API call. Should we have a negative test?

src/snowflake/snowpark/dataframe.py
```diff
@@ -1038,7 +1034,7 @@ def drop(
         """
         # an empty list should be accept, as dropping nothing
         if not cols:
-            raise ValueError("The input of drop() cannot be empty")
+            raise ValueError("The input of drop() cannot be empty")  # pragma: no cover
```
sfc-gh-sfan, line 1037: This is an API call. Should we have a negative test?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2006,7 +2002,7 @@ def join(
             elif not isinstance(using_columns, list):
                 raise TypeError(
                     f"Invalid input type for join column: {type(using_columns)}"
-                )
+                )  # pragma: no cover
```
sfc-gh-sfan, line 2005: This is an API call. Should we have a negative test?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2016,7 +2012,7 @@ def join(
                 rsuffix=rsuffix,
             )
 
-        raise TypeError("Invalid type for join. Must be Dataframe")
+        raise TypeError("Invalid type for join. Must be Dataframe")  # pragma: no cover
```
sfc-gh-sfan, line 2015: This is an API call. Should we have a negative test?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2869,7 +2865,7 @@ def create_or_replace_view(
         elif isinstance(name, (list, tuple)) and all(isinstance(n, str) for n in name):
             formatted_name = ".".join(name)
         else:
-            raise TypeError(
+            raise TypeError(  # pragma: no cover
```
sfc-gh-sfan, line 2868: This is an API call. Should we have a negative test?

src/snowflake/snowpark/dataframe.py
```diff
@@ -3176,7 +3172,9 @@ def with_column_renamed(self, existing: ColumnOrName, new: str) -> "DataFrame":
                     f"Unable to rename column {existing} because it doesn't exist."
                 )
         else:
-            raise TypeError("'exisitng' must be a column name or Column object.")
+            raise TypeError(
```
sfc-gh-sfan, line 3175: This is an API call. Should we have a negative test?

src/snowflake/snowpark/dataframe.py
```diff
@@ -3374,7 +3372,7 @@ def _explain_string(self) -> str:
             if exec_plan:
                 msg = f"{msg}\nLogical Execution Plan:\n{exec_plan}"
             else:
-                msg = f"{self._plan.queries[0].sql} can't be explained"
+                msg = f"{self._plan.queries[0].sql} can't be explained"  # pragma: no cover
```
sfc-gh-sfan, line 3375: I wonder when this would happen

src/snowflake/snowpark/column.py
sfc-gh-stan: [`__repr__` vs `__str__`](https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr)
`assert eval(repr(x)) == x` does not apply for this case, so I chose to skip.

sfc-gh-sfan: Makes sense. I thought using parenthesis is not too bad, but anyways this is okay :)

src/snowflake/snowpark/dataframe.py
sfc-gh-stan: I don't think we need to test for negative cases of input that are indicated by the type hints already, but I'm open to discussion :)

src/snowflake/snowpark/dataframe.py
```diff
@@ -1038,7 +1034,7 @@ def drop(
         """
         # an empty list should be accept, as dropping nothing
```
sfc-gh-stan, line 1035: This comment is contradicting the sanity check on line 1037 below 😕

src/snowflake/snowpark/column.py
sfc-gh-jdu: Should we change to use `__str__`. It shouldn't be a breaking change

src/snowflake/snowpark/dataframe_na_functions.py
```diff
@@ -127,6 +127,14 @@ def drop(
             |4.0  |NULL  |
             --------------
             <BLANKLINE>
+            >>> df.na.drop(subset="a").show()
```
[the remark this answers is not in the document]
sfc-gh-jdu: Dropping support for this case will be a behavior change. Let's just fix the type hints. We should also support it for spark migration.

src/snowflake/snowpark/file_operation.py
sfc-gh-jdu: This will be a behavior change

src/snowflake/snowpark/dataframe_na_functions.py
```diff
@@ -137,7 +145,9 @@ def drop(
         # iff(non_float_col is null, 0, 1) >= thresh
 
         if how is not None and how not in ["any", "all"]:
-            raise ValueError("how ('" + how + "') should be 'any' or 'all'")
+            raise ValueError(
+                "how ('" + how + "') should be 'any' or 'all'"
```
sfc-gh-jdu, line 149: use f-string here?

src/snowflake/snowpark/dataframe.py
sfc-gh-jdu: Yea we should have negative test for this one

src/snowflake/snowpark/dataframe.py
sfc-gh-sfan: Oh I'm not aware of that lol. Github just shows this line to me. It does not hurt to test but yeah it's also fine to leave it as is. No strong opinion then.

src/snowflake/snowpark/dataframe_na_functions.py
```diff
@@ -146,7 +146,7 @@ def drop(
 
         if how is not None and how not in ["any", "all"]:
             raise ValueError(
```
sfc-gh-aalam, line 148: I think we should add a negative test for checking this as well

src/snowflake/snowpark/file_operation.py
sfc-gh-stan: `except OperationalError`? But I think this is the intended behavior: when no files are downloaded, just return a empty list and don't raise exceptions.

sfc-gh-jdu: yea, changing the exception type is usually a behavior change because some users might use try/except to catch the expected exception type.

sfc-gh-yixie: I suggest keeping the IndexError there. I remember I've seen IndexError before.

src/snowflake/snowpark/dataframe.py
sfc-gh-yixie: @sfc-gh-jdu Do you know what sql can't be explained?

src/snowflake/snowpark/column.py
sfc-gh-yixie: Makes sense. The problem is `Expression` doesn't override `__repr__`.

src/snowflake/snowpark/file_operation.py
sfc-gh-stan: Okay, I will revert this change for now, [SNOW-687929](https://snowflakecomputing.atlassian.net/browse/SNOW-687929) is created to track this fix. [SNOW-687933](https://snowflakecomputing.atlassian.net/browse/SNOW-687933) is created to track the proposal of using cloud provider specific pytest marks.

sfc-gh-jdu (approved):

PR #623, opened 2022-11-22
SNOW-685279 Add utility scripts for generating documentation

sfc-gh-stan:

   Fixes #SNOW-685279


sfc-gh-aalam (commented):
The change looks good to me. How has it been tested?

docs/source/util.py
```diff
@@ -0,0 +1,174 @@
...
+
+def autogen_and_parse_for_info(module_name: str, class_name: Optional[str] = None):
+    if class_name:
+        res = Class(module_name, [], [])
+        name = f"{module_name}.{class_name}"
+    else:
+        res = Module(module_name, [], [], [], [])
+        name = module_name
+
+    if not os.path.exists("temp"):
```
sfc-gh-sfan, line 38: I wonder if we should check and make sure we start from a new directory before generating contents? I'm not sure if reusing the same directory will cause some unexpected lingering file when it is supposed to be deleted.

docs/source/util.py
```diff
@@ -0,0 +1,174 @@
...
+
+    section = ""
+
+    with open(f"output/{name}.rst") as fp:
+        for line in fp:
+            line = line.strip()
+            if line.startswith(".. autosummary::"):
+                continue
+            if line.startswith(".. rubric:: "):
+                line = line[12:]
```
sfc-gh-sfan, line 67: Instead of using 12, can we put `len(".. rubric:: ")` instead, or maybe make both `.. rubric:: ` and `.. autosummary:: ` a static constant string variable, so we can do `len(RUBRIC_HEADER)` for example

docs/source/util.py
```diff
@@ -0,0 +1,174 @@
...
+                for title in mapping.keys():
+                    if line.startswith(title):
+                        section = mapping[title]
+                        break
+                continue
+            if not section:
+                continue
+            if line and not line.isspace():
+                getattr(res, section).append(line.rstrip())
+            elif getattr(res, section):  # End of a section
```
sfc-gh-sfan, line 77: Curious: is this check needed? This means we only reset `section` when we put something into the section. What is the scenario this check tries to work for?

docs/source/util.py
```diff
@@ -0,0 +1,174 @@
...
+    else:
+        return ""
+
+
+def generate(title: str, module: str, classes: Iterable[str]) -> str:
+    import itertools
+
+    if classes:
+        results = [autogen_and_parse_for_info(module, c) for c in classes]
+        names = "\n    =".join(classes)
```
sfc-gh-sfan, line 103: I wonder if we could make `\n` + 4 spaces a variable, so that we don't need to count it every time.

docs/source/util.py
```diff
@@ -0,0 +1,174 @@
...
+
+"""
+    else:
+        return ""
+
+
+def generate(title: str, module: str, classes: Iterable[str]) -> str:
+    import itertools
+
+    if classes:
```
sfc-gh-sfan, line 101: There is close to zero shared code for this `if` and `else`. Maybe we could extract them into two methods: `generate_classes` and `generate_modules`?

docs/source/util.py
```diff
@@ -0,0 +1,174 @@
...
+
+    {content}
+
+"""
+    else:
+        return ""
+
+
+def generate(title: str, module: str, classes: Iterable[str]) -> str:
+    import itertools
```
sfc-gh-sfan, line 99: nit: maybe this can go to the beginning of the file?

docs/source/util.py
```diff
@@ -0,0 +1,174 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        prog="generate_doc",
+        description="Generate documentation rst for a module or classes inside a module",
+    )
+
+    parser.add_argument(
+        "module", help="The module or the parent module of the classes to be documented"
+    )
+    parser.add_argument("-c", "--classes", nargs="*", help="Classes to be documented")
```
sfc-gh-sfan, line 160: I wonder if we should document how the current docs can be generated using this script. Also, let's just name this `doc_generator.py` or something that is clearer than `util` :p

sfc-gh-sfan (commented):
Looks good, just some nits and questions

docs/source/util.py
sfc-gh-stan: This script should be used for reference only. There are parts of the current files that just require manual change. The usage I imagine would be something like: 

- There're some interface changes made to `DataFrame`
- Run `util.py` to get the latest list of methods/attributes.
- Paste the list in our rst files manually.

sfc-gh-jdu: which parts of the current files require manual change? We can discuss in this PR and also should document it

docs/README.md
```diff
@@ -22,13 +22,19 @@ python -m pip install sphinx
...
-`docs/source/index.rst`: Specify which modules to generate an autosummary for. This also controls what shows up in the `index.html` landing page.
+`docs/source/index.rst`: Specify which rst to include in the `index.html` landing page.
 `docs/source/conf.py`: The configuration parameters for Sphinx and autosummary.
-`docs/source/_templates/module.rst`: This is a Jinja template that controls how the info on each module doc page is ordered and laid out. You tell Sphinx to use this template in index.rst.
+`docs/source/_templates/`: Directory containing JINJA templates used by autosummary.
 `docs/source/_themes/snowflake_rtd_theme/layout.html`: All of the theme stuff is the same as the snowflake_rtd_theme in the normal Snowflake doc repo. This layout file is the only thing I modified.
+`docs/source/util.py`: A utility script that generates a rst for a module or a list of classes within the same module. This should be used for reference only. 
+Example usage: 
+```
+python3 util.py snowflake.snowpark -c DataFrame DataFrameNaFunctions DataFrameStatFunctions
```
sfc-gh-jdu, line 34: It would be awesome if we can automate it in `Makefile`.

docs/source/util.py
sfc-gh-stan: From the top of my head:
- In `types.rst`, autosummary will categorize `Geography` and `Variant` to `Attributes` instead of `Classes` like the rest of type definitions.
- Still in `types.rst`, the current script does not support custom rubric names like `Classes/Type Hints`
- If we want to leave any comment/literal markdown in the `rst`, for example my TODO comment in `session.rst`, the script does not preserve these.

sfc-gh-jdu: I see, let's mention them somewhere?

sfc-gh-sfan: Yeah I think it would be useful to document the known caveat. People are lazy (at least I am :p) and will rely on the script more and more without realizing something might not work if not documented.

docs/source/exceptions.rst
```diff
@@ -1,11 +1,20 @@
+
+
+===============
 Exceptions
-=============================
-This package contains all Snowpark client-side exceptions.
```
sfc-gh-stan, line 3: These comments are auto extracted by `automodule` directive, I used `noindex` to avoid duplicate indexing of the methods/classes.

sfc-gh-stan:
> The change looks good to me. How has it been tested?

For testing, the rst files changed in this PR are all generated by this script. Please let me know if you found a bug :)

docs/source/doc_gen.py
```diff
@@ -0,0 +1,179 @@
...
+        content = generate_classes(args.title, args.module, args.classes)
+    else:
+        content = generate_module(args.title, args.module)
+
+    if args.filename:
+        with open(args.filename, "w") as fp:
+            fp.write(content)
+    else:
+        print(content)
+
```
sfc-gh-sfan, line 179: For the no new line warning, is it because we have some space at the last line?

sfc-gh-sfan (approved):

docs/source/doc_gen.py
sfc-gh-stan: Oh you are right, that fixed it!

PR #632, opened 2022-12-06
SNOW-671000 Make sure all ProgrammingError is wrapped as SnowparkSQLException

sfc-gh-lspiegelberg:

   Fixes #SNOW-671000

   Add new decorator to Session class to wrap connector's ProgrammingError exception as SnowparkSQLException. Verify through codebase that ProgrammingError is not used in tests explicitly anymore.


src/snowflake/snowpark/session.py
```diff
@@ -518,6 +546,7 @@ def _resolve_import_path(
         else:
             return trimmed_path, None, None
 
+    @Decorator.wrap_exception
     def _resolve_imports(
```
sfc-gh-sfan, line 550: I don't have strong opinion, but I wonder if we should just wrap it for public APIs if there is no information loss?

sfc-gh-sfan (commented):
This is a behavior change so we probably should hold on merging it until the first behavior change release in January.

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +194,29 @@ class Session:
...
+            def wrap(*args, **kwargs):
+                try:
+                    return func(*args, **kwargs)
+                except ProgrammingError as e:
+                    tb = sys.exc_info()[2]
+                    if "unexpected 'as'" in e.msg.lower():
+                        ne = (
+                            SnowparkClientExceptionMessages.SQL_PYTHON_REPORT_UNEXPECTED_ALIAS()
+                        )
+                        raise ne.with_traceback(tb) from None
```
sfc-gh-stan, line 211: I see that `raise ... from None` suppresses the context, does https://github.com/snowflakedb/snowpark-python/blob/7672c690c9be1e75948aac1e7c1c55c89d7abf53/src/snowflake/snowpark/_internal/server_connection.py#L100 need this too then?

sfc-gh-lspiegelberg:
> Actually we don't need to implement a new decorator, we already have one here https://github.com/snowflakedb/snowpark-python/blob/0e87cdfe2730f23208055cfbe2151838d51c1ffe/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#LL79C9-L79C9. And we call it here
> 
> https://github.com/snowflakedb/snowpark-python/blob/0e87cdfe2730f23208055cfbe2151838d51c1ffe/src/snowflake/snowpark/_internal/server_connection.py#L419
> 
> . We could just make sure methods in session.py (or other files) that executes a query directly will call `get_result_set` (currently we use `self._run_query`) or call a method also decorated with `@SnowflakePlan.Decorator.wrap_exception` (for example, we can remove the usage of `run_query `or decorate `run_query` with `@SnowflakePlan.Decorator.wrap_exception`). APIs creating a dataframe (such as session.create_dataframe) shouldn't need a decorator because itself doesn't execute a query.

Originally i wanted to use this decorator, yet it has dependencies on the plan/dataframe. But session has not always a dataframe/plan associated with it afaik. Therefore, I added this new decorator which only focuses on ProgrammingError. There's some options to refactor though: We could either subclass the two decorators, or move them somewhere else (e.g., exceptions.py).  Let me know what you think the best approach would be!

sfc-gh-jdu:
> > Actually we don't need to implement a new decorator, we already have one here https://github.com/snowflakedb/snowpark-python/blob/0e87cdfe2730f23208055cfbe2151838d51c1ffe/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#LL79C9-L79C9. And we call it here
> > https://github.com/snowflakedb/snowpark-python/blob/0e87cdfe2730f23208055cfbe2151838d51c1ffe/src/snowflake/snowpark/_internal/server_connection.py#L419
> > 
> > . We could just make sure methods in session.py (or other files) that executes a query directly will call `get_result_set` (currently we use `self._run_query`) or call a method also decorated with `@SnowflakePlan.Decorator.wrap_exception` (for example, we can remove the usage of `run_query `or decorate `run_query` with `@SnowflakePlan.Decorator.wrap_exception`). APIs creating a dataframe (such as session.create_dataframe) shouldn't need a decorator because itself doesn't execute a query.
> 
> Originally i wanted to use this decorator, yet it has dependencies on the plan/dataframe. But session has not always a dataframe/plan associated with it afaik. Therefore, I added this new decorator which only focuses on ProgrammingError. There's some options to refactor though: We could either subclass the two decorators, or move them somewhere else (e.g., exceptions.py). Let me know what you think the best approach would be!

Yes, I feel the best approach will be merging exception decorator in server_connection.py and snowflake_plan.py, and move it to exception.py. This is an legacy issue that we should fix.

src/snowflake/snowpark/session.py
sfc-gh-sfan: My original thought is that we probably don't need to wrap "private" functions because customer should not be invoking them, but it is possible that these functions like `_resolve_imports` are used outside of this class.

src/snowflake/snowpark/session.py
sfc-gh-lspiegelberg: I haven't touched the server_connection.py file yet. Should the exception wrapper here be refactored as well? It looks to me like separate logic from the rest, but we can add an `from None` to be consistent when it comes to context suppression. Thoughts?

src/snowflake/snowpark/session.py
sfc-gh-lspiegelberg: when I went through the code base to target the `ProgrammingError` exception I was specifically looking for any use of `_conn` and wrapped any functions using it. Given private functions in Python are not really private (a user may still access them!) I think it's probably best to wrap given the context, but no strong opinion on that.

src/snowflake/snowpark/session.py
sfc-gh-lspiegelberg: added context suppression to server_connection.py specific decorator.

sfc-gh-lspiegelberg:
> > > Actually we don't need to implement a new decorator, we already have one here https://github.com/snowflakedb/snowpark-python/blob/0e87cdfe2730f23208055cfbe2151838d51c1ffe/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#LL79C9-L79C9. And we call it here
> > > https://github.com/snowflakedb/snowpark-python/blob/0e87cdfe2730f23208055cfbe2151838d51c1ffe/src/snowflake/snowpark/_internal/server_connection.py#L419
> > > 
> > > . We could just make sure methods in session.py (or other files) that executes a query directly will call `get_result_set` (currently we use `self._run_query`) or call a method also decorated with `@SnowflakePlan.Decorator.wrap_exception` (for example, we can remove the usage of `run_query `or decorate `run_query` with `@SnowflakePlan.Decorator.wrap_exception`). APIs creating a dataframe (such as session.create_dataframe) shouldn't need a decorator because itself doesn't execute a query.
> > 
> > 
> > Originally i wanted to use this decorator, yet it has dependencies on the plan/dataframe. But session has not always a dataframe/plan associated with it afaik. Therefore, I added this new decorator which only focuses on ProgrammingError. There's some options to refactor though: We could either subclass the two decorators, or move them somewhere else (e.g., exceptions.py). Let me know what you think the best approach would be!
> 
> Yes, I feel the best approach will be merging exception decorator in server_connection.py and snowflake_plan.py, and move it to exception.py. This is an legacy issue that we should fix.

Refactored the decorator to `utils.py` because of the interactions between `exceptions.py` and `error_messages.py`. While looking through the source, server_connection seems to require its own dedicated decorator to wrap exceptions.

Duuuuuu (commented):
The refactoring looks good. I'm wondering whether we can avoid adding @wrap_exception to every method running a query directly, and instead we can add @wrap_exception to some internal methods in server_connection.py?

src/snowflake/snowpark/session.py
sfc-gh-yixie: `wrap_exception` is already used by internal methods, for instance `ServerConnection.get_result_set`, which is called directly or indirectly by many public APIs. So we don't need to use `wrap_exception` on every public APIs.

sfc-gh-yixie (changes requested):
I guess we can just add the decorator to a few internal functions instead of on the many public APIs.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -93,13 +93,13 @@ def wrap_exception(cls, func):
             def wrap(*args, **kwargs):
                 # self._conn.is_closed()
                 if args[0]._conn.is_closed():
-                    raise SnowparkClientExceptionMessages.SERVER_SESSION_HAS_BEEN_CLOSED()
+                    raise SnowparkClientExceptionMessages.SERVER_SESSION_HAS_BEEN_CLOSED() from None
```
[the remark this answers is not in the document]
sfc-gh-lspiegelberg: this is a change requested by @sfc-gh-stan to make errors behave the same way across the code base

src/snowflake/snowpark/session.py
sfc-gh-lspiegelberg: So I think this should be a bit more of a design decision. If we wrap at top-level public functions, then clearly we expose that any exception trace will be converted properly. If we have a mix of internal functions that may throw an exception together with a wrapper, and some functions wrapped/not wrapped we run the risk of some changes being introduced that may produce exceptions that are not translated properly. Therefore, I'd say let's remove all internal wrapping unless there's a very good reason and put the wrapper around any public interface. Thoughts?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -162,6 +165,121 @@
 SUPPORTED_TABLE_TYPES = ["temp", "temporary", "transient"]
 
 
+_NUM_PREFIX_DIGITS = 4
```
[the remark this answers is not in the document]
sfc-gh-lspiegelberg: done.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -70,103 +74,23 @@
...
-        self,
-        queries: List["Query"],
-        schema_query: str,
-        post_actions: Optional[List["Query"]] = None,
-        expr_to_alias: Optional[Dict[uuid.UUID, str]] = None,
-        session: Optional["snowflake.snowpark.session.Session"] = None,
-        source_plan: Optional[LogicalPlan] = None,
-        is_ddl_on_temp_object: bool = False,
-        api_calls: Optional[List[Dict]] = None,
+            self,
```
sfc-gh-jdu, line 85: Is this indent change intentional? I think our formatter should fix this?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -85,48 +85,37 @@ def _build_put_statement(
     final_statement = f"PUT {local_path} {target_path} {parallel_str} {compress_str} {source_compression_str} {overwrite_str}"
     return final_statement
 
+def guard_connection(func):
```
sfc-gh-jdu, line 88: can we also combine this function with translate_connector_exception, and we just use one? I saw you have moved 
```
except ReauthenticationRequest as ex:
                    raise SnowparkClientExceptionMessages.SERVER_SESSION_EXPIRED(
                        ex.cause
                    )
``` 
into translate_connector_exception?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-lspiegelberg: ok weird, I changed the indent back manually.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-lspiegelberg: it's separated for separation of concerns. There are functions where no connector exceptions are caused, but a check on whether the connection is valid is enforced.

PR #648, opened 2023-01-07
SNOW-703369 Support DataFrame.dtypes

sfc-gh-sfan:
Description

Add dtypes as a property of DataFrame, which returns datatype with names in pandas Series.

This change referenced mobilize's implementation in https://github.com/MobilizeNet/snowpark-extensions-py/blob/main/tests/test_dataframe_extensions.py with modifications.

Testing

Integration test

   Fixes SNOW-703369

   Add dtypes as a property of DataFrame, which returns datatype with names in pandas Series.

   Note that the implementation is not exactly the same as mobilize, for example, this change does not use numpy array.

### DIFF
--- a/src/snowflake/snowpark/_internal/type_utils.py
+++ b/src/snowflake/snowpark/_internal/type_utils.py
@@ -458,39 +458,41 @@ def python_type_to_snow_type(tp: Union[str, Type]) -> Tuple[DataType, bool]:
     raise TypeError(f"invalid type {tp}")
 
 
-def data_type_to_python_type(dtype: DataType) -> Type:
-    if isinstance(dtype, ArrayType):
-        return list
-    elif isinstance(dtype, BinaryType):
-        return bytes
-    elif isinstance(dtype, BooleanType):
-        return bool
-    elif isinstance(dtype, DateType):
-        return datetime.date
-    elif isinstance(dtype, DecimalType):
-        return int if dtype.scale == 0 else decimal.Decimal
-    elif isinstance(dtype, FloatType):
-        return float
-    elif isinstance(dtype, DoubleType):
-        return float
-    elif isinstance(dtype, GeographyType):
-        return dict
-    elif isinstance(dtype, LongType):
-        return int
-    elif isinstance(dtype, MapType):
-        return dict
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -458,6 +458,41 @@ def python_type_to_snow_type(tp: Union[str, Type]) -> Tuple[DataType, bool]:
     raise TypeError(f"invalid type {tp}")
 
 
+def data_type_to_python_type(dtype: DataType) -> Type:
```
sfc-gh-jdu, line 461: We already have https://github.com/snowflakedb/snowpark-python/blob/4bc70637f4486a67b34036ccd0b8d523db837225/src/snowflake/snowpark/_internal/type_utils.py#L386, maybe align their names?

src/snowflake/snowpark/dataframe.py
```diff
@@ -3401,6 +3402,15 @@ def schema(self) -> StructType:
         """
         return StructType._from_attributes(self._plan.attributes)
 
+    @cached_property
+    def dtypes(self) -> "pandas.Series":
```
sfc-gh-jdu, line 3406: actually I feel moblize's implementation is wrong, we should return a list of tuples containing the column name and type string https://spark.apache.org/docs/3.1.2/api/python/reference/api/pyspark.sql.DataFrame.dtypes.html#pyspark.sql.DataFrame.dtypes

sfc-gh-sfan: Makes sense. Will change to that.

sfc-gh-sfan:
There are three types that do not exist in pyspark: `TimeType`, `GeographyType` and `VariantType`. We chose to directly use the name from class name by removing `Type` and lowercase the remaining.

Also added a unit test because integration test cannot cover all paths. For example, we automatically convert integer types to `LongType()`. Maybe we cannot hit the code path of other integer types and float type, but regardless we have test coverage for these code paths.

sfc-gh-jdu (approved):

sfc-gh-stan (approved):

PR #654, opened 2023-01-13
SNOW-701257: Remove error log

sfc-gh-yixie:

   Fixes #SNOW-701257

   Remove the code that logs en exception as an error.

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -307,30 +307,23 @@ class ServerConnection:
         ] = None,  # this argument is currently only used by AsyncJob
         **kwargs,
     ) -> Union[Dict[str, Any], AsyncJob]:
-        try:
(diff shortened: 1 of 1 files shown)


sfc-gh-sfan (commented):
Do we still have the log of failed query + exception after this change?

sfc-gh-yixie:
> Do we still have the log of failed query + exception after this change?

Good question. I'm thinking of adding the problematic SQL to SnowparkSQLException so people can get the SQL from the exception. Simply removing the error log will make it harder to support because users only see errors but don't see the SQL.

sfc-gh-sfan (approved):
Yeah that would be ideal. Having access to the query text would be helpful in case query id is missing for whatever reason. Do you want to include it in this PR or create a JIRA to follow up on?

The change itself looks good and given the exception has the query id, most of the time we should have enough info for debugging.

sfc-gh-jdu:
Should we have a parameter to control this? I feel logging error is still useful when running tests, so it can be default to false

PR #669, opened 2023-02-03
SNOW-719900 Remove random alias for subqeury in joins

sfc-gh-stan:

   Fixes #SNOW-719900

  Previously, dataframe/table joins will use random aliases for the left and right subquery results at query generation, which means users cannot take advantage of query [result cache](https://community.snowflake.com/s/article/Caching-in-Snowflake-Data-Warehouse). This PR removes that and adds a unit test.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -574,6 +574,7 @@ class Analyzer:
                 logical_plan.join_type,
                 self.analyze(logical_plan.condition) if logical_plan.condition else "",
                 logical_plan,
+                self.session.use_constant_subquery_alias,
             )
 
         if isinstance(logical_plan, Sort):
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -23,10 +23,8 @@ from snowflake.snowpark._internal.analyzer.expression import Attribute
 from snowflake.snowpark._internal.error_message import SnowparkClientExceptionMessages
 from snowflake.snowpark._internal.type_utils import convert_sp_to_sf_type
 from snowflake.snowpark._internal.utils import (
-    TempObjectType,
     get_temp_type_for_object,
     is_single_quoted,
-    random_name_for_temp_object,
 )
 from snowflake.snowpark.row import Row
 from snowflake.snowpark.types import DataType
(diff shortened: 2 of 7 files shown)


tests/unit/test_dataframe.py
```diff
@@ -180,3 +180,22 @@ def test_create_or_replace_temp_view_bad_input():
         "The input of create_or_replace_temp_view() can only a str or list of strs."
         in str(exc_info)
     )
+
+
+def test_same_joins_should_generate_same_queries():
```
sfc-gh-jdu, line 185: You can just run it as an integration test? Also, could you check whether q1.join(q2) works?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -479,8 +477,8 @@ def set_operator_statement(left: str, right: str, operator: str) -> str:
 def left_semi_or_anti_join_statement(
     left: str, right: str, join_type: JoinType, condition: str
 ) -> str:
-    left_alias = random_name_for_temp_object(TempObjectType.TABLE)
-    right_alias = random_name_for_temp_object(TempObjectType.TABLE)
+    left_alias = "SNOWPARK_TEMP_LEFT_RESULT"
```
sfc-gh-sfan, line 480: Do we need to worry about potential ambiguous alias if multiple joins are involved?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -517,8 +515,8 @@ def left_semi_or_anti_join_statement(
 def snowflake_supported_join_statement(
     left: str, right: str, join_type: JoinType, condition: str
 ) -> str:
-    left_alias = random_name_for_temp_object(TempObjectType.TABLE)
-    right_alias = random_name_for_temp_object(TempObjectType.TABLE)
+    left_alias = "SNOWPARK_TEMP_LEFT_RESULT"
```
sfc-gh-yixie, line 518: Some queries may have many repeated query aliases if we use a fixed one. For instance, in the following code df7's query will have two SNOWPARK_TEMP_LEFT_RESULT as subquery aliases. How about removing the alias if we use this way.
```
df5 = df1 .join(df2)
df6 = df3.join(df4)
df7 = df5.join(df6)
```

sfc-gh-stan: That's a great point, why do we need these subquery aliases? I don't see them being used or saved separately from the query string 🤔

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -161,23 +161,29 @@ def analyze(
 
         if isinstance(expr, Like):
             return like_expression(
-                self.analyze(expr.expr, parse_local_name), self.analyze(expr.pattern)
+                self.analyze(expr.expr, parse_local_name),
+                self.analyze(expr.pattern, parse_local_name),
```
sfc-gh-yixie, line 165: This (and other changes in this PR) is already in the main branch. I'm wondering how it's shown as a change in this PR.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-stan: I don't have a strong opinion here but I think a longer name with `SNOWPARK` in it is likely to be duplicated.

sfc-gh-sfan (commented):
Can we have a test of nested joins, e.g. df1.join(df2).join(df3)?

sfc-gh-sfan (approved):

PR #681, opened 2023-02-16
SNOW-742340 Add python 3.9 integ test merge gate 

sfc-gh-sfan:
Description

Add python 3.9 integ test non-required merge gate. These tests
explicitly skip udf test because 3.9 is not supported for UDFs yet

Testing

n/a

   Fixes SNOW-742340

   Add python 3.9 integ test non-required merge gate. These tests explicitly skip udf test because 3.9 is not supported for UDFs yet

### DIFF
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -60,7 +60,7 @@ jobs:
             download_name: macos
           - image_name: windows-latest
             download_name: windows
-        python-version: [3.8]
+        python-version: [3.8, 3.9]
         cloud-provider: [aws, azure, gcp]
     steps:
       - name: Checkout Code
@@ -89,7 +89,7 @@ jobs:
         run: python -m pip install -U setuptools pip wheel
       - name: Install tox
         run: python -m pip install tox
-      - if: ${{ contains('macos', matrix.os.download_name) }}
+      - if: ${{ matrix.python-version == '3.8' && contains('macos', matrix.os.download_name) }}
         name: Run tests (including doctest Tests)
         run: python -m tox -e "py${PYTHON_VERSION/\./}-udf"
         env:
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -11,6 +11,9 @@
 SRC_DIR = os.path.join(THIS_DIR, "src")
 SNOWPARK_SRC_DIR = os.path.join(SRC_DIR, "snowflake", "snowpark")
 CONNECTOR_DEPENDENCY_VERSION = ">=2.7.12, <4.0.0"
+REQUIRED_PYTHON_VERSION = "==3.8.*"
+if os.getenv("SNOWFLAKE_IS_PYTHON_RUNTIME_TEST", False):
+    REQUIRED_PYTHON_VERSION = ">=3.8"
```
sfc-gh-jdu, line 16: should we do `>=3.8, <=3.9`

sfc-gh-jdu (approved):

setup.py
sfc-gh-sfan: I was debating about adding `<3.10` but then thought that `SNOWFLAKE_IS_PYTHON_RUNTIME_TEST` indicates that we are testing unsupported python versions, and we probably don't need to limit the version.

This way we don't need to change this `REQUIRED_PYTHON_VERSION` again for `3.10` testing. What do you think?

sfc-gh-yixie: I prefer adding `<3.10` because it tells users that you can't use 3.10. Or else somebody will wrongly install snowpark with python 3.10 successfully then found it didn't work with UDF.

.github/workflows/precommit.yml
```diff
@@ -60,7 +60,7 @@ jobs:
             download_name: macos
           - image_name: windows-latest
             download_name: windows
-        python-version: [3.8]
+        python-version: [3.8, 3.9]
```
sfc-gh-yixie, line 63: Does this run all tests in both 3.8 and 3.9?

setup.py
sfc-gh-sfan: Hmm but `SNOWFLAKE_IS_PYTHON_RUNTIME_TEST` is for testing only, and it also does not work with 3.9, right?

.github/workflows/precommit.yml
sfc-gh-yixie: Ah, I saw every test has indicated a python version.

sfc-gh-sfan: We don't run udf and doc test for python 3.9. You could find more details in the steps below.

sfc-gh-yixie (approved):

sfc-gh-sfan:
I'll add py39 test for FIPS and no-simplifier case in a separate PR early next week. This `SNOWFLAKE_IS_PYTHON_RUNTIME_TEST` unblocks some of my other work :)

PR #683, opened 2023-02-16
SNOW-693539: Allow case insensitive access to fields on Row

### LINKED ISSUE #618: SNOW-693539: Allow case insensitive access to fields on a row

Employee=Row('FIRSTNAME', 'LASTNAME', 'GENDER', 'SALARY')
employee1=Employee('Robert', 'Williams', 'M', 62)
print(employee1.firstname)

This will throw an exception indicating that the field firstname is not present

Allow to use the attributes in case insensitive way

it will make it easier in scenarios where you deal with rows.

for example is common to have methods like

def processRows(l):
      for m in l:
            m.firstName + m.lastName

in spark probably can do:

rows = df.collect()
processRows(rows)

but in SF we will need to adjust to method to match the fieldname

## References, Other Background

the code in the Row class already has a __getattr__

So we can do something like:

Employee=Row('FIRSTNAME', 'LASTNAME', 'GENDER', 'SALARY',Row.CASE_INSENSITIVE)
employee1=Employee('Robert', 'Williams', 'M', 62)
print(employee1.firstname)

A test modification for Row can be:

```python
class Row(tuple):
    """Represents a row in :class:`DataFrame`.

    It is immutable and works like a tuple or a named tuple.

    >>> row = Row(1, 2)
    >>> row
    Row(1, 2)
    >>> row[0]
    1
    >>> len(row)
    2
    >>> row[0:1]
    Row(1)
    >>> named_row = Row(name1=1, name2=2)
    >>> named_row
    Row(name1=1, name2=2)
    >>> named_row["name1"]
    1
    >>> named_row.name1
    1
    >>> row == named_row
    True

    A ``Row`` object is callable. You can use it to create other ``Row`` objects:

    >>> Employee = Row("name", "salary")
    >>> emp1 = Employee("John", 10000)
    >>> emp1
    Row(name='John', salary=10000)
    >>> emp2 = Employee("James", 20000)
    >>> emp2
    Row(name='James', salary=20000)

    """
    
    def __new__(cls, *values: Any, **named_values: Any):
        if values and named_values:
            raise ValueError("Either values or named_values is required but not both.")
        if named_values:
            # After py3.7, dict is ordered(not sorted) by item insertion sequence.
            # If we support 3.6 or older someday, this implementation needs changing.
            row = tuple.__new__(cls, tuple(named_values.values()))
            row.__dict__["_case_insensitive"] = False
            row.__dict__["_named_values"] = named_values
            row.__dict__["_fields"] = tuple(named_values.keys())
        else:

            is_ci = values[-1]==Row.CASE_INSENSITIVE
            if is_ci:
               values = values[:-1]
            row = tuple.__new__(cls, values)
            row.__dict__["_case_insensitive"] = is_ci
            row.__dict__["_named_values"] = None
            row.__dict__["_fields"] = None

        # _fields is for internal use only. Users shouldn't set this attribute.
        # It contains a list of str representing column names. It also allows duplicates.
        # snowflake DB can return duplicate column names, for instance, "select a, a from a_table."
        # When return a DataFrame from a sql, duplicate column names can happen.
        # But using duplicate column names is obviously a bad practice even though we allow it.
        # It's value is assigned in __setattr__ if internal code assign value explicitly.
        row.__dict__["_has_duplicates"] = None
        return row

    def __getitem__(self, item: Union[int, str, slice]):
        if isinstance(item, int):
            return super().__getitem__(item)
        elif isinstance(item, slice):
            return Row(*super().__getitem__(item))
        else:  # str
            self._populate_named_values_from_fields()
            # get from _named_values first
            if self._named_values:
                return self._named_values[item]
            # we have duplicated fields and _named_values is not populated,
            # so indexing fields
            elif self._fields and self._check_if_having_duplicates():
                try:
                    index = self._fields.index(item.upper())  # may throw ValueError
                    return super().__getitem__(index)  # may throw IndexError
                except (IndexError, ValueError):
                    raise KeyError(item)
            # no column names/keys/fields
            else:
                raise KeyError(item)

    def __setitem__(self, key, value):
        raise TypeError("Row object does not support item assignment")

    def __getattr__(self, item):
        if self._case_insensitive:
            item = item.upper()
        self._populate_named_values_from_fields()
        if self._named_values and item in self._named_values:
            return self._named_values[item]
        elif self._fields and self._check_if_having_duplicates():
            try:
                index = self._fields.index(item.upper())  # may throw ValueError
                return self[index]  # may throw IndexError
            except (IndexError, ValueError):
                raise AttributeError(f"Row object has no attribute {item}")
        else:
            raise AttributeError(f"Row object has no attribute {item}")

    def __setattr__(self, key, value):
        if key != "_fields":
            raise AttributeError("Can't set attribute to Row object")
        if value is not None:
            self.__dict__["_fields"] = value

    def __contains__(self, item):
        self._populate_named_values_from_fields()
        if self._named_values:
            return item in self._named_values
        elif self._fields:
            return item in self._fields
        else:
            return super().__contains__(item)

    def __call__(self, *args, **kwargs):
        """Create a new Row from current row."""
        if args and kwargs:
            raise ValueError(
                "A Row instance should have either values or field-value pairs but not both."
            )
        elif args and len(args) != len(self):
            raise ValueError(f"{len(self)} values are expected.")
        self._populate_named_values_from_fields()
        if self._named_values:
            if args:
                raise ValueError(
                    "The Row object can't be called with a list of values"
                    "because it already has fields and values."
                )
            new_row = Row(**self._named_values)
            for input_key, input_value in kwargs.items():
                if input_key not in self._named_values:
                    raise ValueError(f"Wrong keyword argument {input_key} for {self}")
                new_row._named_values[input_key] = input_value
            return new_row
        elif self._fields and self._check_if_having_duplicates():
            raise ValueError(
                "The Row object can't be called because it has duplicate fields"
            )
        else:
            if kwargs:
                raise ValueError(
                    "The Row object can't be called with field-value pairs "
                    "because it doesn't have field-value pairs"
                )
            if len(args) != len(self) or any(
                not isinstance(value, str) for value in self
            ):
                raise ValueError(
                    "The called Row object and input values must have the same size and the called Row object shouldn't have any non-str fields."
                )
            new_row= Row(**{k: v for k, v in zip(self, args)})
            new_row.__dict__["_case_insensitive"] = self._case_insensitive
            return new_row

    def __copy__(self):
        return _restore_row_from_pickle(self, self._named_values, self._fields)

    def __repr__(self):
        if self._fields:
            return "Row({})".format(
                ", ".join(f"{k}={v!r}" for k, v in zip(self._fields, self))
            )
        elif self._named_values:  # pragma: no cover
            # this might not be reachable because when there is self._named_values, there is always self._fields
            # Need to review later.
            return "Row({})".format(
                ", ".join(f"{k}={v!r}" for k, v in self._named_values.items())
            )

        else:
            return "Row({})".format(", ".join(f"{v!r}" for v in self))

    def __reduce__(self):
        return (
            _restore_row_from_pickle,
            (tuple(self), self._named_values, self._fields),
        )

    def as_dict(self, recursive: bool = False) -> Dict:
        """Convert to a dict if this row object has both keys and values.

        Args:
            recursive: Recursively convert child :class:`Row` objects to dicts. Default is False.

        >>> row = Row(name1=1, name2=2, name3=Row(childname=3))
        >>> row.as_dict()
        {'name1': 1, 'name2': 2, 'name3': Row(childname=3)}
        >>> row.as_dict(True)
        {'name1': 1, 'name2': 2, 'name3': {'childname': 3}}
        """
        self._populate_named_values_from_fields()
        if not self._named_values:
            raise TypeError(
                "Cannot convert a Row without key values or duplicated keys to a dict."
            )
        if not recursive:
            return dict(self._named_values)
        return self._convert_dict(self._named_values)

    def _convert_dict(
        self, obj: Union["Row", Dict, Iterable[Union["Row", Dict]]]
    ) -> Union[Dict, Iterable[Dict]]:
        if isinstance(obj, Row):
            return obj.as_dict(True)
        elif isinstance(obj, dict):
            child_dict = {}
            for k, v in obj.items():
                child_dict[k] = self._convert_dict(v)
            return child_dict
        elif isinstance(obj, Iterable) and not isinstance(obj, (str, bytes, bytearray)):
            return [self._convert_dict(x) for x in obj]

        return obj

    def _populate_named_values_from_fields(self) -> None:
        # populate _named_values dict if we have unduplicated fields
        if (
            self._named_values is None
            and self._fields
            and not self._check_if_having_duplicates()
        ):
            self.__dict__["_named_values"] = {k: v for k, v in zip(self._fields, self)}

    def _check_if_having_duplicates(self) -> bool:
        if self._has_duplicates is None:
            # Usually we don't have duplicate keys
            self.__dict__["_has_duplicates"] = bool(
                len(set(self._fields)) != len(self._fields)
            )
        return self._has_duplicates

    # Add aliases for user code migration
    asDict = as_dict

Row.CASE_INSENSITIVE = "@@@@@@CS@@@@@@"
```

sfc-gh-aalam:

   Fixes #618 

   `_internal/utils.py` and `row.py` are the two files where the interesting changes are present. Rest of the files are just propagating parameters and tests.

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -313,7 +313,7 @@ class ServerConnection:
         async_job_plan: Optional[
             SnowflakePlan
         ] = None,  # this argument is currently only used by AsyncJob
-        case_insensitive: bool = False,
+        case_sensitive: bool = True,
         **kwargs,
     ) -> Union[Dict[str, Any], AsyncJob]:
         try:
@@ -357,7 +357,7 @@ class ServerConnection:
                 async_job_plan.session,
                 data_type,
                 async_job_plan.post_actions,
-                case_insensitive=case_insensitive,
+                case_sensitive=case_sensitive,
                 **kwargs,
             )
 
@@ -405,7 +405,7 @@ class ServerConnection:
         to_iter: bool = False,
         block: bool = True,
         data_type: _AsyncResultType = _AsyncResultType.ROW,
-        case_insensitive: bool = False,
+        case_sensitive: bool = True,
         **kwargs,
     ) -> Union[
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
@@ -421,7 +421,7 @@ class ServerConnection:
             **kwargs,
             block=block,
             data_type=data_type,
-            case_insensitive=case_insensitive,
+            case_sensitive=case_sensitive,
         )
         if not block:
             return result_set
@@ -430,11 +430,11 @@ class ServerConnection:
         else:
             if to_iter:
                 return result_set_to_iter(
-                    result_set["data"], result_meta, case_insensitive=case_insensitive
+                    result_set["data"], result_meta, case_sensitive=case_sensitive
                 )
             else:
                 return result_set_to_rows(
-                    result_set["data"], result_meta, case_insensitive=case_insensitive
+                    result_set["data"], result_meta, case_sensitive=case_sensitive
                 )
 
     @SnowflakePlan.Decorator.wrap_exception
@@ -445,7 +445,7 @@ class ServerConnection:
         to_iter: bool = False,
         block: bool = True,
         data_type: _AsyncResultType = _AsyncResultType.ROW,
-        case_insensitive: bool = False,
+        case_sensitive: bool = True,
         **kwargs,
     ) -> Tuple[
         Dict[
@@ -496,7 +496,7 @@ $$"""
                     block=block,
                     data_type=data_type,
                     async_job_plan=plan,
-                    case_insensitive=case_insensitive,
+                    case_sensitive=case_sensitive,
                     **kwargs,
                 )
 
@@ -522,7 +522,7 @@ $$"""
                             block=not is_last,
                             data_type=data_type,
                             async_job_plan=plan,
-                            case_insensitive=case_insensitive,
+                            case_sensitive=case_sensitive,
                             **kwargs,
                         )
                         placeholders[query.query_id_place_holder] = (
@@ -539,7 +539,7 @@ $$"""
                         action.sql,
                         is_ddl_on_temp_object=action.is_ddl_on_temp_object,
                         block=block,
-                        case_insensitive=case_insensitive,
+                        case_sensitive=case_sensitive,
                         **kwargs,
                     )
 
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -149,7 +149,9 @@ class ServerConnection:
         if PARAM_APPLICATION not in self._lower_case_parameters:
             # Mirrored from snowflake-connector-python/src/snowflake/connector/connection.py#L295
             if ENV_VAR_PARTNER in os.environ.keys():
-                self._lower_case_parameters[PARAM_APPLICATION] = os.environ[ENV_VAR_PARTNER]
+                self._lower_case_parameters[PARAM_APPLICATION] = os.environ[
+                    ENV_VAR_PARTNER
+                ]
             elif "streamlit" in sys.modules:
                 self._lower_case_parameters[PARAM_APPLICATION] = "streamlit"
             else:
@@ -311,6 +313,7 @@ class ServerConnection:
         async_job_plan: Optional[
             SnowflakePlan
         ] = None,  # this argument is currently only used by AsyncJob
+        case_insensitive: bool = False,
         **kwargs,
     ) -> Union[Dict[str, Any], AsyncJob]:
         try:
@@ -354,6 +357,7 @@ class ServerConnection:
                 async_job_plan.session,
                 data_type,
                 async_job_plan.post_actions,
+                case_insensitive=case_insensitive,
                 **kwargs,
             )
 
@@ -401,6 +405,7 @@ class ServerConnection:
         to_iter: bool = False,
         block: bool = True,
         data_type: _AsyncResultType = _AsyncResultType.ROW,
+        case_insensitive: bool = False,
         **kwargs,
     ) -> Union[
         List[Row], "pandas.DataFrame", Iterator[Row], Iterator["pandas.DataFrame"]
@@ -410,7 +415,13 @@ class ServerConnection:
                 "Async query is not supported in stored procedure yet"
             )
         result_set, result_meta = self.get_result_set(
-            plan, to_pandas, to_iter, **kwargs, block=block, data_type=data_type
+            plan,
+            to_pandas,
+            to_iter,
+            **kwargs,
+            block=block,
+            data_type=data_type,
+            case_insensitive=case_insensitive,
         )
         if not block:
             return result_set
@@ -418,9 +429,13 @@ class ServerConnection:
             return result_set["data"]
         else:
             if to_iter:
-                return result_set_to_iter(result_set["data"], result_meta)
+                return result_set_to_iter(
+                    result_set["data"], result_meta, case_insensitive=case_insensitive
+                )
             else:
-                return result_set_to_rows(result_set["data"], result_meta)
+                return result_set_to_rows(
+                    result_set["data"], result_meta, case_insensitive=case_insensitive
+                )
 
     @SnowflakePlan.Decorator.wrap_exception
     def get_result_set(
@@ -430,6 +445,7 @@ class ServerConnection:
         to_iter: bool = False,
         block: bool = True,
         data_type: _AsyncResultType = _AsyncResultType.ROW,
+        case_insensitive: bool = False,
         **kwargs,
     ) -> Tuple[
         Dict[
@@ -480,6 +496,7 @@ $$"""
                     block=block,
                     data_type=data_type,
                     async_job_plan=plan,
+                    case_insensitive=case_insensitive,
                     **kwargs,
                 )
 
@@ -505,6 +522,7 @@ $$"""
                             block=not is_last,
                             data_type=data_type,
                             async_job_plan=plan,
+                            case_insensitive=case_insensitive,
                             **kwargs,
                         )
                         placeholders[query.query_id_place_holder] = (
@@ -521,6 +539,7 @@ $$"""
                         action.sql,
                         is_ddl_on_temp_object=action.is_ddl_on_temp_object,
                         block=block,
+                        case_insensitive=case_insensitive,
                         **kwargs,
                     )
 
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -517,19 +517,31 @@ class DataFrame:
 
     @overload
     def collect(
-        self, *, statement_params: Optional[Dict[str, str]] = None, block: bool = True
+        self,
+        *,
+        statement_params: Optional[Dict[str, str]] = None,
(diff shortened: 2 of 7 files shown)


src/snowflake/snowpark/row.py
```diff
@@ -14,6 +14,17 @@ def _restore_row_from_pickle(values, named_values, fields):
     return row
 
 
+def is_already_quoted(value: str) -> bool:
```
sfc-gh-aalam, line 17: there is a similar check present in `udf_utils.py` but importing it created circular import errors. is there a better way to do this?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -430,6 +445,7 @@ def get_result_set(
         to_iter: bool = False,
         block: bool = True,
         data_type: _AsyncResultType = _AsyncResultType.ROW,
+        case_insensitive: bool = False,
```
sfc-gh-sfan, line 448: I wonder if we should name it `case_sensitive` and flip the default value? This way we don't need to reason through two negations.

src/snowflake/snowpark/row.py
sfc-gh-sfan: Maybe we could extract it into `util.py`?

src/snowflake/snowpark/row.py
```diff
@@ -48,19 +59,42 @@ class Row(tuple):
...
+    >>> emp1 = Employee("John", 10000)
+    >>> emp1.firstName
+    'John'
+    >>> emp2 = Employee("James", 20000)
+    >>> emp2.Salary
+    20000
+
     """
 
+    CASE_INSENSITIVE = "@@@@@@_CASE_INSENSITIVE_@@@@@@"
```
sfc-gh-sfan, line 74: Why do we choose to use a constant that might conflict with user data, not a keyword argument?

sfc-gh-aalam: A new `Row` object can be created by either giving `args` or `named_args` but not both. We have a check that throws `ValueError` if both are provided. I can work on this logic to only accept `args` and `named_args` when `len(named_args) == 1` and `"case_sensitive" in named_args`.

tests/integ/test_dataframe.py
```diff
@@ -1699,6 +1699,62 @@ def test_dataframe_duplicated_column_names(session):
...
+    assert row["firstname"] == "Gordon"
+    assert row["FIRSTNAME"] == "Gordon"
+    assert row["FirstName"] == "Gordon"
+
+    assert row.matches_won == 153
+    assert row.MATCHES_WON == 153
+    assert row.MaTchEs_WoN == 153
+    assert row["matches_won"] == 153
+    assert row["Matches-Won"] == 153
+    assert row["MATCHES_WON"] == 153
```
sfc-gh-aalam, line 1724: how does the team feel about this change? I am not only using lower casing to implement this but I am also replacing `-` with `_`

src/snowflake/snowpark/row.py
sfc-gh-sfan: One caveat is that we will no longer accept `case_sensitive` as an input data when user tries to create a `Row` using `kwargs`. Not sure how much we need to worry about this, but might be a legit concern.

Another way is to create a `Factory` class that can create both case sensitive and insensitive rows, and we could make them share as much code as possible.

tests/integ/test_dataframe.py
sfc-gh-sfan: Does case insensitive implies `"-" == "_"`? If not then I think we probably should not do this.

src/snowflake/snowpark/row.py
sfc-gh-aalam: Not really. If you check the implementation, we accept `args` and `kwargs` only when `len(kwargs) == 1` and `"case_sensitive" in kwargs`. Otherwise we throw error as before.

Let me checkout the `Factory` approach. Sounds much cleaner to me.

tests/integ/test_dataframe.py
sfc-gh-aalam: I earlier thought it could be a good idea but now I'm pretty sure it's gonna break some users and make them mad.

src/snowflake/snowpark/row.py
sfc-gh-sfan: For the normal `args` case it's okay. But if user provides pure `kwargs` and indeed need data with keyword `case_sensitive`, then they are kinda screwed, right?

sfc-gh-aalam: aah. I see. You are right. In current implementation, when user is passing `kwargs`, we always fix fields to be case sensitive but this is not documented so can confuse the users.

src/snowflake/snowpark/row.py
```diff
@@ -48,19 +58,45 @@ class Row(tuple):
...
+    20000
+
     """
 
     def __new__(cls, *values: Any, **named_values: Any):
+        is_case_sensitive = True
         if values and named_values:
-            raise ValueError("Either values or named_values is required but not both.")
+            if len(named_values) == 1 and "case_sensitive" in named_values:
+                is_case_sensitive = named_values["case_sensitive"]
```
sfc-gh-jdu, line 77: what if case_sensitive is a column name? I guess you can make case_sensitive as an optional argument explicitly in `__new__`?

src/snowflake/snowpark/dataframe.py
```diff
@@ -517,19 +517,31 @@ def stat(self) -> DataFrameStatFunctions:
...
         ...  # pragma: no cover
 
     @df_collect_api_telemetry
     def collect(
-        self, *, statement_params: Optional[Dict[str, str]] = None, block: bool = True
+        self,
+        *,
+        statement_params: Optional[Dict[str, str]] = None,
+        block: bool = True,
+        case_sensitive: bool = True,
```
sfc-gh-jdu, line 544: docstring for this argument

tests/integ/test_dataframe.py
```diff
@@ -1699,6 +1699,62 @@ def test_dataframe_duplicated_column_names(session):
...
+    assert row["Matches_Won"] == 153
+    assert row["MATCHES_WON"] == 153
+
+    async_job = df_quote.collect_nowait(case_sensitive=False)
+    with pytest.raises(
+        ValueError,
+        match="Case insensitive fields is not supported in presence of quoted columns",
+    ):
+        row = async_job.result()[0]
+
```
sfc-gh-jdu, line 1756: maybe add some tests for special characters?

src/snowflake/snowpark/row.py
sfc-gh-aalam: Consider the case when we create a row like so: `lease_doc = Row(author="Name", legal=True, case_sensitive=True)`. From our side, it is not possible to decipher whether `case_sensitive` is a column in the row or if it is an argument

sfc-gh-sfan: I still think we should consider providing a factory/builder. This factory can explicitly set a row to be case sensitive or not, and also directly feed the preprocessed value to `Row` object. For example:

```
create_case_insensitive_row(original_args):
    values = [canonicalize_field(value) for value in values]
    row = Row(original_args)
    row.set_case_sensitive(False)
```

or 

```
builder = RowBuilder(args)
builder.set_case_sensitive(False)
row = builder.to_row()
```

I think we probably still should try to avoid making a specific string impossible to be a column name because of implementation.

sfc-gh-sfan (approved):

PR #691, opened 2023-02-24
SNOW-747658 Add session.conf to store runtime configuration

sfc-gh-stan:

   Fixes #SNOW-747658

`Session.conf` lets users to lookup runtime config parameters. Previously, options passed into configs are not preserved for future lookup. E.g. users who want to make sure paramstyle is pyformat cannot easily confirm this. With this change, users can easily access these configs with `session.conf`.

Reference: https://spark.apache.org/docs/3.1.3/api/python/reference/api/pyspark.sql.SparkSession.conf.html#pyspark.sql.SparkSession.conf

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -136,7 +136,8 @@ class ServerConnection:
         self._lower_case_parameters = {k.lower(): v for k, v in options.items()}
         self._add_application_name()
         self._conn = conn if conn else connect(**self._lower_case_parameters)
-        self._lower_case_parameters["password"] = None
+        if "password" in self._lower_case_parameters:
+            self._lower_case_parameters["password"] = None
         self._cursor = self._conn.cursor()
         self._telemetry_client = TelemetryClient(self._conn)
         self._query_listener: Set[QueryHistory] = set()
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -136,8 +136,7 @@ class ServerConnection:
         self._lower_case_parameters = {k.lower(): v for k, v in options.items()}
         self._add_application_name()
         self._conn = conn if conn else connect(**self._lower_case_parameters)
-        if "password" in self._lower_case_parameters:
-            self._lower_case_parameters["password"] = None
+        self._lower_case_parameters.pop("password", None)
         self._cursor = self._conn.cursor()
         self._telemetry_client = TelemetryClient(self._conn)
         self._query_listener: Set[QueryHistory] = set()
@@ -149,7 +148,9 @@ class ServerConnection:
         if PARAM_APPLICATION not in self._lower_case_parameters:
             # Mirrored from snowflake-connector-python/src/snowflake/connector/connection.py#L295
             if ENV_VAR_PARTNER in os.environ.keys():
-                self._lower_case_parameters[PARAM_APPLICATION] = os.environ[ENV_VAR_PARTNER]
+                self._lower_case_parameters[PARAM_APPLICATION] = os.environ[
+                    ENV_VAR_PARTNER
+                ]
             elif "streamlit" in sys.modules:
                 self._lower_case_parameters[PARAM_APPLICATION] = "streamlit"
             else:
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -136,7 +136,7 @@ class ServerConnection:
         self._lower_case_parameters = {k.lower(): v for k, v in options.items()}
         self._add_application_name()
         self._conn = conn if conn else connect(**self._lower_case_parameters)
-        self._lower_case_parameters.pop("password", None)
+        self._lower_case_parameters["password"] = None
         self._cursor = self._conn.cursor()
         self._telemetry_client = TelemetryClient(self._conn)
         self._query_listener: Set[QueryHistory] = set()
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -283,7 +283,7 @@ class Session:
     builder: SessionBuilder = SessionBuilder()
 
     def __init__(
-        self, conn: ServerConnection, options: Optional[Dict[str, Any]]
+        self, conn: ServerConnection, options: Optional[Dict[str, Any]] = None
     ) -> None:
         if len(_active_sessions) >= 1 and is_in_stored_procedure():
             raise SnowparkClientExceptionMessages.DONT_CREATE_SESSION_IN_SP()
@@ -318,7 +318,7 @@ class Session:
         self._sql_simplifier_enabled: bool = self._get_client_side_session_parameter(
             _PYTHON_SNOWPARK_USE_SQL_SIMPLIFIER_STRING, True
         )
-        self._conf = self.RuntimeConfig(self, options)
+        self._conf = self.RuntimeConfig(self, options or {})
         _logger.info("Snowpark Session information: %s", self._session_info)
 
     def __enter__(self):
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -195,7 +195,7 @@ class Session:
         def __init__(self, session: "Session", conf: Dict[str, Any]) -> None:
             self._session = session
             for key, val in conf.items():
-                if hasattr(Session, key):
+                if hasattr(Session, key) and self.is_mutable(key):
                     setattr(session, key, val)
 
         def get(self, key: str, default=None) -> Any:
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -194,25 +194,12 @@ class Session:
     class RuntimeConfig:
         def __init__(self) -> None:
             self._conf: Dict[str:Any] = {}
-            self._immutables: Set[str] = set()
-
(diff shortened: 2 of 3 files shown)


src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -136,8 +136,7 @@ def __init__(
         self._lower_case_parameters = {k.lower(): v for k, v in options.items()}
         self._add_application_name()
         self._conn = conn if conn else connect(**self._lower_case_parameters)
-        if "password" in self._lower_case_parameters:
-            self._lower_case_parameters["password"] = None
+        self._lower_case_parameters.pop("password", None)
```
sfc-gh-sfan, line 139: This is slightly different from the original code (set password to None vs remove password from dict). Probably okay just want to raise it.

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,16 @@ class Session:
     A :class:`Session` object is not thread-safe.
     """
 
+    class RuntimeConfig:
```
sfc-gh-sfan, line 194: I might miss something but it does not seem like we have a public method to add entries? Should `_conf` be populated in `__init__`?

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,16 @@ class Session:
...
+        def __init__(self) -> None:
+            self._conf: Dict[str:Any] = {}
+
+        def get(self, key: str, default=None):
+            return self._conf.get(key, default)
+
+        def unset(self, key: str):
```
sfc-gh-sfan, line 201: Is it correct that we don't support `set` because we want to make it immutable?

src/snowflake/snowpark/session.py
```diff
@@ -227,9 +237,14 @@ def configs(
...
-            if "connection" in self._options:
-                return self._create_internal(self._options["connection"])
-            return self._create_internal(conn=None)
+            session = self._create_internal(self._options.get("connection"))
+            for k, v in self._options.items():
+                if k not in (
+                    "password",
+                    "connection",
+                ):  # Parameters we choose not to expose in RuntimeConfig
+                    session.conf._conf[k] = v
```
sfc-gh-sfan, line 246: Why don't we initiate `conf` here directly with a dict that contains the exposed config?

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,16 @@ class Session:
     A :class:`Session` object is not thread-safe.
     """
 
+    class RuntimeConfig:
+        def __init__(self) -> None:
+            self._conf: Dict[str:Any] = {}
```
sfc-gh-jdu, line 196:
```suggestion
            self._conf: Dict[str, Any] = {}
```

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,16 @@ class Session:
...
+
+        def get(self, key: str, default=None):
```
sfc-gh-jdu, line 198: type hint

src/snowflake/snowpark/session.py
```diff
@@ -326,6 +340,10 @@ def close(self) -> None:
             finally:
                 _remove_session(self)
 
+    @property
+    def conf(self):
```
sfc-gh-jdu, line 344: type hint

src/snowflake/snowpark/session.py
sfc-gh-jdu: If we update `conf`, even using `unset`, should we also change https://github.com/snowflakedb/snowpark-python/blob/30530f8c644c40634d42cf9b95f7895dcbb4647f/src/snowflake/snowpark/_internal/server_connection.py#L136? I think we should make them the same any time

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-stan: True, I'll revert to previous behavior.

sfc-gh-stan:
For some reason I can't reply Jianzhun's last comment above. We can't keep these the same because there will be Snowpark Python specific parameters. In fact I only needed this change so that we could use parameter protection, which makes me wonder... maybe we should just make the parameter a property 🤔 but then it is nice if we can expose the connection parameters as runtime config so users don't need to do things like `session._conn._conn._paramstyle`.

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,38 @@ class Session:
     A :class:`Session` object is not thread-safe.
     """
 
+    class RuntimeConfig:
```
sfc-gh-stan, line 194: This class now doesn't use a dictionary to store config params but instead check whether the config is a `property` of `Session` or the underlying `SnowflakeConnection`. In this way, we avoid storing an extra copy of values and needing to keep the copies in sync.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -136,8 +136,7 @@ def __init__(
         self._lower_case_parameters = {k.lower(): v for k, v in options.items()}
         self._add_application_name()
         self._conn = conn if conn else connect(**self._lower_case_parameters)
-        if "password" in self._lower_case_parameters:
-            self._lower_case_parameters["password"] = None
+        self._lower_case_parameters["password"] = None
```
sfc-gh-sfan, line 139: Wonder if we want to revert this?

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,38 @@ class Session:
...
+                    setattr(session, key, val)
+
+        def get(self, key: str, default=None) -> Any:
+            if hasattr(Session, key):
+                return getattr(self._session, key)
+            if hasattr(self._session._conn._conn, key):
+                return getattr(self._session._conn._conn, key)
+            return default
+
+        def is_modifiable(self, key: str) -> bool:
```
sfc-gh-sfan, line 208: nit: maybe `is_mutable`? No strong opinion

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,38 @@ class Session:
...
+            if hasattr(Session, key) and isinstance(getattr(Session, key), property):
+                return getattr(Session, key).fset is not None
+            if hasattr(SnowflakeConnection, key) and isinstance(
+                getattr(SnowflakeConnection, key), property
+            ):
+                return getattr(SnowflakeConnection, key).fset is not None
```
sfc-gh-sfan, line 214: I'm curious: if `fset` is `None`, what happens if we try to do `setattr`? I guess it still works and that's why we need a check first?

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,38 @@ class Session:
...
+            return False
+
+        def set(self, key: str, value: Any) -> None:
+            if self.is_modifiable(key):
+                if hasattr(Session, key):
+                    setattr(self._session, key, value)
+                if hasattr(SnowflakeConnection, key):
+                    setattr(self._session._conn._conn, key, value)
```
sfc-gh-sfan, line 222: Is it expected that the `key` exists in both `Session` and `SnowflakeConnection`? Also, if neither class has the key, this becomes a no-op. Just want to make sure if this is the expected behavior.

src/snowflake/snowpark/session.py
sfc-gh-stan: `fset` is None means the `property` has no setter, this is for properties that `host` or `port` that is not meant to be changed. `setattr` on a `property` with no setter should generate `AttributeError: can't set attribute '<name>'`. In this way, we could distinguish mutable and immutable configs.

src/snowflake/snowpark/session.py
sfc-gh-stan: `telemetry_enabled` is the only property that exists in both `Session` and `SnowflakeConnection` for now. Since `is_modifiable` returns `true`, this key must exist in one of the classes.

src/snowflake/snowpark/session.py
```diff
@@ -191,6 +191,38 @@ class Session:
     A :class:`Session` object is not thread-safe.
     """
 
+    class RuntimeConfig:
+        def __init__(self, session: "Session", conf: Dict[str, Any]) -> None:
+            self._session = session
+            for key, val in conf.items():
+                if hasattr(Session, key):
+                    setattr(session, key, val)
```
sfc-gh-jdu, line 199: hmmm it should have side effect right? Now people can use session.account to get the account, so it creates new properties exposed to public users, but not documented.

sfc-gh-stan: `account` won't be added since it's not an existing attribute of `Session`, but I see how this could be dangerous: it could be used to overwrite any attributes of `Session`, I will change this to check `self.is_mutable` too.

src/snowflake/snowpark/session.py
sfc-gh-sfan: If `setattr` throws an error for immutable configs, can we skip this check and directly do `setattr` and let it raise `AttributeError` for the immutable case?

sfc-gh-stan: We could, but we still need to check for whether `key` is an existing property to make sure users can't abuse `setattr`. Is there a specific reason you want this behavior?

sfc-gh-jdu (approved):

src/snowflake/snowpark/session.py
sfc-gh-sfan: I see. I'm just trying to see if there is a possibility to reduce code complexity, because this mutability check is quite complicated :)

sfc-gh-sfan (approved):

PR #702, opened 2023-03-02
SNOW-753246 Add Python 3.10 merge gate

sfc-gh-sfan:
There are mainly two issues that need to be fixed to be 3.10 compatible:

- `Iterable` is removed from `typing`. In 3.10 we are supposed to use `collections.abc.Iterables`, but this class is not subscriptable in 3.8. To be backward compatible, we first try to import from `typing`, and fall back to import from `collections.abc` if failed.
- Nested types's `__name__` only has the parent type in 3.10. For example, `Optional[str].__name__` is `Optional`. This breaks some of our type tests in 3.10, which is fixed in this PR as well.

   Fixes SNOW-753246

   There are mainly two issues that need to be fixed to be 3.10 compatible:

- `Iterable` is removed from `typing`. In 3.10 we are supposed to use `collections.abc.Iterables`, but this class is not subscriptable in 3.8. To be backward compatible, we first try to import from `typing`, and fall back to import from `collections.abc` if failed.
- Nested types's `__name__` only has the parent type in 3.10. For example, `Optional[str].__name__` is `Optional`. This breaks some of our type tests in 3.10, which is fixed in this PR as well.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -4,7 +4,6 @@
 #
 import re
 import typing
-from collections.abc import Iterable
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 from snowflake.snowpark._internal.analyzer.binary_plan_node import (
@@ -31,6 +30,14 @@ from snowflake.snowpark._internal.utils import (
 )
 from snowflake.snowpark.row import Row
 from snowflake.snowpark.types import DataType
+
+# Python 3.8 needs to use typing.Iterable because collections.abc.Iterable is not subscriptable
+# Python 3.9 can use both
+# Python 3.10 needs to use collections.abc.Iterable because typing.Iterable is removed
+try:
+    from typing import Iterable
+except ImportError:
+    from collections.abc import Iterable
 
 LEFT_PARENTHESIS = "("
 RIGHT_PARENTHESIS = ")"
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -4,7 +4,8 @@
 #
 import re
 import typing
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
+from collections.abc import Iterable
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from snowflake.snowpark._internal.analyzer.binary_plan_node import (
     Except,
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
@@ -2,7 +2,6 @@
 #
 # Copyright (c) 2012-2022 Snowflake Computing Inc. All rights reserved.
 #
-from collections.abc import Iterable
 from enum import Enum
 from typing import Any, Dict, List, Optional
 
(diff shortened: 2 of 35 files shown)


src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -31,6 +31,14 @@
 from snowflake.snowpark.row import Row
 from snowflake.snowpark.types import DataType
 
+# Python 3.8 needs to use typing.Iterable because collections.abc.Iterable is not subscriptable
+# Python 3.9 can use both
+# Python 3.10 needs to use collections.abc.Iterable because typing.Iterable is removed
```
sfc-gh-stan, line 36: Python 4.0 🤣

tests/unit/test_types.py
```diff
@@ -297,11 +305,14 @@ def test_strip_unnecessary_quotes():
...
 def test_python_type_to_snow_type():
-    def check_type(python_type, snow_type, is_nullable):
+    # In python 3.10, the __name__ of nested type only contains the parent type, which breaks our test. And for this
+    # reason, we introduced type_str_override to test the expected string.
+    def check_type(python_type, snow_type, is_nullable, type_str_override=None):
         assert python_type_to_snow_type(python_type) == (snow_type, is_nullable)
-        assert python_type_to_snow_type(
-            getattr(python_type, "__name__", str(python_type))
-        ) == (snow_type, is_nullable)
+        type_str = type_str_override or getattr(
```
sfc-gh-stan, line 312: What if we just use `str(python_type)` instead of `__name__`? Python 3.10.9 on my machine evaluates:
`str(typing.Union[str, None])` to `'typing.Optional[str]'`, which is what we want right?

sfc-gh-stan:
Just curious: Why aren't any of 3.9 and 3.10 tests required merge gates?

tests/unit/test_types.py
sfc-gh-sfan: Ah I'm not aware of this, let me try `str(python_type)` and see how it works :)

sfc-gh-sfan:
> Just curious: Why aren't any of 3.9 and 3.10 tests required merge gates?

I wanted it to bake for a bit before making it required. Now that you mentioned it, I probably should make python3.9 a required merge gate given it seems quite stable. Will do that next sprint.

tests/unit/test_types.py
sfc-gh-sfan: Hmmm it looks like it does not work perfectly for basic types such as `int`, which gives `<class 'int'>` 🤦

sfc-gh-sfan: Looks like there is `typing.Type`. Let me try that.

sfc-gh-sfan: `type(typing.Optional[str])` gives `<class 'typing._GenericAlias'>`. I think this is beyond our grasp to be honest 🤔

sfc-gh-stan: I see... I just don't like supplying literal strings and not explaining how to obtain them. Especially, now there're at least three cases (1) type (2) TypeVar (3) other type hints. I don't have a strong opinion otherwise.

sfc-gh-stan (approved):

tests/unit/test_types.py
sfc-gh-sfan: To be honest I think we could make them consistence by always providing a string :p

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -3,13 +3,21 @@
...
 from snowflake.snowpark.row import Row
 from snowflake.snowpark.types import StructType
 
+# Python 3.8 needs to use typing.Iterable because collections.abc.Iterable is not subscriptable
+# Python 3.9 can use both
+# Python 3.10 needs to use collections.abc.Iterable because typing.Iterable is removed
+try:
+    from typing import Iterable
+except ImportError:
+    from collections.abc import Iterable
```
sfc-gh-mkeller, line 19: I'm a fan of doing these a different way. See [pyupgrade's docs](https://github.com/asottile/pyupgrade#python2-and-old-python3x-blocks). Doing it this way we can rely on `pyupgrade` to automatically remove these blocks and we know exactly where the code goes for different Python versions.

PR #708, opened 2023-03-06
mark xfail on if_not_exist tests 

sfc-gh-aalam:

   Fixes #NNNN

   Tests that are testing `if_not_exists` show flaky behavior and fail error messages like with `Unknown function TEST_SP_IF_NOT_EXISTS_ADD` on first invocation of sp/udf after being registered with `if_not_exists=True`. Updating the first isntance of registration with `replace=True` instead of `if_not_exists` to test whether it fixes the flakiness.

### DIFF
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -773,6 +773,7 @@ def test_sp_replace(session):
     assert add_sp(1, 2) == 3
 
 
+@pytest.mark.xfail(reason="SNOW-757054 flaky test", strict=False)
 @pytest.mark.skipif(
     IS_IN_STORED_PROC,
     reason="Named temporary procedure is not supported in stored proc",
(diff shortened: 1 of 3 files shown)


sfc-gh-stan:
🤔  Why should changing it to `replace` fix this issue?

sfc-gh-sfan (commented):
I'm also confused why this would fix the issue 🤔

sfc-gh-aalam:
> 🤔 Why would changing it to `replace` fix this issue?

Let me do more research to find a good explanation for this.

sfc-gh-aalam:
A bug is identified when creating temp stored procs with if not exists. https://snowflakecomputing.atlassian.net/browse/SNOW-757054

sfc-gh-stan:
Thank you for investigating! I see the bug fix has been assigned. Since the tests are for `if exists`, I don't think we should fix it by using `replace`. 🤔 Let's either quarantine it with `pytest.mark.xfail(strict=False, reason="SNOW-xxxx")` or just skip it?

sfc-gh-aalam:
> Thank you for investigating! I see the bug fix has been assigned. Since the tests are for `if exists`, I don't think we should fix it by using `replace`. 🤔 Let's either quarantine it with `pytest.mark.xfail(strict=False, reason="SNOW-xxxx")` or just skip it?

You are right. Marking `xfail` seems like the right move

sfc-gh-stan (approved):

PR #716, opened 2023-03-09
SNOW-704094 Flatten select after filter

sfc-gh-stan:

   Fixes #SNOW-704094

   The code change checks if the subquery's where clause depends on a column that is **NEW**  in the current query, if it does not, flatten the query. Note that, `WHERE` is evaluated on the subquery first, so it is safe to flatten when there are columns that are **CHANGED** between the current query and the subquery.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -598,7 +598,7 @@ class SelectStatement(Selectable):
             new.from_ = self.from_.to_subqueryable()
             new.pre_actions = new.from_.pre_actions
             new.post_actions = new.from_.post_actions
-            new.order_by = cols
+            new.order_by = cols + (self.order_by or [])
             new._column_states = self._column_states
         else:
             new = SelectStatement(
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -3,7 +3,7 @@
 #
 
 from abc import ABC, abstractmethod
-from collections import UserDict, namedtuple
+from collections import UserDict
 from copy import copy
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Set, Union
@@ -18,6 +18,8 @@ if TYPE_CHECKING:
     from snowflake.snowpark._internal.analyzer.analyzer import (
         Analyzer,
     )  # pragma: no cover
+
+from dataclasses import dataclass
 
 from snowflake.snowpark._internal.analyzer import analyzer_utils
 from snowflake.snowpark._internal.analyzer.analyzer_utils import result_scan_statement
@@ -46,7 +48,17 @@ SET_UNION_ALL = analyzer_utils.UNION_ALL
 SET_INTERSECT = analyzer_utils.INTERSECT
 SET_EXCEPT = analyzer_utils.EXCEPT
 
-Clause = namedtuple("Clause", "expression dependent_columns")
+
+@dataclass
+class WhereClause:
+    expression: Expression
+    dependent_columns: Set[str]
+
+
+@dataclass
+class OrderByClause:
+    expressions: List[Expression]
+    dependent_columns: Set[str]
 
 
 class ColumnChangeState(Enum):
@@ -327,8 +339,8 @@ class SelectStatement(Selectable):
         *,
         projection: Optional[List[Expression]] = None,
         from_: Optional["Selectable"] = None,
-        where: Optional[Clause] = None,
-        order_by: Optional[Clause] = None,
+        where: Optional[WhereClause] = None,
+        order_by: Optional[OrderByClause] = None,
         limit_: Optional[int] = None,
         offset: Optional[int] = None,
         analyzer: "Analyzer",
@@ -336,8 +348,8 @@ class SelectStatement(Selectable):
         super().__init__(analyzer)
         self.projection: Optional[List[Expression]] = projection
         self.from_: Optional["Selectable"] = from_
-        self.where: Optional[Clause] = where
-        self.order_by: Optional[Clause] = order_by
+        self.where: Optional[WhereClause] = where
+        self.order_by: Optional[OrderByClause] = order_by
         self.limit_: Optional[int] = limit_
         self.offset = offset
         self.pre_actions = self.from_.pre_actions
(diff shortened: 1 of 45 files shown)


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -47,6 +49,18 @@
 SET_EXCEPT = analyzer_utils.EXCEPT
 
 
+@dataclass
+class WhereClause:
+    expression: Expression
+    dependent_columns: Set[str]
```
sfc-gh-stan, line 55: Should be `Optional[Set[str]]` since depdent_columns can be `None` when the Expression is not parsed (e.g. created by `sql_expr`)

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -47,6 +49,18 @@
...
+
+
+@dataclass
+class OrderByClause:
+    expressions: List[Expression]
+    dependent_columns: Set[str]
```
sfc-gh-stan, line 61: Ditto.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -582,11 +595,13 @@ def sort(self, cols: List[Expression]) -> "SelectStatement":
             new.from_ = self.from_.to_subqueryable()
             new.pre_actions = new.from_.pre_actions
             new.post_actions = new.from_.post_actions
-            new.order_by = cols
+            new.order_by = OrderByClause(cols, dependent_columns)
```
sfc-gh-stan, line 598: This a is a bug that will be addressed by [SNOW-756443](https://snowflakecomputing.atlassian.net/browse/SNOW-756443), where chained sorts/orderby should preserve all the dependent columns in order.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-sfan: Should we make the change then?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -488,49 +501,41 @@ def select(self, cols: List[Expression]) -> "SelectStatement":
...
             can_be_flattened = False
             disable_next_level_flatten = True
-        elif self.flatten_disabled or self.has_clause_using_columns:
+        elif self.flatten_disabled:
+            can_be_flattened = False
+        elif (
+            self.where
+            and self.where.dependent_columns
+            and any(
+                new_column_states[_col].change_state in (ColumnChangeState.NEW,)
```
sfc-gh-sfan, line 510: nit: use `==` instead?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -488,49 +501,41 @@ def select(self, cols: List[Expression]) -> "SelectStatement":
...
+                for _col in self.where.dependent_columns.intersection(
+                    new_column_states.active_columns
+                )
+            )
+        ):
+            can_be_flattened = False
+        elif self.order_by:  # TODO: implement orderby
```
sfc-gh-sfan, line 517: Not sure about the policy in open source repo but might be good to have a JIRA?

sfc-gh-sfan (commented):
There are some neat refactors though it also makes the change hard to parse. Could you help highlight the core change?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-stan: SNOW-704094 covers both filter and orderby, I just want to separate the PR's for easier review.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -549,31 +554,39 @@ def select(self, cols: List[Expression]) -> "SelectStatement":
...
+            dependent_columns, self.column_states
+        )
         if can_be_flattened:
             new = copy(self)
             new.from_ = self.from_.to_subqueryable()
             new.pre_actions = new.from_.pre_actions
             new.post_actions = new.from_.post_actions
-            new.where = And(self.where, col) if self.where is not None else col
             new._column_states = self._column_states
+            if self.where:
```
sfc-gh-stan, line 569: Core change #1: merge where clauses between parent query and child query by AND the condition expressions, track the dependent columns of a WHERE clause, merge two dependent columns sets using UNION.
@sfc-gh-sfan @sfc-gh-yixie

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -488,49 +501,41 @@ def select(self, cols: List[Expression]) -> "SelectStatement":
             # We don't flatten when there are duplicate columns.
             can_be_flattened = False
             disable_next_level_flatten = True
-        elif self.flatten_disabled or self.has_clause_using_columns:
+        elif self.flatten_disabled:
+            can_be_flattened = False
+        elif (
+            self.where
```
sfc-gh-stan, line 507: Core change #2: Previously, as long as subquery has a WHERE clause, `can_be_flattened` evaluates to `False`. This change deems that the only case not to be flattened is when any dependent column from the WHERE clause is newly added in the current query, i.e. `df.filter(col("C") > 10).select("*", col(Literal(12)).alias("C")` should not be flattened to `SELECT *, 12 as C FROM <subquery> WHERE C > 10`.

sfc-gh-sfan: This is due to my lack of knowledge of the simplifier: Do we handle the case where select added alias to existing column? E.g. `df.filter(col("C") > 10).select(col("A").alias("C"))`? Is this alias also considered a new column? If so then I think it should be okay.

sfc-gh-stan: We don't flatten `df.filter(col("C") > 10).select(col("A").alias("C"))`. `C` is considered a new column in this case.

sfc-gh-sfan (approved):
LGTM. Let's also have @sfc-gh-yixie take a look

tests/integ/test_simplifier_suite.py
```diff
@@ -995,6 +996,46 @@ def test_rename_to_existing_column_column(session):
     Utils.check_answer(df4, [Row(3, 3, 1)])
 
 
+@pytest.mark.parametrize(
+    "operation,simplified_query",
```
sfc-gh-yixie, line 1000: If I use sql expression, does it flatten, for instance `df.filter(sql_expr('a > 1')).select(...)`?
It shouldn't because we can't detect the dependent columns of a sql text.

sfc-gh-stan: Good catch! I fixed and added a test case.

PR #720, opened 2023-03-10
SNOW-749412: snowpark python functions sample code batch 6

sfc-gh-aling:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -848,19 +848,48 @@ def is_null(e: ColumnOrName) -> Column:
 
 
 def negate(e: ColumnOrName) -> Column:
-    """Returns the negation of the value in the column (equivalent to a unary minus)."""
+    """Returns the negation of the value in the column (equivalent to a unary minus).
+
+    Example::
+
+        >>> df = session.create_dataframe([[1]], schema=["a"])
+        >>> df.select(negate(col("a").alias("result"))).show()
+        ------------
+        |"RESULT"  |
+        ------------
+        |-1        |
+        ------------
+    """
+
     c = _to_col_if_str(e, "negate")
     return -c
 
 
 def not_(e: ColumnOrName) -> Column:
-    """Returns the inverse of a boolean expression."""
+    """Returns the inverse of a boolean expression.
+
+    Example::
+
+        >>> df = session.create_dataframe([[True]], schema=["a"])
+        >>> df.select(not_(col("a").alias("result"))).show()
+        ------------
+        |"RESULT"  |
+        ------------
+        |False     |
+        ------------
+    """
+
     c = _to_col_if_str(e, "not_")
     return ~c
 
 
 def random(seed: Optional[int] = None) -> Column:
-    """Each call returns a pseudo-random 64-bit integer."""
+    """Each call returns a pseudo-random 64-bit integer.
+
+    Example::
+        >>> df = session.sql("select 1")
+        >>> df.select(random(123).alias("result")).show()
+    """
     s = seed if seed is not None else randint(-(2**63), 2**63 - 1)
     return builtin("random")(Literal(s))
 
@@ -1173,7 +1202,20 @@ def degrees(e: ColumnOrName) -> Column:
 
 
 def radians(e: ColumnOrName) -> Column:
-    """Converts degrees to radians."""
+    """Converts degrees to radians.
+
+    Examples::
+
+        >>> df = session.create_dataframe([[1.111], [2.222], [3.333]], schema=["a"])
+        >>> df.select(radians(col("a")).alias("result")).show()
+        ------------------------
+        |"RESULT"              |
+        ------------------------
+        |0.019390607989657     |
+        |0.038781215979314     |
+        |0.058171823968971005  |
+        ------------------------
+    """
     c = _to_col_if_str(e, "radians")
     return builtin("radians")(c)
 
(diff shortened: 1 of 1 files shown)


src/snowflake/snowpark/functions.py
```diff
@@ -1173,7 +1204,21 @@ def degrees(e: ColumnOrName) -> Column:
 
 
 def radians(e: ColumnOrName) -> Column:
-    """Converts degrees to radians."""
+    """Converts degrees to radians.
+
+    Examples::
+
+        >>> df = session.create_dataframe([[1.111], [2.222], [3.333]], schema=["a"])
+        >>> df.select(radians(col("a")).alias("result")).show()
```
sfc-gh-stan, line 1212: Could we use a `CAST` to limit the number of digits after the floating point here?

src/snowflake/snowpark/functions.py
```diff
@@ -1285,21 +1330,65 @@ def ltrim(e: ColumnOrName, trim_string: Optional[ColumnOrName] = None) -> Column
 
 def rpad(e: ColumnOrName, len: Union[Column, int], pad: ColumnOrName) -> Column:
     """Right-pads a string with characters from another string, or right-pads a
-    binary value with bytes from another binary value."""
+    binary value with bytes from another binary value.
```
sfc-gh-stan, line 1333: Could we add "`e` is padded to length `len` with characters/bytes from `pad`".

src/snowflake/snowpark/functions.py
```diff
@@ -1419,7 +1520,21 @@ def pow(
 
 def round(e: ColumnOrName, scale: Union[ColumnOrName, int, float] = 0) -> Column:
     """Returns values from the specified column rounded to the nearest equal or
-    smaller integer."""
+    smaller integer.
```
sfc-gh-stan, line 1523: This is not true, 1.5 is rounded to 2 by default, the [Snowflake round function](https://docs.snowflake.com/en/sql-reference/functions/round) has a third parameter `mode` that can be either [half away from zero](https://en.wikipedia.org/wiki/Rounding#Rounding_half_away_from_zero) or [half to even](https://en.wikipedia.org/wiki/Rounding#Rounding_half_to_even). The default is rounding half away from zero.

src/snowflake/snowpark/functions.py
```diff
@@ -3019,28 +3220,112 @@ def array_to_string(array: ColumnOrName, separator: ColumnOrName) -> Column:
...
 
 def object_construct(*key_values: ColumnOrName) -> Column:
-    """Returns an OBJECT constructed from the arguments."""
+    """Returns an OBJECT constructed from the arguments.
+
+    Example::
+
+        >>> from snowflake.snowpark.types import StructType, StructField, VariantType, StringType
+        >>> df = session.create_dataframe(
+        ...     [["name", "Joe"], ["zip", "98004"]],
```
sfc-gh-stan, line 3255: Could we have a row with `NULL` key and another row with `NULL` value? So the difference between `object_construct` and `object_construct_keep_null` is obvious?

src/snowflake/snowpark/functions.py
```diff
@@ -3019,28 +3220,112 @@ def array_to_string(array: ColumnOrName, separator: ColumnOrName) -> Column:
...
 def object_construct_keep_null(*key_values: ColumnOrName) -> Column:
     """Returns an object containing the contents of the input (i.e. source) object with one or more
-    keys removed."""
+    keys removed.
+
+    Example::
+
+        >>> from snowflake.snowpark.types import StructType, StructField, VariantType, StringType
+        >>> df = session.create_dataframe(
+        ...     [["key_1", "one"], ["key_1", None]],
```
sfc-gh-stan, line 3283: Ditto.

src/snowflake/snowpark/functions.py
```diff
@@ -3019,28 +3220,112 @@ def array_to_string(array: ColumnOrName, separator: ColumnOrName) -> Column:
...
+    The input key must not exist in the object.
+
+    Example::
+
+        >>> df = session.sql(
+        ...     "select object_construct(a,b,c,d,e,f) as obj, k, v from "
+        ...     "values('age', 21, 'zip', 21021, 'name', 'Joe', 'age', 0),"
+        ...     "('age', 26, 'zip', 94021, 'name', 'Jay', 'age', 0) as T(a,b,c,d,e,f,k,v)"
+        ... )
+        >>> df.select(object_delete(col("obj"), col("k")).alias("result")).show()
```
sfc-gh-stan, line 3314: There seems to be no keys removed here? Could we show a case where `keys` is not empty?

sfc-gh-aling: I've updated the sample code to make it explicitly that key "age" is removed

src/snowflake/snowpark/functions.py
sfc-gh-aling: nice catch, updated

src/snowflake/snowpark/functions.py
sfc-gh-aling: sure, code updated

src/snowflake/snowpark/functions.py
sfc-gh-aling: good finding, I removed `nearest equal or smaller integer.` description, I think just using `rounded` should be good

src/snowflake/snowpark/functions.py
sfc-gh-aling: updated, how does it look to you now?

src/snowflake/snowpark/functions.py
sfc-gh-aling: added

sfc-gh-stan (approved):
LGTM, thank you Adam!

src/snowflake/snowpark/functions.py
```diff
@@ -4096,8 +4532,24 @@ def ntile(e: Union[int, ColumnOrName]) -> Column:
     Divides an ordered data set equally into the number of buckets specified by n.
     Buckets are sequentially numbered 1 through n.
 
-    Args:
-        e: The desired number of buckets; must be a positive integer value.
```
sfc-gh-aalam, line 4100: why are we removing arg definition?

sfc-gh-aling: it must be removed by mistake, let me add it back

sfc-gh-sfan (approved):

PR #748, opened 2023-03-27
SNOW-763159: Support anonymous stored procedure (internally)

sfc-gh-jdu:
We would like to support anonymous stored procedure internally first to meet the needs of task API and SnowML.

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -11,6 +11,7 @@ import zipfile
 from logging import getLogger
 from types import ModuleType
 from typing import (
+    Any,
     Callable,
     Dict,
     List,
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -706,3 +714,56 @@ def create_python_udf_or_sp(
     telemetry_client.send_function_usage_telemetry(
         api_call_source, TelemetryField.FUNC_CAT_CREATE.value
     )
+
+
+def generate_anonymous_python_sp_sql(
```
sfc-gh-sfan, line 719: This probably shares a lot with `create_python_udf_or_sp` for sql generation, I wonder if we could refactor a bit to share more code?

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-jdu: I also thought it before, but actually they are not really the same (e.g., object_type, execute_as, and many other arguments cannot apply to anonymous stored proc), so I prefer to keep them separate first. Once we want to add public support, we can check it later to see whether there is a refactoring opportunity.

sfc-gh-jdu: Sure, https://snowflakecomputing.atlassian.net/browse/SNOW-769761

PR #753, opened 2023-03-28
SNOW-706838 skip upload on content match

sfc-gh-aalam:

   Fixes #SNOW-706838

  Follow up to https://github.com/snowflakedb/snowflake-connector-python/pull/1489 where option `_skip_upload_on_content_match` was added to snowflake connector for python that skips file uploads if content matches in remote location.

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -4,6 +4,7 @@
 #
 
 import functools
+import inspect
 import os
 import sys
 import time
@@ -16,7 +17,6 @@ from snowflake.connector.cursor import ResultMetadata, SnowflakeCursor
 from snowflake.connector.errors import NotSupportedError, ProgrammingError
 from snowflake.connector.network import ReauthenticationRequest
 from snowflake.connector.options import pandas
-from snowflake.connector.version import VERSION as SNOWFLAKE_CONNECTOR_VERSION
 from snowflake.snowpark._internal.analyzer.analyzer_utils import (
     escape_quotes,
     quote_name_without_upper_casing,
@@ -143,6 +143,12 @@ class ServerConnection:
         # Snowpark session
         self._telemetry_client.send_session_created_telemetry(not bool(conn))
 
+        # check if cursor.execute supports _skip_upload_on_content_match
+        signature = inspect.signature(self._cursor.execute)
+        self._supports_skip_upload_on_content_match = (
+            "_skip_upload_on_content_match" in signature.parameters
+        )
+
     def _add_application_name(self) -> None:
         if PARAM_APPLICATION not in self._lower_case_parameters:
             # Mirrored from snowflake-connector-python/src/snowflake/connector/connection.py#L295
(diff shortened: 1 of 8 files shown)


src/snowflake/snowpark/session.py
sfc-gh-aalam, line 1: @sfc-gh-jdu I was thinking on how to remove the check you mentioned in this comment https://github.com/snowflakedb/snowflake-connector-python/pull/1489#pullrequestreview-1357450219 but I don't think we can remove it until we pin the minimum snowflake connector version to be >= 3.0.3. Until then, we will keep this path. I have added `skip_upload_on_content_match` in the other uploads below.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -225,6 +227,10 @@ def upload_file(
                 raise ne.with_traceback(tb) from None
         else:
             uri = normalize_local_file(path)
+            if SNOWFLAKE_CONNECTOR_VERSION >= (3, 0, 3, None):
```
sfc-gh-jdu, line 230: Let's have a jira/todo here to remove this check, once connector is released?

src/snowflake/snowpark/session.py
sfc-gh-jdu: sounds good, create a jira for this?

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: we can't remove this check after connector release right. We would also have to set minimum snowflake connector version to be >= 3.0.3. The current minimum required version is 2.7. What should be the policy to update the minimum required connector version? @sfc-gh-yixie @sfc-gh-mkeller 

https://snowflakecomputing.atlassian.net/browse/SNOW-769892

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -267,6 +275,13 @@ def upload_stream(
                     )
                     raise ne.with_traceback(tb) from None
             else:
+                if SNOWFLAKE_CONNECTOR_VERSION >= (3, 0, 3, None):
+                    kwargs = {
```
sfc-gh-stan, line 279: Same as above, I prefer explicitly specifying the parameters. Is there a reason why we would want to use a dictionary?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -225,6 +227,10 @@ def upload_file(
                 raise ne.with_traceback(tb) from None
         else:
             uri = normalize_local_file(path)
+            if SNOWFLAKE_CONNECTOR_VERSION >= (3, 0, 3, None):
+                kwargs = {"_skip_upload_on_content_match": skip_upload_on_content_match}
```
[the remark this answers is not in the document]
sfc-gh-aalam: I chose to do it this way because I can just send `kwargs` to the `run_query` below using `**kwargs`. Otherwise, I would write the `run_query` with arguments twice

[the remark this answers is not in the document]
sfc-gh-stan: I see, that makes sense. I guess if we decide to bump the minimum python connector dependency, this could be avoided 🤔

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: I found a way using `inspect.signature` to see if a function accepts a certain keyword argument

sfc-gh-sfan (approved):

PR #755, opened 2023-03-28
Add statement params to dataframe as a property

sfc-gh-hayu:

   Fixes #NNNN

   Use dataframe's statement params at query execution if no statement params are specified (eg in `collect`). We need this to track credit consumption from dataframes with snowml contributions. cc: @sfc-gh-pdorairaj

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Added support for `functions.hash()` to accept a variable number of input expressions.
 - Added API `Session.RuntimeConfig` for getting/setting/checking the mutability of any runtime configuration.
 - Added support managing case sensitivity in `Row` results from `DataFrame.collect` using `case_sensitive` parameter.
+- Added the property `DataFrame.statement_params` to allow statement parameters.
 
 ### Bug Fixes
 
@@ -47,7 +48,6 @@ is enabled by default, turn off by specifying `source_code_display=False` at reg
   - `functions.date_format()` for `functions.to_date()`.
   - `functions.monotonically_increasing_id()` for `functions.seq8()`
   - `functions.from_unixtime()` for `functions.to_timestamp()`
-- Added the property `DataFrame.statement_params` to allow statement parameters.
 
 ### Bug Fixes:
 - Fixed a bug in SQL simplifier that didn’t handle Column alias and join well in some cases. See https://github.com/snowflakedb/snowpark-python/issues/658 for details.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ is enabled by default, turn off by specifying `source_code_display=False` at reg
   - `functions.date_format()` for `functions.to_date()`.
   - `functions.monotonically_increasing_id()` for `functions.seq8()`
   - `functions.from_unixtime()` for `functions.to_timestamp()`
+- Added the property `DataFrame.statement_params` to return statement parameters.
 
 ### Bug Fixes:
 - Fixed a bug in SQL simplifier that didn’t handle Column alias and join well in some cases. See https://github.com/snowflakedb/snowpark-python/issues/658 for details.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ is enabled by default, turn off by specifying `source_code_display=False` at reg
   - `functions.date_format()` for `functions.to_date()`.
   - `functions.monotonically_increasing_id()` for `functions.seq8()`
   - `functions.from_unixtime()` for `functions.to_timestamp()`
-- Added the property `DataFrame.statement_params` to return statement parameters.
+- Added the property `DataFrame.statement_params` to allow statement parameters.
 
 ### Bug Fixes:
 - Fixed a bug in SQL simplifier that didn’t handle Column alias and join well in some cases. See https://github.com/snowflakedb/snowpark-python/issues/658 for details.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,6 @@
 - Added support for managing case sensitivity in `Row` results from `DataFrame.collect` using `case_sensitive` parameter.
 - Added indexer support for `snowflake.snowpark.types.StructType`.
 - Added a keyword argument `log_on_exception` to `Dataframe.collect` and `Dataframe.collect_no_wait` to optionally disable error logging for SQL exceptions.
-- Added the property `DataFrame.statement_params` to allow statement parameters.
 
 ### Bug Fixes
 
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -521,6 +521,14 @@ class DataFrame:
         self.replace = self._na.replace
 
     @property
+    def statement_params(self) -> Optional[Dict[str, Any]]:
+        return self._statement_params
+
+    @statement_params.setter
+    def query_tag(self, statement_params: Optional[Dict[str, Any]]) -> None:
+        self._statement_params = statement_params
+
+    @property
     def stat(self) -> DataFrameStatFunctions:
         return self._stat
 
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -521,14 +521,6 @@ class DataFrame:
         self.replace = self._na.replace
 
     @property
-    def statement_params(self) -> Optional[Dict[str, Any]]:
-        return self._statement_params
-
-    @statement_params.setter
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ is enabled by default, turn off by specifying `source_code_display=False` at reg
   - `functions.date_format()` for `functions.to_date()`.
   - `functions.monotonically_increasing_id()` for `functions.seq8()`
   - `functions.from_unixtime()` for `functions.to_timestamp()`
+- Added the property `DataFrame.statement_params` to allow statement parameters.
```
sfc-gh-stan, line 50: This is in the wrong release.

sfc-gh-stan:
I don't think it makes sense to keep statement parameters as a property of `DataFrame`. Why not just specify `statement_params` at query execution?

sfc-gh-hayu:
> I don't think it makes sense to keep statement parameters as a property of `DataFrame`. Why not just specify `statement_params` at query execution?

For snowml use cases, we create/modify dataframes and therefore should take credit for part of the consumption. However the DFs may not be executed via our APIs, meaning we are unable to specify `statement_params`, resulting in underestimated credit consumption.

sfc-gh-stan:
> > I don't think it makes sense to keep statement parameters as a property of `DataFrame`. Why not just specify `statement_params` at query execution?
> 
> For snowml use cases, our APIs may return created/modified dataframes and therefore should take credit for part of the consumption. However the DFs may not be executed via snowml APIs, so we are unable to specify `statement_params`, resulting in underestimated credit consumption.

What about using query tags?

sfc-gh-hayu:
> > > I don't think it makes sense to keep statement parameters as a property of `DataFrame`. Why not just specify `statement_params` at query execution?
> > 
> > 
> > For snowml use cases, our APIs may return created/modified dataframes and therefore should take credit for part of the consumption. However the DFs may not be executed via snowml APIs, so we are unable to specify `statement_params`, resulting in underestimated credit consumption.
> 
> What about using query tags?

`query_tag` is a session property and part of `statement_params` so I think it doesn't work here? Also @sfc-gh-yixie should have the context of this change.

sfc-gh-sfan:
If we are trying to track credit usage, I wonder if we could change the application name by checking certain setup (e.g. if `snowml` is in the modules, like what streamlit does): https://github.com/snowflakedb/snowpark-python/blob/69a60e8d1f8cf539ed849eed6b4b591440e3c4d3/src/snowflake/snowpark/_internal/server_connection.py#L144

sfc-gh-hayu:
> If we are trying to track credit usage, I wonder if we could change the application name by checking certain setup (e.g. if `snowml` is in the modules, like what streamlit does):
> 
> https://github.com/snowflakedb/snowpark-python/blob/69a60e8d1f8cf539ed849eed6b4b591440e3c4d3/src/snowflake/snowpark/_internal/server_connection.py#L144

sfc-gh-hayu:
> If we are trying to track credit usage, I wonder if we could change the application name by checking certain setup (e.g. if `snowml` is in the modules, like what streamlit does):
> 
> https://github.com/snowflakedb/snowpark-python/blob/69a60e8d1f8cf539ed849eed6b4b591440e3c4d3/src/snowflake/snowpark/_internal/server_connection.py#L144

What if multiple applications claim contributions to credit usage? Later we need to provide universal telemetry for snowpark use cases and having application overwritten by the latest source may not be ideal. Where is "application" tracked in `job_etl_v`?

src/snowflake/snowpark/dataframe.py
```diff
@@ -622,7 +623,9 @@ def _internal_collect_with_tag_no_telemetry(
             block=block,
             data_type=data_type,
             _statement_params=create_or_update_statement_params_with_query_tag(
-                statement_params, self._session.query_tag, SKIP_LEVELS_THREE
+                statement_params if statement_params else self._statement_params,
```
sfc-gh-yixie, line 626: Nit: `statement_params  or self._statement_params`

src/snowflake/snowpark/dataframe.py
```diff
@@ -3459,6 +3478,14 @@ def dtypes(self) -> List[Tuple[str, str]]:
         ]
         return dtypes
 
+    @property
+    def statement_params(self) -> Optional[Dict[str, Any]]:
```
sfc-gh-yixie, line 3482: Does a user need to use this property? If no I suggest removing it.

sfc-gh-yixie:
What parameters will be used to calculate the credits? If possible can we add some tests?

sfc-gh-sfan:
BTW I'm not opposed to this change cuz it is useful in general. But I want to see if this is right way to solve your problem.

> What if multiple applications claim contributions to credit usage? 

Do we have a way to track multiple applications today? Before we have a proper way to do this, I would find it awkward to use a statement level parameter. For snowml/streamlit, I feel the contribution should be attribute to the top level application, not snowpark and we could write query to include snowml/streamlit if we want to count them towards snowpark. For example, today snowpark's usage goes to snowpark and is not counted as python connector, though we could tweak the query to do that.

> Where is "application" tracked in job_etl_v?

I'm not sure about `job_etl_v`, but you should be able to find it in `session_etl_v.client_app_id`.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Added support for managing case sensitivity in `Row` results from `DataFrame.collect` using `case_sensitive` parameter.
 - Added indexer support for `snowflake.snowpark.types.StructType`.
 - Added a keyword argument `log_on_exception` to `Dataframe.collect` and `Dataframe.collect_no_wait` to optionally disable error logging for SQL exceptions.
+- Added the property `DataFrame.statement_params` to allow statement parameters.
```
sfc-gh-yixie, line 20: This line also needs to disappear.

sfc-gh-hayu:
> What parameters will be used to calculate the credits? If possible can we add some tests?

Added unit test `test_statement_params`.
Using "source" (eg `statement_params["source"]="SnowML"`) to tag jobs for credit usage tracking.

sfc-gh-hayu:
> BTW I'm not opposed to this change cuz it is useful in general. But I want to see if this is right way to solve your problem.
> 
> > What if multiple applications claim contributions to credit usage?
> 
> Do we have a way to track multiple applications today? Before we have a proper way to do this, I would find it awkward to use a statement level parameter. For snowml/streamlit, I feel the contribution should be attribute to the top level application, not snowpark and we could write query to include snowml/streamlit if we want to count them towards snowpark. For example, today snowpark's usage goes to snowpark and is not counted as python connector, though we could tweak the query to do that.
> 
> > Where is "application" tracked in job_etl_v?
> 
> I'm not sure about `job_etl_v`, but you should be able to find it in `session_etl_v.client_app_id`.

Thanks Shixuan. Right we can't track multiple applications today. Regarding the `sys.modules` approach used by streamlit, how to determine the top level application when there are more than one module?

We use statement params as our telemetry is based on `job_etl_v`. Switching to application name will require lots of rework. Also users mix snowml and other APIs (eg snowpark) in a session. If we change the application to snowml then it will take all credits from the session.

sfc-gh-sfan:
Sounds good. I think it would be useful to have a framework for easier application tracking, given that our ecosystem is growing :)

sfc-gh-hayu:
> Sounds good. I think it would be useful to have a framework for easier application tracking, given that our ecosystem is growing :)

Yes that'd be great as an increasing number of projects are dependent on snowpark

sfc-gh-hayu:

sfc-gh-yixie (approved):

tests/unit/test_dataframe.py
```diff
@@ -200,3 +200,20 @@ def test_same_joins_should_generate_same_queries(join_type):
...
+
+
+def test_statement_params():
+    mock_connection = mock.create_autospec(ServerConnection)
+    mock_connection._conn = mock.MagicMock()
+    session = snowflake.snowpark.session.Session(mock_connection)
+    session._conn._telemetry_client = mock.MagicMock()
+    df = session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"])
+    statement_params = {"param1": "val", "param2": 0, "param3": True}
+    df._statement_params = statement_params
```
sfc-gh-aalam, line 212: should we add a better way to set `_statement_params`. Like adding it in `__init__` or adding a method `dataframe.set_statement_params(params)`

sfc-gh-aalam: @sfc-gh-yixie @sfc-gh-sfan, what do you think?

sfc-gh-aalam: I see. thanks for clarification

PR #762, opened 2023-03-31
SNOW-704112 explode related methods

sfc-gh-aalam:

   Fixes #SNOW-704112 

### DIFF
--- a/src/snowflake/snowpark/table_function.py
+++ b/src/snowflake/snowpark/table_function.py
@@ -252,11 +252,7 @@ def _get_cols_after_join_table(
 def _get_cols_after_explode_join(
     func: _ExplodeFunctionCall, plan: SnowflakePlan
 ) -> List:
-    explode_col_type = None
-    for attr in plan.output:
-        if attr.name == func.col:
-            explode_col_type = attr.datatype
-            break
+    explode_col_type = plan.output_dict.get(func.col, None)
 
     cols = []
     if isinstance(explode_col_type, ArrayType):
--- a/src/snowflake/snowpark/table_function.py
+++ b/src/snowflake/snowpark/table_function.py
@@ -252,7 +252,7 @@ def _get_cols_after_join_table(
 def _get_cols_after_explode_join(
     func: _ExplodeFunctionCall, plan: SnowflakePlan
 ) -> List:
-    explode_col_type = plan.output_dict.get(func.col, None)
+    explode_col_type = plan.output_dict.get(func.col, [None])[0]
 
     cols = []
     if isinstance(explode_col_type, ArrayType):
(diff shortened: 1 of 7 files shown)


src/snowflake/snowpark/table_function.py
```diff
@@ -130,6 +131,23 @@ def alias(self, *aliases: str) -> "TableFunctionCall":
...
+
+    def __init__(self, col: ColumnOrName) -> None:
+        super().__init__("flatten", col)
+        if isinstance(col, Column):
+            self.col = col._expression.name
+        else:
+            self.col = quote_name(col)
+
+    def __str__(self) -> str:
+        return "explode"
```
sfc-gh-sfan, line 149: Ah I see why you want to use `__str__` now because user's function name is different from internal Snowflake function name (explode vs flatten). Maybe we should consider adding a field that override the name in error message? E.g. `user_visible_name` or something similar.

src/snowflake/snowpark/table_function.py
```diff
@@ -233,3 +251,43 @@ def get_column_names_from_plan(plan: SnowflakePlan) -> List[str]:
         new_cols = [Column(col)._named() for col in new_cols]
 
     return old_cols, new_cols
+
+
+def _get_cols_after_explode_join(
+    func: _ExplodeFunctionCall, plan: SnowflakePlan
+) -> List:
+    explode_col_type = None
+    for attr in plan.output:
```
sfc-gh-sfan, line 260: I'm curious if it worths introducing a dict in `plan` for this type search. Maybe this search can be pushed into `plan`, and `plan` can build a dict lazily such that the next search is a lookup if `explode` is called multiple times? Or is this an overkill?

src/snowflake/snowpark/table_function.py
sfc-gh-aalam: That is a good suggestion. I'll add it to this PR.

src/snowflake/snowpark/table_function.py
sfc-gh-aalam: `plan` is instance of `SnowflakePlan` with has `output` as `cached_property` decorator in its definition so I think we don't need to add more optimization.

src/snowflake/snowpark/table_function.py
```diff
@@ -130,6 +131,23 @@ def alias(self, *aliases: str) -> "TableFunctionCall":
...
 
+    def __str__(self) -> str:
+        return self.name
+
+
+class _ExplodeFunctionCall(TableFunctionCall):
+    """Internal class to identify explode function call as a special instance of TableFunctionCall"""
+
+    def __init__(self, col: ColumnOrName) -> None:
+        super().__init__("flatten", col)
```
sfc-gh-jdu, line 142:
```suggestion
        super().__init__("explode", col)
```
?

sfc-gh-jdu (approved):
It seems working perfectly because we only need data type (metadata) : )

src/snowflake/snowpark/table_function.py
sfc-gh-sfan: `output` itself should be fine, my worry is about this for-loop in which we iterate over all names to find the name match which is O(n). If we have a dict then it would be O(1).

sfc-gh-sfan (approved):

PR #775, opened 2023-04-08
SNOW-738785: Add TaskContext for Task API

sfc-gh-jdu:
This class is needed for Task API prpr. Once Task API is public, we will move this class there. For now, this class is not supposed to be used by Snowpark users directly (so no changelog is needed). If you're interested in the design and usage of this class, take a look at this [doc](https://docs.google.com/document/d/1XN3yUCeBbxYVc2aDEHxHWbvFJNV9jfERK_FTlfkBArQ/edit?usp=sharing).

### DIFF
--- a/src/snowflake/snowpark/context.py
+++ b/src/snowflake/snowpark/context.py
@@ -8,7 +8,7 @@ from typing import Any, Optional
 
 import snowflake.snowpark
 from snowflake.snowpark._internal.analyzer.analyzer_utils import single_quote
-from snowflake.snowpark._internal.utils import validate_object_name
+from snowflake.snowpark._internal.utils import private_preview, validate_object_name
 
 _use_scoped_temp_objects = True
 
@@ -38,6 +38,7 @@ class TaskContext:
         """
         self._session = session
 
+    @private_preview(version="1.4.0")
     def set_return_value(self, value: Any) -> None:
         """
         Explicitly sets the return value for a task. This value can be retrieved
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/context.py
```diff
@@ -18,3 +22,65 @@ def get_active_session() -> "snowflake.snowpark.Session":
...
+        This method can only be called in a Snowflake task.
+
+        See `SYSTEM$GET_PREDECESSOR_RETURN_VALUE <https://docs.snowflake.com/en/sql-reference/functions/system_get_predecessor_return_value>`_ for details.
+
+        Args:
+            task_name: The task name of the predecessor task that sets the return value to be retrieved.
+
+                - If the task has only one predecessor task that is enabled, the argument is optional.
+                  If this argument is omitted, the function retrieves the return value for the only enabled predecessor task.
+                - If the task has multiple predecessor tasks that are enabled, this argument is required.
```
sfc-gh-sfan, line 70: I'm curious: what is the behavior if no argument is provided and the task has multiple predecessor tasks? Can we add a test to cover this case?

sfc-gh-jdu: It will throw an error. We're not able to test it because it can only be called in task. On the SnowAPI side, we'll have such tests (I have tested locally)

sfc-gh-sfan (approved):

sfc-gh-jdu:
move to https://github.com/snowflakedb/snowapi-python/pull/26

PR #777, opened 2023-04-11
SNOW-674611 Table Sprocs API

sfc-gh-aalam:

   Fixes #SNOW-674611

   Extending stored proc API to register table stored procedures along with documentation updates and examples.


src/snowflake/snowpark/functions.py
```diff
@@ -6702,17 +6702,75 @@ def sproc(
         registered. Invoking :func:`sproc` with ``replace`` set to ``True`` will overwrite the
         previously registered function.
 
+        4. To describe the return type for a table stored procedure, use
```
sfc-gh-mabrennan, line 6705: Maybe change 
"To describe the return type for a table stored procedure" 
to 
"To describe the return type for a stored procedure that returns tabular data"

src/snowflake/snowpark/functions.py
```diff
@@ -6702,17 +6702,75 @@ def sproc(
...
+        either of the following ways:
```
sfc-gh-mabrennan, line 6706: "either" suggests that there are only 2 options and it looks like there are 3 ways to do this.
I suggest that you change "either" to "one".

src/snowflake/snowpark/functions.py
```diff
@@ -6702,17 +6702,75 @@ def sproc(
...
+
+            - (Recommended) Describing return type using :attr:`~snowflake.snowpark.types.StructType`
```
sfc-gh-mabrennan, line 6708: Change "Describing" to "Describe the".

sfc-gh-mabrennan (commented):
The doc strings look really good. I just made a few suggestions.

src/snowflake/snowpark/functions.py
```diff
@@ -6702,17 +6702,75 @@ def sproc(
         registered. Invoking :func:`sproc` with ``replace`` set to ``True`` will overwrite the
         previously registered function.
 
+        4. To describe the return type for a stored procedure that returns tabular data, use
```
sfc-gh-jdu, line 6705: can we have a link to table sproc doc

src/snowflake/snowpark/functions.py
```diff
@@ -6702,17 +6702,75 @@ def sproc(
...
+        one of the following ways:
+
+            - (Recommended) Describe the return type using :attr:`~snowflake.snowpark.types.StructType`
+              and :attr:`~snowflake.snowpark.types.StructField`. Set ``return_type =
+              StructType([StructField("a", DataTypeA()), ...])`` to describe the case
+              ``RETURNS TABLE(A DataTypeA, ...)``.
+
+            - Set ``return_type = StructType([])`` to describe the case ``RETURNS TABLE()``.
```
sfc-gh-jdu, line 6713: what if return_type is not provided? Should it also imply RETURNS TABLE()?

src/snowflake/snowpark/session.py
```diff
@@ -1856,11 +1858,23 @@ def sproc(self) -> StoredProcedureRegistration:
...
+        except Exception:
+            pass
+        return False
+
     def call(
         self,
         sproc_name: str,
         *args: Any,
         statement_params: Optional[Dict[str, Any]] = None,
+        _is_return_table: Optional[bool] = None,
```
sfc-gh-jdu, line 1877: docstring for this arg

src/snowflake/snowpark/session.py
```diff
@@ -1856,11 +1858,23 @@ def sproc(self) -> StoredProcedureRegistration:
...
         return self._sp_registration
 
+    def _infer_is_return_table(self, sproc_name: str, *args: Any) -> bool:
+        try:
+            arg_types = [convert_sp_to_sf_type(infer_type(arg)) for arg in args]
+            func_signature = f"{sproc_name.upper()}({', '.join(arg_types)})"
+            sproc_desc = self._run_query(f"describe procedure {func_signature}")
+            return_type = sproc_desc[1][1]
+            return return_type.upper().startswith("TABLE")
+        except Exception:
```
sfc-gh-jdu, line 1868: what error are we try to catch here? when describe procedure fails? Should we raise an exception directly or have a warning?

src/snowflake/snowpark/functions.py
```diff
@@ -6702,17 +6702,75 @@ def sproc(
...
+    Example 1::
         >>> from snowflake.snowpark.types import IntegerType
         >>> @sproc(return_type=IntegerType(), input_types=[IntegerType(), IntegerType()], packages=["snowflake-snowpark-python"])
         ... def add_sp(session_, x, y):
         ...     return session_.sql(f"SELECT {x} + {y}").collect()[0][0]
         ...
         >>> add_sp(1, 1)
         2
+
+    Example 2::
```
sfc-gh-jdu, line 6731: are these example same as ones in stored_procedures.py? If the same, then we don't need them here? Just refer to examples in stored_procedures.py?

tests/integ/test_stored_procedure.py
```diff
@@ -689,6 +694,115 @@ def _(
...
+                StructField("c", StringType()),
+            ]
+        ),
+    ],
+)
+@pytest.mark.skipif(
+    IS_IN_STORED_PROC,
+    reason="Named temporary procedure is not supported in stored proc",
+)
+def test_table_sproc(session, is_permanent, anonymous, ret_type):
```
sfc-gh-jdu, line 716: can we add a test for session.call, when the table stored proc doesn't exist, because the inferred type from arguments doesn't match with the original arg type of stored proc?

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -457,6 +458,9 @@ def python_type_to_snow_type(tp: Union[str, Type]) -> Tuple[DataType, bool]:
                 False,
             )
 
+    if tp == SnowparkDataFrame:
+        return StructType([]), False
```
sfc-gh-sfan, line 462: Curious: why is this an empty list? Is this for the case of dynamic table return type?

src/snowflake/snowpark/session.py
sfc-gh-sfan: Yeah we should at least print a warning because here we are falling back to non-table sproc and will only return the first field, which could be surprising to user.

src/snowflake/snowpark/types.py
```diff
@@ -341,6 +343,9 @@ def __init__(self, col_types: Iterable[DataType]) -> None:
         self.col_types = col_types
 
 
+#: The type hint for annotating SnowparkDataFrame return type when registering table Stored Procedures.
+SnowparkDataFrame = TypeVar("SnowparkDataFrame")
```
sfc-gh-yixie, line 347: Does `snowflake.snowpark.DataFrame` work?

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-yixie: +1. Can we input nothing if it's empty?

sfc-gh-aalam: `StructType` requires one positional argument for creating a StructType object. This is indeed the case of dynamic return type.

src/snowflake/snowpark/functions.py
sfc-gh-aalam: If no `return_type` is provided then sproc registration should fail right. We enforce return types in non-table return types as well.

src/snowflake/snowpark/functions.py
sfc-gh-aalam: They are already referred in See Also so I'll just revert this section.

src/snowflake/snowpark/session.py
sfc-gh-aalam: added

src/snowflake/snowpark/session.py
sfc-gh-aalam: added

tests/integ/test_stored_procedure.py
sfc-gh-aalam: added. I hope this is what you were expecting. see `test_negative_table_sproc`

src/snowflake/snowpark/session.py
sfc-gh-yixie: Is this an internal parameter not to be used by a user?

src/snowflake/snowpark/functions.py
sfc-gh-jdu: ah ok I misunderstood, return_type is needed to indicate it's a table stored proc. Yea I agree with Yijun that we should use StructType()

src/snowflake/snowpark/session.py
sfc-gh-jdu: Yea probably we should make it public

src/snowflake/snowpark/types.py
sfc-gh-jdu: yea make sense

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -337,6 +343,54 @@ class StoredProcedureRegistration:
...
+            -------------
+            |1    |2    |
+            -------------
+            <BLANKLINE>
+
+    Example 9
+        Creating a table stored procedure using implicit type hints::
+
+            >>> from snowflake.snowpark.types import SnowparkDataFrame
+            >>> @sproc()
```
sfc-gh-jdu, line 382:
```suggestion
            >>> @sproc
```

tests/integ/test_stored_procedure.py
sfc-gh-jdu: yes, thanks

sfc-gh-jdu (approved):

src/snowflake/snowpark/session.py
```diff
@@ -1856,18 +1858,34 @@ def sproc(self) -> StoredProcedureRegistration:
         """
         return self._sp_registration
 
+    def _infer_is_return_table(self, sproc_name: str, *args: Any) -> bool:
+        try:
+            arg_types = [convert_sp_to_sf_type(infer_type(arg)) for arg in args]
+            func_signature = f"{sproc_name.upper()}({', '.join(arg_types)})"
+            sproc_desc = self._run_query(f"describe procedure {func_signature}")
+            return_type = sproc_desc[1][1]
```
sfc-gh-sfan, line 1866: Let's add a comment about the returned describe result so it is easy to understand this result extraction.

sfc-gh-sfan (approved):
LGTM. Let's also address @sfc-gh-yixie 's comment on using `DataFrame`

src/snowflake/snowpark/session.py
sfc-gh-aalam: When I added this parameter, I did so with the intention of keeping it private. Having this enabled me skip `desc procedure` check when stored procedure is created in session and we know the return type is a table sproc. It also acts as an override to the result from `_infer_is_return_table`. I guess we can make it public if we want users to be able to override the results from `_infer_is_return_table`

src/snowflake/snowpark/session.py
sfc-gh-aalam: Added comment about return type and addressed Yijun's `DataFrame` comments.

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -395,6 +395,7 @@ def python_type_to_snow_type(tp: Union[str, Type]) -> Tuple[DataType, bool]:
     """Converts a Python type or a Python type string to a Snowpark type.
     Returns a Snowpark type and whether it's nullable.
     """
+    from snowflake.snowpark.dataframe import DataFrame
```
sfc-gh-aalam, line 398: added import here to circumvent circular import issues.

src/snowflake/snowpark/session.py
sfc-gh-aalam: @sfc-gh-yixie could I get your opinion on this. Should we make `_infer_is_return_table` public?

sfc-gh-yixie: If it's just for performance I'm good not having it as a public API now.
But how about removing it from this function?

```python
def call(
        self,
        sproc_name: str,
        *args: Any,
        statement_params: Optional[Dict[str, Any]] = None,
        ...
):
    self._call(sproc_name, *args, statement_params, ...)

def _call(
        self,
        sproc_name: str,
        *args: Any,
        statement_params: Optional[Dict[str, Any]] = None,
        is_return_table: Optional[bool] = None,
        ...
)
```

The the internal code can use `_call` instead of `call`, and users won't see `_is_return_table `.

tests/integ/test_stored_procedure.py
```diff
@@ -689,6 +695,145 @@ def _(
...
+    - session.call works with provided function name
+    - we can both specify return cols types and keep it blank
+        - exception: sproc from decorator and implicit type hint cannot specify return col types
+    - dataframe returned after a sproc call can be operated on like normal dataframes
+    """
+    if len(ret_type.fields) == 0 and not session.sql_simplifier_enabled:
+        # if return type does not define output columns and sql_simplifier is
+        # disabled, then we don't support dataframe operations on table sprocs
+        pytest.skip()
+
```
sfc-gh-aalam, line 732: I want to know how the team feels about this. The reason why we cannot have dataframe operations working with sql simplifier disabled + return columns not defined is because when sql simplifier is disabled, the code path triggers `plan.output` which results in describe query for the plan `call table_sproc(args)`. Describe query only runs describe without running the actual functions with arguments. In the case when output columns are not defined in table sproc functions definition, the describe query returns `TABLE ()`. This breaks subsequent query creation. 
The same does not happen with sql simplifier is enabled because it skips the part which calls `plan.output`.

PR #778, opened 2023-04-11
SNOW-754127 Support binding variable in session.sql

sfc-gh-sfan:
Description

- Add params to session.sql
- Plumb params to python connector
- Make sure params are processed correctly in plan nodes

Testing

integ test

   Fixes SNOW-754127 

- Add params to session.sql
- Plumb params to python connector
- Make sure params are processed correctly in plan nodes

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -6,16 +6,7 @@ from abc import ABC, abstractmethod
 from collections import UserDict
 from copy import copy
 from enum import Enum
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    List,
-    Optional,
-    Sequence,
-    Set,
-    Union,
-)
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Set, Union
 
 from snowflake.snowpark._internal.analyzer.table_function import (
     TableFunctionExpression,
@@ -184,7 +175,7 @@ class Selectable(LogicalPlan, ABC):
     @abstractmethod
     def query_params(self) -> Optional[Sequence[Any]]:
         """Returns the sql query of this Selectable logical plan."""
-        pass
+        pass  # pragma: no cover
 
     @property
     def sql_in_subquery(self) -> str:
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -171,6 +180,12 @@ def sql_query(self) -> str:
         """Returns the sql query of this Selectable logical plan."""
         pass
 
+    @property
+    @abstractmethod
+    def query_params(self) -> Optional[Sequence[Any]]:
+        """Returns the sql query of this Selectable logical plan."""
+        pass
```
sfc-gh-aalam, line 187: could we add pragma: no cover here so codecov bot doesn't complain

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -441,6 +488,24 @@ def sql_query(self) -> str:
...
+        if self._query_params is not None:
+            return self._query_params
+
+        if not self.has_clause and not self.projection:
+            self._query_params = self.from_.query_params
+            return self._query_params
+
+        query_params = []
+        if self.params:
+            query_params.extend(self.params)
```
sfc-gh-aalam, line 502: can we add a test that triggers this code path

sfc-gh-aalam (approved):
LGTM. just codecov related comments

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-sfan: Good catch. It does not seem like this path can be hit because we don't initiate `SelectStatement` with `params`. I'm just going to remove `self.params` altogether.

sfc-gh-sfan:
The client side tests work well, but when the generated query is not a SELECT/INSERT query, the test fails in stored proc. The test failure could be found in https://ci-dev-133.int.snowflakecomputing.com/job/RT-PC-TestWorker/124754/testReport/junit/(root)/t_python_sp_integ/snowpark_test_bind_variable_sql/. In short, here are the SQLs that failed:

`create or replace table identifier(?) (id int)`
`CREATE  OR  REPLACE    TABLE  SNOWPARK_TEMP_TABLE_4599NVAVX6 AS  SELECT  *  FROM (select column1::INT, column2 from values (?, ?), (?, ?))`
A bind with date `Date \'01-01-1970\' is not recognized`

As discussed in the standup, I'll add a check to explicitly say it is not supported in stored proc, so that we don't keep this PR lingering for too long. I created [SNOW-796947](https://snowflakecomputing.atlassian.net/browse/SNOW-796947) to track the follow-up item

PR #798, opened 2023-04-20
SNOW-791242 Use multistatement query for async multi-query plan

sfc-gh-sfan:
Description

Currently we use SQL stored proc for multi-query plan during async collect. This is not binding friendly, so we want to migrate to use multistatement that we already support in the connector.

Testing

integ test

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/async_job.py
+++ b/src/snowflake/snowpark/async_job.py
@@ -182,6 +182,7 @@ class AsyncJob:
         post_actions: Optional[List[Query]] = None,
         log_on_exception: bool = False,
         case_sensitive: bool = True,
+        num_statements: Optional[int] = None,
         **kwargs,
     ) -> None:
         self.query_id: str = query_id  #: The query ID of the executed query
@@ -193,6 +194,7 @@ class AsyncJob:
         self._post_actions = post_actions if post_actions else []
         self._log_on_exception = log_on_exception
         self._case_sensitive = case_sensitive
+        self._num_statements = num_statements
         self._parameters = kwargs
         self._result_meta = None
         self._inserted = False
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/async_job.py
```diff
@@ -331,6 +333,15 @@ def result(
             _AsyncResultType(result_type.lower()) if result_type else self._result_type
         )
         self._cursor.get_results_from_sfqid(self.query_id)
+        if self._num_statements is not None:
+            for _ in range(self._num_statements - 1):
+                self._cursor.nextset()
+            # TODO: Once we support fetch_pandas_all for multi-statement query in connector, we could remove this
+            #   workaround.
```
sfc-gh-aalam, line 340: Should we have a JIRA for this

src/snowflake/snowpark/async_job.py
```diff
@@ -331,6 +333,15 @@ def result(
...
+            self._cursor.execute(
+                f"select * from table(result_scan('{self._cursor.sfqid}'))"
+            )
+
```
sfc-gh-aalam, line 344: I don't understand this section. we execute a result scan but whom do we share the results with for the intermediate queries.

src/snowflake/snowpark/async_job.py
sfc-gh-sfan: Created [SNOW-840818](https://snowflakecomputing.atlassian.net/browse/SNOW-840818).

src/snowflake/snowpark/async_job.py
sfc-gh-sfan: Today we cannot do `fetch_pandas_all` for multi-statement query because we need arrow result format, but the result of intermediate queries are child queries of a SQL stored proc (the backend support of multi-statement query), and SQL stored proc does not support arrow for child queries.

As far as snowpark is concerned, we only need the result of the last statement of a multi-statement query. And this is why we skip the results until the last one. Let me add more comment here.

sfc-gh-sfan: That said, we probably only need to do `result_scan` when we need pandas. Let me optimize that as well.

sfc-gh-stan (approved):

src/snowflake/snowpark/async_job.py
sfc-gh-aalam: Is this the only way to skip intermediate queries and get to the result of the last one? out of scope for this PR but we can add a way to skip to last result in python connector itself.

sfc-gh-sfan: Skipping to the last query are handled by 

```
            for _ in range(self._num_statements - 1):
                self._cursor.nextset()
```

The `result_scan` here is to make sure that we get the result in ARROW format. It can be applied to any of the statements of this query.

It does sound like a good idea to have a quick way to get the last result :)

PR #808, opened 2023-04-24
SNOW-779124 add length parameter to StringType

sfc-gh-aalam:

   Fixes #SNOW-779124

   Added the option `length` in `StringType` to specify the maximum number of characters to store.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -97,7 +97,7 @@ GENERATOR = "GENERATOR"
 ROW_COUNT = "ROWCOUNT"
 RIGHT_ARROW = " => "
 NUMBER = " NUMBER "
-STRING = " STRING "
+STRING = "STRING"
 UNSAT_FILTER = " 1 = 0 "
 BETWEEN = " BETWEEN "
 FOLLOWING = " FOLLOWING "
@@ -1302,11 +1302,12 @@ def string(length: Optional[int] = None) -> str:
     if length:
         return (
             STRING
+            + SPACE
             + LEFT_PARENTHESIS
             + str(length)
             + RIGHT_PARENTHESIS
         )
-    return STRING.strip()
+    return STRING
 
 
 def get_file_format_spec(
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -97,7 +97,7 @@ GENERATOR = "GENERATOR"
 ROW_COUNT = "ROWCOUNT"
 RIGHT_ARROW = " => "
 NUMBER = " NUMBER "
-STRING = "STRING"
+STRING = " STRING "
 UNSAT_FILTER = " 1 = 0 "
 BETWEEN = " BETWEEN "
 FOLLOWING = " FOLLOWING "
@@ -1302,12 +1302,11 @@ def string(length: Optional[int] = None) -> str:
     if length:
         return (
             STRING
-            + SPACE
             + LEFT_PARENTHESIS
             + str(length)
             + RIGHT_PARENTHESIS
         )
-    return STRING
+    return STRING.strip()
 
 
 def get_file_format_spec(
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -97,6 +97,7 @@ GENERATOR = "GENERATOR"
 ROW_COUNT = "ROWCOUNT"
 RIGHT_ARROW = " => "
 NUMBER = " NUMBER "
+STRING = " STRING "
 UNSAT_FILTER = " 1 = 0 "
 BETWEEN = " BETWEEN "
 FOLLOWING = " FOLLOWING "
@@ -1294,6 +1295,14 @@ def number(precision: int = 38, scale: int = 0) -> str:
         + str(precision)
         + COMMA
         + str(scale)
+        + RIGHT_PARENTHESIS
+    )
+
+def string(length: int = 16777216) -> str:
+    return (
+        STRING
+        + LEFT_PARENTHESIS
+        + str(length)
         + RIGHT_PARENTHESIS
     )
 
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -1298,13 +1298,15 @@ def number(precision: int = 38, scale: int = 0) -> str:
         + RIGHT_PARENTHESIS
     )
 
-def string(length: int = 16777216) -> str:
-    return (
-        STRING
-        + LEFT_PARENTHESIS
-        + str(length)
-        + RIGHT_PARENTHESIS
-    )
+def string(length: Optional[int] = None) -> str:
+    if length:
+        return (
+            STRING
+            + LEFT_PARENTHESIS
+            + str(length)
+            + RIGHT_PARENTHESIS
+        )
+    return STRING.strip()
 
 
 def get_file_format_spec(
--- a/src/snowflake/snowpark/_internal/analyzer/schema_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/schema_utils.py
@@ -16,43 +16,43 @@ from snowflake.snowpark.types import DecimalType, LongType, StringType
 
 
 def command_attributes() -> List[Attribute]:
-    return [Attribute('"status"', StringType())]
+    return [Attribute('"status"', StringType(StringType._MAX_LENGTH))]
 
 
 def list_stage_attributes() -> List[Attribute]:
     return [
-        Attribute('"name"', StringType()),
+        Attribute('"name"', StringType(StringType._MAX_LENGTH)),
         Attribute('"size"', LongType()),
-        Attribute('"md5"', StringType()),
-        Attribute('"last_modified"', StringType()),
+        Attribute('"md5"', StringType(StringType._MAX_LENGTH)),
+        Attribute('"last_modified"', StringType(StringType._MAX_LENGTH)),
     ]
 
 
 def remove_state_file_attributes() -> List[Attribute]:
-    return [Attribute('"name"', StringType()), Attribute('"result"', StringType())]
+    return [Attribute('"name"', StringType(StringType._MAX_LENGTH)), Attribute('"result"', StringType(StringType._MAX_LENGTH))]
 
 
 def put_attributes() -> List[Attribute]:
     return [
-        Attribute('"source"', StringType(), nullable=False),
-        Attribute('"target"', StringType(), nullable=False),
+        Attribute('"source"', StringType(StringType._MAX_LENGTH), nullable=False),
+        Attribute('"target"', StringType(StringType._MAX_LENGTH), nullable=False),
         Attribute('"source_size"', DecimalType(10, 0), nullable=False),
         Attribute('"target_size"', DecimalType(10, 0), nullable=False),
-        Attribute('"source_compression"', StringType(), nullable=False),
-        Attribute('"target_compression"', StringType(), nullable=False),
-        Attribute('"status"', StringType(), nullable=False),
-        Attribute('"encryption"', StringType(), nullable=False),
-        Attribute('"message"', StringType(), nullable=False),
+        Attribute('"source_compression"', StringType(StringType._MAX_LENGTH), nullable=False),
+        Attribute('"target_compression"', StringType(StringType._MAX_LENGTH), nullable=False),
+        Attribute('"status"', StringType(StringType._MAX_LENGTH), nullable=False),
+        Attribute('"encryption"', StringType(StringType._MAX_LENGTH), nullable=False),
+        Attribute('"message"', StringType(StringType._MAX_LENGTH), nullable=False),
     ]
 
 
 def get_attributes() -> List[Attribute]:
     return [
-        Attribute('"file"', StringType(), nullable=False),
+        Attribute('"file"', StringType(StringType._MAX_LENGTH), nullable=False),
         Attribute('"size"', DecimalType(10, 0), nullable=False),
-        Attribute('"status"', StringType(), nullable=False),
-        Attribute('"encryption"', StringType(), nullable=False),
-        Attribute('"message"', StringType(), nullable=False),
+        Attribute('"status"', StringType(StringType._MAX_LENGTH), nullable=False),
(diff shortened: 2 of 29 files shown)


sfc-gh-sfan (commented):
I wonder how much we could replace to use `StringType._MAX_LENGTH` instead of `16777216`.

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -92,6 +92,8 @@ def convert_sf_to_sp_type(
     if column_type_name == "BINARY":
         return BinaryType()
     if column_type_name == "TEXT":
+        if internal_size > 0 and internal_size <= StringType._MAX_LENGTH:
```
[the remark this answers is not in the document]
sfc-gh-aalam: I tested it out on server side - `Error: Invalid character length: 0. Must be between 1 and 16,777,216. (line 4)`

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1297,6 +1298,14 @@ def number(precision: int = 38, scale: int = 0) -> str:
         + RIGHT_PARENTHESIS
     )
 
+def string(length: int = 16777216) -> str:
```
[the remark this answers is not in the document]
sfc-gh-aalam: We should do that in case `_MAX_LENGTH` value changes in the future

src/snowflake/snowpark/types.py
```diff
@@ -68,7 +68,13 @@ class DateType(_AtomicType):
 class StringType(_AtomicType):
     """String data type. This maps to the VARCHAR data type in Snowflake."""
 
-    pass
+    _MAX_LENGTH = 16777216
+
+    def __init__(self, length = 16777216) -> None:
```
sfc-gh-yixie, line 73: How about using `None` as the default value?

tests/integ/scala/test_dataframe_suite.py
```diff
@@ -1615,8 +1615,8 @@ def test_createDataFrame_with_given_schema_array_map_variant(session):
     df = session.create_dataframe(data, schema)
     assert (
         str(df.schema)
-        == "StructType([StructField('ARRAY', ArrayType(StringType()), nullable=True), "
-        "StructField('MAP', MapType(StringType(), StringType()), nullable=True), "
+        == "StructType([StructField('ARRAY', ArrayType(StringType(16777216)), nullable=True), "
+        "StructField('MAP', MapType(StringType(16777216), StringType(16777216)), nullable=True), "
```
sfc-gh-yixie, line 1619: `StringType(16777216)` looks a breaking change. By default we may just use `StringType()` then it won't be a breaking change.

src/snowflake/snowpark/_internal/type_utils.py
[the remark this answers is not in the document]
sfc-gh-sfan: I wonder if we should fail if the size is not within the expected range when initializing `StringType` 🤔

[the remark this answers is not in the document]
sfc-gh-aalam: I'm following the logic followed by `DecimalType` a few lines below, which auto adjusts the input `precision` and `scale` to sensible ranges.

Right now, this function is called by values we get from result metadata. I'm not sure how reliable it is but if length value = 0 (which is possible in case when string length is not explicitly defined), then we should treat it as `StringType()` instead of raising error. 

I could possible raise error when length is negative value or great than `MAX_LENGTH`.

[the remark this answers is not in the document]
sfc-gh-sfan: Yeah negative value makes sense. I'm not particularly sure about `MAX_LENGTH` because this might change and is not controlled by client.

tests/integ/scala/test_literal_suite.py
```diff
@@ -30,8 +30,8 @@ def test_literal_basic_types(session):
     assert (
         field_str == "[StructField('ID', LongType(), nullable=False), "
         "StructField('NULL', StringType(), nullable=True), "
-        "StructField('STR', StringType(), nullable=False), "
-        "StructField('CHAR', StringType(), nullable=False), "
+        "StructField('STR', StringType(6), nullable=False), "
+        "StructField('CHAR', StringType(1), nullable=False), "
```
sfc-gh-aalam, line 34: these lengths come from `result_meta.internal_size`

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -554,13 +554,13 @@ def test_read_parquet_all_data_types_with_no_schema(session, mode):
         StructField('"N"', DecimalType(38, 6), nullable=True),
         StructField('"F"', DoubleType(), nullable=True),
         StructField('"I"', LongType(), nullable=True),
-        StructField('"S"', StringType(), nullable=True),
-        StructField('"C"', StringType(), nullable=True),
+        StructField('"S"', StringType(StringType._MAX_LENGTH), nullable=True),
+        StructField('"C"', StringType(StringType._MAX_LENGTH), nullable=True),
```
sfc-gh-aalam, line 558: the schema is inferred from the size of column stored on server side after parquet file is uploaded. In this case, since size is not specified, it is set to `StringType._MAX_LENGTH`. On the client `StringType() != StringType(StringType._MAX_LENGTH)`

src/snowflake/snowpark/_internal/analyzer/schema_utils.py
```diff
@@ -16,43 +16,43 @@
 
 
 def command_attributes() -> List[Attribute]:
-    return [Attribute('"status"', StringType())]
+    return [Attribute('"status"', StringType(StringType._MAX_LENGTH))]
```
sfc-gh-aalam, line 19: On server, these columns are not set with `STRING(LENGTH)` so the default value it takes is `STRING(16777216)`. On the client we have to explicitly set it to `StringType(_MAX_LENGTH)` because here `StringType() != StringType(_MAX_LENGTH)`.

src/snowflake/snowpark/types.py
sfc-gh-aalam: I am using `None` as default value but it creates a difference between client and server in the meaning of uninitialized string type. On server `STRING = STRING(16777216)` but on client `StringType() != StringType(16777216)`

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -97,6 +97,7 @@
 ROW_COUNT = "ROWCOUNT"
 RIGHT_ARROW = " => "
 NUMBER = " NUMBER "
+STRING = "STRING"
```
sfc-gh-jdu, line 100:
```suggestion
STRING = " STRING "
```

src/snowflake/snowpark/types.py
```diff
@@ -68,7 +68,15 @@ class DateType(_AtomicType):
 class StringType(_AtomicType):
     """String data type. This maps to the VARCHAR data type in Snowflake."""
 
-    pass
+    _MAX_LENGTH = 16777216
+
+    def __init__(self, length = None) -> None:
```
sfc-gh-jdu, line 73:
```suggestion
    def __init__(self, length: Optional[int] = None) -> None:
```

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -92,7 +92,11 @@ def convert_sf_to_sp_type(
     if column_type_name == "BINARY":
         return BinaryType()
     if column_type_name == "TEXT":
-        return StringType()
+        if internal_size > 0:
```
sfc-gh-jdu, line 95: Do you know that this internal_size is usually 0 or a positive number?

tests/integ/scala/test_dataframe_reader_suite.py
sfc-gh-jdu: Can't we make it equivalent by overwriting `__eq__`?

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-aalam: A valid column that is already created in the server will be positive number. If we try to create a number that is negative or 0, server will throw an error

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -482,14 +482,21 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
-                        precision = int(data_type_parts[1].split(",")[0])
-                        scale = int(data_type_parts[1].split(",")[1][:-1])
+                        if "," in data_type_parts[1]:
+                            precision = int(data_type_parts[1].split(",")[0])
+                            scale = int(data_type_parts[1].split(",")[1][:-1])
+                            internal_size = 0
+                        else:
+                            precision = 0
+                            scale = 0
+                            internal_size = int(data_type_parts[1][:-1])
```
sfc-gh-sfan, line 495: Curious: how is this related to the `StringType` change? 

Also, test coverage.

sfc-gh-sfan (approved):

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aalam: The size of a string column comes in result metadata inside the `internal_size` field. So we use this to create `StringType()` on the client

sfc-gh-aalam: I couldn't find a way to specify string column max length in dataframe reader format so I'll remove this section and default column size to 0

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: Then should we raise an exception when it's negative or 0?

src/snowflake/snowpark/types.py
```diff
@@ -68,7 +68,33 @@ class DateType(_AtomicType):
 class StringType(_AtomicType):
     """String data type. This maps to the VARCHAR data type in Snowflake."""
 
-    pass
+    _MAX_LENGTH = 16777216
+
+    def __init__(self, length : Optional[int] = None) -> None:
+        self.length = length
```
sfc-gh-jdu, line 74: type hint and docstring since it's a public property now

src/snowflake/snowpark/types.py
```diff
@@ -68,7 +68,33 @@ class DateType(_AtomicType):
...
+
+    def __repr__(self) -> str:
+        if self.length:
+            return f"StringType({self.length})"
+        return "StringType()"
+
+    def __eq__(self, other):
```
sfc-gh-jdu, line 81: can we add a comment here and explaining when the equality holds?

src/snowflake/snowpark/types.py
```diff
@@ -68,7 +68,33 @@ class DateType(_AtomicType):
...
+        if not isinstance(other, StringType):
+            return False
+
+        if self.length == other.length:
+            return True
+
+        if self.length in (None, StringType._MAX_LENGTH) and other.length in (
```
sfc-gh-jdu, line 88: nit: (None, StringType._MAX_LENGTH) can be a constant?

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-aalam: For negative, yes we should raise and error but as of now, we also use 0 to signify that internal size is not declared so it is a proxy for `None`.

tests/integ/scala/test_literal_suite.py
```diff
@@ -27,17 +27,30 @@ def test_literal_basic_types(session):
...
+            "StructField('BYTES', BinaryType(), nullable=False), "
+            "StructField('INT', LongType(), nullable=False), "
+            "StructField('FLOAT', DoubleType(), nullable=False), "
+            "StructField('DECIMAL', DecimalType(38, 18), nullable=False)]"
+        )
+    else:
+        assert (
+            field_str == "[StructField('ID', LongType(), nullable=False), "
+            "StructField('NULL', StringType(16777216), nullable=True), "
+            "StructField('STR', StringType(6), nullable=False), "
```
sfc-gh-aalam, line 46: we need if-else because we are comparing `str(schema)` instead of `schema`

tests/integ/scala/test_literal_suite.py
```diff
@@ -122,11 +135,18 @@ def test_special_literals(session):
...
+    if session.sql_simplifier_enabled:
+        assert (
+            str(df.schema) == "StructType([StructField('ID', LongType(), nullable=False), "
+            "StructField('NULL', StringType(), nullable=True), "
+            "StructField('LITERAL', LongType(), nullable=False)])"
+        )
+    else:
+        assert (
+            str(df.schema) == "StructType([StructField('ID', LongType(), nullable=False), "
+            "StructField('NULL', StringType(16777216), nullable=True), "
```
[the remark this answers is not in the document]
sfc-gh-aalam: @sfc-gh-yixie can this be considered behavior change? In case when sql simplified is disabled, in this test we compare `str(schema)` instead of `schemas`. With the updated `__eq__` comparing schemas still evaluates to be equal

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1297,6 +1298,16 @@ def number(precision: int = 38, scale: int = 0) -> str:
         + RIGHT_PARENTHESIS
     )
 
+def string(length: Optional[int] = None) -> str:
+    if length:
```
sfc-gh-yixie, line 1302: Can length be 0? If so you may want to use `if length is None:` here. If no this this is fine.

src/snowflake/snowpark/types.py
```diff
@@ -66,9 +66,47 @@ class DateType(_AtomicType):
...
 class StringType(_AtomicType):
-    """String data type. This maps to the VARCHAR data type in Snowflake."""
+    """String data type. This maps to the VARCHAR data type in Snowflake.
 
-    pass
+    A ``StringType`` object can be created in the following ways::
+
+        >>> string_t = StringType(23)  # this can be used to create a string type column which holds at most 23 chars
+        >>> string_t = StringType()    # this can be used to create a string type column with maximum allowed length
+        >>> string_t = StringType(StringType._MAX_LENGTH)  # same as creating StringType()
```
sfc-gh-yixie, line 75: If you expose MAX_LENGTH it should be a public API so no prefix understore

sfc-gh-yixie (approved):

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-aalam: Length can't be zero. Server raises an error if we do so

PR #818, opened 2023-05-02
SNOW-628569 Implement DataFrame.alias

sfc-gh-stan:

   Fixes SNOW-629569

   - [c] I am adding a new automated test(s) to verify correctness of my new code

   Implement dataframe.alias to enable (1) joining two dataframes with duplicate columns w/o specifying suffix, (2) self join (3) referring to columns with syntax `col(<df alias name>, <col name>)` after such joins.
   
   **TODO:**
    - Add `col(<df alias name>, "*")` support
    - Make df.copy preserve `df._select_statement`, such that self join queries can be flattened.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -335,13 +335,10 @@ class Analyzer:
                 return "*"
             else:
                 return ",".join(
-                    list(
-                        map(
-                            self.analyze,
-                            fake_col_name_to_real_col_name,
-                            expr.expressions,
-                        )
-                    )
+                    [
+                        self.analyze(e, fake_col_name_to_real_col_name)
+                        for e in expr.expressions
+                    ]
                 )
 
         if isinstance(expr, SnowflakeUDF):
@@ -654,7 +651,7 @@ class Analyzer:
 
         if isinstance(logical_plan, Selectable):
             # Selectable doesn't have children. It already has the expr_to_alias dict.
-            self.alias_maps_to_use = logical_plan.expr_to_alias
+            self.alias_maps_to_use = logical_plan.expr_to_alias.copy()
         else:
             use_maps = {}
             # get counts of expr_to_alias keys
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -144,36 +144,51 @@ from snowflake.snowpark.types import _NumericType
 ARRAY_BIND_THRESHOLD = 512
 
 
-class Analyzer:  # Why is analyzer session scoped
+class Analyzer:
     def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
         self.session = session
         self.plan_builder = SnowflakePlanBuilder(self.session)
-        self.subquery_plans = []  # reset in resolve, technically plan-scoped
-        # Used to track expression id to alias
-        self.alias_maps_to_use = (
-            None  # reset in do_resolve, technically dataframe-scoped
-        )
-        self.generated_alias_maps = {}  # reset in resolve, technically dataframe-scoped
-        self.fake_col_name_to_real_col_name = {}  # TODO: use a parameter instead
+        self.subquery_plans = []
+        self.alias_maps_to_use = None
+        self.generated_alias_maps = {}
 
     def analyze(
-        self, expr: Union[Expression, NamedExpression], parse_local_name=False
+        self,
+        expr: Union[Expression, NamedExpression],
+        fake_col_name_to_real_col_name: Dict[str, str],
+        parse_local_name=False,
     ) -> str:
         if isinstance(expr, GroupingSetsExpression):
             return grouping_set_expression(
-                [[self.analyze(a, parse_local_name) for a in arg] for arg in expr.args]
+                [
+                    [
+                        self.analyze(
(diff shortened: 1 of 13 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -2333,6 +2362,7 @@ def _join_dataframes_internal(
         lsuffix: str = "",
         rsuffix: str = "",
     ) -> "DataFrame":
+        # fake cols was set here
```
sfc-gh-stan, line 2365: TODO: clean up

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -154,83 +154,136 @@ def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
         self.alias_maps_to_use = None
 
     def analyze(
-        self, expr: Union[Expression, NamedExpression], parse_local_name=False
+        self,
+        expr: Union[Expression, NamedExpression],
+        fake_col_name_to_real_col_name: Dict[str, str],
```
sfc-gh-stan, line 159: I understand passing in `fake_col_name_to_real_col_name` is causing a lot of changes but this is how I believe recursion should behave. I don't like the approach of `expr_to_alias` where it's an attribute of the Analyzer object (which is session-scoped) and gets reset for each call to `Analyzer.resolve`, I find it complicated and hard to debug. In addition, `Analyzer.analyze` is sometimes used indepdently of `Analyzer.resolve` (e.g. in SQL simplifier), so maybe the best thing to do here is add an additional parameter called `analysis_context` and stuff all the context dependent information there?

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -464,17 +615,20 @@ def window_frame_boundary(self, offset: str) -> str:
         except Exception:
             return offset
 
-    def to_sql_avoid_offset(self, expr: Expression, parse_local_name=False) -> str:
+    def to_sql_avoid_offset(
```
sfc-gh-stan, line 618: Missing type hints.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1246,6 +1262,19 @@ def sort(
             return self._with_plan(self._select_statement.sort(sort_exprs))
         return self._with_plan(Sort(sort_exprs, self._plan))
 
+    def alias(self, name: str):
+        _copy = copy.copy(self)
+        _copy._alias = name
```
sfc-gh-aalam, line 1267: why are we using copy here instead of updating the same object?

tests/integ/test_dataframe.py
```diff
@@ -2833,3 +2833,74 @@ def test_nested_joins(session):
         key=lambda r: r[0],
     )
     assert res1 == res2 == res3
+
+
+def test_dataframe_alias(session):
+    """Test `dataframe.alias`"""
+    df1 = session.create_dataframe([[1, 6], [3, 8], [7, 7]], schema=["col1", "col2"])
+    df2 = session.create_dataframe([[1, 2], [3, 4], [5, 5]], schema=["col1", "col2"])
+
```
sfc-gh-aalam, line 2842: can you add some of the usage example to dataframe.py doc as well. This way users have some examples on how to use df.alias

src/snowflake/snowpark/dataframe.py
```diff
@@ -1246,6 +1262,19 @@ def sort(
             return self._with_plan(self._select_statement.sort(sort_exprs))
         return self._with_plan(Sort(sort_exprs, self._plan))
 
+    def alias(self, name: str):
```
sfc-gh-yixie, line 1265: As talked offline, pls mark it as `@experimental`.
@sfc-gh-mrojas Is it good for you? I'm afraid this may get user feedback and have breaking changes in the future so would like to mark it experimental first.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-yixie: How about making fake_col_name_to_real_col_name the last parameter and defaults to be `None`? It's more an incremental way than adding it in the middle so it's less likely to break existing code.

sfc-gh-yixie: Well, all calls need `fake_col_name_to_real_col_name ` so it looks right to be in the middle. Forget it.

src/snowflake/snowpark/dataframe.py
sfc-gh-mrojas: @sfc-gh-yixie I am not sure if it makes sense to have an experimental feature. I think that in that case I would prefer to not proceed and introduce automation in our conversion tools

src/snowflake/snowpark/_internal/error_message.py
```diff
@@ -236,6 +236,13 @@ def PLAN_CREATE_DYNAMIC_TABLE_FROM_DDL_DML_OPERATIONS() -> SnowparkCreateDynamic
             "1207",
         )
 
+    @staticmethod
+    def DF_ALIAS_NOT_RECOGNIZED(alias: str) -> SnowparkDataframeException:
+        return SnowparkDataframeException(
+            f"DataFrame alias unrecognized. A subset of columns corresponding to Dataframe alias {alias} can not be found. "
```
sfc-gh-yixie, line 242: Nit, add a pair of quotes `'{alias}'`

sfc-gh-yixie (approved):

PR #822, opened 2023-05-05
SNOW-809437: Add SnowparkFile to the python API docs

sfc-gh-lschmidt:

   Fixes SNOW-809437

   Adds SnowparkFile to the doc.
   
   Add supported file APIs to the overview page. Note that all methods show up on the class page so I added some basic documentation for those saying it's not supported.


sfc-gh-lschmidt:

src/snowflake/snowpark/files.py
```diff
@@ -93,58 +93,119 @@ def close(self) -> None:
         return
 
     def detach(self) -> None:
+        """
+        Not yet supported in UDF and Stored Procedures.
+
+        See https://docs.python.org/3/library/io.html#io.BufferedIOBase.detach
+        """
```
sfc-gh-skulkarni, line 100: Based on SfStream implementation this actually returns `UnsupportedOperation("Detaching stream from file is unsupported")`

src/snowflake/snowpark/files.py
```diff
@@ -93,58 +93,119 @@ def close(self) -> None:
...
         raise NotImplementedError(_DEFER_IMPLEMENTATION_ERR_MSG)
 
     def fileno(self) -> None:
+        """
+        Not yet supported in UDF and Stored Procedures.
+
+        See https://docs.python.org/3/library/io.html#io.IOBase.fileno
+        """
```
sfc-gh-skulkarni, line 108: Based on SfStream implementation, this actually returns `OSError("This object does not use a file descriptor")`

sfc-gh-skulkarni (commented):
A couple of the marked methods have server side implementations that return errors that we should probably reflect in the docs. Otherwise lgtm

src/snowflake/snowpark/files.py
sfc-gh-lschmidt: Doesn't that match with Not yet supported in UDF and Stored procedures?

src/snowflake/snowpark/files.py
sfc-gh-lschmidt: What would you recommend? I was thinking that not supported is simple enough and they probably don't care what it actually fails with? Should we change it to unsupported? Or update it to "Returns os error"?

sfc-gh-skulkarni (approved):

docs/source/files.rst
```diff
@@ -0,0 +1,43 @@
+
+
+===========================
+Files
+===========================
+
+.. automodule:: snowflake.snowpark.files
```
sfc-gh-stan, line 7: Let's remove these two lines. It's causing an extra header `Attribute` to be generated at the end of [docs/api/snowflake.snowpark.files.SnowflakeFile.html#snowflake.snowpark.files.SnowflakeFile.html](url)

sfc-gh-stan (changes requested):
I see that in the code we use the exception message `_DEFER_IMPLEMENTATION_ERR_MSG`  to indicate `SnowparkFile` does not work locally yet. We should clarify this in the documentation too.
Let's also involve our technical writer @sfc-gh-mabrennan in the review.

docs/source/files.rst
sfc-gh-lschmidt: When I remove these two lines we loose the title information, so I left them. Not sure what you are talking about with the attributes. When I ran it I don't see anything there. Unless it's something not obvious?

sfc-gh-lschmidt:
Added comments to the overall file and to SnowparkFile about how it only works locally. Let me know if this addresses the issues!

docs/source/files.rst
sfc-gh-stan: image
I was referring to the `Attributes` printed at the bottom. 
I see what you mean, since the documentation is very short, what if we just manually copy paste it in the `rst` file?

sfc-gh-stan:
> Added comments to the overall file and to SnowparkFile about how it only works locally. Let me know if this addresses the issues!

Thank you 🙂 the documentation LGTM. I have asked our technical writer @sfc-gh-mabrennan to take a quick look as well.

src/snowflake/snowpark/files.py
```diff
@@ -79,7 +85,7 @@ def open(
         Args:
             file_location: scoped URL, file URL, or string path for files located in a stage
             mode: A string used to mark the type of an IO stream.
-            is_owner_file: A boolean value, if True, the API is intended to access owner's files and all url/uri are allowed. If False, the API is intended to access files passed into the function by the caller and only scoped url is allowed.
+            is_owner_file: (Deprecated) A boolean value, if True, the API is intended to access owner's files and all url/uri are allowed. If False, the API is intended to access files passed into the function by the caller and only scoped url is allowed.
```
sfc-gh-mabrennan, line 88: url/uri -> URL/URI
and
url -> URL

sfc-gh-mabrennan (approved):
LGTM. I just added a couple of capitalization suggestions.

docs/source/files.rst
sfc-gh-lschmidt: Hrm I am a bit worried then it wouldn't get any updates which seems a bit worse. I added None to the attributes section. How is that? 
Screenshot 2023-05-10 at 11 25 09 AM

sfc-gh-stan (approved):

PR #829, opened 2023-05-09
SNOW-810566: Fix SQL simplifier on filter and sequence column 

sfc-gh-jdu:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/expression.py
+++ b/src/snowflake/snowpark/_internal/analyzer/expression.py
@@ -354,3 +354,18 @@ class ListAgg(Expression):
 
     def dependent_column_names(self) -> Optional[Set[str]]:
         return derive_dependent_columns(self.col)
+
+
+class Seq(Expression):
+    def __init__(self, width: int, sign: int) -> None:
+        super().__init__()
+        self.width = width
+        self.sign = sign
+
+    @property
+    def sql(self) -> str:
+        return f"SEQ{self.width}({self.sign})"
+
+    def dependent_column_names(self) -> Optional[Set[str]]:
+        # A sequence column can depend on all columns at the same level because the order matters.
+        return COLUMN_DEPENDENCY_ALL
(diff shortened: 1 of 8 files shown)


src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -354,3 +354,18 @@ def __init__(self, col: Expression, delimiter: str, is_distinct: bool) -> None:
 
     def dependent_column_names(self) -> Optional[Set[str]]:
         return derive_dependent_columns(self.col)
+
+
+class Seq(Expression):
```
sfc-gh-lspiegelberg, line 359: what about functions like PERCENT_RANK? https://docs.snowflake.com/en/sql-reference/functions/percent_rank are we sure they won't have similar issues?

sfc-gh-jdu: It's a window function (in this PR, we make changes to all window functions and seq[*]), so it will be covered. I'll add a test case here

sfc-gh-yixie:
Can we add some tests for window + filter?

src/snowflake/snowpark/_internal/analyzer/expression.py
sfc-gh-yixie: Two questions:
1. Do we need to consider other [Controlled Distribution](https://docs.snowflake.com/en/sql-reference/functions-data-generation) functions?
2. If a user uses `builtin("seq8")`, or `function("seq8")`, we still have this problem. How about updating `builtin` and `function` too?

sfc-gh-jdu: 1. no, there are random sequences, but we target at ordered seuqences.
2. Yea, but given we have seq8 here, I don't think users will use function("seq8"). Also I feel updating builtin and function is too restrictive. How about use this approach first, and when there is a user complaining about function("seq8"), then we update builtin and function?

PR #861, opened 2023-05-25
Snowpark reader writer revision

sfc-gh-mrojas:

   Fixes #SNOW-826851

   The PR implements extensions into the Writer/Readers API to improve developer experience


src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -253,19 +267,101 @@ class DataFrameReader:
...
+        self._metadata_cols: Optional[Iterable[_MetadataColumn]] = None
         # Infer schema information
         self._infer_schema = False
         self._infer_schema_transformations: Optional[
             List["snowflake.snowpark.column.Column"]
         ] = None
         self._infer_schema_target_columns: Optional[List[str]] = None
 
+    def load(self, path: str, format: str = None, schema: StructType = None, **kwargs):
+        """
```
sfc-gh-aalam, line 284: Can you declare a `Literal` for supported file formats and use that here and in `def format` below.

https://peps.python.org/pep-0586/

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -37,6 +39,25 @@
...
+    "TIMESTAMPFORMAT": ("TIMESTAMP_FORMAT", lambda val: val),
+}
+
+copy_options = [
+    "OVERWRITE",
+    "SINGLE",
+    "MAX_FILE_SIZE",
+    "INCLUDE_QUERY_ID",
+    "DETAILED_OUTPUT",
+]
```
sfc-gh-aalam, line 59: How is this different from `COPY_OPTIONS` in `snowflake.snowpark._internal.utils`? Would you mind adding it to the utils file instead?

sfc-gh-mrojas: Sure no problem! Good suggestion.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: Good suggestion. Thanks

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -448,7 +568,30 @@ def option(self, key: str, value: Any) -> "DataFrameReader":
...
+        - NULLVALUE: will be interpreted as `NULL_IF`.
+        - DATEFORMAT: will be interpreted as `DATE_FORMAT`.
+        - TIMESTAMPFORMAT: will be interpreted as `TIMESTAMP_FORMAT`.
+        - INFERSCHEMA: will be interpreted as `INFER_SCHEMA`.
+
         """
+        if key.upper() in ["FORMAT", "TYPE"]:
+            return self.format(value)
+        elif key.upper() == "SCHEMA":
+            return self.schema(value)
```
sfc-gh-aalam, line 590: does pyspark support `schema` and `format` via `options`?

sfc-gh-mrojas: I have that idea but will need to check. Checking the spark code format seems to be stored inside a different field and options in extraOptions but will double check

src/snowflake/snowpark/column.py
```diff
@@ -819,10 +819,21 @@ def otherwise(self, value: ColumnOrLiteral) -> "CaseExpr":
...
-METADATA_FILE_ROW_NUMBER = Column("METADATA$FILE_ROW_NUMBER")
-METADATA_FILE_CONTENT_KEY = Column("METADATA$FILE_CONTENT_KEY")
-METADATA_FILE_LAST_MODIFIED = Column("METADATA$FILE_LAST_MODIFIED")
-METADATA_START_SCAN_TIME = Column("METADATA$START_SCAN_TIME")
-METADATA_FILENAME = Column("METADATA$FILENAME")
+METADATA_FILE_ROW_NUMBER = _MetadataColumn("METADATA$FILE_ROW_NUMBER")
+METADATA_FILE_CONTENT_KEY = _MetadataColumn("METADATA$FILE_CONTENT_KEY")
+METADATA_FILE_LAST_MODIFIED = _MetadataColumn("METADATA$FILE_LAST_MODIFIED")
+METADATA_START_SCAN_TIME = _MetadataColumn("METADATA$START_SCAN_TIME")
+METADATA_FILENAME = _MetadataColumn("METADATA$FILENAME")
```
sfc-gh-aalam, line 839: These changes from metadata PR were removed later in this PR: https://github.com/snowflakedb/snowpark-python/pull/882

could you resolve these changes as well

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -40,6 +40,20 @@
 else:
     from collections.abc import Iterable
 
+# We allow some common aliases to be used
+option_aliases = {
+    "HEADER": ("SKIP_HEADER", lambda val: 1 if val else 0),
```
sfc-gh-aalam, line 45:
```suggestion
OPTION_ALIASES = {
    "HEADER": ("SKIP_HEADER", lambda val: 1 if val else 0),
```
nit: in all our files, constants are declared with all caps

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -253,19 +267,111 @@ class DataFrameReader:
...
+    def load(
+        self,
+        path: str,
+        format: Optional[
+            Literal["avro", "csv", "json", "xml", "parquet", "orc"]
+        ] = None,
+        schema: StructType = None,
+        **kwargs,
+    ):
+        """
```
sfc-gh-aalam, line 292: return type-hint

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: My tests indicated that it is not supported as part of the options.

src/snowflake/snowpark/column.py
sfc-gh-mrojas: done

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: done

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -266,6 +280,98 @@ def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
...
+
+        Alternatives to using the `format` method are:
+        1. Using the `option` method with either "format" or "type" key-value pairs.
+        Example: option("format", "csv") or option("type", "json")
+
+        2. Passing the file type directly into the `load` method.
+        Example: load("path/to/your/file", format="json")
+
+        Note: Ensure that you use a valid file format and use the appropriate method or option to specify it.
+
```
sfc-gh-aalam, line 365:
```suggestion
        Note:
            Ensure that you use a valid file format and use the appropriate method or option to specify it.
```
You could add a Note using this format. https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -448,7 +560,26 @@ def option(self, key: str, value: Any) -> "DataFrameReader":
...
+        - HEADER: Accepted values are True or False which will be interpreted as `option('SKIP_HEADER',1)` or `option('SKIP_HEADER',0)`.
+        - DELIMITER: will be interpreted as `FIELD_DELIMITER`.
+        - SEP: will be interpreted also as `FIELD_DELIMITER`.
+        - LINESEP: will be interpreted as `RECORD_DELIMITER`.
+        - PATHGLOBFILTER: will be interpreted as `PATTERN`.
+        - QUOTE: will be interpreted as `FIELD_OPTIONALLY_ENCLOSED_BY`.
+        - NULLVALUE: will be interpreted as `NULL_IF`.
+        - DATEFORMAT: will be interpreted as `DATE_FORMAT`.
+        - TIMESTAMPFORMAT: will be interpreted as `TIMESTAMP_FORMAT`.
+        - INFERSCHEMA: will be interpreted as `INFER_SCHEMA`.
```
sfc-gh-aalam, line 576:
```suggestion
        Note:
            Aliases are supported for specifying options in the `option` method.

            The following option aliases are available:

            - HEADER: Accepted values are True or False which will be interpreted as `option('SKIP_HEADER',1)` or `option('SKIP_HEADER',0)`.
            - DELIMITER: will be interpreted as `FIELD_DELIMITER`.
            - SEP: will be interpreted also as `FIELD_DELIMITER`.
            - LINESEP: will be interpreted as `RECORD_DELIMITER`.
            - PATHGLOBFILTER: will be interpreted as `PATTERN`.
            - QUOTE: will be interpreted as `FIELD_OPTIONALLY_ENCLOSED_BY`.
            - NULLVALUE: will be interpreted as `NULL_IF`.
            - DATEFORMAT: will be interpreted as `DATE_FORMAT`.
            - TIMESTAMPFORMAT: will be interpreted as `TIMESTAMP_FORMAT`.
            - INFERSCHEMA: will be interpreted as `INFER_SCHEMA`.
```
Same comment on formatting.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -266,6 +280,98 @@ def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
...
+
+        The following example demonstrate how to use a DataFrameReader.
+            >>> # Create a temp stage to run the example code.
+            >>> _ = session.sql("CREATE or REPLACE temp STAGE mystage").collect()
+
+        Example 1:
+            Loading the first two columns of a CSV file and skipping the first header line:
+            >>> from snowflake.snowpark.types import StructType, StructField, IntegerType, StringType
+            >>> _ = session.sql("create file format if not exists csv_format type=csv skip_header=1 null_if='none';").collect()
+            >>> _ = session.file.put("tests/resources/testCSVspecialFormat.csv", "@mystage", auto_compress=False)
```
sfc-gh-aalam, line 323:
```suggestion
            >>> _ = session.sql("create file format if not exists csv_format type=csv skip_header=1 null_if='none'").collect()
            >>> _ = session.file.put("tests/resources/testCSVspecialFormat.csv", "@mystage", auto_compress=False)
```

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aalam: You can also check how the final docstrings look using the readme [here](https://github.com/snowflakedb/snowpark-python/tree/main/docs).

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -219,6 +219,36 @@ def test_read_csv(session, mode):
...
+    df1 = reader.format("csv").schema(user_schema).load(test_file_on_stage)
+    res = df1.collect()
+    res.sort(key=lambda x: x[0])
+
+    assert len(res) == 2
+    assert len(res[0]) == 3
+    assert res == [Row(1, "one", 1.2), Row(2, "two", 2.2)]
+
+    with pytest.raises(SnowparkDataframeReaderException):
+        session.read.format("csv").load(test_file_on_stage)
```
sfc-gh-aalam, line 235: could you also add a `match=...` to the `pytest.raises` so that we match a regex of the error message that is generated

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -344,6 +374,47 @@ def test_read_csv_with_format_type_options(session, mode):
...
+    df1 = (
+        get_reader(session, mode)
+        .schema(user_schema)
+        .options(options)
+        .format("csv")
+        .load(test_file_colon, schema=user_schema, header=True, format="csv", sep=";")
+    )
+    res = df1.collect()
+    res.sort(key=lambda x: x[0])
+    assert res == [Row(1, "one", 1.2), Row(2, "two", 2.2)]
```
sfc-gh-aalam, line 395: could you split this test into two part, one without using `kwargs` in `load` and one without adding `options`. Here it is hard to figure out what the test is trying to test for. If it is making sure that `kwargs` get precedence over options, then we should have different values for the options

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -497,19 +568,12 @@ def test_read_metadata_column_from_stage(session, file_format):
...
-    reader = session.read.with_metadata("metadata$filename", "metadata$file_row_number")
-    df = get_df_from_reader_and_file_format(reader, file_format)
-    res = df.collect()
-    assert res[0]["METADATA$FILENAME"] == filename
-    assert res[0]["METADATA$FILE_ROW_NUMBER"] >= 0
-
 
 @pytest.mark.parametrize("mode", ["select", "copy"])
-def test_read_json_with_no_schema(session, mode):
+def test_read_format_json_with_no_schema(session, mode):
```
sfc-gh-aalam, line 573: we could parametrize this test with `.json()` and `.format("json").load()` so that we know both end points work.

tests/unit/scala/test_error_message.py
```diff
@@ -101,16 +102,6 @@ def test_df_cannot_resolve_column_name():
...
-def test_df_must_provide_schema_for_reading_file():
-    ex = SnowparkClientExceptionMessages.DF_MUST_PROVIDE_SCHEMA_FOR_READING_FILE()
-    assert type(ex) == SnowparkDataframeReaderException
-    assert ex.error_code == "1106"
-    assert (
-        ex.message
-        == "You must call DataFrameReader.schema() and specify the schema for the file."
-    )
-
-
```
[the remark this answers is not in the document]
sfc-gh-mrojas: Interesting I see no reason for that. I should have been an error during rebase

tests/integ/scala/test_dataframe_reader_suite.py
sfc-gh-mrojas: Good suggestion. splittin in 3. With aliases but using standard csv, with kwargs and using options and kwargs to show how kwargs have priority

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -266,6 +280,95 @@ def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
...
+        ] = None,
+        schema: StructType = None,
+        **kwargs,
+    ) -> DataFrame:
+        """
+        Loads data from a file path.
+
+        This method is typically the last method in a daisy chain of DataFrame reader methods.
+        It allows specifying the file type using the `format` parameter, providing a schema using the `schema` parameter,
+        and passing any copy or format-specific options using the `kwargs` parameter.
```
sfc-gh-aalam, line 297:
```suggestion
        This method is typically the last method in a daisy chain of ``DataFrame`` reader methods.
        It allows specifying the file type using the ``format`` parameter, providing a schema using the ``schema`` parameter,
        and passing any copy or format-specific options using the ``kwargs`` parameter.
```

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -266,6 +280,95 @@ def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
...
+            >>> from snowflake.snowpark.types import StructType, StructField, IntegerType, StringType
+            >>> _ = session.sql("create file format if not exists csv_format type=csv skip_header=1 null_if='none'").collect()
+            >>> _ = session.file.put("tests/resources/testCSVspecialFormat.csv", "@mystage", auto_compress=False)
+            >>> # Define the schema for the data in the CSV files.
+            >>> schema = StructType([StructField("ID", IntegerType()),StructField("USERNAME", StringType()),StructField("FIRSTNAME", StringType()),StructField("LASTNAME", StringType())])
+            >>> # Create a DataFrame that is configured to load data from the CSV files in the stage with a given schema and format.
+            >>> df = session.read.load("@mystage/testCSVspecialFormat.csv",format="csv",schema=schema,format_name="csv_format")
+            >>> # Load the data into the DataFrame and return an array of rows containing the results.
+            >>> df.collect()
+            [Row(ID=0, USERNAME='admin', FIRSTNAME=None, LASTNAME=None), Row(ID=1, USERNAME='test_user', FIRSTNAME='test', LASTNAME='user')]
```
sfc-gh-aalam, line 326: Screenshot 2023-07-05 at 10 32 43 AM

Please use [this](https://github.com/snowflakedb/snowpark-python/tree/main/docs) readme to generate docs locally and check if this was the intended result you want. You can dm me and I can help setup doc generation for you.

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -53,7 +66,363 @@ class DataFrameWriter:
...
+        if mode:
+            self.mode(mode)
+        if not self._file_format_type:
+            raise SnowparkClientExceptionMessages.DF_MUST_PROVIDE_FILETYPE_FOR_WRITING_FILE()
+        return self._write_to_location(
+            path, self._file_format_type, self._save_mode.value, block
+        )
+
+    def partitionBy(self, partition_by: Optional[ColumnOrSqlExpr]) -> "DataFrameWriter":
+        """Specifies an expression used to partition the unloaded table rows into separate files. It can be a :class:`Column`, a column name, or a SQL expression."""
```
sfc-gh-aalam, line 143:
```suggestion
    def partition_by(self, partition_by: Optional[ColumnOrSqlExpr]) -> "DataFrameWriter":
        """Specifies an expression used to partition the unloaded table rows into separate files. It can be a :class:`Column`, a column name, or a SQL expression."""
```
we generally don't use CamelCase in this repository

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -53,7 +66,363 @@ class DataFrameWriter:
...
+        """
+        if key.upper() == "PATH":
+            self._path = value
+            return self
+        elif key.upper() == "MODE":
+            return self.mode(value)
+        elif key.upper() in ["FORMAT", "TYPE"]:
+            return self.format(value)
+        elif key.upper() in ["PARTITIONBY", "PARTITION_BY"]:
+            return self.partitionBy(value)
```
sfc-gh-aalam, line 169: I would err on the side of keeping one key - `PARTITION_BY` for this unless there is a good reason why we need two.

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -53,7 +66,363 @@ class DataFrameWriter:
...
+        partition_by: Optional[ColumnOrSqlExpr] = None,
+        file_format_name: Optional[str] = None,
+        file_format_type: Optional[str] = None,
+        format_type_options: Optional[Dict[str, str]] = None,
+        header: bool = False,
+        statement_params: Optional[Dict[str, str]] = None,
+        block: bool = True,
+        **copy_options: Optional[str],
+    ) -> List[Row]:
+        ...  # pragma: no cover
```
sfc-gh-aalam, line 266: this is added twice. See lines 395-425 below.

Also, can this be moved below next to other overloaded `copy_into_location` definitions?

tests/integ/scala/test_dataframe_reader_suite.py
sfc-gh-aalam: unresolving this conversation. This change removes tests for `.json()`. can they be re-added. Even though it is called by `.load()`, we still treat implementation as back box and add tests for interface.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: Thanks. I learnt a lot about docstring formatting. I hope it is right now.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: Should be fine now

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-mrojas: noted

sfc-gh-aalam: adding an alias is fine if we need to make this compatible for snowconvert

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-mrojas: It seems that I can do this in spark:
df.write.option("header",True) \
        .option("partitionBy","ts")\
        .mode("overwrite") \
        .csv("/tmp/zipcodes-state")
so leaving this will handle that case

sfc-gh-mrojas: @sfc-gh-aalam let me know if you prefer to remove it

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -497,19 +568,12 @@ def test_read_metadata_column_from_stage(session, file_format):
     res = df.collect()
     assert res[0]["FILENAME"] == filename
 
-    # test that column name with str works
-    reader = session.read.with_metadata("metadata$filename", "metadata$file_row_number")
-    df = get_df_from_reader_and_file_format(reader, file_format)
-    res = df.collect()
-    assert res[0]["METADATA$FILENAME"] == filename
-    assert res[0]["METADATA$FILE_ROW_NUMBER"] >= 0
-
```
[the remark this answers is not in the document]
sfc-gh-mrojas: added back

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-aalam: then let's keep it.

sfc-gh-aalam: we also might have to update line 177 to use `self.partition_by` or create an alias

sfc-gh-aalam (approved):

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -43,6 +43,20 @@
...
+    "DELIMITER": ("FIELD_DELIMITER", lambda val: val),
+    "SEP": ("FIELD_DELIMITER", lambda val: val),
+    "LINESEP": ("RECORD_DELIMITER", lambda val: val),
+    "PATHGLOBFILTER": ("PATTERN", lambda val: val),
+    "QUOTE": ("FIELD_OPTIONALLY_ENCLOSED_BY", lambda val: val),
+    "NULLVALUE": ("NULL_IF", lambda val: val),
+    "DATEFORMAT": ("DATE_FORMAT", lambda val: val),
+    "TIMESTAMPFORMAT": ("TIMESTAMP_FORMAT", lambda val: val),
+    "INFERSCHEMA": ("INFER_SCHEMA", lambda val: val),
+}
```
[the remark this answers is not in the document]
sfc-gh-aalam: @sfc-gh-yuliu If you could review this section

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -43,6 +43,20 @@
 
 logger = getLogger(__name__)
 
+# We allow some common aliases to be used
+OPTION_ALIASES = {
+    "HEADER": ("SKIP_HEADER", lambda val: 1 if val else 0),
```
sfc-gh-yuliu, line 48: Just FYI, we are introducing another option called `PARSE_HEADER` which will parse the first record and persist the column names. But it's only working for limited number of features rn

sfc-gh-yuliu: https://docs.snowflake.com/en/sql-reference/functions/infer_schema#usage-notes

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -43,6 +43,20 @@
...
+    "DELIMITER": ("FIELD_DELIMITER", lambda val: val),
+    "SEP": ("FIELD_DELIMITER", lambda val: val),
```
sfc-gh-yuliu, line 50: both will map to `FIELD_DELIMITER`? I wonder how this mapping is determined. Are those aliases popular/required somewhere?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -43,6 +43,20 @@
...
+    "LINESEP": ("RECORD_DELIMITER", lambda val: val),
```
sfc-gh-yuliu, line 51: I'm a bit concerned about the `LINESEP` since if the RECORD_DELIMITER is not a new line delimiter (aka `\n`), there's no concept of a line.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -43,6 +43,20 @@
...
+    "PATHGLOBFILTER": ("PATTERN", lambda val: val),
+    "QUOTE": ("FIELD_OPTIONALLY_ENCLOSED_BY", lambda val: val),
+    "NULLVALUE": ("NULL_IF", lambda val: val),
```
sfc-gh-yuliu, line 54: NULL_IF can be an array, would it cause an issue here?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -43,6 +43,20 @@
...
+    "DATEFORMAT": ("DATE_FORMAT", lambda val: val),
+    "TIMESTAMPFORMAT": ("TIMESTAMP_FORMAT", lambda val: val),
+    "INFERSCHEMA": ("INFER_SCHEMA", lambda val: val),
```
sfc-gh-yuliu, line 57: INFER_SCHEMA is not an option but a table function, can you educate me on how these would be used?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -43,6 +43,20 @@
...
+OPTION_ALIASES = {
+    "HEADER": ("SKIP_HEADER", lambda val: 1 if val else 0),
+    "DELIMITER": ("FIELD_DELIMITER", lambda val: val),
+    "SEP": ("FIELD_DELIMITER", lambda val: val),
+    "LINESEP": ("RECORD_DELIMITER", lambda val: val),
+    "PATHGLOBFILTER": ("PATTERN", lambda val: val),
+    "QUOTE": ("FIELD_OPTIONALLY_ENCLOSED_BY", lambda val: val),
+    "NULLVALUE": ("NULL_IF", lambda val: val),
+    "DATEFORMAT": ("DATE_FORMAT", lambda val: val),
+    "TIMESTAMPFORMAT": ("TIMESTAMP_FORMAT", lambda val: val),
```
sfc-gh-yuliu, line 56: Do you also want to add TIME_FORMAT?

Just to double confirm, these will be only used in the data load scenarios right? Asking since theses options won't take effect in other scenarios.

sfc-gh-yuliu (commented):
some comments and questions. Thanks for keeping us posted

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -53,7 +66,299 @@ class DataFrameWriter:
...
+        This method is typically the last method in a daisy chain of :class:`DataFrameWriter` methods.
+        It allows specifying the file type using the ``format`` parameter, providing a schema using the ``schema`` parameter,
+        and passing any copy or format-specific options using the ``kwargs`` parameter.
+
+        Alternatively, you can use the ``format`` method and specify the format, and then use the ``save`` method to specify the path:
+
+            ``df = spark.read.format("csv").option("header", "true").save("path/to/your/file.csv")``
+
+        Args:
+            path: The path to the file(s) to load.
```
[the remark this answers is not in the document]
sfc-gh-mrojas: @sfc-gh-yixie this methods follow the Spark API conventions. In Spark API there is load but not unload

src/snowflake/snowpark/dataframe_reader.py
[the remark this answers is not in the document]
sfc-gh-mrojas: @sfc-gh-yixie just for context. These aliases are mostly for compatibility in migration scenarios. these keywords are valid in spark and we just want to avoid friction when migrating

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: In spark header is pretty basic, it is just to skip first line

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: This comes from usages like these: https://sparkbyexamples.com/spark/spark-read-options/?expand_article=1

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: mm it should not as it will just pass the same value.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mrojas: these only apply in the context of DataFrameReader options not in any other scenarios. They are provided just for backwards compatibility and mapped to the closest equivalent in the current copy options

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aalam: If `INFER_SCHEMA` is set to true, we run an infer_schema query to determine the column names/datatypes for the subsequent dataframe generation

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -53,7 +66,299 @@ class DataFrameWriter:
...
+        This method is typically the last method in a daisy chain of :class:`DataFrameWriter` methods.
+        It allows specifying the file type using the ``format`` parameter, providing a schema using the ``schema`` parameter,
+        and passing any copy or format-specific options using the ``kwargs`` parameter.
+
+        Alternatively, you can use the ``format`` method and specify the format, and then use the ``save`` method to specify the path:
+
+            ``df = session.read.format("csv").option("header", "true").save("path/to/your/file.csv")``
+
+        Args:
+            path: stage location for the file(s) to be loaded
```
sfc-gh-jfreeberg, line 99: The API signature indicates that the path is an optional parameter. If that's true, then what's the default value?

sfc-gh-aalam: This can be set by `option` method but there is still confusion here. We don't check if the value is not NULL and this can lead to error without a good error message. We should improve the error checking here

PR #788, opened 2023-04-18
SNOW-781049: update test to be compatible with stored proc

sfc-gh-aling:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-781049, the tests previous PR are incompatible with stored proc because it can not create randomly generated temp table name.

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

pre-commit test: https://buwa.ci1.us-west-2.aws-dev.app.snowflake.com/builds/precommit/ci-dev-133.int.snowflakecomputing.com/9897

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

PR #772, opened 2023-04-06
SNOW-781049: support dot in table name

sfc-gh-aling:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-781049

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

PR #871, opened 2023-06-01
SNOW-830758: revert code change in table exist detection

sfc-gh-aling:

   Fixes SNOW-830758

revert the change introduced:

https://github.com/snowflakedb/snowpark-python/pull/772/files
https://github.com/snowflakedb/snowpark-python/pull/788/files

We don't want to  break users as existing users depend on the behavior of auto parse table name


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 
 - Fixed a bug when `Dataframe.join_table_function` did not run all the necessary queries to setup join table function when sql simplifier was enabled.
 - Added comments to SnowflakeFile and added a API reference page for it.
+- Reverted the change of `DataFrame.save_as_table` supporting table name containing dots which was introduced in v1.4.0 as it caused behavior change.
```
sfc-gh-stan, line 27: We could link the PR number here.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -539,20 +529,17 @@ def join(
 
     def save_as_table(
         self,
-        table_name: Union[str, Iterable[str]],
```
sfc-gh-stan, line 542: We should make sure `Iterable[str]` also works.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -539,20 +529,17 @@ def join(
...
+        table_name: str,
```
sfc-gh-aalam, line 532: The workaround given to customers is `save_as_table([db_name, schema_name, table_name])`. Will that still work if we revert this?

sfc-gh-stan: Since the release is delayed, can we not revert and try to fix this bug for when `table_name` is `str`?

sfc-gh-aalam: Short version is we can't, to fix the `str` case, we have to go BCR route.

When `table_name` is `str`, it is not possible for us to tell whether the user intends `schema.table_name` or `table.name`. If we want to fix the case where we allow `table_name` with dots, we must change behavior such that 
- if users want to specify `schema.table_name`, they now use `save_as_table([schema, table_name])`
- if the users want to specify `table.name`, they use `save_as_table("table.name")`

For the above, we have to follow BCR process.

src/snowflake/snowpark/session.py
```diff
@@ -2062,45 +2061,41 @@ def query_history(self) -> QueryHistory:
...
-                    f"show tables like '{strip_double_quotes_in_like_statement_in_table_name(qualified_table_name[2])}' in {condition}"
-                )
-            else:
-                # we do not support len(qualified_table_name) > 3 for now
-                raise SnowparkClientExceptionMessages.GENERAL_INVALID_OBJECT_NAME(
-                    ".".join(raw_table_name)
-                )
+            # we do not support len(qualified_table_name) > 3 for now
+            raise SnowparkClientExceptionMessages.GENERAL_INVALID_OBJECT_NAME(
+                table_name
```
sfc-gh-aalam, line 2097: Let's add a test to cover this exception

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: ^ I am wrong here. Looks like there is a way to decipher using quotes. For example, we allow something like `"schema.name"."table.name"` to allow dots in both schema name and table name.

tests/integ/scala/test_dataframe_copy_into.py
```diff
@@ -1219,62 +1219,3 @@ def test_copy_into_table_non_csv_using_options(session, tmp_stage_name1):
...
-            assert session.table(table_name).count() == 0
-
-            df = session.read.schema(user_schema).csv(test_file_on_stage)
-            df.copy_into_table(table_name)
-            Utils.check_answer(
-                session.table(table_name),
-                [Row(1, "one", 1.2), Row(2, "two", 2.2)],
-            )
-        finally:
-            session.sql(f"drop table if exists {full_table_name or table_name}")
```
sfc-gh-aalam, line 1244: This drop might not be working as expected. I see many temp tables in sfctest that have 3 dots in name `...`. When this test is re-added, let's make sure we drop the table name we created. Let me know if I am wrong here tho. This is the first test I found which is adding 3 dots in table name

sfc-gh-aling:
we should not need this PR, this PR should kill two birds with one stone: https://github.com/snowflakedb/snowpark-python/pull/878

PR #881, opened 2023-06-08
SNOW-834932: Added execute_as parameter to StoredProcedureRegistration.register_from_file() function

sfc-gh-vnayak:
Added execute_as parameter to StoredProcedureRegistration.register_from_file() function

N/A

Added execute_as parameter to StoredProcedureRegistration.register_from_file() function. This allows the stored procedure to be called with either 'owner' or 'caller' rights.

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -73,6 +73,8 @@ _MAX_INLINE_CLOSURE_SIZE_BYTES = 8192
 # Every table function handler class must define the process method.
 TABLE_FUNCTION_PROCESS_METHOD = "process"
 
+EXECUTE_AS_WHITELIST = frozenset(["owner", "caller"])
+
 
 class UDFColumn(NamedTuple):
     datatype: DataType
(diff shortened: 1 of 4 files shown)


sfc-gh-sfan (approved):

sfc-gh-aalam (approved):
LGTM

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -173,6 +175,17 @@ def check_register_args(
...
 
+def check_execute_as_arg(execute_as: typing.Literal["caller", "owner"]):
+    if (
+        not isinstance(execute_as, str)
+        or execute_as.lower() not in EXECUTE_AS_WHITELIST
+    ):
+        raise TypeError(
+            f"'execute_as' value '{execute_as}' is invalid, choose from "
+            f"{', '.join(EXECUTE_AS_WHITELIST, )}"
+        )
```
sfc-gh-mkeller, line 186: You communicate in your type-hint that it only accepts either `"caller"`, or `"owner"` as input and then you check whether it's something else.
Think of type-hints as a contract, you the writer of the function promise that the function works for these inputs and gives this type back.
If I take your type-hints into account the check itself makes no sense. I'd like if we removed this altogether.

sfc-gh-aalam: This check was already present in the old code for checking `execute_as` args. Now if it is removed, is that a behavior change? Also, I believe the old code should've raise ValueError instead of TypeError

PR #900, opened 2023-06-20
SNOW-840549: Auto-upload Python packages that are not available in Snowflake (V1)

sfc-gh-vnayak:

https://snowflakecomputing.atlassian.net/browse/SNOW-824816

   - [x] I am adding a new dependency


src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -519,146 +510,6 @@ def {_DEFAULT_HANDLER_NAME}({args}):
 """.strip()
 
 
-def resolve_imports_and_packages(
```
sfc-gh-vnayak, line 522: Moved to packaging_utils.py

tests/integ/test_udf.py
```diff
@@ -1496,10 +1542,6 @@ def test_add_packages_negative(session, caplog):
         session.remove_package("python-dateutil")
     assert "is not in the package list" in str(ex_info)
 
-    with pytest.raises(ValueError) as ex_info:
```
sfc-gh-vnayak, line 1499: Unavailable packages no longer have a straightforward error; this case is covered in other tests.

tests/integ/test_udf.py
```diff
@@ -1398,34 +1409,41 @@ def get_numpy_pandas_dateutil_version() -> str:
...
     )
 
-    # only add numpy, which will overwrite the previously added packages
-    # so pandas will not be available on the server side
-    def is_pandas_available() -> bool:
+    # only add pyyaml, which will overwrite the previously added packages
+    # so matplotlib will not be available on the server side
+    def is_matplotlib_available() -> bool:
         try:
-            import pandas  # noqa: F401
```
sfc-gh-vnayak, line 1405: Need to look into this test; numpy and pandas are always available.

tests/integ/test_udf.py
```diff
@@ -1515,6 +1557,36 @@ def test_add_requirements(session, resources_path):
     assert session.get_packages() == {
         "numpy": "numpy==1.23.5",
         "pandas": "pandas==1.5.3",
+        "snowflake-snowpark-python": "snowflake-snowpark-python",
```
sfc-gh-vnayak, line 1560: New behavior - Snowpark is added by default.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,467 @@
...
+    "pandas",
+    "pyarrow",
+    "python-dateutil",
+    "pytz",
+    "cloudpickle",
+    "streamlit",  # not default, but is available on import
+}
+
+
+def resolve_imports_and_packages(
```
sfc-gh-vnayak, line 55: Moved here from udf_utils.py

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,467 @@
...
+                    f"dependency, switching to latest available version "
+                    f"{valid_packages[package_name][-1]} instead."
+                )
+                dropped_dependencies.append(package)
+                new_dependencies.append(Requirement.parse(package_name))
+
+    return supported_dependencies, dropped_dependencies, new_dependencies
+
+
+# if package_name in valid_packages:
```
sfc-gh-vnayak, line 305: Original code from SnowCLI:- any dependency that can be used via Anaconda channel is used despite versioning (I changed the code so that we now only use Anaconda if it's the right version or if it is a native dependency)

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,467 @@
...
+            if (
+                file.is_file()
+                and not file.match(".*")
+                and file != output_path
+                and file != directory_path
+            ):
+                zipf.write(file, file.relative_to(parent_directory))
+
+
+#     for file in files:
```
sfc-gh-vnayak, line 418: To be removed in final clean-up commit

setup.py
```diff
@@ -52,6 +52,7 @@
         "cloudpickle>=1.6.0,<=2.0.0",
         f"snowflake-connector-python{CONNECTOR_DEPENDENCY_VERSION}",
         "typing-extensions>=4.1.0, <5.0.0",
+        "pyyaml>=6.0",
```
sfc-gh-vnayak, line 55: Required to parse Conda environment YAMLs, for adding requirements (as requested in the project JIRA).

src/snowflake/snowpark/session.py
```diff
@@ -830,19 +857,39 @@ def add_requirements(self, file_path: str) -> None:
...
+                for line in f:
+                    package = line.rstrip()
+                    if package:
+                        packages.append(package)
+        elif file_path.endswith(".yml") or file_path.endswith(".yaml"):
+            with open(file_path) as f:
+                try:
+                    environment_data = yaml.safe_load(f)
+                    dependencies = environment_data.get("dependencies", [])
+                    is_python_dependency = (
```
sfc-gh-vnayak, line 871: We ignore Python itself for now; it will be considered in the next iteration, wherein the RUNTIME_VERSION parameter while creating procedures can be modified.

src/snowflake/snowpark/session.py
```diff
@@ -977,8 +1025,128 @@ def get_req_identifiers_list(
...
+            raise ValueError(
+                f"Unable to auto-upload packages: {packages}, Error: {e} | NOTE: Alternatively, you can find the "
+                f"directory of these packages and add it via session.add_import(). See details at "
+                f"https://docs.snowflake.com/en/developer-guide/snowpark/python/creating-udfs.html#using"
+                f"-third-party-packages-from-anaconda-in-a-udf."
+            )
+
+        return supported_dependencies + new_dependencies
+
+    def _is_anaconda_terms_acknowledged(self) -> bool:
```
sfc-gh-vnayak, line 1147: Abstracted out so for easy testing

sfc-gh-vnayak:
Investigating test failures due to coverage tool issues.

sfc-gh-stan:
There seems to be a rebase issue?

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+                    return_dict[package_req] = included_record_entries
+    return return_dict
+
+
+def identify_supported_packages(
+    packages: List[Requirement],
+    valid_packages: Dict[str, List[str]],
+    native_packages: Set[str],
+) -> Tuple[List[Requirement], List[Requirement], List[Requirement]]:
+    supported_dependencies: List = []
```
sfc-gh-stan, line 107: Let's document what the function does and the meaning of the parameters. It'll be great if we can briefly describe what valid/native packages are.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+                ] = f"{SNOWPARK_PACKAGE_NAME}=={package_client_version}"
+            else:
+                _logger.warning(
+                    f"The version of package '{SNOWPARK_PACKAGE_NAME}' in the local environment is "
+                    f"{package_client_version}, which is not available in Snowflake. Your UDF might not work when "
+                    f"the package version is different between the server and your local environment."
+                )
+        except pkg_resources.DistributionNotFound:
+            _logger.warning(
+                f"Package '{SNOWPARK_PACKAGE_NAME}' is not installed in the local environment. "
```
sfc-gh-stan, line 242: When will this case happen?

src/snowflake/snowpark/session.py
```diff
@@ -869,81 +904,69 @@ def _resolve_packages(
...
                     else:
-                        detailed_err_msg = (
-                            "Anaconda terms must be accepted by ORGADMIN to use "
-                            "Anaconda 3rd party packages. Please follow the instructions at "
+                        version_text = (
+                            f"(version {package_version_req})"
+                            if package_version_req is not None
+                            else ""
+                        )
+                        raise ValueError(
```
sfc-gh-stan, line 948: Maybe `RuntimeError`? I'm not sure if `ValueError` is the most appropriate here.

tests/integ/test_udf.py
```diff
@@ -1528,6 +1563,158 @@ def get_numpy_pandas_version() -> str:
...
+            return "does not work"
+
+    Utils.check_answer(
+        session.sql(f"select {udf_name}()").collect(),
+        [Row("TimeSeriesForestClassifier(n_estimators=5)")],
+    )
+
+
+def test_add_requirements_with_native_dependency_without_force_push(session):
+    session._is_anaconda_terms_acknowledged = lambda: True
```
sfc-gh-stan, line 1705: Let's patch it.

tests/integ/test_udf.py
sfc-gh-stan: Just to confirm: this will not a breaking change since we always allow `snowflake-snowpark-python` to be overwritten?

tests/integ/test_udf.py
```diff
@@ -86,6 +86,17 @@ def setup(session, resources_path):
     )
 
 
+@pytest.fixture(autouse=True)
+def clean_up(session):
+    session.clear_packages()
+    session.clear_imports()
+    acknowledgement_function = session._is_anaconda_terms_acknowledged
```
sfc-gh-stan, line 93: We can avoid this by patching.

tests/integ/test_udf.py
sfc-gh-vnayak: Yeah, the version can always be overwritten. I will add a specific test case to show that this is true.

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Unable to think of any case where this could happen but if it does happen due to an off-beat failure (let's say a bug in pkg_resources), it should not affect the user.

tox.ini
```diff
@@ -4,6 +4,7 @@ show_missing = True
 [coverage:run]
 branch = true
 parallel = true
+concurrency = multiprocessing
```
sfc-gh-vnayak, line 7: This configuration change helps prevent coverage job errors caused by the usage of subprocess module. See details at https://github.com/pytest-dev/pytest-cov/issues/243

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+from pkg_resources import Requirement
+
+_logger = getLogger(__name__)
+PIP_ENVIRONMENT_VARIABLE = "PIP_NAME"
+IMPLICIT_ZIP_FILE_NAME = "zipped_packages"
+SNOWPARK_PACKAGE_NAME = "snowflake-snowpark-python"
+
+# Default that we do not want to upload because we wish to use stable versions instead
+# TODO: This is very hacky, how can errors with default packages be avoided?
+DEFAULT_PACKAGES = {
```
sfc-gh-sfan, line 26: How are default packages selected? It looks like some packages should already be handled by native package detection, so we could probably drop them from here :)

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+def get_package_name_from_metadata(metadata_file_path: str) -> Optional[str]:
+    """Loads a METADATA file from the dist-info directory of an installed
+    Python package, finds the name of the package.
+    This is found on a line containing "Name: my_package".
+
+    Args:
+        metadata_file_path (str): The path to the METADATA file
+
+    Returns:
+        str: the name of the package.
```
sfc-gh-sfan, line 46: The return value could also contain version. Let's add that to the docstring as well. Ideally something like `package[==1.0.0]`?

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+                            os.path.split(line)[0].split(",")[0]
+                            for line in record_file.readlines()
+                        },
+                    )
+                    included_record_entries = []
+                    for record_entry in record_entries:
+                        record_entry_full_path = os.path.abspath(
+                            os.path.join(directory, record_entry),
+                        )
+                        # RECORD file might contain relative paths to items outside target folder. (ignore these)
```
sfc-gh-sfan, line 90: Do you mean we want to ignore paths outside of target folder?

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+            elif package_name in native_packages:
+                # Native packages should anaconda dependencies if possible, even if the version is not available
+                _logger.warning(
+                    f"Package {package_name}(version {package_version_req}) is an unavailable native "
+                    f"dependency, switching to latest available version "
+                    f"{valid_packages[package_name][-1]} instead."
+                )
+                dropped_dependencies.append(package)
+                new_dependencies.append(Requirement.parse(package_name))
+                native_packages.remove(package_name)
```
sfc-gh-sfan, line 127: Why do we remove it from `native_packages`?

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+        raise ValueError(f"Pip failed with return code {pip_install_result}.")
+
+
+def detect_native_dependencies(
+    target: str, downloaded_packages_dict: Dict[Requirement, List[str]]
+) -> Set[str]:
+    def invert_package_map(
+        downloaded_packages_dict: Dict[Requirement, List[str]]
+    ) -> Dict[str, Set[str]]:
+        inverted_dictionary: Dict[str, Set[str]] = {}
```
sfc-gh-sfan, line 171: Maybe name this `root_folder_to_package_name`? It's unclear what `inverted` means.

src/snowflake/snowpark/session.py
```diff
@@ -952,15 +975,26 @@ def _resolve_packages(
...
             else:
                 result_dict[package_name] = package
 
+        dependency_packages = None
+        if len(unsupported_packages) != 0:
+            _logger.warning(
+                f"The following packages are not available in Snowflake: {unsupported_packages}. They "
+                f"will be uploaded to session stage."
+            )
+            dependency_packages = self._upload_unsupported_packages(
```
sfc-gh-sfan, line 994: We should skip the auto-upload when it is in stored proc. Some of the steps are not supported in stored proc (e.g. `Popen`), also the unsupported packages are already uploaded from client, and the zip files are inherited from stored proc to its child jobs.

sfc-gh-sfan (changes requested):
I guess you would need to skip some integ test for stored proc. Please make sure you run the test job :)

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Yes, all package specific files would be downloaded in the target folder itself as pip does not install anything outside the specified target folder. However, the RECORD file might be malformed.

src/snowflake/snowpark/session.py
sfc-gh-vnayak: The pip install `Popen` would occur on the client side; would that still affect the stored proc?

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Because we have found a supported version of it on Anaconda, it is no longer problematic (native packages are a problem if they are unsupported only). The native_packages variable is used later to raise an error about unsupported native packages.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,296 @@
...
+
+    metadata_files = glob.glob(os.path.join(directory, "*dist-info", "METADATA"))
+    package_name_to_record_entries_map: Dict[Requirement, List[str]] = {}
+    for metadata_file in metadata_files:
+        parent_folder = os.path.dirname(metadata_file)
+        package = get_package_name_from_metadata(metadata_file)
+
+        if package is not None:
+            # Determine which folders or files belong to this package
+            record_file_path = os.path.join(parent_folder, "RECORD")
```
sfc-gh-aalam, line 70: Adding a comment here on the function of record files and record entries will be useful context

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,296 @@
...
+                            _logger.info(f"Potential native library: {library}")
+                            native_libraries.add(library)
+    return native_libraries
+
+
+def zip_directory_contents(directory_path: str, output_path: str) -> None:
+    """
+    All files/folders inside the directory path are copied over into the zip package.
+    All files/folders installed outside the directory path, which are not the directory path or the output path or
+    hidden are also copied over.
```
sfc-gh-aalam, line 240: It looks like by outside, you mean one level up of directory path? The current description is vague and does not convey that.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,294 @@
...
+        if results is not None:
+            version = results.group(1)
+            requirement_line += f"=={version}"
+        return requirement_line.strip().lower()
+
+
+def get_downloaded_packages(directory: str) -> Dict[Requirement, List[str]]:
+    """
+    This function records correspondence between installed python packages and their folder structure.
+    When redundant packages are detected, we can use this correspondence to delete installed python packages.
```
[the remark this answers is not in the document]
sfc-gh-vnayak: The dependency packages for an unsupported package could be present in the Anaconda channel. 'Redundant' here means that uploading these dependency packages via zip file would be redundant (instead can import the package directly).

Should the docstring be changed to 'When packages supported by Anaconda are detected'

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: True, this is no longer needed because of improved native packager detection

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Thanks, adding to docstring

src/snowflake/snowpark/session.py
sfc-gh-vnayak: Does skipping auto upload if inside stored proc only apply during testing? (Or do you mean that I need to add code here in `session.py` to skip)

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+import pkg_resources
+from pkg_resources import Requirement
+
+_logger = getLogger(__name__)
+PIP_ENVIRONMENT_VARIABLE = "PIP_NAME"
+IMPLICIT_ZIP_FILE_NAME = "zipped_packages"
+SNOWPARK_PACKAGE_NAME = "snowflake-snowpark-python"
+
+# Default that we do not want to upload because we wish to use stable versions instead
+# TODO: This is very hacky, how can errors with default packages be avoided?
```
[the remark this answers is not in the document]
sfc-gh-vnayak: There was an error about files missing from pandas module, but better native dependency checks was the solution to that. There is no issue with default packages.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+                    f"dependency, switching to latest available version "
+                    f"{valid_packages[package_name][-1]} instead."
+                )
+                dropped_dependencies.append(package)
+                new_dependencies.append(Requirement.parse(package_name))
+
+    return supported_dependencies, dropped_dependencies, new_dependencies
+
+
+def install_pip_packages_to_target_folder(packages: List[str], target: str):
```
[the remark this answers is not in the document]
sfc-gh-vnayak: Thanks, changed

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Thanks, changed the docstring

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,251 @@
...
+        process_output = "\n".join([line.strip() for line in process.stdout])
+        _logger.debug(process_output)
+    except FileNotFoundError:
+        raise ModuleNotFoundError(
+            f"Pip not found. Please install pip in your environment or specify the "
+            f"{PIP_ENVIRONMENT_VARIABLE} environment variable and try again!"
+        )
+
+    if pip_install_result is not None and pip_install_result != 0:
+        raise ValueError(f"Pip failed with return code {pip_install_result}.")
```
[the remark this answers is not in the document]
sfc-gh-vnayak: Chose `RuntimeError` here

tests/integ/test_udf.py
sfc-gh-vnayak: test_add_requirements_and_override_snowpark_package added

src/snowflake/snowpark/session.py
```diff
@@ -385,6 +399,8 @@ def _generate_new_action_id(self) -> int:
 
     def close(self) -> None:
         """Close this session."""
+        if self._tmpdir_handler:
```
sfc-gh-vnayak, line 402: Cleanup temporary directory used to pip install packages.

src/snowflake/snowpark/session.py
```diff
@@ -854,6 +887,8 @@ def _resolve_packages(
                 use_local_version = True
             else:
                 package = package.strip().lower()
+                if package.startswith("#"):
```
sfc-gh-vnayak, line 890: Ignore commented lines

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,376 @@
...
+
+        if package is not None:
+            # Determine which folders or files belong to this package
+            record_file_path = os.path.join(parent_folder, "RECORD")
+            if os.path.exists(record_file_path):
+                # Get unique root folder names
+                with open(record_file_path, encoding="utf-8") as record_file:
+                    record_entries = set()
+
+                    # Read in all record entries
```
sfc-gh-sfan, line 92: Can you add a comment to provide an example of a line in the RECORD file?

src/snowflake/snowpark/session.py
```diff
@@ -974,12 +1012,143 @@ def get_req_identifiers_list(
...
+                    if result_dict[name] != str(package):
+                        raise ValueError(
+                            f"Cannot add dependency package '{name}'{'( version '+version+')' if version else ''} "
+                            f"because {result_dict[name]} is already added."
+                        )
+                else:
+                    result_dict[name] = str(package)
+
+        # Add the Snowpark package to packages dict, based on client's environment (or latest), if not already added
+        add_snowpark_package(result_dict, valid_packages)
```
sfc-gh-sfan, line 1032: Curious: why do we always need to add snowpark packages?

src/snowflake/snowpark/session.py
```diff
@@ -831,19 +858,25 @@ def add_requirements(self, file_path: str) -> None:
...
         packages = []
-        with open(file_path) as f:
-            for line in f:
-                package = line.rstrip()
-                if package:
-                    packages.append(package)
-        self.add_packages(packages)
+        if file_path.endswith(".txt"):
+            with open(file_path) as f:
+                for line in f:
```
sfc-gh-aalam, line 863: it's still possible to have a text file without a `.txt` extension

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,376 @@
...
+
+    Note that we also update the `native_packages` set to reflect genuinely problematic native dependencies, i.e.
+    packages that are not present in Anaconda and are likely to cause errors.
+
+    Args:
+        packages (List[Requirement]): List of python packages.
+        valid_packages (Dict[str, List[str]): Mapping from package name to a list of versions available on the Anaconda
+        channel.
+        native_packages (Set[str]): Set of native dependency package names.
+
```
sfc-gh-aalam, line 139:
```suggestion
        native_packages (Set[str]): Set of package names that have a native dependency.

```
how about this?

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,376 @@
...
+    Detects which `packages` are present in the Snowpark Anaconda channel using the `valid_packages` mapping.
+    If a package is a native dependency (belongs to `native_packages` set) and supported in Anaconda, we switch to
+    the latest available version in Anaconda.
+
+    Note that we also update the `native_packages` set to reflect genuinely problematic native dependencies, i.e.
+    packages that are not present in Anaconda and are likely to cause errors.
+
+    Args:
+        packages (List[Requirement]): List of python packages.
+        valid_packages (Dict[str, List[str]): Mapping from package name to a list of versions available on the Anaconda
```
sfc-gh-aalam, line 136: packages: Are they the list of packages requested by the user to be included in session?

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-aalam: are they package names requested by the user that we have detected to have a native dependency?

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,376 @@
...
+                package_version_required is None
+                or package_version_required in valid_packages[package_name]
+            ):
+                supported_dependencies.append(package)
+
+            # Native packages should be anaconda dependencies, even if the requested version is not available.
+            elif package_name in native_packages:
+                _logger.warning(
+                    f"Package {package_name}(version {package_version_required}) is an unavailable native "
+                    f"dependency, switching to latest available version "
```
sfc-gh-aalam, line 167:
```suggestion
                    f"Package {package_name}(version {package_version_required}) has an unavailable native "
                    f"dependency, switching to latest available version "
```
Also, `package_version_required` can be `None` here, right?

tests/integ/test_stored_procedure.py
```diff
@@ -1150,3 +1159,62 @@ def test_anonymous_stored_procedure(session):
...
+        rule1 = ctrl.Rule(temperature["cold"] & humidity["dry"], fan_speed["low"])
+        rule2 = ctrl.Rule(temperature["warm"] & humidity["moist"], fan_speed["high"])
+
+        # Create fuzzy control system
+        fan_ctrl = ctrl.ControlSystem([rule1, rule2])
+        fan_speed_ctrl = ctrl.ControlSystemSimulation(fan_ctrl)
+
+        # Set inputs
+        fan_speed_ctrl.input["temperature"] = 30
+        fan_speed_ctrl.input["humidity"] = 70
```
sfc-gh-aalam, line 1211: I would suggest having a computationally minimal `sproc` for testing purpose but not too strong of an opinion. WDYT @sfc-gh-sfan

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Either package names requested by the user (unavailable in anaconda) or dependencies of requested unavailable packages.

src/snowflake/snowpark/session.py
sfc-gh-vnayak: It's a convenience for sprocs, not having to specify snowpark; maybe this can be added at the  sproc registration point instead so as to not affect UDFs.

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Fixed the version text;

"is" is more accurate because the package `package_name` is the native dependency (hard to know which requested package has this dependency because pip doesn't specify the association between dependencies and requested packages).

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-aalam: I see. Could you elaborate the description to cover that?

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Updated this; it is the list of python packages that are either requested by the user or a dependency of a requested package.

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: Sure, added

src/snowflake/snowpark/session.py
sfc-gh-vnayak: Updated

src/snowflake/snowpark/session.py
sfc-gh-vnayak: Removed these changes for now.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,390 @@
...
+            target, "**", f"*{native_extension}"
+        )
+
+        glob_output: List[str] = glob.glob(base_search_string) + glob.glob(
+            recursive_search_string, recursive=True
+        )
+        if glob_output and len(glob_output) > 0:
+            record_entries_to_package_map = invert_downloaded_package_to_entry_map(
+                downloaded_packages_dict
+            )
```
sfc-gh-aalam, line 299: can be moved outside for loop

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Args:
+            packages_dict (Dict[Requirement, List[str]]): Mapping between packages and a list of files or folders
+            corresponding to it.
+
+        Returns:
+            Dict[str, Set[str]]: The inverse mapping from a file or folder to the packages it corresponds to. Note that
+            it is unlikely a file corresponds to multiple packages (but we allow for the possibility). We only need
+            to return a set of strings here rather than Requirement objects because the specific version of a native
+            package is irrelevant.
```
sfc-gh-aalam, line 275: instead of using _corresponds to_, using _depend on it_ is better in this case.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -0,0 +1,390 @@
...
+                record_entry = os.path.split(relative_path)[0]
+                if (
+                    record_entry == ""
+                ):  # Implies the relative_path is a file name at the base directory
+                    record_entry = relative_path
+
+                # Check which packages own this record entry
+                if record_entry in record_entries_to_package_map:
+                    library_set = record_entries_to_package_map[record_entry]
+                    native_libraries.update(library_set)
```
sfc-gh-aalam, line 314: nit: is `library_set` the set of packages that depend on this native file? naming `package_set` makes it more readable

sfc-gh-vnayak: updated

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-vnayak: The package "is" those files and folders; what would be the right description

sfc-gh-vnayak:
Change documentation for `add_requirements` and `add_packages`; https://github.com/snowflakedb/snowpark-python/pull/900/commits/16a665b89af07e091568d7442c9631cb5b69d8fc

@sfc-gh-aalam

src/snowflake/snowpark/session.py
```diff
@@ -834,16 +857,17 @@ def add_requirements(self, file_path: str) -> None:
         with open(file_path) as f:
             for line in f:
                 package = line.rstrip()
-                if package:
+                if package and len(package) > 0:
```
sfc-gh-aalam, line 865: is there a case when `if package` is `True` but `len(package) > 0` is `False` when `package` is a `str`

src/snowflake/snowpark/session.py
```diff
@@ -870,81 +896,73 @@ def _resolve_packages(
...
                     (col("language") == "python")
-                    & (col("package_name").in_([v[0] for v in package_dict.values()]))
+                    & (col("package_name").in_(package_search_list))
                 )
                 .group_by("package_name")
                 .agg(array_agg("version"))
                 ._internal_collect_with_tag()
             }
-            if validate_package and package_dict
+            if len(package_search_list) > 0
```
sfc-gh-aalam, line 919: we should keep `validate_package` here

sfc-gh-vnayak: It was removed because we need the `valid_packages` list to have access to the versions for snowpark package (because it was being added automatically). The `validate_package` check prevents validation later in the code, so is redundant here.

I have reverted this to match earlier code as we are no longer automatically adding the snowpark package

src/snowflake/snowpark/session.py
sfc-gh-vnayak: True, this is redundant; removed

src/snowflake/snowpark/session.py
```diff
@@ -974,12 +1007,143 @@ def get_req_identifiers_list(
...
+                    if result_dict[name] != str(package):
+                        raise ValueError(
+                            f"Cannot add dependency package '{name}'{'( version '+version+')' if version else ''} "
+                            f"because {result_dict[name]} is already added."
+                        )
+                else:
+                    result_dict[name] = str(package)
+
+        # Add the Snowpark package to packages dict, based on client's environment (or latest), if not already added
+        # add_snowpark_package(result_dict, valid_packages)
```
sfc-gh-aalam, line 1023: should this be uncommented?

sfc-gh-vnayak: Not adding snowpark python package automatically in this PR (moved to next PR for simplicity).

This is because adding snowpark python package only makes sense for sprocs and not udfs (this distinction was not made in this PR)

src/snowflake/snowpark/session.py
```diff
@@ -870,81 +901,73 @@ def _resolve_packages(
...
-                        f"Cannot add package {package_name}=={package_version_req} because {unavailable_pkg_err_msg}"
+                    version_text = (
+                        f"(version {package_version_req})"
+                        if package_version_req is not None
+                        else ""
                     )
+                    if is_in_stored_procedure():  # pragma: no cover
+                        raise RuntimeError(
+                            f"Cannot add package {package_name}{version_text} because it is not present on Anaconda and "
+                            f"cannot be installed via Pip as you are executing this code inside a stored procedure."
```
sfc-gh-aalam, line 940:
```suggestion
                            f"Cannot add package {package_name}{version_text} because it is not available in Snowflake and "
                            f"cannot be installed via Pip as you are executing this code inside a stored procedure. "
```

Can you re-add the part which tells users to check `information_schema.packages` and suggests different ways to upload pure-python packages.

src/snowflake/snowpark/session.py
sfc-gh-aalam: okay. Can you remove the comment here then? Re-add it in the next PR

sfc-gh-vnayak:
Changed the condition for anaconda third party terms acceptance https://github.com/snowflakedb/snowpark-python/pull/900/commits/b907e664b62e8a39e46a6998cb0df2fb37ba8394 

It now matches the condition from before, wherein the terms are only required if a package is not present in Anaconda at all (but not if a particular version of a package is not present).

For a particular missing version, we earlier just raised ValueError.

src/snowflake/snowpark/session.py
```diff
@@ -790,13 +814,16 @@ def clear_packages(self) -> None:
...
         self._packages.clear()
 
-    def add_requirements(self, file_path: str) -> None:
+    def add_requirements(self, file_path: str, force_push: bool = True) -> None:
         """
         Adds a `requirement file <https://pip.pypa.io/en/stable/user_guide/#requirements-files>`_
-        that contains a list of packages as dependencies of a user-defined function (UDF).
+        that contains a list of packages as dependencies of a user-defined function (UDF). Pure Python packages that
+        are not available on Anaconda will be pip installed locally and made available as an import (via zip file
+        on a remote stage).
```
sfc-gh-aalam, line 822: instead of Anaconda, I would rather say Snowflake because even though Anaconda might support it externally, Snowflake might decide to block certain packages.

src/snowflake/snowpark/session.py
```diff
@@ -889,62 +919,57 @@ def _resolve_packages(
...
-                        f"Cannot add package {package_name}=={package_version_req} because {unavailable_pkg_err_msg}"
+                    version_text = (
+                        f"(version {package_version_req})"
+                        if package_version_req is not None
+                        else ""
                     )
+                    if is_in_stored_procedure():  # pragma: no cover
+                        raise RuntimeError(
+                            f"Cannot add package {package_name}{version_text} because it is not available on Snowflake "
+                            f"and it cannot be installed via pip as you are executing this code inside a stored "
```
sfc-gh-aalam, line 940:
```suggestion
                            f"Cannot add package {package_name}{version_text} because it is not available in Snowflake "
                            f"and it cannot be installed via pip as you are executing this code inside a stored "
```
tiny nit

src/snowflake/snowpark/session.py
```diff
@@ -706,6 +722,9 @@ def add_packages(
         :class:`~snowflake.snowpark.udf.UDFRegistration`. See details of
         `third-party Python packages in Snowflake <https://docs.snowflake.com/en/developer-guide/udf/python/udf-python-packages.html>`_.
 
+        Pure Python packages that are not available in Snowflake will be pip installed locally and made available as an
+        import (via zip file on a remote stage).
+
```
sfc-gh-aalam, line 727: nice!

tests/integ/test_udf.py
```diff
@@ -88,6 +95,15 @@ def setup(session, resources_path):
     )
 
 
+@pytest.fixture(autouse=True)
+def clean_up(session):
+    session.clear_packages()
+    session.clear_imports()
+    yield
+    session.clear_packages()
```
sfc-gh-aalam, line 103: Is there a reason this is needed twice?

tests/integ/test_udf.py
```diff
@@ -1497,9 +1529,18 @@ def test_add_packages_negative(session, caplog):
...
-        session.add_packages("dateutil")
-    assert "Cannot add package dateutil" in str(ex_info)
+    with patch.object(session, "_is_anaconda_terms_acknowledged", lambda: True):
+        with pytest.raises(RuntimeError) as ex_info:
+            session.add_packages("dateutil")
+
+        # dateutil is not a valid name, the library name is python-dateutil
+        assert "Pip failed with return code 1" in str(ex_info)
+
+    with patch.object(session, "_is_anaconda_terms_acknowledged", lambda: False):
```
sfc-gh-aalam, line 1539: are the tests failing without the patch?

tests/integ/test_udf.py
```diff
@@ -1548,7 +1582,166 @@ def get_numpy_pandas_version() -> str:
...
+        temperature["cold"] = fuzz.trimf(temperature.universe, [0, 0, 50])
+        temperature["warm"] = fuzz.trimf(temperature.universe, [0, 50, 100])
+        humidity["dry"] = fuzz.trimf(humidity.universe, [0, 0, 50])
+        humidity["moist"] = fuzz.trimf(humidity.universe, [0, 50, 100])
+        fan_speed["low"] = fuzz.trimf(fan_speed.universe, [0, 0, 50])
+        fan_speed["high"] = fuzz.trimf(fan_speed.universe, [0, 50, 100])
+
+        # Define fuzzy rules
+        rule1 = ctrl.Rule(temperature["cold"] & humidity["dry"], fan_speed["low"])
+        rule2 = ctrl.Rule(temperature["warm"] & humidity["moist"], fan_speed["high"])
```
sfc-gh-aalam, line 1681: same comment about complexity of the udf. It is better if we have something simple.

tests/integ/test_udf.py
sfc-gh-vnayak: Without the patch, anaconda terms are not acknowledged and RuntimeError will be raised, as expected in this test case. The patch helps ensure the test is not flaky (i.e. if the acknowledgement changes in the future, this test will remain robust.

tests/integ/test_udf.py
sfc-gh-vnayak: Not sure if this changes things but the code here runs fast.

tests/integ/test_udf.py
sfc-gh-vnayak: This is redundant; removed

tests/integ/test_stored_procedure.py
```diff
@@ -63,6 +63,14 @@ def setup(session, resources_path):
     )
 
 
+@pytest.fixture(autouse=True)
+def clean_up(session):
+    session.clear_packages()
+    session.clear_imports()
+    session.add_packages("snowflake-snowpark-python")
+    yield
```
sfc-gh-aalam, line 71: Do you think it is a better idea to break this down into setup and cleanup? adding inside cleanup looks odd

tests/integ/test_udf.py
sfc-gh-aalam: The reason I am in favor of a simple code here is so that if things break and we need to debug, it is easier to debug a simpler code rather than a complex code.

sfc-gh-vnayak: Thx, updated to a simpler test.

tests/integ/test_udf.py
```diff
@@ -1519,20 +1558,13 @@ def test_add_packages_negative(session, caplog):
...
-    session.clear_packages()
-
 
 @pytest.mark.skipif(
     (not is_pandas_and_numpy_available) or IS_IN_STORED_PROC,
     reason="numpy and pandas are required",
 )
 def test_add_requirements(session, resources_path):
     test_files = TestFiles(resources_path)
-    session.clear_packages()
```
sfc-gh-aalam, line 1568: we no longer need this because `clean_up` does it automatically, right?

sfc-gh-aalam (approved):
LGTM

tests/integ/test_udf.py
sfc-gh-vnayak: Yeah

sfc-gh-stan (approved):

PR #933, opened 2023-07-06
SNOW-856555: Fix get_type_hint exception on numpy ufunc

sfc-gh-jdu:

   Fixes SNOW-856555

   A TypeError will be raised if we try to call get_type_hints from a numpy ufunc, e.g., `get_type_hints(np.exp)`. In this case, if it fails, we just assume there are no type hints. This bug blocks supporting series/df.apply on numpy functions in snowpandas.

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -93,9 +93,14 @@ def get_types_from_type_hints(
         # For Python 3.10+, the result values of get_type_hints()
         # will become strings, which we have to change the implementation
         # here at that time. https://www.python.org/dev/peps/pep-0563/
-        python_types_dict = get_type_hints(
-            getattr(func, TABLE_FUNCTION_PROCESS_METHOD, func)
-        )
+        try:
+            python_types_dict = get_type_hints(
+                getattr(func, TABLE_FUNCTION_PROCESS_METHOD, func)
+            )
+        except TypeError:
+            # A TypeError will be raised if we try to get type hints from a numpy ufunc
+            # e.g., get_type_hints(np.exp)
+            python_types_dict = {}
     else:
         if object_type == TempObjectType.TABLE_FUNCTION:
             python_types_dict = (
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -93,9 +93,14 @@ def get_types_from_type_hints(
...
         # here at that time. https://www.python.org/dev/peps/pep-0563/
-        python_types_dict = get_type_hints(
-            getattr(func, TABLE_FUNCTION_PROCESS_METHOD, func)
-        )
+        try:
+            python_types_dict = get_type_hints(
+                getattr(func, TABLE_FUNCTION_PROCESS_METHOD, func)
+            )
+        except TypeError:
+            # A TypeError will be raised if we try to get type hints from a numpy ufunc
```
sfc-gh-yzou, line 101: do we want to catch the TypeError in general? If we do not, or not clear on this, can we do a check to see if the function is numpy ufunc before we call get_type_hints instead of catch the TypeError exception?

sfc-gh-jdu: Yep, the error is raised from https://github.com/python/cpython/blob/5548097925b9924ebf761376d632c5198d01ebd5/Lib/typing.py#L2211-L2217. It's also possible that functions from other libraries will fail here (not in _allowed_types), then we should also catch the error and assume there is no type hint. In general, only normal Python functions allow users to add type hints.

sfc-gh-yzou: Maybe make the comment more clear that "if we fail to get_type_hints, return empty type dict. This will fail for functions like numpy.min"

sfc-gh-yzou (approved):

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-sfan: Agreed. I think failing in general makes sense, but we should rewrite the comment to reflect this :)

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-jdu: done

PR #941, opened 2023-07-12
SNOW-747608 Make tests compatible with UDAF behavior change

sfc-gh-sfan:
Description

Enabling UDAF adds an is_aggregate field when describing a function. We need to make sure our tests work both ways before enabling the feature in the test accounts.

Testing

Tested on an account that has UDAF enabled

   Fixes SNOW-747608

   Enabling UDAF adds an is_aggregate field when describing a function. We need to make sure our tests work both ways before enabling the feature in the test accounts.

### DIFF
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -1014,7 +1014,8 @@ def test_describe_sp(session, source_code_display):
 
     return1_sp = session.sproc.register(return1)
     describe_res = session.sproc.describe(return1_sp).collect()
-    assert [row[0] for row in describe_res] == [
+    actual_fields = [row[0] for row in describe_res]
+    expected_fields = [
         "signature",
         "returns",
         "language",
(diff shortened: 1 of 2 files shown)


tests/integ/test_stored_procedure.py
```diff
@@ -1027,7 +1028,15 @@ def return1(session_: Session) -> str:
         "runtime_version",
         "packages",
         "installed_packages",
+        # This seems like an unintended change from the server, we should remove it once it is removed from server
+        "is_aggregate",
```
sfc-gh-aalam, line 1032: I had a chat some time back with @sfc-gh-fdrozdowski and I recall it being an intentional change. Can you confirm @sfc-gh-fdrozdowski

sfc-gh-aalam (approved):

tests/integ/test_stored_procedure.py
sfc-gh-fdrozdowski: IIUC, this field should not be shown for stored proc as it doesn't make sense, see https://snowflake.slack.com/archives/C04A6RQ3571/p1689185938064419?thread_ts=1689182211.748549&cid=C04A6RQ3571. @sfc-gh-ddalmeida Do you have a jira to track remove `is_aggregate` for stored procs?

sfc-gh-sfan: It is probably intentional for UDF, but should not be intentional for stored proc. I also chatted with @sfc-gh-ddalmeida on this. It should be easy to hide this property for stored proc.

PR #949, opened 2023-07-17
SNOW-841310: EXCLUDE clause simplifier

sfc-gh-vnayak:

   Fixes #NNNN


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -501,7 +512,16 @@ def sql_query(self) -> str:
...
+        if exclude_clause != analyzer_utils.EMPTY_STRING:
+            self._sql_query = (
+                f"{analyzer_utils.SELECT}{analyzer_utils.STAR}{exclude_clause}{analyzer_utils.FROM}"
+                f"{from_clause}{where_clause}{order_by_clause}{limit_clause}{offset_clause}"
+            )
+        else:
+            self._sql_query = (
+                f"{analyzer_utils.SELECT}{self.projection_in_str}{exclude_clause}{analyzer_utils.FROM}"
+                f"{from_clause}{where_clause}{order_by_clause}{limit_clause}{offset_clause}"
+            )
```
sfc-gh-sfan, line 524: You could use `selected` or `projection` as a variable name, such that it represents either `{analyzer_utils.STAR}{exclude_clause}` or `{self.projection_in_str}` depending on the value of `exclude_clause. This way you don't need to replicate most of the other clauses of the query.

Also, should we assert that only one of `projection_in_str` and `exclude_clause` exists?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -649,6 +671,92 @@ def filter(self, col: Expression) -> "SelectStatement":
 
         return new
 
+    def drop(
+        self, cols: List[Expression], kept_cols: List[Expression]
+    ) -> "SelectStatement":
+        disable_next_level_flatten = False
+        new_column_states = derive_column_states_from_subquery(kept_cols, self)
+        if new_column_states is None:
+            can_be_flattened = False
```
sfc-gh-sfan, line 680: We have a lot of conditions where `can_be_flattened` is `False`. I wonder if we should add a comment saying under what condition we will continue flattening the query.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -649,6 +671,92 @@ def filter(self, col: Expression) -> "SelectStatement":
...
+            if self.exclude is None:
+                new.exclude = cols
+            else:
+                new.exclude = self.exclude + cols
+
+        else:
+            new = SelectStatement(
+                from_=self.to_subqueryable(),
+                projection=kept_cols,
+                exclude=cols,
```
sfc-gh-sfan, line 751: Will we eventually do some magic between these two such that we get the correct projection columns?

sfc-gh-vnayak:
Note that EXCLUDE simplifier needs to be merged after the RENAME simplifier.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -946,6 +947,13 @@ def do_resolve_with_resolved_children(
                 logical_plan,
             )
 
+        if isinstance(logical_plan, Drop):
```
sfc-gh-yixie, line 950: Is SelectStatement itself good enough? I'm thinking of removing this new logical plan (and Rename).

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-vnayak: The exclude parameter is primarily used to generate the correct SQL statement, does not play a role in other computation (projection does). As long as the projection parameter is accurate (contains columns that should remain in the dataframe), we will get the correct projections.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-vnayak: I see that nearly every API in dataframe.py checks if `self._select_statement` is None, and if it is, the API code defaults to using a logical plan (saw this in order by, sort, filter, etc). The `Drop` plan is used in a similar way.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-vnayak: The assertion that only one exists is implicit in the flattening checks; for example https://github.com/snowflakedb/snowpark-python/pull/949/files#diff-ba9463440d935e1d5667a2909f7d2b8061f56de7bdabd8207a29903e9eae8f59R681-R682.

Having an extra assertion might help though, I will add that, thank you!

tests/integ/test_simplifier_suite.py
```diff
@@ -1002,12 +1002,25 @@ def test_rename_to_dropped_column_name(session):
     session.sql_simplifier_enabled = True
     df1 = session.create_dataframe([[1, 2, 3]], schema=["a", "b", "c"])
     df2 = df1.drop("a").drop("b")
-    df3 = df2.withColumn("a", df2["c"])
+    df3 = df2.withColumn("a", col("c"))
```
sfc-gh-vnayak, line 1005: This test has been temporarily altered. @sfc-gh-yixie

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-sfan: Let's say we eventually have `projection` that is `[a, b, c]`, and `exclude` is `[a]`. Do we squash the result such that we know the result is `[b, c]` (because we try to find the intersection), or is this not a legit input?

sfc-gh-vnayak: Projection would be `[b, c]` if exclude is `[a]`, because of the way projection and exclude are computed. In the `drop()` function in `dataframe.py`, `kept_col_names` and `remove_col_names` are mutually exclusive (and these eventually form the variables `projection` and `exclude`)

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-vnayak: Added descriptions for situations where `exclude` variable is used; the other conditions mirror the conditions from `select()` in `select_statement.py` and have the same query flattening rationale.

tests/integ/test_simplifier_suite.py
sfc-gh-vnayak:
```
df1 = session.create_dataframe([[1, 2, 3]], schema=["a", "b", "c"])
df2 = df1.with_column_renamed("a", "ae").with_column_renamed("b", "be")
df3 = df2.withColumn("a", df2["ae"])
df4 = df3.withColumn("b", df2["be"])
print(df4.collect()[0])
```

The output is `Row(A=1, BE=2, C=3, A=1, B=2)`, not `Row(AE=1, BE=2, C=3, A=1, B=2)`.

This bug is related to the test failure here. They are both caused by incorrect new projection formation when SQL Simplifier is enabled but `can_be_flattened` is set to `False`.

PR #953, opened 2023-07-18
SNOW-871287: Replicate packages in your Python environment on Snowflake (V2)

sfc-gh-vnayak:

   Fixes #NNNN


sfc-gh-vnayak:
Blocked on https://snowflakecomputing.atlassian.net/browse/SNOW-838111

src/snowflake/snowpark/session.py
```diff
@@ -904,6 +904,78 @@ def add_requirements(
             persist_path=persist_path,
         )
 
+    def replicate_local_environment(
```
sfc-gh-yixie, line 907: Better add `@experimental`

sfc-gh-vnayak:
Intended to be merged after 1.6.0 release

src/snowflake/snowpark/session.py
```diff
@@ -917,6 +918,80 @@ def add_requirements(self, file_path: str) -> None:
                 self.add_import(import_path)
         self.add_packages(packages)
 
+    @experimental(version="1.7.0")  # TODO - Update the version here
```
sfc-gh-sfan, line 921: This looks about right :p We probably can remove TODO

src/snowflake/snowpark/session.py
```diff
@@ -917,6 +918,80 @@ def add_requirements(self, file_path: str) -> None:
...
+            1. This method will add packages for all UDFs created later in the current
+            session. If you only want to add packages for a specific UDF, you can use
+            ``packages`` argument in :func:`functions.udf` or
+            :meth:`session.udf.register() <snowflake.snowpark.udf.UDFRegistration.register>`.
+
+            2. We recommend you to `setup the local environment with Anaconda <https://docs.snowflake.com/en/developer-guide/udf/python/udf-python-packages.html#local-development-and-testing>`_,
+            to ensure the consistent experience of a UDF between your local environment
+            and the Snowflake server.
+        """
+        default_packages = ["wheel", "pip", "setuptools"]
```
sfc-gh-sfan, line 978: We probably should make this a static constant

src/snowflake/snowpark/session.py
```diff
@@ -917,6 +918,80 @@ def add_requirements(self, file_path: str) -> None:
...
+        packages = []
+        for package in pkg_resources.working_set:
+            if package.key in ignore_packages:
+                _logger.info(f"{package.key} found in environment, ignoring...")
+                continue
+            if package.key in default_packages:
+                _logger.info(f"{package.key} is available by default, ignoring...")
+                continue
+            packages.append(
+                f"{package.key}{'==' + package.version if package.has_version() else ''}"
```
sfc-gh-sfan, line 990: What if user's local version is not available on Snowflake? I think this will directly cause a failure 🤔

src/snowflake/snowpark/session.py
```diff
@@ -917,6 +918,80 @@ def add_requirements(self, file_path: str) -> None:
...
+            --------------
+            <BLANKLINE>
+            >>> session.clear_packages()
+            >>> session.clear_imports()
+
+        Args:
+            reinstall_packages: Ignores environment present on persist_path and overwrites it with a fresh installation.
+            ignore_packages: Set of packages that will be ignored.
+            persist_path: A remote stage directory path where packages not present in Snowflake will be persisted. Mentioning
+             this path will speed up automated package loading.
```
sfc-gh-aalam, line 965: might want to update this

src/snowflake/snowpark/session.py
sfc-gh-aalam: based on auto-upload settings, it might get installed via pip, zipped and uploaded to a stage. Is that right Vivek?

sfc-gh-vnayak: Yes, if the user's package versions are not on Snowflake and package auto-upload is not enabled, `_resolve_packages` will throw an error.

tests/integ/test_packaging.py
```diff
@@ -966,3 +966,33 @@ def test_get_available_versions_for_packages(session):
     assert returned.keys() == set(packages)
     for key in returned.keys():
         assert len(returned[key]) > 0
+
+
+@pytest.mark.skipif(
+    IS_IN_STORED_PROC or IS_WINDOWS,
+    reason="Subprocess calls are not allowed within stored procedures. Custom package upload does not work well on Windows.",
+)
+def test_replicate_local_environment(session):
```
sfc-gh-aalam, line 975: I don't understand what this test is testing. the sample udf does not need any packages to work so what's the point of replicating local environment here?

sfc-gh-vnayak: This is true, It is difficult to know how to test replicate_local_environment in a non-flaky way because it is tied to the environment (I would have to pip install stuff in the environment to make a valid test).

This test only proves that replicate_local_environment does not prevent UDF from working, not super useful.

src/snowflake/snowpark/session.py
sfc-gh-sfan: Yeah that makes sense. I was just wondering if we should provide a relaxed mode, such that if:

- the user's environment contains packages that exists on Snowflake but not the exact version
- This package cannot be installed and uploaded to a stage

Then we just use the package without specifying the version. 

I'm not sure if this is a legit use case so just thinking out loud.

tests/integ/test_packaging.py
sfc-gh-sfan: Hmm I see. This is interesting. Is there a package that we absolutely will install in the enviornment that we could do an import test? 🤔

sfc-gh-vnayak: `cloudpickle` maybe?

src/snowflake/snowpark/session.py
sfc-gh-vnayak: Yeah maybe we can have a `relax=True` argument, sounds useful.

tests/integ/test_packaging.py
```diff
@@ -966,3 +966,33 @@ def test_get_available_versions_for_packages(session):
...
+                ignore_packages={
+                    "snowflake-snowpark-python",
+                    "snowflake-connector-python",
+                    "urllib3",
+                    "tzdata",
+                    "numpy",
+                },
+            )
+
+    assert any([package.startswith("cloudpickle") for package in session._packages])
```
sfc-gh-aalam, line 998: can we have more tests for example
- `DEFAULT_PACKAGES` not in `session._packages`
- ignore packages not in `session._packages`

sfc-gh-aalam (approved):

PR #947, opened 2023-07-17
SNOW-869906: Tag package auto-upload feature as experimental

sfc-gh-vnayak:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

PR #963, opened 2023-07-21
SNOW-871306: Fix test failures on Windows for tests in test_packaging.py

sfc-gh-vnayak:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/packaging_utils.py
+++ b/src/snowflake/snowpark/_internal/packaging_utils.py
@@ -302,6 +302,7 @@ def pip_install_packages_to_target_folder(packages: List[str], target: str) -> N
         ModuleNotFoundError: If pip is not present.
         RuntimeError: If pip fails to install the packages.
     """
+    _logger.debug(f"Using pip to install packages ({packages}), via subprocess...")
     try:
         pip_executable: str = os.getenv(PIP_ENVIRONMENT_VARIABLE)
         pip_command: List[str] = (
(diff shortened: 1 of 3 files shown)


sfc-gh-vnayak:
The standard output (stdout) and standard error (stderr) are captured with subprocess.PIPE. However, the Popen call was earlier followed by `process.wait()`, which waits for the process to complete, but does not read the output. If the child process generates enough output to fill the OS pipe buffer, the child process will block waiting for the OS to make room in the pipe buffer. This will never occur because the parent process is waiting for the child process to exit before it reads the output. 

This can be resolved by using process.communicate() instead of process.wait(). The communicate() function waits for the process to terminate and also reads all output, thereby avoiding the deadlock situation.

See details at https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

tests/integ/test_packaging.py
```diff
@@ -305,7 +305,7 @@ def test_add_unsupported_requirements_twice_should_not_fail_for_same_requirement
 
 
 @pytest.mark.skipif(
-    IS_IN_STORED_PROC or IS_WINDOWS,
+    IS_IN_STORED_PROC,
```
sfc-gh-vnayak, line 308: The IS_WINDOWS safeguard will be added back for all tests (see https://github.com/snowflakedb/snowpark-python/pull/947).

They are temporarily undone here to test the changes on Windows.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -311,17 +315,20 @@ def pip_install_packages_to_target_folder(packages: List[str], target: str) -> N
         process = subprocess.Popen(
             pip_command + ["install", "-t", target, *packages],
             stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
             universal_newlines=True,
         )
-        process.wait()
+        stdout, stderr = process.communicate(timeout=timeout)
```
sfc-gh-vnayak, line 321: The standard output (stdout) and standard error (stderr) are captured with subprocess.PIPE. However, the Popen call was earlier followed by process.wait(), which waits for the process to complete, but does not read the output. If the child process generates enough output to fill the OS pipe buffer, the child process will block waiting for the OS to make room in the pipe buffer. This will never occur because the parent process is waiting for the child process to exit before it reads the output.

This can be resolved by using process.communicate() instead of process.wait(). The communicate() function waits for the process to terminate and also reads all output, thereby avoiding the deadlock situation.

See details at https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

tests/integ/test_packaging.py
sfc-gh-aalam: I also remember in experimental PR, you mentioned that this might not work well with windows. Do you want to remove that comment now?

sfc-gh-aalam (approved):

tests/integ/test_packaging.py
sfc-gh-vnayak: No, these failures are flaky. The pre-commit succeeding on this PR does not imply that it won't happen again.

I feel the answer here is to not support the package auto-upload feature on Windows and only look into changing that (and removing the Windows safeguards) if there is a request from users to do so. What do you think?

Will merge the experimental PR after this one so that the tests will not run on Windows.

PR #978, opened 2023-07-27
SNOW-876321: Fix schema_query for union with sql simplifier enabled

sfc-gh-aalam:

   Fixes #SNOW-876321

   A regression was identified in the attached SNOW ticket. The root cause is that the schema_query generated when sql simplifier is enabled only took the first operand of union into consideration. This change updates the schema query generation to consider all operands.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -32,7 +32,10 @@ if TYPE_CHECKING:
 import sys
 
 from snowflake.snowpark._internal.analyzer import analyzer_utils
-from snowflake.snowpark._internal.analyzer.analyzer_utils import result_scan_statement
+from snowflake.snowpark._internal.analyzer.analyzer_utils import (
+    result_scan_statement,
+    schema_value_statement,
+)
 from snowflake.snowpark._internal.analyzer.binary_expression import And
 from snowflake.snowpark._internal.analyzer.expression import (
     COLUMN_DEPENDENCY_ALL,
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -812,7 +815,12 @@ def sql_query(self) -> str:
     def schema_query(self) -> str:
         """The first operand decide the column attributes of a query with set operations.
         Refer to https://docs.snowflake.com/en/sql-reference/operators-query.html#general-usage-notes"""
-        return self.set_operands[0].selectable.schema_query
+        attributes = self.set_operands[0].selectable.snowflake_plan.attributes
+        sql = f"({schema_value_statement(attributes)})"
+        for i in range(1, len(self.set_operands)):
+            attributes = self.set_operands[i].selectable.snowflake_plan.attributes
+            sql = f"{sql}{self.set_operands[i].operator}({schema_value_statement(attributes)})"
```
sfc-gh-sfan, line 822: I'm just curious: what is the operator of the first operand? Is it None?

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -812,7 +815,12 @@ def sql_query(self) -> str:
     def schema_query(self) -> str:
         """The first operand decide the column attributes of a query with set operations.
         Refer to https://docs.snowflake.com/en/sql-reference/operators-query.html#general-usage-notes"""
-        return self.set_operands[0].selectable.schema_query
```
sfc-gh-stan, line 815: Is this a regression in the sense that the customer did not enable sql simplifier before?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: It is `None` indeed. If you expand a few lines above, the new `schema_query` generation follows the same pattern as `sql_query` generation.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: Two things happened together to surface this bug we had in `schema_query` generation for set operations with sql simplifier enabled.
1. The old `schema_query` generation for set operations assumed that the overall schema query will be the same as schema query for the first operand.
2. New StringType changes were added where max width of the column will respected by the client.

The users had sql simplifier enabled and they were doing a `DataFrame.union` operation on two string type columns of different max widths. The first operator had a smaller width than the second operator. After the `union`, they ran a `df.cache_result` operation,  which uses the `schema_query` to create a table and insert the result of `union` operation into the table. Since the max width picked here did not allow entry for the second column, the users had a regression.

PR #979, opened 2023-07-27
SNOW-876999: Add support for timezone-aware type annotation on UDF 

sfc-gh-jdu:

   Fixes SNOW-876999

   Previously we add support for timestamp variations in data types, but we also need such support for UDFs with type annotations.

### DIFF
--- a/src/snowflake/snowpark/_internal/type_utils.py
+++ b/src/snowflake/snowpark/_internal/type_utils.py
@@ -516,7 +516,9 @@ def python_type_to_snow_type(tp: Union[str, Type]) -> Tuple[DataType, bool]:
         elif tp_args[0] == TZ:
             timezone = TimestampTimeZone.TZ
         else:
-            timezone = TimestampTimeZone.DEFAULT
+            raise TypeError(
+                f"Only Timestamp, Timestamp[NTZ], Timestamp[LTZ] and Timestamp[TZ] are allowed, but got {tp}"
+            )
         return TimestampType(timezone), False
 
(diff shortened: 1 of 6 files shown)


sfc-gh-azhan (approved):

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -502,6 +506,19 @@ def python_type_to_snow_type(tp: Union[str, Type]) -> Tuple[DataType, bool]:
...
+    if tp == Timestamp or tp_origin == Timestamp:
+        if not tp_args:
+            timezone = TimestampTimeZone.DEFAULT
+        elif tp_args[0] == NTZ:
+            timezone = TimestampTimeZone.NTZ
+        elif tp_args[0] == LTZ:
+            timezone = TimestampTimeZone.LTZ
+        elif tp_args[0] == TZ:
+            timezone = TimestampTimeZone.TZ
+        else:
```
sfc-gh-sfan, line 518: Just curious: what can this `else` be? Or should we actually raise?

sfc-gh-sfan (commented):
Please also address the comment about test coverage for default case :)

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: Actually it can be `Timestamp[int]`, which is invalid and we should raise an exception here, good catch! I added tests to cover these code path.

sfc-gh-sfan (approved):

PR #983, opened 2023-08-01
Fix tests with no pandas installed

sfc-gh-stan:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -165,24 +165,20 @@ def extract_return_type_from_udtf_type_hints(
     elif return_type_hint is None:
         return None
     else:
-        # Vectorized UDTF
-        if not installed_pandas:
-            raise RuntimeError(
-                "No pandas detected in local environment, please install pandas to use vectorized UDTF"
-            )
-        elif typing.get_origin(return_type_hint) == PandasDataFrame:
-            return PandasDataFrameType(
-                col_types=[
-                    python_type_to_snow_type(x)[0]
-                    for x in typing.get_args(return_type_hint)
-                ],
-                col_names=output_schema,
-            )
-        elif return_type_hint is pandas.DataFrame:
-            return PandasDataFrameType(
-                []
-            )  # placeholder, indicating the return type is pandas DataFrame
-        elif return_type_hint is NoneType:
+        if installed_pandas:  # Vectorized UDTF
+            if typing.get_origin(return_type_hint) == PandasDataFrame:
+                return PandasDataFrameType(
+                    col_types=[
+                        python_type_to_snow_type(x)[0]
+                        for x in typing.get_args(return_type_hint)
+                    ],
+                    col_names=output_schema,
+                )
+            elif return_type_hint is pandas.DataFrame:
+                return PandasDataFrameType(
+                    []
+                )  # placeholder, indicating the return type is pandas DataFrame
+        if return_type_hint is NoneType:
             return None
         else:
             raise ValueError(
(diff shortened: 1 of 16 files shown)


src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -165,24 +165,20 @@ def extract_return_type_from_udtf_type_hints(
...
-                ],
-                col_names=output_schema,
-            )
-        elif return_type_hint is pandas.DataFrame:
-            return PandasDataFrameType(
-                []
-            )  # placeholder, indicating the return type is pandas DataFrame
-        elif return_type_hint is NoneType:
+        if installed_pandas:  # Vectorized UDTF
+            if typing.get_origin(return_type_hint) == PandasDataFrame:
```
sfc-gh-sfan, line 169: This can be a follow up and I'm not sure if it is feasible, but I wonder if we should introduce something like `snowflake.connector.options` where we have the import handled in one place 🤔

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -165,24 +165,20 @@ def extract_return_type_from_udtf_type_hints(
...
-                    for x in typing.get_args(return_type_hint)
-                ],
-                col_names=output_schema,
-            )
-        elif return_type_hint is pandas.DataFrame:
-            return PandasDataFrameType(
-                []
-            )  # placeholder, indicating the return type is pandas DataFrame
-        elif return_type_hint is NoneType:
+        if installed_pandas:  # Vectorized UDTF
```
sfc-gh-sfan, line 168: Can you help me understand why this fixes the issue? Do we happen to go to `elif` when `if not installed_pandas` is `True`?

tests/integ/test_udf.py
```diff
@@ -1557,54 +1566,76 @@ def add_one_df_pandas_annotation(df: pandas.DataFrame) -> pandas.Series:
...
-@pytest.mark.skipif(not is_pandas_and_numpy_available, reason="pandas is required")
+@pytest.mark.skipif(not is_pandas_available, reason="pandas is required")
 @pytest.mark.parametrize(
     "_type, data, expected_types, expected_dtypes",
     [
         (
             IntegerType,
             [[4096]],
-            (numpy.int16, int, numpy.short),
+            ("<class 'numpy.int16'>", "<class 'int'>"),
```
sfc-gh-sfan, line 1576: Are these changes expected?

sfc-gh-stan: Yes these are expected since `parameterize` is evaluated before `skip` is evaluated. So the test will try to evaluate `numpy.int16` which fails when `pandas` is not installed. So I converted these into string comparison as a workaround.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-stan: Yea I agree, there needs to be a cleaner and more robust way of doing this.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-stan: This block was not what broke the tests. I refactored this part a bit so that when no pandas is installed and the return type is invalid (not a collection type and not a pandas DataFrame), user would hit the invalid type invalid error instead of the "pandas not installed" error. The one liner changes below are what broke the tests before.

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aalam: do you import `PandasDataFrameType` or `PandasSeriesType` from somewhere? If pandas is not installed, how do the import work?

sfc-gh-stan: 👍  https://snowflakecomputing.atlassian.net/browse/SNOW-881404

PR #989, opened 2023-08-04
SNOW-878135 is_permanent behavior fix

sfc-gh-aalam:

   Fixes #SNOW-878135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 ### Behavior Changes
 
-- When creating stored procedures, UDFs, UDTFs, UDAFs with parameter `is_permanent=False` will now create temporary objects even when `stage_name` is provided. Earlier `is_permanent=False` with non-None `stage_name` try to create permanent objects at the given stage location.
+- When creating stored procedures, UDFs, UDTFs, UDAFs with parameter `is_permanent=False` will now create temporary objects even when `stage_name` is provided. The default value of `is_permanent` is `False` which is why if this value is not explicitly set to `True` for permanent objects, users will notice a change in behavior.
 
 ## 1.6.1 (2023-08-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   - `array_flatten`
 - Added support for replicating your local Python environment on Snowflake via `Session.replicate_local_environment`.
 
+### Behavior Changes
+
+- When creating stored procedures, UDFs, UDTFs, UDAFs with parameter `is_permanent=False` will now create temporary objects even when `stage_name` is provided. Earlier `is_permanent=False` with non-None `stage_name` try to create permanent objects at the given stage location.
+
 ## 1.6.1 (2023-08-02)
 
 ### New Features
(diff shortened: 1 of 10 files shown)


sfc-gh-stan:
Does it make sense to get rid of `is_permanent`?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -341,12 +341,19 @@ def check_register_args(
...
+            logger.warn(
+                "is_permanent is False therefore stage_location will be ignored"
+            )
 
     if parallel < 1 or parallel > 99:
         raise ValueError(
             "Supported values of parallel are from 1 to 99, " f"but got {parallel}"
         )
 
+    return stage_location if is_permanent else None
```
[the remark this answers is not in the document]
sfc-gh-aalam: Yeah. I think that is the right way to do things here. I'll reformat the code.

sfc-gh-aalam:
> Does it make sense to get rid of `is_permanent`?

Not to me. If we get rid of it, do you suggest we use `stage_location` as a proxy? I don't think that's intuitive.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
   - `array_flatten`
 - Added support for replicating your local Python environment on Snowflake via `Session.replicate_local_environment`.
 
+### Behavior Changes
+
+- When creating stored procedures, UDFs, UDTFs, UDAFs with parameter `is_permanent=False` will now create temporary objects even when `stage_name` is provided. Earlier `is_permanent=False` with non-None `stage_name` try to create permanent objects at the given stage location.
```
sfc-gh-sfan, line 14: I think we also need to call out that the default behavior is changing because the default value of `is_permanent` is `None`. So the behavior of a UDF that only has `stage_name` specified will see their UDF becomes temporary.

sfc-gh-sfan (commented):
We probably also want to add similar test to `udtf` and `udaf`?

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -790,7 +793,7 @@ def _do_register_sp(
                     object_name=udf_name,
                     all_imports=all_imports,
                     all_packages=all_packages,
-                    is_temporary=stage_location is None,
+                    is_temporary=not is_permanent,
```
sfc-gh-sfan, line 796: I wonder if it makes sense to rename `is_temporary` to `is_permanent`? This way we only need to do negation once instead of in all four classes.

We probably should have made the public param `is_temporary` at the beginning so we don't need this awkward flip lol. But now all is done 🤷

sfc-gh-stan:
> e `stage_location` as a

What about something like `permanent_stage_location` ? I prefer only having one parameter here but I understand that will be a BCR.

src/snowflake/snowpark/stored_procedure.py
sfc-gh-aalam: we could rename `is_temporary -> is_permanent`. public API already has `is_permanent` so we shouldn't change that.

sfc-gh-sfan (approved):

PR #995, opened 2023-08-08
SNOW-670956 Support qualify

sfc-gh-stan:

   Fixes #NNNN

   Support `DataFrame.qualify`, which corresponds to [QUALIFY](https://docs.snowflake.com/en/sql-reference/constructs/qualify) clause in Snowflake SQL.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -394,12 +394,14 @@ class SelectStatement(Selectable):
         order_by: Optional[List[Expression]] = None,
         limit_: Optional[int] = None,
         offset: Optional[int] = None,
+        qualify: Optional[Expression] = None,
         analyzer: "Analyzer",
     ) -> None:
         super().__init__(analyzer)
         self.projection: Optional[List[Expression]] = projection
         self.from_: Optional["Selectable"] = from_
         self.where: Optional[Expression] = where
+        self.qualify_: Optional[Expression] = qualify
         self.order_by: Optional[List[Expression]] = order_by
         self.limit_: Optional[int] = limit_
         self.offset = offset
@@ -422,6 +424,7 @@ class SelectStatement(Selectable):
             projection=self.projection,
             from_=self.from_,
             where=self.where,
+            qualify=self.qualify_,
             order_by=self.order_by,
             limit_=self.limit_,
             offset=self.offset,
@@ -456,6 +459,7 @@ class SelectStatement(Selectable):
             (
                 self.where is not None,
                 self.order_by is not None,
+                self.qualify_ is not None,
             )
         )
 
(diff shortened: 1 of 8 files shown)


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -394,12 +394,14 @@ def __init__(
         order_by: Optional[List[Expression]] = None,
         limit_: Optional[int] = None,
         offset: Optional[int] = None,
+        qualify: Optional[Expression] = None,
```
sfc-gh-sfan, line 397: It's interesting that we have `qualify` here but not `having` 🤔

sfc-gh-sfan (commented):
What is the behavior when a dataframe has no window function and tries to use qualify? Given that we don't have a check in client, I guess it would be the server to raise this error. Is the error easier to comprehend for a snowpark user? Or should we actually add some client side validation based on class type (because we know if certain construct is a window function)?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-stan: True. There is no active demand for these SQL features that don't have a direct counterpart in Spark. It shouldn't be much work to support it tho.

sfc-gh-stan:
> What is the behavior when a dataframe has no window function and tries to use qualify? Given that we don't have a check in client, I guess it would be the server to raise this error. Is the error easier to comprehend for a snowpark user? Or should we actually add some client side validation based on class type (because we know if certain construct is a window function)?

Good question. Server will raise a SQL compilation exception that looks something like `SQL compilation error: error line 1 at position 228
found QUALIFY clause but no window function`. I think this is pretty comprehensible, we could do client side verification but only partially since the use case where `df.select(<window_expr>.alias("criteria")).qualify(col("criteria") > 1)` cannot be verified easily (we never check logical plan of the subquery).

src/snowflake/snowpark/dataframe.py
```diff
@@ -3496,6 +3497,41 @@ def with_column_renamed(self, existing: ColumnOrName, new: str) -> "DataFrame":
         ]
         return self.select(new_columns)
 
+    def qualify(self, expr: ColumnOrSqlExpr) -> "DataFrame":
```
sfc-gh-yixie, line 3500: Just to brainstorm, is it possible to still use api `filter` and generate SQL with `qualify` when window functions are used?

sfc-gh-stan: This is similar to Shixuan's question. To fully support that, we need to detect window expression in the current dataframe's projection, so for something like `df.select(window_expr.alias("winwin").filter(col("winwin") > 12)` we need to traverse the dataframe's logical plan. It's probably easier for simplified paths.

PR #1000, opened 2023-08-09
SNOW-680119: table join with overlapping columns

sfc-gh-aalam:

   Fixes #SNOW-680119

   There is a bug with aliasing output columns of table function where aliasing fails if table output columns overlap with dataframe columns. This change fixes that bug.


src/snowflake/snowpark/dataframe.py
```diff
@@ -1011,16 +1011,27 @@ def select(
...
-                    _, new_cols = _get_cols_after_join_table(
-                        func_expr, self._plan, join_plan
+                    # this join plan is created here to extract output columns after the join. If a better way
+                    # to extract this information is found, please update this function.
+                    temp_join_plan = self._session._analyzer.resolve(
+                        TableFunctionJoin(self._plan, func_expr)
+                    )
+                    _, new_cols, alias_cols = _get_cols_after_join_table(
+                        func_expr, self._plan, temp_join_plan
                     )
```
sfc-gh-aalam, line 1025: We have temp join here to generate a join sql query and run schema query to get the columns generated after the join. We use the result to do apply aliases into the new generated sql query using `right_cols = new_col_names`. The other way we can do this is to run a describe query on table function but that is practically not very different from the earlier approach. 

If anyone in the team has a better approach to do this, let me know. I'm happy to update this section.

tests/integ/test_telemetry.py
```diff
@@ -930,6 +930,8 @@ def add_one(session_, x):
 
 @pytest.mark.udf
 def test_udtf_call_and_invoke(session, resources_path):
+    sql_simplifier_enabled = session.sql_simplifier_enabled
+    session.sql_simplifier_enabled = True
```
sfc-gh-aalam, line 934: this test has order of telemetry reversed in sql simplifier disabled case. I'm adding this change here to test fix that.

src/snowflake/snowpark/dataframe.py
```diff
@@ -995,14 +996,33 @@ def select(
...
         table_func = None
+        func_expr = None
         join_plan = None
 
         for e in exprs:
             if isinstance(e, Column):
-                names.append(e._named())
+                expr = e._expression
+                left_cols.append(self._session._analyzer.analyze(expr, {}))
+
```
sfc-gh-aalam, line 1009:
```suggestion
                left_cols.append(
                    self._session._analyzer.analyze(
                        expr, self._plan.df_aliased_col_name_to_real_col_name
                    )
                )

```

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -325,16 +325,41 @@ def lateral_statement(lateral_expression: str, child: str) -> str:
...
-def join_table_function_statement(func: str, child: str) -> str:
+def join_table_function_statement(
+    func: str,
+    child: str,
+    left_cols: List[str],
+    right_cols: List[str],
+    use_constant_subquery_alias: bool,
+) -> str:
+    left_alias = (
+        "T_LEFT"
```
sfc-gh-sfan, line 336: Just curious: do we need to worry about alias conflict if there are multiple joins?

src/snowflake/snowpark/dataframe.py
sfc-gh-sfan: I'm not familiar with plan generation, so the only idea I have is to extract the "output column generation" part of the `TableFunctionJoin ` plan generation as a shared function. Note that this might not be easy at all and feel free to ignore if it is not straightforward.

tests/integ/test_telemetry.py
sfc-gh-sfan: Could you be more explicit on the error message? From a quick glance our assertion is on `dict` being equal and I think `dict` is not sensitive to ordering.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-aalam: when table joins are involved, we only allow a single table function coming in. I am not allowing aliasing the left child but I have removed it now - is not possible to cover all the cases

tests/integ/test_telemetry.py
sfc-gh-aalam: The telemetry message log is a list of dict of dicts. In our tests, we extract a specific index using `extract_telemetry_log_data` using `index` like `-1` or `-2` saying select the last or second last entry of the message log.

When I test this without simplifier enabled, the order of telemetry messages in the message log is flipped but the message is present in both cases. I'm not sure if this is the best way to fix this error though - I am open to more suggestions here.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-stan: We don't flatten joins so a join is always binary and subquery alias only persists at one level, so subquery aliases won't conflict.

sfc-gh-stan (approved):

sfc-gh-sfan (approved):

tests/integ/test_telemetry.py
sfc-gh-aalam: updated based on your suggestion. This looks much better now

PR #1026, opened 2023-08-28
SNOW-575699: Fix flaky query tag tests

sfc-gh-yixie:

   Fixes #SNOW-575699, #574078, #574166, #759410

The Snowflake server side query tag has latency so the client tests are flaky.
Instead of retrieving the query's query tag from the server side, we now record what query tags are sent to the server.

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -319,6 +319,11 @@ class ServerConnection:
         for listener in self._query_listener:
             listener._add_query(query_record)
 
+    def notify_debug_info_listeners(self, debug_info: Optional[Dict[str, Any]]):
+        if debug_info:
+            for listener in self._query_listener:
+                listener._add_debug_info(debug_info)
+
     @_Decorator.wrap_exception
     def run_query(
         self,
@@ -343,6 +348,7 @@ class ServerConnection:
                 if not kwargs.get("_statement_params"):
                     kwargs["_statement_params"] = {}
                 kwargs["_statement_params"]["SNOWPARK_SKIP_TXN_COMMIT_IN_DDL"] = True
+            statement_params = kwargs.get("_statement_params")
             if block:
                 results_cursor = self._cursor.execute(query, params=params, **kwargs)
                 self.notify_query_listeners(
@@ -359,6 +365,7 @@ class ServerConnection:
                 logger.debug(
                     f"Execute async query [queryID: {results_cursor['queryId']}] {query}"
                 )
+            self.notify_debug_info_listeners(statement_params)
         except Exception as ex:
             if log_on_exception:
                 query_id_log = f" [queryID: {ex.sfqid}]" if hasattr(ex, "sfqid") else ""
@@ -616,6 +623,7 @@ class ServerConnection:
             set_query_tag_cursor = self._cursor.execute(
                 f"alter session set query_tag = {str_to_sql(query_tag)}"
             )
+            self.notify_debug_info_listeners({"set_session_query_tag": query_tag})
             self.notify_query_listeners(
                 QueryRecord(set_query_tag_cursor.sfqid, set_query_tag_cursor.query)
             )
@@ -627,6 +635,7 @@ class ServerConnection:
             unset_query_tag_cursor = self._cursor.execute(
                 "alter session unset query_tag"
             )
+            self.notify_debug_info_listeners({"unset_session_query_tag": query_tag})
             self.notify_query_listeners(
                 QueryRecord(unset_query_tag_cursor.sfqid, unset_query_tag_cursor.query)
             )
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -343,6 +348,7 @@ def run_query(
                 if not kwargs.get("_statement_params"):
                     kwargs["_statement_params"] = {}
                 kwargs["_statement_params"]["SNOWPARK_SKIP_TXN_COMMIT_IN_DDL"] = True
+            statement_params = kwargs.get("_statement_params")
```
sfc-gh-sfan, line 351: Should we inline this line to the `notify_debug_info_listeners`? Or move it right above the `notify` line if we want easier breakpoint experience. Want to avoid changes in between as much as possible. I could see that it is getting the same object, but just in case we decided to do something like recreating `kwargs`.

tests/integ/scala/test_query_tag_suite.py
```diff
@@ -30,36 +32,47 @@ def test_set_query_tag(session, query_tag):
...
+                temp_view_name
+            )
         Utils.drop_view(session, view_name)
         Utils.drop_view(session, temp_view_name)
 
         assert (
-            len(query_history) == 8
-        )  # 7 DataFrame queries + 1 query that get sql history
+            len(get_local_query_tags(query_history)) == 0
+        )  # The session has query tag. Each SQL statement shouldn't have its own query tag in the statement params.
```
sfc-gh-sfan, line 69: We probably should also check session level param? Are we able to do that? Otherwise this test does not seem very useful because we could get the same result by not setting any tags.

tests/integ/scala/test_query_tag_suite.py
```diff
@@ -91,37 +104,75 @@ def test_query_tags_from_trackback(session, code):
         globals(),
     )
     random_name_func = globals().get(f"{random_name}_func")
-    random_name_func(session)
-    query_history = get_query_history_for_tags(session, random_name)
-    assert len(query_history) == 1
+    if installed_pandas or "pandas" not in code:
```
sfc-gh-sfan, line 107: Curious: why do we need `pandas` here?

tests/integ/scala/test_query_tag_suite.py
```diff
@@ -91,37 +104,75 @@ def test_query_tags_from_trackback(session, code):
...
+    session.query_tag = query_tag
+    try:
+        with session.query_history() as query_history:
+            session.create_dataframe(
+                [["a"] * (ARRAY_BIND_THRESHOLD + 1)]
+            ).count()  # trigger large local relation query
+        set_session_query_tag = [
+            x for x in query_history._debug_info if "set_session_query_tag" in x
+        ]
+        assert len(set_session_query_tag) == 0
```
sfc-gh-sfan, line 124: Is 0 expected? Should we add comment about this result?

tests/integ/scala/test_query_tag_suite.py
```diff
@@ -91,37 +104,75 @@ def test_query_tags_from_trackback(session, code):
...
+@pytest.mark.parametrize("data", ["a", "'a'", "\\a", "a\n", r"\ua", " a", '"a'])
+def test_large_local_relation_query_tag_from_traceback(session, data):
+    with session.query_history() as query_history:
+        session.create_dataframe(
+            [[data] * (ARRAY_BIND_THRESHOLD + 1)]
+        ).count()  # trigger large local relation query
+    set_session_query_tag = [
+        x for x in query_history._debug_info if "set_session_query_tag" in x
+    ]
+    assert len(set_session_query_tag) == 1
```
sfc-gh-sfan, line 142: Where did we set the tag such that it becomes one 🤔

sfc-gh-sfan (commented):
Overall I'm a bit worried about not asserting on session level params because it does not give us the same coverage. Should we do a `SHOW PARAMETERS LIKE` to assert value?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -319,6 +319,11 @@ def notify_query_listeners(self, query_record: QueryRecord) -> None:
         for listener in self._query_listener:
             listener._add_query(query_record)
 
+    def notify_debug_info_listeners(self, debug_info: Optional[Dict[str, Any]]):
+        if debug_info:
```
sfc-gh-stan, line 323: I prefer the null check to happen outside the function, if `debug_info` is `None`, this function will be a noop.

src/snowflake/snowpark/query_history.py
```diff
@@ -24,6 +24,9 @@ class QueryHistory:
     def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
         self.session = session
         self._queries: List[QueryRecord] = []
+        self._debug_info: List[
```
sfc-gh-stan, line 27: Should we introduce this just to pass the tests?

PR #1031, opened 2023-08-29
SNOW-900112 fix table sproc register from file

### LINKED ISSUE #1019: SNOW-900112: register_from_file() does not work with tabular stored procedures
Simply said, register_from_file does not work for tabular stored procedures.
Please find very small reproducable example attached.

```
from snowflake.snowpark.dataframe import DataFrame
# This works
@sproc(packages=['snowflake-snowpark-python'])
def test_sproc(session: Session) -> DataFrame:
    return session.range(5)

test_sproc().show()

# This results in error: "Pandas stored procedure is not supported"
# content of sproc_code.py:
# --------------
# from snowflake.snowpark.session import Session
# from snowflake.snowpark.dataframe import DataFrame
#
# def test_sproc(session: Session) -> DataFrame:
#     return session.range(5)
session.sproc.register_from_file(
    file_path='sproc_code.py', 
    func_name='test_sproc',
    packages=['snowflake-snowpark-python']
)
```

sfc-gh-aalam:

   Fixes #1019 

   Fix type hint detection when registering table sprocs using `register_from_file`.

### DIFF
--- a/src/snowflake/snowpark/_internal/type_utils.py
+++ b/src/snowflake/snowpark/_internal/type_utils.py
@@ -31,6 +31,7 @@ from typing import (  # noqa: F401
 
 import snowflake.snowpark.types  # type: ignore
 from snowflake.connector.options import installed_pandas, pandas
+from snowflake.snowpark._internal.utils import TempObjectType
 from snowflake.snowpark.types import (
     LTZ,
     NTZ,
@@ -406,7 +407,7 @@ def merge_type(a: DataType, b: DataType, name: Optional[str] = None) -> DataType
         return a
 
 
-def python_type_str_to_object(tp_str: str) -> Type:
+def python_type_str_to_object(tp_str: str, object_type: str) -> Type:
     # handle several special cases, which we want to support currently
     if tp_str == "Decimal":
         return decimal.Decimal
@@ -416,6 +417,8 @@ def python_type_str_to_object(tp_str: str) -> Type:
         return datetime.time
     elif tp_str == "datetime":
         return datetime.datetime
+    elif tp_str == "DataFrame" and object_type == TempObjectType.PROCEDURE:
+        return snowflake.snowpark.DataFrame
     elif tp_str in ["Series", "pd.Series"] and installed_pandas:
         return pandas.Series
     elif tp_str in ["DataFrame", "pd.DataFrame"] and installed_pandas:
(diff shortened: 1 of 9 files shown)


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -571,7 +589,7 @@ def retrieve_func_type_hints_from_source(
 ) -> Optional[Dict[str, str]]:
     """
     Retrieve type hints of a function from a source file, or a source string (test only).
-    Returna None if the function is not found.
+    Return None if the function is not found.
     """
```
sfc-gh-aalam, line 593:
```suggestion
    Returns None if the function is not found.
    """
```

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -406,7 +407,7 @@ def merge_type(a: DataType, b: DataType, name: Optional[str] = None) -> DataType
         return a
 
 
-def python_type_str_to_object(tp_str: str) -> Type:
+def python_type_str_to_object(tp_str: str, object_type: str) -> Type:
```
sfc-gh-sfan, line 410: Should we assert on the expected `object_type`?

sfc-gh-sfan: Basically checking if `object_type` is of the expected types (I imagine only UDXF and sp). Are there other use cases that might call this?

sfc-gh-aalam: We also call it with `python_type_to_snow_type` in `src/snowflake/snowpark/dataframe_na_functions.py` which is none of the objects in `TempObjectType` enums. Right now, I'm calling this function with `object_type=""`

sfc-gh-sfan: Another option is to just pass in `is_return_type_of_sproc` and default to `False`, and if we need more than `sproc`, we could refactor it to take in `object_type` 🤔 . Will this make it clearer?

sfc-gh-aalam: sounds like a better alternative. will work on it.

tests/unit/test_types.py
```diff
@@ -549,7 +564,7 @@ def check_type(python_type, snow_type, is_nullable, type_str_override=None):
 
     # invalid type str
     with pytest.raises(NameError):
-        python_type_to_snow_type("string")
+        python_type_to_snow_type("string", "")
```
sfc-gh-sfan, line 567: Do we still need this?

sfc-gh-sfan (approved):
We might even be able to use callstack to deduce if we are in a stored proc creation....but i'm not sure how stable that is so we could keep things as is.

tests/unit/test_types.py
sfc-gh-aalam: no, must have missed it in the cleanup

PR #1032, opened 2023-08-30
SNOW-903907: Custom package upload bug fix

sfc-gh-aalam:

   Fixes #SNOW-903907

### DIFF
--- a/src/snowflake/snowpark/_internal/packaging_utils.py
+++ b/src/snowflake/snowpark/_internal/packaging_utils.py
@@ -249,6 +249,8 @@ def identify_supported_packages(
         channel.
         native_packages (Set[str]): Set of packages that contain native code. (either packages requested by users and
         unavailable in anaconda or dependencies of requested packages)
+        package_dict (Dict[str, str]): A dictionary of package name -> package spec of packages that have
+            been added explicitly so far using add_packages() or other such methods.
 
     Returns:
         Tuple[List[Requirement], List[Requirement], List[Requirement]]: Tuple containing dependencies that are present
--- a/src/snowflake/snowpark/_internal/packaging_utils.py
+++ b/src/snowflake/snowpark/_internal/packaging_utils.py
@@ -233,6 +233,7 @@ def identify_supported_packages(
     packages: List[Requirement],
     valid_packages: Dict[str, List[str]],
     native_packages: Set[str],
+    package_dict: Dict[str, str],
 ) -> Tuple[List[Requirement], List[Requirement], List[Requirement]]:
     """
     Detects which `packages` are present in the Snowpark Anaconda channel using the `valid_packages` mapping.
@@ -283,12 +284,13 @@ def identify_supported_packages(
 
             # Native packages should be anaconda dependencies, even if the requested version is not available.
             elif package_name in native_packages:
-                _logger.warning(
-                    f"Package {package_name}{version_text} contains native code, switching to latest available version "
-                    f"in Snowflake '{valid_packages[package_name][-1]}' instead."
-                )
+                if package_name not in package_dict:
+                    _logger.warning(
+                        f"Package {package_name}{version_text} contains native code, switching to latest available version "
+                        f"in Snowflake instead."
+                    )
+                    new_dependencies.append(Requirement.parse(package_name))
                 dropped_dependencies.append(package)
-                new_dependencies.append(Requirement.parse(package_name))
 
             else:
                 packages_to_be_uploaded.append(str(package))
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -1186,6 +1186,8 @@ class Session:
         Args:
             packages (List[str]): List of package names requested by the user, that are not present in Snowflake.
             package_table (str): Name of Snowflake table containing information about Anaconda packages.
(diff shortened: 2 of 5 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -1177,6 +1178,7 @@ def _upload_unsupported_packages(
         self,
         packages: List[str],
         package_table: str,
+        package_dict: Dict[str, str],
```
sfc-gh-stan, line 1181: Let's add some documentation about `package_dict` in the doctsring.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -233,6 +233,7 @@ def identify_supported_packages(
     packages: List[Requirement],
     valid_packages: Dict[str, List[str]],
     native_packages: Set[str],
+    package_dict: Dict[str, str],
```
sfc-gh-stan, line 236: Same as below: add an arg description please.

src/snowflake/snowpark/_internal/packaging_utils.py
```diff
@@ -283,12 +286,13 @@ def identify_supported_packages(
...
-                    f"Package {package_name}{version_text} contains native code, switching to latest available version "
-                    f"in Snowflake '{valid_packages[package_name][-1]}' instead."
-                )
+                if package_name not in package_dict:
+                    _logger.warning(
+                        f"Package {package_name}{version_text} contains native code, switching to latest available version "
+                        f"in Snowflake instead."
+                    )
+                    new_dependencies.append(Requirement.parse(package_name))
                 dropped_dependencies.append(package)
```
sfc-gh-sfan, line 295: Is it correct that we add an existing package to `dropped_dependencies`? I might miss some context so just want to do a sanity check.

src/snowflake/snowpark/session.py
```diff
@@ -1158,9 +1159,9 @@ def get_req_identifiers_list(
 
                 # Add to packages dictionary
                 if name in result_dict:
-                    if result_dict[name] != str(package):
+                    if version is not None and result_dict[name] != str(package):
```
sfc-gh-sfan, line 1162: Is it possible to add the same package + version when it already exists? Or do we already deduplicate the packages (e.g. via a dict)?

tests/integ/test_packaging.py
```diff
@@ -408,13 +413,34 @@ def test_add_unsupported_requirements_twice_should_not_fail_for_same_requirement
     IS_IN_STORED_PROC,
     reason="Subprocess calls are not allowed within stored procedures.",
 )
+@pytest.mark.skipif(
+    not runtime_39_or_above, reason="arch is not available on python 3.8"
```
sfc-gh-sfan, line 417: Just curious, why do we have to use `arch` for testing? Does it bring in a different version of `scipy` or `pandas` that satisfies our test scenario?

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-aalam: here the packages in the list `dropped_dependencies` are the packages that were downloaded by pip but need to be deleted from temp directory before the directory is zipped and uploaded to stage. `new_dependencies` is the list of packages available in snowflake that are dependencies of requested external packages.

sfc-gh-sfan: Makes sense. Thanks for the explanation. Maybe `dependencies_to_drop` is a better name :p

src/snowflake/snowpark/session.py
sfc-gh-aalam: we first add packages to `result_dict` that are available in snowflake and are added via `add_packages()` (or other such methods). Then, we pip install other packages and figure out dependencies. Now there can be dependency clashes if requested dependency add via add_package is `package_a==x.y.1` and inferred dependency via pip installation is `package_a==x.y.2`. In this case, resolve packages should throw an error.

In the updated logic, we ignore case when inferred dependency is not version specific, in other words inferred depndency `package_a` is compatible with `package_a==x.y.1`. This case would throw error in resolve packages earlier but passes now.

tests/integ/test_packaging.py
sfc-gh-aalam: this is the scenario that failed in the issue raised by users so I'm using the same example. it downloads `scipy` and `pandas` that triggers the scenario mentioned above.

src/snowflake/snowpark/session.py
sfc-gh-sfan: If we have a `package_a==x.y.1` from `add_package` and the same `package_a==x.y.1` from `pip`, we should pass but I'm not sure if this is a legit case, or if our code has this scenario covered already (it might be covered by the original `result_dict[name] != str(package)`).

sfc-gh-aalam: that's right, in the case you mentioned, `result_dict[name] == str(package)`

src/snowflake/snowpark/_internal/packaging_utils.py
sfc-gh-aalam: I agree

sfc-gh-stan (approved):
LGTM

sfc-gh-sfan (approved):

PR #1036, opened 2023-08-31
SNOW-857185: Give an option to log on exception when calling sp with …

sfc-gh-aalam:
…optional args

   Fixes #SNOW-857185

   Don't log warning messages when trying to infer if stored proc has table return type unless users explicitly request it.

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -2355,7 +2355,9 @@ class Session:
         """
         return self._sp_registration
 
-    def _infer_is_return_table(self, sproc_name: str, *args: Any) -> bool:
+    def _infer_is_return_table(
+        self, sproc_name: str, *args: Any, log_on_exception: bool = False
+    ) -> bool:
         try:
             arg_types = [convert_sp_to_sf_type(infer_type(arg)) for arg in args]
             func_signature = f"{sproc_name.upper()}({', '.join(arg_types)})"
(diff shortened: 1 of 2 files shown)


sfc-gh-sfan (commented):
This means we are not able to return a dataframe if the user tries to invoke a table sproc with a `None` value on a field that is not string. Do we need to be concerned? Are our users able to provide a cast (e.g. an expression that can be translated to `NULL :: INT` for example)?

sfc-gh-aalam:
> This means we are not able to return a dataframe if the user tries to invoke a table sproc with a `None` value on a field that is not string. Do we need to be concerned? Are our users able to provide a cast (e.g. an expression that can be translated to `NULL :: INT` for example)?

That is a good concern. We won't be able to do so right now. I'll work on a fix for it.

src/snowflake/snowpark/session.py
```diff
@@ -2355,19 +2355,24 @@ def sproc(self) -> StoredProcedureRegistration:
...
-            sproc_desc = self._run_query(f"describe procedure {func_signature}")
+            sproc_desc = self._run_query(
+                f"describe procedure {func_signature}",
+                log_on_exception=log_on_exception,
+            )
             return_type = sproc_desc[1][1]
             return return_type.upper().startswith("TABLE")
         except Exception as exc:
-            _logger.warn(
+            _logger.info(
```
sfc-gh-sfan, line 2375: Should this be guard by `log_on_exception`? I have a mixed feeling because I think this is actually a very useful log, so I prefer to keep it, but just want to raise it.

sfc-gh-sfan (approved):

src/snowflake/snowpark/session.py
sfc-gh-aalam: The main concern here was if the describe procedure fails to find procedure, users see warning message even when things work successfully. There were two sources of logs, one here and one inside `_run_query`. 
Now the log message in `_run_query` is controlled by `log_on_exception`. I though keeping this one uniformly log info should be good. In this case, users can have messages generated at warning level when they want, else message go log info.

PR #1038, opened 2023-09-05
[Snowpark] Add arg for volatile/immutable UDF

sfc-gh-tbao:

   SNOW-903916

This PR adds keyword support for volatile/immutable when registering Snowpark UDFs.

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -971,6 +971,7 @@ def create_python_udf_or_sp(
     secure: bool = False,
     external_access_integrations: Optional[List[str]] = None,
     secrets: Optional[Dict[str, str]] = None,
+    immutable: bool = False,
 ) -> None:
     runtime_version = (
         f"{sys.version_info[0]}.{sys.version_info[1]}"
@@ -1010,8 +1011,9 @@ $$
         if inline_python_code
         else ""
     )
-
+    mutability = "IMMUTABLE" if immutable else "VOLATILE"
     strict_as_sql = "\nSTRICT" if strict else ""
+
     external_access_integrations_in_sql = (
         f"\nEXTERNAL_ACCESS_INTEGRATIONS=({','.join(external_access_integrations)})"
         if external_access_integrations
@@ -1028,6 +1030,7 @@ CREATE{" OR REPLACE " if replace else ""}
 {"" if is_permanent else "TEMPORARY"} {"SECURE" if secure else ""} {object_type.value.replace("_", " ")} {"IF NOT EXISTS" if if_not_exists else ""} {object_name}({sql_func_args})
 {return_sql}
 LANGUAGE PYTHON {strict_as_sql}
+{mutability}
 RUNTIME_VERSION={runtime_version}
 {imports_in_sql}
 {packages_in_sql}
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -6532,6 +6532,7 @@ def udf(
     secure: bool = False,
     external_access_integrations: Optional[List[str]] = None,
     secrets: Optional[Dict[str, str]] = None,
+    immutable: bool = False,
 ) -> Union[UserDefinedFunction, functools.partial]:
     """Registers a Python function as a Snowflake Python UDF and returns the UDF.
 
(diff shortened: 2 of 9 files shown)


sfc-gh-tbao:

sfc-gh-tbao:
recheck

sfc-gh-stan:
The changes LGTM. Please add an entry in `CHANGELOG.md` to document this addition.

sfc-gh-stan (approved):

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -971,6 +971,7 @@ def create_python_udf_or_sp(
     secure: bool = False,
     external_access_integrations: Optional[List[str]] = None,
     secrets: Optional[Dict[str, str]] = None,
+    immutable: bool = False,
```
sfc-gh-aling, line 974: from the definition I think this immutable is optional: https://docs.snowflake.com/en/sql-reference/sql/create-function#python-handler

so I think the parameters should be `immutable: Optional[bool] = None`
```suggestion
    immutable: bool = False,
```

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1010,8 +1011,9 @@ def create_python_udf_or_sp(
         if inline_python_code
         else ""
     )
-
+    mutability = "IMMUTABLE" if immutable else "VOLATILE"
```
sfc-gh-aling, line 1014: depending on `immutable` being optional, I think this would be

```python
mutability = ""
if immutable is not None:
    mutability = "IMMUTABLE" if immutable else "VOLATILE"

File: src/snowflake/snowpark/functions.py [hunk 3]
```diff
@@ -6532,6 +6532,7 @@ def udf(
     secure: bool = False,
     external_access_integrations: Optional[List[str]] = None,
     secrets: Optional[Dict[str, str]] = None,
+    immutable: bool = False,
```
sfc-gh-aling, line 6535: same comments as above

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-tbao: If we agree on the default VOLATILE behavior, then we don't need this :)

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aling, line 974: can you check if those are the two possible values for this? such that we don't plan to introduce more in the future at least in short term.

because I'm thinking of if there's a third value for this parameter, however boolean which only represents two is not enough.

in contrast, another way to do is introducing two parameters in your PR:

```
immutable: Optional[bool] = None
volatile: Optional[bool] = None
```

sfc-gh-aling (changes requested):
please see my comments regarding the values of this field

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-tbao: the keyword is for determinism of the UDF, which there're are only two options.

sfc-gh-aling: oh you're right, I overlooked the part that it has a default value, thanks for calling it out.
now I'm thinking whether the signature should be `immutable = False` or `volatile = True`

@sfc-gh-jdu @sfc-gh-sfan @sfc-gh-aalam do you know if this parameter is supported in the scala udf?

sfc-gh-tbao: Any luck on this PR review?

For naming I'll advocate for immutable since the antonym is mutable, but volatile doesn't have that.
TBH, I'll prefer deterministic as a better naming, but that's inconsistent with SQL keyword, so let's keep it.

sfc-gh-aalam: I'm okay with naming it `immutable=False`. Other option is we have something like `mutability : Literal["volatile", "immutable"] = "volatile"`

sfc-gh-tbao: @sfc-gh-aling shall we proceed with this?

sfc-gh-aling (approved):

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aling: yes, I'm also okay with immutable=False

PR #857, opened 2023-05-23
Fix Code generation and Types to be compatible with Python 3.11+

DmitryErmolchik:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes: Not in Jira

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [x] I am fixing code generator on Python 3.11
   - [x] I am adding source code compatibility with Python 3.11

3. Please describe how your code solves the related issue.

   I'm using this project as third-party library to in my project. This project runs on Python 3.10, Python 3.11.
   The only one feature I'm use is generate SQL queries to request data from Snowflake, but project team doesn't support Python 3.9+. To improve my build quality I started run unit tests, but one or two of them failed.
   The issue is in new Python 3.11 commands flow who discovered the error in the extract_func_global_refs method.
   I rewrote it with the correct logic. Additionally I send you patch to fix source codes incompatibility with Python 3.11

   P.S.: All unit tests passed with Python 3.9, Python 3.10, Python 3.11.

PR #1040, opened 2023-09-06
SNOW-900365: Add Python 3.11 merge gate

sfc-gh-aling:

   Fixes SNOW-900365

### DIFF
--- a/src/snowflake/snowpark/_internal/code_generation.py
+++ b/src/snowflake/snowpark/_internal/code_generation.py
@@ -180,11 +180,10 @@ def get_class_references(
 
 def extract_func_global_refs(code: CodeType) -> Set[str]:
     # inspired by cloudpickle to recursively extract all the global references used by the target func's code object
-    co_names = set(code.co_names)
     out_names = set()
     for instr in dis.get_instructions(code):
         op = instr.opcode
-        if op in GLOBAL_OPS and instr.argval in co_names:
+        if op in GLOBAL_OPS:
             out_names.add(instr.argval)
 
     if code.co_consts:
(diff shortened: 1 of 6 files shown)


sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/code_generation.py
```diff
@@ -180,12 +180,12 @@ def get_class_references(
...
     # inspired by cloudpickle to recursively extract all the global references used by the target func's code object
-    co_names = code.co_names
+    co_names = set(code.co_names)
     out_names = set()
     for instr in dis.get_instructions(code):
         op = instr.opcode
-        if op in GLOBAL_OPS:
-            out_names.add(co_names[instr.arg])
+        if op in GLOBAL_OPS and instr.argval in co_names:
+            out_names.add(instr.argval)
```
sfc-gh-sfan, line 188: Sanity check here: Is it the case that previously `co_names` is a tuple/list and `instr.arg` is an index, and now we changed it such that we directly add value if it exists?

Is there a case where `instr.argval in co_names` can be `False`? This seems to be a logic change given previously we just check if `op` is in `GLOBAL_OPS`.

sfc-gh-aling: thanks for the sanity check -- I was copying the change from the external PR: https://github.com/snowflakedb/snowpark-python/pull/857 but I think you are right.

after double checking cloudpickle's implementation, the co_names is not used anymore and only `argval` is used: https://github.com/cloudpipe/cloudpickle/commit/6a0e12d058d1bd3ab26ec000ac2249b4ee7e9c9f.

I have updated the code to align with cloudpickle's latest implementation.

sfc-gh-sfan (approved):

PR #1044, opened 2023-09-08
SNOW-900365: update cloudpickle dep for python 3.11 support

sfc-gh-aling:

   Fixes SNOW-900365, we also need to update cloudpickle to support py 3.11

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,9 +11,20 @@ THIS_DIR = os.path.dirname(os.path.realpath(__file__))
 SRC_DIR = os.path.join(THIS_DIR, "src")
 SNOWPARK_SRC_DIR = os.path.join(SRC_DIR, "snowflake", "snowpark")
 CONNECTOR_DEPENDENCY_VERSION = ">=3.2.0, <4.0.0"
+INSTALL_REQ_LIST = [
+    "setuptools>=40.6.0",
+    "wheel",
+    "cloudpickle>=1.6.0,<=2.0.0",
+    f"snowflake-connector-python{CONNECTOR_DEPENDENCY_VERSION}",
+    "pyyaml",
+]
 REQUIRED_PYTHON_VERSION = ">=3.8, <3.11"
 if os.getenv("SNOWFLAKE_IS_PYTHON_RUNTIME_TEST", False):
     REQUIRED_PYTHON_VERSION = ">=3.8"
+    # replace the line with condition py < 3.11
+    INSTALL_REQ_LIST[2] = "cloudpickle>=1.6.0,<=2.0.0;python_version<'3.11'"
+    # add one more line for py ~= 3.11
+    INSTALL_REQ_LIST.append("cloudpickle>=2.1.0,<3.0.0;python_version~='3.11'")
 
 # read the version
 VERSION = ()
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -11,9 +11,20 @@
...
+    f"snowflake-connector-python{CONNECTOR_DEPENDENCY_VERSION}",
+    "pyyaml",
+]
 REQUIRED_PYTHON_VERSION = ">=3.8, <3.11"
 if os.getenv("SNOWFLAKE_IS_PYTHON_RUNTIME_TEST", False):
     REQUIRED_PYTHON_VERSION = ">=3.8"
+    # replace the line with condition py < 3.11
+    INSTALL_REQ_LIST[2] = "cloudpickle>=1.6.0,<=2.0.0;python_version<'3.11'"
+    # add one more line for py ~= 3.11
+    INSTALL_REQ_LIST.append("cloudpickle>=2.2.1,<3.0.0;python_version~='3.11'")
```
sfc-gh-aling, line 27: 2.2.1 is available in the anaconda channel, but not 2.2.0, 2.1.0. cloudpickle starts supporting python 3.11 from 2.1.0.

checking with ananconda team to see if it's on purpose that 2.1.0/2.2.0 are skipped

sfc-gh-aling, line 27: 2.2.1 right now is the only available version which supports py 311 in our anaconda channel

setup.py
```diff
@@ -11,9 +11,20 @@
...
+    "wheel",
+    "cloudpickle>=1.6.0,<=2.0.0",
+    f"snowflake-connector-python{CONNECTOR_DEPENDENCY_VERSION}",
+    "pyyaml",
+]
 REQUIRED_PYTHON_VERSION = ">=3.8, <3.11"
 if os.getenv("SNOWFLAKE_IS_PYTHON_RUNTIME_TEST", False):
     REQUIRED_PYTHON_VERSION = ">=3.8"
+    # replace the line with condition py < 3.11
+    INSTALL_REQ_LIST[2] = "cloudpickle>=1.6.0,<=2.0.0;python_version<'3.11'"
```
sfc-gh-sfan, line 25: I wonder if it makes sense to make a `CLOUDPICKLE_REQ_LIST`, and we use the current if clause to determine the value. After we determine the value of `CLOUDPICKLE_REQ_LIST`, we then add the list into `INSTALL_REQ_LIST`. This way we don't need to keep a magic number 2 here.

sfc-gh-stan: I second this.

sfc-gh-aling: good idea, I updated the PR, let me know how it looks

setup.py
```diff
@@ -11,9 +11,30 @@
...
 CONNECTOR_DEPENDENCY_VERSION = ">=3.2.0, <4.0.0"
+INSTALL_REQ_LIST = [
+    "setuptools>=40.6.0",
+    "wheel",
+    f"snowflake-connector-python{CONNECTOR_DEPENDENCY_VERSION}",
+    "pyyaml",
+]
+CLOUDPICKLE_REQ_LIST = ["cloudpickle>=1.6.0,<=2.0.0"]
+# this allows us to update the cloudpickle dependency by env var when building package privately
+# without re-releasing package just to update the dependency, example if we want to lower cloudpickle dep to 2.1.0:
```
sfc-gh-stan, line 22: Just curious, why would we want to pin cloudpickle version in private builds?

sfc-gh-aling: I did this mainly to pin the lower bound, cloudpickle >= 2.1.0 supports python 3.11, but right now we are testing the latest cloudpickle 2.2.1 in our channel, we don't have 2.1.0.

if not pin the lower bound, cloudpickle < 2.1.0 won't work with snowpark-python in py311 env

setup.py
```diff
@@ -11,9 +11,30 @@
...
+# SNOWFLAKE_IS_PYTHON_RUNTIME_TEST=1 CLOUDPICKLE_PYTHON_311_DEPENDENCY="cloudpickle==2.1.0;python_version~='3.11'" pip install .
+CLOUDPICKLE_PYTHON_311_DEPENDENCY = os.getenv(
```
sfc-gh-sfan, line 24: nit: This can go inside `if os.getenv("SNOWFLAKE_IS_PYTHON_RUNTIME_TEST", False):` I think? We could move it out of the if branch the same time as we make 3.11 available public

sfc-gh-sfan (approved):

PR #1049, opened 2023-09-13
[Local Testing] SNOW-783554 Support selecting window expressions

sfc-gh-stan:
- [x] Make non-rank related functions with unspecified window frame use cumulative window by default
- [ ] Support aggregating window expressions
- [x] Support more rank related functions/Refactor to allow patching for rank related functions (This will be addressed in a separate PR).

### DIFF
--- a/src/snowflake/snowpark/mock/analyzer.py
+++ b/src/snowflake/snowpark/mock/analyzer.py
@@ -247,7 +247,7 @@ class MockAnalyzer:
                 "[Local Testing] group by grouping sets is not implemented."
             )
 
-        if isinstance(expr, WindowExpression):  # n
+        if isinstance(expr, WindowExpression):
             return window_expression(
                 self.analyze(
                     expr.window_function,
@@ -506,11 +506,6 @@ class MockAnalyzer:
             if parse_local_name:
                 expr_str = expr_str.upper()
             return expr_str
-            # expr_str = (
-            #    expr.name
-            #    if expr.name
-            #    else self.analyze(expr.child, expr_to_alias, parse_local_name)
-            # )
         elif isinstance(expr, Cast):
             return cast_expression(
                 self.analyze(expr.child, expr_to_alias, parse_local_name),
--- a/src/snowflake/snowpark/mock/analyzer.py
+++ b/src/snowflake/snowpark/mock/analyzer.py
@@ -308,7 +308,7 @@ class MockAnalyzer:
                 self.session._conn._telemetry_client.send_function_usage_telemetry(
                     expr.api_call_source, TelemetryField.FUNC_CAT_USAGE.value
                 )
-            func_name = expr.name.upper() if parse_local_name else expr.name
+            func_name = expr.name.upper()
             return function_expression(
                 func_name,
                 [self.to_sql_avoid_offset(c, expr_to_alias) for c in expr.children],
--- a/src/snowflake/snowpark/mock/analyzer.py
+++ b/src/snowflake/snowpark/mock/analyzer.py
@@ -27,12 +27,15 @@ from snowflake.snowpark._internal.analyzer.analyzer_utils import (
     range_statement,
     rank_related_function_expression,
     regexp_expression,
+    specified_window_frame_expression,
     subfield_expression,
     subquery_expression,
     table_function_partition_spec,
     unary_expression,
     update_merge_statement,
+    window_expression,
     window_frame_boundary_expression,
+    window_spec_expression,
     within_group_expression,
 )
 from snowflake.snowpark._internal.analyzer.binary_expression import (
@@ -244,19 +247,39 @@ class MockAnalyzer:
                 "[Local Testing] group by grouping sets is not implemented."
             )
 
-        if isinstance(expr, WindowExpression):
-            raise NotImplementedError(
-                "[Local Testing] window function is not implemented."
+        if isinstance(expr, WindowExpression):  # n
+            return window_expression(
+                self.analyze(
+                    expr.window_function,
+                    parse_local_name=parse_local_name,
+                ),
+                self.analyze(
+                    expr.window_spec,
+                    parse_local_name=parse_local_name,
+                ),
             )
 
         if isinstance(expr, WindowSpecDefinition):
-            raise NotImplementedError(
-                "[Local Testing] window function is not implemented."
+            return window_spec_expression(
+                [
+                    self.analyze(x, parse_local_name=parse_local_name)
+                    for x in expr.partition_spec
+                ],
+                [
+                    self.analyze(x, parse_local_name=parse_local_name)
+                    for x in expr.order_spec
+                ],
+                self.analyze(
+                    expr.frame_spec,
+                    parse_local_name=parse_local_name,
+                ),
             )
 
         if isinstance(expr, SpecifiedWindowFrame):
-            raise NotImplementedError(
-                "[Local Testing] window function is not implemented."
+            return specified_window_frame_expression(
+                expr.frame_type.sql,
+                self.window_frame_boundary(self.to_sql_avoid_offset(expr.lower, {})),
+                self.window_frame_boundary(self.to_sql_avoid_offset(expr.upper, {})),
             )
 
         if isinstance(expr, UnspecifiedFrame):
@@ -483,6 +506,11 @@ class MockAnalyzer:
             if parse_local_name:
                 expr_str = expr_str.upper()
             return expr_str
+            # expr_str = (
+            #    expr.name
+            #    if expr.name
+            #    else self.analyze(expr.child, expr_to_alias, parse_local_name)
+            # )
         elif isinstance(expr, Cast):
             return cast_expression(
                 self.analyze(expr.child, expr_to_alias, parse_local_name),
--- a/src/snowflake/snowpark/mock/functions.py
+++ b/src/snowflake/snowpark/mock/functions.py
@@ -139,7 +139,7 @@ def mock_avg(column: ColumnEmulator) -> ColumnEmulator:
             all_item_is_none = False
             ret += float(data)
             cnt += 1
-    # round to 5 according to snowflake spec
+
     ret = (
         ColumnEmulator(data=[round((ret / cnt), 3)])
         if not all_item_is_none
--- a/src/snowflake/snowpark/mock/functions.py
+++ b/src/snowflake/snowpark/mock/functions.py
@@ -5,7 +5,7 @@
 import datetime
 import math
 from decimal import Decimal
-from typing import Callable, Union, Optional
+from typing import Callable, Union
 
 from snowflake.snowpark.exceptions import SnowparkSQLException
 from snowflake.snowpark.mock.snowflake_data_type import (
@@ -31,7 +31,6 @@ from .util import (
     process_numeric_time,
     process_string_time_with_fractional_seconds,
 )
-from .._internal.type_utils import ColumnOrLiteral
 
 RETURN_TYPE = Union[ColumnEmulator, TableEmulator]
 
@@ -449,13 +448,6 @@ def mock_endswith(expr1: ColumnEmulator, expr2: ColumnEmulator):
     return res
 
 
-@patch("lag")
-def lag(
-    e: ColumnEmulator,
-    offset: int = 1,
-    default_value: Optional[ColumnOrLiteral] = None,
-    ignore_nulls: bool = False,
-    current_idx: int = None,
-) -> ColumnEmulator:
-    pass
-
+@patch("row_number")
+def mock_row_number(window: TableEmulator, row_idx: int):
+    return ColumnEmulator(data=[row_idx + 1], sf_type=ColumnType(LongType(), False))
--- a/src/snowflake/snowpark/mock/functions.py
+++ b/src/snowflake/snowpark/mock/functions.py
@@ -5,7 +5,7 @@
 import datetime
 import math
 from decimal import Decimal
-from typing import Callable, Union
+from typing import Callable, Optional, Union
 
 from snowflake.snowpark.exceptions import SnowparkSQLException
 from snowflake.snowpark.mock.snowflake_data_type import (
@@ -26,6 +26,7 @@ from snowflake.snowpark.types import (
     _NumericType,
 )
 
+from .._internal.type_utils import ColumnOrLiteral
 from .util import (
     convert_snowflake_datetime_format,
     process_numeric_time,
@@ -70,7 +71,9 @@ def patch(function):
 
 @patch("min")
 def mock_min(column: ColumnEmulator) -> ColumnEmulator:
-    if isinstance(column.sf_type.datatype, _NumericType):
+    if isinstance(
+        column.sf_type.datatype, _NumericType
+    ):  # TODO: figure out where 5 is coming from
         return ColumnEmulator(data=round(column.min(), 5), sf_type=column.sf_type)
     res = ColumnEmulator(data=column.dropna().min(), sf_type=column.sf_type)
     try:
@@ -138,7 +141,7 @@ def mock_avg(column: ColumnEmulator) -> ColumnEmulator:
             cnt += 1
     # round to 5 according to snowflake spec
     ret = (
-        ColumnEmulator(data=[round((ret / cnt), 5)])
+        ColumnEmulator(data=[round((ret / cnt), 3)])
         if not all_item_is_none
         else ColumnEmulator(data=[None])
     )
@@ -445,4 +448,13 @@ def mock_endswith(expr1: ColumnEmulator, expr2: ColumnEmulator):
     res.sf_type = ColumnType(StringType(), expr1.sf_type.nullable)
     return res
 
-  
+
+@patch("lag")
+def lag(
+    e: ColumnEmulator,
+    offset: int = 1,
+    default_value: Optional[ColumnOrLiteral] = None,
+    ignore_nulls: bool = False,
+    current_idx: int = None,
+) -> ColumnEmulator:
+    pass
--- a/src/snowflake/snowpark/mock/functions.py
+++ b/src/snowflake/snowpark/mock/functions.py
@@ -5,7 +5,7 @@
 import datetime
 import math
 from decimal import Decimal
-from typing import Callable, Optional, Union
+from typing import Callable, Union, Optional
 
 from snowflake.snowpark.exceptions import SnowparkSQLException
 from snowflake.snowpark.mock.snowflake_data_type import (
@@ -26,12 +26,12 @@ from snowflake.snowpark.types import (
     _NumericType,
 )
 
-from .._internal.type_utils import ColumnOrLiteral
 from .util import (
     convert_snowflake_datetime_format,
     process_numeric_time,
     process_string_time_with_fractional_seconds,
 )
+from .._internal.type_utils import ColumnOrLiteral
 
 RETURN_TYPE = Union[ColumnEmulator, TableEmulator]
 
@@ -458,3 +458,4 @@ def lag(
     current_idx: int = None,
 ) -> ColumnEmulator:
     pass
+
--- a/src/snowflake/snowpark/mock/plan.py
+++ b/src/snowflake/snowpark/mock/plan.py
@@ -7,10 +7,24 @@ import inspect
 import math
 import re
 import uuid
+from bisect import bisect_left, bisect_right
 from enum import Enum
 from functools import cached_property, partial
 from typing import TYPE_CHECKING, Dict, List, NoReturn, Optional, Union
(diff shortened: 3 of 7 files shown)


tests/integ/scala/test_window_frame_suite.py
```diff
@@ -171,11 +174,20 @@ def test_unbounded_rows_range_between_with_aggregation(session):
...
     df = session.create_dataframe(
-        [(1, "1"), (1, "1"), (sys.maxsize, "1"), (3, "2"), (2, "1"), (sys.maxsize, "2")]
+        [
+            (1, "1"),
+            (1, "1"),
+            (sys.maxsize, "1"),
+            (3, "2"),
+            (2, "1"),
+            (sys.maxsize, "2"),
+        ]  # 4 3 1 2 2 1
```
sfc-gh-stan, line 188: remove the comment.

src/snowflake/snowpark/mock/plan.py
```diff
@@ -895,6 +909,215 @@ def calculate_expression(
...
+                            calculate_expression(
+                                window_function.default,
+                                w,
+                                analyzer,
+                                expr_to_alias,
+                                keep_literal=True,
+                            )
+                        )
+                    else:
+                        res_cols.append(
```
sfc-gh-stan, line 1102: use `target_expr` instead

src/snowflake/snowpark/mock/functions.py
```diff
@@ -69,7 +69,9 @@ def wrapper(*args, **kwargs):
 
 @patch("min")
 def mock_min(column: ColumnEmulator) -> ColumnEmulator:
-    if isinstance(column.sf_type.datatype, _NumericType):
+    if isinstance(
+        column.sf_type.datatype, _NumericType
+    ):  # TODO: figure out where 5 is coming from
```
sfc-gh-stan, line 74: Will be addressed in a separate PR.

src/snowflake/snowpark/mock/plan.py
```diff
@@ -895,6 +909,215 @@ def calculate_expression(
...
+        if window_spec.order_spec:
+            res = input_data.sort_values(
+                by=[
+                    exp.child.name for exp in window_spec.order_spec
+                ],  # TODO: consider when the expr is something like col1+2
+                ascending=[
+                    isinstance(exp.direction, Ascending)
+                    for exp in window_spec.order_spec
+                ]
+                or True,
```
sfc-gh-stan, line 927: Consider False?

src/snowflake/snowpark/mock/functions.py
```diff
@@ -137,7 +139,7 @@ def mock_avg(column: ColumnEmulator) -> ColumnEmulator:
             cnt += 1
     # round to 5 according to snowflake spec
     ret = (
-        ColumnEmulator(data=[round((ret / cnt), 5)])
+        ColumnEmulator(data=[round((ret / cnt), 3)])
```
sfc-gh-aling, line 142: there are other places in this module where we round to 5 (I did this previously)
so do we also need to update other mock implementations?

src/snowflake/snowpark/mock/plan.py
```diff
@@ -895,6 +909,215 @@ def calculate_expression(
         column = calculate_expression(exp.child, input_data, analyzer, expr_to_alias)
         column.sf_type = ColumnType(exp.to, exp.nullable)
         return column
+    if isinstance(exp, WindowExpression):
+        # Should return one column
+        window_function = exp.window_function
+        window_spec = exp.window_spec
+
+        # Process order by clause
+        if window_spec.order_spec:
```
sfc-gh-aling, line 918: nit, can be written in the form:

```python
res = input_data.sort_values(...) if window_spec.order_spec else input_data
```

src/snowflake/snowpark/mock/plan.py
```diff
@@ -895,6 +909,215 @@ def calculate_expression(
...
+        # Process partition_by clause
+        if window_spec.partition_spec:
+            res = res.groupby(
+                [exp.name for exp in window_spec.partition_spec],
+                sort=False,
+                as_index=False,
+            )
+            res_index = []
+            for r in res:
+                res_index += list(r[1].index)
```
sfc-gh-aling, line 943: nit, can be simplified as

```python
res_index = [list(r[1].index) for r in res]
```

src/snowflake/snowpark/mock/plan.py
```diff
@@ -895,6 +909,215 @@ def calculate_expression(
...
+
+        if not window_spec.frame_spec or not isinstance(
+            window_spec.frame_spec, SpecifiedWindowFrame
```
sfc-gh-aling, line 946: can remind me what else classes could this `window_spec.frame_spec` be?
from the logic it looks we only support `SpecifiedWindowFrame`,  if there are other `window_spec.frame_spec` that we don't support we should probably document it in the limitations

src/snowflake/snowpark/mock/plan.py
```diff
@@ -895,6 +909,215 @@ def calculate_expression(
...
+                    right_idx = row_idx
+                elif isinstance(upper, UnboundedFollowing):
+                    right_idx = len(window) - 1
+                elif isinstance(upper, Literal):
+                    upper_bound = order_by_values.loc[current_row] + upper.value
+                    right_idx = max(bisect_right(order_by_values, upper_bound), row_idx)
+
+                windows.append(window.iloc[[i for i in range(left_idx, right_idx + 1)]])
+
+        assert len(windows) == len(input_data)
```
sfc-gh-aling, line 997: instead of assertion error, I think we can raise a more user friend error if the size doesn't match?

src/snowflake/snowpark/mock/plan.py
```diff
@@ -895,6 +909,215 @@ def calculate_expression(
...
+                        target_expr = calculate_expression(
+                            window_function.expr,
+                            w.iloc[cur_idx],
+                            analyzer,
+                            expr_to_alias,
+                            keep_literal=True,
+                        )
+                        if target_expr is not None:
+                            cur_count += 1
+                            if cur_count == abs(offset):
```
sfc-gh-aling, line 1088: I have a dumb question: why we stop here when `cur_count == abs(offset)`, this offset is the start offset of the window or the end offset of the window?

src/snowflake/snowpark/mock/functions.py
sfc-gh-stan: Yea, I will address these altogether in a separate PR.

src/snowflake/snowpark/mock/plan.py
sfc-gh-stan: This is for list concatenation. So I need `+=`.

src/snowflake/snowpark/mock/plan.py
sfc-gh-stan: there is `UnspecifiedFrame` which means to use the default window frame, pleas see https://docs.snowflake.com/en/sql-reference/functions-analytic#window-frame-usage-notes.

src/snowflake/snowpark/mock/plan.py
sfc-gh-stan: This is more of a sanity check instead of user facing error, I'll remove it.

src/snowflake/snowpark/mock/plan.py
sfc-gh-stan: Negative offset just means moving in the other direction. That's why I define `delta` and take the absolute value

sfc-gh-yixie:
I tried this
```
from snowflake.snowpark import Session, Window
from snowflake.snowpark.functions import min as min_
from snowflake.snowpark.mock.connection import MockServerConnection

session = Session(MockServerConnection())

df = session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"])
df.select(min_("b").over(Window.partition_by("a").order_by("b"))).show()
df.select(min_("b").over(Window.partition_by("a").order_by("b").range_between(0, 1))).show()
```

Two problems:
1. The first show outputs column name `min("B") OVER (PARTITION BY "A"  ORDER BY "B" A...`. I guess Snowflake uses all upper case.
2. the 2nd doesn't work. `AttributeError: 'numpy.ndarray' object has no attribute 'loc'`.

src/snowflake/snowpark/mock/functions.py
```diff
@@ -69,7 +69,9 @@ def wrapper(*args, **kwargs):
 
 @patch("min")
 def mock_min(column: ColumnEmulator) -> ColumnEmulator:
```
sfc-gh-yixie, line 71: What window functions will we provide out of the box?

sfc-gh-stan: I will refactor the current implementations (hardcoded `LEAD`, `LAG`, etc) into `functions.py` in a new PR.

src/snowflake/snowpark/mock/analyzer.py
```diff
@@ -285,7 +308,7 @@ def analyze(
                 self.session._conn._telemetry_client.send_function_usage_telemetry(
                     expr.api_call_source, TelemetryField.FUNC_CAT_USAGE.value
                 )
-            func_name = expr.name.upper() if parse_local_name else expr.name
```
sfc-gh-stan, line 288: This always converts function names to uppercase. cc: @sfc-gh-yixie

src/snowflake/snowpark/mock/snowflake_data_type.py
```diff
@@ -270,6 +270,8 @@ def add_date_and_number(
 
 
 class ColumnEmulator(pd.Series):
+    _metadata = ["sf_type"]
```
sfc-gh-stan, line 273: Adding this to preserve `sf_type` after index operations like `reset_index` or `sort_index`.

tests/integ/scala/test_window_frame_suite.py
```diff
@@ -251,20 +265,30 @@ def test_range_between_should_accept_at_most_one_order_by_expression_when_unboun
         df.select(
             min_("key").over(window.range_between(Window.unboundedPreceding, 1))
         ).collect()
-    assert "Cumulative window frame unsupported for function MIN" in str(ex_info)
+        if not local_testing_mode:
+            assert "Cumulative window frame unsupported for function MIN" in str(
```
sfc-gh-stan, line 269: The error message doesn't make sense. This error is caused by `window.range_between(Window.unboundedPreceding, 1)`, you cannot specify range between for non-cumulative window frames. As an evidence, changing the expression to `window.range_between(Window.unboundedPreceding, 0)` runs regardless of the number of order expressions.

tests/integ/scala/test_window_frame_suite.py
```diff
@@ -283,19 +307,28 @@ def test_range_between_should_accept_numeric_values_only_when_bounded(session):
         df.select(
             min_("value").over(window.range_between(Window.unboundedPreceding, 1))
         ).collect()
-    assert "Cumulative window frame unsupported for function MIN" in str(ex_info)
+        if not local_testing_mode:
```
sfc-gh-stan, line 310: Same issue as above, the error message is confusing and the actual failure is not caused by numeric values.

sfc-gh-stan: We should rewrite these two tests for both live connection and local testing.

sfc-gh-stan:
> I tried this
> 
> ```
> from snowflake.snowpark import Session, Window
> from snowflake.snowpark.functions import min as min_
> from snowflake.snowpark.mock.connection import MockServerConnection
> 
> session = Session(MockServerConnection())
> 
> df = session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"])
> df.select(min_("b").over(Window.partition_by("a").order_by("b"))).show()
> df.select(min_("b").over(Window.partition_by("a").order_by("b").range_between(0, 1))).show()
> ```
> 
> Two problems:
> 
> 1. The first show outputs column name `min("B") OVER (PARTITION BY "A"  ORDER BY "B" A...`. I guess Snowflake uses all upper case.
> 2. the 2nd doesn't work. `AttributeError: 'numpy.ndarray' object has no attribute 'loc'`.

Fixed.

src/snowflake/snowpark/mock/plan.py
```diff
@@ -982,7 +1003,268 @@ def calculate_expression(
...
+    if isinstance(exp, WindowExpression):
+        window_function = exp.window_function
+        window_spec = exp.window_spec
+
+        # Process order by clause
+        if window_spec.order_spec:
+            res = input_data.sort_values(
+                by=[
+                    exp.child.name for exp in window_spec.order_spec
+                ],  # TODO: consider when the expr is something like col1+2
```
sfc-gh-stan, line 1015: This needs to be addressed.

tests/integ/scala/test_window_spec_suite.py
```diff
@@ -434,17 +441,27 @@ def test_skewness_and_kurtosis_functions_in_window(session):
...
     w1 = Window.partition_by("sno")
     w2 = Window.partition_by("sno", "pno")
+
     select = df.select(
         "sno", "pno", "qty", sum_("qty").over(w2).alias("sum_qty_2")
     ).select(
         "sno", "pno", "qty", col("sum_qty_2"), sum_("qty").over(w1).alias("sum_qty_1")
     )
+
+    # S1 P1 100
```
sfc-gh-stan, line 458: remove these

src/snowflake/snowpark/mock/plan.py
```diff
@@ -167,6 +185,90 @@ def __init__(
         self.options = options
 
 
+def handle_order_by_clause(
```
sfc-gh-aling, line 188: nit: can we have a short desc on what this method is doing, input, output?

src/snowflake/snowpark/mock/plan.py
```diff
@@ -167,6 +185,90 @@ def __init__(
...
+def handle_range_frame_indexing(
+    window_spec,
+    res_index,
+    res,
+    analyzer,
+    expr_to_alias,
+    unbounded_preceding,
+    unbounded_following,
+):
+    if window_spec.order_spec:
```
sfc-gh-aling, line 223: same, can we have a short desc on what this function is doing.
from the code looks like it's trying to find window accroding to the win spec, as it's returning a list of frames.

src/snowflake/snowpark/mock/plan.py
```diff
@@ -167,6 +185,90 @@ def __init__(
...
+            list(zip(sort_columns_array, sort_orders_array, null_first_last_array))
+        )
+        for column, ascending, null_first in kk:
+            comparator = partial(custom_comparator, ascending, null_first)
+            result_df = result_df.sort_values(by=column, key=comparator)
+    return result_df
+
+
+def handle_range_frame_indexing(
+    window_spec,
```
sfc-gh-aling, line 215: what does `window_spec` contain?
I see only `window_spec.order_spec` is used in this func, does that mean to get windows we only need `order_spec`?

sfc-gh-stan: `window_spec` is a `WindowSpec` instance. In this particular function only `order_spec` is being used because`range_between` window frames use the expressions in the order by clause to select the group of rows in the window.

src/snowflake/snowpark/mock/plan.py
```diff
@@ -167,6 +189,87 @@ def __init__(
...
+    analyzer: "MockAnalyzer",
+    expr_to_alias: Optional[Dict[str, str]],
+    unbounded_preceding: bool,
+    unbounded_following: bool,
+) -> "pd.api.typing.RollingGroupby":
+    """Return a list of range between window frames based on the dataframe paritions `res` and the ORDER BY clause `order_spec`."""
+    if order_spec:
+        windows = []
+        for current_row, win in zip(res_index, res.rolling(EntireWindowIndexer())):
+            if order_spec:
```
sfc-gh-aling, line 229: I think this if check is unnecessary here given we checked it in line 226

src/snowflake/snowpark/mock/plan.py
```diff
@@ -982,7 +1072,268 @@ def calculate_expression(
...
+                        )
+                    else:
+                        res_cols.append(target_expr)
+
+            res_col = ColumnEmulator(
+                data=res_cols, dtype=object
+            )  # dtype=object prevents implicit converting None to Nan
+            res_col.index = res_index
+            return res_col.sort_index()
+        elif isinstance(window_function, FirstValue):
```
sfc-gh-aling, line 1267: nit: it feels that FirstValue and LastValue calculation are sharing lots of logic
except for the `w.iloc[0]` vs `w.iloc[len(w) - 1]` and `for cur_idx in range(len(w)):` vs `for cur_idx in range(len(w) - 1, -1, -1):`.

src/snowflake/snowpark/mock/snowflake_data_type.py
sfc-gh-aling: I don't see this is used anyhere?

src/snowflake/snowpark/mock/plan.py
```diff
@@ -167,6 +189,87 @@ def __init__(
...
+    """Return a list of range between window frames based on the dataframe paritions `res` and the ORDER BY clause `order_spec`."""
+    if order_spec:
+        windows = []
+        for current_row, win in zip(res_index, res.rolling(EntireWindowIndexer())):
+            if order_spec:
+                _win = handle_order_by_clause(order_spec, win, analyzer, expr_to_alias)
+                row_idx = list(_win.index).index(current_row)
+                start_idx = 0 if unbounded_preceding else row_idx
+                end_idx = len(_win) - 1 if unbounded_following else row_idx
+                while start_idx > 0:
```
sfc-gh-aling, line 234: should we add an upper bound for start_idx?

src/snowflake/snowpark/mock/plan.py
```diff
@@ -167,6 +189,87 @@ def __init__(
...
+                        calculate_expression(
+                            exp.child, _win.iloc[start_idx - 1], analyzer, expr_to_alias
+                        )
+                        for exp in order_spec
+                    )
+                    if not expr1 == expr2:
+                        break
+                    start_idx -= 1
+
+                while end_idx < len(_win) - 1:
```
sfc-gh-aling, line 251: lower bound for end_idx?

src/snowflake/snowpark/mock/plan.py
```diff
@@ -167,6 +189,87 @@ def __init__(
...
+                            exp.child, _win.iloc[start_idx], analyzer, expr_to_alias
+                        )
+                        for exp in order_spec
+                    )
+                    expr2 = list(
+                        calculate_expression(
+                            exp.child, _win.iloc[start_idx - 1], analyzer, expr_to_alias
+                        )
+                        for exp in order_spec
+                    )
```
sfc-gh-aling, line 246: I feel the logic of finding `start_idx` and `end_idx` can also be merged.


def search(search_pos, direction):
    while search_pos in a range:
        expr1 = func(search_pos)
        expr2 = func(search_pos + direction)
        if expr1 != expr2:
            break
    search_pos += direction

start = search(start_idx, -1)
end = search(end_idx, 1)

src/snowflake/snowpark/mock/snowflake_data_type.py
sfc-gh-stan: This is needed by pandas internally to preserve added property in subclasses: https://pandas.pydata.org/docs/development/extending.html#define-original-properties

src/snowflake/snowpark/mock/plan.py
sfc-gh-stan: `start_idx` is initiated to be either 0 or `row_idx` (guaranteed to exist), then it only decrements, so I don't think we have to?

src/snowflake/snowpark/mock/plan.py
sfc-gh-stan: This is very true. I am planning to refactor this part in a separate PR so all window function definitions can be found in `mock/functions.py`.

src/snowflake/snowpark/mock/plan.py
sfc-gh-stan: ditto.

src/snowflake/snowpark/mock/snowflake_data_type.py
sfc-gh-aling: shall we add a comment here?

sfc-gh-aling (approved):

src/snowflake/snowpark/mock/plan.py
sfc-gh-aling: let's create a JIRA or log it in an existing jira so that we don't forget?

PR #1053, opened 2023-09-20
SNOW-919476: Do not ignore stage location for using imports when is_permanent is False

sfc-gh-aalam:

   Fixes #SNOW-919476

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -851,11 +851,14 @@ def resolve_imports_and_packages(
                 )
             udf_level_imports[resolved_import_tuple[0]] = resolved_import_tuple[1:]
         all_urls = session._resolve_imports(
-            upload_stage, udf_level_imports, statement_params=statement_params
+            import_stage,
+            upload_stage,
+            udf_level_imports,
+            statement_params=statement_params,
         )
     elif imports is None:
         all_urls = session._resolve_imports(
-            import_stage, statement_params=statement_params
+            import_stage, upload_stage, statement_params=statement_params
         )
     else:
         all_urls = []
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -349,11 +349,6 @@ def check_register_args(
             raise ValueError(
                 f"stage_location must be specified for permanent {get_error_message_abbr(object_type)}"
             )
-    else:
-        if stage_location:
-            logger.warn(
-                "is_permanent is False therefore stage_location will be ignored"
-            )
 
     if parallel < 1 or parallel > 99:
         raise ValueError(
@@ -813,11 +808,10 @@ def resolve_imports_and_packages(
     statement_params: Optional[Dict[str, str]] = None,
     source_code_display: bool = False,
     skip_upload_on_content_match: bool = False,
-    is_permanent: bool = False,
 ) -> Tuple[str, str, str, str, str, bool]:
     upload_stage = (
         unwrap_stage_location_single_quote(stage_location)
-        if stage_location and is_permanent
+        if stage_location
         else session.get_session_stage()
     )
 
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -810,13 +810,15 @@ def resolve_imports_and_packages(
     skip_upload_on_content_match: bool = False,
     is_permanent: bool = False,
 ) -> Tuple[str, str, str, str, str, bool]:
-    import_stage = (
+    import_only_stage = (
         unwrap_stage_location_single_quote(stage_location)
         if stage_location
         else session.get_session_stage()
     )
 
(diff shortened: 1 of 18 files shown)


sfc-gh-aalam:
PR for fixing the merge-gate failures is here: https://github.com/snowflakedb/snowpark-python/pull/1054

sfc-gh-sfan (commented):
I'm not sure if this is the desired behavior (e.g. upload the udf zip file into a permanent stage location for a temp UDF). Can we instead separate the concern of upload and imports? For example, we could have an upload stage and an import stage.

sfc-gh-aalam:
> I'm not sure if this is the desired behavior (e.g. upload the udf zip file into a permanent stage location for a temp UDF). Can we instead separate the concern of upload and imports? For example, we could have an upload stage and an import stage.

This is a really good suggestion. I have updated the code to use this logic because I got confused about the context and started looking at stage location only as an upload location.

This is the desired behavior from temp sprocs:
- temp sprocs should not upload packages to permanent stages
- temp sprocs should be able to import pacakges from permanent stages
- temp sprocs should be able to use custom packages

src/snowflake/snowpark/session.py
```diff
@@ -695,9 +696,11 @@ def _resolve_imports(
         """Resolve the imports and upload local files (if any) to the stage."""
         resolved_stage_files = []
         stage_file_list = self._list_files_in_stage(
-            stage_location, statement_params=statement_params
+            import_stage, statement_params=statement_params
         )
-        normalized_stage_location = unwrap_stage_location_single_quote(stage_location)
+        # probably shouldn't do it. It is already done in resolve_imports_and_pacakges
+        normalized_import_location = unwrap_stage_location_single_quote(import_stage)
```
sfc-gh-sfan, line 702: Can we verify if this could lead to bugs (e.g. over-unwrapping)?

sfc-gh-sfan:
> This is the desired behavior from temp sprocs:
> 
> * temp sprocs should not upload packages to permanent stages
> * temp sprocs should be able to import pacakges from permanent stages
> * temp sprocs should be able to use custom packages using imports
> * temp sprocs should be able to import files from permanent stage and also use custom packages from imports

Let's make sure this is documented and tested :)

sfc-gh-aalam:
> > This is the desired behavior from temp sprocs:
> > 
> > * temp sprocs should not upload packages to permanent stages
> > * temp sprocs should be able to import pacakges from permanent stages
> > * temp sprocs should be able to use custom packages using imports
> > * temp sprocs should be able to import files from permanent stage and also use custom packages from imports
> 
> Let's make sure this is documented and tested :)

Yep. I'm working on the tests now.

src/snowflake/snowpark/session.py
sfc-gh-aalam: there are some tests in `test_utils_suite.py/test_normalize_stage_location` which test for this.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -821,6 +816,12 @@ def resolve_imports_and_packages(
         else session.get_session_stage()
     )
 
+    import_stage = (
```
sfc-gh-stan, line 819: Let's define import_stage first then `upload_stage = import_stage if is_permanent else session_stage`.

sfc-gh-stan: I wonder if there's a better name for `upload_stage`, because technically we need to **upload** files to `import_stage` too. How about something like `sproc_stage` vs `import_stage`? 
In addition, could you please add a one line documentation for these args to clarify?

sfc-gh-aalam: I agree with your point but the stage is used for udfs as well. How about `import_only_stage` and `upload_and_import_stage`?

sfc-gh-sfan: I don't think we upload files to import stage, but it is the other way around (we import the upload stage). And as Afroz mentioned, the import/upload difference is application to UDF families as well. I cannot come up with a better name than what Afroz suggested.

src/snowflake/snowpark/session.py
sfc-gh-aalam: Added more tests. Overwrapping does not lead to bugs

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aalam: done

sfc-gh-sfan (approved):

src/snowflake/snowpark/session.py
```diff
@@ -695,9 +696,15 @@ def _resolve_imports(
         """Resolve the imports and upload local files (if any) to the stage."""
         resolved_stage_files = []
         stage_file_list = self._list_files_in_stage(
-            stage_location, statement_params=statement_params
+            import_only_stage, statement_params=statement_params
+        )
+        # probably shouldn't do it. It is already done in resolve_imports_and_pacakges
```
sfc-gh-sfan, line 701: Let's remove this comment if we already have tests :p

Issue #1055, opened 2023-09-24
Type checking


alexmojaki:
The code has a lot of type hints but I see nothing enforcing that types are actually checked. I find this surprising as this is a very complex codebase that would definitely benefit from rigorous type checking.

PR #1056, opened 2023-09-24
Typing

alexmojaki:
This is an attempt to make some progress on https://github.com/snowflakedb/snowpark-python/issues/1055. Before this PR, pyright reports 625 errors, afterwards it reports 468. In particular I resolved all the type errors in `src/snowflake/snowpark/_internal/analyzer/`.

### DIFF
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -51,7 +51,7 @@ jobs:
         run: python -m pip install -U setuptools pip
       - name: Install tox
         run: python -m pip install tox
-      - name: Run fix_lint
+      - name: Run pyright on Selected Files
         run: python -m tox -e pyright
 
   build:
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -36,6 +36,24 @@ jobs:
       - name: Run fix_lint
         run: python -m tox -e fix_lint
 
+  type_checking:
+    name: Type Checking
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Display Python version
+        run: python -c "import sys; import os; print(\"\n\".join(os.environ[\"PATH\"].split(os.pathsep))); print(sys.version); print(sys.executable);"
+      - name: Upgrade setuptools and pip
+        run: python -m pip install -U setuptools pip
+      - name: Install tox
+        run: python -m pip install tox
+      - name: Run fix_lint
+        run: python -m tox -e pyright
+
   build:
     needs: lint
     name: Build Wheel File
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -2,9 +2,9 @@
 #
 # Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
 #
-
+import uuid
 from collections import Counter, defaultdict
-from typing import TYPE_CHECKING, DefaultDict, Dict, Union
+from typing import TYPE_CHECKING, DefaultDict, Dict, Optional, Union
 
 import snowflake.snowpark
 from snowflake.snowpark._internal.analyzer.analyzer_utils import (
@@ -156,7 +156,7 @@ class Analyzer:
         self.plan_builder = SnowflakePlanBuilder(self.session)
         self.generated_alias_maps = {}
         self.subquery_plans = []
-        self.alias_maps_to_use = None
+        self.alias_maps_to_use: Optional[Dict[uuid.UUID, str]] = None
 
     def analyze(
         self,
@@ -334,6 +334,7 @@ class Analyzer:
             return sql
 
         if isinstance(expr, Attribute):
+            assert self.alias_maps_to_use is not None
             name = self.alias_maps_to_use.get(expr.expr_id, expr.name)
             return quote_name(name)
 
@@ -531,7 +532,7 @@ class Analyzer:
     def table_function_expression_extractor(
         self,
         expr: TableFunctionExpression,
-        df_aliased_col_name_to_real_col_name: Dict[str, str],
+        df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
         parse_local_name=False,
     ) -> str:
         if isinstance(expr, FlattenFunction):
@@ -580,13 +581,14 @@ class Analyzer:
     def unary_expression_extractor(
         self,
         expr: UnaryExpression,
-        df_aliased_col_name_to_real_col_name: Dict[str, str],
+        df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
         parse_local_name=False,
     ) -> str:
         if isinstance(expr, Alias):
             quoted_name = quote_name(expr.name)
             if isinstance(expr.child, Attribute):
                 self.generated_alias_maps[expr.child.expr_id] = quoted_name
+                assert self.alias_maps_to_use is not None
                 for k, v in self.alias_maps_to_use.items():
                     if v == expr.child.name:
                         self.generated_alias_maps[k] = quoted_name
@@ -681,7 +683,7 @@ class Analyzer:
     def to_sql_avoid_offset(
         self,
         expr: Expression,
-        df_aliased_col_name_to_real_col_name: Dict[str, str],
+        df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
         parse_local_name: bool = False,
     ) -> str:
         # if expression is a numeric literal, return the number without casting,
@@ -752,7 +754,7 @@ class Analyzer:
         self,
         logical_plan: LogicalPlan,
         resolved_children: Dict[LogicalPlan, SnowflakePlan],
-        df_aliased_col_name_to_real_col_name: Dict[str, str],
+        df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
     ) -> SnowflakePlan:
         if isinstance(logical_plan, SnowflakePlan):
             return logical_plan
@@ -991,9 +993,10 @@ class Analyzer:
                 if logical_plan.format_type_options
                 else {}
             )
-            format_name = logical_plan.cur_options.get("FORMAT_NAME")
+            format_name = (logical_plan.cur_options or {}).get("FORMAT_NAME")
             if format_name is not None:
                 format_type_options["FORMAT_NAME"] = format_name
+            assert logical_plan.file_format is not None
             return self.plan_builder.copy_into_table(
                 path=logical_plan.file_path,
                 table_name=logical_plan.table_name,
@@ -1005,7 +1008,8 @@ class Analyzer:
                 validation_mode=logical_plan.validation_mode,
                 column_names=logical_plan.column_names,
                 transformations=[
-                    self.analyze(x, df_aliased_col_name_to_real_col_name)
+                    # TODO investigate, this looks like a genuine type mismatch
+                    self.analyze(x, df_aliased_col_name_to_real_col_name)  # type: ignore
                     for x in logical_plan.transformations
                 ]
                 if logical_plan.transformations
@@ -1044,7 +1048,9 @@ class Analyzer:
                 )
                 if logical_plan.condition
                 else None,
-                resolved_children.get(logical_plan.source_data, None),
+                None
+                if logical_plan.source_data is None
+                else resolved_children.get(logical_plan.source_data, None),
                 logical_plan,
             )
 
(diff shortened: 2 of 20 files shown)


alexmojaki:

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -1002,7 +1008,8 @@ def do_resolve_with_resolved_children(
                 validation_mode=logical_plan.validation_mode,
                 column_names=logical_plan.column_names,
                 transformations=[
-                    self.analyze(x, df_aliased_col_name_to_real_col_name)
+                    # TODO investigate, this looks like a genuine type mismatch
```
sfc-gh-sfan, line 1011: @sfc-gh-stan Looks like you introduced `df_aliased_col_name_to_real_col_name`. Could you take a look at the type mismatch here?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -480,7 +480,10 @@ def range_statement(start: int, end: int, step: int, column_name: str) -> str:
 
 
 def schema_query_for_values_statement(output: List[Attribute]) -> str:
-    cells = [schema_expression(attr.datatype, attr.nullable) for attr in output]
+    cells = [
+        schema_expression(cast(DataType, attr.datatype), attr.nullable)
```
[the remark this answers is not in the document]
alexmojaki: `Attribute.datatype` has type `Optional[DataType]`, which isn't accepted here. Usually in this PR I've dealt with this by writing something like `assert attr.datatype is not None` which satisfies the type checker. I use `cast` here so that the list comprehension can stay.

However, I've taken a closer look, and the only reason for the type being optional is so that it can default to None, which is only useful in one place where the data type is not provided: https://github.com/snowflakedb/snowpark-python/blob/2243bb617b72982ded33e6326cc9bae76ec47205/src/snowflake/snowpark/_internal/analyzer/select_statement.py#L1096-L1099

So maybe a better solution is to make that argument required and provide a dummy datatype in that place.

src/snowflake/snowpark/_internal/analyzer/binary_expression.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, left: Expression, right: Expression) -> None:
     def __str__(self):
         return f"{self.left} {self.sql_operator} {self.right}"
 
-    def dependent_column_names(self) -> Optional[Set[str]]:
+    def dependent_column_names(self) -> Optional[AbstractSet[str]]:
```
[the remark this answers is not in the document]
alexmojaki: Before this PR, there are 7 errors along these lines: `Expression of type "frozenset[Unknown]" cannot be assigned to return type "Set[str] | None"`, i.e. a `frozenset` doesn't count as a `Set`. Other solutions could be to use `Union[Set[str], FrozenSet[str]]` (could make a type alias) or to make `COLUMN_DEPENDENCY_DOLLAR` and `COLUMN_DEPENDENCY_EMPTY` normal sets and hope they don't get mutated (probably not a good idea).

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -93,13 +94,14 @@ def __init__(
...
-            Set[str]
+            AbstractSet[str]
         ] = COLUMN_DEPENDENCY_ALL,  # columns that this column has a dependency on.
         depend_on_same_level: bool = False,  # Whether this column has dependency on one or more columns of the same level instead of the subquery.
         referenced_by_same_level_columns: Optional[
-            Set[str]
+            AbstractSet[str]
         ] = COLUMN_DEPENDENCY_EMPTY,  # Other same-level columns that use this column.
-        state_dict: Optional["ColumnStateDict"] = None,  # has states of all columns.
+        *,
```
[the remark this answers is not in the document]
alexmojaki: There's a similar story for `format_type_options` in `SnowflakePlanBuilder.copy_into_table`. In that case I opted for a dirtier fix and a TODO since I was starting to question things, which I'm glad about now lol

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -459,16 +463,17 @@ def column_states(self) -> ColumnStateDict:
...
         return self._column_states
 
     @column_states.setter
-    def column_states(self, value: ColumnStateDict):
+    def column_states(self, value: Optional[ColumnStateDict]):
         """A dictionary that contains the column states of a query.
         Refer to class ColumnStateDict.
         """
-        self._column_states = copy(value)
+        value = self._column_states = copy(value)
```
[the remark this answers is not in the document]
alexmojaki: I've pushed a better fix now which makes `column_states` non-optional.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1026,7 +1041,7 @@ def populate_column_dependency(
     elif dependent_column_names == COLUMN_DEPENDENCY_ALL:
         column_states[quoted_c_name].depend_on_same_level = True
         column_states.columns_referencing_all_columns.add(quoted_c_name)
-    else:
```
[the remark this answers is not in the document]
alexmojaki: Added, although it shouldn't be reachable.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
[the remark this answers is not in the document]
sfc-gh-sfan: Hey we don't need to avoid a behavior change here, I'm just curious on why do we need such a change and you explained it pretty well :)

sfc-gh-sfan:
Test pr in https://github.com/snowflakedb/snowpark-python/pull/1061

src/snowflake/snowpark/_internal/analyzer/select_statement.py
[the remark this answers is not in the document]
alexmojaki: lol no it's extremely reachable since the set can just be empty. Removed the else clause. The goal here is just to concisely tell the type checker that `dependent_column_names` isn't None.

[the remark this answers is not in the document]
sfc-gh-sfan: Can we do `elif dependent_column_names is not None` then add a final unreachable `else`? I'm a bit paranoid to not have a "default failure" when there is a lot of if/elif involved.

sfc-gh-sfan:
```
tests/integ/scala/test_snowflake_plan_suite.py:78: in test_multiple_queries
    plan2 = SnowflakePlan(
E   TypeError: SnowflakePlan.__init__() missing 1 required keyword-only argument: 'session'
```

.github/workflows/precommit.yml
```diff
@@ -36,6 +36,24 @@ jobs:
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Display Python version
+        run: python -c "import sys; import os; print(\"\n\".join(os.environ[\"PATH\"].split(os.pathsep))); print(sys.version); print(sys.executable);"
+      - name: Upgrade setuptools and pip
+        run: python -m pip install -U setuptools pip
+      - name: Install tox
+        run: python -m pip install tox
+      - name: Run fix_lint
```
alexmojaki, line 54:
```suggestion
      - name: Run pyright
```

Happy to see you chose to add this to CI :blush:

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-stan: The mismatch is from `CopyIntoTableNode.transformations`. Fixing...

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -988,9 +993,10 @@ def do_resolve_with_resolved_children(
                 if logical_plan.format_type_options
                 else {}
             )
-            format_name = logical_plan.cur_options.get("FORMAT_NAME")
+            format_name = (logical_plan.cur_options or {}).get("FORMAT_NAME")
```
sfc-gh-stan, line 996: Looking at the only reference to `CopyIntoTable` at https://github.com/snowflakedb/snowpark-python/blob/697294d8ee89396deab4309dea384b1eadb75d6b/src/snowflake/snowpark/dataframe.py#L2819, `cur_options` will never be None.

alexmojaki: Please can you take over the PR where possible?

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -1041,7 +1047,9 @@ def do_resolve_with_resolved_children(
                 )
                 if logical_plan.condition
                 else None,
-                resolved_children.get(logical_plan.source_data, None),
```
sfc-gh-stan, line 1044: This is not introduced in this PR, but I'm curious why we need to do `resolved_children.get(logical_plan.source_data, None)`. 

From the references to `TableUpdate`/`TableDelete`/`TableMerge`, `source` or `source_data` will already be resolved. In addition, `logical_plan.source_data` is guaranteed to be in `resolved_children`. https://github.com/snowflakedb/snowpark-python/blob/697294d8ee89396deab4309dea384b1eadb75d6b/src/snowflake/snowpark/_internal/analyzer/table_merge_expression.py#L54 

I feel like just `logical_plan.source_data` would suffice.

sfc-gh-stan: cc: @sfc-gh-jdu

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -1053,14 +1061,16 @@ def do_resolve_with_resolved_children(
...
+                None
```
sfc-gh-stan, line 1064: Ditto.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -1053,14 +1061,16 @@ def do_resolve_with_resolved_children(
...
+                if logical_plan.source_data is None
+                else resolved_children.get(logical_plan.source_data, None),
                 logical_plan,
             )
 
         if isinstance(logical_plan, TableMerge):
             return self.plan_builder.merge(
                 logical_plan.table_name,
-                resolved_children.get(logical_plan.source),
+                resolved_children[logical_plan.source],
```
sfc-gh-stan, line 1073: Ditto.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1025,8 +1025,8 @@ def copy_into_table(
     file_path: str,
     file_format_type: str,
     format_type_options: Dict[str, Any],
-    copy_options: Dict[str, Any],
-    pattern: str,
+    copy_options: Optional[Dict[str, Any]],
```
sfc-gh-stan, line 1028: `copy_options` is technically never `None` given https://github.com/snowflakedb/snowpark-python/blob/697294d8ee89396deab4309dea384b1eadb75d6b/src/snowflake/snowpark/dataframe.py#L2800

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-stan: Will do, thank you for making this happen, really appreciate your work here! 🎉

sfc-gh-stan:
Sorry all, I messed up a rebase and had to create a new test pr: #1095

src/snowflake/snowpark/_internal/analyzer/schema_utils.py
```diff
@@ -90,7 +93,10 @@ def convert_result_meta_to_attribute(meta: List[ResultMetadata]) -> List[Attribu
             Attribute(
                 quoted_name,
                 convert_sf_to_sp_type(
-                    FIELD_ID_TO_NAME[type_value], precision, scale, internal_size
+                    FIELD_ID_TO_NAME[type_value],
+                    precision or 0,
```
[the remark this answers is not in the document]
sfc-gh-sfan: What happens today if it is `None`?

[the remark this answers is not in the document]
sfc-gh-stan: This is only None when the result is not `Decimal` type, where the parameter would not be used in `convert_sf_to_sp_type`. So technically the most "rigorous" way of writing this is checking the column type in `convert_result_meta_to_attribute` and define `convert_sf_to_sp_type` separately.

sfc-gh-mkeller (approved):
:ship: :ship:

sfc-gh-stan:
After discussing with @sfc-gh-mkeller , we have decided to use `mypy` as the source of truth for type checking. We will refer to the mypy configuration from the upstream https://github.com/snowflakedb/snowflake-connector-python/commit/eee76d15c6d957f1683440f2152eeabd518ce814#diff-6f2d4ba9ca9a357d31014946667b7bed1bfdbc6d2530afc77778fa0a36bee457. Ideally speaking this should not conflict with pyright checks, so I will keep the `pyright` merge gate, but `mypy` takes precedence if such conflict arise. I will mark the pyright merge gate as optional when we add `mypy` in the CI.

Issue #824, opened 2023-05-05
SNOW-809908: Snowpark Session _table_exists doesn't work as expected with fully qualifed table names


brian-kalinowski-sonos:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using?

 `Python 3.8.13`

2. What operating system and processor architecture are you using?

  `macOS-10.16-x86_64-i386-64bit`

3. What are the component versions in the environment (`pip freeze`)?
```
aiobotocore==2.3.4
aiohttp==3.8.1
aioitertools==0.10.0
aiosignal==1.2.0
altair==4.2.0
appnope==0.1.3
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
asn1crypto==1.5.1
asttokens==2.0.5
async-timeout==4.0.2
attrs==21.4.0
azure-common==1.1.28
azure-core==1.24.2
azure-storage-blob==12.12.0
backcall==0.2.0
backports.zoneinfo==0.2.1
beautifulsoup4==4.11.1
bleach==5.0.0
blinker==1.4
boto3==1.15.18
botocore==1.18.18
cachetools==5.2.0
certifi==2020.12.5
cffi==1.14.6
chardet==3.0.4
charset-normalizer==2.0.12
click==8.1.3
cloudpickle==2.0.0
commonmark==0.9.1
cryptography==3.2.1
debugpy==1.6.0
decorator==5.1.1
defusedxml==0.7.1
Deprecated==1.2.13
entrypoints==0.4
executing==0.8.3
fastjsonschema==2.15.3
filelock==3.8.0
frozenlist==1.3.0
fsspec==2022.5.0
gitdb==4.0.9
GitPython==3.1.27
greenlet==1.1.2
idna==2.10
importlib-metadata==4.11.4
importlib-resources==5.8.0
iniconfig==1.1.1
ipykernel==6.15.0
ipython==8.4.0
ipython-genutils==0.2.0
ipywidgets==7.7.0
isodate==0.6.1
jedi==0.18.1
Jinja2==3.1.2
jmespath==0.10.0
jsonschema==4.6.0
jupyter-client==7.3.4
jupyter-core==4.10.0
jupyterlab-pygments==0.2.2
jupyterlab-widgets==1.1.0
MarkupSafe==2.1.1
matplotlib-inline==0.1.3
mistune==0.8.4
msrest==0.7.1
multidict==6.0.2
nbclient==0.6.4
[...]

PR #1057, opened 2023-09-24
Add ParsedTableName class

alexmojaki:
I was looking into https://github.com/snowflakedb/snowpark-python/issues/824 before I realised it was fixed, and I noticed a lot of ad-hoc boilerplate for handling parameters representing table names of type `Union[str, Iterabe[str]]`. This is a proposal for a more uniform and elegant way to do that. `ParsedTableName` also acts as a nominal type, asserting that a name has already been validated, and making it clear that it's not just any old string.

This was just meant to be testing an idea but I got carried away, sorry for the huge diff! It occurred to me afterwards that I could make `ParsedTableName` a subclass of `str` to make this less intrusive. But first I want to hear if you think this whole thing is a reasonable idea at all.


sfc-gh-sfan (commented):
This seems reasonable to me. I'll have @sfc-gh-yixie to take a look as well.

PR #1060, opened 2023-09-25
[Local Testing] SNOW-904992 Support CaseWhen expressions

sfc-gh-stan:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/mock/functions.py
+++ b/src/snowflake/snowpark/mock/functions.py
@@ -391,3 +391,18 @@ def mock_to_timestamp(column: ColumnEmulator, fmt: Union[ColumnEmulator, str] = 
         sf_type=ColumnType(TimestampType(), column.sf_type.nullable),
         dtype=object,
     )
+
+
+@patch("iff")
+def mock_iff(condition: ColumnEmulator, expr1: ColumnEmulator, expr2: ColumnEmulator):
+    assert isinstance(condition.sf_type.datatype, BooleanType)
+    condition = condition.array
+    res = ColumnEmulator(data=[None] * len(condition), dtype=object)
+    if not all(condition) and expr1.sf_type != expr2.sf_type:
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/mock/functions.py
```diff
@@ -391,3 +391,18 @@ def mock_to_timestamp(column: ColumnEmulator, fmt: Union[ColumnEmulator, str] =
...
         dtype=object,
     )
+
+
+@patch("iff")
+def mock_iff(condition: ColumnEmulator, expr1: ColumnEmulator, expr2: ColumnEmulator):
+    assert isinstance(condition.sf_type.datatype, BooleanType)
+    condition = condition.array
+    res = ColumnEmulator(data=[None] * len(condition), dtype=object)
+    if not all(condition) and expr1.sf_type != expr2.sf_type:
```
sfc-gh-yixie, line 401: Snowflake doesn't need the two expression to have exactly the same datatype. I guess there is a rule to do type coercion.
For instance `select system$typeof(iff(True, '2', 1))` returns a number type. '2' is cast to 2.
If you choose not to deal with type coercion at this time, better add a todo to followup in the future.

Refer to https://docs.snowflake.com/en/sql-reference/data-type-conversion

I used to put the type conversion/coercion into a Python dict but later commented out.
https://github.com/snowflakedb/snowpark-python/blob/b9946f3e914bba3e4255d3acbd26090e0882435d/src/snowflake/snowpark/mock/snowflake_data_type.py#L62

sfc-gh-stan: I see, thanks for pointing this out. I will add a TODO for supporting type coercion, we also need to support Column.cast so it should be similar.

sfc-gh-stan: Created https://snowflakecomputing.atlassian.net/browse/SNOW-929218

PR #1069, opened 2023-10-02
[Local Testing] SNOW-904260 support dataframe na functions

sfc-gh-stan:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/dataframe_na_functions.py
+++ b/src/snowflake/snowpark/dataframe_na_functions.py
@@ -4,6 +4,7 @@
 #
 
 import copy
+import math
 from logging import getLogger
 from typing import Dict, Optional, Union
 
@@ -153,13 +154,6 @@ class DataFrameNaFunctions:
         # iff(float_col = 'NaN' or float_col is null, 0, 1)
         # iff(non_float_col is null, 0, 1) >= thresh
 
-        from snowflake.snowpark.mock.connection import MockServerConnection
-
-        if isinstance(self._df._session._conn, MockServerConnection):
-            raise NotImplementedError(
-                "[Local Testing] DataFrame NA functions are currently not supported."
-            )
-
         if how is not None and how not in ["any", "all"]:
             raise ValueError(f"how ('{how}') should be 'any' or 'all'")
 
@@ -206,7 +200,7 @@ class DataFrameNaFunctions:
                     df_col_type_dict[normalized_col_name], (FloatType, DoubleType)
                 ):
                     # iff(col = 'NaN' or col is null, 0, 1)
-                    is_na = iff((col == "NaN") | col.is_null(), 0, 1)
+                    is_na = iff((col == math.nan) | col.is_null(), 0, 1)
                 else:
                     # iff(col is null, 0, 1)
                     is_na = iff(col.is_null(), 0, 1)
@@ -309,13 +303,6 @@ class DataFrameNaFunctions:
         # select col, iff(float_col = 'NaN' or float_col is null, replacement, float_col)
         # iff(non_float_col is null, replacement, non_float_col) from table where
 
-        from snowflake.snowpark.mock.connection import MockServerConnection
-
-        if isinstance(self._df._session._conn, MockServerConnection):
-            raise NotImplementedError(
-                "[Local Testing] DataFrame NA functions are currently not supported."
-            )
-
         if subset is None:
             subset = self._df.columns
         elif isinstance(subset, str):
@@ -368,7 +355,7 @@ class DataFrameNaFunctions:
                     if isinstance(datatype, (FloatType, DoubleType)):
                         # iff(col = 'NaN' or col is null, value, col)
                         res_columns.append(
-                            iff((col == "NaN") | col.is_null(), value, col).as_(
+                            iff((col == math.nan) | col.is_null(), value, col).as_(
                                 col_name
                             )
                         )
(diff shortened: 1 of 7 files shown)


src/snowflake/snowpark/mock/functions.py
```diff
@@ -431,20 +429,31 @@ def mock_coalesce(*exprs):
 def mock_substring(
     base_expr: ColumnEmulator, start_expr: ColumnEmulator, length_expr: ColumnEmulator
 ):
-    return base_expr.str.slice(start=start_expr - 1, stop=start_expr - 1 + length_expr)
+    res = [
```
sfc-gh-stan, line 432: These changes are caused by setting ` keep_literal=False` below.

src/snowflake/snowpark/dataframe_na_functions.py
```diff
@@ -206,7 +200,7 @@ def drop(
                     df_col_type_dict[normalized_col_name], (FloatType, DoubleType)
                 ):
                     # iff(col = 'NaN' or col is null, 0, 1)
-                    is_na = iff((col == "NaN") | col.is_null(), 0, 1)
+                    is_na = iff((col == math.nan) | col.is_null(), 0, 1)
```
sfc-gh-aling, line 203: it feels like this change will affect the live connection or it won't?

tests/integ/test_dataframe.py
```diff
@@ -2126,33 +2122,30 @@ def test_fillna(session):
     Utils.check_answer(
         session.create_dataframe(
             [[1, 1.1], [None, None]], schema=["col1", "col2"]
-        ).fillna({"col1": 1.1, "col2": 1}),
-        [Row(1, 1.1), Row(None, 1)],
+        ).fillna({"col1": 1.1, "col2": 1.1}),
+        [Row(1, 1.1), Row(None, 1.1)],
```
sfc-gh-aling, line 2126: what makes we change to `1.1`, does `1` not work any more?

tests/integ/scala/test_dataframe_suite.py
```diff
@@ -2586,21 +2614,25 @@ def test_replace(session):
 
     # replace NaN
     Utils.check_answer(
-        TestData.null_data3(session).na.replace({float("nan"): 11}, subset=["flo"]),
+        TestData.null_data3(session, local_testing_mode).na.replace(
+            {float("nan"): 11.0}, subset=["flo"]
```
sfc-gh-aling, line 2618: why we change from using `11` to `11.0`?

tests/integ/scala/test_dataframe_suite.py
```diff
@@ -2548,14 +2573,15 @@ def test_fillna(session):
...
 
 
-def test_replace(session):
+@pytest.mark.localtest
+def test_replace(session, local_testing_mode):
     res = (
-        TestData.null_data3(session)
-        .na.replace({2: 300, 1: 200}, subset=["flo"])
+        TestData.null_data3(session, local_testing_mode)
+        .na.replace({2: 300.0, 1: 200.0}, subset=["flo"])
```
sfc-gh-aling, line 2584: does it mean `300` and `200` don't work in local test?

src/snowflake/snowpark/dataframe_na_functions.py
sfc-gh-stan: It won't, `math.nan` gets converted to `NaN::float` in the generated query.

src/snowflake/snowpark/mock/functions.py
```diff
@@ -279,7 +279,10 @@ def mock_abs(expr):
 
 
 @patch("to_decimal")
-def mock_to_decimal(e: ColumnEmulator, precision: int, scale: int):
+def mock_to_decimal(
```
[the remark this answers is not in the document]
sfc-gh-stan: This is the side effect of setting `keep_literal` to False above. We need to make a special case for functions that take in literal values I guess.

tests/integ/scala/test_dataframe_suite.py
sfc-gh-stan: Because `300` and `200` are integer values, we need coercion support to use them on a floating point column.

tests/integ/scala/test_dataframe_suite.py
sfc-gh-stan: Same reason as above.

tests/integ/test_dataframe.py
sfc-gh-stan: Ditto.

src/snowflake/snowpark/mock/functions.py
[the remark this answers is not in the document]
sfc-gh-stan: Let's use type hints to distinguish.

src/snowflake/snowpark/functions.py
```diff
@@ -2742,7 +2742,7 @@ def char(col: ColumnOrName) -> Column:
     return builtin("char")(c)
 
 
-def to_char(c: ColumnOrName, format: Optional[ColumnOrLiteralStr] = None) -> Column:
```
sfc-gh-stan, line 2745: `ColumnOrLiteralStr` is referring to `lit("abc")` or `"abc"`, which is misleading since our type hints don't distinguish Column[str] or Column[int]. This is also inconsistent with functions like `to_binary`. Since local testing uses type hints to determine whether an expression should be evaluated as a column or a literal value, we should change this to `str.

sfc-gh-aling: can users put a col expression as the `format`?
e.g., is the following allowed?
`to_char(col("data", format=col("format")))`

sfc-gh-stan: I don't think so: https://docs.snowflake.com/en/sql-reference/functions/to_char

sfc-gh-aling (approved):

PR #895, opened 2023-06-15
SNOW-817445: Fix dataframe writer save_as_table allow nulls

sfc-gh-aalam:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #SNOW-817445

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Schema provided to database is not entirely respected when save as table is called. This change makes sure we respect schema.

PR #1075, opened 2023-10-03
SNOW-911327: Use CTAS for save as table

### LINKED ISSUE #1046: SNOW-912320: The behavior of `save_as_table(mode='overwrite')` is inconsistent between v1.5.1 and v1.6.1

   `3.8.17 (default, Jul  5 2023, 20:35:26) [GCC 11.2.0]`(python worksheet in snowflake console)

   `Linux-5.4.181-99.354.amzn2.aarch64-aarch64-with-glibc2.17` (python worksheet in snowflake console)

   tested in python worksheet in snowflake console only with `snowflake-snowpark-python` dependency(anaconda)

4. What did you do?

    Run below python worksheet with `snowflake-snowpark-python==1.5.1` and `snowflake-snowpark-python==1.6.1` and see the result is inconsistent.

```python
# The Snowpark package is required for Python Worksheets. 
# You can add more packages by selecting them using the Packages control and then importing them.

import snowflake.snowpark as snowpark
from snowflake.snowpark.functions import col, lit
from snowflake.snowpark.types import LongType


def main(session: snowpark.Session) -> snowpark.DataFrame: 
    TABLE_NAME = "_TEST"
    # clean the table first
    session.sql(f"DROP TABLE IF EXISTS {TABLE_NAME}").collect()

    # create table with dataframe
    df = session.create_dataframe([1,2]).to_df(["a"])
    df.write.save_as_table(TABLE_NAME)

    # adding column 'b' as LongType() with NULL values
    # note: lit(None) infers its datatype with StringType().
    #       So, adding dummy_column with None value first
    #       and then cast it to LongType()
    tbl = session.table(TABLE_NAME).select('*')
    tbl = tbl.with_column('b_none', lit(None))
    tbl = tbl.with_column('b', col('b_none').astype(LongType()))
    tbl = tbl.drop('b_none')

    # overwrite it
    tbl.write.save_as_table(TABLE_NAME, column_order='name', mode='overwrite')

    # The output of snowflake-snowpark-python v1.5.1:
    # --------------
    # |"A"  |"B"   |
    # --------------
    # |1    |NULL  |
    # |2    |NULL  |
    # --------------
    #
    # The output of snowflake-snowpark-python v1.6.1:
    # It expects to be the same with v1.5.1
    # -------------
    # |"A"  |"B"  |
    # -------------
    # |     |     |
    # -------------
    return session.table(TABLE_NAME).select('*')
```

   consistent results between `1.5.1` and `1.6.1`

6. Can you set logging to DEBUG and collect the logs?

   I could not extract log in python worksheet in snowflake console. sorry.

sfc-gh-aalam:

   Fixes #1046, #SNOW-911327

   In PR #895 we started using create-table-insert-values pattern instead of using a CTAS pattern for saving dataframes. This is causing issues for multiple users so we are using CTAS while also respecting nullable columns. The tests added before also work for the correctness of this change.

### DIFF
--- a/tests/integ/test_dataframe.py
+++ b/tests/integ/test_dataframe.py
@@ -2306,11 +2306,10 @@ def test_table_types_in_save_as_table(session, save_mode, table_type):
         Utils.drop_table(session, table_name)
 
 
-@pytest.mark.parametrize("table_type", ["", "temp", "temporary", "transient"])
 @pytest.mark.parametrize(
     "save_mode", ["append", "overwrite", "ignore", "errorifexists"]
 )
-def test_save_as_table_respects_schema(session, save_mode, table_type):
+def test_save_as_table_respects_schema(session, save_mode):
     table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
 
     schema1 = StructType(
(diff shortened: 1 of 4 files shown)


tests/integ/test_dataframe.py
```diff
@@ -2306,11 +2306,10 @@ def test_table_types_in_save_as_table(session, save_mode, table_type):
         Utils.drop_table(session, table_name)
 
 
-@pytest.mark.parametrize("table_type", ["", "temp", "temporary", "transient"])
 @pytest.mark.parametrize(
     "save_mode", ["append", "overwrite", "ignore", "errorifexists"]
 )
-def test_save_as_table_respects_schema(session, save_mode, table_type):
```
sfc-gh-aalam, line 2313: removing excessive `table_type` parametrization for tests which don't provide any additional value in the test. I do want to listen to team's opinion on this

sfc-gh-sfan (commented):
To make sure I understand, we only revert the CTAS -> create and insert change, but not the original changes to allow nulls, right?

As for `table_type`, as long as we have one test that tests table_type, we should be okay.

sfc-gh-aalam:
> To make sure I understand, we only revert the CTAS -> create and insert change, but not the original changes to allow nulls, right?
> 
> As for `table_type`, as long as we have one test that tests table_type, we should be okay.

We only revert CTAS -> create+insert statement but keeping the NULL changes. In the first PR, I didn't know it was possible to enforce schema nulls using CTAS until yijun showed it to me. Note we haven't removed any nullable schema checks that were added earlier.

sfc-gh-sfan (approved):

PR #1101, opened 2023-10-20
SNOW-940607: Fix precision loss during to_pandas conversion

sfc-gh-aalam:

   Fixes #SNOW-940607

   When using `pandas.to_numeric` to fix https://github.com/snowflakedb/snowflake-connector-python/issues/828 caused precision issues when number of digits contained in the number was high and was interpreted as object instead of int64. We use `.astype('int64')` in a try block to try to convert the pandas column into `int64` if possible. If we hit exception, we fall back to `to_numeric`. Using `.astype('int64')` preserves precision in cases when the column can be converted into an int64 (which will most likely be the case).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.10.0 (TBD)
 
+### Bug Fixes
+
+- Fixed a bug in `DataFrame.to_pandas()` where the converting snowpark dataframes to pandas dataframes was losing precision on 19-digit numbers.
+
 ### Behavior change
 
 - Changed the behavior of `date_format`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bug Fixes
 
-- Fixed a bug in `DataFrame.to_pandas()` where the converting snowpark dataframes to pandas dataframes was losing precision on 19-digit numbers.
+- Fixed a bug in `DataFrame.to_pandas()` where converting snowpark dataframes to pandas dataframes was losing precision on integers with more than 19 digits.
 
 ### Behavior change
 
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -668,13 +668,12 @@ def _fix_pandas_df_integer(
             # pandas.to_numeric raises "RuntimeWarning: invalid value encountered in cast"
             # when it tried to downcast and loses precision. In this case, we try to convert to
             # int64 using astype() and fallback to to_numeric if we were unsuccessful.
-            pd_col_with_numeric_downcast = pd_df[pandas_col_name]
             with warnings.catch_warnings(record=True) as warning_list:
                 pd_col_with_numeric_downcast = pandas.to_numeric(
                     pd_df[pandas_col_name], downcast="integer"
                 )
 
-            if len(warning_list) == 0:
+            if not warning_list:
                 pd_df[pandas_col_name] = pd_col_with_numeric_downcast
             else:
                 try:
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -8,6 +8,7 @@ import inspect
 import os
 import sys
 import time
+import warnings
 from logging import getLogger
 from typing import IO, Any, Dict, Iterator, List, Optional, Sequence, Set, Tuple, Union
 
@@ -652,6 +653,8 @@ def _fix_pandas_df_integer(
 
     Based on the Result Metadata characteristics, this functions tries to make a best effort conversion to int64 without losing
     precision.
+
+    We need to get rid of this workaround because this causes a performance hit.
     """
     for column_metadata, pandas_dtype, pandas_col_name in zip(
         results_cursor.description, pd_df.dtypes, pd_df.columns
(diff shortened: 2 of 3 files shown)


sfc-gh-stan (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.10.0 (TBD)
 
+### Bug Fixes
+
+- Fixed a bug in `DataFrame.to_pandas()` where the converting snowpark dataframes to pandas dataframes was losing precision on 19-digit numbers.
```
sfc-gh-stan, line 7:
```suggestion
- Fixed a bug in `DataFrame.to_pandas()` where converting snowpark dataframes to pandas dataframes was losing precision on numbers with more than 19 digits.
```

sfc-gh-yixie (approved):

sfc-gh-yixie (approved):

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -655,7 +665,20 @@ def _fix_pandas_df_integer(
...
-            pd_df[pandas_col_name] = pandas.to_numeric(
-                pd_df[pandas_col_name], downcast="integer"
-            )
+            # pandas.to_numeric raises "RuntimeWarning: invalid value encountered in cast"
+            # when it tried to downcast and loses precision. In this case, we try to convert to
+            # int64 using astype() and fallback to to_numeric if we were unsuccessful.
+            with warnings.catch_warnings(record=True) as warning_list:
+                pd_col_with_numeric_downcast = pandas.to_numeric(
+                    pd_df[pandas_col_name], downcast="integer"
+                )
```
sfc-gh-yixie, line 674: This isn't reliable but it won't make things worse. We can use it for now if no other options. Could you create a JIRA to follow up?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -655,7 +665,20 @@ def _fix_pandas_df_integer(
...
+
+            if not warning_list:
+                pd_df[pandas_col_name] = pd_col_with_numeric_downcast
+            else:
+                try:
+                    pd_df[pandas_col_name] = pd_df[pandas_col_name].astype("int64")
+                except Exception:
```
sfc-gh-yixie, line 681: I suggest you use the specific exception.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -655,7 +665,20 @@ def _fix_pandas_df_integer(
...
-            )
+            # pandas.to_numeric raises "RuntimeWarning: invalid value encountered in cast"
+            # when it tried to downcast and loses precision. In this case, we try to convert to
+            # int64 using astype() and fallback to to_numeric if we were unsuccessful.
+            with warnings.catch_warnings(record=True) as warning_list:
+                pd_col_with_numeric_downcast = pandas.to_numeric(
+                    pd_df[pandas_col_name], downcast="integer"
+                )
+
+            if not warning_list:
```
sfc-gh-yixie, line 676: Is it possible to check the warning_list content?

sfc-gh-aalam: it is possible. what kind of logic would you suggest be better here? `if len(warning_list) == 1 and warning_list[0].message.startswith("RuntimeWarning")` ?

sfc-gh-yixie: Does the warning have a specific text? I no this is good. If yes I suggest you check if that text is in the warning.

sfc-gh-aalam: it does: `RuntimeWarning('invalid value encountered in cast')`

sfc-gh-yixie: Yeah checking this text will be better just in case there is another different warning.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -655,7 +665,24 @@ def _fix_pandas_df_integer(
...
+            with warnings.catch_warnings(record=True) as warning_list:
+                pd_col_with_numeric_downcast = pandas.to_numeric(
+                    pd_df[pandas_col_name], downcast="integer"
+                )
+
+            if (
+                len(warning_list) == 1
+                and warning_list[0].message
+                == "RuntimeWarning('invalid value encountered in cast')"
+            ):
```
sfc-gh-sfan, line 680: Can we just use the `try-catch` below and remove this `if-else`? It seems like we want to use `astype` whenever possible

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: We had a JIRA earlier about this issue. I am adding additional context to it and will write-up a proposal for GS/XP team to help us come-up with a better solution.

https://snowflakecomputing.atlassian.net/browse/SNOW-574649

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: I had this approach earlier but @sfc-gh-yixie pointed out that it can be a breaking change. It causes failure is some test cases where `pd.to_numeric()` with fit data into a pandas column of `int8` if the data is small enough to fit into it, whereas if we use `astype` whenever possible, we will change our old behavior.

PR #1102, opened 2023-10-20
[Local Testing] SNOW-915624 Fix DataFrame.sort where the orderby expression is not an Attribute

sfc-gh-stan:

   Fixes #NNNN

   Previously something like `df.sort(col("a")*col("b"))` would not work because the expression is not an existing column in the dataframe. This PR fixes the issue by modifying the implementation of `handle_order_by_clause ` to temporarily add the column `col("a")*col("b")` during sorting.

### DIFF
--- a/src/snowflake/snowpark/mock/plan.py
+++ b/src/snowflake/snowpark/mock/plan.py
@@ -199,8 +199,15 @@ def handle_order_by_clause(
     sort_columns_array = []
     sort_orders_array = []
     null_first_last_array = []
+    added_columns = []
     for exp in order_by:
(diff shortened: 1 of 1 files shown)


sfc-gh-sfan (approved):
This looks good, but I wonder if there are other places that needs similar treatment? If needed let's make sure we could have some code reuse :)

sfc-gh-stan:
> This looks good, but I wonder if there are other places that needs similar treatment? If needed let's make sure we could have some code reuse :)

Yea, I extracted the sort logic into `handle_order_by_clause` in order to reuse the code for Window Function handling. So far these two are the only places I found.

PR #1106, opened 2023-10-24
SNOW-946900: Add internal parameter in stored proc registration to allow forcing inline code

sfc-gh-jdu:

   Fixes SNOW-946900

   This is useful for Pandas API to skip code uploading when the stored procedure is registered by us (not users) and we are pretty sure the function won't be very large.

### DIFF
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -1397,3 +1397,20 @@ def test_sp_external_access_integration(session, db_parameters):
             )
             return
         raise
+
+
+def test_force_inline_code(session):
+    large_str = "snow" * 10000
+
+    def f(session: Session) -> int:
+        return len(large_str)
+
+    with session.query_history() as query_history:
+        _ = session.sproc.register(f, packages=["snowflake-snowpark-python"])
+    assert all("AS $$" not in query.sql_text for query in query_history.queries)
+
+    with session.query_history() as query_history:
+        _ = session.sproc.register(
+            f, packages=["snowflake-snowpark-python"], force_inline_code=True
+        )
+    assert any("AS $$" in query.sql_text for query in query_history.queries)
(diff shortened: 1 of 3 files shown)


tests/integ/test_stored_procedure.py
```diff
@@ -1397,3 +1397,20 @@ def return_success(session_):
             )
             return
         raise
+
+
+def test_force_inline_code(session):
```
sfc-gh-stan, line 1402: Can this be tested in Snowpandas instead? This addition is only meaningful to Snowpandas code for now.

sfc-gh-jdu: we can, but that part code will not be covered in snowpark tests. I don't think it's a good practice to leave uncovered code here.

tests/integ/test_stored_procedure.py
```diff
@@ -1397,3 +1397,20 @@ def return_success(session_):
...
+    def f(session: Session) -> int:
+        return len(large_str)
+
+    with session.query_history() as query_history:
+        _ = session.sproc.register(f, packages=["snowflake-snowpark-python"])
+    assert all("AS $$" not in query.sql_text for query in query_history.queries)
+
+    with session.query_history() as query_history:
+        _ = session.sproc.register(
+            f, packages=["snowflake-snowpark-python"], force_inline_code=True
```
sfc-gh-aalam, line 1414: I'm wondering if the way to set this param through public api should be with an internal parameter name like `session.sproc.register(..., _force_inline_code=True)`.

sfc-gh-jdu: This parameter is added to kwargs, so it won't show up in documents and isn't exposed to users. I follow the same approach as we did for `anonymous` parameter.

tests/integ/test_stored_procedure.py
sfc-gh-jdu: sure, I can do (1), but I'm not sure about (2) whether it's a good idea to add a code link to a private repo right now.

tests/integ/test_stored_procedure.py
sfc-gh-aalam: sounds good. 👍

sfc-gh-aalam (approved):

sfc-gh-stan (approved):

sfc-gh-sfan (approved):

PR #1109, opened 2023-10-26
[Local Testing] SNOW-931294 Support DataFrame.createOrReplaceView and DataFrame.createOrReplaceTempView

sfc-gh-stan:
### Please describe how your code solves the related issue.

This PR enables DataFrame.createOrReplaceView and DataFrame.createOrReplaceTempView by changing the existing `TableRegistry` class to store a separate map of string to `MockExecutionPlan` pairs, where the key is the view name (str) and the value is an `MockExeuctionPlan` instance corresponding to the underlying query of a [View](https://docs.snowflake.com/en/user-guide/views-introduction) . When users fetch from a view using `Session.table`, we execute the `MockExecutionPlan` to obtain the results. In this way, changes in dependencies (e.g. `sum(*)` from a table) will be correctly reflected.

### DIFF
--- a/src/snowflake/snowpark/mock/connection.py
+++ b/src/snowflake/snowpark/mock/connection.py
@@ -72,10 +72,11 @@ def _build_target_path(stage_location: str, dest_prefix: str = "") -> str:
 
 
 class MockServerConnection:
-    class TableRegistry:
-        # Table registry. TODO: move to datastore
+    class TabularEntityRegistry:
+        # Registry to store tables and views.
         def __init__(self, conn: "MockServerConnection") -> None:
             self.table_registry = {}
+            self.view_registry = {}
             self.conn = conn
 
         def get_fully_qualified_name(self, name: Union[str, Iterable[str]]) -> str:
@@ -88,6 +89,14 @@ class MockServerConnection:
             if len(name) == 2:
                 name = [current_database] + name
             return ".".join(quote_name(n) for n in name)
+
+        def is_existing_table(self, name: Union[str, Iterable[str]]) -> bool:
+            qualified_name = self.get_fully_qualified_name(name)
+            return qualified_name in self.table_registry
+
+        def is_existing_view(self, name: Union[str, Iterable[str]]) -> bool:
+            qualified_name = self.get_fully_qualified_name(name)
+            return qualified_name in self.view_registry
 
         def read_table(self, name: Union[str, Iterable[str]]) -> TableEmulator:
             qualified_name = self.get_fully_qualified_name(name)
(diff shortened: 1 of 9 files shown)


src/snowflake/snowpark/mock/plan.py
```diff
@@ -496,7 +512,9 @@ def execute_mock_plan(
             for exp in source_plan.grouping_expressions
         ]
         for column_name, _, column_type in column_exps:
-            result_df_sf_Types[column_name] = column_type
+            result_df_sf_Types[
+                column_name
+            ] = column_type  # TODO: fix this, this does not work
```
sfc-gh-stan, line 517: This  is blocking enabling `test_create_temp_view_on_functions`, essentially `col("a")` does not initialize the datatype of the column since it is an unresolved attribute. We probably need to `calculate_expression` to fix this.

src/snowflake/snowpark/mock/connection.py
```diff
@@ -135,6 +144,18 @@ def drop_table(self, name: Union[str, Iterable[str]]) -> None:
             if name in self.table_registry:
                 self.table_registry.pop(name)
 
+        def create_or_replace_view(
+            self, execution_plan: MockExecutionPlan, name: Union[str, Iterable[str]]
+        ):
+            name = self.get_fully_qualified_name(name)
+            self.view_registry[name] = execution_plan
```
sfc-gh-aling, line 151: why we are storing the plan instead of just executing and storing the result TableEmulator like how we do with table?

src/snowflake/snowpark/table.py
```diff
@@ -682,8 +682,8 @@ def drop_table(self, if_exists: bool = False) -> None:
         Note that subsequent operations such as :meth:`DataFrame.select`, :meth:`DataFrame.collect` on this ``Table`` instance and the derived DataFrame will raise errors because the underlying
         table in the Snowflake database no longer exists.
         """
-        if hasattr(self._session._conn, "table_registry"):
-            self._session._conn.table_registry.drop_table(self.table_name)
+        if hasattr(self._session._conn, "entity_registry"):
+            self._session._conn.entity_registry.drop_table(self.table_name)
```
sfc-gh-aling, line 686: we only have the `entity_registry`/`table_registry` attribute on mock connection right?

src/snowflake/snowpark/mock/connection.py
sfc-gh-aling: view is dynamically generated

src/snowflake/snowpark/mock/plan.py
sfc-gh-aling: let's make sure we create a JIRA for this

sfc-gh-aling (approved):

sfc-gh-aling:
@sfc-gh-stan I have fixed the datatype issue of the group by column by calculating the expression, I think we should be able to merge the PR

PR #1111, opened 2023-10-26
SNOW-947875 Add sql_error_code and raw_message into SnowparkSQLException

sfc-gh-aalin:

   Fixes SNOW-947875. When there is a parent stored procedure calling a child stored procedure, the parent stored procedure needs to determine the error code in the child query to determine how to format its error message. (See SNOW-800808 for more context.) Currently, the only way to do this when the parent is a Python sproc is to parse `SnowflakeSQLException.message`. It's slightly nicer and less error prone to have this information exposed directly in the exception class.

Adds raw message and error code attributes from ProgrammingError and OperationalError

### DIFF
--- a/src/snowflake/snowpark/exceptions.py
+++ b/src/snowflake/snowpark/exceptions.py
@@ -79,12 +79,16 @@ class SnowparkSQLException(SnowparkClientException):
         error_code: Optional[str] = None,
         sfqid: Optional[str] = None,
         query: Optional[str] = None,
+        sql_error_code: Optional[int] = None,
+        raw_message: Optional[str] = None,
     ) -> None:
         self.message: str = message
         self.error_code: Optional[str] = error_code
         self.sfqid: Optional[str] = sfqid
         self.query: Optional[str] = query
         self.telemetry_message: str = message
+        self.sql_error_code = sql_error_code
+        self.raw_message = raw_message
 
         pretty_error_code = f"({self.error_code}): " if self.error_code else ""
         pretty_sfqid = f"{self.sfqid}: " if self.sfqid else ""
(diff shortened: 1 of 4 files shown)


sfc-gh-stan:
Could you please provide a summary on why we need these changes?

sfc-gh-aalin:
> Could you please provide a summary on why we need these changes?

Done.

sfc-gh-aalin:

sfc-gh-sfan (approved):

src/snowflake/snowpark/exceptions.py
```diff
@@ -79,12 +79,16 @@ def __init__(
         error_code: Optional[str] = None,
         sfqid: Optional[str] = None,
         query: Optional[str] = None,
+        sql_error_code: Optional[int] = None,
+        raw_message: Optional[str] = None,
```
sfc-gh-aling, line 83: what will be the difference between `error_code` and `sql_error_code`, `message` and `raw_message`?
and who will be using(set/get) the two new error fields?

I ask because I feel those are duplicate to the existing ones
or maybe consider introducing a new sub class dedicate to your error scenario

sfc-gh-aalin: For example, if I get a [SQL_INTERNAL_ERROR](https://github.com/snowflakedb/snowflake/blob/f54a602c44a594c446b9803a0bbef042c3aec69c/GlobalServices/modules/platform/core/src/main/resources/com/snowflake/gs_error_messages.properties#L23), this is what each of the values will be:

* `error_code`: 1304 (I'm not entirely sure what this number means, but it looks like a Snowpark error code system distinct from the SQL error code system)
* `sql_error_code`: 000603
* `message`: Something like `000603 (P0000): SQL execution internal error: blah blah`
* `raw_message`: `SQL execution internal error: blah blah`

> and who will be using(set/get) the two new error fields?

The intent is only for this to be populated by the `snowpark-python` library when an error is thrown by `snowflake-connector` library. From what I can tell, `ProgrammingError` and `OperationalError` are the only two errors that can come from the connector library.

sfc-gh-smirzaei (approved):

sfc-gh-aling:
testing PR: https://github.com/snowflakedb/snowpark-python/pull/1123

sfc-gh-aling (approved):

PR #1115, opened 2023-10-30
[Local Testing] SNOW-904981 Support Column bitwise operations and unary minus expression

sfc-gh-stan:
**TODO:**
There are 2 data type issues blocking the PR:
- [x] We need to align `ColumnEmulator.dtype` and `ColumnEmulator.sf_type` to allow pandas operations to succeed.
- [x] We need to align `ColumnEmulator.dtype` and the Python object type of the data inside ColumnEmulator.


sfc-gh-aling (approved):
pre-approve assuming the test code will be updated to use `if not local_testing_mode` check to keep the live test

PR #1124, opened 2023-11-01
SNOW-948486 Support specifying input column names for vectorized UDTF

sfc-gh-stan:
[SNOW-948486](https://snowflakecomputing.atlassian.net/browse/SNOW-948486)

The PR introduces a new optional parameter `input_names` to `UDTFRegistration.register/register_file` and `functions.py/pandas_udtf`, where we now support specifying input column names for UDTF's. Note this is essentially a NOOP for regular UDTFs since the columns are mapped to function parameters, the feature is only useful for vectorized UDTF where we rely on names to access the columns in the pandas DataFrame. The parameter is set as optional for backwards compatibility, if unspecified, the default column names will be `ARG1`, `ARG2`, etc.

As a result, we now infer the column names from schema in `RelationalGroupedDataFrame.applyInPandas`.

[SNOW-948486]: https://snowflakecomputing.atlassian.net/browse/SNOW-948486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snowflake/snowpark/relational_grouped_dataframe.py
+++ b/src/snowflake/snowpark/relational_grouped_dataframe.py
@@ -2,7 +2,7 @@
 #
 # Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
 #
-from typing import Callable, Dict, List, Optional, Tuple, Union
+from typing import Callable, Dict, List, Tuple, Union
 
 from snowflake.connector.options import pandas
 from snowflake.snowpark import functions
@@ -259,7 +259,6 @@ class RelationalGroupedDataFrame:
         self,
         func: Callable,
         output_schema: StructType,
-        input_names: Optional[List[str]],
         **kwargs,
     ) -> DataFrame:
         """Maps each grouped dataframe in to a pandas.DataFrame, applies the given function on
@@ -288,12 +287,10 @@ class RelationalGroupedDataFrame:
                 >>> import pandas as pd
                 >>> from snowflake.snowpark.types import StructType, StructField, StringType, FloatType
                 >>> def convert(pandas_df):
-                ...     return pandas_df.assign(temp_f = lambda x: x.temp_c * 9 / 5 + 32)
-                ...
+                ...     return pandas_df.assign(TEMP_F = lambda x: x.TEMP_C * 9 / 5 + 32)
                 >>> df = session.createDataFrame([('SF', 21.0), ('SF', 17.5), ('SF', 24.0), ('NY', 30.9), ('NY', 33.6)],
                 ...         schema=['location', 'temp_c'])
                 >>> df.group_by("location").apply_in_pandas(convert,
-                ...     input_names=['"location"', '"temp_c"'],
                 ...     output_schema=StructType([StructField("location", StringType()),
                 ...                               StructField("temp_c", FloatType()),
                 ...                               StructField("temp_f", FloatType())])).order_by("temp_c").show()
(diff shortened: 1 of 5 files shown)


sfc-gh-sfan (commented):
Could you elaborate on how this API change can work together with the behavior change? i.e. What happens if we release before the BCR server change, and what happens if we release after the BCR server change.

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -348,6 +355,7 @@ def end_partition(self, pdf: pandas.DataFrame) -> pandas.DataFrame:
         _apply_in_pandas_udtf = self._df._session.udtf.register(
             _ApplyInPandas,
             output_schema=output_schema,
+            input_names=input_names,
```
sfc-gh-sabdelrahman, line 358: Can't we get the input column names from `self._df.schema` similar to how we get the `input_types` here:  https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/relational_grouped_dataframe.py#L344-L346 ?

sfc-gh-stan:
> Could you elaborate on how this API change can work together with the behavior change? i.e. What happens if we release before the BCR server change, and what happens if we release after the BCR server change.

The server change has already been released and enabled in all deployments by default. The change basically allows users to specify the input column names from function definition command (using the signature). Previously the function signature will be ignored and users had to set columns like `pdf.columns = ["col1", "col2"]` in the function body. 
Note that specifying column names is essentially a NOOP for regular UDTF's, since we just use function parameters to access the input columns, so client never supported specifying column names for UDTF input. That's why this change is required to accommodate the BCR. 
Please see the changed doc tests for how BCR impacts vectorized UDTF registration.

src/snowflake/snowpark/relational_grouped_dataframe.py
sfc-gh-aalam: I agree with Sara. we can infer `input_names` from `self._df.schema` similar to how to do it for `input_types`.

tests/integ/test_udtf.py
```diff
@@ -284,18 +284,18 @@ def process(
...
         return pdf.assign(temp_f=lambda x: x.temp_c * 9 / 5 + 32)
 
     df = session.createDataFrame(
         [("SF", 21.0), ("SF", 17.5), ("SF", 24.0), ("NY", 30.9), ("NY", 33.6)],
-        schema=["location", "temp_c"],
+        schema=['"location"', '"temp_c"'],
     )
 
-    df = df.group_by("location").apply_in_pandas(
+    df = df.group_by('"location"').apply_in_pandas(
```
sfc-gh-aalam, line 296: does the old code not work with this change? That could be a BCR if so

sfc-gh-stan: ah thanks for catching, these are not needed, removing...

src/snowflake/snowpark/relational_grouped_dataframe.py
sfc-gh-stan: Oh somehow I misread the part about `self._df.schema`. In that case it does make sense to do that instead.

sfc-gh-sfan (approved):

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1130, opened 2023-11-05
Add RelationalGroupedDataFrame.pivot()

### LINKED ISSUE #1093: SNOW-944062: Implementation and functionality of `pivot` differs from PySpark and is not user-friendly

Currently the `pivot` method is a method of a dataframe as shown in the [docs](https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/latest/api/snowflake.snowpark.DataFrame.pivot). You thus have to pass the pivot column and even the distinct values for pivoting. It is not possible use `pivot` in combination with a GroupedDataframe. 

This is not compatible to how [PySpark implements pivot](https://spark.apache.org/docs/3.2.0/api/python/reference/api/pyspark.sql.GroupedData.pivot.html). Here `pivot` is a function on a GroupedDataframe. And the specific values don't need to be specified.

Implement the same behavior as in PySpark. 

It will be easier to migrate from PySpark to Snowpark and Pyspark's pivot behaviour is just more user-friendly as in most cases you group before you apply it. 

## References, Other Background

* https://spark.apache.org/docs/3.2.0/api/python/reference/api/pyspark.sql.GroupedData.pivot.html
* https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/latest/api/snowflake.snowpark.DataFrame.pivot

sfc-gh-aalam:

   Fixes #1093

   Added a`pivot` method in `RelationalGroupedDataFrame` which allows to access `pivot` using `df.group_by().pivot()`

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -934,6 +934,7 @@ class Analyzer:
             if (
                 len(logical_plan.grouping_columns) != 0
                 and len(logical_plan.aggregates) != 0
+                and logical_plan.aggregates[0] is not None
             ):
                 # Currently snowflake pivot creates a group by from all columns outside of
                 # pivot column and aggregate column. In order to implement df.group_by().pivot(),
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -931,7 +931,10 @@ class Analyzer:
             )
 
         if isinstance(logical_plan, Pivot):
-            if len(logical_plan.grouping_columns) != 0:
+            if (
+                len(logical_plan.grouping_columns) != 0
+                and len(logical_plan.aggregates) != 0
+            ):
                 # Currently snowflake pivot creates a group by from all columns outside of
                 # pivot column and aggregate column. In order to implement df.group_by().pivot(),
                 # we need to first select only the columns that need to be involved in this
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -933,8 +933,7 @@ class Analyzer:
         if isinstance(logical_plan, Pivot):
             if (
                 len(logical_plan.grouping_columns) != 0
-                and len(logical_plan.aggregates) != 0
-                and logical_plan.aggregates[0] is not None
+                and logical_plan.aggregates[0].children is not None
             ):
                 # Currently snowflake pivot creates a group by from all columns outside of
                 # pivot column and aggregate column. In order to implement df.group_by().pivot(),
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -931,6 +931,26 @@ class Analyzer:
             )
 
         if isinstance(logical_plan, Pivot):
+            if len(logical_plan.grouping_columns) != 0:
+                # project_cols = grouping_cols + pivot_col + aggregate_col
+                project_exprs = [
+                    *logical_plan.grouping_columns,
+                    logical_plan.aggregates[0].children[
+                        0
+                    ],  # aggregate column is first child in aggregates
+                    logical_plan.pivot_column,
+                ]
+                child = self.plan_builder.project(
+                    [
+                        self.analyze(col, df_aliased_col_name_to_real_col_name)
+                        for col in project_exprs
+                    ],
+                    resolved_children[logical_plan.child],
+                    logical_plan,
+                )
+            else:
+                child = resolved_children[logical_plan.child]
+
             return self.plan_builder.pivot(
                 self.analyze(
                     logical_plan.pivot_column, df_aliased_col_name_to_real_col_name
@@ -942,7 +962,7 @@ class Analyzer:
                 self.analyze(
                     logical_plan.aggregates[0], df_aliased_col_name_to_real_col_name
                 ),
-                resolved_children[logical_plan.child],
+                child,
                 logical_plan,
             )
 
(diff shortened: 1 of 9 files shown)


src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -365,6 +366,103 @@ def end_partition(self, pdf: pandas.DataFrame) -> pandas.DataFrame:
...
+            <BLANKLINE>
+        """
+        pc = self._df._convert_cols_to_exprs(
+            "RelationalGroupedDataFrame.pivot()", pivot_col
+        )
+        if values is None:
+            distinct_values = (
+                self._df.select(pivot_col).distinct()._internal_collect_with_tag()
+            )
+            value_exprs = [Literal(v[0]) for v in distinct_values]
```
sfc-gh-aalam, line 458: should we add this in `Dataframe.pivot` as well with a documentation update saying this method is not recommended?

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -365,6 +366,103 @@ def end_partition(self, pdf: pandas.DataFrame) -> pandas.DataFrame:
...
+
+            >>> create_result = session.sql('''create or replace temp table monthly_sales(empid int, team text, amount int, month text)
+            ... as select * from values
+            ... (1, 'A', 10000, 'JAN'),
+            ... (1, 'B', 400, 'JAN'),
+            ... (2, 'A', 4500, 'JAN'),
+            ... (2, 'A', 35000, 'JAN'),
+            ... (1, 'B', 5000, 'FEB'),
+            ... (1, 'A', 3000, 'FEB'),
+            ... (2, 'B', 200, 'FEB') ''').collect()
```
sfc-gh-sfan, line 417: I wonder if we could use `create_dataframe` then `save_as_table` to represent this `session.sql`? Maybe it is not worth it but just thinking out loud.

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -365,6 +366,103 @@ def end_partition(self, pdf: pandas.DataFrame) -> pandas.DataFrame:
...
+            ----------------------------------------
+            <BLANKLINE>
+        """
+        pc = self._df._convert_cols_to_exprs(
+            "RelationalGroupedDataFrame.pivot()", pivot_col
+        )
+        if values is None:
+            distinct_values = (
+                self._df.select(pivot_col).distinct()._internal_collect_with_tag()
+            )
```
sfc-gh-sfan, line 457: This is actually a blocking call, and might not be cheap. I wonder if we should highlight why this is not efficient in the documentation, in addition to saying this is not recommended.

Alternatively, should we lazy evaluate this query (put them into the query plan), or at least do a async query here and fetch the result only when the value is required?

sfc-gh-sfan (approved):
Let's make sure we have a TODO for distinct values

sfc-gh-sfan (commented):
Do we need to raise if `values` is `None`?

sfc-gh-aalam:
> Let's make sure we have a TODO for distinct values

https://snowflakecomputing.atlassian.net/browse/SNOW-967385

sfc-gh-aalam:
> Do we need to raise if `values` is `None`?

Right now, `df.pivot` doesn't raise it but we could raise a `ValueError` in both places saying `values cannot be empty`. Right now we get `NoneType is not iterable`

sfc-gh-sfan:
I see. Both works, probably an explicit error message is easier for user to understand :)

src/snowflake/snowpark/dataframe.py
```diff
@@ -1624,6 +1633,8 @@ def pivot(
             pivot_col: The column or name of the column to use.
             values: A list of values in the column.
         """
+        if not values:
+            raise ValueError("values cannot be None or empty")
```
sfc-gh-mkeller, line 1637: This should never be `None` according to the argument's type-hint
```suggestion
            raise ValueError("values cannot be empty")
```

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -931,6 +931,32 @@ def do_resolve_with_resolved_children(
...
+                # Currently snowflake pivot creates a group by from all columns outside of
+                # pivot column and aggregate column. In order to implement df.group_by().pivot(),
+                # we need to first select only the columns that need to be involved in this
+                # operation, and then apply pivot operation. Here, we will first use project
+                # plan to select group_by, pivot and aggregate column and then apply the pivot
+                # logic.
+                #     project_cols = grouping_cols + pivot_col + aggregate_col
+                project_exprs = [
+                    *logical_plan.grouping_columns,
+                    logical_plan.aggregates[0].children[
```
sfc-gh-mkeller, line 944: This is failing the `mypy` check, please fix it

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -365,6 +366,62 @@ def end_partition(self, pdf: pandas.DataFrame) -> pandas.DataFrame:
...
+            ----------------------------------------
+            |1        |A       |10000    |3000     |
+            |1        |B       |400      |5000     |
+            |2        |A       |39500    |NULL     |
+            |2        |B       |NULL     |200      |
+            ----------------------------------------
+            <BLANKLINE>
+        """
+        if not values:
+            raise ValueError("values cannot be None or empty")
```
sfc-gh-mkeller, line 415: Same as the other
```suggestion
            raise ValueError("values cannot be empty")
```

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -365,6 +366,62 @@ def end_partition(self, pdf: pandas.DataFrame) -> pandas.DataFrame:
...
+        pc = self._df._convert_cols_to_exprs(
+            "RelationalGroupedDataFrame.pivot()", pivot_col
+        )
+        value_exprs = [
+            v._expression if isinstance(v, Column) else Literal(v) for v in values
+        ]
+        self._group_type = _PivotType(pc[0], value_exprs)
+        return self
```
sfc-gh-mkeller, line 423: You could have the return-type be [typing_extensions.Self](https://docs.python.org/3/library/typing.html#typing.Self) here

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1137, opened 2023-11-16
SNOW-950842 Add support for the vector data type

sfc-gh-rpanchapakesan:

   Fixes SNOW-950842

Added support for the vector data type to snowpark. As with arrays, we temporarily serialize vector inputs as strings and wrap `parse_json` and `cast` calls to construct the right type. Let me know if there are any edge cases that this temporary bypass introduces that I need to handle.

There are many formatting changes that were automatically applied by the repository's precommit because I edited certain files.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
+++ b/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
@@ -205,14 +205,14 @@ def schema_expression(data_type: DataType, is_nullable: bool) -> str:
     if isinstance(data_type, GeometryType):
         return "to_geometry('POINT(-122.35 37.55)')"
     if isinstance(data_type, VectorType):
-        make_float = 0.1 if data_type.element_type == "float" else 0
-        values = [i + 1 + make_float for i in range(data_type.dimension)]
         if data_type.element_type == "int":
-            return f"{values} :: VECTOR({data_type.element_type},{data_type.dimension})"
+            zero = int(0)
         elif data_type.element_type == "float":
-            return f"{values} :: VECTOR({data_type.element_type},{data_type.dimension})"
+            zero = float(0)
         else:
             raise TypeError(f"Invalid vector element type: {data_type.element_type}")
+        values = [i + zero for i in range(data_type.dimension)]
+        return f"{values} :: VECTOR({data_type.element_type},{data_type.dimension})"
     raise Exception(f"Unsupported data type: {data_type.__class__.__name__}")
 
 
--- a/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
+++ b/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
@@ -31,6 +31,7 @@ from snowflake.snowpark.types import (
     TimestampType,
     TimeType,
     VariantType,
+    VectorType,
     _FractionalType,
     _IntegralType,
     _NumericType,
@@ -73,6 +74,9 @@ def to_sql(value: Any, datatype: DataType, from_values_statement: bool = False) 
     if isinstance(datatype, VariantType):
         if value is None:
             return "NULL :: VARIANT"
+    if isinstance(datatype, VectorType):
+        if value is None:
+            return f"NULL :: VECTOR({datatype.element_type},{datatype.dimension})"
     if value is None:
         return "NULL"
 
@@ -149,6 +153,9 @@ def to_sql(value: Any, datatype: DataType, from_values_statement: bool = False) 
         # PARSE_JSON returns VARIANT, so no need to append :: VARIANT here explicitly.
         return f"PARSE_JSON({str_to_sql(json.dumps(value, cls=PythonObjJSONEncoder))})"
 
+    if isinstance(datatype, VectorType):
+        return f"{value} :: VECTOR({datatype.element_type},{datatype.dimension})"
+
     raise TypeError(f"Unsupported datatype {datatype}, value {value} by to_sql()")
 
 
@@ -197,6 +204,15 @@ def schema_expression(data_type: DataType, is_nullable: bool) -> str:
         return "to_geography('POINT(-122.35 37.55)')"
     if isinstance(data_type, GeometryType):
         return "to_geometry('POINT(-122.35 37.55)')"
+    if isinstance(data_type, VectorType):
+        make_float = 0.1 if data_type.element_type == "float" else 0
+        values = [i + 1 + make_float for i in range(data_type.dimension)]
+        if data_type.element_type == "int":
+            return f"{values} :: VECTOR({data_type.element_type},{data_type.dimension})"
+        elif data_type.element_type == "float":
+            return f"{values} :: VECTOR({data_type.element_type},{data_type.dimension})"
+        else:
+            raise TypeError(f"Invalid vector element type: {data_type.element_type}")
     raise Exception(f"Unsupported data type: {data_type.__class__.__name__}")
 
 
--- a/src/snowflake/snowpark/_internal/analyzer/schema_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/schema_utils.py
@@ -2,27 +2,33 @@
 # Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
 #
 
-from typing import TYPE_CHECKING, List
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
 
 import snowflake.snowpark
-from snowflake.connector.constants import FIELD_ID_TO_NAME
-from snowflake.connector.cursor import ResultMetadata
+from snowflake.connector.cursor import ResultMetadata, SnowflakeCursor
 from snowflake.snowpark._internal.analyzer.analyzer_utils import (
     quote_name_without_upper_casing,
 )
 from snowflake.snowpark._internal.analyzer.expression import Attribute
-from snowflake.snowpark._internal.type_utils import convert_sf_to_sp_type
+from snowflake.snowpark._internal.type_utils import convert_metadata_to_sp_type
 from snowflake.snowpark.types import DecimalType, LongType, StringType
 
 if TYPE_CHECKING:
     import snowflake.snowpark.session
 
+    try:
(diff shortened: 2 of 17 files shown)


src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -197,6 +204,15 @@ def schema_expression(data_type: DataType, is_nullable: bool) -> str:
         return "to_geography('POINT(-122.35 37.55)')"
     if isinstance(data_type, GeometryType):
         return "to_geometry('POINT(-122.35 37.55)')"
+    if isinstance(data_type, VectorType):
+        make_float = 0.1 if data_type.element_type == "float" else 0
```
sfc-gh-mloring, line 208: Nit: This might read a bit more cleanly if the element type check is hoisted to the top of the block and the return logic is deduplicated
```
if data_type.element_type == "int":
    zero = int(0)
elif data_type.element_type == "float":
    zero = float(0)
else:
    raise TypeError(f"Invalid vector element type: {data_type.element_type}")
values = [i + zero for i in range(data_type.dimension]
return f"{values} :: VECTOR({data_type.element_type},{data_type.dimension})"
```

src/snowflake/snowpark/_internal/analyzer/schema_utils.py
```diff
@@ -2,21 +2,25 @@
...
 )
 from snowflake.snowpark._internal.analyzer.expression import Attribute
-from snowflake.snowpark._internal.type_utils import convert_sf_to_sp_type
+from snowflake.snowpark._internal.type_utils import convert_metadata_to_sp_type
 from snowflake.snowpark.types import DecimalType, LongType, StringType
 
 if TYPE_CHECKING:
     import snowflake.snowpark.session
 
+    try:
```
sfc-gh-mloring, line 19: Let's add a comment here describing the difference between `ResultMetadataV2` and `ResultMetadata` and noting that we intend to unify them in a future release.

src/snowflake/snowpark/_internal/analyzer/schema_utils.py
```diff
@@ -85,20 +89,50 @@ def analyze_attributes(
...
-                ),
-                nullable,
+                convert_metadata_to_sp_type(column_metadata),
+                column_metadata.is_nullable,
             )
         )
     return attributes
+
+
+def get_new_description_if_exists(
```
sfc-gh-mloring, line 111: Nit: The `if_exists` naming might be a bit confusing and could suggest that `None` might be returned here or the query may not be run if the new describe doesn't exist below. I would suggest dropping it and communicating the fallback behavior in the docstring.

src/snowflake/snowpark/types.py
```diff
@@ -455,6 +479,11 @@ def get_snowflake_col_datatypes(self):
 #: The type hint for annotating Geometry data when registering UDFs.
 Geometry = TypeVar("Geometry")
 
+# TODO(SNOW-969479): Add a type hint that can be used to annotate Vector data. Python does not
+# currently support integer type parameters (which are needed to represent a vector's dimension).
+# typing.Annotate can be used as a temporary bypass once the minimum supported Python version is
+# bumped to 3.9
```
sfc-gh-mloring, line 485: Looks like you're not the first person to reach for dependent types in Python: https://typing-extensions.readthedocs.io/en/latest/#Literal 😄

sfc-gh-mloring (approved):
Looks good on the vector side. Thanks for putting this together!

src/snowflake/snowpark/types.py
sfc-gh-rpanchapakesan: Yeah, if `Literal` supported unbounded domains, I think it'd be all we need.

src/snowflake/snowpark/_internal/analyzer/schema_utils.py
sfc-gh-rpanchapakesan: I have an explanation in the docstring for ResultMetadataV2, so I added a comment here pointing readers to that.

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -197,6 +204,15 @@ def schema_expression(data_type: DataType, is_nullable: bool) -> str:
...
     if isinstance(data_type, GeometryType):
         return "to_geometry('POINT(-122.35 37.55)')"
+    if isinstance(data_type, VectorType):
+        if data_type.element_type == "int":
+            zero = int(0)
+        elif data_type.element_type == "float":
+            zero = float(0)
+        else:
+            raise TypeError(f"Invalid vector element type: {data_type.element_type}")
+        values = [i + zero for i in range(data_type.dimension)]
```
sfc-gh-sfan, line 214: Curious: What is this zero doing?

src/snowflake/snowpark/_internal/analyzer/schema_utils.py
```diff
@@ -2,21 +2,27 @@
...
 from snowflake.snowpark._internal.analyzer.expression import Attribute
-from snowflake.snowpark._internal.type_utils import convert_sf_to_sp_type
+from snowflake.snowpark._internal.type_utils import convert_metadata_to_sp_type
 from snowflake.snowpark.types import DecimalType, LongType, StringType
 
 if TYPE_CHECKING:
     import snowflake.snowpark.session
 
+    # Refer to the docstring for ResultMetadataV2 (in cursor.py) for more information about
+    # how it differs from ResultMetadata
```
sfc-gh-sfan, line 20: We could also bump the connector dependency lower bound to make sure `ResultMetadataV2` exists. Also, make sure sp connector has the same change released.

sfc-gh-sfan (commented):
I wonder if we should bump the connector version lower bound such that we don't need the fallback in code

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
sfc-gh-rpanchapakesan: This is to ensure we add float elements when creating float vectors (adding `float(0)` results in `1.0, 2.0, 3.0`). This isn't strictly necessary since internally we can create float vectors from integers, but I added it for clarity for debugging etc.

src/snowflake/snowpark/_internal/analyzer/schema_utils.py
sfc-gh-rpanchapakesan: Is this safe to do? That would definitely clean up the code, but I wasn't sure whether we could bump the version since this change is going out next week with the corresponding connector version that it would rely on. If that's fine to do, I can make the update

sfc-gh-sfan: We typically first release connector, then snowpark, so it should be safe. We also bumped the version for geometry type I think. cc @sfc-gh-aalam to confirm

sfc-gh-rpanchapakesan: Will the merge gates pull a version of the connector from main? If not, I don't think they'll pass once I make the update.

tests/integ/scala/test_udf_suite.py
```diff
@@ -549,6 +550,59 @@ def geometry(g):
     )
 
 
+@pytest.mark.xfail(reason="SNOW-974852 vectors are not yet rolled out", strict=False)
```
sfc-gh-kbhatia, line 553: We can instead flip the param on the small set of preprod/prod accounts that run these tests.  ?

sfc-gh-kbhatia (approved):

sfc-gh-rpanchapakesan:
I've also added vector distance functions (which I was initially planning on sending in a follow-up) to this PR in order to ensure all the changes make the release cutoff.

src/snowflake/snowpark/functions.py
```diff
@@ -5390,6 +5390,60 @@ def object_pick(obj: ColumnOrName, key1: ColumnOrName, *keys: ColumnOrName) -> C
     return builtin("object_pick")(o, k1, *ks)
 
 
+def vector_cosine_distance(v1: ColumnOrName, v2: ColumnOrName) -> Column:
+    """Returns the cosine distance between two vectors.
```
sfc-gh-mloring, line 5394: Can you please add `"of equal dimension and element type"` to each of these return clauses?

sfc-gh-mloring (approved):
Added vector distance functions LGTM

src/snowflake/snowpark/functions.py
```diff
@@ -5390,13 +5390,17 @@ def object_pick(obj: ColumnOrName, key1: ColumnOrName, *keys: ColumnOrName) -> C
     return builtin("object_pick")(o, k1, *ks)
 
 
+# The following three vector functions have doctests that are disabled (">>" instead of ">>>")
+# since vectors are not yet rolled out.
```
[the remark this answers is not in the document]
sfc-gh-rpanchapakesan: Will do once vectors to roll out (at least in preprod), and the new connector is released.

src/snowflake/snowpark/_internal/analyzer/schema_utils.py
sfc-gh-aalam: To my understanding, only vector related function will be affected if older version of connector is used, right? We would need to bump connector dependency when we want to broader vector support release like this: https://github.com/snowflakedb/snowpark-python/pull/1156/files

sfc-gh-rpanchapakesan: Yeah, if an older connector is used, vector queries would fail but everything else should be fine (iiuc, this is verified as the CI gates are running the currently-released connector and are passing). I think we'll likely want to bump the connector dependency when we exit PrPr.

sfc-gh-rpanchapakesan:
I had to remove the alter parameter statements because they are not allowed in stored procedures. The tests are still marked xfail so this shouldn't have any downstream effects. However, I think we may want to manually enable vectors on the CI accounts, so we can remove the xfail once the new connector version is released.

I've verified that the tests marked xfail do pass if I run them manually with the not-yet-released connector and in an environment in which vectors are enabled.

PR #1141, opened 2023-11-21
SNOW-975597: use result scan for table sproc result

sfc-gh-aalam:

   Fixes #SNOW-975597


src/snowflake/snowpark/stored_procedure.py
```diff
@@ -102,8 +103,13 @@ def __call__(
 
         if self._anonymous_sp_sql:
             call_sql = generate_call_python_sp_sql(session, self.name, *args)
-            df = session.sql(f"{self._anonymous_sp_sql}{call_sql}")
+            query = f"{self._anonymous_sp_sql}{call_sql}"
+            df = session.sql(query)
             if self._is_return_table:
+                qid = session._conn.execute_and_get_sfqid(
```
sfc-gh-jrose, line 109: Does the same query get run twice?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -339,6 +339,22 @@ def notify_query_listeners(self, query_record: QueryRecord) -> None:
...
+        statement_params: Optional[Dict[str, str]] = None,
+    ) -> str:
+        if statement_params is None:
+            statement_params = {}
+        if is_ddl_on_temp_object:
+            statement_params["SNOWPARK_SKIP_TXN_COMMIT_IN_DDL"] = True
+        results_cursor = self._cursor.execute(query, _statement_params=statement_params)
+        self.notify_query_listeners(
+            QueryRecord(results_cursor.sfqid, results_cursor.query)
+        )
```
sfc-gh-sfan, line 355: Curious: is there a good way to reduce the duplication if `statement_params` and query listener's code? I would guess this code block is quite reusable.

sfc-gh-sfan (approved):

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -102,8 +103,13 @@ def __call__(
...
+                    query, statement_params=statement_params
+                )
+                df = session.sql(result_scan_statement(qid))
```
sfc-gh-mkeller, line 112: The return-type of this function is incorrect. It should be `Any` and not `any`. The lower-case version is a builtin. I know this isn't your doing, but could you please fix it?

src/snowflake/snowpark/stored_procedure.py
sfc-gh-aalam: `df = session.sql(query)` does not run the query, doing an internal collect or running function `execute_and_get_sfqid` does. I see df on line 107 is only used for the final return on line 114. I'll update the PR and move it so it become less confusing.

sfc-gh-jrose: I understand now. Thanks for explaining.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: I think I can remove extra processing done on statement params are cursor is happy with accepting a None statement param. I'll remove ddl on temp object to keep things simple as it does not have any usage now. seeing other occurrences of statement params, does not seem like there is much duplication.

I'll create a helper for execute and notify query listener as I see them being used together frequently. Do you want me to refactor them in this PR or should I create a follow-up JIRA for that

tests/integ/test_udf.py
```diff
@@ -1781,7 +1781,12 @@ def return_type_in_dataframe(x):
         (
             IntegerType,
             [[1048576]],
-            ("<class 'numpy.int32'>", "<class 'numpy.int64'>", "<class 'int'>", "<class 'numpy.intc'>"),
+            (
+                "<class 'numpy.int32'>",
+                "<class 'numpy.int64'>",
+                "<class 'int'>",
+                "<class 'numpy.intc'>",
+            ),
```
sfc-gh-aalam, line 1789: lint fix

Issue #1148, opened 2023-11-25
The withColumnRenamed fucntion fails to rename a column if the snowpark dataframe has multiple columns with same name but with different case style


Ilyas-kipi:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using?

Python 3.10.8 (tags/v3.10.8:aaaf517, Oct 11 2022, 16:50:30) [MSC v.1933 64 bit (AMD64)]

2. What operating system and processor architecture are you using?

Windows-10-10.0.19045-SP0

3. What are the component versions in the environment (`pip freeze`)?

altair==4.2.2
asn1crypto==1.5.1
attrs==23.1.0
blinker==1.6.2
cachetools==5.3.1
certifi==2023.7.22
cffi==1.15.1
charset-normalizer==3.2.0
click==8.1.7
cloudpickle==2.0.0
colorama==0.4.6
cron-descriptor==1.4.0
croniter==1.4.1
cryptography==41.0.3
entrypoints==0.4
filelock==3.12.3
gitdb==4.0.10
GitPython==3.1.32
greenlet==2.0.2
idna==3.4
importlib-metadata==6.8.0
Jinja2==3.1.2
jsonschema==4.19.0
jsonschema-specifications==2023.7.1
markdown-it-py==3.0.0
MarkupSafe==2.1.3
mdurl==0.1.2
numpy==1.25.2
oscrypto==1.3.0
packaging==23.1
pandas==2.0.3
Pillow==10.0.0
platformdirs==3.8.1
plotly==5.16.1
protobuf==3.20.3
pyarrow==13.0.0
pycparser==2.21
pycryptodomex==3.18.0
pydeck==0.8.1b0
Pygments==2.16.1
PyJWT==2.8.0
Pympler==1.0.1
pyOpenSSL==23.2.0
python-dateutil==2.8.2
pytz==2023.3
PyYAML==6.0.1
referencing==0.30.2
requests==2.31.0
rich==13.5.2
rpds-py==0.10.0
six==1.16.0
smmap==5.0.0
snowflake-connector-python==3.5.0
snowflake-snowpark-python==1.10.0
snowflake-sqlalchemy==1.5.0
sortedcontainers==2.4.0
SQLAlchemy==1.4.49
streamlit==1.22.0
tenacity==8.2.3
toml==0.10.2
tomlkit==0.12.1
toolz==0.12.0
tornado==6.3.3
typing_extensions==4.7.1
[...]

PR #1149, opened 2023-11-25
Update dataframe.py

Ilyas-kipi:
the upper() function is converting the column name to upper case and hence when the dataframe has two columns with same name but different case as show Ex- Column names = ['Snow Flake', 'SNOW FLAKE']. When the user tries to rename 'Snow Flake' column to 'Snow Flake Renamed', the current withColumnRenamed method throws an exception as the method converts 'Snow Flake' to upper case but since the dataframe already has another column called 'SNOW FLAKE', the to_be_renamed list will have 2 elements and hence an exception will be raised.

   SNOW-977836: The withColumnRenamed fucntion fails to rename a column if the snowpark dataframe has multiple columns with same name but with different case style

   the upper() function is converting the column name to upper case and hence when the dataframe has two columns with same name but different case as show Ex- Column names = ['Snow Flake', 'SNOW FLAKE']. When the user tries to rename 'Snow Flake' column to 'Snow Flake Renamed', the current withColumnRenamed method throws an exception as the method converts 'Snow Flake' to upper case but since the dataframe already has another column called 'SNOW FLAKE', the to_be_renamed list will have 2 elements and hence an exception will be raised. Removing the upper function will make sure that we compare the columns to be renamed and the existing dataframe columns in the same case as they exist and will not raise an exception if we have multiple column with same name but different case.

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -3554,7 +3554,7 @@ class DataFrame:
         else:
             raise TypeError(f"{str(existing)} must be a column name or Column object.")
 
-        to_be_renamed = [x for x in self._output if x.name.upper() == old_name.upper()]
+        to_be_renamed = [x for x in self._output if x.name == old_name]
         if not to_be_renamed:
             raise ValueError(
                 f'Unable to rename column "{existing}" because it doesn\'t exist.'


Ilyas-kipi:

suenalaba:
certainly fixes the issue, but I think tests need to be added in `test_dataframe.py`.

Also please see my [comment](https://github.com/snowflakedb/snowpark-python/issues/1148#issuecomment-1855072553) on renaming of duplicated columns.

Ilyas-kipi:
Sure, will add test cases around this functionality

Ilyas-kipi:
Hey @suenalaba , pls see my comments on #1148 , I've added test case to test_dataframe.py with the name test_with_column_renamed_case_sensitivity()

suenalaba:
looks good to me now

PR #1155, opened 2023-11-28
[Local Testing] SNOW-976759 Fix implementation of mock_avg to compute the correct return type and precision

sfc-gh-stan:

   N/A

   Previously `mock_avg` always return a column of floating point values rounded to 5 digits after the floating point, where the result type is incorrectly set to the type of the incoming column. As a result, average of Integer column will have a return type of Integer. Average of Decimal columns will not have the correct precision and scale. This PR fixes these issues.

### DIFF
--- a/src/snowflake/snowpark/mock/functions.py
+++ b/src/snowflake/snowpark/mock/functions.py
@@ -32,6 +32,7 @@ from snowflake.snowpark.types import (
     TimestampType,
     TimeType,
     VariantType,
+    _FractionalType,
     _IntegralType,
     _NumericType,
 )
@@ -164,6 +165,7 @@ def mock_avg(column: ColumnEmulator) -> ColumnEmulator:
             scale = 12
         res_type = DecimalType(precision, scale)
     else:
+        assert isinstance(column.sf_type.datatype, _FractionalType)
         res_type = FloatType()
 
     notna = column[~column.isna()]
--- a/src/snowflake/snowpark/mock/functions.py
+++ b/src/snowflake/snowpark/mock/functions.py
@@ -24,6 +24,7 @@ from snowflake.snowpark.types import (
     DateType,
     DecimalType,
     DoubleType,
+    FloatType,
     LongType,
     MapType,
     NullType,
@@ -31,6 +32,7 @@ from snowflake.snowpark.types import (
     TimestampType,
     TimeType,
     VariantType,
+    _IntegralType,
     _NumericType,
 )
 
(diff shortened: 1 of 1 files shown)


src/snowflake/snowpark/mock/functions.py
```diff
@@ -141,22 +143,34 @@ def mock_sum(column: ColumnEmulator) -> ColumnEmulator:
...
-        if not all_item_is_none
-        else ColumnEmulator(data=[None])
-    )
-    ret.sf_type = column.sf_type
-    return ret
+    if isinstance(column.sf_type.datatype, NullType) or column.isna().all():
+        return ColumnEmulator(data=[None], sf_type=ColumnType(NullType(), True))
+    elif isinstance(column.sf_type.datatype, _IntegralType):
+        res_type = DecimalType(38, 6)
+    elif isinstance(column.sf_type.datatype, DecimalType):
```
sfc-gh-yixie, line 155: We have other "fixed" types like IntegerType, LongType and DecimalType.
Pls refer to `def calculate_type(c1: ColumnType, c2: Optional[ColumnType], op: Union[str]):`

src/snowflake/snowpark/mock/functions.py
```diff
@@ -141,22 +143,34 @@ def mock_sum(column: ColumnEmulator) -> ColumnEmulator:
...
+            column.sf_type.datatype.precision,
+            column.sf_type.datatype.scale,
+        )
+        precision = max(38, column.sf_type.datatype.precision + 12)
+        if scale <= 6:
+            scale = scale + 6
+        elif scale < 12:
+            scale = 12
+        res_type = DecimalType(precision, scale)
+    else:
```
sfc-gh-yixie, line 166: How about an explicit checking for `FloatType` and `DoubleType`?

src/snowflake/snowpark/mock/functions.py
sfc-gh-stan: Shouldn't IntegerType, LongType be handled by the case on line 153?

src/snowflake/snowpark/mock/functions.py
```diff
@@ -141,22 +143,34 @@ def mock_sum(column: ColumnEmulator) -> ColumnEmulator:
...
+        res_type = DecimalType(38, 6)
+    elif isinstance(column.sf_type.datatype, DecimalType):
+        precision, scale = (
+            column.sf_type.datatype.precision,
+            column.sf_type.datatype.scale,
+        )
+        precision = max(38, column.sf_type.datatype.precision + 12)
+        if scale <= 6:
+            scale = scale + 6
+        elif scale < 12:
```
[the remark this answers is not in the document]
sfc-gh-stan: Then we use the original scale from input, i.e. it is unchanged from the assignment on line 156.

src/snowflake/snowpark/mock/functions.py
sfc-gh-stan: That's a good idea. Do you prefer using
```
elif instance(column.sf_type.datatype, _FractionalType)
    ...
else:
   raise Error
```
or an assert statement in the else case on line 166?

sfc-gh-yixie (approved):

sfc-gh-aalam (approved):

PR #1161, opened 2023-12-04
SNOW-870432: make use_logical_type backwards compatible

sfc-gh-aalam:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -1936,7 +1936,7 @@ class Session:
                 )
             signature = inspect.signature(write_pandas)
             if not ("use_logical_type" in signature.parameters):
-                # is write_pandas does not support use_logical_type, don't pass this argument
+                # do not pass use_logical_type if write_pandas does not support it
                 kwargs.pop("use_logical_type", None)
             success, nchunks, nrows, ci_output = write_pandas(
                 self._conn._conn,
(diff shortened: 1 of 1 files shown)


sfc-gh-aalam:
> Thanks so much for updating to be backwards compatible with older versions of the `snowflake-connector-python` library - very helpful for us over on Platform ML!
> 
> One note is you don't actually require v3.4.0 anymore, so in theory you could update your dependencies to be lower again. But perhaps that increased dependency is generally preferred. Overall LGTM.

We still want to let client users know that they need to update dependencies to use the `use_logical_type` parameter so I think we should keep it.

src/snowflake/snowpark/session.py
```diff
@@ -1933,6 +1934,10 @@ def write_pandas(
                     + (schema + "." if schema else "")
                     + (table_name)
                 )
+            signature = inspect.signature(write_pandas)
+            if not ("use_logical_type" in signature.parameters):
```
sfc-gh-stan, line 1938: Do we need to remove `use_logical_type` explicitly? I feel like we only need the change on line 1955?

sfc-gh-stan: Per our discussion on Slack, when this parameter is passed in kwargs, Python connector will pass it to PyArrow and hence cause the error there.

sfc-gh-stan (approved):

PR #1175, opened 2023-12-12
SNOW-971157: Add additional parameters to Session.add_import

sfc-gh-jrose:

   Fixes SNOW-971157

This PR adds a couple extra parameters to the `Session.add_import` method. These parameters allow the user to specify a specific chunk size when adding an import. A larger chunk_size can be useful when files start to get larger. A second parameter allows the user to specify that the entire file should be hashed. In this case it will be read one chunk at a time with the hashing algorithm updated each time.

I've left the default behavior of the the `calculate_checksum` alone, but it might be worth changing it to hash at least one chunk of each file in a folder instead of stopping after the total amount of file read is the chunk size.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.12.0 (TBD)
 
-### Behavior change
+### New Features
 
 - Added two optional arguments to `Session.add_import`.
   - `chunk_size`: The number of bytes to hash per chunk of the uploaded files.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Release History
 
+## 1.12.0 (TBD)
+
+### Behavior change
+
+- Added two optional arguments to `Session.add_import`.
+  - `chunk_size`: The number of bytes to hash per chunk of the uploaded files.
+  - `whole_file_hash`: By default only the first chunk of the uploaded import is hashed to save time. When this is set to True each uploaded file is fully hashed instead.
+
 ## 1.11.1 (2023-11-07)
 
 ### Bug Fixes
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -375,12 +375,33 @@ def parse_positional_args_to_list(*inputs: Any) -> List:
         return [*inputs]
 
 
+def _hash_file(
+    hash_algo: hashlib._hashlib.HASH, path: str, chunk_size: int, whole_file_hash: bool
+):
+    """
+    Reads from a file and updates the given hash algorithm with the read text.
+
+    Args:
+        hash_algo: The hash algorithm to updated.
+        path: The path to the file to be read.
+        chunk_size: How much of the file to read at a time.
+        whole_file_hash: When True the whole file is hashed rather than stopping after the first chunk.
+    """
+    with open(path, "rb") as f:
+        data = f.read(chunk_size)
+        hash_algo.update(data)
+        while data and whole_file_hash:
+            data = f.read(chunk_size)
+            hash_algo.update(data)
+
+
 def calculate_checksum(
     path: str,
     chunk_size: int = 8192,
     ignore_generated_py_file: bool = True,
     additional_info: Optional[str] = None,
     algorithm: str = "sha256",
+    whole_file_hash: bool = False,
 ) -> str:
     """Calculates the checksum of a file or a directory.
 
@@ -397,6 +418,7 @@ def calculate_checksum(
         additional_info: Any additional information we might want to include
             for checksum computation.
         algorithm: the hash algorithm.
+        whole_file_hash: When set to True the files will be completely read while hashing.
 
     Returns:
         The result checksum.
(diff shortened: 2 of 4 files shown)


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -423,14 +444,19 @@ def calculate_checksum(
...
+                    _hash_file(hash_algo, filename, chunk_size, whole_file_hash)
+                    current_size += file_size
                 if current_size < chunk_size:
-                    filename = os.path.join(dirname, file)
-                    file_size = os.path.getsize(filename)
                     read_size = min(file_size, chunk_size - current_size)
                     current_size += read_size
-                    with open(filename, "rb") as f:
-                        hash_algo.update(f.read(read_size))
+                    _hash_file(hash_algo, filename, read_size, False)
```
sfc-gh-aalam, line 459: just checking if we have a test case that covers this part of the code. i.e. chunk_size > file_size

sfc-gh-jrose: Good call. Most of the existing examples in the test for this code had smaller file size than chunk size.

sfc-gh-aalam (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## 1.12.0 (TBD)
 
+### Behavior change
```
sfc-gh-aling, line 5: This should be `New Features` instead of `Behavior change`?
```suggestion
### New Features
```

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -423,14 +444,19 @@ def calculate_checksum(
...
+
                 hash_algo.update(file.encode("utf8"))
+
+                filename = os.path.join(dirname, file)
+                file_size = os.path.getsize(filename)
+
+                if whole_file_hash:
+                    _hash_file(hash_algo, filename, chunk_size, whole_file_hash)
+                    current_size += file_size
                 if current_size < chunk_size:
```
sfc-gh-aling, line 456: I'm curious will this `if current_size < chunk_size:` still meets the condition after the first `if whole_file_hash` meets the condition, in which case `_hash_file` will be executed twice? or is this a valid scenario?

src/snowflake/snowpark/session.py
```diff
@@ -550,7 +550,13 @@ def get_imports(self) -> List[str]:
         """
         return list(self._import_paths.keys())
 
-    def add_import(self, path: str, import_path: Optional[str] = None) -> None:
+    def add_import(
+        self,
+        path: str,
+        import_path: Optional[str] = None,
+        chunk_size: int = 8192,
```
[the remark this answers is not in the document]
sfc-gh-jrose: I think it is somewhat arbitrary as a value that's big enough to hold some amount of hash, but not so big to cause memory issues. It matches what the current default hash size is [here](https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/utils.py#L380).

src/snowflake/snowpark/_internal/utils.py
sfc-gh-jrose: That's a good catch. I believe it should be an `elif`. Hashing the file a second time should break anything, but it also is wasted computation.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -423,14 +444,19 @@ def calculate_checksum(
...
+
+                filename = os.path.join(dirname, file)
+                file_size = os.path.getsize(filename)
+
+                if whole_file_hash:
+                    _hash_file(hash_algo, filename, chunk_size, whole_file_hash)
+                    current_size += file_size
+                elif current_size < chunk_size:
                     read_size = min(file_size, chunk_size - current_size)
                     current_size += read_size
```
sfc-gh-aling, line 458: reading the logic, actually is `current_size` used after set?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -423,14 +444,19 @@ def calculate_checksum(
...
-                    file_size = os.path.getsize(filename)
+
+                filename = os.path.join(dirname, file)
+                file_size = os.path.getsize(filename)
+
+                if whole_file_hash:
+                    _hash_file(hash_algo, filename, chunk_size, whole_file_hash)
+                    current_size += file_size
+                elif current_size < chunk_size:
                     read_size = min(file_size, chunk_size - current_size)
```
sfc-gh-aling, line 457: will this `min` change the behavior? previously `read_size` is always set to `os.path.getsize(filename)` but now `read_size` is `min(file_size, chunk_size - current_size)`

src/snowflake/snowpark/_internal/utils.py
sfc-gh-jrose: Yes, on future iterations `current_size` is used to see if this branch should be entered.

sfc-gh-aling: ahh I see, that code is unexpanded in the review page

sfc-gh-aling (approved):

PR #1181, opened 2023-12-13
SNOW-976702 : Add timeseries agg dataframe function. 

sfc-gh-rsureshbabu:

   [SNOW-976702](https://snowflakecomputing.atlassian.net/browse/SNOW-976702)

The Pr is adding the time series  aggregation function as proposed in https://docs.google.com/document/d/14J5lr_a3fE1xeU-YzKWdeduBdC_035sry11lJUjjkT8/edit

[SNOW-976702]: https://snowflakecomputing.atlassian.net/browse/SNOW-976702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +164,135 @@ def moving_agg(
...
+        col_formatter: Callable[[str, str, int], str] = _default_col_formatter,
+    ) -> "snowflake.snowpark.dataframe.DataFrame":
+        """
+        Applies aggregations to the specified columns of the DataFrame over specified time windows,
+        and grouping criteria.
+
+        Args:
+            aggs: A dictionary where keys are column names and values are lists of the desired aggregation functions.
+            windows: Time windows for aggregations using strings such as '7D' for 7 days, where the units are
+                H: Hours, D: Days, W: Weeks, M: Months, Y: Years. For future-oriented analysis, use positive numbers,
```
sfc-gh-wezhou, line 184: what about minutes/hours/seconds?

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +164,135 @@ def moving_agg(
...
+            sliding_interval, "sliding_interval", allow_negative=False
+        )
+        sliding_point_col = "sliding_point"
+
+        agg_df = self._df
+        agg_df = agg_df.withColumn(
+            sliding_point_col,
+            self._get_sliding_interval_start(time_col, slide_unit, slide_duration),
+        )
+        agg_df.show()
```
sfc-gh-wezhou, line 237: useless?

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +64,43 @@ def _validate_formatter_argument(self, fromatter):
...
+                f"Unsupported unit '{unit}'. Supported units are '{supported_units}"
+            )
+
+        # Converting month unit to 'mm' for Snowpark
+        if unit == "m":
+            unit = "mm"
+
+        return duration, unit
+
+    def _get_sliding_interval_start(self, time_col, unit, duration):
```
sfc-gh-wezhou, line 91: add return type.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +164,135 @@ def moving_agg(
...
+
+        Returns:
+            A Snowflake DataFrame with additional columns corresponding to each specified time window aggregation.
+
+        Raises:
+            ValueError: If an unsupported value is specified in arguments.
+            TypeError: If an unsupported type is specified in arguments.
+            SnowparkSQLException: If an unsupported aggregration is specified.
+
+        Example:
```
sfc-gh-wezhou, line 201: example can be enriched by how the dataframe looks like before and after the transformation.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +164,135 @@ def moving_agg(
...
+        for column, functions in aggs.items():
+            for function in functions:
+                agg_exprs.append((column, function))
+
+        # Perform the aggregation
+        sliding_windows_df = agg_df.groupBy(group_by + [sliding_point_col]).agg(
+            agg_exprs
+        )
+
+        for window in windows:
```
sfc-gh-wezhou, line 248: have you checked the generated SQL and what's the performance? We have another alternative to conduct time-series calculation in our user guide: https://docs.snowflake.com/LIMITEDACCESS/snowflake-feature-store#time-series-features. It's pure SQL. But my gut feeling is it will be faster than how it works here.

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: Hours are supported. Support for seconds and minutes will be added when we have a range between support. The reason for this is performance-related; performing a self-join at seconds or minutes granularity would be a very expensive operation.

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: Yes. I have seen that. Both are same. The SQL is creating CTE to create and join the column. The Python code here also is doing the same.

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-wezhou: hmmm, why? intuitively the # of rows for seconds/minutes is less than hours/days. Also I think worse performance is always better than nothing. We can warn explicitly for such case.

sfc-gh-rsureshbabu: The number of points within the window will be low, but the number of such windows will be high. I am working on a performance benchmark for this API. I will soon have a clearer picture of how bad it could get. If we want to support seconds or minutes, extending the API will be easy.

sfc-gh-rsureshbabu: Added the support for seconds/minutes as well.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +221,125 @@ def moving_agg(
...
+            agg_df, agg_df, aggs, group_by_cols
+        )
+
+        # Perform aggregations at window intervals.
+        result_df = agg_df
+        for window in windows:
+            window_duration, window_unit = self._validate_and_extract_time_unit(
+                window, "window"
+            )
+            print("windows", window_duration, window_unit)
```
sfc-gh-wezhou, line 304: nit: remove?

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +221,125 @@ def moving_agg(
...
+            left_df = sliding_windows_df.alias("A")
+            right_df = sliding_windows_df.alias("B")
+
+            for column in right_df.columns:
+                if column not in group_by:
+                    right_df = right_df.withColumnRenamed(column, f"{column}B")
+
+            self_joined_df = left_df.join(right_df, on=group_by, how="leftouter")
+
+            self_joined_df.show()
```
sfc-gh-wezhou, line 315: remove?

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +221,125 @@ def moving_agg(
...
+        if not sliding_interval:
+            raise ValueError("sliding_interval must not be empty")
+
+        if not time_col or not isinstance(time_col, str):
+            raise ValueError("time_col must be a string")
+
+        slide_duration, slide_unit = self._validate_and_extract_time_unit(
+            sliding_interval, "sliding_interval", allow_negative=False
+        )
+        sliding_point_col = "sliding_point"
```
sfc-gh-wezhou, line 284: nit: name starts with underscore if it's internal column.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +69,95 @@ def _validate_formatter_argument(self, fromatter):
         if not callable(fromatter):
             raise TypeError("formatter must be a callable function")
 
+    def _validate_and_extract_time_unit(
```
sfc-gh-sfan, line 72: This function seems like a good candidate for unit test if you have test coverage issue.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +69,95 @@ def _validate_formatter_argument(self, fromatter):
...
+                f"Unsupported unit '{unit}'. Supported units are '{supported_units}"
+            )
+
+        # Converting month unit to 'mm' for Snowpark
+        if unit == "t":
+            unit = "mm"
+
+        return duration, unit
+
+    def _get_sliding_interval_start(
```
sfc-gh-sfan, line 96: This function seems like a good candidate for unit test if you have test coverage issue.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +221,125 @@ def moving_agg(
...
+        col_formatter: Callable[[str, str, int], str] = _default_col_formatter,
+    ) -> "snowflake.snowpark.dataframe.DataFrame":
+        """
+        Applies aggregations to the specified columns of the DataFrame over specified time windows,
+        and grouping criteria.
+
+        Args:
+            aggs: A dictionary where keys are column names and values are lists of the desired aggregation functions.
+            windows: Time windows for aggregations using strings such as '7D' for 7 days, where the units are
+                S: Seconds, M: Minutes, H: Hours, D: Days, W: Weeks, T: Months, Y: Years. For future-oriented analysis, use positive numbers,
```
sfc-gh-sfan, line 241: Is there a reference for these unit abbreviation? Trying to understand why we use `T` to represent months.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +69,95 @@ def _validate_formatter_argument(self, fromatter):
...
+        if not time_str:
+            raise ValueError(f"{argument_name} must not be empty")
+
+        duration = int(time_str[:-1])
+        unit = time_str[-1].lower()
+
+        if not allow_negative and duration < 0:
+            raise ValueError(f"{argument_name} must not be negative.")
+
+        supported_units = ["s", "m", "h", "d", "w", "t", "y"]
```
sfc-gh-sfan, line 84: This should be a static constant

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +69,95 @@ def _validate_formatter_argument(self, fromatter):
...
+        if unit not in supported_units:
+            raise ValueError(
+                f"Unsupported unit '{unit}'. Supported units are '{supported_units}"
+            )
+
+        # Converting month unit to 'mm' for Snowpark
+        if unit == "t":
+            unit = "mm"
```
sfc-gh-sfan, line 92: I'm not familiar with Snowpark enough, so curious where does snowpark require `mm`?

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +69,95 @@ def _validate_formatter_argument(self, fromatter):
...
+
+        return duration, unit
+
+    def _get_sliding_interval_start(
+        self, time_col: Column, unit: str, duration: int
+    ) -> Column:
+        unit_seconds = {
```
sfc-gh-sfan, line 99: This should be a static constant. Also, it is better to make the calculation explicit. For example: `60 * 60`, `60 * 60 * 24` and `60 * 60 * 24 * 7`

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +69,95 @@ def _validate_formatter_argument(self, fromatter):
...
+            "s": 1,  # seconds
+            "m": 60,  # minutes
+            "h": 3600,  # hours
+            "d": 86400,  # days
+            "w": 604800,  # weeks
+        }
+
+        if unit == "mm":
+            base_date = lit("1970-01-01").cast("date")
```
sfc-gh-sfan, line 108: Do we need to worry about timezone here? Also, this seems like a good static constant.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +69,95 @@ def _validate_formatter_argument(self, fromatter):
...
+            )
+
+        else:
+            raise ValueError(
+                "Invalid unit. Supported units are 'S', 'M', 'H', 'D', 'W', 'T', 'Y'."
+            )
+
+    def _perform_aggregations(
+        self,
+        result_df: "snowflake.snowpark.dataframe.DataFrame",
```
sfc-gh-sfan, line 138: It's unclear to me why we need `result_df` as an input. Should we rename it to `base_df` maybe?

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -57,6 +69,95 @@ def _validate_formatter_argument(self, fromatter):
...
+            for func in funcs:
+                agg_column_name = (
+                    col_formatter(column, func, window)
+                    if col_formatter
+                    else f"{column}_{func}{rename_suffix}"
+                )
+                agg_expression = expr(f"{func}({column}{rename_suffix})").alias(
+                    agg_column_name
+                )
+                agg_df = input_df.groupBy(group_by_cols).agg(agg_expression)
```
sfc-gh-sfan, line 156: nit: Let's use `group_by` to be more pythonic.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +221,125 @@ def moving_agg(
...
+
+        Returns:
+            A Snowflake DataFrame with additional columns corresponding to each specified time window aggregation.
+
+        Raises:
+            ValueError: If an unsupported value is specified in arguments.
+            TypeError: If an unsupported type is specified in arguments.
+            SnowparkSQLException: If an unsupported aggregration is specified.
+
+        Example:
```
sfc-gh-sfan, line 257: Let's make this example a doc test.

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +221,125 @@ def moving_agg(
...
+        if not time_col or not isinstance(time_col, str):
+            raise ValueError("time_col must be a string")
+
+        slide_duration, slide_unit = self._validate_and_extract_time_unit(
+            sliding_interval, "sliding_interval", allow_negative=False
+        )
+        sliding_point_col = "sliding_point"
+
+        agg_df = self._df
+        agg_df = agg_df.withColumn(
```
sfc-gh-sfan, line 287: nit: `with_column`

src/snowflake/snowpark/dataframe_transform_functions.py
```diff
@@ -120,3 +221,125 @@ def moving_agg(
...
+                window, "window"
+            )
+            print("windows", window_duration, window_unit)
+            # Perform self-join on DataFrame for aggregation within each group and time window.
+            left_df = sliding_windows_df.alias("A")
+            right_df = sliding_windows_df.alias("B")
+
+            for column in right_df.columns:
+                if column not in group_by:
+                    right_df = right_df.withColumnRenamed(column, f"{column}B")
```
sfc-gh-sfan, line 311: nit: `with_column_renamed`

sfc-gh-yixie:
Code coverage of the new file is 80%. We usually try to cover as many lines as possible.

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: Added docstest

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: It's not an internal column. This column will be part of the output.

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: Snowflake has used "mm" for months. Instead of accepted 2 characters, we are accepting single character "T"

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: Thats the convention snowflake follows :)

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: These lines already covered in the existing tests.

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: These lines are already covered in the existing tests

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-sfan: I don't see a good reason that we want to use `T` then. Why don't we follow Snowflake's convention?

src/snowflake/snowpark/dataframe_analytics_functions.py
```diff
@@ -21,7 +21,7 @@
 from snowflake.snowpark.window import Window
 
 # "s" (seconds), "m" (minutes), "h" (hours), "d" (days), "w" (weeks), "t" (months), "y" (years)
-SUPPORTED_TIME_UNITS = ["s", "m", "h", "d", "w", "t", "y"]
+SUPPORTED_TIME_UNITS = ["s", "m", "h", "d", "w", "mm", "y"]
```
sfc-gh-aalam, line 24: will this change break older customers as "t" is no longer supported?
also, nit: update comment above

sfc-gh-rsureshbabu: I think you are looking into the wrong diff. This line was added in this PR.

src/snowflake/snowpark/dataframe_transform_functions.py
sfc-gh-rsureshbabu: Discussed offline. Switched it to snowflake convention.

sfc-gh-wezhou (approved):

src/snowflake/snowpark/dataframe_analytics_functions.py
sfc-gh-aalam: yeah. I got redirected to single commit diff.

src/snowflake/snowpark/dataframe_analytics_functions.py
```diff
@@ -248,3 +371,144 @@ def cumulative_agg(
...
+            raise ValueError("sliding_interval must not be empty")
+
+        if not time_col or not isinstance(time_col, str):
+            raise ValueError("time_col must be a string")
+
+        slide_duration, slide_unit = self._validate_and_extract_time_unit(
+            sliding_interval, "sliding_interval", allow_negative=False
+        )
+        sliding_point_col = "sliding_point"
+
```
sfc-gh-aalam, line 457: would this overwrite sliding_point if original df has this column? I would suggest using randomized name - see `random_table_name` in `utils.py`

tests/integ/test_df_analytics.py
```diff
@@ -310,3 +312,255 @@ def custom_formatter(input_col, agg):
...
+
+
+@pytest.mark.skipif(not is_pandas_available, reason="pandas is required")
+def test_time_series_agg(session):
+    """Tests time_series_agg_fixed function with various window sizes."""
+
+    df = get_sample_dataframe(session)
+    df = df.withColumn("ORDERDATE", to_timestamp(df["ORDERDATE"]))
+    # session.sql_simplifier_enabled = False
+
```
sfc-gh-aalam, line 324: nit: remove redundant comment

tests/integ/test_df_analytics.py
```diff
@@ -310,3 +312,255 @@ def custom_formatter(input_col, agg):
...
+        ["2023-01-20", 102, 150],
+        ["2023-02-20", 102, 250],
+        ["2023-03-20", 102, 350],
+        ["2023-04-20", 102, 450],
+    ]
+    df = session.create_dataframe(data).to_df("ORDERDATE", "PRODUCTKEY", "SALESAMOUNT")
+
+    df = df.withColumn("ORDERDATE", to_timestamp(df["ORDERDATE"]))
+    # session.sql_simplifier_enabled = False
+
```
sfc-gh-aalam, line 381: same

src/snowflake/snowpark/dataframe_analytics_functions.py
sfc-gh-rsureshbabu: This is not a temporary column. I am not sure about generating a random name for an output column. This will make it hard to write queries based on this column. We could put a contact suffix/prefix to the name. That can bring down the probability of collision. Say "sp_gen_sliding_point". wdyt ?

sfc-gh-aalam (approved):

PR #1184, opened 2023-12-14
[Mypy] Add dataframe.py to mypy coverage

sfc-gh-jrose:

   Fixes #NNNN

 There are a lot of places in this file where attributes or variables are passed that don't have strong typing so I've added quite a few exceptions. In the future it might be useful to refactor some of our code to use Protocols instead of Unions of various types. Propagating the unions through all functions gets very complex, this is especially true for things like `SelectStatement` vs `MockSelectStatement` or `Expression` vs `NamedExpression`.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -50,6 +50,7 @@ from snowflake.snowpark._internal.analyzer.expression import (
     Attribute,
     Expression,
     FunctionExpression,
+    NamedExpression,
     Star,
     UnresolvedAttribute,
     derive_dependent_columns,
@@ -575,7 +576,9 @@ class SelectStatement(Selectable):
             return new
         return self
 
-    def select(self, cols: List[Expression]) -> "SelectStatement":
+    def select(
+        self, cols: Sequence[Union[Expression, NamedExpression]]
+    ) -> "SelectStatement":
         """Build a new query. This SelectStatement will be the subquery of the new query.
         Possibly flatten the new query and the subquery (self) to form a new flattened query.
         """
@@ -604,7 +607,7 @@ class SelectStatement(Selectable):
             new.flatten_disabled = self.flatten_disabled
             return new
         disable_next_level_flatten = False
-        new_column_states = derive_column_states_from_subquery(cols, self)
+        new_column_states = derive_column_states_from_subquery(cols, self)  # type: ignore[arg-type]  # TODO:SNOW-1043025
         if new_column_states is None:
             can_be_flattened = False
             disable_next_level_flatten = True
@@ -625,7 +628,7 @@ class SelectStatement(Selectable):
             # TODO: Clean up, this entire if case is parameter protection
             can_be_flattened = False
         elif (self.where or self.order_by or self.limit_) and has_data_generator_exp(
-            cols
+            cols    # type: ignore[arg-type]  # TODO:SNOW-1043025
         ):
             can_be_flattened = False
         elif self.where and (
@@ -678,7 +681,7 @@ class SelectStatement(Selectable):
             new.post_actions = new.from_.post_actions
         else:
             new = SelectStatement(
-                projection=cols, from_=self.to_subqueryable(), analyzer=self.analyzer
+                projection=cols, from_=self.to_subqueryable(), analyzer=self.analyzer  # type: ignore[arg-type]  # TODO:SNOW-1043025
             )
         new.flatten_disabled = disable_next_level_flatten
         assert new.projection is not None
@@ -713,7 +716,7 @@ class SelectStatement(Selectable):
 
         return new
 
-    def sort(self, cols: List[Expression]) -> "SelectStatement":
+    def sort(self, cols: Sequence[Expression]) -> "SelectStatement":
         can_be_flattened = (
             (not self.flatten_disabled)
             and can_clause_dependent_columns_flatten(
@@ -721,17 +724,18 @@ class SelectStatement(Selectable):
             )
             and not has_data_generator_exp(self.projection)
         )
+        cols_list = list(cols)
         if can_be_flattened:
             new = copy(self)
             new.from_ = self.from_.to_subqueryable()
             new.pre_actions = new.from_.pre_actions
             new.post_actions = new.from_.post_actions
-            new.order_by = cols + (self.order_by or [])
+            new.order_by = cols_list + (self.order_by or [])
             new.column_states = self.column_states
         else:
             new = SelectStatement(
                 from_=self.to_subqueryable(),
-                order_by=cols,
+                order_by=cols_list,
                 analyzer=self.analyzer,
             )
         return new
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -1,6 +1,7 @@
 #
 # Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
 #
+from collections import defaultdict
 import collections.abc
 import io
 import os
@@ -1164,7 +1165,7 @@ def generate_call_python_sp_sql(
     sql_args = []
     for arg in args:
         if isinstance(arg, snowflake.snowpark.Column):
-            sql_args.append(session._analyzer.analyze(arg._expression, {}))  # type: ignore[attr-defined]
+            sql_args.append(session._analyzer.analyze(arg._expression, defaultdict()))
         else:
             sql_args.append(to_sql(arg, infer_type(arg)))
-    return f"CALL {sproc_name}({', '.join(sql_args)})"
+    return f"CALL {sproc_name}({', '.join(sql_args)})"  # type: ignore[arg-type]  #TODO: SNOW-1019751
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -371,7 +371,7 @@ InputType = TypeVar("InputType")
 
 
 def parse_positional_args_to_list(
-    *inputs: Union[InputType, Iterable[InputType]]
+    *inputs: InputType
 ) -> List[InputType]:
     """Convert the positional arguments to a list."""
     if len(inputs) == 1:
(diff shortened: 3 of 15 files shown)


tox.ini
```diff
@@ -132,7 +132,7 @@ commands = pyright src/snowflake/snowpark/_internal/analyzer
 [testenv:mypy]
 description = static type checking with mypy, configuration aligned with python connector
 deps = mypy
-commands = mypy --config-file mypy.ini src/snowflake/snowpark/_internal/telemetry.py src/snowflake/snowpark/_internal/type_utils.py src/snowflake/snowpark/_internal/udf_utils.py src/snowflake/snowpark/_internal/utils.py
```
sfc-gh-stan, line 135: Is there a reason why we are removing these files from the check?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1164,7 +1165,7 @@ def generate_call_python_sp_sql(
     sql_args = []
     for arg in args:
         if isinstance(arg, snowflake.snowpark.Column):
-            sql_args.append(session._analyzer.analyze(arg._expression, {}))  # type: ignore[attr-defined]
+            sql_args.append(session._analyzer.analyze(arg._expression, defaultdict()))
```
sfc-gh-yixie, line 1168: This is not a equivalent change. Is it possible to keep {}?

src/snowflake/snowpark/dataframe.py
```diff
@@ -239,8 +244,8 @@ def _disambiguate(
         # We use the session of the LHS DataFrame to report this telemetry
         lhs._session._conn._telemetry_client.send_alias_in_join_telemetry()
 
-    lsuffix = lsuffix or lhs._alias
-    rsuffix = rsuffix or rhs._alias
+    lsuffix = lsuffix or lhs._alias or ""
+    rsuffix = rsuffix or rhs._alias or ""
```
sfc-gh-yixie, line 248: Not equivalent change

src/snowflake/snowpark/dataframe.py
```diff
@@ -522,18 +529,19 @@ def __init__(
...
 
+        # mypy does not like function forwarding like below. Disabling warnings for now.
         self._stat = DataFrameStatFunctions(self)
         self._analytics = DataFrameAnalyticsFunctions(self)
-        self.approxQuantile = self.approx_quantile = self._stat.approx_quantile
-        self.corr = self._stat.corr
-        self.cov = self._stat.cov
-        self.crosstab = self._stat.crosstab
-        self.sampleBy = self.sample_by = self._stat.sample_by
+        self.approxQuantile = self.approx_quantile = self._stat.approx_quantile  # type: ignore[misc, method-assign]
```
sfc-gh-yixie, line 535: Does the type:ignore harm the code assist in IDE?

src/snowflake/snowpark/dataframe.py
```diff
@@ -887,7 +894,7 @@ def to_pandas_batches(
         )
 
     @df_api_usage
-    def to_df(self, *names: Union[str, Iterable[str]]) -> "DataFrame":
+    def to_df(self, *names: str) -> "DataFrame":
```
sfc-gh-yixie, line 897: This does accept `Iterable[str]`, for instance `["a", "b"]`.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1099,7 +1107,7 @@ def select(
         return self._with_plan(Project(names, join_plan or self._plan))
 
     @df_api_usage
-    def select_expr(self, *exprs: Union[str, Iterable[str]]) -> "DataFrame":
```
sfc-gh-yixie, line 1102: Same here

src/snowflake/snowpark/dataframe.py
```diff
@@ -1693,14 +1699,15 @@ def unpivot(
...
+        column_exprs = self._convert_cols_to_exprs("unpivot()", *column_list)
         unpivot_plan = Unpivot(value_column, name_column, column_exprs, self._plan)
 
         if self._select_statement:
+            analyzer = self._session._analyzer
             return self._with_plan(
-                SelectStatement(
-                    from_=SelectSnowflakePlan(
-                        unpivot_plan, analyzer=self._session._analyzer
+                analyzer.create_select_statement(
```
sfc-gh-yixie, line 1708: I don't suggest this change in this mypy PR. Is it required to pass mypy?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2333,10 +2350,10 @@ def join_table_function(
                 TableFunctionJoin(self._plan, func_expr)
             )
             old_cols, new_cols, alias_cols = _get_cols_after_join_table(
-                func_expr, self._plan, temp_join_plan
+                func_expr, self._plan, temp_join_plan  # type: ignore[arg-type]  # TODO: SNOW-1038774
             )
             new_col_names = [
-                self._session._analyzer.analyze(col, {}) for col in new_cols
+                self._session._analyzer.analyze(col, defaultdict()) for col in new_cols
```
sfc-gh-yixie, line 2356: Not equivalent change

src/snowflake/snowpark/dataframe.py
```diff
@@ -2197,7 +2214,7 @@ def join(
             elif isinstance(using_columns, Column):
                 using_columns = using_columns
             elif (
-                isinstance(using_columns, Iterable)
+                isinstance(using_columns, Sequence)
```
sfc-gh-yixie, line 2217: Not equivalent change. Something can be Iterable but not Sequence.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-stan: This is a bug uncovered by mypy, `defaultdict()` is the correct fix. Also this change only affects DataFrame alias, which is an experimental feature.

sfc-gh-yixie: It seems that `defaultdict()` has the same effect as `{}`? The following raises an KeyError.
```
d = defaultdict()
print(d["a"])
```
`defaultdict` accepts a datatype for the default value. Is it needed?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -27,12 +27,14 @@
     Any,
     Callable,
     Dict,
+    Iterable,
```
sfc-gh-aling, line 30: looks like this import is not used

src/snowflake/snowpark/dataframe.py
```diff
@@ -611,13 +615,17 @@ def collect_nowait(
...
         """
-        return self._internal_collect_with_tag_no_telemetry(
+        result = self._internal_collect_with_tag_no_telemetry(
             statement_params=statement_params,
             block=False,
             data_type=_AsyncResultType.ROW,
             log_on_exception=log_on_exception,
             case_sensitive=case_sensitive,
         )
+        assert isinstance(
```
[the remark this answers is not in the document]
sfc-gh-aling: 1. I think we should avoid using assertion error in production in general, if we think the check is really needed, we should raise a proper exception explaining to users what happened.

assertion error can be silenced with `-O` option when running python:
```
# some.py
print('123')
assertion True is False
```
if you run with `-O`, assertion error won't be raised:
```bash
python -O some.py
```

there are lots of discussion around assertion best practices, whether to do it in production or not, you can search around first.

2. and my second concern is can raising error be a breaking change here? what's the current behavior if it's not the instance of AsyncJob

src/snowflake/snowpark/dataframe.py
```diff
@@ -1124,16 +1132,15 @@ def select_expr(self, *exprs: Union[str, Iterable[str]]) -> "DataFrame":
...
+        parsed_exprs: List[str] = parse_positional_args_to_list(*exprs)
+        return self.select([sql_expr(expr) for expr in parsed_exprs])
 
     selectExpr = select_expr
 
     @df_api_usage
     def drop(
         self,
-        *cols: Union[ColumnOrName, Iterable[ColumnOrName]],
+        *cols: ColumnOrName,
```
sfc-gh-aling, line 1143: just curious about why we remove the Iterable for the *args parameter, is this mypy standard?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2177,15 +2192,17 @@ def join(
...
+        ), "Value specified for keyword parameter `join_type needs` to be of str type"
         if isinstance(right, DataFrame):
             if self is right or self._plan is right._plan:
                 raise SnowparkClientExceptionMessages.DF_SELF_JOIN_NOT_SUPPORTED()
 
-            if isinstance(join_type, Cross) or (
-                isinstance(join_type, str)
-                and join_type.strip().lower().replace("_", "").startswith("cross")
-            ):
+            if isinstance(join_type, str) and join_type.strip().lower().replace(
```
sfc-gh-aling, line 2203: is removing the `isinstance(join_type, Cross)` part on purpose?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2218,7 +2235,7 @@ def join(
             return self._join_dataframes(
                 right,
                 using_columns,
-                create_join_type(join_type or "inner"),
```
sfc-gh-aling, line 2221: why we remove the this, can join_type be None or empty string?

src/snowflake/snowpark/mock/_select_statement.py
```diff
@@ -192,6 +192,7 @@ def __init__(
         self.post_actions = self.from_.post_actions
         self._sql_query = None
         self._schema_query = None
+        self._query_params: Sequence[Any] = []
```
sfc-gh-aling, line 195: is this fixing a bug?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -713,25 +716,26 @@ def filter(self, col: Expression) -> "SelectStatement":
 
         return new
 
-    def sort(self, cols: List[Expression]) -> "SelectStatement":
+    def sort(self, cols: Sequence[Expression]) -> "SelectStatement":
```
sfc-gh-yzou, line 719: @sfc-gh-stan i see your pain here, mypy is converting List -> list, Dict -> dict, but you can not use list or dict for 3.8 at this moment. This seems start changing your signature, not sure if it is a good thing or not.
so one way to skip the check here is using typing.List, or typing.Dict, and we can start change it  to list or dict after 3.8 is deprecated.
i haven't figure out a way to ask mypy to skip the check yet

PR #1186, opened 2023-12-19
SNOW-993761: Add statement params to anonymous sproc

sfc-gh-thoyt:

N/A (JIRA: SNOW-993761)

Exposes statement_params for telemetry in `StoredProcedure.__call__` for use with snowpark-ml.

### DIFF
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -1342,11 +1342,7 @@ def test_stored_procedure_call_with_statement_params(session, anonymous):
     )
     if anonymous:
         assert add_sp._anonymous_sp_sql is not None
-    telemetry_log_length = len(session._conn._telemetry_client.telemetry._log_batch)
     assert add_sp(1, 2, statement_params=statement_params) == 3
-    assert (
-        len(session._conn._telemetry_client.telemetry._log_batch) > telemetry_log_length
-    )
 
 
 @pytest.mark.skipif(IS_NOT_ON_GITHUB, reason="need resources")
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -1331,6 +1331,24 @@ def test_anonymous_stored_procedure(session):
     assert add_sp(1, 2) == 3
 
 
+@pytest.mark.parametrize("anonymous", [True, False])
+def test_stored_procedure_call_with_statement_params(session, anonymous):
+    statement_params = {"test": "params"}
+    add_sp = session.sproc.register(
+        lambda session_, x, y: session_.sql(f"SELECT {x} + {y}").collect()[0][0],
+        return_type=IntegerType(),
+        input_types=[IntegerType(), IntegerType()],
+        anonymous=anonymous,
+    )
+    if anonymous:
+        assert add_sp._anonymous_sp_sql is not None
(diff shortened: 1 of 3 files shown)


sfc-gh-thoyt:

sfc-gh-jrose (commented):
Are there any examples or tests that should be updated to show usage of the exposed functionality?

sfc-gh-thoyt:
@sfc-gh-jrose Thanks for reviewing. I added a test confirming telemetry is sent for both anonymous and non-anonymous sproc calls.

tests/integ/test_stored_procedure.py
```diff
@@ -1331,6 +1331,24 @@ def test_anonymous_stored_procedure(session):
...
+        input_types=[IntegerType(), IntegerType()],
+        anonymous=anonymous,
+    )
+    if anonymous:
+        assert add_sp._anonymous_sp_sql is not None
+    telemetry_log_length = len(session._conn._telemetry_client.telemetry._log_batch)
+    assert add_sp(1, 2, statement_params=statement_params) == 3
+    assert (
+        len(session._conn._telemetry_client.telemetry._log_batch) > telemetry_log_length
+    )
```
sfc-gh-aalam, line 1349: can you refer to `test_call_with_statement_params` in `tests/integ/test_dataframe.py` for a better way to test statement_params

sfc-gh-thoyt: The dataframe statement params tested in `test_call_with_statement_params` have a different code path with date format validation. The stored proc statement params do not have the same validation.

sfc-gh-aalam: depending of telemetry log length will create flaky test. We can still use a similar approach as in `test_call_with_statement_params`

create an anonymous sproc that return a date like this
```
lambda session_: session_.sql(f"select column1::DATE from values ('01-01-2023')").collect()[0][0],
...
...
```

with `statement_params = {"DATE_INPUT_FORMAT": "MM-DD-YYYY"}` expect correct results.

with `statement_params = {"DATE_INPUT_FORMAT": "MXM-DXD-YYYY"}` expect failure.

sfc-gh-thoyt: I understand what you mean now - I wasn't previously familiar with statement params being used in this way, only in the context of telemetry.

The issue with your example is that it doesn't appear that the statement params propagate to internal sproc logic. `select column1::DATE from values ('01-01-2023')` will always raise an exception regardless of the statement params used when `__call__`ing the sproc.

Is there another parameter we could use to confirm that the statement params are being respected?

sfc-gh-aalam: I misunderstood how statements_params are used. It's not supposed to permeate into the sproc environment. Your current test is okay. I would just remove the telemetry test because it is likely to be flaky (we flush the log non-deterministically so length might reset during the test.) Let's follow the lead of `test_session_register_sp` which is essentially making sure we can pass this param. We can test this against a test deployment and check to query history to make sure that the statement params are seen.

sfc-gh-thoyt: Thanks @sfc-gh-aalam  - I removed the telemetry test.

sfc-gh-jrose (approved):

PR #1188, opened 2023-12-29
SNOW-940607: use precision value to decide when to use astype(int64)

sfc-gh-aalam:

   Fixes #SNOW-940607

   In this PR, we update `_fix_pandas_df_integer` to use `astype('int64')` for cases where the precision value sent by GS is greater than 10. This is done to fix the loss of precision when using `pandas.to_numeric` for long numbers.

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -681,12 +681,16 @@ def _fix_pandas_df_integer(
         if (
             FIELD_ID_TO_NAME.get(column_metadata.type_code) == "FIXED"
             and column_metadata.precision is not None
-            and column_metadata.precision > 10
             and column_metadata.scale == 0
             and not str(pandas_dtype).startswith("int")
         ):
             try:
-                pd_df[pandas_col_name] = pd_df[pandas_col_name].astype("int64")
+                if column_metadata.precision > 10:
+                    pd_df[pandas_col_name] = pd_df[pandas_col_name].astype("int64")
+                else:
+                    pd_df[pandas_col_name] = pandas.to_numeric(
+                        pd_df[pandas_col_name], downcast="integer"
+                    )
             except OverflowError:
                 pd_df[pandas_col_name] = pandas.to_numeric(
                     pd_df[pandas_col_name], downcast="integer"
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -685,7 +685,12 @@ def _fix_pandas_df_integer(
             and not str(pandas_dtype).startswith("int")
         ):
             try:
-                if column_metadata.precision > 10:
+                # When scale = 0 and precision values are between 10-20, the integers fit into int64.
+                # If we rely only on pandas.to_numeric, it loses precision value on large integers, therefore
+                # we try to strictly use astype("int64") in this scenario. If the values are too large to
+                # fit in int64, an OverflowError is thrown and we rely on to_numeric to choose and appropriate
+                # floating datatype to represent the number.
+                if column_metadata.precision > 10 and column_metadata.precision < 20:
                     pd_df[pandas_col_name] = pd_df[pandas_col_name].astype("int64")
                 else:
                     pd_df[pandas_col_name] = pandas.to_numeric(
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -8,7 +8,6 @@ import inspect
 import os
 import sys
 import time
-import warnings
 from logging import getLogger
 from typing import (
     IO,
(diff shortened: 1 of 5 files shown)


sfc-gh-smirzaei (approved):

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -685,25 +684,16 @@ def _fix_pandas_df_integer(
             and column_metadata.scale == 0
             and not str(pandas_dtype).startswith("int")
         ):
-            # pandas.to_numeric raises "RuntimeWarning: invalid value encountered in cast"
-            # when it tried to downcast and loses precision. In this case, we try to convert to
-            # int64 using astype() and fallback to to_numeric if we were unsuccessful.
-            with warnings.catch_warnings(record=True) as warning_list:
-                pd_col_with_numeric_downcast = pandas.to_numeric(
+            try:
+                if column_metadata.precision > 10:
```
sfc-gh-aling, line 688: out of curiosity, why is number 10, is this documented somewhere?

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -614,9 +614,12 @@ def _fix_pandas_df_integer(table_res: TableEmulator) -> "pandas.DataFrame":
             else:
                 pd_df[pd_df_col_name] = table_res[col_name].astype("int64")
         elif isinstance(col_sf_type.datatype, _IntegralType):
-            pd_df[pd_df_col_name] = pandas.to_numeric(
-                table_res[col_name].tolist(), downcast="integer"
-            )
+            try:
+                pd_df[pd_df_col_name] = table_res[col_name].astype("int64")
```
sfc-gh-aling, line 618: in the mocking, I think we also need to check precision 10 to align with the live behavior

tests/integ/test_df_to_pandas.py
```diff
@@ -77,8 +77,8 @@ def test_to_pandas_cast_integer(session, to_pandas_api, local_testing_mode):
     assert str(pandas_df.dtypes[2]) == "int32"
     assert str(pandas_df.dtypes[3]) == "int64"
     assert (
-        str(pandas_df.dtypes[4]) == "int8"
-    )  # When static type can possibly be greater than int64 max, use the actual value to infer the int type.
+        str(pandas_df.dtypes[4]) == "int64"
```
sfc-gh-aling, line 80: is this considered a BCR?

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: given that scale=0, precision greater than 10 implies that the number of digits in the number will be greater than 10. max number of digits in int32 is equal to 10 therefore we only convert to int64 when precision info by GS is greater than 10. If we keep using pandas to_numeric for large numbers, it loses precision for larger numbers.

tests/integ/test_df_to_pandas.py
sfc-gh-aalam: yeah, it is

src/snowflake/snowpark/mock/_connection.py
sfc-gh-aalam: The precision check is done when datatype is an instance of `DecimalType`. For `_IntegralType` we don't have precision value. By default, Integral types are NUMBER(38,0) so they fall in `int64` domain.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aling: aha I see, this is knowledgeable, can we document this in the code comment?

tests/integ/test_df_to_pandas.py
sfc-gh-aling: then we will have this put under the "BCR" section in the changelog?

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-yixie: How about when the precision is great than int64?

tests/integ/test_df_to_pandas.py
sfc-gh-aalam: you are right. I'll update that.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: If the number cannot fit in int64 - it will throw an OverflowError and we will resort to old behavior to `to_numeric`. In this case, we will have float types instead of integer types. I guess I could tighten the condition and say when precision > 10 and < 20.

```
>>> df = pd.DataFrame({'A': [1, '8'*25]})
>>> df['A'].astype('int64')
  File "<input>", line 1, in <module>
    df['A'].astype('int64')
.
.
.
OverflowError: Python int too large to convert to C long
```

sfc-gh-aalam: Update here: We can't have the <20 condition here because GS would not guarantee such tight bounds here.

The use-case for this function is to correct mislabeled integer types back to integers so we should ideally not be having overflows. In case of overflow however, pandas.to_numeric will convert incoming data into a float type and we will move back to the old behavior.

sfc-gh-yixie (approved):

PR #1218, opened 2024-01-26
SNOW-1016694: Add support for current_timestamp to local testing

sfc-gh-jrose:

   Fixes #1016694

This change adds local testing support for the following functions:

- current_timestamp
- current_date
- current_time

It also refactors the related test to verify that the various components of those times are approximately equal when called.

For some reason pylint also updated a bunch of unrelated lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@
   - Added the `moving_agg` function in `DataFrame.analytics` to enable moving aggregations like sums and averages with multiple window sizes.
   - Added the `cummulative_agg` function in `DataFrame.analytics` to enable commulative aggregations like sums and averages on multiple columns.
   - Added the `compute_lag` and `compute_lead` function in `DataFrame.analytics` for enabling lead and lag calculations on multiple columns.
+- Added support for the following local testing functions:
+  - current_timestamp
+  - current_date
+  - current_time
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## 1.13.0 (TBD)
 
+### New Features
+
+- Added support for the following local testing functions:
+  - current_timestamp
+  - current_date
+  - current_time
+
 ### Behavior Changes (API Compatible)
 
 - Added support for an optional `date_part` argument in function `last_day`
@@ -26,10 +33,6 @@
   - Added the `moving_agg` function in `DataFrame.analytics` to enable moving aggregations like sums and averages with multiple window sizes.
   - Added the `cummulative_agg` function in `DataFrame.analytics` to enable commulative aggregations like sums and averages on multiple columns.
   - Added the `compute_lag` and `compute_lead` function in `DataFrame.analytics` for enabling lead and lag calculations on multiple columns.
-- Added support for the following local testing functions:
-  - current_timestamp
-  - current_date
-  - current_time
 
 ### Bug Fixes
 
--- a/src/snowflake/snowpark/mock/_functions.py
+++ b/src/snowflake/snowpark/mock/_functions.py
@@ -29,6 +29,7 @@ from snowflake.snowpark.types import (
     MapType,
     NullType,
     StringType,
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
   - `array_except`
   - `create_map`
   - `sign`/`signum`
+- Added support for the following local testing functions:
```
sfc-gh-stan, line 19: This should be under a new release.

tests/integ/test_function.py
```diff
@@ -207,11 +207,20 @@ def test_order(session):
 
 
 def test_current_date_and_time(session):
```
sfc-gh-stan, line 209: Let's mark these with `@pytest.mark.localtest`

tests/integ/test_function.py
```diff
@@ -207,11 +207,20 @@ def test_order(session):
...
-    df1 = session.sql("select current_date(), current_time(), current_timestamp()")
-    df2 = session.create_dataframe([1]).select(
+    max_delta = 5
```
sfc-gh-stan, line 210: Can this be a smaller number?

sfc-gh-stan (changes requested):

sfc-gh-stan (approved):

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -308,6 +309,26 @@ def mock_to_date(
...
+@patch("current_date")
+def mock_current_date():
+    now = datetime.datetime.now()
+    return ColumnEmulator(data=now.date(), sf_type=ColumnType(DateType(), False))
+
+
+@patch("current_time")
+def mock_current_time():
+    now = datetime.datetime.now()
+    return ColumnEmulator(data=now.time(), sf_type=ColumnType(TimeType(), False))
```
sfc-gh-aling, line 329: do we need to take timezone into account, I'm wondering if is there any session parameter that could change the returning value?

tests/integ/test_function.py
```diff
@@ -206,12 +206,22 @@ def test_order(session):
...
+    rows = df.collect()
+
+    assert len(rows) == 1, "df1 should only contain 1 row"
+    date, time, timestamp = rows[0]
+    time1 = datetime.datetime.combine(date, time).timestamp()
+    time2 = timestamp.timestamp()
+
+    assert time1 == pytest.approx(
+        time2, max_delta
+    ), f"Times should be within {max_delta} seconds of each other."
```
sfc-gh-aling, line 224: why do we have this test here? I'm wondering if this would introduce test flakiness in live mode.

sfc-gh-jrose: This seemed like the easiest way to get two times that should be close to each other. Testing an exact value would be ideal, but since we don't want to mock out the server response the next best thing would be to have a fuzzy equality. 

If this causes flakiness we can relax the allowed delta. If we relaxed the delta to our server response SLA a failure would not be due to flakiness, but due to a larger issue in my opinion.

src/snowflake/snowpark/mock/_functions.py
sfc-gh-jrose: Timezone does affect the result, but I don't think we have a good mechanism for overriding session parameters in local testing yet and creating that mechanism is outside of the scope for this change.

sfc-gh-aling: yeah, I also don't think it's an issue to be addressed now.
can you create a JIRA for this timezone session parameter and put it into our backlog epic?

tests/integ/test_function.py
sfc-gh-aling: in this case, should we also test `timestamp` with `time2`?

sfc-gh-aling: gotcha, but I'm still wondering whether it's needed to compare time1 with time2..

previously the test is comparing the results from raw sql execution with from df.select function -- the same results expected to be the same because after union, the new df is supposed to only have one row.

we can keep the test here, it does no harm

PR #1226, opened 2024-01-30
SNOW-869536 Fix buggy behavior in DataFrame.to_local_iterator 

### LINKED ISSUE #945: SNOW-869536: Iterators from to_local_iterator stop returning results after another query occurs
Please answer these questions before submitting your issue. Thanks!

`Python 3.10.8 (main, Oct 13 2022, 09:48:40) [Clang 14.0.0 (clang-1400.0.29.102)]`

`macOS-13.3.1-arm64-arm-64bit`

... (Snowpark 1.5.1)

4. What did you do?

We've been using an iterator from `to_local_iterator()`, and also using the table's schema to parse it.

```python
# Note: df.schema must not be called before this for bug to be recreated
df = session.table("some_table")
my_iter = df.to_local_iterator()
counter = 0
for row in my_iter:
   len(df.schema.fields)  # Here was our parsing logic which used df.schema, this is good enough to recreate the bug
   counter += 1
print(counter)   # 1 is printed - we only iterated over the first row
```

We expected to iterate over all the rows, and we only iterated over the first one.
Calling `df.schema` had probably caused the python snowflake connector to execute another query, making `cursor.execute()` no longer point to our query and rendering the iterator useless.
This probably means that generally, other queries cannot be run while iterating.
Note that there's an easy workaround, using AsyncJobs - which makes the iterator query specifically for our query-id, and thus is still stable even while other queries are running:

```python
my_iter = df.to_local_iterator(block=False).result()
counter = 0
for row in my_iter:
   len(df.schema.fields)
   counter += 1
print(counter)  # Prints the length of the table
```

6. Can you set logging to DEBUG and collect the logs?

Hard to do with our current environment :(

sfc-gh-stan:

   Fixes #945 

   **Issue**: All queries issued from a Snowpark `Session` object are executed using the same SnowflakeCursor instance, this causes unexpected, wrong behavior when we fetch results from an iterator but have code that executes a different query half way, e.g.
   ```
    df = session.create_dataframe([[1,2,3],[4,5,6]], schema=["a", "b", "c"]) 
    my_iter = df.to_local_iterator()
    row_counter
    for row in my_iter:
       len(df.schema.fields)  # this executes a schema query, which overwrites properties of the cursor object
       row_counter += 1
    print(counter)  # this prints 1 since schema query only returns 1 row, but should be 2
   ```
   
   **Fix**: This PR changes `src/snowflake/snowpark/_internal/server_connection.py` create a new cursor object locally to read data from the last executed query and continues to use this cursor object in the iterator to achieve isolation between the iterator and the following queries executed within the session.

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -435,7 +435,7 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
     ) -> Dict[str, Any]:
-        if to_iter:  # Fix for SNOW-869536
+        if to_iter and not to_pandas:  # Fix for SNOW-869536
             new_cursor = results_cursor.connection.cursor()
             new_cursor.execute(
                 f"SELECT * FROM TABLE(RESULT_SCAN('{results_cursor.sfqid}'))"
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -434,8 +434,14 @@ class ServerConnection:
         results_cursor: SnowflakeCursor,
         to_pandas: bool = False,
         to_iter: bool = False,
-        num_statements: Optional[int] = None,
     ) -> Dict[str, Any]:
+        if to_iter:  # Fix for SNOW-869536
+            new_cursor = results_cursor.connection.cursor()
+            new_cursor.execute(
+                f"SELECT * FROM TABLE(RESULT_SCAN('{results_cursor.sfqid}'))"
+            )
+            results_cursor = new_cursor
+
         if to_pandas:
             try:
                 data_or_iter = (
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -435,7 +435,9 @@ class ServerConnection:
         to_pandas: bool = False,
         to_iter: bool = False,
     ) -> Dict[str, Any]:
-        if to_iter and not to_pandas:  # Fix for SNOW-869536
+        if (
+            to_iter and not to_pandas
+        ):  # Fix for SNOW-869536, to_pandas doesn't have this issue, SnowflakeCursor.fetch_pandas_batches already handles the isolation.
             new_cursor = results_cursor.connection.cursor()
             new_cursor.execute(
                 f"SELECT * FROM TABLE(RESULT_SCAN('{results_cursor.sfqid}'))"
--- a/tests/integ/scala/test_async_job_suite.py
+++ b/tests/integ/scala/test_async_job_suite.py
@@ -107,6 +107,7 @@ def test_async_to_pandas_common(session):
     )
 
 
+@pytest.mark.skipif(IS_IN_STORED_PROC_LOCALFS, reason="Requires large result")
 @pytest.mark.skipif(not is_pandas_available, reason="Pandas is not available")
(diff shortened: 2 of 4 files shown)


tests/integ/scala/test_async_job_suite.py
```diff
@@ -110,15 +110,14 @@ def test_async_to_pandas_common(session):
 @pytest.mark.skipif(IS_IN_STORED_PROC_LOCALFS, reason="Requires large result")
 @pytest.mark.skipif(not is_pandas_available, reason="Pandas is not available")
 def test_async_to_pandas_batches(session):
-    df = session.range(100000).cache_result()
+    df = session.range(100000).sort("id").cache_result()
```
sfc-gh-stan, line 113: This test broke with the change I made to `src/snowflake/snowpark/_internal/server_connection.py`, because previously the exact same set of pandas batches are returned (from result reuse since) but now it is returning a different set. E.g. If there are 10 rows in total, `to_pandas_batches` could return a iteration of 3 rows, 4 rows, 3 rows OR an iteration of 5 rows, 5 rows. This doesn't break any user promise and I just rewrote the test to reflect this change.

tests/integ/scala/test_async_job_suite.py
```diff
@@ -112,13 +112,16 @@ def test_async_to_pandas_common(session):
...
     df = session.range(100000).cache_result()
     async_job = df.to_pandas_batches(block=False)
-    res = list(async_job.result())
-    expected_res = list(df.to_pandas_batches())
-    assert len(res) > 0
-    assert len(expected_res) > 0
-    for r, er in zip(res, expected_res):
-        assert_frame_equal(r, er)
-        break
+    res = pd.concat(async_job.result(), ignore_index=True).sort_values(
```
sfc-gh-stan, line 115: Results are not guaranteed to be in the same order.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -434,8 +434,14 @@ def _to_data_or_iter(
         results_cursor: SnowflakeCursor,
         to_pandas: bool = False,
         to_iter: bool = False,
-        num_statements: Optional[int] = None,
     ) -> Dict[str, Any]:
+        if to_iter and not to_pandas:  # Fix for SNOW-869536
```
sfc-gh-stan, line 438: `to_pandas` doesn't have this issue, `SnowflakeCursor.fetch_pandas_batches` already handles the isolation.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -434,8 +434,14 @@ def _to_data_or_iter(
         results_cursor: SnowflakeCursor,
         to_pandas: bool = False,
         to_iter: bool = False,
-        num_statements: Optional[int] = None,
```
sfc-gh-sfan, line 437: Did we remove this because the param is unused?

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-sfan: Can we put this into comment instead? This seems more informative than "fix for" :)

sfc-gh-sfan (approved):

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-stan: Yes

sfc-gh-aalam (approved):

PR #1231, opened 2024-02-01
[SNOW-1036413] Fix stored proc register failure when packages are given as modules.

sfc-gh-lspiegelberg:

   SNOW-1036413

   SNOW-949781 introduced changes to stored proc register, which results in failure when giving packages as modules to the stored proc register function. Fix here.

### DIFF
--- a/src/snowflake/snowpark/stored_procedure.py
+++ b/src/snowflake/snowpark/stored_procedure.py
@@ -771,7 +771,8 @@ class StoredProcedureRegistration:
             if package_name not in self._session._packages:
                 packages = list(self._session._packages.values()) + [this_package]
         else:
-            if not any(package_name in p for p in packages):
+            package_names = [p if isinstance(p, str) else p.__name__ for p in packages]
+            if not any(p.startswith(package_name) for p in package_names):
                 packages.append(this_package)
 
         (
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -1508,7 +1508,7 @@ def test_stored_proc_register_with_module(session):
 
     # use pandas module here
     packages = list(session.get_packages().values())
-    assert "pandas" "pandas" not in packages
+    assert "pandas" not in packages
     packages = [pandas] + packages
 
     def proc_function(session: Session) -> str:
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -1497,3 +1497,28 @@ def test_force_inline_code(session):
             f, packages=["snowflake-snowpark-python"], force_inline_code=True
         )
     assert any("AS $$" in query.sql_text for query in query_history.queries)
+
+
+@pytest.mark.skipif(
+    not is_pandas_available, reason="test module dependency with pandas dependency"
+)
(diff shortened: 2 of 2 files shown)


tests/integ/test_stored_procedure.py
```diff
@@ -1497,3 +1497,28 @@ def f(session: Session) -> int:
...
+@pytest.mark.skipif(
+    not is_pandas_available, reason="test module dependency with pandas dependency"
+)
+def test_stored_proc_register_with_module(session):
+
+    import pandas
+
+    # use pandas module here
+    packages = list(session.get_packages().values())
+    assert "pandas" "pandas" not in packages
```
sfc-gh-jrose, line 1511: Is `"pandas" "pandas"` intended?

sfc-gh-evandenberg: If it's part of the error message may be put that in a variable that is referenced to avoid any confusion?

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -771,7 +771,8 @@ def _do_register_sp(
             if package_name not in self._session._packages:
                 packages = list(self._session._packages.values()) + [this_package]
         else:
-            if not any(package_name in p for p in packages):
+            package_names = [p if isinstance(p, str) else p.__name__ for p in packages]
+            if not any(p.startswith(package_name) for p in package_names):
```
sfc-gh-stan, line 775: Is this because `package_names` contain multiple entries of the same package with different versions?

sfc-gh-aling:
the test failure is because in our backend our sp connector only supports pandas < 2.1.0.
is it possible to use other package for test?

I've logged a JIRA to unblock the usage of newer pandas for connector as well as our sp connector:
https://snowflakecomputing.atlassian.net/browse/SNOW-1038863

tests/integ/test_stored_procedure.py
sfc-gh-lspiegelberg: good catch, this is a typo. fixed.

src/snowflake/snowpark/stored_procedure.py
sfc-gh-lspiegelberg: good question, I tried to mimick the code above `package_name in p` but though that packages will be given as strings in the form `pandas==2.0.3`. Hence, comparing with startswith seems the right idea. `in` seems wrong, because what if there's a package `connector` requested? that might conflict with `snowflake-connector` or `connector-java`. Prob. `startswith` is not even correct - do you know if there are some existing util functions to deal with correct resolution? Potentially splitting the string after `==, !=, <=, ...` into two parts and then comparing the first part may be the right thing to do. Wdyt?

sfc-gh-lspiegelberg:
> the test failure is because in our backend our sp connector only supports pandas < 2.1.0. is it possible to use other package for test?
> 
> I've logged a JIRA to unblock the usage of newer pandas for connector as well as our sp connector, we will get someone to work on it in the next sprint or if I have time this week I will work on it: https://snowflakecomputing.atlassian.net/browse/SNOW-1038863

Is there a package I could use for testing? If not, would it be ok to add a package (e.g., scipy) to the testing environment?

PR #1236, opened 2024-02-02
Add new API connection for SiS/SPCS

sfc-gh-smirzaei:

   Fixes #NNNN

   The goal is to add a new API `connect` for SiS and SPCS where we can create a session using the default connection params.

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -501,6 +501,10 @@ def is_in_stored_procedure():
     return PLATFORM == "XP"
 
 
+def is_in_spcs():
+    return os.path.exists("/snowflake/session/token")
+
+
 def random_name_for_temp_object(object_type: TempObjectType) -> str:
     return f"{TEMP_OBJECT_NAME_PREFIX}{object_type.value}_{generate_random_alphanumeric().upper()}"
 
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -2758,26 +2758,22 @@ class Session:
         ):  # (will be removed before merging) no need for `"streamlit" in sys.modules` check since is_in_stored_procedure() must be true for SiS as well
             # Basically noop and return the existing session
             return _get_active_session()
-        else:
-            try:
-                # 2) check if we are running in SPCS
-                with open("/snowflake/session/token") as file:
-                    token = file.read()
-                connection_parameters = {
-                    "account": os.getenv("SNOWFLAKE_ACCOUNT"),
-                    "host": os.getenv("SNOWFLAKE_HOST"),
-                    "authenticator": "oauth",
-                    "token": token,
-                    "warehouse": os.getenv("SNOWFLAKE_WAREHOUSE"),
-                    "database": os.getenv("SNOWFLAKE_DATABASE"),
-                    "schema": os.getenv("SNOWFLAKE_SCHEMA"),
-                }
-                return Session.builder.configs(connection_parameters).create()
-            except FileNotFoundError:
-                # 3) we are not running in SPCS, fall back to normal way of creating session from default connection params
-                return Session.builder.create()
-            except Exception:
-                # a general error happened, re-raise it
-                raise
+        try:
+            # 2) check if we are running in SPCS
+            with open("/snowflake/session/token") as file:
+                token = file.read()
+            connection_parameters = {
+                "account": os.getenv("SNOWFLAKE_ACCOUNT"),
+                "host": os.getenv("SNOWFLAKE_HOST"),
+                "authenticator": "oauth",
+                "token": token,
+                "warehouse": os.getenv("SNOWFLAKE_WAREHOUSE"),
+                "database": os.getenv("SNOWFLAKE_DATABASE"),
+                "schema": os.getenv("SNOWFLAKE_SCHEMA"),
+            }
+            return Session.builder.configs(connection_parameters).create()
+        except FileNotFoundError:
+            # 3) we are not running in SPCS, fall back to normal way of creating session from default connection params
+            return Session.builder.create()
 
     createDataFrame = create_dataframe
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -2767,6 +2767,8 @@ class Session:
             connection_parameters = {
                 "account": os.getenv("SNOWFLAKE_ACCOUNT"),
                 "host": os.getenv("SNOWFLAKE_HOST"),
+                "port": os.getenv("SNOWFLAKE_PORT"),
+                "protocol": os.getenv("SNOWFLAKE_PROTOCOL"),
                 "authenticator": "oauth",
                 "token": token,
                 "warehouse": os.getenv("SNOWFLAKE_WAREHOUSE"),
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -2749,4 +2749,33 @@ class Session:
             _logger.warning("query `%s` cannot be explained", query)
             return None
 
+    def create(self) -> "Session":
+        """Gets the existing session if one exists (in case of SPs/Streamlits) or creates a new one using default connection params (e.g., in case of SPCS)."""
+
+        # 1) check if in SiS or sproc
+        if is_in_stored_procedure() or "streamlit" in sys.modules:
+            # Basically noop and return the existing session
+            return _get_active_session()
+        else:
+            try:
+                # 2) check if we are running in SPCS
+                with open("/snowflake/session/token") as file:
+                    token = file.read()
+                connection_parameters = {
+                    "account": os.getenv("SNOWFLAKE_ACCOUNT"),
+                    "host": os.getenv("SNOWFLAKE_HOST"),
+                    "authenticator": "oauth",
+                    "token": token,
+                    "warehouse": os.getenv("SNOWFLAKE_WAREHOUSE"),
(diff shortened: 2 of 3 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -2749,4 +2749,35 @@ def _explain_query(self, query: str) -> Optional[str]:
...
+        # 1) check if in SiS or sproc
+        if (
+            is_in_stored_procedure()
+        ):  # (will be removed before merging) no need for `"streamlit" in sys.modules` check since is_in_stored_procedure() must be true for SiS as well
+            # Basically noop and return the existing session
+            return _get_active_session()
+        else:
+            try:
+                # 2) check if we are running in SPCS
+                with open("/snowflake/session/token") as file:
```
sfc-gh-smirzaei, line 2764: We could replace this block by adding a helper method like `_get_default_SPSC_connection_params` in connector (similar to `_get_default_connection_params`) to get SPCS default params and move the whole thing in `_create_internal`. Please lemme know which one is a better option/pattern.

src/snowflake/snowpark/session.py
```diff
@@ -2749,4 +2749,35 @@ def _explain_query(self, query: str) -> Optional[str]:
...
+    def create(self) -> "Session":
+        """Gets the existing session if one exists (in case of SPs/Streamlits) or creates a new one using default connection params (e.g., in case of SPCS)."""
+
+        # 1) check if in SiS or sproc
+        if (
+            is_in_stored_procedure()
+        ):  # (will be removed before merging) no need for `"streamlit" in sys.modules` check since is_in_stored_procedure() must be true for SiS as well
+            # Basically noop and return the existing session
+            return _get_active_session()
+        else:
```
sfc-gh-mkeller, line 2761: No need for the else-clause since you return in the then-clause. Just get rid of this line and de-dent the rest

src/snowflake/snowpark/session.py
```diff
@@ -2749,4 +2749,35 @@ def _explain_query(self, query: str) -> Optional[str]:
...
+                    "database": os.getenv("SNOWFLAKE_DATABASE"),
+                    "schema": os.getenv("SNOWFLAKE_SCHEMA"),
+                }
+                return Session.builder.configs(connection_parameters).create()
+            except FileNotFoundError:
+                # 3) we are not running in SPCS, fall back to normal way of creating session from default connection params
+                return Session.builder.create()
+            except Exception:
+                # a general error happened, re-raise it
+                raise
```
sfc-gh-mkeller, line 2781: No need to catch an exception just to throw it again, let it bubble up
```suggestion
```

src/snowflake/snowpark/session.py
sfc-gh-mkeller: I have no preference, either will work!

src/snowflake/snowpark/session.py
```diff
@@ -2749,4 +2749,32 @@ def _explain_query(self, query: str) -> Optional[str]:
...
+
+        # 1) check if in SiS or sproc
+        if (
+            is_in_stored_procedure()
+        ):  # (will be removed before merging) no need for `"streamlit" in sys.modules` check since is_in_stored_procedure() must be true for SiS as well
+            # Basically noop and return the existing session
+            return _get_active_session()
+        try:
+            # 2) check if we are running in SPCS
+            with open("/snowflake/session/token") as file:
```
sfc-gh-smirzaei, line 2764: Any suggestion/idea on how I can test this for SPCS?

tests/integ/test_session.py
```diff
@@ -586,7 +585,10 @@ def test_create_session_from_default_config_file(monkeypatch, db_parameters):
         with monkeypatch.context() as m:
             m.setenv("SNOWFLAKE_CONNECTIONS", tomlkit.dumps(doc))
             m.setenv("SNOWFLAKE_DEFAULT_CONNECTION_NAME", "default")
-            with Session.builder.create() as new_session:
+            if not IS_IN_STORED_PROC:
```
sfc-gh-smirzaei, line 588: Hmm now I'm not sure if it works (haven't tested this yet)

src/snowflake/snowpark/session.py
```diff
@@ -2749,4 +2749,32 @@ def _explain_query(self, query: str) -> Optional[str]:
             _logger.warning("query `%s` cannot be explained", query)
             return None
 
+    @staticmethod
+    def create() -> "Session":
```
sfc-gh-sfan, line 2753: Is `create` the right action name for "gets existing session if one exists", or should this be `fetch` instead?

src/snowflake/snowpark/session.py
sfc-gh-sfan, line 2764: Should we create a `is_in_spcs` util function? This way we don't need to include `except` in the normal code flow.

sfc-gh-sfan: This is probably even more challenging than stored proc. In stored proc, we have a test job that creates Anaconda package, and does some repodata hack to use the new package (test job: https://ci-dev-142.int.snowflakecomputing.com/job/PythonStoredProcBuildPrecommitTest/).

Do you know how SPCS does its regression test today?

src/snowflake/snowpark/session.py
sfc-gh-aalam: I think there is already a method called `getOrCreate`

src/snowflake/snowpark/session.py
sfc-gh-smirzaei: Done

src/snowflake/snowpark/session.py
sfc-gh-smirzaei: So basically this comes from a request from field/PM: https://docs.google.com/document/d/12FGgc9NeH-xa31FMZa_lO2N5Xq71Cm0aquTLI2-G0IE/edit#heading=h.hvtde6rw3c3p

They want a unified `connect` method in session (which I mistakenly named the method `created`) which works across different use-cases (i.e., when Snowpark is used as is or in SiS, SP, SPCS)

sfc-gh-sfan: I see. @sfc-gh-mkeller @sfc-gh-yixie : Do you have concerns for introducing a `connect` method in session?

tests/integ/test_session.py
sfc-gh-aalam: https://ci-dev-142.int.snowflakecomputing.com/job/PythonStoredProcBuildPrecommitTest/895/
ran a SP test for you

src/snowflake/snowpark/session.py
```diff
@@ -2749,4 +2750,34 @@ def _explain_query(self, query: str) -> Optional[str]:
...
 
+    @staticmethod
+    def connect() -> "Session":
+        """Gets the existing session if one exists (in case of SPs/Streamlits) or creates a new one using default connection params (e.g., in case of SPCS)."""
+
+        # 1) check if in SiS or sproc
+        if (
+            is_in_stored_procedure()
+        ):  # (will be removed before merging) no need for `"streamlit" in sys.modules` check since is_in_stored_procedure() must be true for SiS as well
+            # Basically noop and return the existing session
```
sfc-gh-aalam, line 2761: it would be removed before merging to main? I don't understand why have it then.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -501,6 +501,10 @@ def is_in_stored_procedure():
     return PLATFORM == "XP"
 
 
+def is_in_spcs():
+    return os.path.exists("/snowflake/session/token")
```
sfc-gh-yixie, line 505: Is there another way?

src/snowflake/snowpark/session.py
sfc-gh-yixie: There is already `Session.builder.getOrCreate()`

sfc-gh-stan: +1

tests/integ/test_session.py
sfc-gh-aalam: some tests failed. Looks like a rebase with main is needed before we do another run

src/snowflake/snowpark/session.py
sfc-gh-smirzaei: `getOrCreate` as is, doesn't work (since we expect to have token in the toml file and not as a separate file). But I had a conversion with Srikanth, he is tasked with adding `connect` for SPCS for _all_ connectors. He will write a design doc based on the existing `connect` API in Python connector where we take into account a new connect argument (like token_file) to be added to the toml file. I'll hold off working on this for now since his general solution is better/cleaner than what I have here.

src/snowflake/snowpark/session.py
```diff
@@ -2749,4 +2750,34 @@ def _explain_query(self, query: str) -> Optional[str]:
...
+            with open("/snowflake/session/token") as file:
+                token = file.read()
+            connection_parameters = {
+                "account": os.getenv("SNOWFLAKE_ACCOUNT"),
+                "host": os.getenv("SNOWFLAKE_HOST"),
+                "port": os.getenv("SNOWFLAKE_PORT"),
+                "protocol": os.getenv("SNOWFLAKE_PROTOCOL"),
+                "authenticator": "oauth",
+                "token": token,
+                "warehouse": os.getenv("SNOWFLAKE_WAREHOUSE"),
```
sfc-gh-bhess, line 2774: SNOWFLAKE_WAREHOUSE may or may not be set. Setting the QUERY_WAREHOUSE on the SERVICE is optional.

PR #1247, opened 2024-02-12
SNOW-1043123: Add greatest/least support to local testing

sfc-gh-jrose:

   Fixes #1043123

  This commit adds support for `greatest` and `least` in local testing. There are also a bunch of format changes that pre-commit made as well.


src/snowflake/snowpark/mock/_functions.py
```diff
@@ -929,3 +929,17 @@ def mock_to_variant(expr: ColumnEmulator):
...
+
+@patch("greatest")
+def mock_greatest(*exprs: ColumnEmulator):
+    result = reduce(lambda x, y: x.combine(y, max), exprs)
+    result.sf_type = exprs[0].sf_type
+    return result
+
+
+@patch("least")
+def mock_least(*exprs: ColumnEmulator):
```
sfc-gh-stan, line 942:
```
The first argument determines the return type:

If the first type is numeric, then the return type is ‘widened’ according to the numeric types in the list of all arguments.

If the first type is not numeric, then all other arguments must be convertible to the first type.

If any of the argument values is NULL, the result is NULL.
```
from https://docs.snowflake.com/en/sql-reference/functions/least#usage-notes

The first two cases since those require coercion (implicit type conversion, "123" can be implicitly evaluated to `123`) for this PR, we should leave a TODO in the docstring.

Does the current implementation support the third case?

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -929,3 +929,17 @@ def mock_to_variant(expr: ColumnEmulator):
     res = expr.copy()
     res.sf_type = ColumnType(VariantType(), expr.sf_type.nullable)
     return res
+
+
+@patch("greatest")
+def mock_greatest(*exprs: ColumnEmulator):
```
sfc-gh-stan, line 935: Same as below.

sfc-gh-stan (changes requested):
Please add a TODO comment for type coercion in the docstrings and check the edge case when the input column (only) has null values.

src/snowflake/snowpark/mock/_functions.py
sfc-gh-jrose: I've modified logic and added test cases to show that coercion works correctly.

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -933,6 +933,45 @@ def mock_to_variant(expr: ColumnEmulator):
...
+    if x is None or y is None:
+        return (None, None)
+
+    _x = x if isinstance(x, Number) else float(x)
+    _y = y if isinstance(y, Number) else float(y)
+
+    if _x > _y:
+        return (_x, _y)
+    else:
+        return (_y, _x)
```
sfc-gh-aling, line 950: do we consider non-numeric types here?
I see in the description https://docs.snowflake.com/en/sql-reference/functions/least:

`LEAST supports all data types, including VARIANT.`

sfc-gh-jrose: I can add some more test cases to be sure. From what I can tell anything non numeric gets cast to a numeric type.

sfc-gh-aling: interesting, if I compare a string "abc" with a datetime "2024-01-1", both would be cast to numeric first?

sfc-gh-jrose: That case actually generates an exception:
```
SELECT greatest("A", "B") FROM ( SELECT "A", to_timestamp("B") AS "B" FROM ( SELECT $1 AS "A", $2 AS "B" FROM  VALUES ('abc' :: STRING, '2024-01-01 00:00:00' :: STRING)));
```
```
Timestamp 'abc' is not recognized
```

sfc-gh-jrose: Although if thy are all strings it does look like they aren't converted to numerics so the _compare function is incorrect. I'll fix it up.

src/snowflake/snowpark/mock/_functions.py
sfc-gh-stan: Sorry I had a typo in the comment above, I meant that coercion is not supported in Local Testing in general so we could ignore the first two cases in this PR and simply leave a TODO in the docstring (including a JIRA ticket)

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -933,6 +933,45 @@ def mock_to_variant(expr: ColumnEmulator):
     return res
 
 
+def _compare(x: Any, y: Any) -> Tuple[Any, Any]:
```
sfc-gh-stan, line 936: Let's use `TypeVar` here to indicate that input and output needs to be the same type.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
   - current_timestamp
   - current_date
   - current_time
+  - greatest
+  - least
```
sfc-gh-aling, line 20: let's move this under 1.14.0

src/snowflake/snowpark/mock/_functions.py
sfc-gh-aling: thanks! do you know what exception class it's raising?
I'm thinking of raising the same error to align the error experience in live mode.
for example, I noticed that in the code `type(_x)(y)`, this might raise `ValueError`, I'm thinking if we should wrap this into a snowpark client exception

sfc-gh-aling (commented):
I have left a comment about the error experience, otherwise the change LGTM

sfc-gh-stan (approved):

sfc-gh-aling (approved):
LGTM, @sfc-gh-jrose can you help create a JIRA to track the error experience?

PR #1257, opened 2024-02-17
SNOW-981562: add way to respect user defined schema before inferring schema

sfc-gh-aalam:

   Fixes #SNOW-981562, SNOW-1544694

   This PR allows users to specify schema for `session.crate_dataframe` from pandas dataframes. We follow a best effort way to create dataframe using specified schema but fall back to infer-schema if we fail. This is done so we do not introduce a breaking change.


src/snowflake/snowpark/session.py
```diff
@@ -2278,6 +2305,30 @@ def create_dataframe(
...
+                            database=sf_database,
+                            schema=sf_schema,
+                            quote_identifiers=True,
+                            use_logical_type=self._use_logical_type_for_create_df,
+                        )
+                        set_api_call_source(t, "Session.create_dataframe[pandas]")
+                        return t
+                    except ProgrammingError as e:
+                        self._run_query(f"drop table if exists {temp_table_name}")
+                        logging.debug(
```
sfc-gh-jrose, line 2327: This might be better as a warn so that the user is more likely to see it.

src/snowflake/snowpark/session.py
```diff
@@ -2191,6 +2193,31 @@ def write_pandas(
...
+            temp_table_name: table name
+            schema: user provided StructType schema
+
+        Returns:
+            True table was created successfully, else False
+        """
+        try:
+            schema_string = attribute_to_schema_string(schema._to_attributes())
+            self._run_query(
+                f"CREATE SCOPED TEMP TABLE {temp_table_name} ({schema_string})"
```
sfc-gh-jrose, line 2211: I'm not particularly familiar with scoped temp tables. When do they get cleaned up if not explicitly deleted?

sfc-gh-jrose (approved):

src/snowflake/snowpark/session.py
sfc-gh-aalam: When run outside of a stored proc, using `SCOPED` keyword is a no-op. Within stored proc, it is mainly used to limit the snowpark internal created temp object scope. This is not properly documented and I learnt about this by asking out in our slack channels

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 
 - Added support for an optional `date_part` argument in function `last_day`.
 - `SessionBuilder.app_name` will set the query_tag after the session is created.
+- Added support to create dataframe using user provided schema when snowpark dataframe is created from pandas dataframe on a best effort basis.
```
sfc-gh-aling, line 19: this needs to go to v1.14.0

src/snowflake/snowpark/session.py
```diff
@@ -2276,6 +2303,30 @@ def create_dataframe(
...
+                # the temp table. If we fail, go back to old method using infer schema.
+                if isinstance(
+                    schema, StructType
+                ) and self._create_temp_table_for_given_schema(temp_table_name, schema):
+                    try:
+                        t = self.write_pandas(
+                            data,
+                            temp_table_name,
+                            database=sf_database,
+                            schema=sf_schema,
```
sfc-gh-aling, line 2317: I'm wondering if this is still considered a BCR -- let's say previously even users specify the schema, but it's not appreciated when input is a pandas DF, so the schema of the result DF is different than what users specified.

however, with the change, it can be the schema users specified.

CHANGELOG.md
sfc-gh-aalam: good catch

src/snowflake/snowpark/session.py
```diff
@@ -2276,6 +2303,30 @@ def create_dataframe(
                 )
                 sf_schema = self._conn._get_current_parameter("schema", quoted=False)
 
+                # If the user specifies schema for their dataframe, we try out best to match
+                # it by create a temp table with the specified schema, and load the data into
+                # the temp table. If we fail, go back to old method using infer schema.
+                if isinstance(
+                    schema, StructType
+                ) and self._create_temp_table_for_given_schema(temp_table_name, schema):
```
sfc-gh-jdu, line 2311: should we consider dropping this temp table after evaluation?

sfc-gh-jdu (approved):

src/snowflake/snowpark/session.py
sfc-gh-aalam: If we are successful at writing pandas data into the temp table, then we will use this table in this session. If we drop it, we won't be able to use the dataframe.

sfc-gh-aalam: Now that we have temp table clean-up, we can update `Session.write_pandas` to clean-up temp table once the usage is done?

sfc-gh-aalam: I separated this change out into this PR: https://github.com/snowflakedb/snowpark-python/pull/2784

PR #1260, opened 2024-02-20
SNOW-1058663  Fix `ServerConnection. _to_data_or_iter` to return the original query id

sfc-gh-stan:

   Fixes SNOW-1058663

   **Context:** Currently t_python_sp::table_procs_on_snowsql.ssql is failing in PythonStoredProcBuildPrecommitTest because of the changes introduced in SNOW-869536 Fix buggy behavior in DataFrame.to_local_iterator  by sfc-gh-stan · Pull Request #1226 · snowflakedb/snowpark-python . This is because the implementation of table sprocs calls `_execute_and_get_query_id`, which calls `ServerConnection._to_data_or_iter` with `to_iter` set to True, therefore is affected by the code change in Snowpark Python API. The error indicates SnowSQL can only retrieve the query result in Arrow, therefore it is likely due to `_execute_and_get_query_id` needs to return the query ID of the original query instead of a result scan.
  
    **Changes in this PR:** Make sure `ServerConnection._to_data_iter` returns the query ID of the original query instead of the result scan when `to_iter` is True and `to_pandas` is False.

    **Successful precommit run**: https://ci-dev-133.int.snowflakecomputing.com/job/RT-Precommit/13976/

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -440,9 +440,7 @@ class ServerConnection:
             to_iter and not to_pandas
         ):  # Fix for SNOW-869536, to_pandas doesn't have this issue, SnowflakeCursor.fetch_pandas_batches already handles the isolation.
             new_cursor = results_cursor.connection.cursor()
-            new_cursor.execute(
-                f"SELECT * FROM TABLE(RESULT_SCAN('{results_cursor.sfqid}'))"
-            )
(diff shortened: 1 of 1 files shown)


sfc-gh-aalam (approved):
We could probably re-write the result scan statement as `f"SELECT * FROM TABLE(RESULT_SCAN('qid'))"` since we have a variable for that already but I'm fine otherwise too. I like the fact that overall the fix is simple and elegant

sfc-gh-sfan (approved):

PR #1274, opened 2024-02-26
SNOW-1060064: Eliminate repeated subquery using CTE (part 1, with original query generation framework)

sfc-gh-jdu:

   Fixes SNOW-1060064

   This PR adds an optimization to eliminate repeated subquery using CTE (see details in https://docs.google.com/document/d/1vVUYqLeD_nQRVaH3SX2c4jDNJoBGHJlg8PqiJ8qxgVM/edit), with original query generation framework. 

The basic idea is to convert repeated subqueries to a CTE, e.g., 
`
( SELECT * FROM ( SELECT  *  FROM ( SELECT "A", "B" FROM ( SELECT $1 AS "A", $2 AS "B" FROM  VALUES (1 :: INT, 2 :: INT), (3 :: INT, 4 :: INT))) WHERE ("A" = 1 :: INT))) UNION ALL ( SELECT * FROM ( SELECT  *  FROM ( SELECT "A", "B" FROM ( SELECT $1 AS "A", $2 AS "B" FROM  VALUES (1 :: INT, 2 :: INT), (3 :: INT, 4 :: INT))) WHERE ("A" = 1 :: INT)))
`
will be converted to 
`
WITH SNOWPARK_TEMP_TABLE_SD2LMA9GKT AS ( SELECT * FROM ( SELECT  *  FROM ( SELECT "A", "B" FROM ( SELECT $1 AS "A", $2 AS "B" FROM  VALUES (1 :: INT, 2 :: INT), (3 :: INT, 4 :: INT))) WHERE ("A" = 1 :: INT))) ( SELECT  *  FROM (SNOWPARK_TEMP_TABLE_SD2LMA9GKT)) UNION ALL ( SELECT  *  FROM (SNOWPARK_TEMP_TABLE_SD2LMA9GKT))
`

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -157,6 +157,7 @@ RENAME = " RENAME "
 INTERSECT = f" {Intersect.sql} "
 EXCEPT = f" {Except.sql} "
 NOT_NULL = " NOT NULL "
+WITH = "WITH "
 
 TEMPORARY_STRING_SET = frozenset(["temporary", "temp"])
 
@@ -1382,3 +1383,10 @@ def get_file_format_spec(
         file_format_str += FORMAT_NAME + EQUALS + file_format_name
     file_format_str += RIGHT_PARENTHESIS
     return file_format_str
+
+
+def cte_statement(queries: List[str], table_names: List[str]) -> str:
+    return WITH + COMMA.join(
+        table_name + AS + LEFT_PARENTHESIS + query + RIGHT_PARENTHESIS
+        for query, table_name in zip(queries, table_names)
+    )
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -1386,7 +1386,8 @@ def get_file_format_spec(
 
 
 def cte_statement(queries: List[str], table_names: List[str]) -> str:
-    return WITH + COMMA.join(
-        table_name + AS + LEFT_PARENTHESIS + query + RIGHT_PARENTHESIS
+    result = COMMA.join(
+        f"{table_name}{AS}{LEFT_PARENTHESIS}{query}{RIGHT_PARENTHESIS}"
         for query, table_name in zip(queries, table_names)
     )
+    return f"{WITH}{result}"
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -18,7 +18,6 @@ from typing import (
     List,
     Optional,
     Sequence,
-    Set,
     Tuple,
 )
 
@@ -37,7 +36,6 @@ if TYPE_CHECKING:
 import snowflake.connector
 import snowflake.snowpark
 from snowflake.snowpark._internal.analyzer.analyzer_utils import (
-    SPACE,
     aggregate_statement,
     attribute_to_schema_string,
     batch_insert_into_statement,
@@ -48,7 +46,6 @@ from snowflake.snowpark._internal.analyzer.analyzer_utils import (
     create_or_replace_view_statement,
     create_table_as_select_statement,
     create_table_statement,
-    cte_statement,
     delete_statement,
     drop_file_format_if_exists_statement,
     drop_table_if_exists_statement,
@@ -78,6 +75,10 @@ from snowflake.snowpark._internal.analyzer.analyzer_utils import (
 from snowflake.snowpark._internal.analyzer.binary_plan_node import (
     JoinType,
     SetOperation,
+)
+from snowflake.snowpark._internal.analyzer.cte_utils import (
+    create_cte_query,
+    find_duplicate_subtrees,
 )
 from snowflake.snowpark._internal.analyzer.expression import Attribute
 from snowflake.snowpark._internal.analyzer.schema_utils import analyze_attributes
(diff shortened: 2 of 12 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -220,6 +225,124 @@ def __init__(
...
+                    # replace the placeholder (id) with child query
+                    plan_to_query_map[node] = plan_to_query_map[node].replace(
+                        child._id, plan_to_query_map[child]
+                    )
+
+            # duplicate subtrees will be converted CTEs
+            if node in filtered_duplicate_plan_set:
+                # when a subquery is converted a CTE to with clause,
+                # it will be replaced by `SELECT * from TEMP_TABLE` in the original query
+                table_name = random_name_for_temp_object(TempObjectType.TABLE)
```
sfc-gh-jdu, line 327: I'm wondering, whether we should create another object type for CTE, which can be less confusing to customers (because it's not a real table) and easier for us to debug.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -23,6 +24,13 @@
 class LogicalPlan:
     def __init__(self) -> None:
         self.children = []
+        self._id = str(uuid.uuid4())
+
+    def __eq__(self, other: "LogicalPlan") -> bool:
+        return isinstance(other, LogicalPlan) and (self._id == other._id)
+
+    def __hash__(self) -> int:
+        return hash(self._id)
```
sfc-gh-aalam, line 33: My understanding is that we are relying on this during traversal to find duplicate nodes, is that correct? I have seen usage patters where users have dataframe constructor class to generate specific dataframes defined by configurations given to the constructor class. My concern is that this approach will not detect dataframes that are duplicates - as in they embody the exact same query - but they will not get detected as duplicates because they have different UUIDs.

sfc-gh-jdu: I think your point makes sense, and we can generate hash based on query text. But then I need to move `_id` to `SnowflakePlan` instead of `LogicalPlan` because query is generated upon a Snowflake Plan is created

sfc-gh-jdu: done

src/snowflake/snowpark/session.py
```diff
@@ -460,6 +460,7 @@ def __init__(
                 _PYTHON_SNOWPARK_USE_SQL_SIMPLIFIER_STRING, True
             )
         )
+        self._cte_optimization_enabled: bool = False
```
sfc-gh-aalam, line 463: overall I don't find any issues with the change. can we set this param using an environment variable and add a merge gate to integ tests with this param enabled?

sfc-gh-jdu: Maybe we follow the similar step when we added SQL simplifier? Add a server side parameter and add a merge gate. And maybe I can do it in next PR (integrate CTE optimization with SQL simplifier).

sfc-gh-aalam: sounds good

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: I would not be opposed to it

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -220,6 +225,145 @@ def __init__(
...
+                plan_to_query_map[node] = select_stmt
+
+        build_plan_to_query_map_in_post_order(self)
+
+        # construct with clause
+        with_stmt = cte_statement(
+            list(duplicate_plan_to_cte_map.values()),
+            list(duplicate_plan_to_table_name_map.values()),
+        )
+        final_query = with_stmt + SPACE + plan_to_query_map[self]
```
sfc-gh-aalam, line 361: since we are finding the topmost duplicate node, we don't need to concern ourselves with finding the correct order, right. This is cool

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: Yeah, because we build `duplicate_plan_to_cte_map` in post order, where children node is ahead of parent node. It satisfies the requirement of with clause:

> A WITH clause can refer recursively to itself, and to other CTEs that appear earlier in the same clause. For instance, cte_name2 can refer to cte_name1 and itself, while cte_name1 can refer to itself, but not to cte_name2.

src/snowflake/snowpark/session.py
sfc-gh-jdu: @sfc-gh-yixie wdyt?

sfc-gh-yixie: Agree

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1382,3 +1383,10 @@ def get_file_format_spec(
         file_format_str += FORMAT_NAME + EQUALS + file_format_name
     file_format_str += RIGHT_PARENTHESIS
     return file_format_str
+
+
+def cte_statement(queries: List[str], table_names: List[str]) -> str:
+    return WITH + COMMA.join(
+        table_name + AS + LEFT_PARENTHESIS + query + RIGHT_PARENTHESIS
```
sfc-gh-yixie, line 1390: Better use string format

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -220,6 +225,145 @@ def __init__(
...
+                    is_any_parent_unique_node = any(
+                        node_count_map[n] == 1 for n in node_parents_map[node]
+                    )
+                    if is_any_parent_unique_node:
+                        return True
+                    else:
+                        has_multi_parents = len(node_parents_map[node]) > 1
+                        if has_multi_parents:
+                            return True
+                return False
```
sfc-gh-yixie, line 307: How about adding some unit test code?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yixie: Yeah I agree.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -220,6 +225,145 @@ def __init__(
...
+        # if there is no duplicate node, no optimization will be performed
+        duplicate_plan_set = find_duplicate_subtrees(self)
+        if not duplicate_plan_set:
+            return self
+
+        plan_to_query_map = {}
+        duplicate_plan_to_cte_map = {}
+        duplicate_plan_to_table_name_map = {}
+
+        def build_plan_to_query_map_in_post_order(node: SnowflakePlan) -> None:
```
sfc-gh-yixie, line 321: Same here. Unit tests.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -272,15 +416,16 @@ def output_dict(self) -> Dict[str, Any]:
 
     def __copy__(self) -> "SnowflakePlan":
         return SnowflakePlan(
-            self.queries.copy() if self.queries else [],
+            copy.deepcopy(self.queries) if self.queries else [],
```
sfc-gh-yixie, line 419: Why is deepcopy required here?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -220,6 +225,145 @@ def __init__(
...
+
+    def __hash__(self) -> int:
+        return hash(self._id)
+
+    def replace_repeated_subquery_with_cte(self) -> "SnowflakePlan":
+        # parameter protection
+        # TODO SNOW-106671: enable cte optimization with sql simplifier
+        if (
+            not self.session._cte_optimization_enabled
+            or self.session._sql_simplifier_enabled
```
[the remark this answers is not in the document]
sfc-gh-jdu: The latter. We'll not check sql simplifier soon after next PR

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: because self.queries contains Query object, instead of query string, which can be used in original query plan object

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -220,6 +225,145 @@ def __init__(
...
+                      df2  df2   df2  df2
+                       |    |     |    |
+                      df1  df1   df1  df1
+
+            df4, df3 and df2 are duplicate subtrees.
+
+            This function is used to only include nodes that should be converted to CTEs.
+            """
+            node_count_map = defaultdict(int)
+            node_parents_map = {root: set()}
```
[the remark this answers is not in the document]
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: Yes, very good point, added

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: I found it's difficult/not worthwhile to write unit test for this function because what we're going to test is the generate sql, and I need to create many fake queries in unit tests. It's already tested in integration tests. But I moved them to a util file, which is more clear to review the code

sfc-gh-yixie (approved):

Issue #1266, opened 2024-02-22
Patching function with no argument


petsvakala:
Hi

I was following instructions how to patch built-in functions ( https://docs.snowflake.com/en/developer-guide/snowpark/python/testing-locally#patching-built-in-functions) however I am not sure how to do that for **current_date()** function.

This is how I have approached that:
```
@patch(current_date)
def patch_current_date() -> ColumnEmulator:
    ret_column = ColumnEmulator(data=[datetime.date.today()])
    ret_column.sf_type = ColumnType(DateType(), True)
    return ret_column
```
   
but this only fills the first row of dataframe. Rest of the rows for that column will be NA.
<img width="153" alt="image" src="https://github.com/snowflakedb/snowpark-python/assets/63597531/63352197-3884-4904-9dc0-f24597b021a3">

This is how that specific code line looks like in my test function:
`input_df.with_column('CURRENT_DATE', current_date())`

sfc-gh-aling:
hi @petsvakala , thanks for reaching out.
@sfc-gh-jrose I know you added the support to current_date recently, can you help take a look at this issue to see if this is covered?

sfc-gh-jrose:
I did add `current_date` recently, but it hasn't made it into a release yet I don't think. I believe the issue in this bug is this line:
```
    ret_column = ColumnEmulator(data=[datetime.date.today()])
```
The column emulator assumes that the data is the same length as the column and inserts `None` if no data remains in the list. If you remove the list braces it will instead be a single value that is used for all entries in the column instead.
```
    ret_column = ColumnEmulator(data=datetime.date.today())
```

sfc-gh-jrose:
I was wrong. This appears to be a gap in the local testing API. I'll see if support can be added by the next release.

petsvakala:
Ok, what close status should I choose for this issue for time being or I will keep it open?

sfc-gh-jfreeberg:
@petsvakala  -- Can you retry with v1.14.0? It should be fixed now

petsvakala:
Hi
Unfortunately still same behaviour.

**Here you can see new package exists:**
_poetry show snowflake-snowpark-python_
```
 name         : snowflake-snowpark-python     
 version      : 1.14.0                        
 description  : Snowflake Snowpark for Python 

dependencies
 - cloudpickle >=1.6.0,<2.1.0 || >2.1.0,<2.2.0 || >2.2.0,<=2.2.1
[...]

PR #1311, opened 2024-03-14
SNOW-1075566: Refactor local testing to allow constant function returns

sfc-gh-jrose:

   Fixes #SNOW-1075566

  This PR adds some pattern matching to function expression handling that detects constant functions and fills the column with the returned value rather than null filling after the first value.

See here for more info: https://github.com/snowflakedb/snowpark-python/issues/1266

An alternative to this PR would be to introduce a new return type to local functions that tells the planner that the result is a constant that should populate a whole column. This would make it more explicit when this behaviour should happen, but would make the user experience of making a local function a little more confusing. @sfc-gh-aling and I agreed that the current approach is better for that reason.

### DIFF
--- a/src/snowflake/snowpark/mock/_plan.py
+++ b/src/snowflake/snowpark/mock/_plan.py
@@ -407,7 +407,7 @@ def handle_function_expression(
         )
         and len(result) == 1
     ):
-        resized = result.repeat(len(input_data)).reset_index()[0]
+        resized = result.repeat(len(input_data)).reset_index(drop=True)
         resized.sf_type = result.sf_type
         return resized
 
--- a/src/snowflake/snowpark/mock/_plan.py
+++ b/src/snowflake/snowpark/mock/_plan.py
@@ -394,7 +394,24 @@ def handle_function_expression(
             current_row
         )  # the row's 0-base index in the window
         to_pass_args.append(row_idx)
-    return _MOCK_FUNCTION_IMPLEMENTATION_MAP[func_name](*to_pass_args)
+
+    result = _MOCK_FUNCTION_IMPLEMENTATION_MAP[func_name](*to_pass_args)
+
+    # If none of the args are column emulators and the function result only has one item
+    # assume that the single value should be repeated instead of Null filled. This allows
+    # constant expressions like current_date or current_database to fill a column instead
+    # of just the first row.
+    if (
+        not any(
+            isinstance(arg, (ColumnEmulator, TableEmulator)) for arg in to_pass_args
+        )
+        and len(result) == 1
+    ):
+        resized = result.repeat(len(input_data)).reset_index()[0]
+        resized.sf_type = result.sf_type
+        return resized
+
+    return result
 
 
 def handle_udf_expression(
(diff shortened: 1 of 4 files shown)


tests/integ/scala/test_dataframe_suite.py
```diff
@@ -2704,6 +2704,7 @@ def test_drop_duplicates(session):
     )
 
     result1 = df.dropDuplicates(["a"])
+    # import pdb; pdb.set_trace()
```
sfc-gh-aling, line 2707: nit: remove this?
```suggestion
```

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -394,7 +394,24 @@ def handle_function_expression(
             current_row
         )  # the row's 0-base index in the window
         to_pass_args.append(row_idx)
-    return _MOCK_FUNCTION_IMPLEMENTATION_MAP[func_name](*to_pass_args)
+
+    result = _MOCK_FUNCTION_IMPLEMENTATION_MAP[func_name](*to_pass_args)
```
sfc-gh-aling, line 398: I'm thinking of a more generalized way to handle this. Sophie has left a comment in the UDF PR and I'm wondering whether the same logic could be applied here: https://github.com/snowflakedb/snowpark-python/pull/1316#discussion_r1532479587.

if we pre-create an empty table/column emulator with size of len(input_data), will the constant data auto fill into the empty dataframe?

sfc-gh-jrose: That approach doesn't work as well here because there is a bit of a difference in how they are applied. 

UDF are applied as a row-wise operation so by initializing a new empty index of size `n` we can insure that `n` calls to the udf happen and `n` values are generated to add to the df.

Built-in functions on the other hand are executed with entire columns at a time which is typically more performant, but also means that without any column inputs a function doesn't know how many values to add.

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -394,7 +394,24 @@ def handle_function_expression(
...
+    # assume that the single value should be repeated instead of Null filled. This allows
+    # constant expressions like current_date or current_database to fill a column instead
+    # of just the first row.
+    if (
+        not any(
+            isinstance(arg, (ColumnEmulator, TableEmulator)) for arg in to_pass_args
+        )
+        and len(result) == 1
+    ):
+        resized = result.repeat(len(input_data)).reset_index()[0]
```
sfc-gh-stan, line 410: NIT:  `reset_index(drop=True)` should be able to replace the [0]

tests/integ/test_function.py
```diff
@@ -209,19 +209,31 @@ def test_order(session):
 @pytest.mark.localtest
 def test_current_date_and_time(session):
     max_delta = 1
-    df = session.create_dataframe([1]).select(
-        current_date(), current_time(), current_timestamp()
+    df = (
+        session.create_dataframe([1, 2])
```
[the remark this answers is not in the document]
sfc-gh-jrose: The main reason this bug wasn't observed initially was because constant functions returned one row previously which appears to be correct when there is only one row in the input data. By making the test have two rows I was able to observe the bug by running the test.

sfc-gh-stan (approved):

sfc-gh-yuwang (approved):

PR #1315, opened 2024-03-15
SNOW-1064309: local testing read json support

sfc-gh-aling:

   Fixes SNOW-1064309

local testing session.read.json feature

this is to support snowflake querying from stage json file https://docs.snowflake.com/en/user-guide/querying-stage.
it allows users to read data from json files, one json file is one row of data.

by selecting from json files:
- if infer schema is not specified, then json files are loaded as string data
- if infer schema is specified, then we will parse the json data schema, convert fields to columns. snowflake supports loading multiple json files and these json files can be of different schema, in case there is same column name but of different data types, snowflake coerce the type to be string


src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +444,80 @@ def read_file(
...
+                        contents.append(content)
+                        # if not result_df_sf_types:
+                        # extract the schema from the content
+                        schema = infer_schema(content)
+                        for field in schema:
+                            lower_filed_name = field.name.lower()
+                            column_name = (
+                                lower_filed_name
+                                if lower_filed_name in content
+                                else file.name
```
sfc-gh-jrose, line 485: Why substitute in the file name here?

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +444,80 @@ def read_file(
...
+                    ):
+                        content[miss_key] = None
+                    df = pd.DataFrame([content])
+                    result_df = pd.concat([result_df, df], ignore_index=True)
+                # snowflake output sorted column names
+                sorted_columns = sorted(list(result_df.columns))
+                result_df = result_df[sorted_columns]
+                # result_df.sf_types_by_col_index = {
+                #     i: result_df_sf_types[col] for i, col in enumerate(sorted_columns)
+                # }
```
sfc-gh-jrose, line 516: can this be removed?

sfc-gh-aling: yes I cleaned up the code

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-aling: thanks for catching this part, I realized that there is a bug here that the implementation can not handle mixed-case column name.
I reimplemented this part and updated the test file to test the case.

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -83,6 +84,18 @@ def _build_target_path(stage_location: str, dest_prefix: str = "") -> str:
     return f"{qualified_stage_name}{dest_prefix_name if dest_prefix_name else ''}"
 
 
+def _row_output_converter(cell_value, column_idx, sf_types):
+    if cell_value is None:
+        return None
+    if isinstance(sf_types[column_idx].datatype, DecimalType):
+        return Decimal(str(cell_value))
+    if isinstance(sf_types[column_idx].datatype, StringType):
+        if cell_value is True or cell_value is False:
```
sfc-gh-stan, line 93: Why would this case happen?

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -83,6 +84,18 @@ def _build_target_path(stage_location: str, dest_prefix: str = "") -> str:
     return f"{qualified_stage_name}{dest_prefix_name if dest_prefix_name else ''}"
 
 
+def _row_output_converter(cell_value, column_idx, sf_types):
```
sfc-gh-stan, line 87: Why do we need the conversion here? The data of a ColumnEmulator should be coerced into the column type when  an instance of ColumnEmulator is initialized.

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -1351,6 +1351,15 @@ def calculate_expression(
         try:
             return input_data[exp.name]
         except KeyError:
+            if exp.name.startswith('"') and exp.name.endswith('"'):
```
sfc-gh-stan, line 1354: It's better to process `exp.name` in the try block instead of adding a nested try block in the except case.

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -814,7 +839,33 @@ def test_read_json_with_no_schema(session, mode):
...
+        Row(
+            '{\n  "CoLor": 10,\n  "new_field": true,\n  "new_fruit": "NewApple",\n  "new_size": 10\n}'
+        ),
+        Row('{\n  "color": "Red",\n  "fruit": "Apple",\n  "size": "Large"\n}'),
+        Row('{\n  "color": true,\n  "fruit": "Banana",\n  "size": "Small"\n}'),
+    ]
+
+    if not local_testing_mode:
+        # query_test
+        res = df1.where(sql_expr("$1:color") == "Red").collect()
```
sfc-gh-stan, line 858: I feel like we could rewrite this to run in local testing, like `df1["$1"]["color"]`.

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +444,73 @@ def read_file(
...
+                    )
+                    column_series.sf_type = ColumnType(
+                        schema[i].datatype, schema[i].nullable
+                    )
+                    result_df[column_name] = column_series
+                    result_df_sf_types[column_name] = column_series.sf_type
+
+                for local_file in local_files:
+                    with open(local_file) as file:
+                        content = json.load(file)
```
sfc-gh-stan, line 468: This should use the CustomJsonDecoder.

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +444,73 @@ def read_file(
...
+                        df = pd.DataFrame({result_df.columns[0]: [content]})
```
sfc-gh-stan, line 469: Does this mean we ignore all the other columns? Or can there only be one column in the schema if `infer_schema_opt` is False?

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +444,73 @@ def read_file(
...
+                        result_df = pd.concat([result_df, df], ignore_index=True)
+            else:
+                # need to infer schema
+                contents = []
+                for local_file in local_files:
+                    with open(local_file) as file:
+                        content = json.load(file)
```
sfc-gh-stan, line 476: This should use the CustomJsonDecoder.

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +444,73 @@ def read_file(
...
+                                # snowflake coerce to string type
+                                target_datatype = StringType()
+
+                            column_series = ColumnEmulator(
+                                data=None, dtype=object, name=column_name
+                            )
+                            column_series.sf_type = ColumnType(
+                                target_datatype, nullable=True
+                            )
+                            result_df[column_name] = column_series
```
sfc-gh-stan, line 497: These could be merged into one line.

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +444,73 @@ def read_file(
...
+                # snowflake output the value as raw string
+                for i in range(len(schema)):
+                    column_name = analyzer.analyze(schema[i])
+                    column_series = ColumnEmulator(
+                        data=None, dtype=object, name=column_name
+                    )
+                    column_series.sf_type = ColumnType(
+                        schema[i].datatype, schema[i].nullable
+                    )
+                    result_df[column_name] = column_series
```
sfc-gh-stan, line 463: NIT: these could be merged into one line.

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-stan, line 497: NIT: these could be merged into one line.

sfc-gh-stan (changes requested):
Left a few comments and questions.

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-aling: updated

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-aling: updated

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-aling: good catch, updated

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-aling: good catch, updated

tests/integ/scala/test_dataframe_reader_suite.py
sfc-gh-aling: good suggestion and it worked, updated

src/snowflake/snowpark/mock/_plan.py
sfc-gh-aling: love this idea, code updated

src/snowflake/snowpark/mock/_connection.py
sfc-gh-stan: Adam and I chatted about this offline, we need this conversion logic because when there are multiple json files with conflicting schema, for example, the same column name maps to an number in one file and an array in another file. When Snowflake sees a conflict like this, the default behavior seems to be just converting all values in this column to string.

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-aling: it's the latter one that there should only be 1 column if not inferring schema
added an assertion at the beginning.

src/snowflake/snowpark/mock/_connection.py
sfc-gh-aling: moved the logic into an inner method: convert_value_to_str_for_string_type

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -1444,7 +1444,13 @@ def calculate_expression(
                 raise_error=NotImplementedError,
             )
         try:
-            return input_data[exp.name]
+            if exp.name in input_data:
```
sfc-gh-stan, line 1447: Does this mean there is a mixture of quoted and unquoted names here? How do we distinguish between a column that is explicitly quoted? For example:
```
    [[1,2],[3,4]], schema=['col1', '"col2"']
)

# df['col1'1] should work
# df['"col1"'] should not work
# df['"col2"'] should work
# df['col2'] should not work
```

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +448,104 @@ def read_file(
...
+                assert len(schema) == 1, (
+                    f"[Local Testing] Unexpected schema length {len(schema)} when loading "
+                    f"json data without inferring schema."
+                )
+                column_name = analyzer.analyze(schema[0])
+                column_series = ColumnEmulator(
+                    data=None,
+                    dtype=object,
+                    name=column_name,
+                    sf_type=ColumnType(schema[0].datatype, schema[0].nullable),
```
sfc-gh-stan, line 469: Shouldn't the datatype always be `Variant` in this case? Does specifying other data types work?

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +448,104 @@ def read_file(
...
+                # fill empty cells with None value, this aligns with snowflake
+                for content in contents:
+                    for miss_key in set(result_df_sf_types.keys()) - set(
+                        content.keys()
+                    ):
+                        content[miss_key] = None
+                    df = pd.DataFrame([content])
+                    result_df = pd.concat([result_df, df], ignore_index=True)
+
+                def convert_value_to_str_for_string_type(dataframe):
```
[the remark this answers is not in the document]
sfc-gh-aling: good idea, I have made the necessary code change in to_char to make it work for my case.

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-aling: I tested and you are right! changed to use variant type and set nullable to True (this is my test observation that snowflake always set it to True)

src/snowflake/snowpark/mock/_plan.py
sfc-gh-aling: good catch, after doing some test, snowflake always add double quotes to the col name when loading json.

I removed the logic in the _plan.py, and add the logic of handling column name (add double quotes/escape double quotes) to the read json function.

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -766,7 +766,11 @@ def mock_to_char(
             )
         func = partial(try_convert, lambda x: str(x), try_cast)
     else:
-        func = partial(try_convert, lambda x: str(x), try_cast)
+        func = partial(
```
sfc-gh-stan, line 769: I prefer if we add an additional case for `BooleanType` where we use `str(x).lower()`
and an additional case for `VariantType` where we use `json.dumps(cls=CustomJsonEncoder)`.

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +449,96 @@ def read_file(
...
+                        result_df = pd.concat([result_df, df], ignore_index=True)
+            else:
+                # need to infer schema
+                contents = []
+                for local_file in local_files:
+                    with open(local_file) as file:
+                        content = json.load(file, cls=CUSTOM_JSON_DECODER)
+                        tmp_content = {}
+                        # snowflake escape double quotes by adding extra double quote
+                        for key, value in content.items():
```
sfc-gh-stan, line 490: This makes sense 👍 , can we reuse https://github.com/snowflakedb/snowpark-python/blob/bf4a409fb78816568de69e57da6e2bf6919643d7/src/snowflake/snowpark/_internal/utils.py#L855?

src/snowflake/snowpark/mock/_stage_registry.py
```diff
@@ -413,6 +449,96 @@ def read_file(
...
+                    df = TableEmulator([content])
+                    result_df = pd.concat([result_df, df], ignore_index=True)
+                # when concat is called, sf_type information gets lost, so we reset the type info in the end
+                result_df.sf_types = result_df_sf_types
+
+                # in the case that there are values of different types in the same column, snowflake will
+                # convert data into string
+                for col_name in result_df.columns:
+                    if isinstance(result_df_sf_types[col_name].datatype, StringType):
+                        result_df[col_name] = mock_to_char(result_df[col_name])
```
sfc-gh-stan, line 534: Since we the column is of mixed types now, lets's set `result_df_sf_types[col_name].datatype` to `VariantType` then use `mock_to_char`.

src/snowflake/snowpark/mock/_functions.py
sfc-gh-aling: good idea, updated

src/snowflake/snowpark/mock/_stage_registry.py
sfc-gh-aling: good idea, code updated

sfc-gh-stan (approved):
LGTM, thanks for the great work Adam!

sfc-gh-jrose (approved):

PR #1316, opened 2024-03-15
SNOW-1183317: [Local Testing] Support registering udf

sfc-gh-jrose:

   Fixes SNOW-1183317

   This PR does a few things to add support for UDFs in local testing:
   - Adds a new MockUDFRegistration class that is used to register UDFs when in local testing mode. It is a child of the base UDFRegistration and just overwrites parts that need to be local only.
   - Adds logic to the mock _plan handler to use the newly registrable udfs as needed. As part of that I also added some logic to the function_expression handler so that it would instead treat a function as a udf as needed.
   - Moved some quoting logic from the plan describe function into the mock analyzer function. Some of the udf function columns were not resolving correctly without the logic being there.
   - Refactored the test data for udf tests to use df operations instead of sql so that it can be created with local testing.
     - As part of that refactor I noticed some issues with the `to_date` local testing function that made it not use fmt strings correctly. I did a light refactor to that function so that it can recognize both string inputs and column inputs.


src/snowflake/snowpark/mock/_plan.py
```diff
@@ -388,6 +397,32 @@ def handle_function_expression(
...
+            f"[Local Testing] udf {exp.udf_name} does not exist."
+        )
+
+    to_pass_args = [
+        calculate_expression(child, input_data, analyzer, expr_to_alias).name
+        for child in exp.children
+    ]
+
+    res = input_data.apply(lambda row: udf(*[row[col] for col in to_pass_args]), axis=1)
+    res.sf_type = ColumnType(exp.datatype, exp.nullable)
```
sfc-gh-stan, line 420: Is `exp.datatype`, `exp.nullable` filled here?

src/snowflake/snowpark/mock/_udf.py
```diff
@@ -0,0 +1,97 @@
...
+        immutable: bool = False,
+        *,
+        statement_params: Optional[Dict[str, str]] = None,
+        source_code_display: bool = True,
+        api_call_source: str,
+        skip_upload_on_content_match: bool = False,
+        is_permanent: bool = False,
+    ) -> UserDefinedFunction:
+        if is_permanent:
+            raise NotImplementedError(
```
sfc-gh-stan, line 54: Let's use `MockConnection.log_not_supported_error` for Local Testing telemetry.

sfc-gh-stan (changes requested):

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -319,6 +320,14 @@ def handle_function_expression(
...
         func_name = exp.name.lower()
 
+    udf_name = exp.name.split(".")[-1]
+    # If udf name in the registry then this is a udf, not an actual function
+    if udf_name in analyzer.session.udf._registry:
+        exp.udf_name = udf_name
+        return handle_udf_expression(
+            exp, input_data, analyzer, expr_to_alias, current_row
+        )
+
```
sfc-gh-aling, line 330: nit: just curious, can a udf be defined using the same signature of a built-in function, what would happen?

src/snowflake/snowpark/mock/_plan.py
sfc-gh-aling: I also have the question that is exp the same as the udf definition (what is returned as columns)?

src/snowflake/snowpark/mock/_udf.py
```diff
@@ -0,0 +1,97 @@
...
+        if not custom_python_runtime_version_allowed:
+            check_python_runtime_version(
+                self._session._runtime_version_from_requirement
+            )
+
+        if udf_name in self._registry:
+            raise SnowparkSQLException(
+                f"002002 (42710): SQL compilation error: \nObject '{udf_name}' already exists.",
+                error_code="1304",
+            )
```
sfc-gh-aling, line 93: the parameter `replace` allows you to overwrite the udf

src/snowflake/snowpark/mock/__init__.py
```diff
@@ -1,12 +1,32 @@
...
 # Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
 #
+from json import JSONEncoder
+
 from ._functions import patch
 from ._snowflake_data_type import ColumnEmulator, ColumnType, TableEmulator
 
-CUSTOM_JSON_ENCODER = None
+
+class NumpyEncoder(JSONEncoder):
```
sfc-gh-aling, line 10: may I ask what is this change for?

src/snowflake/snowpark/mock/_udf.py
sfc-gh-aling: good catch, let's replace `NotImplementedError` with `log_not_supported_error` in other places as well

src/snowflake/snowpark/session.py
```diff
@@ -437,7 +437,12 @@ def __init__(
 """
         self._session_stage = random_name_for_temp_object(TempObjectType.STAGE)
         self._stage_created = False
-        self._udf_registration = UDFRegistration(self)
+
+        if options is not None and options.get("local_testing", False):
+            self._udf_registration = MockUDFRegistration(self)
+        else:
+            self._udf_registration = UDFRegistration(self)
```
sfc-gh-aling, line 444: this is for future discussion but want to collect your thoughts:

in general I want to avoid if-else checking (as few as possible) in the code to decide what implementation to use when switching between localtest and non-localtesting.

one potential approach is to subclass Session. do you see any other approaches here?

src/snowflake/snowpark/mock/__init__.py
sfc-gh-jrose: This [chunk of code](https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/mock/_connection.py#L480-L482) was failing due to being unable to serialize numpy datatypes.

src/snowflake/snowpark/mock/_plan.py
sfc-gh-jrose: I tried this out on live and it looks like if you shadow a builtin it it will favor the builtin over your udf. I can change this so that the priority matches.

src/snowflake/snowpark/session.py
sfc-gh-jrose: I have some other ideas. Let's talk in person about it sometime.

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -388,6 +397,32 @@ def handle_function_expression(
...
+    current_row=None,
+):
+    udf = analyzer.session.udf._registry.get(exp.udf_name)
+
+    if udf is None:
+        raise SnowparkSQLException(
+            f"[Local Testing] udf {exp.udf_name} does not exist."
+        )
+
+    to_pass_args = [
```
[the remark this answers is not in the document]
sfc-gh-jrose: I've added a test case for that specific scenario and it appears to work. Are there any other edge cases I should be careful about?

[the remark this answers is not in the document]
sfc-gh-stan: I'm referring to when inputs of a UDF contains columns that doesn't exist in the `input_data`, for the added test case `test_child_expression`, if we create a udf `double_udf` and evaluate `df.collect(double_udf(col("A") + col("A"))`, this will raise a `KeyError` because `col("A")+col("A")` does not exist in `df`.

Please see here in the stacked PR (support registering from file): https://github.com/snowflakedb/snowpark-python/pull/1325/files#diff-fb4a9902d599442f4f106211e1a05a3de511c8009eb7174d0c96c25aa4c2cf98R415, we need to evaluate the children expressions first.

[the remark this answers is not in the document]
sfc-gh-stan: Note: my change doesn't handle when the function input is empty, when udf has no parameters.

sfc-gh-stan (approved):

src/snowflake/snowpark/mock/__init__.py
sfc-gh-aling: that part of code is for use_xxx method on the session object, how would a UDF call trigger it?

src/snowflake/snowpark/mock/_udf.py
```diff
@@ -0,0 +1,103 @@
...
+                "You cannot create a non-vectorized UDF using pandas_udf(). "
+                "Use udf() instead."
+            )
+
+        if packages or imports:
+            self._session._conn.log_not_supported_error(
+                external_feature_name="udf",
+                error_message="Uploading imports and packages not currently supported.",
+                raise_error=NotImplementedError,
+            )
```
sfc-gh-aling, line 86: we will support this soon I believe

sfc-gh-aling (approved):

PR #1323, opened 2024-03-18
SNOW-1000284: Add schema support for structure types

sfc-gh-jrose:

   Fixes #SNOW-1000284

   - [x] I am adding a new dependency

   This PR adds relevant changes needed to support Structured Types. The goal of this commit is to enable reading structured data from iceberg tables. Writing new tables and support for non-iceberg tables is a future goal of the project.

The change does the following:

1. metadata that contains fields for relevant types (MAP, OBJECT, ARRAY) signals that structured types are enabled and being returned.
2. When receiving that metadata the relevant types are marked as structured types using a new new parameter. Users that would like to make new dataframes with structured types will similarly need to set the structured_types parameter on the schema objects. In the future we will enable this behavior by default.
3. When marked as structured the relevant types will generate slightly different queries for sprocs and udfs which allows them to specify data is structured for both input and output.

I've added test cases that test the following:

1. Returned data for iceberg and non-iceberg data can be handled with the new metadata. We don't expect non-iceberg structured data yet, but it couldn't hurt to be prepared for it.
2. Sprocs can take structured data as input and provide it as output. The same goes for udfs.
3. Subfields in structured objects/maps and indexes in structured arrays can be accessed with regular dataframe semantics.
4. We're able to return structured data from requests for pandas dataframes from iceberg tables.

Merge Notes:
This change cannot merge until the next python-connector version is released. I will need to add it as a dependency to support the new pandas changes.

This change also will not pass merge gates until I figure out how to add permissions to the test runner in order to access the external volume that we're using for iceberg tests. Maybe @sfc-gh-yixie can help me figure that out.


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -144,6 +175,8 @@ def convert_sf_to_sp_type(
         return VariantType()
     if column_type_name == "OBJECT":
         return MapType(StringType(), StringType())
+    if column_type_name == "MAP":
```
sfc-gh-jdu, line 178:
```suggestion
    if column_type_name in ["MAP", "OBJECT"]:
```

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1163,7 +1163,7 @@ def create_python_udf_or_sp(
 
     if replace and if_not_exists:
         raise ValueError("options replace and if_not_exists are incompatible")
-    if isinstance(return_type, StructType):
+    if isinstance(return_type, StructType) and return_type.structured is False:
```
sfc-gh-jdu, line 1166:
```suggestion
    if isinstance(return_type, StructType) and not return_type.structured:
```

tests/integ/scala/test_datatype_suite.py
```diff
@@ -229,6 +294,170 @@ def test_dtypes(session):
     ]
 
 
+@pytest.mark.parametrize(
```
sfc-gh-jdu, line 297: should we also skip these tests in stored proc? Does stored proc connector have the corresponding change for struct type?

tests/utils.py
```diff
@@ -76,6 +76,12 @@
 IS_NOT_ON_GITHUB = os.getenv("GITHUB_ACTIONS") != "true"
 # this env variable is set in regression test
 IS_IN_STORED_PROC_LOCALFS = IS_IN_STORED_PROC and os.getenv("IS_LOCAL_FS")
+STRUCTURED_TYPE_ENVIRONMENTS = {"dev", "aws"}
+IS_STRUCTURED_TYPES_SUPPORTED = (
+    os.getenv("cloud_provider", "dev") in STRUCTURED_TYPE_ENVIRONMENTS
```
sfc-gh-jdu, line 81: cloud_provider is set on github, right? We also have some tests running on jenkins, so they will just be skipped on jenkins?

tests/integ/scala/test_datatype_suite.py
```diff
@@ -33,7 +36,69 @@
...
+                    nullable=True,
+                ),
+                StructField(
+                    "ARR", ArrayType(DoubleType(), structured=True), nullable=True
+                ),
+            ]
+        ),
+        id="structured-types-enabled",
+    ),
+    False: pytest.param(
```
sfc-gh-jdu, line 80: we finally will remove this example once all deployments supporting struct types?

sfc-gh-jrose: Yes, at some point in the future we will make a BCR that turns structured types on by default. That will probably be the best time to remove this.

tests/integ/scala/test_datatype_suite.py
sfc-gh-jrose: By the time I merge this change it should have the corresponding change.

tests/utils.py
sfc-gh-jrose: I'm not sure how that gets set. I know that in tox.ini it's passed in regardless of environment, but that doesn't mean it is set.

sfc-gh-jdu: so we will skip these tests on jenkins right now? e.g., https://ci-dev-142.int.snowflakecomputing.com/job/SnowparkPythonClientRegressRunner/

sfc-gh-jrose: Yes. Once the parameter for enabling structured types is on by default we will need to make sure these tests work in all environments, but for now sfctest0 is the only environment that has them enabled. It is also the only environment that we have iceberg test infrastructure set up.

sfc-gh-jdu: got it, thanks. Then maybe create a jira and add a TODO here to make sure these conditions will be finally removed?

sfc-gh-jdu (approved):

tests/integ/scala/test_datatype_suite.py
```diff
@@ -33,7 +36,69 @@
...
+        [
+            ("MAP", "map<string(16777216),bigint>"),
+            ("OBJ", "struct<string(16777216),double>"),
+            ("ARR", "array<double>"),
+        ],
+        StructType(
+            [
+                StructField(
+                    "MAP",
+                    MapType(StringType(16777216), LongType(), structured=True),
```
sfc-gh-yixie, line 59: Do we have to explicit use the length 16777216? This number may change soon.

PR #1326, opened 2024-03-21
SNOW-1064297: feature session support add packages and add requirements

sfc-gh-aling:

   Fixes SNOW-1064297

implement local testing session.add_packages and add_requirements.

for local testing, we just parse the input packages and record them in the session._package_dict, and we don't do resolve the packages. The assumption here is that users will have all packages installed in their own env.

we do not need to implement local testing version of add_requirements, clear_packages, remove_package because all of them are already local operations.


sfc-gh-jrose (approved):

src/snowflake/snowpark/session.py
```diff
@@ -1320,6 +1315,27 @@ def _resolve_packages(
...
+                        ValueError(
+                            f"Cannot add package '{str(pkg_req)}' because {self._packages[pkg_name]} "
+                            "is already added."
+                        )
+                    )
+                else:
+                    self._packages[pkg_name] = str(pkg_req)
+            if len(errors) == 1:
+                raise errors[0]
+            elif len(errors) > 0:
```
sfc-gh-stan, line 1336: Question: why do we need this case?

sfc-gh-stan (changes requested):
Please remove `pytest.mark.localtest` for the tests that depend on UDF execution, it's better to enable those after UDF support is merged, otheriwse we need to have a task to see if we can rewrite all the `if local_testing_mode` branches in tests.

For tests that don't depend on UDF execution, can we move those to unit tests?

sfc-gh-stan:
Also please fill in the description.

src/snowflake/snowpark/session.py
sfc-gh-aling: this is copied and pasted from the existing package resolving case: https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/session.py#L1270-L1274

```python
                    except Exception as ex:  # pragma: no cover
                        logging.warning(
                            "Failed to get the local distribution of package %s: %s",
                            package_name,
                            ex,
                        )

            if package_name in current_packages:
                if current_packages[package_name] != package:
                    errors.append(
                        ValueError(
                            f"Cannot add package '{package}' because {current_packages[package_name]} "
                            "is already added."
                        )
                    )
            else:
                current_packages[package_name] = package

        # Raise all exceptions at once so users know all issues in a single invocation.
        if len(errors) == 1:
            raise errors[0]
        elif len(errors) > 0:
            raise RuntimeError(errors)
```

sfc-gh-aling:
@sfc-gh-stan we can't simply move them as those are not unit tests -- the process of live package resolving needs to go to snowflake to fetch all package information.

sfc-gh-stan (approved):

PR #1329, opened 2024-03-22
SNOW-1184278 [Local Testing] Support register from file, Python import, session.add_import and session.clear_imports

sfc-gh-stan:

   Fixes SNOW-1184277 and SNOW-1184278 and SNOW-1064303 and SNOW-1064305

   This PR includes the changes that support:
    - registering a UDF from file
    - adding a local Python file as a UDF import at session/udf level
    - adding a remote Python file on stage as a UDF import at session/udf level
    - adding a local Python directory as a UDF import at session/udf level
    - adding a local zip as a UDF import at session/udf level
    - Session.add_import
    - Session.clear_imports


tests/integ/test_udf.py
```diff
@@ -409,10 +408,8 @@ def test_session_register_udf(session):
     )
 
 
-@pytest.mark.skipif(
-    not is_pandas_available, reason="pandas is required to register vectorized UDFs"
-)
```
sfc-gh-aalam, line 414: I don't understand why this is being removed

tests/integ/test_udf.py
```diff
@@ -543,126 +565,125 @@ def test_register_from_file_with_skip_upload(session, resources_path, caplog):
...
 
+@pytest.mark.localtest
 def test_add_import_local_file(session, resources_path):
     test_files = TestFiles(resources_path)
-    # This is a hack in the test such that we can just use `from test_udf import mod5`,
-    # instead of `from test.resources.test_udf.test_udf import mod5`. Then we can test
-    # `import_as` argument.
-    with patch.object(
-        sys, "path", [*sys.path, resources_path, test_files.test_udf_directory]
-    ):
```
sfc-gh-aalam, line 553: I don't have context on why this hack was added and why is it removed now. Can you give more context here

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -404,12 +405,31 @@ def handle_udf_expression(
...
+    udf_name = exp.udf_name
 
-    if udf is None:
-        raise SnowparkSQLException(
-            f"[Local Testing] udf {exp.udf_name} does not exist."
-        )
+    if udf_name not in udf_registry:
+        raise SnowparkSQLException(f"[Local Testing] udf {udf_name} does not exist.")
+
+    frozen_sys_module_keys = set(sys.modules.keys())
```
sfc-gh-jrose, line 414: I think this should be handled with a try/finally maybe in a context manager. I think there are scenarios where this code could throw a recoverable error and we should be sure the path isn't messed up in that case.

tests/integ/test_udf.py
sfc-gh-stan: This was only necessary because the original test contains a pandas UDF registration. I have split the test into 5 tests below, `test_register_vectorized_udf_from_file` has the `skipif` makrer.

tests/integ/test_udf.py
sfc-gh-jdu: As long as the test passes, then it should be ok. Previously I remember if I don't add this patch, the test will fail

sfc-gh-jrose (approved):
Looks good 😃

src/snowflake/snowpark/mock/_plan.py
sfc-gh-aalam: +1

src/snowflake/snowpark/mock/_udf.py
```diff
@@ -10,21 +11,71 @@
...
+        """
+        Imports a python file or a directory of python module structure or a zip of the former.
+        Returns the name of the Python module to be imported.
+        When udf_name is not None, the import is added to the UDF associated with the name;
+        Otherwise, it is a session level import and will be added to every future UDF call.
+        """
+        file_name, file_extension = os.path.splitext(os.path.basename(file_path))
+
+        if file_path.startswith("@"):  # file is on a stage
+            stage_registry = self._session._conn.stage_registry
```
sfc-gh-aalam, line 48: maybe add `snow://` too here:

https://github.com/snowflakedb/snowpark-python/pull/1346

sfc-gh-aalam (approved):

src/snowflake/snowpark/mock/_udf.py
sfc-gh-stan: I created https://snowflakecomputing.atlassian.net/browse/SNOW-1285387 to track this

PR #1332, opened 2024-03-25
[Draft][Do Not Merge] Refactoring UDF Annotation Code

sfc-gh-bgoel:

This refactoring is being introduced so that it is easier to inspect the functions and customize behavior depending on the execution context. 
   Fixes #NNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -39,7 +39,6 @@ from snowflake.snowpark._internal.type_utils import (
 )
 from snowflake.snowpark._internal.utils import (
     STAGE_PREFIX,
-    CallableProperties,
     TempObjectType,
     get_udf_upload_prefix,
     is_single_quoted,
@@ -49,7 +48,7 @@ from snowflake.snowpark._internal.utils import (
     unwrap_stage_location_single_quote,
     validate_object_name,
 )
-from snowflake.snowpark.types import DataType, StructField, StructType
+from snowflake.snowpark.types import DataType, NullType, StructField, StructType
 
 if installed_pandas:
     from snowflake.snowpark.types import (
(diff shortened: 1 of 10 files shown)


src/snowflake/snowpark/context.py
```diff
@@ -8,6 +8,8 @@
 
 _use_scoped_temp_objects = True
 
+#
+
```
sfc-gh-bgoel, line 12: This will be the place to introduce the hook from Snowflake CLI, as well as have a global flag that allows us to execute code in a "local sandbox" environment, i.e. not interact with Snowflake.

src/snowflake/snowpark/udf.py
```diff
@@ -66,21 +70,32 @@ class UserDefinedFunction:
         - :func:`~snowflake.snowpark.functions.udf`
     """
 
+    # def __init__(
```
sfc-gh-bgoel, line 73: Temporarily commented to show the usefulness of the `CallableProperties`. Once that implementation is accepted, this old `__init__` will be removed in favor of the new one.

src/snowflake/snowpark/udf.py
```diff
@@ -897,3 +915,115 @@ def _do_register_udf(
                 )
 
         return UserDefinedFunction(func, return_type, input_types, udf_name)
+
+    def __do_register_udf(
```
sfc-gh-bgoel, line 919: Temporary naming since python does not have overloading feature. Once the new implementation has been reviewed, the older `_do_register_udf` will be removed.

src/snowflake/snowpark/udf.py
```diff
@@ -897,3 +915,115 @@ def _do_register_udf(
...
+        )
+
+        # allow registering pandas UDF from udf(),
+        # but not allow registering non-pandas UDF from pandas_udf()
+        if from_pandas_udf_function and not is_pandas_udf:
+            raise ValueError(
+                "You cannot create a non-vectorized UDF using pandas_udf(). "
+                "Use udf() instead."
+            )
+
```
sfc-gh-bgoel, line 960: For lines 961 - 984, it breaks out the original `resolve_imports_and_packages` into self contained package and imports resolution.

src/snowflake/snowpark/udf.py
```diff
@@ -897,3 +915,115 @@ def _do_register_udf(
...
+        )
+
+        callableProperties.setResolvedHandler(handler)
+        callableProperties.setResolvedInlineCode(inline_code)
+        callableProperties.setResolvedImports(all_imports)
+
+        if not custom_python_runtime_version_allowed:
+            check_python_runtime_version(
+                self._session._runtime_version_from_requirement
+            )
```
sfc-gh-bgoel, line 993: 990-993 has been moved up from its original location in `create_python_udf_or_sp` so that all property resolution is handled in this method, and`create_python_udf_or_sp` is only responsible for transforming existing properties into sql strings and running the query.

src/snowflake/snowpark/udf.py
```diff
@@ -897,3 +915,115 @@ def _do_register_udf(
...
+
+        runtime_version = (
+            f"{sys.version_info[0]}.{sys.version_info[1]}"
+            if not self._session._runtime_version_from_requirement
+            else self._session._runtime_version_from_requirement
+        )
+        callableProperties.setResolvedRuntimeVersion(runtime_version)
+
+        raised = False
```
sfc-gh-bgoel, line 1002: At this point, the hook introduced by the CLI will be invoked to send the `callableProperties` data back to the CLI, instead of invoking `create_python_udf_or_sp`.

src/snowflake/snowpark/udf.py
```diff
@@ -897,3 +915,115 @@ def _do_register_udf(
...
+        except BaseException:
+            raised = True
+            raise
+        finally:
+            if raised:
+                cleanup_failed_permanent_registration(
+                    self._session, upload_file_stage_location, stage_location
+                )
+
+        return UserDefinedFunction(callableProperties)
```
sfc-gh-bgoel, line 1029: This will be executed regardless the context/hook.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -977,6 +978,322 @@ def resolve_imports_and_packages(
...
+    stage_location: Optional[str],
+    callableProperties: CallableProperties,
+    imports: Optional[List[Union[str, Tuple[str, str]]]],
+    arg_names: List[str],
+    statement_params: Optional[Dict[str, str]] = None,
+    is_dataframe_input: bool = False,
+    skip_upload_on_content_match: bool = False,
+    is_pandas_udf: bool = False,
+) -> Tuple[str, str, str, str, bool]:
+
```
sfc-gh-bgoel, line 1018: `session.add_imports` performs a lazy execution, i.e. only adds imports provided by the user to a running list, and does not actually interact with Snowflake. With the CLI local sandbox, we will do something similar here, instead of the implementation below.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -977,6 +978,322 @@ def resolve_imports_and_packages(
...
+    statement_params: Optional[Dict[str, str]] = None,
+) -> str:
+
+    # resolve packages
+    resolved_packages = (
+        session._resolve_packages(
+            packages, include_pandas=is_pandas_udf, statement_params=statement_params
+        )
+        if packages is not None
+        else session._resolve_packages(
```
sfc-gh-bgoel, line 994: `_resolve_packages` includes calls to Snowflake for getting `information_schema.packages` view and for fetching and/or uploading packages to a Snowflake stage. With the CLI local sandbox, we want to short circuit this and exit before interaction with Snowflake. We only want to parse packages and check them against the local environment of the user, and return from there. This is because we are not registering anything with Snowflake and hence do not need to resolve any of the packages, supported or unsupported.

src/snowflake/snowpark/udf.py
```diff
@@ -66,21 +70,32 @@ class UserDefinedFunction:
...
+    #     self._input_types = input_types
+    #     self._is_return_nullable = is_return_nullable
+
     def __init__(
         self,
-        func: Union[Callable, Tuple[str, str]],
-        return_type: DataType,
-        input_types: List[DataType],
-        name: str,
+        callableProperties: CallableProperties,
```
sfc-gh-bdufour, line 92: Technically UserDefinedFunction is a public type, which makes this a backwards-incompatible change. Check with Snowpark whether they're ok changing this ctor. Otherwise, you'd want to add anything you need after the last parameter, and make sure there is a default value assigned.

src/snowflake/snowpark/udf.py
```diff
@@ -897,3 +915,115 @@ def _do_register_udf(
                 )
 
         return UserDefinedFunction(func, return_type, input_types, udf_name)
```
sfc-gh-bdufour, line 917: Again, just edit the code. This instantiation is broken now because you changed the ctor. You're using git, no need to keep old dead code lying around.

src/snowflake/snowpark/udf.py
```diff
@@ -897,3 +915,115 @@ def _do_register_udf(
...
+            object_type=callableProperties.object_type,
+            func=callableProperties.func,
+            return_type=callableProperties.raw_return_type,
+            input_types=callableProperties.raw_input_types,
+            name=callableProperties.raw_name,
+        )
+
+        callableProperties.setValidatedObjectName(udf_name)
+        callableProperties.setValidatedReturnType(return_type)
+        callableProperties.setValidatedInputTypes(input_types)
```
sfc-gh-bdufour, line 946: Really not loving the fact that the input param sneakily gets mutated here. Should this logic be moved to methods on the CallableProperties object itself? It would make the mutation a lot more obvious IMHO. Something like:
`callableProperties.processRegistrationInputs(session)` instead of the logic above (and similarly for the other pieces)

src/snowflake/snowpark/udf.py
sfc-gh-bgoel: That, or to keep this public class as is (and other classes too like SPROC, UDxF), the UDF creation in udf.py can be changed to accept the specific instance variables of `callableProperties` instead of passing in the entire object itself.

src/snowflake/snowpark/udf.py
sfc-gh-bdufour: I understand that, but from a maintenance POV, modifying an object that's passed as a parameter in the middle of other logic is usually a surprising side effect. The way this method is named, it says "perform the registration using the provided properties", but then the object gets mutated along the way.

sfc-gh-bgoel:
Closing this PR in favor of #1341

PR #1335, opened 2024-03-25
SNOW-1208807 [Local Testing] Support importing non Python file in UDF and access from snowflake_import_directory

sfc-gh-stan:

   Fixes SNOW-1208807

   This PR supports non Python imports in UDF by creating a temporary directory per UDF execution. All imported files will be copied to this directory, users can access these files by reading the value of `sys._xoptions.get("snowflake_import_directory")` .
   Example usage in official documentation: https://docs.snowflake.com/en/developer-guide/udf/python/udf-python-examples#reading-a-statically-specified-file-using-imports


src/snowflake/snowpark/mock/_plan.py
```diff
@@ -443,22 +468,32 @@ def handle_udf_expression(
     res.name = quote_name(f"{exp.udf_name}({', '.join(input_data.columns)})".upper())
 
     if udf_name in analyzer.session.udf._udf_level_imports:
+        # Remove UDF level imports
         for module_path in analyzer.session.udf._udf_level_imports[udf_name]:
             if (
                 module_path in sys.path
-            ):  # We need this additional check since children expression might temper with sys.path as well, this needs some future work
+            ):  # TODO: For now we need this additional check since children expression can write to sys.path as well
```
sfc-gh-jrose, line 475: When will this no longer be the case?

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -411,26 +414,48 @@ def handle_udf_expression(
     if udf_name not in udf_registry:
         raise SnowparkSQLException(f"[Local Testing] udf {udf_name} does not exist.")
 
+    # Initialize import directory
```
sfc-gh-jrose, line 417: If a udf is called multiple times in the same query will the modules be copied for each invocation?

src/snowflake/snowpark/mock/_plan.py
sfc-gh-stan: When we work on adding more isolation, e.g. run UDF in a separate process.

src/snowflake/snowpark/mock/_plan.py
sfc-gh-stan: Yes, we could potentially optimize by initializing the import directory at UDF registration.

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -411,26 +414,48 @@ def handle_udf_expression(
...
+    temporary_import_path = tempfile.TemporaryDirectory()
+    analyzer.session.udf._udf_import_directories[udf_name] = temporary_import_path
+    last_import_directory = sys._xoptions.get("snowflake_import_directory")
+    sys._xoptions["snowflake_import_directory"] = temporary_import_path.name
```
sfc-gh-jrose, line 421: I think this should be handled with try/finally or a context manager.

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -443,32 +454,56 @@ def cleanup_imports():
...
             del sys.modules[key]
 
+        # Cleanup import directory
+        analyzer.session.udf._udf_import_directories[udf_name].cleanup()
+
+        # Restore snowflake_import_directory
+        if last_import_directory is not None:
+            sys._xoptions["snowflake_import_directory"] = last_import_directory
+        else:
+            del sys._xoptions["snowflake_import_directory"]
```
sfc-gh-aling, line 469: what is the `Restore` process here?

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -430,10 +433,18 @@ def handle_udf_expression(
     if udf_name not in udf_registry:
         raise SnowparkSQLException(f"[Local Testing] udf {udf_name} does not exist.")
 
+    # Initialize import directory
+    temporary_import_path = tempfile.TemporaryDirectory()
+    analyzer.session.udf._udf_import_directories[udf_name] = temporary_import_path
+    last_import_directory = sys._xoptions.get("snowflake_import_directory")
```
sfc-gh-aling, line 439: do we actually use `last_import_directory`, is the settable by user?

sfc-gh-aling (commented):
only having some questions about the `last_import_directory` part, other parts look good to me

src/snowflake/snowpark/mock/_plan.py
sfc-gh-stan: We need `last_import_directory` when a UDF's input is another UDF's results. On line 510, we calculate the child expression as function input, when this involves another UDF expression, the child UDF should be computed with the corresponding import directory but after its execution this should be restored to the last UDF execution on the stack, so we need to save `last_import_directory` in each `handle_udf_expression` call and restore it at the end.

sfc-gh-aling: I see, that makes more sense to me now, thanks for the explanation.

sfc-gh-aling (approved):

PR #1338, opened 2024-03-26
SNOW-1183322: [Local Testing] Add support for registering sprocs

sfc-gh-jrose:

   Fixes #1183322

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
     - file.get_stream
   - snowflake.snowpark.functions
     - udf
+    - sproc
 - Added the function `DataFrame.write.csv` to unload data from a ``DataFrame`` into one or more CSV files in a stage.
 
 ## 1.14.0 (2024-03-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
+- Added support for registering udfs and stored procedure to local testing.
 - Added support for the following local testing APIs:
   - snowflake.snowpark.Session:
     - file.put
@@ -28,8 +29,6 @@
     - object_construct
     - object_construct_keep_null
     - pow
-    - udf
-    - sproc
     - sqrt
 - Added the function `DataFrame.write.csv` to unload data from a ``DataFrame`` into one or more CSV files in a stage.
 - Added distributed tracing using open telemetry apis for action functions in `DataFrame` and `DataFrameWriter`:
--- a/src/snowflake/snowpark/mock/_stored_procedure.py
+++ b/src/snowflake/snowpark/mock/_stored_procedure.py
@@ -8,6 +8,7 @@ from types import ModuleType
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import snowflake.snowpark
+from snowflake.snowpark._internal.analyzer.expression import Literal
 from snowflake.snowpark._internal.udf_utils import (
     check_python_runtime_version,
     process_registration_inputs,
@@ -15,13 +16,13 @@ from snowflake.snowpark._internal.udf_utils import (
 from snowflake.snowpark._internal.utils import TempObjectType
 from snowflake.snowpark.column import Column
 from snowflake.snowpark.exceptions import SnowparkSQLException
-from snowflake.snowpark.mock._plan import calculate_expression
-from snowflake.snowpark.mock._snowflake_data_type import ColumnEmulator
 from snowflake.snowpark.stored_procedure import (
     StoredProcedure,
     StoredProcedureRegistration,
 )
 from snowflake.snowpark.types import DataType, _FractionalType, _IntegralType
+
+from ._telemetry import LocalTestOOBTelemetryService
 
 if sys.version_info <= (3, 9):
     from typing import Iterable
@@ -57,19 +58,14 @@ class MockStoredProcedure(StoredProcedure):
                     raise ValueError(
                         f"Unexpected type {expr.datatype} for sproc argument of type {expected_type}"
                     )
-                parsed_args.append(
-                    calculate_expression(
-                        expr,
-                        ColumnEmulator(data=[None]),
-                        session._analyzer,
-                        {},
-                        keep_literal=True,
+
+                if not isinstance(expr, Literal):
+                    raise ValueError(
+                        "[Local Testing] Unexpected argument type {expr.__class__.__name__} for call to sproc"
                     )
-                )
+                parsed_args.append(expr.value)
             else:
                 parsed_args.append(arg)
-
-        # import pdb; pdb.set_trace()
 
         return self.func(session, *parsed_args)
 
@@ -102,8 +98,11 @@ class MockStoredProcedureRegistration(StoredProcedureRegistration):
         source_code_display: bool = True,
         skip_upload_on_content_match: bool = False,
     ) -> StoredProcedure:
-        raise NotImplementedError(
-            "[Local Testing] Registering stored procedure from file is not currently supported."
+        LocalTestOOBTelemetryService.get_instance().log_not_supported_error(
+            external_feature_name="register sproc from file",
+            internal_feature_name="MockStoredProcedureRegistration.register_from_file",
+            parameters_info={},
+            raise_error=NotImplementedError,
         )
 
     def _do_register_sp(
@@ -151,13 +150,16 @@ class MockStoredProcedureRegistration(StoredProcedureRegistration):
             raise TypeError("pandas stored procedure is not supported")
 
         if packages or imports:
-            raise NotImplementedError(
-                "[Local Testing] Uploading imports and packages not currently supported."
+            LocalTestOOBTelemetryService.get_instance().log_not_supported_error(
+                external_feature_name="uploading imports and packages for sprocs",
+                internal_feature_name="MockStoredProcedureRegistration._do_register_sp",
+                parameters_info={},
+                raise_error=NotImplementedError,
             )
 
         check_python_runtime_version(self._session._runtime_version_from_requirement)
 
-        if udf_name in self._registry:
+        if udf_name in self._registry and not replace:
             raise SnowparkSQLException(
                 f"002002 (42710): SQL compilation error: \nObject '{udf_name}' already exists.",
                 error_code="1304",
--- a/src/snowflake/snowpark/mock/_stored_procedure.py
+++ b/src/snowflake/snowpark/mock/_stored_procedure.py
@@ -1,0 +1,174 @@
+#
+# Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
+#
+
+import sys
+import typing
+from types import ModuleType
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+import snowflake.snowpark
+from snowflake.snowpark._internal.udf_utils import (
+    check_python_runtime_version,
+    process_registration_inputs,
+)
+from snowflake.snowpark._internal.utils import TempObjectType
+from snowflake.snowpark.column import Column
+from snowflake.snowpark.exceptions import SnowparkSQLException
+from snowflake.snowpark.mock._plan import calculate_expression
+from snowflake.snowpark.mock._snowflake_data_type import ColumnEmulator
+from snowflake.snowpark.stored_procedure import (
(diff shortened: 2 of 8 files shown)


src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -0,0 +1,174 @@
...
+                        ColumnEmulator(data=[None]),
+                        session._analyzer,
+                        {},
+                        keep_literal=True,
+                    )
+                )
+            else:
+                parsed_args.append(arg)
+
+        # import pdb; pdb.set_trace()
```
sfc-gh-aling, line 72: nit: remove?
```suggestion
        # import pdb; pdb.set_trace()
```

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -0,0 +1,174 @@
...
+                    raise ValueError(
+                        f"Unexpected type {expr.datatype} for sproc argument of type {expected_type}"
+                    )
+                parsed_args.append(
+                    calculate_expression(
+                        expr,
+                        ColumnEmulator(data=[None]),
+                        session._analyzer,
+                        {},
+                        keep_literal=True,
```
sfc-gh-aling, line 66: I have two dumb questions here:

1. what is `Column` type args for sproc and how this is used by users?
2. why we set `ColumnEmulator(data=[None]),` here, this gives us an empty column with just the name, is this intended? this is kinda relavant to the first question on how `Column` is used in sproc.

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -0,0 +1,174 @@
...
+        external_access_integrations: Optional[List[str]] = None,
+        secrets: Optional[Dict[str, str]] = None,
+        *,
+        statement_params: Optional[Dict[str, str]] = None,
+        source_code_display: bool = True,
+        skip_upload_on_content_match: bool = False,
+    ) -> StoredProcedure:
+        raise NotImplementedError(
+            "[Local Testing] Registering stored procedure from file is not currently supported."
+        )
```
sfc-gh-aling, line 107: let's use the telemetry not supported api.
also is this planned be supported by @sfc-gh-stan soon?

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -0,0 +1,174 @@
...
+
+        if packages or imports:
+            raise NotImplementedError(
+                "[Local Testing] Uploading imports and packages not currently supported."
+            )
+
+        check_python_runtime_version(self._session._runtime_version_from_requirement)
+
+        if udf_name in self._registry:
+            raise SnowparkSQLException(
```
sfc-gh-aling, line 161: with parameter `replace`, user can overwrite the existing one

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -0,0 +1,174 @@
...
+            anonymous,
+        )
+
+        if is_pandas_udf:
+            raise TypeError("pandas stored procedure is not supported")
+
+        if packages or imports:
+            raise NotImplementedError(
+                "[Local Testing] Uploading imports and packages not currently supported."
+            )
```
sfc-gh-aling, line 156: let's use oob telemetry/mock connection log not supported api

src/snowflake/snowpark/mock/_stored_procedure.py
sfc-gh-jrose: 1. Some of our tests pass in `lit` values rather than raw values. Those are the only Column objects I've observed being passed to sproc.
2. The way calculate_expression works for lit values requires there to be some input data even if it is all null. I think I can refactor this so that it is more clear.

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -0,0 +1,176 @@
...
+        *args: Any,
+        session: Optional["snowflake.snowpark.session.Session"] = None,
+        statement_params: Optional[Dict[str, str]] = None,
+    ) -> Any:
+        args, session = self._validate_call(args, session)
+
+        # Unpack columns if passed
+        parsed_args = []
+        for arg, expected_type in zip(args, self._input_types):
+            if isinstance(arg, Column):
```
sfc-gh-stan, line 55: Question: why do we need this case? Do we support column as a stored procedure argument?

src/snowflake/snowpark/mock/_stored_procedure.py
sfc-gh-stan: Yea I'll have a separate PR that adds the support for imports.

src/snowflake/snowpark/mock/_stored_procedure.py
sfc-gh-stan: These are good questions actually. In fact I feel like we shouldn't support column inputs to stored procedures, although that might be a limitation from `DataFrame.select`...

sfc-gh-stan: +1 for refactor to be more clear

tests/integ/test_stored_procedure.py
```diff
@@ -161,54 +163,100 @@ def return1(session_):
...
     assert add_sp(4, 6) == 10
     assert int2str_sp(123) == "123"
-    assert pow_sp(2, 10) == 1024
     assert return1_sp(session=session) == "1"
     assert plus1_sp(1, session=session) == 2
     assert add_sp(4, 6, session=session) == 10
     assert int2str_sp(123, session=session) == "123"
-    assert pow_sp(2, 10, session=session) == 1024
 
+    if local_testing_mode is False:
```
sfc-gh-stan, line 211:
```suggestion
    if not local_testing_mode:
```

sfc-gh-stan:
Can we enable more tests in `tests/integ/test_stored_procedure.py` like `test_type_hints`?

sfc-gh-jrose:
I've enabled all the tests that look relevant to local testing. There's several more tests related to imports and loading from file that can't be enabled yet. Turning on those tests revealed a couple of gaps. 

1. session.call didn't work previously, but should now. 
2. Calling a sproc with a nested column expression that returns a single value appears to be supported in live so I've reverted the refactor I made to add support for calling `calculate_expression` back. I've added a comment so that it's a little more understandable.

In order to implement tests 1:1 I also implemented functions.sqrt and functions.pow. They are trivial so shouldn't expand the scope of this PR by much.

CHANGELOG.md
```diff
@@ -25,9 +25,12 @@
     - current_database
     - current_session
     - date_trunc
-    - udf
     - object_construct
     - object_construct_keep_null
+    - pow
+    - udf
+    - sproc
```
sfc-gh-aling, line 32: nit: I think udf and sproc are worth being mentioned separately in the changelog

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -0,0 +1,238 @@
...
+                    session._analyzer,
+                    {},
+                )
+
+                # If the length of the resolved expression is not a single value we cannot pass it as a literal.
+                if len(resolved_expr) != 1:
+                    raise ValueError(
+                        "[Local Testing] Unexpected argument type {expr.__class__.__name__} for call to sproc"
+                    )
+                parsed_args.append(resolved_expr[0])
```
sfc-gh-aling, line 96: can you give me an example of stored procedure with one `arg` being Column? I'm trying to understand how this arg is used -- is it used to getting column of data within this process or we just need to parse the expression into a literal.

if this is just analyzing an expression to infer a literal value, then the MockAnalyzer.analyze can better help here: https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/mock/_analyzer.py#L169-L176.

sfc-gh-jrose: This test has an example of where this is used:
https://github.com/snowflakedb/snowpark-python/pull/1338/files#diff-4963488bfdaad538e6af7222d4db977758e66d05b4c11b600583449152416f50L221

sfc-gh-aling: I see, if evaluating the literal value is the only case, then let's try using `session._analyzer.analyze`.

calculate_expression is for calculating table/column results, while analyze is just for expressions parsing

sfc-gh-jrose: Running analyze on this statement produces query text which is not what the function expects as input. For example, the sqrt done in the test when put through the analyzer creates the string `sqrt(36)` which is not what the function expects as input.

sfc-gh-stan (approved):

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -54,20 +67,51 @@ def __call__(
         for arg, expected_type in zip(args, self._input_types):
             if isinstance(arg, Column):
                 expr = arg._expression
-                if not sproc_types_are_compatible(expr.datatype, expected_type):
+
+                # If expression does not define its datatype we cannot verify it's compatibale.
+                # This is potentially unsafe.
+                if expr.datatype and not sproc_types_are_compatible(
```
sfc-gh-stan, line 73: Question: in what case is expr.datatype empty here?

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -31,11 +42,13 @@
 
 
 def sproc_types_are_compatible(x, y):
-    if isinstance(x, type(y)):
-        return True
-    elif isinstance(x, _IntegralType) and isinstance(y, _IntegralType):
-        return True
-    elif isinstance(x, _FractionalType) and isinstance(y, _FractionalType):
+    if (
```
sfc-gh-stan, line 45: When we work on coercion, this function should be replaced by like `DataType.can_coerce_to(DataType)`. @sfc-gh-aling

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # Semi-structured types are serialized in json
+        if isinstance(
+            self._return_type,
+            (
+                ArrayType,
+                MapType,
+                StructType,
+            ),
+        ) and not isinstance(result, DataFrame):
```
sfc-gh-stan, line 110: We have no support for table sprocs yet, why do we need `not isinstance(result, DataFrame)`?

sfc-gh-stan:
LGTM overall, had a couple of questions and comments.

src/snowflake/snowpark/mock/_stored_procedure.py
sfc-gh-jrose: The datatype of the sqrt function in this example is None:
https://github.com/snowflakedb/snowpark-python/pull/1338/files#diff-4963488bfdaad538e6af7222d4db977758e66d05b4c11b600583449152416f50L221

src/snowflake/snowpark/mock/_stored_procedure.py
sfc-gh-jrose: I've seen examples in the tests where a sproc returns a Dataframe, but it's signature says it returns a Structtype. If it's truly a struct type it needs to be serialized, but if it's just a dataframe it does not.

src/snowflake/snowpark/mock/_stored_procedure.py
sfc-gh-aling: I see, that makes sense. thanks for checking it!

sfc-gh-aling (approved):

sfc-gh-aling:
I noticed the oob telemetry test is flaky these days thus blocking the merge gate.
I created a PR to fix the flakiness: https://github.com/snowflakedb/snowpark-python/pull/1382

src/snowflake/snowpark/mock/_stored_procedure.py
sfc-gh-stan: I see. Could you please point me to the test?

sfc-gh-jrose: See [here](https://github.com/snowflakedb/snowpark-python/blob/e7421b44623e4388fe4f31596ca65bd45abf8109/tests/integ/test_stored_procedure.py).

PR #1341, opened 2024-03-28
Allow UDF/Sproc Decorators to execute in a Local Sandbox

sfc-gh-bgoel:

### Background
When a user wants to create a UDF/UDTF/UDAF (UDF for short) or an Sproc, they are able to use the corresponding decorator and provide certain parameters that help in the creation of the said object in their Snowflake account. As it is, the current code takes in those parameters, performs some validation and transformation on them, pulls in some relevant session information, and then creates a SQL query that is sent to be run in the Snowflake account.

### Ask

### Design
2. Further, Snowflake Native App (or a user that just wants to run code in a local sandbox instead of connecting to Snowflake) logic needs to execute in a sandbox because the UDFs/SPROCs should not be immediately created in Snowflake. Snowflake Native App wants to be able to use the resolved properties to create our own SQL to be run as part of our application code, and this workflow should not happen in the account that the user is developing for. Hence the need to have a sandbox instead of the actual account. We do not want any unintended side-effects happening in the user's account and therefore want to wall-off any interaction with Snowflake.  


src/snowflake/snowpark/context.py
```diff
@@ -4,10 +4,22 @@
...
 
 _use_scoped_temp_objects = True
 
+# This is an internal-only function, used to interrupt registration of UDxF or SPROC objects with Snowflake.
+_internal_only_interrupt_registration: Optional[Callable] = None
+
+# This is an internal-only global flag, used to determine whether to execute code in a client's local sandbox or connect to a Snowflake account.
+_execute_in_local_sandbox: bool = False
+
+
```
sfc-gh-bgoel, line 19: The two variables above will be set from an environment that invokes the python interpreter to read the user's files. For a lot of developers, this environment will be the one that Snowflake CLI is running in. In that case, we just want to fetch (validated and resolved) properties of the UDxF/SPROC decorator, and stop short of actually registering anything with Snowflake. Since this is an execution environment dependent behavior, the environment is responsible for setting the context here.

src/snowflake/snowpark/context.py
```diff
@@ -4,10 +4,22 @@
...
+def get_execute_in_local_sandbox() -> bool:
+    return _execute_in_local_sandbox
```
sfc-gh-bdufour, line 21: Why not just read the internal flag directly? The getter makes it look like this is a public API, but it's not

src/snowflake/snowpark/udf.py
```diff
@@ -850,7 +860,7 @@ def _do_register_udf(
         )
 
         if not custom_python_runtime_version_allowed:
-            check_python_runtime_version(
+            check_python_runtime_version(  # no-op for sandbox
```
sfc-gh-bdufour, line 863: comment is intentional?

src/snowflake/snowpark/context.py
```diff
@@ -4,11 +4,70 @@
...
+            )  # Since sandboxing and connecting to snowflake to continue registration is not possible
+    else:
+        if _interrupt_registration:
+            return ObjectRegistrationDecision.DO_NOT_REGISTER_WITH_SNOWFLAKE
+        else:  # This is the default flow of object registration
+            return ObjectRegistrationDecision.REGISTER_WITH_SNOWFLAKE
+
+
+# This is an internal-only function to decide whether to proceed with registring an object with Snowflake, and if not, then the action that should be taken in place of registration.
+def _get_decision_to_register_with_snowflake(registration_info: Any) -> None:
```
sfc-gh-bdufour, line 58: IMHO this logic has become much more complicated than necessary. To focus on the essentials, you're introducing two concepts here:
1. A callback to be invoked, if defined, whether a uXdf/sproc gets registered
2. The notion of a sandboxed environment so that we know whether connecting to Snowflake is even allowed.

These two concepts are related but orthogonal. I can define a callback to interrupt registration in native apps, or I can define a callback to inspect what's being registered for debugging purposes.

Sandboxing is used to short-circuit logic that requires a Snowflake connection. I can want to do this for reasons unrelated to our use case, of course (e.g. as part of a sandboxed build, or in a testing context).

I don't see a reason why these two concepts need to be intertwined as much as they are here. The simplest option I see that fulfills all use cases I can throw at it: sandboxing is a boolean, and is inspected before a connection to snowflake is made. Callbacks take in the registration info, and return whether or not to proceed with registration. This is flexible because it can be done on a per-callable basis.

src/snowflake/snowpark/functions.py
```diff
@@ -7047,6 +7048,10 @@ def udf(
     external_access_integrations: Optional[List[str]] = None,
     secrets: Optional[Dict[str, str]] = None,
     immutable: bool = False,
+    native_app_params: Optional[
+        Dict[str, Any]
+    ] = None,  # This could look like native_app_params =
+    # {"schema": "some_schema", "application_roles": ["app_public", "app_admin"], "packages": ["a", "b"], "imports": ["c", "d"]}
```
sfc-gh-bdufour, line 7054: as discussed, I don't see that packages and imports would need to be added here. Schema and app roles are fair game though, depending on Snowpark's own plans for generation.

src/snowflake/snowpark/functions.py
```diff
@@ -7198,7 +7203,13 @@ def udf(
         [Row(MINUS_ONE(10)=9)]
 
     """
-    session = session or snowflake.snowpark.session._get_active_session()
+    # If the local sandbox flag is true, then we should not use any session that the user may already have or pass to the decorator.
+    if snowflake.snowpark.context._is_execution_environment_sandboxed:
+        session = Session.builder.config("local_testing", True).create()
+    else:
```
sfc-gh-bdufour, line 7209: IMHO that's the wrong place to do this, and could cause unexpected behaviour. Shouldn't we instead ensure that the active session is still a singleton, but doesn't have a valid connection when sandboxing is enabled? That's a lot more future-proof as well.

src/snowflake/snowpark/session.py
```diff
@@ -440,7 +443,13 @@ def __init__(
         self._session_stage = random_name_for_temp_object(TempObjectType.STAGE)
         self._stage_created = False
 
-        if isinstance(conn, MockServerConnection):
+        # Two possibilities here:
+        # 1. create another MockUDFRegistrationSandbox class
+        # 2. weave the sandbox code from within the original UDFRegistration class
+        if (
+            isinstance(conn, MockServerConnection)
+            and not _is_execution_environment_sandboxed
```
sfc-gh-bdufour, line 451: Seems incorrect. If the sandboxing is enabled as well as local testing is enabled on the current session, then we won't create a mock registration instance.

src/snowflake/snowpark/session.py
```diff
@@ -2638,6 +2663,10 @@ def get_fully_qualified_name_if_possible(self, name: str) -> str:
         """
         Returns the fully qualified object name if current database/schema exists, otherwise returns the object name
         """
+
+        if _is_execution_environment_sandboxed:
+            return ""  # This will fail validation for object name regex in Snowflake, as it should, because local sandboz objects should not be anonymous.
```
sfc-gh-bdufour, line 2668: typo: sandbox

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -482,6 +485,10 @@ def run_query(
             object_name = match.group(2)
             setattr(self, f"_active_{object_type}", object_name)
             return {"data": [("Statement executed successfully.",)], "sfqid": None}
+        elif (
+            not self._local_testing
+        ):  # i.e. we are in a sandbox environment = not a true testing environment
+            return {"data": [("Statement executed successfully.",)], "sfqid": None}
```
sfc-gh-bdufour, line 491: Why do we need this?

src/snowflake/snowpark/session.py
```diff
@@ -805,6 +821,10 @@ def _resolve_imports(
     ) -> List[str]:
         """Resolve the imports and upload local files (if any) to the stage."""
         resolved_stage_files = []
+
+        if _is_execution_environment_sandboxed:
```
sfc-gh-bdufour, line 825: nit: I'd add a comment here to explain what we're doing with the short-circuit

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1051,6 +1136,38 @@ def create_python_udf_or_sp(
         else ""
     )
 
+    callableProperties = CallableProperties(
```
sfc-gh-bdufour, line 1139: nit: only create this if the callback is set

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1128,6 +1247,33 @@ def generate_anonymous_python_sp_sql(
...
+        all_packages=all_packages,
+        external_access_integrations=external_access_integrations,
+        secrets=secrets,
+        handler=handler,
+        inline_python_code=inline_python_code,
+        native_app_params=native_app_params,
+        unresolved_imports=unresolved_imports,
+    )
+
+    continue_registration = (
```
sfc-gh-bdufour, line 1268: nit: make this look like the one above?

sfc-gh-bgoel: Done, will be in the next commit.

src/snowflake/snowpark/session.py
sfc-gh-bgoel: Done, in the next commit.

sfc-gh-jrose (commented):
Overall I think the change makes sense. Separating out the native apps params seems like the right thing to do.

My main concern for this change is that it adds additional complexity to already lengthy and complex functions like `resolve_imports_and_packages`. That function should probably be broken up a bit into some helper functions. I usually follow the rule of thumb that if the whole function can't fit on screen at once it's probably too long.

Adding complexity to support a different execution mode to the Session class will make code maintenance harder. Instead I'd rather see a subclass of Session with appropriate parts changed. The session builder can then choose which type of session is more applicable. 

Finally, global variables for control flow could probably be better handled with an environment variable for `_is_execution_environment_sandboxed` and a session builder parameter for `_should_continue_registration`. I think we should avoid mutable global variables when possible.

Those are my thoughts, but I'd be interested in seeing what other people say before you start refactoring.

sfc-gh-bdufour:
Thanks for your feedback! Answers inline

> Overall I think the change makes sense. Separating out the native apps params seems like the right thing to do.
> 
> My main concern for this change is that it adds additional complexity to already lengthy and complex functions like `resolve_imports_and_packages`. That function should probably be broken up a bit into some helper functions. I usually follow the rule of thumb that if the whole function can't fit on screen at once it's probably too long.

Fair point, however we were asked very specifically to keep the number of lines changed to a minimum and avoid refactoring for this change at least. Yijun thought that introducing refactorings would increase the risk associated with the PR, and that it was best to avoid this.

> 
> Adding complexity to support a different execution mode to the Session class will make code maintenance harder. Instead I'd rather see a subclass of Session with appropriate parts changed. The session builder can then choose which type of session is more applicable.

Keep in mind what we're doing is truly global and not scoped to a session, since we're loading arbitrary user code in a sandbox. We don't know how the user code will create sessions (e.g. through a builder or not).

> 
> Finally, global variables for control flow could probably be better handled with an environment variable for `_is_execution_environment_sandboxed` and a session builder parameter for `_should_continue_registration`. I think we should avoid mutable global variables when possible.

That was deliberate, see above regarding session state. It's actually one of those rare cases where mutable global state is desirable, IHMO. We need to guarantee that no matter how the user's code is written, we're get the information we need and prevent accidental SF queries.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -869,7 +947,7 @@ def resolve_imports_and_packages(
                     "or a tuple of the file path (str) and the import path (str)."
                 )
             udf_level_imports[resolved_import_tuple[0]] = resolved_import_tuple[1:]
-        all_urls = session._resolve_imports(
+        all_urls = session._resolve_imports(  # _resolve_imports will return an empty list in case of sandbox
```
sfc-gh-yixie, line 950: For offline discussion, is it possible to skip using session if the native apps sandbox doesn't have a session?

sfc-gh-bgoel:
Closing in favor of #1381

PR #1369, opened 2024-04-04
SNOW-1235478: local test to_time support more data conversion

sfc-gh-aling:

   Fixes SNOW-1235478

support to_time according to: https://docs.snowflake.com/en/sql-reference/functions/to_time#arguments

- For string_expr, the result of converting the string to a time.
- For timestamp_expr, the time portion of the input value.
- For 'integer' (a string containing an integer), the integer is treated as a number of seconds, milliseconds, microseconds, or nanoseconds after the start of the Unix epoch
- For variant_expr:
  - If the variant contains a string in TIME format (‘HH:MI:SS’), a string conversion is performed.
  - If the variant contains a string in INTEGER format, a string conversion is performed and the value is treated as the number of seconds since midnight (modulus 86400 if necessary).
  - If the variant contains JSON null value, the output is NULL.


src/snowflake/snowpark/mock/_functions.py
```diff
@@ -481,45 +487,63 @@ def mock_to_time(
...
+                    For variant_expr:
+                        If the variant contains a string in TIME format (‘HH:MI:SS’), a string conversion is performed.
+                        If the variant contains a string in INTEGER format, a string conversion is performed and the value is treated as the number of seconds since midnight (modulus 86400 if necessary).
+                        If the variant contains JSON null value, the output is NULL.
+                    """
+                    if data.isnumeric():
+                        # handle integer format strings,
+                        # strings in TIME format and JSON null value are handled already by non-VariantType logic
+                        res.append(convert_int_string_to_time(data))
+                        continue
```
sfc-gh-jrose, line 513: This case looks the same as the StringType case to me. Are there other subcases that will be here?

sfc-gh-aling: no other type so far.. read the spec it's just StringType cases, other cases will trigger sql error saying can not convert the value

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -495,45 +501,63 @@ def mock_to_time(
...
-                            process_string_time_with_fractional_seconds(
-                                data, fractional_seconds
-                            ),
-                            time_fmt,
-                        )
-                        + datetime.timedelta(hours=hour_delta)
-                    ).time()
+            if auto_detect:
+                # handling non-standard string expr cases
+                if isinstance(column.sf_type.datatype, StringType) and data.isnumeric():
```
sfc-gh-stan: NIT: Use `isdigit()` instead for integers?

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -469,15 +469,21 @@ def mock_to_time(
...
+    [x] For 'integer' (a string containing an integer), the integer is treated as a number of seconds, milliseconds, microseconds, or nanoseconds after the start of the Unix epoch. See the Usage Notes below.
 
-    [ ] For this timestamp, the function gets the number of seconds after the start of the Unix epoch. The function performs a modulo operation to get the remainder from dividing this number by the number of seconds in a day (86400): number_of_seconds % 86400
+        [x] For this timestamp, the function gets the number of seconds after the start of the Unix epoch. The function performs a modulo operation to get the remainder from dividing this number by the number of seconds in a day (86400): number_of_seconds % 86400
 
     """
+
+    def convert_int_string_to_time(d: str):
+        return datetime.datetime.utcfromtimestamp(
+            process_numeric_time(d) % 86400
```
sfc-gh-stan, line 486: Seems like `process_numeric_time` converts integer values that could be milliseconds or nanoseconds to seconds? Does it handle microseconds? 

NIT: Can we rename the function to something like `convert_integer_value_to_seconds`?

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -495,45 +501,63 @@ def mock_to_time(
...
+                    res.append(data.time())
+                    continue
+                # if the data doesn't fall into the above two categories, it goes to the string_expr handling
+                elif isinstance(column.sf_type.datatype, VariantType):
+                    """
+                    https://docs.snowflake.com/en/sql-reference/functions/to_time#arguments
+                    For variant_expr:
+                        If the variant contains a string in TIME format (‘HH:MI:SS’), a string conversion is performed.
+                        If the variant contains a string in INTEGER format, a string conversion is performed and the value is treated as the number of seconds since midnight (modulus 86400 if necessary).
+                        If the variant contains JSON null value, the output is NULL.
```
sfc-gh-stan: This case also needs to handle when `data` is of `datetime` type, i.e. the first case in:
```
select to_time(current_time()::variant);
select to_time(current_time()::varchar);
```

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -495,45 +501,63 @@ def mock_to_time(
...
+                    """
+                    if data.isnumeric():
```
sfc-gh-stan, line 544: We cannot assume `data` is a `str`.

tests/integ/scala/test_function_suite.py
```diff
@@ -1287,6 +1288,70 @@ def test_to_timestamp(session):
...
+            Row(time(21, 20, 19)),
+            Row(
+                None,
+            ),
+        ],
+    )
+
+    # variant data
+
+    # invalid input for string expr with format
```
sfc-gh-stan, line 1340: `expected_exception = ValueError if local_testing_mode else SnowparkSQLException`?

tests/utils.py
```diff
@@ -803,6 +803,54 @@ def datetime_primitives1(cls, session: "Session") -> DataFrame:
         )
         return session.create_dataframe(data, schema)
 
+    @classmethod
+    def time_primitives1(cls, session: "Session") -> DataFrame:
```
sfc-gh-stan, line 807: Where are these being used?

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -495,45 +501,63 @@ def mock_to_time(
             if data is None:
```
sfc-gh-stan: NIT: can we rewrite this to use `ColumnEmulator.apply` instead of a for loop?

tests/integ/scala/test_function_suite.py
sfc-gh-aling: todo: turn on this test

sfc-gh-jrose (commented):
Looks good other than the test that needs to be enabled. I wonder if there is any shared functionality with `to_timestamp*` that could be abstracted out into some utility functions.

tests/utils.py
sfc-gh-aling: it's now used in the first case of the test_to_time

tests/integ/scala/test_function_suite.py
sfc-gh-aling: updated

src/snowflake/snowpark/mock/_functions.py
sfc-gh-aling: added isinstance check first

src/snowflake/snowpark/mock/_functions.py
sfc-gh-aling: updated to handle all unit as per the spec, also added tests

src/snowflake/snowpark/mock/_functions.py
sfc-gh-aling: yeah digit seems better here, updated

src/snowflake/snowpark/mock/_functions.py
sfc-gh-aling: good point, code updated to handle variant type with value being time, also added a test for it

src/snowflake/snowpark/mock/_functions.py
sfc-gh-aling: it probably takes some extra time to rewrite so I would keep it for now

sfc-gh-aling:
@sfc-gh-jrose I haven't worked on to_timestamp yet, but I feel some can be consolidated.
I will do in another PR

sfc-gh-jrose (approved):

sfc-gh-stan (approved):

PR #1381, opened 2024-04-11
Allow Session-less Decorators for Extension Functions

sfc-gh-bgoel:

### Background
When a user wants to create an extension function (UDF/UDTF/UDAF/Sproc), they are able to use the corresponding decorator and provide certain parameters that help in the creation of the said object in their Snowflake account. As it is, the current code takes in those parameters, performs some validation and transformation on them, pulls in some relevant session information, and then creates a SQL query that is sent to be run in the Snowflake account.

### Ask

### Design
2. Further, Snowflake Native App (or a user that just wants to run code in a local sandbox instead of connecting to Snowflake) logic needs to execute in a sandbox because we do not wish to actually create the extension functions in Snowflake, and avoid any side effects through other session calls. Snowflake Native App wants to be able to use the resolved properties to create our own SQL to be run as part of our application code, and this workflow should not happen in the account that the user is developing for. 

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -1323,26 +1323,27 @@ $$
     )
 
     # As an FYI, _should_continue_registration is a function, and is defined outside the Snowpark context.
-    extension_function_properties = ExtensionFunctionProperties(
-        anonymous=True,
-        object_type=TempObjectType.PROCEDURE,
-        object_name=object_name,
-        input_args=input_args,
-        input_sql_types=input_sql_types,
-        return_sql=return_sql,
-        runtime_version=runtime_version,
-        all_imports=all_imports,
-        all_packages=all_packages,
-        external_access_integrations=external_access_integrations,
-        secrets=secrets,
-        handler=handler,
-        inline_python_code=inline_python_code,
-        native_app_params=native_app_params,
-        import_paths=import_paths,
-        func=func,
-    )
-    # The result of the function call below does not matter because we are not using session object here
-    _should_continue_registration(extension_function_properties)
+    if _should_continue_registration is not None:
+        extension_function_properties = ExtensionFunctionProperties(
+            anonymous=True,
+            object_type=TempObjectType.PROCEDURE,
+            object_name=object_name,
+            input_args=input_args,
+            input_sql_types=input_sql_types,
+            return_sql=return_sql,
+            runtime_version=runtime_version,
+            all_imports=all_imports,
+            all_packages=all_packages,
+            external_access_integrations=external_access_integrations,
+            secrets=secrets,
+            handler=handler,
+            inline_python_code=inline_python_code,
+            native_app_params=native_app_params,
+            import_paths=import_paths,
+            func=func,
+        )
+        # The result of the function call below does not matter because we are not using session object here
+        _should_continue_registration(extension_function_properties)
 
     sql = f"""
 WITH {object_name} AS PROCEDURE ({sql_func_args})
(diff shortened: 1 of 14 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -726,6 +726,7 @@ def clear_imports(self) -> None:
             self.udf._clear_session_imports()
         self._import_paths.clear()
 
+    @staticmethod
     def _resolve_import_path(
         self,
```
sfc-gh-yixie, line 731: self is not needed any more if it's a static method.

src/snowflake/snowpark/functions.py
```diff
@@ -7198,10 +7208,18 @@ def udf(
         [Row(MINUS_ONE(10)=9)]
 
     """
-    session = session or snowflake.snowpark.session._get_active_session()
+
+    # Precedence to checking sandbox to avoid any side effects
+    if _is_execution_environment_sandboxed:
+        session = None
+        udf_registration = UDFRegistration(session=session).register
```
sfc-gh-yixie, line 7215: `udf_registration = UDFRegistration(session=session).register()`?

sfc-gh-bgoel: I am assigning the method `register` to the variable `udf_registration`, so `udf_registration` is of type Callable. This is because depending on the session object being None, the registration class is initialized differently. And hence where we are calling the `register` method also changes. I want to hide that complexity behind a simple variable.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1128,6 +1322,28 @@ def generate_anonymous_python_sp_sql(
...
+        external_access_integrations=external_access_integrations,
+        secrets=secrets,
+        handler=handler,
+        inline_python_code=inline_python_code,
+        native_app_params=native_app_params,
+        import_paths=import_paths,
+        func=func,
+    )
+    # The result of the function call below does not matter because we are not using session object here
+    _should_continue_registration(extension_function_properties)
```
sfc-gh-jrose, line 1345: I believe this needs a guard in case `_should_continue_registration` is None.

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -534,6 +535,11 @@ def register(
                 The secrets can be accessed from handler code. The secrets specified as values must
                 also be specified in the external access integration and the keys are strings used to
                 retrieve the secrets using secret API.
+            native_app_params: This is a special parameter, that is relevant when using this function to create UDFs
+                as a Snowflake Native App developer. It is a dictionary of parameters that are relevant in Snowflake Native Apps,
+                such as schema and application roles.
+                A typical dictionary could look like: {"schema": "some_schema", "application_roles": ["app_public", "app_admin"]}
+                This parameter is ignored if you are not developing a Snowflake Native App.
```
sfc-gh-yixie, line 542: I suggest you briefly tell users what native apps will do with these parameters.

src/snowflake/snowpark/functions.py
```diff
@@ -7056,6 +7060,9 @@ def udf(
     like to register it to multiple databases, use ``session.udf.register`` instead. See examples
     in :class:`~snowflake.snowpark.udf.UDFRegistration`.
 
+    This can also be used to create an extension function in a `Snowflake Native App Framework
+    setup script <https://docs.snowflake.com/en/developer-guide/native-apps/creating-setup-script>`_.
```
sfc-gh-yixie, line 7064: Will you add sample code that uses the decorators in the linked page?

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -464,6 +464,7 @@ def register(
         *,
         statement_params: Optional[Dict[str, str]] = None,
         source_code_display: bool = True,
+        native_app_params: Optional[Dict[str, Any]] = None,
```
sfc-gh-yixie, line 467: This will become a public API but a user can't use it for now. So it's better we hide it. You can consider using the `**kwargs` and remove the docstring of native_app_params

sfc-gh-bgoel:
Closed in favor of https://github.com/snowflakedb/snowpark-python/pull/1397

PR #1382, opened 2024-04-12
SNOW-1324573: local testing fix flaky oob telemetry test

sfc-gh-aling:

   Fixes SNOW-1324573

the local testing oob telemetry is flaky because the service takes longer to process or throttled.
do the following to improve:
- increate the default wait time, this barely add overhead to customer as we are doing batch sending underneath
- add retry in our test
- remove duplicate and unreliable exact record tuple assertion, checking caplog text should be enough

### DIFF
--- a/src/snowflake/snowpark/mock/_telemetry.py
+++ b/src/snowflake/snowpark/mock/_telemetry.py
@@ -10,8 +10,6 @@ from datetime import datetime
 from enum import Enum
 from typing import Optional
 
-import requests
-
 from snowflake.connector.compat import OK
 from snowflake.connector.secret_detector import SecretDetector
 from snowflake.connector.telemetry_oob import TelemetryService
@@ -20,6 +18,13 @@ from snowflake.snowpark._internal.utils import (
     get_python_version,
     get_version,
 )
+
+REQUESTS_AVAILABLE = True
+try:
+    # by default in stored procedure requests is not imported
+    import requests
+except ImportError:
+    REQUESTS_AVAILABLE = False
 
 # 3 seconds setting in the connector oob could be too short that the oob service is unable to handle the request,
 # 5 seconds is more tolerant
@@ -86,6 +91,11 @@ class LocalTestOOBTelemetryService(TelemetryService):
         self._deployment_url = self.PROD
 
     def _upload_payload(self, payload) -> None:
+        if not REQUESTS_AVAILABLE:
+            logger.debug(
+                "request module is not available",
+            )
+            return
         success = True
         response = None
         try:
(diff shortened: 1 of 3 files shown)


sfc-gh-jrose (approved):

sfc-gh-aalam:
Is the test supposed to work without requests? `Exception response: None, exception: No module named 'requests'`

src/snowflake/snowpark/mock/_telemetry.py
```diff
@@ -83,9 +87,8 @@ def _upload_payload(self, payload) -> None:
         success = True
         response = None
         try:
-            # import here is because stored proc doesn't have vendored request module
-            # have it at the top will cause import error in stored procedure running
-            from snowflake.connector.vendored import requests
+            # import here is because stored proc doesn't have requests as requirement
+            import requests
```
sfc-gh-stan, line 91: Question: Why do we use non vendored requests here?

tests/mock_unit/test_oob_telemetry.py
```diff
@@ -139,28 +146,38 @@ def test_unit_oob_log_not_implemented_error(caplog, local_testing_telemetry_setu
 
     # test sending successfully
     caplog.clear()
-    with caplog.at_level(logging.DEBUG, logger="snowflake.snowpark.mock._telemetry"):
-        oob_service.log_not_supported_error(
-            external_feature_name=unraise_feature_name,
-        )
+
+    max_retry = 10
```
sfc-gh-stan, line 150: What about replacing this with a pytest plugin: https://github.com/box/flaky?

src/snowflake/snowpark/mock/_telemetry.py
sfc-gh-aling: the vendored requests triggers unnecessary ocsp validation

tests/mock_unit/test_oob_telemetry.py
sfc-gh-aalam: this is very cool ^

sfc-gh-aalam (approved):

tests/mock_unit/test_oob_telemetry.py
sfc-gh-aling: this is good idea, but I did this retry manually for 2 reasons here:

1. I'm not sure introducing a new dependency (though just for test) would require any stored proc test change or other configure change. luckily nothing breaks and works, but we need extra time to test.
2. though it's test dependency, does it still need 3rd party library review process?

so I think we can log a JIRA for improvement at a later time.

sfc-gh-aling:
@sfc-gh-aalam I'm not sure why doing a local import requests would trigger error ModuleNotFound.. but moving the import to global fixes the flakiness.

in stored proc requests is not available by default, I re-implement this so that stored proc shall also work.

PR #1383, opened 2024-04-12
SNOW-1305528 : Adding lineage.trace API

sfc-gh-rsureshbabu:

The PR is introducing a new API (lineage.trace). Detailed description of the API is in the design doc : https://docs.google.com/document/d/1vDmOWWk5BRGUb74h0UgGScfTcJ4GsyGTT5b32llvJYo/edit#heading=h.hs3dx8soa4jr


src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,377 @@
...
+        """
+        if _ObjectField.PARENT_NAME in graph_entity:
+            name = f"{graph_entity[_ObjectField.DB]}.{graph_entity[_ObjectField.SCHEMA]}.{graph_entity[_ObjectField.PARENT_NAME]}"
+            version = graph_entity[_ObjectField.NAME]
+        else:
+            name = f"{graph_entity[_ObjectField.DB]}.{graph_entity[_ObjectField.SCHEMA]}.{graph_entity[_ObjectField.NAME]}"
+            version = None
+
+        domain = (
+            graph_entity.get(_ObjectField.REFINED_DOMAIN)
```
sfc-gh-rsureshbabu, line 251: Waiting on server side change to fix this bug. Will fix this before merge.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,440 @@
...
+    CREATED_ON = "createdOn"
+    VERSION = "version"
+
+    # A list of fileds queried on each object in the lineage.
+    GRAPH_ENTITY_PROPERTIES = [
+        DOMAIN,
+        REFINED_DOMAIN,
+        USER_DOMAIN,
+        NAME,
+        PARENT_NAME,
```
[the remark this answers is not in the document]
sfc-gh-yli: oh nevermind, this comment was on older commits. Seems like you have made new changes

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+            return []
+
+        lineage_edges = self._get_lineage(
+            object_name, object_domain, [direction], object_version, current_depth
+        )
+        if (
+            current_depth == total_depth
+            or not lineage_edges
+            or self._is_masked_object(lineage_edges[0][0])
+            or self._is_masked_object(lineage_edges[0][1])
```
sfc-gh-yli, line 250: how about deleted objects? if a deleted object is returned, and we need request the next level of it. It seems you are using name to construct the request, but using name for a deleted obejcts might not be able to resolve. For example, for UI, we are using id field for requesting for recursive search

sfc-gh-rsureshbabu: Should we even go to the next level from the deleted node ?

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
     - save_as_table
 - Added support for snow:// URLs to `snowflake.snowpark.Session.file.get` and `snowflake.snowpark.Session.file.get_stream`
 - UDAF client support is ready for public preview. Please stay tuned for the Snowflake announcement of UDAF public preview.
+- Added new API ( snowflake.snowpark.Session.lineage.trace ) to explore data lineage of snowfake objects.
```
sfc-gh-yixie, line 46: 1. You may need to merge the main branch and move this to unreleased. 
2. I suggest `Added snowflake.snowpark.Session.lineage.trace ...`

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+        )
+        return self._session.create_dataframe(transformed_results, schema=schema)
+
+    def trace(
+        self,
+        object_name: str,
+        object_domain: str,
+        *,
+        object_version: Optional[str] = None,
+        direction: LineageDirection = LineageDirection.BOTH,
```
sfc-gh-yixie, line 379: I suggest you accept str value as well. Snowpark's other APIs support str values. It's easier to use in a jupyter notebook.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+class _EdgeType(Enum):
+    """
+    Types of edges for lineage tracing.
+    """
+
+    DATA_LINEAGE = "DATA_LINEAGE"
+    OBJECT_DEPENDENCY = "OBJECT_DEPENDENCY"
+
+    @classmethod
+    def list(cls):
```
sfc-gh-yixie, line 47: Nit: `values` might be more informative.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+    To access an object of this class, use :attr:`Session.lineage`.
+    """
+
+    def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
+        self._session = session
+        self.user_to_system_domain_map = {
+            _UserDomain.FEATURE_VIEW: _SnowflakeDomain.TABLE,
+            _UserDomain.MODEL: _SnowflakeDomain.MODULE,
+        }
+        self.versioned_object_domains = {
```
sfc-gh-yixie, line 128: Are you intending to expose `user_to_system_domain_map` and `versioned_object_domains` to users?

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: Discussed offline. For PrPr, we will not explore from deleted nodes. we can extend this based on user feedback.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+            or self._is_masked_object(lineage_edges[0][1])
+        ):
+            return lineage_edges
+
+        next_object = (
+            lineage_edges[0][1]
+            if direction == LineageDirection.DOWNSTREAM
+            else lineage_edges[0][0]
+        )
+        deeper_lineage_edges = self._recursive_trace(
```
sfc-gh-yixie, line 259: Python's max recursion depth is 1000. Will this possibly be over 1000? If yes you may need to convert it to an iterative way.

sfc-gh-rsureshbabu: We have a maximum limit on recursion. The current limit is 10.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+
+class Lineage:
+    """
+    Provides methods for exploring lineage of Snowflake objects.
+    To access an object of this class, use :attr:`Session.lineage`.
+    """
+
+    def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
+        self._session = session
+        self.user_to_system_domain_map = {
```
sfc-gh-tbao, line 124: shall we make this a module level constant?

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+        object_version: Optional[str] = None,
+    ) -> str:
+        """
+        Constructs a GraphQL query for lineage tracing based on the specified parameters.
+        """
+        properties_string = ", ".join(_ObjectField.GRAPH_ENTITY_PROPERTIES)
+        edge_types_formatted = ", ".join(_EdgeType.list())
+
+        parts = []
+        edge_template = "{direction}: {edge_key}(edgeType:[{edge_types}],direction:{dir}){{{source_key} {{{properties}}}, {target_key} {{{properties}}}}}"
```
sfc-gh-tbao, line 148: put it as a class constant?

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+        object_version: Optional[str] = None,
+        current_depth=1,
+    ) -> List[Tuple[VariantType, VariantType, StringType, int]]:
+        """
+        Constructs and executes a query to trace the lineage of a given entity at a single depth level.
+        """
+        query_string = self._build_graphql_query(
+            object_name, object_domain, directions, object_version
+        )
+        response = self._session.sql(query_string)
```
sfc-gh-tbao, line 204: how about put .collect() here?
also do we need to handle exception here?

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+    def _get_lineage(
+        self,
+        object_name: str,
+        object_domain: str,
+        directions: list,
+        object_version: Optional[str] = None,
+        current_depth=1,
+    ) -> List[Tuple[VariantType, VariantType, StringType, int]]:
+        """
+        Constructs and executes a query to trace the lineage of a given entity at a single depth level.
```
sfc-gh-tbao, line 199: maybe mention the output schema in comment: source, target, dir, depth. So it's earlier to interpret later

src/snowflake/snowpark/lineage.py
sfc-gh-tbao: the depth is currently guarded to be <=5

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+        self,
+        object_name: str,
+        object_domain: str,
+        *,
+        object_version: Optional[str] = None,
+        direction: LineageDirection = LineageDirection.BOTH,
+        depth: int = 2,
+    ) -> "snowflake.snowpark.dataframe.DataFrame":
+        """
+        Traces the lineage of a object within Snowflake and returns it as a DataFrame.
```
sfc-gh-tbao, line 383: nit: an object

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+
+        Args:
+            object_name (str): The name of the Snowflake object to start trace, formatted as "database.schema.object".
+            object_domain (str): The domain of the Snowflake object to start trace. e.g., "table", "view".
+            object_version (Optional[str]): Version of the Snowflake object to start trace, defaults to None.
```
sfc-gh-tbao, line 388: maybe mention only applicable to versioned object like Model/Dataset

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+            direction (LineageDirection): The direction to trace (UPSTREAM, DOWNSTREAM, BOTH), defaults to BOTH.
+            depth (int): The depth of the trace, defaults to 2.
```
sfc-gh-tbao, line 390: Mention max depth limit?

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+        """
+        if not object_name:
+            raise ValueError("Object name must be provided.")
+        if not object_domain:
+            raise ValueError("Object type must be provided.")
+
+        if depth < 1 or depth > 5:
+            raise ValueError("Depth must be between 1 and 5.")
+
+        if direction == LineageDirection.BOTH and depth == 1:
```
sfc-gh-tbao, line 433: is there a reason we need to handle this specially?

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+        return entity[_ObjectField.STATUS] in {"MASKED", "DELETED"}
+
+    def _get_name_and_version(self, graph_entity):
+        """
+        Extracts and returns the name and version from the given graph entity.
+        """
+        if graph_entity[_ObjectField.USER_DOMAIN] in self.versioned_object_domains:
+            if graph_entity[_ObjectField.USER_DOMAIN] == _UserDomain.FEATURE_VIEW:
+                if "$" in graph_entity[_ObjectField.NAME]:
+                    parts = graph_entity[_ObjectField.NAME].split("$")
```
sfc-gh-tbao, line 284: the safest way to check it's a feature view is to query the relevant tags on the table. Is there a viable way to do that here?

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: Nope. made it _*

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: I changed it to 10 in the latest commit.

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: you mean at snowpark level ?

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: We dont use this in any other function right now. Its only useful in query generation function

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: sorry, .collect() of which object ? 

I dont see any exception that we could catch and handle here.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+            edges="".join(parts),
+        )
+
+        return f"select SYSTEM$DGQL('{query}')"
+
+    def _get_lineage(
+        self,
+        object_name: str,
+        object_domain: str,
+        directions: list,
```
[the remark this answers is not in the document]
sfc-gh-rsureshbabu: Why do we need list? we have BOTH option

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: Its a small optimization. We can get the information in 1 call instead f 2 calls.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,454 @@
...
+    DATA = "data"
+    NODE = "V"
+    EDGE = "E"
+    SOURCE = "S"
+    TARGET = "T"
+    OUT = "OUT"
+    IN = "IN"
+
+
+class _UserDomain:
```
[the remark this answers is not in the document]
sfc-gh-rsureshbabu: Its for internal use + we are calling in userDomain in DGQL as well. It would be easier to be consistent

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: We are doing all those things in the server side and the populate userDomain to FeatureView

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,468 @@
...
+        current_depth=1,
+    ) -> List[Tuple[VariantType, VariantType, StringType, int]]:
+        """
+        Constructs and executes a query to trace the lineage of a given entity at a single depth level.
+        """
+        query_string = self._build_graphql_query(
+            object_name, object_domain, directions, object_version
+        )
+        response = self._session.sql(query_string)
+        json_response = json.loads(response.collect()[0][0])
```
sfc-gh-yli, line 217: what does [0][0] mean?

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,468 @@
...
+            or not lineage_edges
+            or self._is_terminal_entity(lineage_edges[0][0])
+            or self._is_terminal_entity(lineage_edges[0][1])
+        ):
+            return lineage_edges
+
+        next_object = (
+            lineage_edges[0][1]
+            if direction == LineageDirection.DOWNSTREAM
+            else lineage_edges[0][0]
```
sfc-gh-yli, line 269: why are we just getting one object for next_object, wouldn't there be a list of next_objects? for cases like A -> [B, C, D], we need to call recursively on B, C, D separately. Also if I am understanding correctly, we are doing DFS, we can hit the same object multiple times in the graph, are we doing optimization to avoid calling DGQL on the object multiple times? for example A -> [B, C -> [B, D], D], and C's downstream B/D might already be visited

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: The test covers this case. Right now, we throw this to the user as is.

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: [row][column]

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: Output will have list of edges (source, target). I am processing target in each edge. 
 I am tracking the visited nodes to avoid loops.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,468 @@
...
+                | {"createdOn": "1712881232230", "domain": "TABLE",      | {"createdOn": "1712881232230", "domain": "VIEW",   | "Downstream"| 1       |
+                |  "name": "YOUR_DATABASE.YOUR_SCHEMA.T1", "status":     |  "name": "YOUR_DATABASE.YOUR_SCHEMA.V1", "status": |             |         |
+                |  "ACTIVE"}                                             |  "ACTIVE"}                                         |             |         |
+                | {"createdOn": "1712881232230", "domain": "VIEW",       | {"createdOn": "1712881232230", "domain": "VIEW",   | "Downstream"| 2       |
+                |  "name": "YOUR_DATABASE.YOUR_SCHEMA.V1", "status":     |  "name": "YOUR_DATABASE.YOUR_SCHEMA.V2", "status": |             |         |
+                |  "ACTIVE"}                                             |  "ACTIVE"}                                         |             |         |
+                ---------------------------------------------------------------------------------------------------------------------------------------
+                <BLANKLINE>
+        """
+        if depth < 1 or depth > 10:
```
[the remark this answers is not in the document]
sfc-gh-rsureshbabu: I dont see the reason why 5 or 10. I am fine with 5 as well. Switched it.

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,468 @@
...
+                >>> df = session.lineage.trace(
+                ...     f"{db}.{schema}.T1",
+                ...     "table",
+                ...     direction=LineageDirection.DOWNSTREAM
+                ... )
+                >>> df.show()
+                ---------------------------------------------------------------------------------------------------------------------------------------
+                | "SOURCE_OBJECT"                                        | "TARGET_OBJECT"                                    | "LINEAGE"   | "DEPTH" |
+                ---------------------------------------------------------------------------------------------------------------------------------------
+                | {"createdOn": "1712881232230", "domain": "TABLE",      | {"createdOn": "1712881232230", "domain": "VIEW",   | "Downstream"| 1       |
```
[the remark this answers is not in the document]
sfc-gh-rsureshbabu: I am not opposed to displaying the time in a specific format; however, providing users with the timestamp as a long integer offers them the flexibility to convert it into their preferred time format. Even if we provide the time in a formatted string, users would still need to convert it into a time type for it to be usable.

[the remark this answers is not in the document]
sfc-gh-yli: This doesn't need to be blocking, but I think it is best to run this by PMs and see if we have any existing product surface that exposes the raw long values like this. To my knowledge, I am not aware of any, and we always expose `timestamp_ltz`

sfc-gh-tbao (approved):
Thanks for working on this @sfc-gh-rsureshbabu , looks great.

src/snowflake/snowpark/lineage.py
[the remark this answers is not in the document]
sfc-gh-tbao: I mean the type hint could be more specific.

src/snowflake/snowpark/lineage.py
sfc-gh-tbao: `response = self._session.sql(query_string).collect()`

initially the code read like we are obtaining the response, but it actually lazily actualized.

src/snowflake/snowpark/lineage.py
sfc-gh-tbao: just in this python file

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,490 @@
...
+        Recursively traces lineage by making successive DGQL queries based on response nodes.
+        """
+        if current_depth > total_depth:
+            return []
+
+        current_node = (object_name, object_domain, object_version)
+        if current_node in visited:
+            return []
+
+        visited.add(current_node)
```
sfc-gh-yli, line 264: are you using one `visited` set for the entire graph search? I thought you need to use different `visited` for each path. What if you have 2 paths like
1. A->B->C->D
2. A-> D
3. B->D
If you visit D first in path 1, does that mean you cannot find 2nd and 3rd path now since D is visited?

sfc-gh-rsureshbabu: I see what you mean. I think BFS would be better to handle these cases. I have switched it to BFS. I have a visited set per direction.

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: I swicthed the logic to BFS. PTAL

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,509 @@
...
+                StructField("source_object", VariantType()),
+                StructField("target_object", VariantType()),
+                StructField("lineage", StringType()),
+                StructField("depth", IntegerType()),
+            ]
+        )
+        return self._session.create_dataframe(transformed_results, schema=schema)
+
+    @private_preview(version="1.16.0")
+    def trace(
```
sfc-gh-yixie, line 426: It's not a must for now but I prefer using a LogicalPlan like other places. Could you help create a followup JIRA?

src/snowflake/snowpark/lineage.py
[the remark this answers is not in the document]
sfc-gh-yixie: Using a string with int value for a datetime isn't Pythonic. I prefer datetime. Could you create a followup JIRA to address it?

sfc-gh-yixie (approved):

sfc-gh-yli (approved):
discussed offline, the new BFS algorithm is better at handling loop/cycles. LGTM

src/snowflake/snowpark/lineage.py
sfc-gh-rsureshbabu: Created a JIRA https://snowflakecomputing.atlassian.net/browse/SNOW-1355760. I have put this in Q2 plan.

src/snowflake/snowpark/lineage.py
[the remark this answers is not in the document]
sfc-gh-rsureshbabu: We discussed this offline. We dont know the right format for datetime. Users either need to convert from Long to datetime or string time to datetime. Created a JIRA to switch it based on PrPr feedback. https://snowflakecomputing.atlassian.net/browse/SNOW-1355769

src/snowflake/snowpark/lineage.py
```diff
@@ -0,0 +1,507 @@
...
+                ...     f"CREATE OR REPLACE VIEW {db}.{schema}.V2 AS SELECT * FROM {db}.{schema}.V1"
+                ... ).collect()
+                >>> df = session.lineage.trace(
+                ...     f"{db}.{schema}.T1",
+                ...     "table",
+                ...     direction=LineageDirection.DOWNSTREAM
+                ... )
+                >>> df.show()
+                ---------------------------------------------------------------------------------------------------------------------------------------
+                | "SOURCE_OBJECT"                                        | "TARGET_OBJECT"                                    | "LINEAGE"   | "DEPTH" |
```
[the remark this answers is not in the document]
sfc-gh-yli: also should the third column be `direction` instead of `lineage`?

PR #1397, opened 2024-04-17
Allow Session-less Decorators for Extension Functions

sfc-gh-bgoel:

### Background
When a user wants to create an extension function (UDF/UDTF/UDAF/Sproc), they are able to use the corresponding decorator and provide certain parameters that help in the creation of the said object in their Snowflake account. As it is, the current code takes in those parameters, performs some validation and transformation on them, pulls in some relevant session information, and then creates a SQL query that is sent to be run in the Snowflake account.

### Ask

### Design
2. Further, Snowflake Native App (or a user that just wants to run code in a local sandbox instead of connecting to Snowflake) logic needs to execute in a sandbox because we do not wish to actually create the extension functions in Snowflake, and avoid any side effects through other session calls. Snowflake Native App wants to be able to use the resolved properties to create our own SQL to be run as part of our application code, and this workflow should not happen in the account that the user is developing for. 


src/snowflake/snowpark/functions.py
```diff
@@ -7047,6 +7050,7 @@ def udf(
     external_access_integrations: Optional[List[str]] = None,
     secrets: Optional[Dict[str, str]] = None,
     immutable: bool = False,
+    **kwargs,
```
sfc-gh-yixie, line 7053: How about raising a ValueError in the code to make sure only `native_app_params` is in the kwargs? This kwargs is tentative so I hope to make sure users don't pass in unexpected arguments. So In the future there won't be a breaking change when we make native_app_params explicit.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -548,7 +610,8 @@ def process_registration_inputs(
     else:
         object_name = random_name_for_temp_object(object_type)
         if not anonymous:
-            object_name = session.get_fully_qualified_name_if_possible(object_name)
+            if session is not None:
+                object_name = session.get_fully_qualified_name_if_possible(object_name)
```
sfc-gh-stan, line 614: NIT: merge the nested if into `and`

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -806,8 +869,66 @@ def {_DEFAULT_HANDLER_NAME}({wrapper_params}):
...
+def add_snowpark_package_to_sproc_packages(
+    session: Optional["snowflake.snowpark.Session"],
+    packages: Optional[List[Union[str, ModuleType]]],
+) -> List[Union[str, ModuleType]]:
+    major, minor, patch = VERSION
+    package_name = "snowflake-snowpark-python"
+    # Use == to ensure that the remote version matches the local version
+    this_package = f"{package_name}=={major}.{minor}.{patch}"
+
+    # When resolve_imports_and_packages is called below it will use the provided packages or
```
sfc-gh-stan, line 881: Let's put these comments in a docstring.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -806,8 +869,66 @@ def {_DEFAULT_HANDLER_NAME}({wrapper_params}):
...
+        if not any(package_name in p for p in packages):
+            packages.append(this_package)
+    return packages
+
+
+def resolve_packages_in_sandbox(
+    packages: Optional[List[Union[str, ModuleType]]],
+) -> List[str]:
+    """
+    Special function invoked only when executing Snowpark code in a sandbox environment.
```
sfc-gh-stan, line 900: The word sandbox is an existing term in the backend, let's change this to something like native_apps_cli?

src/snowflake/snowpark/functions.py
sfc-gh-stan: I prefer just adding `native_app_params` as an optional parameter, why are we using kwargs?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -825,142 +946,180 @@ def resolve_imports_and_packages(
...
-        else session.get_session_stage(statement_params=statement_params)
-    )
-
-    upload_and_import_stage = (
-        import_only_stage
-        if is_permanent
-        else session.get_session_stage(statement_params=statement_params)
-    )
+) -> Tuple[
+    Optional[str],
```
sfc-gh-stan, line 950: `Optional` should be used very cautiously here.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -972,23 +1131,26 @@ def resolve_imports_and_packages(
         inline_code,
         all_imports,
         all_packages,
+        udf_level_imports,
```
sfc-gh-stan, line 1134: Shouldn't `udf_level_imports` be included in `all_imports`?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1052,6 +1219,38 @@ def create_python_udf_or_sp(
         else ""
     )
 
+    # As an FYI, _should_continue_registration is a function, and is defined outside the Snowpark context.
+    if _should_continue_registration is None:
```
sfc-gh-stan, line 1223: If `should_continue_registraion` is only used in Native Apps Sandbox environment, what's the use case when this returns True?

src/snowflake/snowpark/context.py
```diff
@@ -4,10 +4,22 @@
...
 
+# This is an internal-only global flag, used to determine whether to execute code in a client's local sandbox or connect to a Snowflake account.
+# If this is True, then the session instance is forcibly set to None to avoid any interaction with a Snowflake account.
+_is_execution_environment_sandboxed: bool = False
+
+# This callback, assigned by the caller environment outside Snowpark, can be used to share information about the extension function to be registered.
+# It should also return a decision on whether to proceed with registring the extension function with the Snowflake account.
+# If _should_continue_registration is None, i.e. a caller environment never assigned it an alternate callable, then we want to continue registration as part of the regular Snowpark workflow.
+# If _should_continue_registration is not None, i.e. a caller environment has assigned it an alternate callable, then the callback is responsible for determining the rest of the Snowpark workflow.
+_should_continue_registration: Optional[Callable] = None
```
sfc-gh-stan, line 21: We can further indicate the return type of the `Callable`.

sfc-gh-stan (changes requested):

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bgoel: Discussed on zoom, these comments help understand flow of the code, but docstrings help with overall description and behavior of the function. Will leave as inline comments.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bgoel: Discussed on zoom, `native_apps_cli` is too specific, this is any generic sandbox that is created for a client running the library to prevent talking to snowflake. Changed it accordingly in context.py and elsewhere.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bgoel: Discussed the override mechanism here, though that can be a fast-follow to this PR.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bgoel: Added inline explanation for this.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bgoel: `_should_continue_registration` is not specific to a sandbox execution. This could be installed by any caller environment that just wants to inspect objects, e.g. testing, before they are created. So in that way, you could still want to continue registration.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -333,6 +394,15 @@ def get_error_message_abbr(object_type: TempObjectType) -> str:
     raise ValueError(f"Expect FUNCTION of PROCEDURE, but get {object_type}")
 
 
+def check_decorator_args(**kwargs):
+    for key, _ in kwargs.items():
+        if key not in REGISTER_KWARGS_ALLOWLIST:
+            raise ValueError(
+                "Invalid key-value arguments passed to the decorator."
+                "Only 'native_app_params' is accepted as an additional argument to this decorator."
```
sfc-gh-bdufour, line 402: outdated?

src/snowflake/snowpark/functions.py
sfc-gh-bdufour: The idea was that we'll migrate to an optional parameter once the second half of the implementation is ready. This temporary hiding of the argument is simply to limit accidental discovery in between releases.

sfc-gh-bdufour (approved):

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-stan: Please add a TODO comment with the JIRA number/Github issue link here.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -825,142 +953,181 @@ def resolve_imports_and_packages(
...
 
     # resolve imports
+    # This udf_level_imports will be used as a return value, as it contains information even if the session is None
+    udf_level_imports = {}
     if imports:
-        udf_level_imports = {}
         for udf_import in imports:
             if isinstance(udf_import, str):
-                resolved_import_tuple = session._resolve_import_path(udf_import)
+                resolved_import_tuple = (
```
sfc-gh-stan, line 1005: I think we can just use `snowflake.snowpark.Session._resolve_import_path` since that's a static method, we can remove the `self` pointer from the method definition.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -825,142 +953,181 @@ def resolve_imports_and_packages(
...
         all_urls = []
 
     dest_prefix = get_udf_upload_prefix(udf_name)
 
     # Upload closure to stage if it is beyond inline closure size limit
-    if isinstance(func, Callable):
-        custom_python_runtime_version_allowed = (
-            False  # As cloudpickle is being used, we cannot allow a custom runtime
-        )
+    handler = inline_code = upload_file_stage_location = None
```
sfc-gh-stan, line 1050: Let's move the definition of `custom_python_runtime_version_allowed` here too.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -972,23 +1139,26 @@ def resolve_imports_and_packages(
...
     )
 
 
 def create_python_udf_or_sp(
     session: "snowflake.snowpark.Session",
+    func: Union[Callable, Tuple[str, str]],
     return_type: DataType,
     input_args: List[UDFColumn],
-    handler: str,
+    handler: Optional[str],
```
sfc-gh-stan, line 1153: The use of `Optional` here should also be improved by the same follow up as we discussed above.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -999,12 +1169,17 @@ def create_python_udf_or_sp(
...
     statement_params: Optional[Dict[str, str]] = None,
     comment: Optional[str] = None,
+    native_app_params: Optional[Dict[str, Any]] = None,
 ) -> None:
-    runtime_version = (
-        f"{sys.version_info[0]}.{sys.version_info[1]}"
-        if not session._runtime_version_from_requirement
-        else session._runtime_version_from_requirement
-    )
+    if session is not None:
```
sfc-gh-stan, line 1174:
```suggestion
    runtime_version = f"{sys.version_info[0]}.{sys.version_info[1]}"
    if session is not None and session._runtime_version_from_requirement:
        runtime_version = session._runtime_version_from_requirement
```

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -972,23 +1139,26 @@ def resolve_imports_and_packages(
...
         all_imports,
         all_packages,
+        udf_level_imports,  # We return this extra dict since all_imports is an empty string when session is None.
         upload_file_stage_location,
         custom_python_runtime_version_allowed,
     )
 
 
 def create_python_udf_or_sp(
     session: "snowflake.snowpark.Session",
```
sfc-gh-stan, line 1149: This should also be Optional now?

src/snowflake/snowpark/udaf.py
```diff
@@ -294,7 +294,7 @@ class UDAFRegistration:
         - :meth:`~snowflake.snowpark.DataFrame.join_table_function`
     """
 
-    def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
+    def __init__(self, session: Optional["snowflake.snowpark.session.Session"]) -> None:
```
sfc-gh-stan, line 297: Why is this Optional?

sfc-gh-bgoel: All the decorators could either have the session = None or otherwise. Depending on that, 

```
    if session is None:
        udf_registration_method = UDFRegistration(session=session).register
    else:
        udf_registration_method = session.udf.register
```
When session is None, we cannot call session.udf.register, and hence need to use the UDFRegistration class directly. Initializing it requires a session, and since we are initializing it with a None session, the signature has been modified to accept the None session.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -869,98 +1016,114 @@ def resolve_imports_and_packages(
...
 
     dest_prefix = get_udf_upload_prefix(udf_name)
 
     # Upload closure to stage if it is beyond inline closure size limit
-    if isinstance(func, Callable):
-        custom_python_runtime_version_allowed = (
-            False  # As cloudpickle is being used, we cannot allow a custom runtime
-        )
+    handler = inline_code = upload_file_stage_location = None
+    custom_python_runtime_version_allowed = False
```
sfc-gh-stan, line 1046:
```suggestion
    custom_python_runtime_version_allowed = not  isinstance(func, Callable)
```

sfc-gh-stan (approved):

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-yixie: Yeah this is to make it more generic.

src/snowflake/snowpark/functions.py
```diff
@@ -7208,7 +7208,7 @@ def udf(
     """
 
     # Initial check to make sure no unexpected args are passed in
-    check_decorator_args(kwargs)
+    check_decorator_args(**kwargs)
```
sfc-gh-yixie, line 7211: Nit: unboxing kwargs in `check_decorator_args `  isn't necessary. I'm OK if you don't change it becauswe we'll change this part pretty soon.

sfc-gh-yixie (approved):

PR #1419, opened 2024-04-24
SNOW-1324779: Update warnings for Snowpark pandas

sfc-gh-joshi:

   Fixes #1324779

This PR modifies Snowpark pandas to error and prompt the user to install Modin if the incorrect pandas or Modin version is encountered. It also adds a warning informing users that the pandas feature is in PuPr.

These error messages were tested manually by manually installing the appropriate dependencies. If so desired, they can also be set up in CI as well, but might be more complicated to run since they require environmental changes.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/__init__.py
+++ b/src/snowflake/snowpark/modin/plugin/__init__.py
@@ -2,20 +2,31 @@
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
 
+import sys
+import warnings
+
 from packaging import version
 
-# We need this import here to prevent circular dependency issues, since snowflake.snowpark.modin.pandas
-# currently imports some internal utilities from snowflake.snowpark.modin.plugin. Test cases will
-# import snowflake.snowpark.modin.plugin before snowflake.snowpark.modin.pandas, so in order to prevent
-# circular dependencies from manifesting, apparently snowflake.snowpark.modin.pandas needs to
-# be imported first.
-from snowflake.snowpark.modin import pandas  # noqa: F401
-from snowflake.snowpark.modin.config import DocModule
-from snowflake.snowpark.modin.plugin import docstrings
+if sys.version_info.major == 3 and sys.version_info.minor == 8:
+    raise RuntimeError(
+        "Snowpark pandas does not support Python 3.8. Please update to Python 3.9 or later."
+    )  # pragma: no cover
 
-DocModule.put(docstrings.__name__)
 
 install_msg = "Run `pip install snowflake-snowpark-python[modin]` to resolve."
+
+# pandas import needs to come before Python version + modin checks,
+# since modin may raise its own warnings/errors on the wrong pandas version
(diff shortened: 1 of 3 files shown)


sfc-gh-nkrishna (approved):
LGTM thanks @sfc-gh-joshi!

src/snowflake/snowpark/modin/plugin/__init__.py
```diff
@@ -26,5 +37,26 @@
...
+        f"The Modin version installed ({modin.__version__}) does not match the supported Modin version in"
+        + f" Snowpark pandas ({supported_modin_version}). "
+        + install_msg
     )  # pragma: no cover
+
+
+warnings.warn(
+    "Snowpark pandas is currently in Public Preview."
+    + " See https://docs.snowflake.com/LIMITEDACCESS/snowpark-pandas for details.",
+    stacklevel=1,
```
sfc-gh-mvashishtha, line 49:
```suggestion
```

nit: `stacklevel=1` is the default

sfc-gh-mvashishtha (approved):
LGTM

src/snowflake/snowpark/modin/plugin/__init__.py
sfc-gh-joshi: Not specifying stacklevel is a flake8 warning:
```
src/snowflake/snowpark/modin/plugin/__init__.py:46:1: B028 No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
```

sfc-gh-mvashishtha: I don't understand how `stacklevel` works; do you? stacklevel 1, 2, 0, 3, all do the same thing here:

```python
import warnings

warnings.filterwarnings('error', message='.*evie.*')

def do_deprecated():
    deprecated_api()
def deprecated_api():
    warnings.warn('preview', DeprecationWarning, stacklevel=0)
do_deprecated()
```


Screenshot 2024-04-24 at 4 53 54 PM

sfc-gh-joshi: I tried your example, and removing the `filterwarnigns` call makes the `stacklevel` argument change which method is shown.

sfc-gh-mvashishtha: True! I added the `filterwarnings` because without it, python wasn't showing a method at all! Why doesn't this show the code that raised the warning?

```python
def f():
    import snowflake.snowpark.modin.pandas as pd
f()
```

even if i put the warning inside a function:

```python

def f():
    warnings.warn(
        "Snowpark pandas is currently in Public Preview."
        + " See https://docs.snowflake.com/LIMITEDACCESS/snowpark-pandas for details.",
        stacklevel=1,
    )

f()
```

maybe it's because __init__.py isn't run via a function.

PR #1434, opened 2024-04-25
SNOW-1344848: Suppress dynamic pivot warnings

sfc-gh-mvashishtha:

   Fixes SNOW-1344848

    Stop warning the user about our internal usage of private preview
    dynamic pivot features. The user should have already been warned that
    Snowpark pandas is in public or private preview. They very likely don't
    know or care that we are using Snowpark DataFrame pivot() internally,
    let alone that we are using private preview features of Snowpark
    Python.
    
    I'm not adding a changelog note because we haven't released a version of
        Snowpark Python with this bug yet.

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -182,8 +182,12 @@ SCOPED_TEMPORARY_STRING = "SCOPED TEMPORARY"
 
 SUPPORTED_TABLE_TYPES = ["temp", "temporary", "transient"]
 
-PIVOT_VALUES_NONE_OR_DATAFRAME_WARNING = "Parameter values is Optional or DataFrame is in private preview since v1.15.0.  Do not use it in production."
-PIVOT_DEFAULT_ON_NULL_WARNING = "Parameter default_on_null is not None is in private preview since v1.15.0.  Do not use it in production."
+PIVOT_VALUES_NONE_OR_DATAFRAME_WARNING = (
+    "Calling pivot() with the `value` parameter set to None or to a Snowpark "
+    + "DataFrame is in private preview since v1.15.0. Do not use this feature "
+    + "in production."
+)
+PIVOT_DEFAULT_ON_NULL_WARNING = "Calling pivot() with the default_on_null is not None is in private preview since v1.15.0.  Do not use it in production."
 
 
 class TempObjectType(Enum):
@@ -879,6 +883,9 @@ def escape_quotes(unescaped: str) -> str:
     return unescaped.replace(DOUBLE_QUOTE, DOUBLE_QUOTE + DOUBLE_QUOTE)
 
 
+should_warn_dynamic_pivot_is_in_private_preview = True
+
+
 def prepare_pivot_arguments(
     df: "snowflake.snowpark.DataFrame",
     df_name: str,
@@ -899,17 +906,17 @@ def prepare_pivot_arguments(
     """
     from snowflake.snowpark.dataframe import DataFrame
 
-    if values is None or isinstance(values, DataFrame):
-        warning(
-            df_name,
-            PIVOT_VALUES_NONE_OR_DATAFRAME_WARNING,
-        )
-
-    if default_on_null is not None:
-        warning(
-            df_name,
-            PIVOT_DEFAULT_ON_NULL_WARNING,
-        )
+    if should_warn_dynamic_pivot_is_in_private_preview:
+        if values is None or isinstance(values, DataFrame):
+            warning(
+                df_name,
+                PIVOT_VALUES_NONE_OR_DATAFRAME_WARNING,
+            )
+        if default_on_null is not None:
+            warning(
+                df_name,
+                PIVOT_DEFAULT_ON_NULL_WARNING,
+            )
 
     if values is not None and not values:
         raise ValueError("values cannot be empty")
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -182,6 +182,9 @@ SCOPED_TEMPORARY_STRING = "SCOPED TEMPORARY"
 
 SUPPORTED_TABLE_TYPES = ["temp", "temporary", "transient"]
 
+PIVOT_VALUES_NONE_OR_DATAFRAME_WARNING = "Parameter values is Optional or DataFrame is in private preview since v1.15.0.  Do not use it in production."
+PIVOT_DEFAULT_ON_NULL_WARNING = "Parameter default_on_null is not None is in private preview since v1.15.0.  Do not use it in production."
+
 
 class TempObjectType(Enum):
     TABLE = "TABLE"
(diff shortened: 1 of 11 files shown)


sfc-gh-mvashishtha:
I removed @sfc-gh-lmukhopadhyay because she's OOO

sfc-gh-yuwang (commented):
LGTM, but I want to wait for pandas team reviewing this since we don't have tons of knowledge of it.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -182,6 +182,9 @@
 
 SUPPORTED_TABLE_TYPES = ["temp", "temporary", "transient"]
 
+PIVOT_VALUES_NONE_OR_DATAFRAME_WARNING = "Parameter values is Optional or DataFrame is in private preview since v1.15.0.  Do not use it in production."
```
sfc-gh-jkew, line 185: can you add the word, "pivot" here somewhere? The warning has no context on the operation that we are trying to perform.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -818,16 +822,41 @@ def pivot(
         See detailed docstring in Snowpark DataFrame's pivot.
         """
         snowpark_dataframe = self.to_projected_snowpark_dataframe()
+
+        # Don't warn the user about our internal usage of private preview
```
sfc-gh-jkew, line 826: This seems wildly complex just to suppress a warning. injecting filters into logging just for this seems like overkill. I suggest one of the following:

1. Emit the warning only when a configuration variable is true, and we set it to false when snowpandas is initialized
2. Don't bother emitting the warning at all.

sfc-gh-jkew (changes requested):
Too complicated of a solution for a minor problem.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-rdurrani: +1 - since we're now integrated with the snowpark code, maybe we should add an extra kwarg that suppresses the warning to `prepare_pivot_arguments` and just set that in our code?

sfc-gh-rdurrani (approved):
LGTM, but I agree with @sfc-gh-jkew's comment - think there may be an easier solution here.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-mvashishtha: I know this solution seems complex, but I thought about a few alternatives and thought that this solution made the most sense.

Here are the goals I think we should have:

1) We should warn Snowpark pandas users who are using Snowpark dynamic pivot, but not using it through Snowpark pandas, that the feature is experimental.
2) We should warn Snowpark users who are NOT snowpark pandas users that the Snowpark dynamic pivot feature is experimental.
3) We should **not** give the warning when users use the pivot via Snowpark pandas. Snowpark pandas users already get a warning that the entire Snowpark pandas feature is in private or public preview. Usually, they won't be aware that the pandas method they've called is using Snowpark pivot under the hood. 

Some solutions I have considered:

1) What I have currently. Fulfills all 3 goals.
2) As @sfc-gh-rdurrani  proposes, add a parameter to `prepare_pivot_arguments` that tells whether to give the warnings. The problem with this is that it only really makes sense to call `prepare_pivot_arguments` in the snowpark python `pivot()` implementation, and so then we have to pass this new parameter in to the user-exposed `pivot()` method or add a separate, private `pivot()` method with the extra parameter. I don't like either of those approaches-- we want to stick to using public interfaces of Snowpark dataframes, and I don't want to pollute the public interface of `pivot()` with this flag.
2) @sfc-gh-jkew 's solution 1). The problem is that if a Snowpark pandas user happens to also use dynamic pivot on a Snowpark dataframe, they will not get a warning. So this fulfills goals 2 and 3, but not goal 1.
3) @sfc-gh-jkew 's solution 2). This fulfills goal 3) but not goals 1) or 2).

about the complexity:

I know I implemented this change, but I don't think it's terribly complex. It's 30 lines contained in a method that does little else. There are 2 new classes but each is under 10 lines and I think they're each easy to understand.

If after this explanation this solution still seems unacceptable, I can go check with @sfc-gh-evandenberg to see if we can just get rid of the warnings completely.

sfc-gh-jkew: I don't think the first goal is particularly important in this case. We can just assume that if they are using snowpandas they are already using experimental features for PuPr.

sfc-gh-mvashishtha: i pinged @sfc-gh-evandenberg to see if he's ok with skipping goal 1). waiting on response.

sfc-gh-stan: Another way to keep track of flags in runtime context is Session.RuntimeConfig: https://github.com/snowflakedb/snowpark-python/blob/ed89373bc587e8f7e842a9393b39a0036bbb2ea3/src/snowflake/snowpark/session.py#L295. I also think warning filter is an overkill.

sfc-gh-mvashishtha: Okay, I went with @sfc-gh-jkew 's option 1).

src/snowflake/snowpark/_internal/utils.py
sfc-gh-mvashishtha: Done.

tests/conftest.py
```diff
@@ -91,3 +93,11 @@ def cte_optimization_enabled(pytestconfig):
 
 def pytest_sessionstart(session):
     os.environ["SNOWPARK_LOCAL_TESTING_INTERNAL_TELEMETRY"] = "1"
+
+
+@pytest.fixture(autouse=True)
+def clear_warning_dict():
+    yield
+    # clear the warning dict so that warnings from one test don't affect
```
sfc-gh-stan, line 101: Our tests are run in parallel using pytest-xdist, I don't think this logic ensures a clean warning_dict for each test?

sfc-gh-stan (approved):
This seems to also surpress the warning when the user (1) imports Snowpark pandas but doesn't do anything and (2)uses dynamic pivot directly from Snowpark Python API. But I understand the code will be much more complicated to address this specific case, in the future we should work out a way to generalize this solution for `if Snowpark Pandas uses an experimental/pre-GA public API from Snowpark Python, the warning should be surpressed`.

tests/conftest.py
sfc-gh-mvashishtha: Thanks for bringing this up. pytest-xdist runs tests in separate processes that run in parallel, so there's only one `warning_dict` in each process. This fixture resets the warning dict before every test case that each process runs, so `warning_dict` should be clear before each test case.

sfc-gh-jkew (approved):

sfc-gh-mvashishtha:
> in the future we should work out a way to generalize this solution for if Snowpark Pandas uses an experimental/pre-GA public API from Snowpark Python, the warning should be surpressed

Sure, if this comes up at least twice again, we might want to add something like the log filter I added earlier to generalize the warning suppression here.

sfc-gh-evandenberg:
LGTM

PR #1444, opened 2024-04-26
[SNOW-1338078]: Do not fail in deduplication if ordering columns not renamed

sfc-gh-rdurrani:

   Fixes SNOW-1338078

This PR makes sure that deduplication of columns before a join does not fail if metadata columns are not renamed. It also adds a test for this, which uncovered a bug where the `where` API does not ensure the row position column before doing a positional join, which was also fixed.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -7585,6 +7585,8 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
                     ]
                 }
         else:
+            frame = frame.ensure_row_position_column()
+            cond_frame = cond_frame.ensure_row_position_column()
             joined_frame, result_column_mapper = join_utils.join(
                 frame,
                 cond_frame,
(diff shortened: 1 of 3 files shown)


sfc-gh-nkumar (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7585,6 +7585,8 @@ def where(
                     ]
                 }
         else:
+            frame = frame.ensure_row_position_column()
```
sfc-gh-helmeleegy, line 7588: Will we need to do this before every join? If so, there is at least one more join in this function, where we don't call `ensure_row_position_column()`. Also, how about `align_on_index()`? I haven't thought about it, so just asking here.

sfc-gh-helmeleegy: More generally, do we need to search for this pattern even outside of `where`?

sfc-gh-rdurrani: `align_on_index` shouldn't need `row_position` - I also checked to see if anywhere join was being called with a row position column had this same bug (not calling ensure_row_position beforehand) but it doesn't look like this happens anywhere else!

sfc-gh-helmeleegy: Sounds great!

sfc-gh-helmeleegy (approved):
LGTM. Thanks for the quick fix!

PR #1454, opened 2024-04-29
SNOW-1347394: Remove BaseQueryCompiler + error in QC default to pandas

sfc-gh-joshi:

   Fixes SNOW-1347394

This PR removes our vendored copy of the `BaseQueryCompiler` class, inheriting the class from upstream Modin instead. Similarly, it removes all the operator registration classes defined in `snowflake.snowpark.modin.core.dataframe.algebra.default2pandas`, with one exception. Upstream Modin does not properly render the names of `property` objects (https://github.com/modin-project/modin/issues/7233), so we should override `DataFrameDefault.register` to fix this until this issue is fixed upstream.

This PR incidentally removes `Series.dt.week` + `Series.dt.weekofyear`, which were already removed in pandas 2.0.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -646,6 +639,9 @@ def finalize(self) -> None:
     def free(self) -> None:
         pass
 
+    def execute(self) -> None:
+        pass
+
```
sfc-gh-nkrishna, line 644: Why do we need this?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -664,29 +660,26 @@ def repartition(self, axis: Any = None) -> "SnowflakeQueryCompiler":
...
+        # Previously, Snowpark pandas would register a stored procedure that materializes the frame
+        # and performs the native pandas operation. Because this fallback has extremely poor
+        # performance, we now raise NotImplementedError instead.
+        args_str = ", ".join(map(str, args))
+        if args and kwargs:
+            args_str += ", "
+        if kwargs:
+            args_str += ", ".join(f"{key}={value}" for key, value in kwargs.items())
+        ErrorMessage.not_implemented(
+            f"{object_type}.{fn_name}({args_str}) is not yet available in Snowpark pandas API"
```
sfc-gh-nkrishna, line 680: Just to be consistent with what was mentioned on Slack:
```suggestion
            f"Snowpark pandas doesn't yet support {object_type}.{fn_name}({args_str}"
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: This was added as an abstract method to `BaseQueryCompiler` in upstream Modin last year as a means of forcing computation for lazy backends (https://github.com/modin-project/modin/issues/5221) that all children classes have to implement. Even though our DataFrame is lazy it would take a fair bit of work to actually properly implement this, so I elected to just leave a `pass` for now.

sfc-gh-nkrishna: Gotcha, makes sense.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: Changed. Note that since this method call occurs behind a few layers of indirection (for example, `BinaryDefault.register` -> `DataFrameDefault.register` -> `SnowflakeQueryCompiler.default_to_pandas()`) we lose information about which specific parameter caused the error, so a followup PR should replace calls in the QC to `*Default.register` with a direct `ErrorMessage.not_implemented` to call out the specific method/param as needed.

sfc-gh-nkrishna (approved):
Thanks @sfc-gh-joshi!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Is it worth filing a ticket for that?

sfc-gh-joshi: I think this would be covered by SNOW-1341619, though that ticket itself might need to get split into smaller chunks.

sfc-gh-nkrishna: Ah okay, cool.

sfc-gh-vbudati (approved):
Thanks Jonathan!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -664,29 +660,26 @@ def repartition(self, axis: Any = None) -> "SnowflakeQueryCompiler":
...
+        # Previously, Snowpark pandas would register a stored procedure that materializes the frame
+        # and performs the native pandas operation. Because this fallback has extremely poor
+        # performance, we now raise NotImplementedError instead.
+        args_str = ", ".join(map(str, args))
+        if args and kwargs:
+            args_str += ", "
+        if kwargs:
+            args_str += ", ".join(f"{key}={value}" for key, value in kwargs.items())
+        ErrorMessage.not_implemented(
+            f"Snowpark pandas doesn't yet support {object_type}.{fn_name}({args_str}"
```
sfc-gh-nkumar, line 680: nit: `...doesn't yet support the method {object_type}.{fn_name}`

sfc-gh-nkumar (approved):

sfc-gh-joshi:
Do we not want the message to include the specified parameters? Some query
compiler methods directly call DataFrameDefault.register when certain
parameter combinations are encountered, while the method is generally
supported.

PR #1460, opened 2024-04-30
SNOW-1347393: Remove fallback and raise NotImplementedError for GroupBy APIs

sfc-gh-nkrishna:

   SNOW-1347393

   This PR removes fallback and raises NotImplementedError for GroupBy APIs.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -2471,7 +2471,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         is_supported = check_is_groupby_supported_by_snowflake(by, level, axis)
         if not is_supported:
             ErrorMessage.not_implemented(
-                "GroupBy.ngroups is not implemented if axis == 1, both by and level are configured, or if `by` contains any non-pandas hashable labels."
+                "GroupBy.ngroups is not implemented yet if axis == 1, both by and level are configured, or if `by` contains any non-pandas hashable labels."
             )
 
         query_compiler = get_frame_with_groupby_columns_as_index(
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2471,36 +2468,20 @@ def groupby_ngroups(
...
-            return fallback_ngroups()
 
         query_compiler = get_frame_with_groupby_columns_as_index(
             self, by, level, dropna
         )
 
         if query_compiler is None:
-            return fallback_ngroups()
+            ErrorMessage.not_implemented(
+                "GroupBy.ngroups is not implemented for the given parameter combination"
```
sfc-gh-nkumar, line 2483: If we can also tell users which exact parameter is not supported. That would be more actionable information. Not a pupr blocker though so If that leads to much bigger refactoring we can as follow up after PuPr..

sfc-gh-nkrishna: For that, we should need to refactor `check_is_groupby_supported_by_snowflake` and `check_is_aggregation_supported_in_snowflake` which is used in a lot of places. I think it's better to do that in a follow-up PR if necessary given the CHANGELOG/our documentation documents unsupported params.

sfc-gh-mvashishtha (approved):
LGTM

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-azhan: We could get some info from the code path without changing the code right? Some errors raised below may specific to some usage cases. Can we make "for the given parameter combination" more concrete without changing code here?

sfc-gh-azhan: Oh, I see you already have the group method name included. Then it should be good for now.

sfc-gh-azhan (approved):
LGTM

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: The alternative is something like "`Groupby.{agg_name}` is not implemented for `pd.Grouper`, `axis == 1`, both `by` and `level` are configured, `by` contains any non-pandas hashable labels, or unsupported aggregation parameters."

sfc-gh-nkumar: I like detailed error message.

sfc-gh-nkumar (approved):

PR #1468, opened 2024-04-30
[private branch] Maintain the AST as we build up Snowpark expressions: table(), filter()

sfc-gh-azwiegincew:
A very basic initial attempt at serializing the AST.

I'm trying to maintain a parallel codebase for phases 0 and 1 for now, since it would be a shame to do this work twice. Once we complete and ship phase 0, we'll be able to drastically simplify the phase 1 client.

Unlike what I mentioned before, this implementation doesn't flush dependencies of eagerly evaluated expressions. Instead, any client-side value is appended to the pending batch. This is simpler to implement and will likely work well, although we may need to do some dependency analysis on the server to ensure we don't issue unnecessary queries.


src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
+import base64
+import itertools
+import sys
+
```
sfc-gh-lspiegelberg, line 9: should the `proto.ast_pb2` file get checked in at all? I assume it's generated from `ast.proto`? Would maybe remove it from the commit and add acomment to run first ast-make.sh before using this code.

src/snowflake/snowpark/dataframe.py
```diff
@@ -508,8 +509,10 @@ def __init__(
         session: Optional["snowflake.snowpark.Session"] = None,
         plan: Optional[LogicalPlan] = None,
         is_cached: bool = False,
```
sfc-gh-lspiegelberg, line 511: I'd like to understand a bit better (for me) what ast_smt is supposed to be. Can we either add a comment, or maybe a simple test - just to see how it's supposed to be used?

src/snowflake/snowpark/_internal/ast.py
sfc-gh-azwiegincew: The build process is TBD. I'm thinking we shouldn't even check in the proto file, since its master copy is in the main Snowpark repo.

src/snowflake/snowpark/dataframe.py
```diff
@@ -3285,6 +3300,22 @@ def _lateral(self, table_function: TableFunctionExpression) -> "DataFrame":
...
         query = self._plan.queries[-1].sql.strip().lower()
 
+        # TODO: A little hack to prevent infinite recursion.
+        if not "snowpark_coprocessor" in query:
+            # Add an Assign node that applies SpDataframeShow() to the input, followed by its Eval.
+            repr = self._session._ast_batch.assign()
+            repr.expr.sp_dataframe_show.id.CopyFrom(self._ast_stmt.var_id)
+            self._session._ast_batch.eval(repr)
+            
+            print(f'Original: {self._plan.queries}')
```
sfc-gh-lspiegelberg, line 3310: nit: can we use logging? e.g., logging.debug or logging.trace.

sfc-gh-azwiegincew: This is here temporarily. I wouldn't want to log these things for real.

sfc-gh-lspiegelberg (approved):
left some comments, I think we can move faster by keeping the review process for now lighter. At some point we may need to do some cleanup/quality check work though then.

src/snowflake/snowpark/dataframe.py
```diff
@@ -4093,8 +4124,8 @@ def dtypes(self) -> List[Tuple[str, str]]:
         ]
         return dtypes
 
-    def _with_plan(self, plan) -> "DataFrame":
-        df = DataFrame(self._session, plan)
+    def _with_plan(self, plan, ast_stmt=None) -> "DataFrame":
```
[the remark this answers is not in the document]
sfc-gh-azwiegincew: I'll add a comment. We shouldn't be using the plan – the AST is not the SQL query plan. In Phase 1, any mention of plan will go away.

tests/thin-client/steel-thread.py
```diff
@@ -0,0 +1,7 @@
+from snowflake.snowpark import Session
+from snowflake.snowpark.functions import col, contains
```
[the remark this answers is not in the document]
sfc-gh-azwiegincew: This is just a steel thread driver. It will go away. Let's keep it nimble.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-azwiegincew: We arleady have a task for this. Resolving.

PR #1475, opened 2024-05-01
Bugfix/pandas df reference in mock

DustinMoriarty:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-1358494

2. Fill out the following pre-review checklist:

   - [x ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Previously, the mock plan attempted to replace np.nan with None for certain data types. However, this code had no effect. This caused a warning to be raised by pandas. This non-effective code should be removed if current code works without it. This will silence the warning and reduce confusion.

Issue #1474, opened 2024-05-01
Pandas Replace on Copy of DataFrame has No Effect


DustinMoriarty:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using?

3.11.5

2. What operating system and processor architecture are you using?

macOS-14.4.1-arm64-arm-64bit

3. What are the component versions in the environment (`pip freeze`)?

annotated-types==0.6.0
asn1crypto==1.5.1
black==24.4.0
certifi==2024.2.2
cffi==1.16.0
charset-normalizer==3.3.2
click==8.1.7
cloudpickle==2.2.1
coverage==7.5.0
cryptography==42.0.5
filelock==3.13.4
flake8==7.0.0
idna==3.7
iniconfig==2.0.0
isort==5.13.2
mccabe==0.7.0
mypy-extensions==1.0.0
numpy==1.26.4
packaging==24.0
pandas==2.2.2
pathspec==0.12.1
platformdirs==4.2.0
pluggy==1.4.0
pyarrow==15.0.2
pycodestyle==2.11.1
pycparser==2.22
pydantic==2.5.3
pydantic_core==2.14.6
pyflakes==3.2.0
PyJWT==2.8.0
pyOpenSSL==24.1.0
pytest==8.1.1
python-dateutil==2.9.0.post0
pytz==2024.1
PyYAML==6.0.1
requests==2.31.0
six==1.16.0
snowflake-connector-python==3.8.1
snowflake-snowpark-python==1.15.0
sortedcontainers==2.4.0
tomlkit==0.12.4
typing_extensions==4.11.0
tzdata==2024.1
urllib3==2.2.1

4. What did you do?

Using a mock snowpark session for in memory unit testing we receive warnings for improper actions on a pandas copy of a dataframe.

```python
import warnings
import numpy as np
from snowflake.snowpark import Session
import pytest

@pytest.fixture()
def snowpark_session() -> Session:
    return Session.builder.config("local_testing", True).create()

def test_pandas_error(snowpark_session: Session):
    with warnings.catch_warnings():
        warnings.simplefilter("error")
        snowpark_session.create_dataframe(
            [{"a": 1.1, "b": "s"}, {"a": np.nan, "b": "t"}],
            schema=["a", "b"]
        )
```

During my tests I receive the following warning.
```shell
snowpark/mock/_plan.py:565: FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
[...]

PR #1476, opened 2024-05-01
Fix fill na for numeric types.

DustinMoriarty:

   Fixes SNOW-1358494
   (Gitlab Issue #1474)

   - [ x] I am adding a new automated test(s) to verify correctness of my new code

   This is an alternative solution for #1474. I suspect that the intent was to return None rather than np.nan for numeric types. The other solution that is proposed is in #1475, which removes the effective code entirely.

### DIFF
--- a/src/snowflake/snowpark/mock/_plan.py
+++ b/src/snowflake/snowpark/mock/_plan.py
@@ -561,8 +561,8 @@ def execute_mock_plan(
         for column_name in table.columns:
             sf_type = table.sf_types[column_name]
             table[column_name].sf_type = table.sf_types[column_name]
-            if not isinstance(sf_type.datatype, _NumericType):
-                table[column_name].replace(np.nan, None, inplace=True)
+            if isinstance(sf_type.datatype, _NumericType):
+                table[column_name] = table[column_name].replace(np.nan, None)
         return table
     if isinstance(source_plan, MockSelectExecutionPlan):
         return execute_mock_plan(source_plan.execution_plan, expr_to_alias)
--- a/tests/unit/test_mock_plan.py
+++ b/tests/unit/test_mock_plan.py
@@ -1,0 +1,32 @@
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
+import warnings
+
+import numpy as np
+import pytest
+
+from snowflake.snowpark import Session
+
+
+@pytest.fixture()
+def snowpark_session() -> Session:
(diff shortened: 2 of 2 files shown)


tests/unit/test_mock_plan.py
```diff
@@ -0,0 +1,32 @@
...
+
+from snowflake.snowpark import Session
+
+
+@pytest.fixture()
+def snowpark_session() -> Session:
+    return Session.builder.config("local_testing", True).create()
+
+
+def test_pandas_error(snowpark_session: Session):
```
DustinMoriarty, line 18: Note that this only really fails in newer versions of pandas higher than 2.2 on newer version of python on which 2.2. can actually install. We may need to change tox to iterate on more versions in order to actually capture the failure with the previous version.

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -561,8 +561,8 @@ def execute_mock_plan(
         for column_name in table.columns:
             sf_type = table.sf_types[column_name]
             table[column_name].sf_type = table.sf_types[column_name]
-            if not isinstance(sf_type.datatype, _NumericType):
-                table[column_name].replace(np.nan, None, inplace=True)
+            if isinstance(sf_type.datatype, _NumericType):
```
sfc-gh-aalam, line 564: looks like you remove the `not` here

DustinMoriarty: @sfc-gh-aalam : yes I did.  The old version had no effect for multiple reasons.  I was trying to determine what the original design intent was.  np.nan is a float type, so this would only make sense if it is applied to numeric types. One could also just fill Nan on the whole dataset for that matter.  

I am guessing that the idea here is to model the data which would be returned by a real snowflake connection.  I am pretty sure you would not see numpy anywhere there because numpy is an optional dependency I believe.

sfc-gh-stan: `NaN` is a valid value for Snowflake Float type, so this line should be `if not isinstance(sf_type.datatype, (DoubleType, FloatType)):`. 
We replace `np.nan` with `None` for other types because Pandas converts `None` to `nan` automatically if the dtype is something like `np.int64`, but for Snowflake integer columns should only contain `null` (mapped to `None` in Python) instead of `nan`.

sfc-gh-stan (commented):
Change on line 565 makes sense to me, please update line 564 to only replace for `DoubleType` and `FloatType`.

PR #1497, opened 2024-05-02
SNOW-1357695: Fix int64 not serializable errors on apply(axis=1)

sfc-gh-mvashishtha:

   Fixes SNOW-1357695

   We convert numpy ints to ints explicitly in the UDTF for `groupby.apply` [here](https://github.com/snowflakedb/snowpark-python/blob/31aad1fc33c263e3b2c3403f3dab160712e00f22/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py#L310) and [here](https://github.com/snowflakedb/snowpark-python/blob/31aad1fc33c263e3b2c3403f3dab160712e00f22/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py#L546) but have so far managed to avoid doing so for apply(axis=1). With the new server release, that doesn't work any more, and we need to explicitly convert to ints.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
@@ -221,7 +221,7 @@ def create_udtf_for_apply_axis_1(
                 raise TypeError(f"Unsupported data type {df} from df.apply")
 
             result["value"] = (
-                result["value"].apply(handle_missing_value_in_variant).astype(object)
+                result["value"].apply(handle_missing_value_in_variant).astype(object).apply(convert_numpy_int_result_to_int)
             )
             return result
 
--- a/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
@@ -221,7 +221,10 @@ def create_udtf_for_apply_axis_1(
                 raise TypeError(f"Unsupported data type {df} from df.apply")
 
             result["value"] = (
-                result["value"].apply(handle_missing_value_in_variant).astype(object).apply(convert_numpy_int_result_to_int)
+                result["value"]
+                .apply(handle_missing_value_in_variant)
+                .apply(convert_numpy_int_result_to_int)
+                .astype(object)
             )
             return result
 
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -221,7 +221,10 @@ def end_partition(self, df):  # type: ignore[no-untyped-def] # pragma: no cover
                 raise TypeError(f"Unsupported data type {df} from df.apply")
 
             result["value"] = (
-                result["value"].apply(handle_missing_value_in_variant).astype(object)
+                result["value"]
+                .apply(handle_missing_value_in_variant)
+                .apply(convert_numpy_int_result_to_int)
```
sfc-gh-joshi, line 226: Should we combine these into a single .apply call?

sfc-gh-mvashishtha: I doubt it matters at this point given the huge overheads from temp table creation, joins, and UDTF invocation but it's better practice to combine the calls. Done.

tests/integ/modin/frame/test_apply.py
```diff
@@ -544,7 +544,6 @@ def test_axis_1_multi_index_column_labels_different_levels_negative():
     )
 
 
-@pytest.mark.skip(reason="SNOW-1358681")
```
sfc-gh-azhan, line 547: what about other skipped ones? are they not due to this?

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -222,8 +222,11 @@ def end_partition(self, df):  # type: ignore[no-untyped-def] # pragma: no cover
 
             result["value"] = (
                 result["value"]
-                .apply(handle_missing_value_in_variant)
-                .apply(convert_numpy_int_result_to_int)
+                .apply(
```
sfc-gh-azhan, line 225: I agree with @sfc-gh-mvashishtha those changes should not have a significant perf impact.

sfc-gh-joshi (approved):

tests/integ/modin/frame/test_apply.py
sfc-gh-mvashishtha: They have other causes. `test_axis_1_return_not_json_serializable_label` failed because the error message changed. `test_axis_1_date_time_timestamp_type` failed because snowflake started returning float64 instead of int64.

sfc-gh-vbudati (approved):
Thanks Mahesh!

PR #1498, opened 2024-05-02
SNOW-1296757: local testing fix math.nan related issue

sfc-gh-aling:

   Fixes SNOW-NNNNNNN

- Fixed a bug that when there is `None` value in arithmetic calculation, the output should remain `None` instead of `math.nan`.
- Fixed a bug in function `sum` and `covar_pop` that when there is `math.nan` in the data, the output should also be `math.nan`.


src/snowflake/snowpark/mock/_snowflake_data_type.py
```diff
@@ -134,6 +134,15 @@ def normalize_output_sf_type(t: DataType) -> DataType:
...
 
+def reset_nan_to_none(col_a, col_b, res_col):
+    # in pandas arithmetic operation, None are automatically convert to nan
+    if isinstance(res_col.sf_type.datatype, _NumericType):
+        for idx, (x, y) in enumerate(zip(col_a, col_b)):
+            if x is None or y is None:
+                res_col[idx] = None
+    return res_col
+
+
```
sfc-gh-jrose, line 145: Copied from the other PR:
I don' t think this works unless all the columns are object type. Otherwise pandas will just set the value back to nan.

Neither are object:
```
>>> a = pd.Series([1, 1, 1, None], index=['a', 'b', 'c', 'd'])
>>> b = pd.Series([1, None, 1, 1], index=['a', 'b', 'c', 'd'])
>>> res = a + b
>>> res
a    2.0
b    NaN
c    2.0
d    NaN
dtype: float64
>>> reset_nan_to_none(a, b, res)
a    2.0
b    NaN
c    2.0
d    NaN
dtype: float64
```
One side is object:
```
>>> a = pd.Series([1, 1, 1, None], index=['a', 'b', 'c', 'd'], dtype=object)
>>> b = pd.Series([1, None, 1, 1], index=['a', 'b', 'c', 'd'])
>>> res = a + b
>>> res
a    2.0
b    NaN
c    2.0
d    NaN
dtype: object
>>> reset_nan_to_none(a, b, res)
a     2.0
b     NaN
c     2.0
d    None
dtype: object
```

Both are object:
```
>>> a = pd.Series([1, 1, 1, None], index=['a', 'b', 'c', 'd'], dtype=object)
>>> b = pd.Series([1, None, 1, 1], index=['a', 'b', 'c', 'd'], dtype=object)
>>> res = a + b
>>> res
a      2
b    NaN
c      2
d    NaN
dtype: object
>>> reset_nan_to_none(a, b, res)
a       2
b    None
c       2
d    None
dtype: object
```

sfc-gh-aling: thanks for calling this out, I re-checked our current logic and I found that all the columns are guaranteed object type in case of None value existing for numeric values. so what happens is:

1. we create the pd dataframe with object: https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/mock/_plan.py#L553
2. when we perform [fix type drift in binary operation](https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/mock/_plan.py#L1524-L1529), if we detect `None`, we simply don't convert the dtype: https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/mock/_util.py#L200-L204.

sfc-gh-jrose (approved):

src/snowflake/snowpark/mock/_snowflake_data_type.py
```diff
@@ -330,6 +339,7 @@ def __add__(self, other):
         result = super().__add__(other)
         if self.sf_type:
             result.sf_type = calculate_type(self.sf_type, other.sf_type, op="+")
+        result = reset_nan_to_none(self, other, result)
```
sfc-gh-stan, line 342: But what if we want to have `nan` in the result?

sfc-gh-stan (approved):

src/snowflake/snowpark/mock/_snowflake_data_type.py
sfc-gh-aling: I'm good with the name change

PR #1513, opened 2024-05-03
SNOW-1357844: fix flaky test caused by oob telemetry datetime comparsion

sfc-gh-aling:

   Fixes SNOW-1357844

pytest.approx only supports number, datetime is not supported, so previously implementation doesn't work at all.
change to use timestamp to the approx equal

### DIFF
--- a/tests/mock_unit/test_oob_telemetry.py
+++ b/tests/mock_unit/test_oob_telemetry.py
@@ -48,7 +48,7 @@ def test_unit_oob_connection_telemetry(caplog, local_testing_telemetry_setup):
             assert event["UUID"]
             assert datetime.strptime(
                 event["Created_on"], "%Y-%m-%d %H:%M:%S"
-            ).timestamp() == pytest.approx(now_time.timestamp(), abs=1)
+            ).timestamp() == pytest.approx(now_time.timestamp(), abs=3)
             assert event["Message"]["Connection_UUID"] == (
                 None if idx == 0 else connection_uuid
             )
@@ -127,7 +127,7 @@ def test_unit_oob_log_not_implemented_error(caplog, local_testing_telemetry_setu
         assert event["UUID"]
         assert datetime.strptime(
             event["Created_on"], "%Y-%m-%d %H:%M:%S"
-        ).timestamp() == pytest.approx(now_time.timestamp(), abs=1)
+        ).timestamp() == pytest.approx(now_time.timestamp(), abs=3)
         assert event["Message"]["Connection_UUID"] == (
             None if idx == 0 else connection_uuid
         )
--- a/tests/mock_unit/test_oob_telemetry.py
+++ b/tests/mock_unit/test_oob_telemetry.py
@@ -5,7 +5,7 @@ import json
 import logging
 import re
 import uuid
-from datetime import datetime, timedelta
+from datetime import datetime
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


tests/mock_unit/test_oob_telemetry.py
```diff
@@ -48,7 +48,7 @@ def test_unit_oob_connection_telemetry(caplog, local_testing_telemetry_setup):
             assert event["UUID"]
             assert datetime.strptime(
                 event["Created_on"], "%Y-%m-%d %H:%M:%S"
-            ) == pytest.approx(now_time, abs=timedelta(seconds=1))
+            ).timestamp() == pytest.approx(now_time.timestamp(), abs=1)
```
sfc-gh-mvashishtha, line 51: this test still doesn't look deterministic. Can we mock the time that we use to get `Created_on` and patch in a constant time?

sfc-gh-aling: it would be tricky to mock as set `Create_on` is set up `datetime.now()` in the code: https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/mock/_telemetry.py#L64.

I can increase the value to make the test more tolerant.

sfc-gh-mvashishtha: unittest.mock lets you patch a particular function. can we try that? https://docs.python.org/3/library/unittest.mock-examples.html#partial-mocking

sfc-gh-mvashishtha:
@sfc-gh-aling could you please `pip install pytest-repeat pytest-xdist` and run the test 1000 times before and after this change (I think parallelism with `-n 32` might help speed it up) to check that the flakiness goes away?

tests/mock_unit/test_oob_telemetry.py
sfc-gh-aling: thanks for sharing, let me try

sfc-gh-aling: changed to use mock and it's working perfectly, new thing learned

sfc-gh-aling:
@sfc-gh-mvashishtha I chanced to use mock, we shall be good now. thanks for the suggestion

tests/mock_unit/test_oob_telemetry.py
```diff
@@ -28,15 +29,18 @@ def test_unit_oob_connection_telemetry(caplog, local_testing_telemetry_setup):
     oob_service = LocalTestOOBTelemetryService.get_instance()
     # clean up queue first
     oob_service.export_queue_to_string()
-    with caplog.at_level(logging.DEBUG, logger="snowflake.snowpark.mock._telemetry"):
+    with caplog.at_level(
+        logging.DEBUG, logger="snowflake.snowpark.mock._telemetry"
+    ), mock.patch("snowflake.snowpark.mock._telemetry.datetime") as mock_datetime:
```
sfc-gh-mvashishtha, line 34: nit: here and below, can we only put the `mock.patch` around the method calls that require it? that way it's clearer why we're doing the patching, the diff is smaller, and it's easier to understand the indentation below. I think that might mean just putting the context around the `log_session_creation`.

sfc-gh-mvashishtha (commented):
@sfc-gh-aling one more nit

 > command pytest tests/mock_unit/test_oob_telemetry.py::test_unit_oob_log_not_implemented_error --count 1000 -n 32 passes

Thanks for checking, but it seems to pass on the main branch as well, so it doesn't reproduce the flake. That's okay. I think mocking the time should work.

tests/mock_unit/test_oob_telemetry.py
sfc-gh-aling: sure thing. I updated the ctx mgr at a reasonable granularity and added comments.

sfc-gh-mvashishtha (approved):
LGTM! thanks

sfc-gh-stan (approved):

PR #1529, opened 2024-05-07
[SNOW-1357748]: Update read_snowflake to use string matching

sfc-gh-rdurrani:

   

   Fixes SNOW-1357748

This PR updates read_snowflake to use string matching for the order by warning.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/utils.py
@@ -152,41 +152,29 @@ def _is_table_name(table_name_or_query: str) -> bool:
 
 
 def _check_if_sql_query_contains_order_by_and_warn_user(
-    logical_plan: native_pd.DataFrame,
+    sql_query_text: str,
 ) -> bool:
     """
-    Checks whether the sql query corresponding to the logical plan passed in contains an order by clause
+    Checks whether the sql query passed in contains an order by clause
     and warns the user that ORDER BY will be ignored currently.
 
     Args:
-        logical_plan: The logical plan to check.
+        sql_query_text: The SQL Query to check.
 
     Returns:
         Whether or not the query contains an order by.
     """
-    # We need to determine if the query contains an ORDER BY, which we can determine by looking at the
-    # logical plan. If a query contains an ORDER BY that affects the ordering of the final result,
-    # the logical plan will contain an operation row whose operator name is Sort.
-    # Example of a logical plan with an ORDER BY:
-    #     id                             operation ...
-    # 0    0                        ResultFinalize ...
-    # 1    1                     Exchange (SINGLE) ...
-    # 2    2                          ResultWorker ...
-    # 3    3                            Projection ...
-    # 4    4                                  Sort ...
-    # 5    5                      Exchange (RANGE) ...
-    # 6    6                        RangePartition ...
-    # 7    7                  Exchange (BROADCAST) ...
-    # 8    8                                Sample ...
-    # 9    9                     Exchange (SINGLE) ...
-    # 10  10                                Sample ...
-    # 11  11                             TableScan ...
-    # 12  12                             TableScan ...
-    # (ellipsis denotes additional columns that have been removed since they are unnecessary for this example.)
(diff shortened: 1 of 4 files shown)


sfc-gh-azhan (approved):

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -152,41 +152,29 @@ def _is_table_name(table_name_or_query: str) -> bool:
...
-    # 12  12                             TableScan ...
-    # (ellipsis denotes additional columns that have been removed since they are unnecessary for this example.)
-    # Have to use `strip` since the names of the operators contain indentation.
-    contains_order_by = (
-        logical_plan["operation"].str.lower().str.contains("sort")
-    ).any()
+    # We need to determine if the query contains an ORDER BY. We previously looked
+    # at the logical plan in order to determine if there was an ORDER BY; however,
+    # the output schema of the `EXPLAIN` SQL statement (which was used to generate
+    # the logical plan) seems to not be stable across releases, so instead, we
```
sfc-gh-helmeleegy, line 170: Could it be that some optimization kicks in that manages to get rid of the sort operation in the logical plan? Can we tell if for some of the queries we generate in our tests, the order by clause can be legitimately optimized away? If so, then perhaps we should just ensure that the queries we use in our tests cannot have the order by clauses get optimized away (e.g. when the ordering is applied in the outermost query).

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -152,41 +152,29 @@ def _is_table_name(table_name_or_query: str) -> bool:
...
+    # check to see if the text of the SQL query contains "ORDER BY".
+    # Note: This method may sometimes raise false positives, e.g.:
+    # SELECT "ORDER BY COLUMN", * FROM TABLE;
+    # The above query shouldn't raise a warning, but since we are using
+    # string matching, we will get a false positive and raise a warning.
+    order_by_pattern = r"\s+order\s+by\s+"
+    contains_order_by = re.search(order_by_pattern, sql_query_text.lower()) is not None
     if contains_order_by:
         # If the query contains an ORDER BY, we need to warn the user that
         # the ordering induced by the ORDER BY is not guaranteed to be preserved
```
sfc-gh-helmeleegy, line 180: When the order by clause is immaterial to the final answer, then we may not want to give any warning to the user.
For example: `select count(*) from T order by a`. In this case, the compiler should optimize away the `order by` clause, and the sort operation should not appear in the `explain` result.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -152,41 +152,29 @@ def _is_table_name(table_name_or_query: str) -> bool:
...
-    # 11  11                             TableScan ...
-    # 12  12                             TableScan ...
-    # (ellipsis denotes additional columns that have been removed since they are unnecessary for this example.)
-    # Have to use `strip` since the names of the operators contain indentation.
-    contains_order_by = (
-        logical_plan["operation"].str.lower().str.contains("sort")
-    ).any()
+    # We need to determine if the query contains an ORDER BY. We previously looked
+    # at the logical plan in order to determine if there was an ORDER BY; however,
+    # the output schema of the `EXPLAIN` SQL statement (which was used to generate
```
sfc-gh-helmeleegy, line 169: What exactly do we mean by the "output schema of ... EXPLAIN .. seems to not be stable"? Is it that the `operation` column goes missing? Or the data content in that column may differ across releases. The latter case is understandable because it depends on what optimizations are applied. The former case however, should be far less common.

sfc-gh-rdurrani: The operation column is always present, but there are other columns we were previously relying on (in addition to operation) that got changed. There are also different data values in the operation column - e.g., before this release, ORDER BY X LIMIT N was a SortLimit operation, but now is a TopK operation.

sfc-gh-helmeleegy (approved):
LGTM! We resolved all my comments in an offline discussion.

PR #1534, opened 2024-05-07
SNOW-904270: [Local Testing] Add support for DataFrame.pivot

sfc-gh-jrose:

   Fixes SNOW-904270


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
     - get
     - concat
     - concat_ws
+  - snowflake.snowpark.DataFrame:
```
sfc-gh-aling, line 20: nit: this goes to 1.17.0

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -1376,6 +1386,97 @@ def outer_join(base_df):
...
+                internal_feature_name=type(source_plan).__name__,
+                raise_error=NotImplementedError,
+            )
+
+        grouping_columns = [
+            plan.session._analyzer.analyze(c) for c in source_plan.grouping_columns
+        ]
+        indices = grouping_columns or [
+            col for col in child_rf.keys() if col not in {agg_column, pivot_column}
+        ]
```
sfc-gh-aling, line 1433: just for my own learning purpose, grouping_columns are columns specified in `order by` and if `order by` not specified then it will use the rest of the columns for grouping (excluding agg and pivot column) ?

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -1376,6 +1386,97 @@ def outer_join(base_df):
...
+
+        result = child_rf.pivot_table(
+            columns=pivot_column,
+            values=agg_column,
+            aggfunc=agg_function,
+            index=indices,
+        )
+        result.reset_index(inplace=True)
+
+        # Pandas pivot can include many levels of indicies. we're only interested in the last one
```
sfc-gh-aling, line 1463: not blocker: curious what does "many levels of indicies" in pandas mean and what's the relation between the last one and other levels?

tests/integ/scala/test_dataframe_aggregate_suite.py
sfc-gh-aling, line 1: there is another test test_column_suite.py::test_pivot_with_multiple_queries

src/snowflake/snowpark/mock/_plan.py
sfc-gh-jrose: Correct. That behavior is not documented on the main pivot page, but it is documented [here](https://community.snowflake.com/s/article/Pivot-returns-more-rows-than-expected).

src/snowflake/snowpark/mock/_plan.py
sfc-gh-jrose: This mostly only matter when there are multiple aggregations specified. You might get a list of columns like `[["sum", "A"], ["avg", "B"]]`. I originally assumed that since it was possible to do that with pandas it would be with snowflake as well, but after some testing it looks like snowflake only ever allows one aggregation per pivot so this probably isn't an issue. I'll remove

tests/integ/scala/test_dataframe_aggregate_suite.py
sfc-gh-jrose: The functionality in that test is not supported yet. It relies on a subquery to provide the pivot values which I've left as a NotImplementError for now.

src/snowflake/snowpark/mock/_plan.py
sfc-gh-aling: nit: can you help document this behavior in the code as comment? so that in the future when someone needs to work on this they can quickly ramp up

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -1379,6 +1389,94 @@ def outer_join(base_df):
...
+        if pivot_values:
+            result = result[list(indices) + pivot_values]
+
+        # Non-indice columns lack an sf_type, add them back in.
+        for res_col in set(result.columns) - set(indices):
+            # fill_na will not fill na like values with None, but it can replace them with a sentinel value
+            filled = result[res_col].fillna(sentinel)
+
+            # Sentinel values are replaced with None, then all Nones are replaced with the default if provided
+            data = filled.replace({sentinel: None}).replace({None: default}).values
```
sfc-gh-aling, line 1476: dumb question, is it possible that the original value inside the pd dataframe is an object()?

sfc-gh-jrose: It could be an object, but even so it would never equal the sentinel object. An initialized base object is only ever equal to itself. You can see this with an example:
```
>>> object() == object()
False
```

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -1379,6 +1389,98 @@ def outer_join(base_df):
...
```
sfc-gh-stan, line 1480: Is this different than `result[res_col].fillna(default)`?

sfc-gh-jrose: There can be Nones in the dataframe if there were Nones in the source data. There can also be sentinel values generated by the aggregation functions. I could probably refactor to go directly to Nones skipping the sentinel value altogether. 

Originally I wanted to use the sentinel value as a fill value for the pivot_table call so that I could preserve values that were NaN for reasons unrelated to the pivot, but a recent change in pandas prevents both using None as a fill value and using the object sentinel value trick. This caused a test failure in my original approach since I was on pandas 2.0.0 and the test env was on 2.2.2

sfc-gh-stan (approved):

sfc-gh-aling (approved):

PR #1536, opened 2024-05-08
[SNOW-1370566]: Add support for cache user API

sfc-gh-rdurrani:

   

   Fixes SNOW-1370566

Adds support for users to cache computationally expensive intermediate results.


src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -238,3 +238,20 @@ def to_pandas(
         Name: Animal, dtype: object
     """
     return self._to_pandas(statement_params=statement_params, **kwargs)
+
+
+@register_dataframe_accessor("cache")
+@snowpark_pandas_telemetry_method_decorator
+def cache(self) -> pd.DataFrame:
```
sfc-gh-jkew, line 245: I think it makes sense to keep the same signature as the snowpark one: https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/latest/api/snowflake.snowpark.DataFrame.cache_result

sfc-gh-rdurrani: I had originally thought so too, but `cache_result` doesn't seem very pandas-esque. I originally had `cache_dataframe` and `cache_series`, but that was non-ideal since what if the user wants to cache the result of an operation without knowing if its a series or a DataFrame - I can change it to cache_result if folks feel strongly though.

tests/integ/modin/frame/test_cache.py
```diff
@@ -0,0 +1,75 @@
...
+    snow_df = pd.DataFrame(native_df)
+    cached_snow_df = snow_df.cache()
+    assert cached_snow_df is snow_df
+    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+        cached_snow_df, native_df
+    )
+
+
+@sql_count_checker(query_count=3, join_count=1)
+def test_cache_dataframe_complex(
```
sfc-gh-mvashishtha, line 60: This test doesn't show us how `cache_result` is changing counts of queries, joins, etc. I think it would be more useful to compare query counts for a chain of 1-2 pandas methods with and without a `cache_result`. e.g. with a few different `SqlCounter`s at each step we can check:

- each `cache_result` costs 1 query
- a `cache_result` after a `merge` will cost a `join_count=1` in the `SqlCounter`, but a subsequent method call on the cached dataframe incurs no joins.

ditto for `Series`.

docs/source/modin/dataframe.rst
```diff
@@ -207,3 +207,10 @@ DataFrame
     DataFrame.to_pandas
     DataFrame.to_snowflake
     DataFrame.to_snowpark
+
```
sfc-gh-lspiegelberg, line 210: It might make sense to call this out here, but a single API in its own rubric seems overkill. Can we maybe fold it with the previous category? We can rename it to `Snowflake specific API extensions (IO / conversion / caching)` if you'd like.

src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
```diff
@@ -46,6 +46,7 @@
 
 ### New Features
 - Added partial support for `SeriesGroupBy.apply` (where the `SeriesGrouBy` is obtained through `DataFrameGroupBy.__getitem__`).
```
sfc-gh-lspiegelberg, line 48: Should we use `cache_result` similar to Snowpark? I'd be even fine to define an alias `cache` and `cache_result` and add both to API docs.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -27,6 +27,7 @@
     ROW_POSITION_COLUMN_LABEL,
     append_columns,
     assert_duplicate_free,
```
sfc-gh-lspiegelberg, line 29: is it cache_result or cache? Let's maybe use one name throughout our code base, and if desired define an alias at the frontend layer.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -691,6 +692,22 @@ def ensure_row_count_column(self) -> "InternalFrame":
             index_column_snowflake_quoted_identifiers=self.index_column_snowflake_quoted_identifiers,
         )
 
+    def materialize_and_cache(self) -> "InternalFrame":
```
sfc-gh-lspiegelberg, line 695: hm I feel a proper name here might be `persist_to_temporary_table`. Materialize is often used in a different context and this may be misleading. Would you remind renaming?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1123,6 +1123,12 @@ def to_snowpark(
             index, index_label
         )
 
+    def cache_result(self) -> None:
```
sfc-gh-lspiegelberg, line 1126: should this return not the SnowflakeQueryCompiler itself/a cached version of it? Can we add some comments about temp table and that this lives for the duration of the session?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
sfc-gh-lspiegelberg: `cache` is I think the Spark API, I think we can define an alias here and use both.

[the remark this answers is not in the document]
sfc-gh-lspiegelberg: +1, it would be great to add this to best practices.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -238,3 +238,20 @@ def to_pandas(
...
+    """
+    Materialize and cache the current Snowpark pandas DataFrame inplace.
```
sfc-gh-lspiegelberg, line 247: persists the current Snowpark pandas DataFrame to a temporary table for the duration of a session to improve subsequent operations.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -238,3 +238,20 @@ def to_pandas(
...
+
+    Returns:
+        Snowpark pandas DataFrame
+
+    Note:
+        This method is inplace - the returned DataFrame is a reference to the same
```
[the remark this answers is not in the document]
sfc-gh-lspiegelberg: +1

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
```diff
@@ -203,3 +203,20 @@ def to_pandas(
     Name: Animal, dtype: object
     """
     return self._to_pandas(statement_params=statement_params, **kwargs)
+
+
+@register_series_accessor("cache")
+@snowpark_pandas_telemetry_method_decorator
+def cache(self) -> Series:
+    """
+    Materialize and cache the current Snowpark pandas Series inplace.
```
sfc-gh-lspiegelberg, line 212: docstring, can we maybe do docstring reuse here?

tests/integ/modin/frame/test_cache.py
sfc-gh-lspiegelberg: +1, it would be great to have a few tests with chained operations and @sfc-gh-mvashishtha has a great suggestion with tracking changing query counts here.

sfc-gh-lspiegelberg: We may even want to test .cache() with certain operator groups, i.e.

cache() after groupby()
cache() after agg()
cache() after pivot()
cache() after read_snowflake()
cache() after filter()
cache() after apply() # <-- this likely should make it into best practice guide

tests/integ/modin/frame/test_cache.py
```diff
@@ -0,0 +1,75 @@
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
```
sfc-gh-lspiegelberg, line 3: Does not need to be in this file, but a test showing that the cached table is dropped when the session is closed would be great too.

sfc-gh-lspiegelberg (commented):
thanks for quickly getting a PR up for this. I think we may need a bit more testing and some tiny changes to the API. Hopefully it's not too much work, and we can get it in soon.

sfc-gh-helmeleegy (commented):
This is generally great. I think we’ll also need to have a method to uncache a given dataframe/series, which is especially useful for long-running sessions. It addresses the concern raised earlier about the potentially high cost of keeping many temp tables around for a long time. Spark already has unpersist among other ways of clearing caches.

This can be done in a follow-up PR but let’s create a jira for it.

docs/source/modin/dataframe.rst
```diff
@@ -207,3 +207,10 @@ DataFrame
...
+.. rubric:: Caching
```
sfc-gh-vbudati, line 211: update the dataframe_supported and series_supported rst files as well

src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
```diff
@@ -46,6 +46,7 @@
...
+- Added `DataFrame.cache` and `Series.cache` methods for users to cache computationally intensive intermediate results for reuse.
```
sfc-gh-jkew, line 49: I think we need a `performance.rst` file, in `docs/source/modin/performance.rst` and I think this should be the first entry. We might want to add more later.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
sfc-gh-rdurrani: Changed to `cache_result`!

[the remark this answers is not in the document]
sfc-gh-rdurrani: Added docstring, and created followup ticket: https://snowflakecomputing.atlassian.net/browse/SNOW-1435412

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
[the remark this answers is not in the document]
sfc-gh-rdurrani: Discussed offline - decided to include an `inplace` parameter.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
sfc-gh-rdurrani: Makes sense!

src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
```diff
@@ -71,10 +72,6 @@
   Snowpark pandas deviates from this behavior and will maintain the same behavior as pandas from versions 1.5.x.
 - Changed the import path of Snowpark pandas package to use Modin 0.28.1 instead. The new recommended import statement is `import modin.pandas as pd; import snowflake.snowpark.modin.plugin`.
 
-### New Features
-- Added partial support for `SeriesGroupBy.apply` (where the `SeriesGrouBy` is obtained through `DataFrameGroupBy.__getitem__`).
```
sfc-gh-mvashishtha: Should we add back the line about `SeriesGroupBy.apply`?

src/snowflake/snowpark/modin/plugin/extensions/utils.py
```diff
@@ -0,0 +1,84 @@
...
+...     {object_var_name}_obj.reset_index(drop=True)
+...     end = perf_counter()
+...     return end - start
+...
+
+>>> uncached_time = time_reset_index({object_var_name})
+
+>>> cached_time = time_reset_index(new_{object_var_name})
+
+>>> assert cached_time < uncached_time
```
sfc-gh-mvashishtha, line 57:
```suggestion
```

I'd recommend not counting on a perform benchmark to pass every time we run this doctest.

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -0,0 +1,159 @@
...
+import snowflake.snowpark.modin.plugin  # noqa: F401
+from tests.integ.modin.sql_counter import SqlCounter
+from tests.integ.modin.utils import (
+    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck,
+)
+
+
+def assert_empty_snowpark_pandas_equals_to_pandas(snow_df, native_df):
+    native_snow_df = snow_df.to_pandas()
+    assert native_snow_df.empty is native_df.empty is True
```
sfc-gh-mvashishtha, line 22: What's this line supposed to check? `a is b` always returns a bool. Do we just need to verify that `native_df.empty`? Then we should do `assert native_df.empty`.

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -0,0 +1,159 @@
...
+def test_cache_result_empty_dataframe(inplace):
+    native_df = native_pd.DataFrame()
+    snow_df = pd.DataFrame()
+    with SqlCounter(query_count=1):
+        cached_snow_df = cache_and_return_df(snow_df, inplace)
+    if not inplace:
+        # If inplace, then cached_snow_df IS snow_df, so this check
+        # is meaningless.
+        with SqlCounter(query_count=1):
+            assert np.all((cached_snow_df == snow_df).values)
```
sfc-gh-mvashishtha, line 57: here and below, why check `values`? That skips index values. Can we do something like `assert_snowpark_pandas_equals_to_pandas_without_dtypecheck` instead? Also, we have a `assert_empty_snowpark_pandas_equals_to_pandas` below here?

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -0,0 +1,159 @@
...
+
+@pytest.mark.parametrize("inplace", [True, False])
+def test_cache_result_empty_dataframe(inplace):
+    native_df = native_pd.DataFrame()
+    snow_df = pd.DataFrame()
+    with SqlCounter(query_count=1):
+        cached_snow_df = cache_and_return_df(snow_df, inplace)
+    if not inplace:
+        # If inplace, then cached_snow_df IS snow_df, so this check
+        # is meaningless.
```
sfc-gh-mvashishtha, line 55: We should still check the values when `inplace=True`.

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -0,0 +1,159 @@
...
+    cached_snow_df = cached_snow_df.set_index("b", drop=False)
+    native_df = native_df.set_index("b", drop=False)
+    with SqlCounter(query_count=1):
+        assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+            cached_snow_df, native_df, check_freq=False
+        )
+
+
+@pytest.mark.parametrize("inplace", [True, False])
+class TestCacheResultReducesLatency:
```
sfc-gh-mvashishtha, line 128: It seems too brittle to hope that we will consistently reduce latency. I think overall query count and count of things like `join`, `udtf`, etc. is more helpful. I suggest that we have a few test cases for different scenarios where users might want to cache result and that we show how caching reduces query counts compared to not caching. See @sfc-gh-lspiegelberg 's comment [here](https://github.com/snowflakedb/snowpark-python/pull/1536#discussion_r1594326009).

tests/integ/modin/frame/test_cache_result.py
sfc-gh-rdurrani: If `inplace` is True, then cached_snow_df is a reference to snow_df (based off of how I wrote the function), so I'm not sure that it makes sense, since that check is always vacuously true?

sfc-gh-rdurrani: To clarify - how I wrote the helper function for the test, so that we can get a reference to the cached dataframe after calling the helper function regardless of the value of inplace.

tests/integ/modin/frame/test_cache_result.py
sfc-gh-rdurrani: These are both Snowpark pandas DataFrames - but I'll go ahead and update!

tests/integ/modin/frame/test_cache_result.py
sfc-gh-mvashishtha: This way of checking that is confusing since people have to think about the associativity of the `is` operator. Can we just do `assert native_snow_df.empty` and then `assert_native_df.empty`?

tests/integ/modin/frame/test_cache_result.py
sfc-gh-mvashishtha: We can make a deep copy of the `snow_df` before the caching and check that the values are the same as the values in `snow_df` after the inplace `cache_result`.

docs/source/modin/dataframe.rst
sfc-gh-rdurrani: Seems like we don't include to_snowpark/to_snowflake in those files - should we still include this even though it is a non-pandas API?

tests/integ/modin/frame/test_cache_result.py
sfc-gh-rdurrani: Sounds good!

src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
sfc-gh-rdurrani: I believe that this diff is because of a merge.

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -31,66 +41,76 @@ def assert_empty_snowpark_pandas_equals_to_pandas(snow_df, native_df):
...
     else:
         # If inplace=True, there is no return, so we set `cached_snow_df` equal to `snow_df`.
-        snow_df.cache_result(inplace=inplace)
         cached_snow_df = snow_df
+        cached_snow_df.cache_result(inplace=inplace)
     return cached_snow_df
 
 
+def perform_chained_operations(df, module):
+    df = df.reset_index(drop=True)
```
sfc-gh-mvashishtha, line 71: add a comment that we expect this to be expensive because each `concat` unions results, and each result may itself contain unions?

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -125,35 +142,135 @@ def test_cache_result_dataframe_complex_correctness(
...
+            cached_snow_df = cache_and_return_df(snow_df, inplace)
+
+        with SqlCounter(query_count=1, union_count=9):
+            cached_snow_df = perform_chained_operations(cached_snow_df, pd)
+            assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+                cached_snow_df, native_df
+            )
+
+    def test_cache_result_post_apply(self, inplace, simple_test_data):
+        native_df = perform_chained_operations(
```
sfc-gh-mvashishtha, line 220: add a comment that the additional caching isn't helping much here because apply(axis=1) already caches an intermediate result?

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -125,35 +142,135 @@ def test_cache_result_dataframe_complex_correctness(
...
+
+        with SqlCounter(query_count=2, union_count=9):
+            repr(cached_snow_df)
+            cached_snow_df = perform_chained_operations(cached_snow_df, pd)
+            assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+                cached_snow_df,
+                native_df,
+            )
+
+    def test_cache_result_post_applymap(self, inplace, simple_test_data):
```
sfc-gh-mvashishtha, line 244: I don't think `applymap` does caching of its own. The query overheads in this test case are from setting up the UDF and defining it, so we just end up paying an extra query to cache the UDF result. I would add a comment explaining that.

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -125,35 +142,135 @@ def test_cache_result_dataframe_complex_correctness(
...
+            repr(snow_df)
+            snow_df = perform_chained_operations(snow_df, pd)
+            assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+                snow_df, native_df
+            )
+
+        with SqlCounter(
+            query_count=10,
+            high_count_expected=True,
+            high_count_reason="caching result of udf seems to have high query count.",
```
sfc-gh-mvashishtha, line 265:
```suggestion
            high_count_reason="query count comes from setting up and defining the UDF",
```

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -125,35 +142,135 @@ def test_cache_result_dataframe_complex_correctness(
...
+        ):
+            snow_df = pd.DataFrame(simple_test_data).applymap(lambda x: x + x)
+            cached_snow_df = cache_and_return_df(snow_df, inplace)
+
+        with SqlCounter(query_count=2, union_count=9, udf_count=0):
```
sfc-gh-mvashishtha, line 270: maybe comment that one query comes from `repr` and the next from `assert_snowpark_pandas_equals_to_pandas_without_dtypecheck`.

tests/integ/modin/series/test_cache_result.py
sfc-gh-mvashishtha, line 1: ditto all my comments on frame/test_cache_result.py :)

sfc-gh-mvashishtha (commented):
@sfc-gh-rdurrani this is looking better to me, but I still have some comments.

sfc-gh-lspiegelberg:
> 1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.
>    
>    Fixes [SNOW-1370566](https://snowflakecomputing.atlassian.net/browse/SNOW-1370566)
> 2. Fill out the following pre-review checklist:
>    
>    * [x]  I am adding a new automated test(s) to verify correctness of my new code
>    * [ ]  I am adding new logging messages
>    * [ ]  I am adding a new telemetry message
>    * [ ]  I am adding new credentials
>    * [ ]  I am adding a new dependency
> 3. Please describe how your code solves the related issue.
> 
> Adds support for users to cache computationally expensive intermediate results.

Can you Update the PR description to reflect what new APIs are introduced, and what this PR does? Thanks!

docs/source/modin/index.rst
```diff
@@ -18,4 +18,5 @@ For your convenience, here is all the :doc:`Supported APIs <supported/index>`
     groupby
     resampling
     numpy
-    All supported APIs <supported/index>
\ No newline at end of file
+    All supported APIs <supported/index>
+    performance
```
sfc-gh-lspiegelberg, line 22: uppercase?

docs/source/modin/performance.rst
```diff
@@ -0,0 +1,50 @@
+Performance Recommendations
+===========================
+
+This page contains recommendations to help improve performance when using Snowpark pandas.
```
sfc-gh-lspiegelberg, line 4: nit: ...when using Snowpark pandas API.

docs/source/modin/performance.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Caching Intermediate Results
+----------------------------
+Snowpark pandas uses a lazy paradigm - when operations are called on a Snowpark pandas object,
+the operation is not evaluated until absolutely necessary - e.g. when displaying results to users.
+This paradigm enables larger queries to be optimized using Snowflake's SQL Query Optimizer, but in certain
+cases, can lead to expensive recomputations. Take, for example, the following toy code snippet:
```
sfc-gh-lspiegelberg, line 11: Let's maybe try to be a bit less colloquial here, you can remove things like "toy code snippet". It's sufficient to describe what the example does to get your point across :) E.g.,

This code snippet creates a 30x5 DataFrame  using concatenation of 30 smaller 1x5 DataFrames, .... Any operation that uses `df`, will perform the concatenation lazily. However, this may be problematic when reusing `df` multiple times for output operations, e.g., when inspecting it between operations using `print(df)`. Instead, consider caching the result after performing the `concat` operation using `df.cache_result`.

docs/source/modin/performance.rst
```diff
@@ -0,0 +1,50 @@
...
+    # Session.builder.create() will create a default Snowflake connection.
+    Session.builder.create()
+    df = pd.concat([pd.DataFrame([range(i, i+5)]) for i in range(0, 150, 5)])
+    df = df.cache_result(inplace=False)
+    print(df)
+    df = df.reset_index(drop=True)
+    print(df)
+
+Consider using the ``cache_result`` API whenever a DataFrame or Series that is expensive to compute sees high reuse.
+One caveat is after the ``apply`` and ``applymap`` APIs, which can sometimes eagerly materialize results for metadata purposes.
```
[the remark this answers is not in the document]
sfc-gh-lspiegelberg: I think `apply`/`applymap` could be a footnote or left out.

docs/source/modin/series.rst
```diff
@@ -216,5 +214,14 @@ Series
     Series.str
     Series.dt
 
+.. rubric:: Snowflake Specific
```
sfc-gh-lspiegelberg, line 217: Can we change the order and put this first? I would put the new/deviating APIs first so readers see them immediately. I assume they tend to know pandas, so we need to make sure these get visibility.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1133,6 +1133,12 @@ def to_snowpark(
             index, index_label
         )
 
+    def cache_result(self) -> "SnowflakeQueryCompiler":
+        """
+        Materializes and caches this query compiler in place.
```
sfc-gh-lspiegelberg, line 1138: I am confused about the inplace here, how is this inplace?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -238,3 +239,17 @@ def to_pandas(
         Name: Animal, dtype: object
     """
     return self._to_pandas(statement_params=statement_params, **kwargs)
+
+
+@register_dataframe_accessor("cache_result")
+@_add_cache_result_docstring
+@snowpark_pandas_telemetry_method_decorator
+def cache_result(self, inplace: bool = False) -> pd.DataFrame:
```
sfc-gh-lspiegelberg, line 247: return type should be `Optional[pd.DataFrame]` because of `inplace`?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -238,3 +239,17 @@ def to_pandas(
...
+    """
+    Persists the current Snowpark pandas DataFrame to a temporary table that lasts the duration of the session.
```
sfc-gh-lspiegelberg, line 249: Can we add parameter description for inplace?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -238,3 +239,17 @@ def to_pandas(
...
+    """
+    new_qc = self._query_compiler.cache_result()
+    if not inplace:
```
sfc-gh-lspiegelberg, line 252: nit: you can remove `not`, and simply do:

```
if inplace:
  ...
else:
 ...
```

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
```diff
@@ -203,3 +204,17 @@ def to_pandas(
...
     """
     return self._to_pandas(statement_params=statement_params, **kwargs)
+
+
+@register_series_accessor("cache_result")
+@_add_cache_result_docstring
+@snowpark_pandas_telemetry_method_decorator
+def cache_result(self, inplace: bool = False) -> pd.Series:
+    """
+    Persists the Snowpark pandas Series to a temporary table for the duration of the session.
```
sfc-gh-lspiegelberg, line 214: param inplace description? same comment with inplace/if

sfc-gh-lspiegelberg: just saw, you're using doc API. nvm, this docstring is fine.

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -125,35 +142,135 @@ def test_cache_result_dataframe_complex_correctness(
...
+    new_session = Session.builder.configs(db_parameters).create()
+    pd.session = new_session
+    df_data, kwargs = time_index_string_column_data
+    snow_df, _ = create_test_dfs(df_data, **kwargs)
+    snow_df = cache_and_return_df(snow_df, inplace)
+    table_name = (
+        snow_df._query_compiler._modin_frame.ordered_dataframe._dataframe_ref.snowpark_dataframe.table_name
+    )
+    new_session.close()
+    with pytest.raises(SnowparkSQLException, match="does not exist or not authorized"):
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: Makes sense - I had thought the same, but wasn't sure how else to respond to this comment: https://github.com/snowflakedb/snowpark-python/pull/1536#discussion_r1594328852

docs/source/modin/index.rst
sfc-gh-rdurrani: This is lowercase (as are the others) since its actually just letting sphinx know the name of the page. In the index, it is appropriately capitalized, since I believe it grabs the title of the page from the page itself. Adding a screenshot below for verification:
image

docs/source/modin/performance.rst
```diff
@@ -0,0 +1,50 @@
+Performance Recommendations
+===========================
+
+This page contains recommendations to help improve performance when using Snowpark pandas.
+
+Caching Intermediate Results
+----------------------------
+Snowpark pandas uses a lazy paradigm - when operations are called on a Snowpark pandas object,
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: That makes sense - my one concern is that this may be too detailed for users? Although I suppose they're probably a power user if they're looking into performance optimizations, so its probably ok.

sfc-gh-azhan (changes requested):
Let's do a rebase and update changelog first.

docs/source/modin/performance.rst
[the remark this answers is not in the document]
sfc-gh-lspiegelberg: I always liked as a user to get an explanation. If you think it's too detailed, we can always make it less prominent, or make it a note explaining it. Else, if you rather skip an explanation it's also ok to simply describe what the operation does. We can also compile a list of tips and tricks, e.g.:

```
1. Limit number of chained operations with caching
....
2. Select only columns you use
....
3. Avoid `to_pandas`, leave data server-side
....
4. Avoid df.iterrows(), use instead ...
...
```
That would look then more like a performance section!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: Good catch - old comment I forgot to delete, thank you!

docs/source/modin/series.rst
sfc-gh-rdurrani: Hmm - I'm not sure if this is wise. I imagine that we want to document the pandas APIs first to mimic the pandas docs as much as possible. cc: @sfc-gh-mvashishtha @sfc-gh-yzou

sfc-gh-azhan (commented):
changelog looks good to me.

docs/source/modin/performance.rst
```diff
@@ -0,0 +1,58 @@
...
+This page contains recommendations to help improve performance when using the Snowpark pandas API.
+
+Caching Intermediate Results
+----------------------------
+Snowpark pandas uses a lazy paradigm - when operations are called on a Snowpark pandas object,
+a lazy operator graph is built up and executed only when an output operation is called (e.g. printing
+the data, or persisting it to a table in Snowflake). This paradigm mirrors the Snowpark DataFrame paradigm,
+and enables larger queries to be optimized using Snowflake's SQL Query Optimizer. Certain workloads; however,
+can generate large operator graphs that include repeated, computationally expensive, subgraphs, that can introduce
+recomputations that impact performance. Take the following code snippet as an example:
```
sfc-gh-mvashishtha, line 13:
```suggestion
can generate large operator graphs that include repeated computationally expensive subgraphs. Take the following code snippet as an example:
```

nit: " that can introduce recomputations that impact performance" makes the sentence a run-on, and I think it's redundant.

src/snowflake/snowpark/modin/plugin/extensions/utils.py
```diff
@@ -0,0 +1,85 @@
...
+read from a table.
+
+>>> new_{object_var_name} = {object_var_name}.cache_result()
+
+>>> import numpy as np
+
+>>> np.all((new_{object_var_name} == {object_var_name}).values)
+True
+
+>>> from time import perf_counter
```
sfc-gh-mvashishtha, line 45: let's remove the timing since we shouldn't count on saving time with the `cache_result`?

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -0,0 +1,269 @@
...
+
+    This method is expensive because the concat generates unions, and the objects being unioned
+    may themselves contain unions, joins, or other expensive operations like pivots.
+    """
+    df = df.reset_index(drop=True)
+    return module.concat([df] * 10)
+
+
+@pytest.mark.parametrize("inplace", [True, False])
+def test_cache_result_empty_dataframe(inplace):
```
sfc-gh-mvashishtha, line 72: can we parametrize this test by the `DataFrame.__init__` kwargs dict, e.g. `{}` and `{'columns'=['A', 'B', 'C']}`? Right now it has a few nearly identical blocks. Let's also use `create_test_dfs` to make the dataframes.

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -0,0 +1,269 @@
...
+    native_df = native_df.set_index("b", drop=False)
+    with SqlCounter(query_count=1):
+        assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+            cached_snow_df, native_df, check_freq=False
+        )
+
+
+@pytest.mark.parametrize("inplace", [True, False])
+class TestCacheResultReducesQueryCount:
+    def test_cache_result_simple(self, inplace):
```
sfc-gh-mvashishtha, line 152: These test cases take about 49 seconds total on my laptop. Let's reduce the query size significantly, maybe by a factor of 10. We should still be able to observe the benefits of caching in query count.

tests/integ/modin/frame/test_cache_result.py
```diff
@@ -0,0 +1,269 @@
...
+            cached_snow_df = cache_and_return_df(snow_df, inplace)
+
+        with SqlCounter(query_count=1, union_count=9):
+            cached_snow_df = perform_chained_operations(cached_snow_df, pd)
+            cached_snow_df.to_pandas()
+
+        with SqlCounter(query_count=1):
+            assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+                cached_snow_df, native_df
+            )
```
sfc-gh-mvashishtha, line 176:
```suggestion
            assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
                cached_snow_df, native_df
            )
```

nit: this is clearer than the `cached_snow_df.to_pandas()` whose result isn't used anywhere.

tests/integ/modin/series/test_cache_result.py
```diff
@@ -0,0 +1,168 @@
...
+
+@pytest.fixture(scope="function")
+def simple_test_data():
+    int_data = list(range(9))
+    str_data = ["foo", "bar", "baz", "foo", "bar", "baz", "foo", "bar", "baz"]
+    return list(chain.from_iterable(zip(int_data, str_data)))
+
+
+@pytest.mark.parametrize("inplace", [True, False])
+def test_cache_result_empty_series(inplace):
```
sfc-gh-mvashishtha, line 58: Ditto from dataframe test. let's parametrize by the `Series.__init__` kwargs.

tests/integ/modin/series/test_cache_result.py
```diff
@@ -0,0 +1,168 @@
...
+@pytest.mark.parametrize("inplace", [True, False])
+class TestCacheResultReducesQueryCount:
+    def test_cache_result_simple(self, inplace):
+        snow_series = pd.concat(
+            [pd.Series(list(range(i, i + 5))) for i in range(0, 50, 5)]
+        )
+        native_series = perform_chained_operations(
+            native_pd.Series(np.arange(50)), native_pd
+        )
+        native_series.name = 0
```
sfc-gh-mvashishtha, line 119: why is this line here? We don't do the same to the Snowpark pandas series. Is it working around a Snowpark pandas bug? if so, can you link the bug here?

tests/integ/modin/series/test_cache_result.py
```diff
@@ -0,0 +1,168 @@
...
+            cached_snow_series = cache_and_return_series(snow_series, inplace)
+
+        with SqlCounter(query_count=1, union_count=9):
+            cached_snow_series = perform_chained_operations(cached_snow_series, pd)
+            cached_snow_series.to_pandas()
+
+        with SqlCounter(query_count=1):
+            assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+                cached_snow_series, native_series
+            )
```
sfc-gh-mvashishtha, line 139:
```suggestion
            assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
                cached_snow_series, native_series
            )
```

ditto about removing the line creating the unused `cached_snow_series.to_pandas()`

tests/integ/modin/series/test_cache_result.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def test_cache_result_post_apply(self, inplace, simple_test_data):
+        # In this test, the caching doesn't aid in the query counts since
+        # the implementation of apply(axis=1) itself contains intermediate
+        # result caching.
+        native_series = perform_chained_operations(
+            native_pd.Series(simple_test_data).apply(lambda x: x + x), native_pd
+        )
+        native_series.name = 0
```
sfc-gh-mvashishtha, line 148: ditto: why this line?

src/snowflake/snowpark/modin/plugin/extensions/utils.py
sfc-gh-rdurrani: Good catch, thank you - forgot to do so!

tests/integ/modin/series/test_cache_result.py
sfc-gh-rdurrani: Seems like this is a bug in snowpark pandas - filing a JIRA and linking it here.

docs/source/modin/performance.rst
```diff
@@ -9,8 +9,8 @@ Snowpark pandas uses a lazy paradigm - when operations are called on a Snowpark
 a lazy operator graph is built up and executed only when an output operation is called (e.g. printing
 the data, or persisting it to a table in Snowflake). This paradigm mirrors the Snowpark DataFrame paradigm,
 and enables larger queries to be optimized using Snowflake's SQL Query Optimizer. Certain workloads; however,
```
sfc-gh-mvashishtha, line 11:
```suggestion
and enables larger queries to be optimized using Snowflake's SQL Query Optimizer. Certain workloads, however,
```

sfc-gh-mvashishtha (approved):
LGTM except for a tiny grammar error! Thanks for addressing all my feedback.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -239,3 +240,17 @@ def to_pandas(
...
     """
     return self._to_pandas(statement_params=statement_params, **kwargs)
+
+
+@register_dataframe_accessor("cache_result")
+@_add_cache_result_docstring
+@snowpark_pandas_telemetry_method_decorator
+def cache_result(self, inplace: bool = False) -> Optional[pd.DataFrame]:
+    """
+    Persists the current Snowpark pandas DataFrame to a temporary table that lasts the duration of the session.
```
sfc-gh-azhan, line 250: can we add some doctests please?

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
```diff
@@ -203,3 +204,17 @@ def to_pandas(
...
     """
     return self._to_pandas(statement_params=statement_params, **kwargs)
+
+
+@register_series_accessor("cache_result")
+@_add_cache_result_docstring
+@snowpark_pandas_telemetry_method_decorator
+def cache_result(self, inplace: bool = False) -> Optional[pd.Series]:
+    """
+    Persists the Snowpark pandas Series to a temporary table for the duration of the session.
```
sfc-gh-azhan, line 214: can we add some doctests please?

src/snowflake/snowpark/modin/plugin/extensions/utils.py
```diff
@@ -0,0 +1,74 @@
...
+True
+
+>>> {object_var_name}.reset_index(drop=True, inplace=True) # Slower
+
+>>> new_{object_var_name}.reset_index(drop=True, inplace=True) # Faster
+
+"""
+
+
+def _add_cache_result_docstring(func: Fn) -> Fn:
```
sfc-gh-azhan, line 52:
```suggestion
def add_cache_result_docstring(func: Fn) -> Fn:
```
I thought if it only used inside util.py then we make it private.

docs/source/modin/performance.rst
```diff
@@ -0,0 +1,58 @@
...
+    # Session.builder.create() will create a default Snowflake connection.
+    Session.builder.create()
+    df = pd.concat([pd.DataFrame([range(i, i+5)]) for i in range(0, 150, 5)])
+    print(df)
+    df = df.reset_index(drop=True)
+    print(df)
+
+The above code snippet creates a 30x5 DataFrame using concatenation of 30 smaller 1x5 DataFrames,
+prints it, resets its index, and prints it again. The concatenation step can be expensive, and is
+lazily recomputed every time the dataframe is materialized - once per print. Instead, we recommend using
```
sfc-gh-azhan, line 31: We should also give user a tool to check how the complexity looks like? E.g., the number of api calls stored in the current df, then they can decide whether they should cache?

No need to block this PR but we should think about this. 
cc @sfc-gh-jkew @sfc-gh-helmeleegy

tests/integ/modin/frame/conftest.py
```diff
@@ -148,6 +150,15 @@ def time_index_native_df():
     )
 
 
+@pytest.fixture(scope="function")
+def time_index_string_column_data():
```
sfc-gh-azhan, line 154:
```suggestion
def date_index_string_column_data():
```

sfc-gh-azhan (commented):
fix:
```
-------------
src/snowflake/snowpark/modin/plugin/_internal/frame.py (50.0%): Missing lines 702
src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py (50.0%): Missing lines 1141
src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py (55.6%): Missing lines 252-254,256
src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py (55.6%): Missing lines 216-218,220
src/snowflake/snowpark/modin/plugin/extensions/utils.py (100%)
-------------
```

src/snowflake/snowpark/modin/plugin/extensions/utils.py
sfc-gh-rdurrani: Ah sounds good!

docs/source/modin/performance.rst
sfc-gh-rdurrani: Can open up a ticket to get to that after this: https://snowflakecomputing.atlassian.net/browse/SNOW-1445508

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
sfc-gh-rdurrani: Do you mean beyond the doctests that are added by `add_cache_result_docstring`?

sfc-gh-azhan (approved):

PR #1539, opened 2024-05-08
SNOW-1370365: Avoid UNION ALL for 1-column quantile operations

sfc-gh-joshi:

   

   Fixes SNOW-1370365

This PR avoids UNION ALL operations for computing quantiles over 1-column datasets. This optimization has significant implications for `pd.qcut`, which frequently computes a large number of quantiles and previously would had extremely high union counts in queries.
In particular, `test_qcut.py::test_qcut_two_columns` goes from 90 unions -> 0 unions, 34 joins -> 14 joins; and `series/test_quantile.py::test_quantile_large' goes from ~80 queries -> 6 queries.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
+++ b/src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
@@ -47,7 +47,7 @@
 - Given an input of type `Series`, `pd.qcut` always returns a `Series`.
 
 ### Improvements
-- Improved performance for `pd.qcut`, `Series.quantile`, and `Series.describe`.
+- Improved performance for `pd.qcut`, `Series.quantile`, `Series.describe`. Also improved `DataFrame.quantile` and `DataFrame.describe` for one-column `DataFrame`s.
 
 ### New Features
 - Added partial support for `SeriesGroupBy.apply` (where the `SeriesGrouBy` is obtained through `DataFrameGroupBy.__getitem__`).
--- a/src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
+++ b/src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
@@ -46,6 +46,9 @@
 - Changed the import path of Snowpark pandas package to use Modin 0.28.1 instead. The new recommended import statement is `import modin.pandas as pd; import snowflake.snowpark.modin.plugin`.
 - Given an input of type `Series`, `pd.qcut` always returns a `Series`.
 
+### Improvements
+- Improved performance for `pd.qcut`, `Series.quantile`, and `Series.describe`.
+
 ### New Features
 - Added partial support for `SeriesGroupBy.apply` (where the `SeriesGrouBy` is obtained through `DataFrameGroupBy.__getitem__`).
 
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -9300,7 +9300,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         q: list[float],
         interpolation: Literal["linear", "lower", "higher", "midpoint", "nearest"],
         index: Optional[Union[list[str], list[float]]] = None,
-        index_dtype: Optional[npt.DTypeLike] = None,
+        index_dtype: npt.DTypeLike = float,
     ) -> "SnowflakeQueryCompiler":
         """
         Helper method for ``qcut`` and ``quantile`` to compute quantiles over frames with a single column.
@@ -9355,7 +9355,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
             The labels for the resulting index column, allowing us to avoid a JOIN query by directly
             setting the correct column names before UNPIVOT. This is used primarily for ``describe``,
             where the resulting row labels are percentiles like "25%" rather than decimals like "0.25".
-        index_dtype : npt.DtypeLike, optional
+        index_dtype : npt.DtypeLike, default: float
             The type to which to coerce the resulting index column. Since UNPIVOT requires string column
             names, the resulting index column must be explicitly casted after the operation.
 
@@ -9382,59 +9382,60 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
             index = list(map(str, index))
         original_frame = self._modin_frame
         col_label = original_frame.data_column_pandas_labels[0]
-        col_ident = original_frame.data_column_snowflake_quoted_identifiers[0]
+        col_identifier = original_frame.data_column_snowflake_quoted_identifiers[0]
         new_labels = [str(quantile) for quantile in q]
-        new_idents = (
+        new_identifiers = (
             original_frame.ordered_dataframe.generate_snowflake_quoted_identifiers(
                 pandas_labels=new_labels
             )
         )
         ordered_dataframe = original_frame.ordered_dataframe.agg(
             *[
-                column_quantile(col(col_ident), interpolation, quantile).as_(new_ident)
-                for new_ident, quantile in zip(new_idents, q)
+                column_quantile(col(col_identifier), interpolation, quantile).as_(
+                    new_ident
+                )
+                for new_ident, quantile in zip(new_identifiers, q)
             ]
         )
-        index_ident = ordered_dataframe.generate_snowflake_quoted_identifiers(
+        index_identifier = ordered_dataframe.generate_snowflake_quoted_identifiers(
             pandas_labels=[None]
         )[0]
         # In order to set index labels without a JOIN, we call unpivot directly instead of using
         # transpose_single_row. This also lets us avoid JSON serialization/deserialization.
         ordered_dataframe = ordered_dataframe.unpivot(
             original_frame.data_column_snowflake_quoted_identifiers[0],
-            index_ident,
-            new_idents,
-            col_mapper=dict(zip(new_idents, index))
+            index_identifier,
+            new_identifiers,
+            col_mapper=dict(zip(new_identifiers, index))
             if index is not None
-            else dict(zip(new_idents, new_labels)),
+            else dict(zip(new_identifiers, new_labels)),
         ).ensure_row_position_column()
         internal_frame = InternalFrame.create(
             ordered_dataframe=ordered_dataframe,
             data_column_pandas_labels=[col_label],
             data_column_pandas_index_names=[None],
-            data_column_snowflake_quoted_identifiers=[col_ident],
+            data_column_snowflake_quoted_identifiers=[col_identifier],
             index_column_pandas_labels=[None],
-            index_column_snowflake_quoted_identifiers=[index_ident],
+            index_column_snowflake_quoted_identifiers=[index_identifier],
         )
-        if index_dtype:
-            # We cannot call astype() directly to convert an index column, so we replicate
-            # the logic here so we don't have to mess with set_index.
-            internal_frame = (
-                internal_frame.update_snowflake_quoted_identifiers_with_expressions(
-                    {
-                        index_ident: column_astype(
-                            index_ident,
-                            TypeMapper.to_pandas(
-                                internal_frame.quoted_identifier_to_snowflake_type()[
-                                    index_ident
-                                ]
-                            ),
-                            index_dtype,
-                            TypeMapper.to_snowflake(index_dtype),
-                        )
-                    }
-                )[0]
-            )
+        # We cannot call astype() directly to convert an index column, so we replicate
+        # the logic here so we don't have to mess with set_index.
+        internal_frame = (
+            internal_frame.update_snowflake_quoted_identifiers_with_expressions(
+                {
+                    index_identifier: column_astype(
+                        index_identifier,
+                        TypeMapper.to_pandas(
+                            internal_frame.quoted_identifier_to_snowflake_type()[
+                                index_identifier
+                            ]
+                        ),
+                        index_dtype,
+                        TypeMapper.to_snowflake(index_dtype),
+                    )
+                }
+            )[0]
+        )
 
         return SnowflakeQueryCompiler(internal_frame)
 
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -9154,6 +9154,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         ] = "linear",
         method: Literal["single", "table"] = "single",
         index: Optional[Union[list[str], list[float]]] = None,
+        index_dtype: Optional[npt.DTypeLike] = None,
     ) -> "SnowflakeQueryCompiler":
         """
         Returns values at the given quantiles for each column.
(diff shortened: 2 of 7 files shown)


sfc-gh-lspiegelberg:
> 1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.
>    
>    Fixes [SNOW-1370365](https://snowflakecomputing.atlassian.net/browse/SNOW-1370365)
> 2. Fill out the following pre-review checklist:
>    
>    * [ ]  I am adding a new automated test(s) to verify correctness of my new code
>    * [ ]  I am adding new logging messages
>    * [ ]  I am adding a new telemetry message
>    * [ ]  I am adding new credentials
>    * [ ]  I am adding a new dependency
> 3. Please describe how your code solves the related issue.
> 
> This PR avoids UNION ALL operations for computing quantiles over 1-column datasets. This optimization has significant implications for `pd.qcut`, which frequently computes a large number of quantiles and previously would had extremely high union counts in queries.

This is great! Out of curiousity, do you have some numbers that describe the delta? I.e., rough idea of Unions reduced / runtime improved? You prob. can use https://github.com/snowflakedb/snowpark-python/blob/7702803548f934fe4786f66b74cf4770d3c32b44/tests/integ/modin/test_qcut.py#L272 for that. Would even consider adding to the PR description, given it's a big win!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9176,6 +9177,10 @@ def quantiles_along_axis0(
             When specified, sets the index column of the result to be this list. This is not part of
             the pandas API for quantile, and only used to implement df.describe().
             When unspecified, the index is the float values of the quantiles.
+        index_dtype: Optional[npt.DTypeLike], default None
+            When specified along with index, determines the type of the index column. This is only used
```
sfc-gh-lspiegelberg, line 9181: do we support multi-index? Is there a check to ensure this is only called for single index column case? If not, prob. stating this only works for non-multi-index and having an assert (len(...) == 1) over the ordered_dataframe -> index_snowflake_quoted_identifiers or how it is called should work.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9201,6 +9206,15 @@ def quantiles_along_axis0(
             ErrorMessage.not_implemented(
                 "quantile is not supported for datetime columns"
             )
+        assert index is None or len(index) == len(
+            q
+        ), f"length of index {index} did not match quantiles {q}"
+        if len(query_compiler.columns) == 1:
```
sfc-gh-lspiegelberg, line 9212: nit: better to use an alternative to `.columns`, I remember it triggered sometimes queries. You can use instead `InternalFrame.data_column_snowflake_quoted_identifiers` or `query_compiler.get_axis_len(axis)`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9201,6 +9206,15 @@ def quantiles_along_axis0(
...
+            # Use method without UNION ALL operations if the query compiler has only a single column
+            index_values = q if index is None else index
```
sfc-gh-lspiegelberg, line 9214: can we add a comment to explain the if expression here?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9201,6 +9206,15 @@ def quantiles_along_axis0(
...
+            return self._quantiles_single_col(
+                q, interpolation, index=index_values, index_dtype=index_dtype or float
```
sfc-gh-lspiegelberg, line 9216: maybe also add a comment why you default the index_dtype to float.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9264,7 +9275,7 @@ def quantiles_along_axis0(
             OrderingColumn(global_ordering_identifier),
             *ordered_dataframe.ordering_columns,
         )
-        return SnowflakeQueryCompiler(
```
sfc-gh-lspiegelberg, line 9267: nit: can prob. return directly

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9276,6 +9287,114 @@ def quantiles_along_axis0(
...
+        q: list[float],
+        interpolation: Literal["linear", "lower", "higher", "midpoint", "nearest"],
+        index: Optional[Union[list[str], list[float]]] = None,
+        index_dtype: Optional[npt.DTypeLike] = None,
+    ) -> "SnowflakeQueryCompiler":
+        """
+        Helper method for qcut() + quantile() to compute quantiles over frames with a single column.
+
+        Normally, we compute single row for every given quantile, with each column corresponding to
+        a column to the input frame.
```
sfc-gh-lspiegelberg, line 9303: really love the description here, great job!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9276,6 +9287,114 @@ def quantiles_along_axis0(
...
+        ordered_dataframe = original_frame.ordered_dataframe.agg(
+            *[
+                column_quantile(col(col_ident), interpolation, quantile).as_(new_ident)
+                for new_ident, quantile in zip(new_idents, q)
+            ]
+        )
+        index_ident = ordered_dataframe.generate_snowflake_quoted_identifiers(
+            pandas_labels=[None]
+        )[0]
+        # In order to preserve index labels, we call unpivot directly instead of using transpose_single_row
```
sfc-gh-lspiegelberg, line 9360: is this a bug in transpose_single_row? Should we fix it?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9276,6 +9287,114 @@ def quantiles_along_axis0(
...
+        ).ensure_row_position_column()
+        internal_frame = InternalFrame.create(
+            ordered_dataframe=ordered_dataframe,
+            data_column_pandas_labels=[col_label],
+            data_column_pandas_index_names=[None],
+            data_column_snowflake_quoted_identifiers=[col_ident],
+            index_column_pandas_labels=[None],
+            index_column_snowflake_quoted_identifiers=[index_ident],
+        )
+        if index_dtype:
```
sfc-gh-lspiegelberg, line 9377: you seem to call this function never with `index_dtype=None` but always with a valid index_dtype (the None case seems to become `index_dtype=index_dtype or float` above in `quantiles_along_axis0`. Is there a scenario where `index_dtype=None` is called? If not, can we remove the if statement, and adjust the index_dtype type hint?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9276,6 +9287,114 @@ def quantiles_along_axis0(
...
+            # We cannot call astype() directly to convert an index column, so we replicate
+            # the logic here so we don't have to mess with set_index
+            internal_frame = (
+                internal_frame.update_snowflake_quoted_identifiers_with_expressions(
+                    {
+                        index_ident: column_astype(
+                            index_ident,
```
sfc-gh-lspiegelberg, line 9384: nit: index_ident -> index_identifier?

tests/integ/modin/frame/test_quantile.py
```diff
@@ -111,3 +129,18 @@ def test_quantile_datetime_negative():
     snow_df = pd.DataFrame(TEST_QUANTILE_DATA)
     with pytest.raises(NotImplementedError):
         snow_df.quantile(numeric_only=False)
+
+
+@pytest.mark.xfail(
```
sfc-gh-lspiegelberg, line 134: do you have some more context on this test? Is the problem here the `to_pandas()` emitting a lot of queries?

tests/integ/modin/test_qcut.py
```diff
@@ -264,15 +264,13 @@ def test_qcut_invalid_quantiles_negative(q):
 
 @sql_count_checker(
     query_count=5,
-    join_count=34,
-    union_count=90,
```
sfc-gh-lspiegelberg, line 268: wow, that's a query count reduction!

tests/integ/modin/series/test_quantile.py
```diff
@@ -11,7 +11,7 @@
```
sfc-gh-lspiegelberg, line 11: I remember there was a test somewhere either in qcut/quantile where q was large and it was commented because of too many UNIONs. Do you think we can find & reactivate it? I would be curious to see if we can now with this change also support scenarios like `q=100` in qcut, or what a limit for `q` is we should set.

sfc-gh-lspiegelberg (commented):
Looks great! Left a few comments.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: What I mean by this is that there's nowhere in `transpose_single_row` that would easily let us set the labels of the resulting index (the names of the columns before UNPIVOT) to be our desired values. This matters in cases like `df.describe`, where the quantiles have labels like "25%" and "50%" rather than the usual decimal "0.25", "0.5", etc. We would have to do a JOIN later to add a column with the correct index. Calling unpivot directly also lets us avoid the JSON serialization/deserialization.

I may end up using `transpose_single_row` anyway, since I'm having difficulty getting `UNPIVOT` to preserve the order of the passed in quantiles in the result column. This would force us to do a JOIN anyway to ensure the index is respected.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: We don't have any tests for multi-index quantile, but this parameter shouldn't be affected by whether the input frame is multi-indexed or not. The query compiler that results from a quantile computation should always have only a single index column, where the index is the computed quantiles (`[0.25, 0.5]`, etc.).

For what it's worth, we do test `describe` with multi-index series, and this updated quantile operation works without issue there.

tests/integ/modin/series/test_quantile.py
sfc-gh-joshi: I believe that's `test_quantile_large`, which I've un-XFAILed. I added a separate `test_quantile_large` for a frame with multiple columns in `frame/test_describe.py` which remains XFAILed since it doesn't benefit from this PR.

We could probably increase the limit for q as you suggested, but we'd have to have a separate limit for single-column frames vs. multi-column ones.

tests/integ/modin/frame/test_quantile.py
sfc-gh-joshi: This is a modified version of `series/test_quantile.py::test_quantile_large`, which this PR addresses by removing UNION ALLs. That optimization is only applicable in the single-column case, so performing a quantile() operation with many qs on a multi-column frame will run into the same high query counts as before.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: added (also moved further up)

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: I've changed index_dtype to default to `float` now, and removed this `if index_dtype` guard.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: +100

tests/integ/modin/frame/test_quantile.py
```diff
@@ -36,16 +37,34 @@
...
             native_df,
             lambda df: df.quantile(q, numeric_only=True),
         )
 
 
+@pytest.mark.parametrize("q", TEST_QUANTILES)
+def test_quantile_single_column(q):
+    # When the input frame has only a single column, and q is sorted, no UNION ALLs are performed
+    expected_union_count = len(q) - 1 if sorted(q) != q else 0
+    snow_df, native_df = create_test_dfs({"a": [1, 2, 3, 4]})
```
sfc-gh-lspiegelberg, line 54: nit: you can parametrize over data to save some similar code here, e.g.

```
@pytest.mark.parametrize("data", [{"a": [1, 2, 3, 4]}, {"a": [1, 2, 3, 4], "b": ["the", "quick", "brown", "fox"]}])
```

tests/integ/modin/frame/test_quantile.py
sfc-gh-lspiegelberg: ok, thanks for explaining. Do you have an idea whether we could do some similar to approach to get rid off the high union count there as well?

tests/integ/modin/series/test_quantile.py
sfc-gh-lspiegelberg: perfect!

sfc-gh-lspiegelberg (approved):
Great job!

src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
```diff
@@ -46,6 +46,9 @@
 - Changed the import path of Snowpark pandas package to use Modin 0.28.1 instead. The new recommended import statement is `import modin.pandas as pd; import snowflake.snowpark.modin.plugin`.
 - Given an input of type `Series`, `pd.qcut` always returns a `Series`.
 
+### Improvements
+- Improved performance for `pd.qcut`, `Series.quantile`, and `Series.describe`.
```
sfc-gh-helmeleegy, line 50: Let's mention single-column dataframes here as well.

sfc-gh-helmeleegy (approved):
This is really nice. Thanks for the performance boost!

tests/integ/modin/frame/test_quantile.py
sfc-gh-joshi: The operation is different between the two (the 2-column frame uses `numeric_only=True`), and I don't think it's worth doing a double parametrization for just 2 function calls.

tests/integ/modin/frame/test_quantile.py
sfc-gh-joshi: Possibly, but the simplest thing I can think of would be just trading off UNIONs for JOINs -- we could use this UNPIVOT method to compute a quantile column for every column in the input frame, then JOIN all those columns together.

This may be worth it if len(q) >> len(columns), though it might take some trial and error for us to figure out an appropriate threshold.

src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
sfc-gh-joshi: added

PR #1545, opened 2024-05-09
SNOW-1371757: Fix repr when max_cols or max_rows is None

sfc-gh-dpetersohn:

   Fixes SNOW-1371757

   Catch `None` values at the API layer and sanitize in the same way Modin OSS does: 
   
   https://github.com/modin-project/modin/blob/06699a892c2a020c92ecf0c54d1aa35bbcd5660b/modin/pandas/dataframe.py#L268-L269


sfc-gh-dpetersohn:

src/snowflake/snowpark/modin/pandas/dataframe.py
```diff
@@ -254,9 +254,9 @@ def __repr__(self):
         str
         """
         # TODO: SNOW-1063346: Modin upgrade - modin.pandas.DataFrame functions
-        num_rows = pandas.get_option("display.max_rows") or 10
+        num_rows = pandas.get_option("display.max_rows") or len(self.index)
```
sfc-gh-nkrishna, line 257: `len(self.index)` can be pretty expensive since it still pulls the index (at least until we have a cache for the length), I think keeping the hardcoded "or 10" here and differing from pandas in this case is perhaps a better call

sfc-gh-dpetersohn: I did consider what you suggested, but the reason I chose to go with `len(self.index)` is because:
- User must opt-in to this behavior by setting `display.max_rows = None`
- Users have no option outside of converting the entire dataset to pandas in order to view all of their rows
- We kind of end up treating users as though they are not capable of making the decision themselves to pull out all of the rows.

I think a better way forward would be to throw a warning when `display.max_rows` is set to `None` to let the user know that would be pulling out all of the data. Then at least they are informed when they set it.

We should also not require `len(self.index)` to pull the index out of the database. That needs to be fixed separately.

sfc-gh-nkrishna: That makes sense to me -- going to approve the PR, we can add the warning in this PR or in a follow-up good with either option.

sfc-gh-nkrishna (approved):

src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Fixed UDTF "int64 is not serializable" errors from new Snowflake release in `apply(axis=1)`
 - Fixed binary operations with single-row DataFrame and Series.
 - Fixed incorrect return type in `qcut` when given `Series` input and improved error checking logic.
+- Fixed support for `display.max_rows` and `display.max_cols` being set to `None`.
```
sfc-gh-azhan, line 13: can you rebase? This file is OFD.

src/snowflake/snowpark/modin/pandas/dataframe.py
sfc-gh-azhan: Can we just use `len(self)`? It will make a count query which is better than pulling the data locally. At least we don't need to worry about OOM if data is huge.

sfc-gh-azhan (changes requested):

sfc-gh-helmeleegy (approved):
LGTM.

src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
sfc-gh-azhan: can you move under 1.16.0a1? 1.15.0a1 has been released.

sfc-gh-dpetersohn: done!

sfc-gh-azhan (approved):

PR #1550, opened 2024-05-09
SNOW-1320296 - pd.concat SQL Compilation ambigious __row_position__ issue

sfc-gh-jkew:
SNOW-1320296 - pd.concat SQL Compilation ambigious __row_position__ issue

This fix addresses issues with concat seen originally when loading from a local file (although it could occur in other ways as well). It exhibits with both df.describe(include='O') and df.concat. There were two issues, one where the _row_position_ column was ambiguous, and included twice in the generated frame, when data is loaded from a file we can have an ordering column which can overlap with the index, so we de-duplicate these identifiers before the union.

The second issue was occurs when a data frame loaded from a file is concatenated with a generated data frame. The index column consolidation was not triggered because in one frame we have an index set as '__index__' and in the other it was set as '__row_position__' . Instead of just looking at length differences between frames we also look at index name differences between frames to determine if the index needs to be dropped ( assuming ignore_index is 'True')


sfc-gh-mvashishtha (approved):
thanks for the fix. LGTM

sfc-gh-lmukhopadhyay (approved):
LGTM! Just wondering, is there a good way to test one of the use cases you mentioned above: "data frame loaded from a file is concatenated with a generated data frame"?

sfc-gh-jdu (approved):

PR #1563, opened 2024-05-10
SNOW-1359484 Allow write Snowpark pandas DataFrame and Series in sess…

sfc-gh-azhan:
…ion.write_pandas

   Fixes SNOW-1359484

Allow user to use `session.write_pandas` to write Snowpark pandas objects:

- It reuse and does not change the interface of `write_pandas`
- it reuse `to_snowflake` to implement the write logic
- copied main tests from `test_pandas_to_df.py` and use them to test writing Snowpark pandas DataFrame.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
+++ b/src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
@@ -8,6 +8,9 @@
 ### Improvements
 - Improved performance of `pd.qcut` by removing joins in generated sql query.
 
+### Improvements
+- Allow `session.write_pandas` to write Snowpark pandas DataFrame or Series to a table.
+
 ## 1.15.0a1 (2024-07-05)
 
 ### Bug Fixes
--- a/src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
+++ b/src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
@@ -7,8 +7,6 @@
 
 ### Improvements
 - Improved performance of `pd.qcut` by removing joins in generated sql query.
-
-### Improvements
 - Allow `session.write_pandas` to write Snowpark pandas DataFrame or Series to a table.
 
 ## 1.15.0a1 (2024-07-05)
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -2172,7 +2172,8 @@ class Session:
 
         if not auto_create_table and not self._table_exists(name):
             raise SnowparkClientException(
-                f"Cannot write Snowpark pandas DataFrame to table {location} because it does not exist. Use auto_create_table = True to create table before writing a Snowpark pandas DataFrame"
+                f"Cannot write Snowpark pandas DataFrame or Series to table {location} because it does not exist. Use "
+                f"auto_create_table = True to create table before writing a Snowpark pandas DataFrame or Series"
             )
         if_exists = "replace" if overwrite else "append"
         df.to_snowflake(
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -2277,10 +2277,11 @@ class Session:
             your pandas DataFrame cannot be written to the specified table, an
             exception will be raised.
 
-            If the dataframe is :class:`~snowflake.snowpark.modin.pandas.DataFrame` or :class:`~snowflake.snowpark.modin.pandas.Series`,
-            it will call :func:`snowflake.snowpark.modin.pandas.DataFrame.to_snowflake <snowflake.snowpark.modin.pandas.DataFrame.to_snowflake>` or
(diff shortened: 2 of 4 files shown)


src/snowflake/snowpark/modin/plugin/PANDAS_CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 ### Improvements
 - Improved performance of `pd.qcut` by removing joins in generated sql query.
 
+### Improvements
```
sfc-gh-helmeleegy, line 11: Can you merge with existing Improvements section?

src/snowflake/snowpark/session.py
```diff
@@ -2112,9 +2112,84 @@ def get_session_stage(
...
+
+        name = [table_name]
+        if schema:
+            name = [quote_id(schema)] + name
+        if database:
+            name = [quote_id(database)] + name
+
+        if not auto_create_table and not self._table_exists(name):
+            raise SnowparkClientException(
+                f"Cannot write Snowpark pandas DataFrame to table {location} because it does not exist. Use auto_create_table = True to create table before writing a Snowpark pandas DataFrame"
```
sfc-gh-helmeleegy, line 2175: Can we have the error message mention DataFrame or Series depending on `df`'s type?

tests/integ/modin/test_session_write_pandas.py
sfc-gh-helmeleegy, line 1: Can we have one test for Series?

sfc-gh-helmeleegy (commented):
Looks great overall. Just a few minor nits.

src/snowflake/snowpark/session.py
sfc-gh-helmeleegy: I see you already updated it to mention Series, which is what I meant. I was thinking however, it might be nice to just mention either DataFrame or Series - because we should already have this information. It's just a nice to have though. So I'll leave it to you to decide.

sfc-gh-helmeleegy (approved):
LGTM. Thanks for taking care of this!

src/snowflake/snowpark/session.py
```diff
@@ -2241,22 +2323,37 @@ def write_pandas(
...
-                compression=compression,
-                on_error=on_error,
-                parallel=parallel,
-                quote_identifiers=quote_identifiers,
-                auto_create_table=auto_create_table,
-                overwrite=overwrite,
-                table_type=table_type,
-                **kwargs,
-            )
+            if "modin" in str(type(df)):
```
sfc-gh-jrose, line 2326: I don't think the intention of this line is very clear. Something like this would be more clear:
```suggestion
            if isinstance(df, (snowflake.snowpark.modin.pandas.DataFrame, snowflake.snowpark.modin.pandas.Series):
```

sfc-gh-jrose (commented):
It would be nice to have a test that validates the naming conventions as well to make sure the quote_id code paths are working correctly.

src/snowflake/snowpark/session.py
sfc-gh-azhan: This won't work because modin is not always imported.

sfc-gh-azhan:
> It would be nice to have a test that validates the naming conventions as well to make sure the quote_id code paths are working correctly.
> 

I have integ tests to cover the case where quote_id is true or false, and the code is basically the same as the existing code: https://github.com/snowflakedb/snowpark-python/blob/cc3d2898d4170cc68ca8d59a7872e9c2b1ba7929/src/snowflake/snowpark/session.py#L2219

> It's also surprising to me that new functionality would not require a CHANGELOG update.
The change log is added in pandas_changelog.md. In the next release, we will merge them together so this is not in changelog.md.

src/snowflake/snowpark/session.py
```diff
@@ -2112,9 +2112,85 @@ def get_session_stage(
...
+            table_type: The table type of table to be created. The supported values are: ``temp``, ``temporary``,
+                and ``transient``. An empty string means to create a permanent table. Learn more about table types
+                `here <https://docs.snowflake.com/en/user-guide/tables-temp-transient.html>`_.
+        """
+
+        def quote_id(id: str) -> str:
+            return '"' + id + '"' if quote_identifiers else id
+
+        name = [table_name]
+        if schema:
```
sfc-gh-jdu, line 2168: If quote_identifiers is true, actually the table and column names will all be quoted. Do we have a test for that?

src/snowflake/snowpark/session.py
sfc-gh-azhan: rewrite this part now PTAL

src/snowflake/snowpark/session.py
sfc-gh-azhan: good point! I added one test specific for this. Also notice that `to_snowflake` always quote columns names so we can raise not implemented error for modin pandas is quote_identifier is false.

sfc-gh-jrose (approved):

PR #1577, opened 2024-05-14
SNOW-1321662: Update merge behavior to pandas 2.x

sfc-gh-nkumar:

   

   Fixes SNOW-1321662

 Update merge behavior to match with pandas 2.x
 change 1: merge/join with how='outer' always performs sorting. 
 change 2: for merge/join between frames with multi index, order of index columns is now:
   right index columns + remaining left index columns (if how != right)
   left index columns + remaining right index columns otherwise
   
Native pandas has fixed lot of ordering related bugs in 2.x https://github.com/pandas-dev/pandas/pull/54611/files
So also removed custom ordering logic from tests which is no longer required.  

Also updated tests and removed xfails.


tests/integ/modin/frame/test_merge.py
```diff
@@ -92,9 +92,8 @@ def _merge_native_pandas_frames_on_index_on_both_sides(
     is_left_mi = left.index.nlevels > 1
     is_right_mi = right.index.nlevels > 1
     if sort:
-        # When joining single index frame with multi-index or multi-index frame with
-        # single index, native pandas doesn't respect 'sort' argument. It always
-        # behaves as sort=False.
+        # When multi-index is involved on either (or both) side of frames, native pandas
```
sfc-gh-mvashishtha, line 95: The comment says "or both", but the condition doesn't seem to include the both case: `(is_left_mi and not is_right_mi) or (not is_left_mi and is_right_mi)`

tests/integ/modin/frame/test_merge.py
```diff
@@ -226,27 +198,16 @@ def _verify_merge(
...
-    keys, and order among keys is inherited from the left dataframe, and then the right
-    data frame.
-
-    In addition to above there are bugs in Native pandas where even left and right join
-    do not preserve order from left frame or right frame respectively.
+    To avoid comparison failure due to bugs in Native pandas we perform some custom
+    operation after merge.
+    Some example bugs:
+    left and right join do not preserve order from left frame or right frame respectively.
     https://github.com/pandas-dev/pandas/issues/40608
```
sfc-gh-mvashishtha, line 205: but this has bug has been fixed now?

tests/integ/modin/frame/test_merge.py
```diff
@@ -483,7 +428,10 @@ def test_merge_on_index_multiindex_common_labels(left_df, right_df, how, sort):
     )
 
 
-@pytest.mark.xfail(reason="SNOW-1321662 - pandas 2.2.1 update failure", strict=True)
+@pytest.mark.xfail(
+    reason="pandas bug: AssertionError: Length of order must be same as number of levels (2), got 1",
```
sfc-gh-mvashishtha, line 432: could you add a link to the pandas bug here?

tests/integ/modin/frame/test_merge.py
```diff
@@ -513,18 +461,23 @@ def test_merge_on_index_multiindex_equal_labels(left_df, right_df, how, sort):
...
     right_df = right_df.rename(columns={"A": "left_i"}).set_index(
         "left_i", append=True
     )  # index columns ['right_i', 'left_i']
     if how in ("inner", "right"):
-        with SqlCounter(query_count=3, join_count=1):
-            _verify_merge(
-                left_df, right_df, how=how, left_index=True, right_index=True, sort=sort
-            )
+        if how == "inner" and sort is False:
+            # Bug in pandas. it retains right order instead of left order.
```
sfc-gh-mvashishtha, line 469: could you link to the pandas bug here?

tests/integ/modin/frame/test_merge.py
```diff
@@ -513,18 +461,23 @@ def test_merge_on_index_multiindex_equal_labels(left_df, right_df, how, sort):
...
+            pass
```
sfc-gh-mvashishtha, line 470: Let's `pytest.skip` and link to the pandas bug in the `reason`?

sfc-gh-mvashishtha (commented):
leaving some questions.

tests/integ/modin/frame/test_merge.py
sfc-gh-nkumar: It's the second if block below `is_left_mi and is_right_mi` based on if both sides have multi index or not sorting steps are different.

tests/integ/modin/frame/test_merge.py
sfc-gh-nkumar: Good catch, I should remove reference to this bug. In 2.x quite a few sorting related bugs are fixed so we no longer need custom sorting logic in this method.

tests/integ/modin/frame/test_merge.py
sfc-gh-nkumar: I was looking for it but couldn't find this in open issues.

tests/integ/modin/frame/test_merge.py
sfc-gh-nkumar: Didn't find this bug reported in issues. I think mult-index dataframes are not really that popular so lot of bugs around these are not filed.

tests/integ/modin/frame/test_merge.py
sfc-gh-nkumar: Added pytest.skip.

tests/integ/modin/frame/test_merge.py
sfc-gh-mvashishtha: Could you file a bug report in pandas and link it here? Sometimes someone else will fix the bug, and we also have a self-contained bug report to refer to when we look at this test or behavior again.

tests/integ/modin/frame/test_merge.py
sfc-gh-mvashishtha: ditto about pandas bug

tests/integ/modin/frame/test_merge.py
sfc-gh-nkumar: Done

sfc-gh-mvashishtha (approved):
LGTM

tests/integ/modin/frame/test_merge.py
```diff
@@ -513,18 +458,22 @@ def test_merge_on_index_multiindex_equal_labels(left_df, right_df, how, sort):
...
     right_df = right_df.rename(columns={"A": "left_i"}).set_index(
         "left_i", append=True
     )  # index columns ['right_i', 'left_i']
     if how in ("inner", "right"):
-        with SqlCounter(query_count=3, join_count=1):
-            _verify_merge(
-                left_df, right_df, how=how, left_index=True, right_index=True, sort=sort
-            )
+        if how == "inner" and sort is False:
+            pytest.skip("pandas bug: https://github.com/pandas-dev/pandas/issues/55774")
```
sfc-gh-mvashishtha, line 466: this bug has been fixed in pandas 2.2.1. why is it a problem for us?

sfc-gh-joshi (approved):
LGTM besides @sfc-gh-mvashishtha's comments about native pandas bugs.

tests/integ/modin/frame/test_merge.py
sfc-gh-nkumar: Looks like it's not fixed completely. ordering is incorrect for single to mulit join case.

sfc-gh-mvashishtha: Could you please comment on the pandas issue with a reproducer?

sfc-gh-nkumar: TBH, I don't think we should spend much energy on fixing or worrying about pandas bugs. Pandas merge implementation (especially when multi-index is involved) is super buggy. We should strive to provide a consistent/documented behavior to snowpark pandas users not worry about pandas bugs.

sfc-gh-mvashishtha: I still recommend that we report pandas bugs when we find them.

1) it's a service to the rest of the open source community. In this case it looks like https://github.com/pandas-dev/pandas/issues/55774 will be closed and there's no record of the fact that the issue hasn't really been fixed. We don't know how many users will also find this bug and not find an existing issue for it. If they want a fix, they have to report the issue and then wait for a fix. This project and Snowflake in general benefit a lot from pandas, and filing bugs is a simple way to give back to the pandas community.
1) we use pandas methods in our implementation, so if someone fixes a pandas issue, that makes our implementation simpler or less buggy.
1) we use pandas a lot in our tests, so if someone fixes a pandas issue, our tests can be simpler to maintain, understand, and extend. Also, having a self-contained bug report makes it easier to understand what pandas behaviors we would like to rely on but cannot, and to learn when a bug has been fixed.

PR #1472, opened 2024-05-01
SNOW-1341619: Always mention the specific missing feature in ErrorMessage.not_implemented.

sfc-gh-mvashishtha:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-1341619

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Use decorators that share most of their implementation to replace "This functionality is not yet available in Snowpark pandas API" with more informative errors. Remove the ability to call `not_implemented()` without a custom `message`.

   Eventually we will expand these decorators so they can give useful messages for missing support for particular parameter values as well.

PR #1580, opened 2024-05-14
SNOW-1264688: Remove overrides of infer_object/nunique

sfc-gh-joshi:

   

   Fixes SNOW-1264688

`infer_objects` and `nunique` were overridden as vestigial remnants of the SnowparkPandasDataFrame and SnowparkPandasSeries classes. Now that we can use the docs extension modules, there is no longer any reason to keep these overrides around.

After this PR, only 2 remaining DF/Series methods are overridden:
- `memory_usage` can also be removed pretty easily; since it has a separate ticket filed I'll remove it in a separate PR.
- `Series.isin` does some extra argument processing that I'd prefer not to touch in case something breaks.

### DIFF
--- a/src/snowflake/snowpark/modin/pandas/base.py
+++ b/src/snowflake/snowpark/modin/pandas/base.py
@@ -1986,6 +1986,7 @@ class BasePandasDataset(metaclass=TelemetryMeta):
             )
         )
 
+    @base_not_implemented()
     def infer_objects(
         self, copy: bool | None = None
     ) -> BasePandasDataset:  # pragma: no cover # noqa: RT01, D200
@@ -1993,8 +1994,6 @@ class BasePandasDataset(metaclass=TelemetryMeta):
         Attempt to infer better dtypes for object columns.
         """
         # TODO: SNOW-1119855: Modin upgrade - modin.pandas.base.BasePandasDataset
-        # This method is currently overriden in dataframe_overrides.py and series_overrides.py
-        # and raises NotImplementedError
         new_query_compiler = self._query_compiler.infer_objects()
         return self._create_or_update_from_compiler(
             new_query_compiler, inplace=False if copy is None else not copy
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -1986,15 +1986,14 @@ def idxmin(self, axis=0, skipna=True, numeric_only=False):  # noqa: PR01, RT01,
             )
         )
 
+    @base_not_implemented()
```
sfc-gh-yzou, line 1989: what is this annotation here ?

src/snowflake/snowpark/modin/plugin/docstrings/base.py
```diff
@@ -1317,7 +1317,8 @@ def idxmin():
 
     def infer_objects():
         """
-        Attempt to infer better dtypes for object columns.
+        Attempt to infer better dtypes for object columns. This is not currently supported
```
sfc-gh-yzou, line 1320: I am little bit confused, there seems no not implemented error raised in base.py, which means it is supported, right?

sfc-gh-joshi: Before this PR, our overrides had `@dataframe_not_implemented` and `@series_not_implemented`, which automatically raise NotImplementedError.

src/snowflake/snowpark/modin/pandas/base.py
sfc-gh-joshi: This was added in https://github.com/snowflakedb/snowpark-python/pull/1472 to automatically raise NotImplementedError with the appropriate API name.

sfc-gh-yzou: I see.

sfc-gh-yzou (approved):

sfc-gh-helmeleegy (approved):
LGTM, thanks Jonathan!

PR #1584, opened 2024-05-14
SNOW-1372258 - Print warning when calling __array__()

sfc-gh-jkew:
SNOW-1372258 - Print warning when calling __array__()

3rd party libraries call __array__() frequently to get a numpy array for local processing. We do not warn users of this so certain types of pandas code (iterating through columns to print) may take a very long time. Here we make calls to __array__ print a single warning, and we point users at the to_pandas or to_numpy API calls.

### DIFF
--- a/src/snowflake/snowpark/modin/pandas/base.py
+++ b/src/snowflake/snowpark/modin/pandas/base.py
@@ -3781,6 +3781,12 @@ class BasePandasDataset(metaclass=TelemetryMeta):
             NumPy representation of Modin object.
         """
         # TODO: SNOW-1119855: Modin upgrade - modin.pandas.base.BasePandasDataset
+        WarningMessage.single_warning(
+            "Calling __array__ on a modin object materializes all data into local memory.\n"
+            + "Since this can be called by 3rd party libraries silently, it can lead to \n"
+            + "unexpected delays or high memory usage. Use to_pandas() or to_numpy() to do \n"
+            + "this once explicitly.",
+        )
         arr = self.to_numpy(dtype)
         return arr
 


sfc-gh-oplaton (approved):

sfc-gh-azhan (approved):

src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -3781,6 +3781,12 @@ def __array__(self, dtype=None):
             NumPy representation of Modin object.
         """
         # TODO: SNOW-1119855: Modin upgrade - modin.pandas.base.BasePandasDataset
+        WarningMessage.single_warning(
+            "Calling __array__ on a modin object materializes all data into local memory.\n"
+            + "Since this can be called by 3rd party libraries silently, it can lead to \n"
+            + "unexpected delays or high memory usage. Use to_pandas() or to_numpy() to do \n"
```
sfc-gh-helmeleegy, line 3787: I don't feel the call to action is very clear here. First, the user is not the one calling `__array__`. So it's not like the user needs to replace each call to `__array__` with a call to `to_pandas()` or `to_numpy()`. Also, the fact that we are gaining two options (`to_pandas()` and `to_numpy()`) could be confusing. My understanding is that 3rd party libraries are expecting a pandas data frame, and then while processing it, they may call `__array__` a few times. So asking the user to use `to_numpy()` wouldn't help because the 3rd party library won't have access to the resulting numpy array. Instead, I think we can replace the last sentence with something like:
```
If this is something you are experiencing, you can consider calling to_pandas() once explicitly before passing the modin object to the 3rd party library.
```

sfc-gh-jkew: It's actually hard to recommend any sort of action here as it really depends on what their code is and how long they want to do client-side work. If anything it might be wise to make the whole thing shorter, to something like: `A call to __array__ was made, materializing all data to the client` and let the user figure it out.

Not sure it's super worth it though. I mainly wanted *any* sort of warning when this happens.

PR #1595, opened 2024-05-15
SNOW-1429873 Fix pandas groupby apply int not serializable issue

sfc-gh-azhan:

   

   Fixes SNOW-1429873 Fix pandas groupby apply int not serializable issue


tests/integ/modin/groupby/test_groupby_apply.py
```diff
@@ -857,6 +858,28 @@ def operation(df):
...
+        udtf_count=UDTF_COUNT,
+        join_count=JOIN_COUNT,
+    )
+    def test_group_apply_return_df_from_lambda(self):
+        diamonds_path = (
+            pathlib.Path(__file__).parent.parent.parent.parent
+            / "resources"
+            / "diamonds.csv"
+        )
+        diamonds_pd = native_pd.read_csv(diamonds_path)
```
sfc-gh-mvashishtha, line 873: can we use a smaller dataframe created inline that reproduces the error? I think that would be easier to maintain. Does the data have to come from a CSV?

tests/integ/modin/groupby/test_groupby_apply.py
```diff
@@ -857,6 +858,28 @@ def operation(df):
...
+        eval_snowpark_pandas_result(
+            pd.DataFrame(diamonds_pd),
+            diamonds_pd,
+            lambda diamonds: diamonds.groupby("cut").apply(
+                lambda x: x.sort_values("price", ascending=False).head(5),
+                include_groups=True,
```
sfc-gh-mvashishtha, line 879: looks like this PR is also fixing a bug that `include_groups=True` would raise an exception. Should remove the `include_groups=True` here and have a separate test case for that?

sfc-gh-azhan: I added a todo above to test both True and False in the next PR.

sfc-gh-mvashishtha (approved):

sfc-gh-jdu (approved):

PR #1603, opened 2024-05-16
SNOW-1370338: [Local Testing] Fix test failures in test_function.py

sfc-gh-jrose:

   Fixes SNOW-1370338

There are 5 test failures fixed in this change:

- **test_struct**
  `_object_construct` in `_functions.py` did not use a fresh index when combining column emulators during table creation. This made the logic in `construct_dict` unable to index correctly.
- **test_basic_string_operations**
  The `substring` method was not raising the expected error. In order to better match expected function experience I've added a catch-all in `handle_function_expression` that raises the new `SnowparkLocalTestingException` to mask all function call errors so that implementors don't have to worry about wrapping any exceptions that occur in the functions.
- **test_count_distinct**
  The count distinct test had an operation like this `df.select(count_distinct(df["*"]))`. This revealed that `handle_function_expression` did not have good support for handling the star expression. I've added some logic to unwrap the multiple columns that the star expression refers to, but I'm not sure this is the right approach. There may be somewhere upstream where the star expression could be replaced with multiple Attribute expressions instead, but I'm not sure where that would be.
- **test_date_add_date_sub**
  The `add_timedelta` helper function for `date_add` could not correctly handle numpy data in timedeltas. I've added an explicit cast to float which should avoid using numpy types.
- **test_create_map**
  This test had the same issue as test_struct.


src/snowflake/snowpark/mock/_analyzer.py
```diff
@@ -318,9 +318,18 @@ def analyze(
                     expr.api_call_source, TelemetryField.FUNC_CAT_USAGE.value
                 )
             func_name = expr.name.upper()
+
+            children = []
+            for c in expr.children:
+                extracted = self.to_sql_avoid_offset(c, expr_to_alias)
+                if isinstance(extracted, list):
```
sfc-gh-aling, line 325: why we need this change -- what's the test case that fails because of this?

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -384,18 +384,33 @@ def handle_function_expression(
...
         if key == spec.varargs:
-            to_pass_args.extend(
-                [
-                    calculate_expression(
-                        c,
-                        input_data,
-                        analyzer,
-                        expr_to_alias,
-                        keep_literal=keep_literal,
+            # TODO is this really the best place for this?
```
sfc-gh-aling, line 387: I feel this can be added to into `calculate_expression` as another top level if instance check

src/snowflake/snowpark/mock/_analyzer.py
sfc-gh-jrose: This is part of the change for test_count_distinct. `self.to_sql_avoid_offset` can call `self.analyze` which can either return string or list of strings. Previously the list comprehension expected only strings which caused one of the failure cases in test_count_distinct.

src/snowflake/snowpark/mock/_plan.py
sfc-gh-stan: I agree with Adam that this should be in `calculate_expression` instead. `calculate_expression` should be able to return a `TableEmulator` for expressions like `Star` and `SnowflakeUDTF`.

sfc-gh-stan: Let's isolate this change into a separate PR so we can merge the rest of the fixes.

tests/integ/test_function.py
```diff
@@ -643,16 +636,12 @@ def test_basic_numerical_operations_negative(session, local_testing_mode):
...
-    reason="BUG: SNOW-1370338: substring raises unsupported operand type(s) for -: 'str' and 'int'",
-)
-def test_basic_string_operations(session):
+def test_basic_string_operations(session, local_testing_mode):
     # Substring
     df = session.create_dataframe(["a not that long string"], schema=["a"])
     with pytest.raises(SnowparkSQLException) as ex_info:
         df.select(substring("a", "b", 1)).collect()
-    assert "Numeric value 'b' is not recognized" in str(ex_info)
+    assert local_testing_mode or "Numeric value 'b' is not recognized" in str(ex_info)
```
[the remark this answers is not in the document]
sfc-gh-jrose:
```
snowflake.snowpark.mock.exceptions.SnowparkLocalTestingException: Error executing mocked function 'substring'. See error traceback for detailed information.```
Then further up in the traceback there is
```
TypeError: unsupported operand type(s) for -: 'str' and 'int'
```

sfc-gh-stan (approved):

sfc-gh-aling (approved):

PR #1623, opened 2024-05-17
SNOW-915448: [Local Testing] Fix precision for DecimalTypes

sfc-gh-jrose:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR

   Fixes SNOW-915448

   After reviewing the current precision failures I found that precision is actually fairly correct for local testing. _IntegralType data uses the same size integers as live so there were no issues there. DecimalType data is correctly using Decimal to represent the data. The Decimal class does not limit precision and scale the same as the server side does in live though. To better reflect server side logic I've modified the logic for DecimalTypes so that when data is collected or shown the data is correctly truncated to fit the expected precision.

  There remains one disparity for precision. Aggregated DecimalTypes all map to `Number(38,6)` except for when data is aggregated over a range. This change does not address that problem yet because it would require a significant refactor to how we call aggregation functions. I've added a new ticket with my notes on that problem and updated tests that are failing due to windowed data imprecision.


sfc-gh-stan (approved):
This is great. Please document the window aggregation precision as a known issue and create a follow up task.

PR #1625, opened 2024-05-17
SNOW-1434962: Make groupby.apply sort result deterministic.

sfc-gh-mvashishtha:

   

   Fixes SNOW-1434962

   [`sort_values`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sort_values.html) by default is not stable since it uses `kind="quicksort"`. Use a stable `nlargest` call with the default of `keep="first"` so that when multiple rows have the same price, we get them in the order that they appear in the input dataframe.

    Unfortunately I couldn't reproduce the flake locally by running the test repeatedly unless I manually changed the order of the input dataframe before the `sort_values` to simulate instability of `sort_values`. However, I think now we should be able to count on the stability of `nlargest`.

### DIFF
--- a/tests/integ/modin/groupby/test_groupby_apply.py
+++ b/tests/integ/modin/groupby/test_groupby_apply.py
@@ -874,7 +874,11 @@ class TestFuncReturnsScalar:
             pd.DataFrame(diamonds_pd),
             diamonds_pd,
             lambda diamonds: diamonds.groupby("cut").apply(
-                lambda x: x.nlargest(5, columns=["price"]),
+                # Use the stable "mergesort" algorithm to make the result order
+                # deterministic (see SNOW-1434962).
+                lambda x: x.sort_values(
+                    "price", ascending=False, kind="mergesort"
+                ).head(5),
                 include_groups=True,
             ),
         )
--- a/tests/integ/modin/groupby/test_groupby_apply.py
+++ b/tests/integ/modin/groupby/test_groupby_apply.py
@@ -863,9 +863,6 @@ class TestFuncReturnsScalar:
         udtf_count=UDTF_COUNT,
         join_count=JOIN_COUNT,
     )
-    @pytest.mark.xfail(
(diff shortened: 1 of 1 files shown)


tests/integ/modin/groupby/test_groupby_apply.py
```diff
@@ -877,7 +874,7 @@ def test_group_apply_return_df_from_lambda(self):
             pd.DataFrame(diamonds_pd),
             diamonds_pd,
             lambda diamonds: diamonds.groupby("cut").apply(
-                lambda x: x.sort_values("price", ascending=False).head(5),
+                lambda x: x.nlargest(5, columns=["price"]),
```
sfc-gh-azhan, line 877: this does not fix the issue. pandas maintain the order from the original frame but we didn't.

sfc-gh-azhan (changes requested):

tests/integ/modin/groupby/test_groupby_apply.py
mvashishtha: the `sort_values` is running inside pandas. pandas `sort_values` isn't guaranteed to be `stable` with the default `kind`. In the current version of the code, neither Snowpark pandas nor pandas is expected to necessarily maintain order.

sfc-gh-azhan: can we just set `kind = 'mergesort'`?

sfc-gh-azhan: Thank for helping this BTW

sfc-gh-mvashishtha: done.

sfc-gh-azhan (approved):

sfc-gh-azhan:
try it now: https://ci-dev-142.int.snowflakecomputing.com/job/SnowparkPythonSnowPandasDailyRegressRunner/432/

sfc-gh-nkrishna (approved):
Thanks for investigating!

PR #1626, opened 2024-05-17
SNOW-1435354: Add CTE optimization tests to snowpark pandas daily precommit tests

sfc-gh-jdu:

   

   Fixes SNOW-1435354

   Add CTE optimization tests to snowpark pandas daily precommit tests. A successful run: https://github.com/snowflakedb/snowpark-python/actions/runs/9406047620

### DIFF
--- a/.github/workflows/daily_modin_precommit.yml
+++ b/.github/workflows/daily_modin_precommit.yml
@@ -215,6 +215,74 @@ jobs:
             .tox/.coverage
             .tox/coverage.xml
 
+  test-enable-cte-optimization:
+    name: Test Enable CTE Optimization modin-${{ matrix.os.download_name }}-${{ matrix.python-version }}-${{ matrix.cloud-provider }}
+    needs: build
+    runs-on: ${{ matrix.os.image_name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - image_name: macos-latest
+            download_name: macos  # it includes doctest
+        python-version: ["3.9", "3.10", "3.11"]
+        cloud-provider: [aws]
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+      - name: Decrypt parameters.py
+        shell: bash
+        run: .github/scripts/decrypt_parameters.sh
+        env:
+          PARAMETER_PASSWORD: ${{ secrets.PARAMETER_PASSWORD }}
+          CLOUD_PROVIDER: ${{ matrix.cloud-provider }}
+      - name: Download wheel(s)
+        uses: actions/download-artifact@v3
+        with:
+          name: wheel
+          path: dist
+      - name: Show wheels downloaded
+        run: ls -lh dist
+        shell: bash
+      - name: Upgrade setuptools, pip and wheel
+        run: python -m pip install -U setuptools pip wheel
+      - name: Install tox
+        run: python -m pip install tox
+      - if: ${{ contains('macos', matrix.os.download_name) }}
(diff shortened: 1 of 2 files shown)


tests/integ/modin/frame/test_quantile.py
```diff
@@ -131,9 +131,8 @@ def test_quantile_datetime_negative():
         snow_df.quantile(numeric_only=False)
 
 
-@pytest.mark.xfail(
+@pytest.mark.skip(
```
sfc-gh-jdu, line 134: It has to be skipped because without sql count check it doesn't fail. This xfail is actually a skip here because it will definitely fail due to `@sql_count_checker(query_count=0)`

sfc-gh-jrose (approved):

.github/workflows/daily_modin_precommit.yml
```diff
@@ -215,6 +215,74 @@ jobs:
             .tox/.coverage
             .tox/coverage.xml
 
+  test-enable-cte-optimization:
```
sfc-gh-azhan, line 218: What's the long term plan? e.g., deprecate this job once CTE is enabled by default?

sfc-gh-jdu: yes

.github/workflows/daily_modin_precommit.yml
```diff
@@ -215,6 +215,74 @@ jobs:
...
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+      - name: Decrypt parameters.py
+        shell: bash
+        run: .github/scripts/decrypt_parameters.sh
+        env:
+          PARAMETER_PASSWORD: ${{ secrets.PARAMETER_PASSWORD }}
+          CLOUD_PROVIDER: ${{ matrix.cloud-provider }}
+      - name: Download wheel(s)
+        uses: actions/download-artifact@v3
```
sfc-gh-yzou, line 246: v3 -> v4

.github/workflows/daily_modin_precommit.yml
```diff
@@ -215,6 +215,74 @@ jobs:
...
+  test-enable-cte-optimization:
+    name: Test Enable CTE Optimization modin-${{ matrix.os.download_name }}-${{ matrix.python-version }}-${{ matrix.cloud-provider }}
+    needs: build
+    runs-on: ${{ matrix.os.image_name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - image_name: macos-latest
+            download_name: macos  # it includes doctest
```
sfc-gh-yzou, line 227: it seems we are using macos for most daily runs, but we have larger machines for linux and windows, shall we use a different one?

tests/integ/modin/frame/test_quantile.py
sfc-gh-yzou: I see the point about why this test will fail, but I am curious is this causing any problem to the CTE tests? I thought you skip all sql_counter check already

sfc-gh-jdu: xfail with strict=True expects this test to fail because of query_count=0. When we skip sql counter check in with CTE enabled, this test will not fail. Then xfail will "fail" this test consistently.

.github/workflows/daily_modin_precommit.yml
sfc-gh-jdu: done

tests/integ/modin/frame/test_quantile.py
sfc-gh-yzou: I see. Thanks for the explanation!

sfc-gh-yzou (approved):

.github/workflows/daily_modin_precommit.yml
sfc-gh-azhan: Now you use ubuntu to run doctests which is new to us. So it may have different behavior than macos. Just a reminder

PR #1646, opened 2024-05-21
SNOW-1440016 Make sure all modin doctests are covered in CI

sfc-gh-azhan:
SNOW-1440016 Make sure all modin doctests are covered in CI

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,7 +35,6 @@ def pytest_collection_modifyitems(items) -> None:
     """Applies tags to tests based on folders that they are in."""
     top_test_dir = Path(__file__).parent
     top_doctest_dir = top_test_dir.parent.joinpath("src/snowflake/snowpark")
-    modin_doctest_dir = top_doctest_dir.joinpath("modin/pandas")
     for item in items:
         item_path = Path(str(item.fspath)).parent
         try:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -35,7 +35,6 @@ def pytest_collection_modifyitems(items) -> None:
     """Applies tags to tests based on folders that they are in."""
     top_test_dir = Path(__file__).parent
     top_doctest_dir = top_test_dir.parent.joinpath("src/snowflake/snowpark")
-    modin_doctest_dir = top_doctest_dir.joinpath("modin/pandas")
```
sfc-gh-azhan, line 38: we missed those doctests under new plugin folders

sfc-gh-nkrishna (approved):

tests/conftest.py
noloerino: Are there any files in the new plugin folders that need to have doctests run? All the docstring overrides get applied to the appropriate class in modin/pandas before doctests for those are run.

sfc-gh-azhan: oh, so they did run? One change we will have in index is that we will only have the index plugin but no index.py in the modin frontend at least in our repo. So we have to include them to run.

One more question, once we use modin OSS, how to run doctests?

sfc-gh-joshi: > oh, so they did run? One change we will have in index is that we will only have the index plugin but no index.py in the modin frontend at least i
n our repo. So we have to include them to run.

Sure, that makes sense.

> One more question, once we use modin OSS, how to run doctests?

This is something I'm not too sure about, maybe @sfc-gh-dpetersohn would know. My guess is we'd have to mess with conftest.py a little to run the doctests defined in the docstring overrides, but I'm not sure how to handle docstrings that we don't override (if any).

sfc-gh-azhan: track it here https://snowflakecomputing.atlassian.net/browse/SNOW-1440221

sfc-gh-aling (approved):

PR #1647, opened 2024-05-21
SNOW-1370173: [Local Testing] Fix test_createDataFrame_with_given_schema_timestamp

sfc-gh-jrose:

   

   Fixes SNOW-1370173

  This test showed a couple of bugs in the `to_timestamp` function. First, by default to_timestamp should be using `to_timestamp_ntz` logic. This means timezone information should have been stripped, but it was not previously. Second, _to_timestamp would through an error when an empty column was passed in. This is resolved by short circuiting on empty input. Finally, to_timestamp_ntz could not handle nullable data correctly so I've added a condition to fix that.


tests/integ/scala/test_dataframe_suite.py
```diff
@@ -1813,9 +1809,10 @@ def test_createDataFrame_with_given_schema_timestamp(session):
     ]
     df = session.create_dataframe(data, schema)
     schema_str = str(df.schema)
+
     assert (
         schema_str
-        == "StructType([StructField('TIMESTAMP', TimestampType(tz=ntz), nullable=True), "
+        == f"StructType([StructField('TIMESTAMP', TimestampType({'' if local_testing_mode else 'tz=ntz'}), nullable=True), "
```
sfc-gh-aling, line 1815: if in live mode schema `TimestampTimeZone.NTZ`  is returned for input `StructField("timestamp", TimestampType()),`.

shouldn't we update our mocking implementation to align with the behavior?

src/snowflake/snowpark/mock/_functions.py
sfc-gh-aling, line 1: changelog?

tests/integ/scala/test_dataframe_suite.py
sfc-gh-jrose: In some scenarios it returns as NTZ, but in others it just remains DEFAULT. I have not worked out the logic of when it favors one or the other. It does however always seem to expect the cast to act like NTZ. I expect there's some inconsistency with how we type TimestampType outside of local testing as well so I think resolving this will be a larger refactor than we want before local testing release.

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -822,7 +823,7 @@ def mock_timestamp_ntz(
     result = _to_timestamp(column, fmt, try_cast, enforce_ltz=True)
     # Cast to NTZ by removing tz data if present
     return ColumnEmulator(
-        data=[x.replace(tzinfo=None) for x in result],
+        data=[None if x is None else x.replace(tzinfo=None) for x in result],
```
sfc-gh-stan, line 826: NIT: can we rewrite this to use `try_convert`?

sfc-gh-stan (approved):

src/snowflake/snowpark/mock/_functions.py
sfc-gh-jrose: Yes, but it's a bit more verbose since I'd have to use a lambda function.

sfc-gh-aling (approved):

tests/integ/scala/test_dataframe_suite.py
sfc-gh-aling: interesting, it's the scenario that the data contains timezone while the schema doesn't. but I agree this should not block our release

PR #1652, opened 2024-05-22
[SNOW-902943]: Add support for pd.NamedAgg in DataFrame and Series.agg

sfc-gh-rdurrani:

   

   Fixes SNOW-902943

Adds support for NamedAggregations for df and series.agg


src/snowflake/snowpark/modin/pandas/groupby.py
```diff
@@ -602,6 +604,17 @@ def aggregate(
             how="axis_wise",
             is_result_dataframe=is_result_dataframe,
         )
+
```
sfc-gh-rdurrani, line 607: When parsing the named aggregations, we go from a dictionary mapping label -> tuple(column_name, agg_func) to a dictionary mapping column_name -> (label, agg_func) which causes us to potentially return the columns out of order for groupby.agg. Rather than reimplementing the parsing for function, I figured it was easier (and pretty cheap) to just check the order of the columns on the output, and reindex with the correct order. Are folks ok with that?

sfc-gh-joshi: Can you clarify why this would be out of order? Shouldn't the columns of the output follow the iteration order of the input dictionary?

sfc-gh-rdurrani: Sure! Basically, say we have the following: `df.groupby('by_col').agg(x=('A', 'min'), y=('B', 'max'), z=('A', 'max'))`. When we process the aggregation functions, we convert it into a dictionary that looks like this: `{'A': [AggFuncWithLabel(func='min', pandas_label='x'), AggFuncWithLabel(func='max', pandas_label='z')], 'B': AggFuncWithLabel(func='max', pandas_label='y')}` - because we group by the column that is being aggregated on internally. When we process this dictionary, we then end up processing in the order x, z, y (rather than x, y, z) and so that is the order of the columns (or if this were df.agg, the order of the indices) that we output.

sfc-gh-rdurrani: Note that the order will always be x, z, y, because, as you point out, the iteration order of the input dictionary is respected. We just end up reordering the ordering of the original input dictionary by the column that is aggregated on, while breaking ties with the original ordering.

src/snowflake/snowpark/modin/pandas/groupby.py
```diff
@@ -602,6 +608,22 @@ def aggregate(
...
+        if uses_named_agg:
+            # When we use NamedAggregations, the output columns must be in the same order
+            # as they were passed in via the kwargs, but due to how we process the NamedAggregations
+            # they can end up out of order, as named aggregations on the same column will be moved to
+            # be contiguous.
+            # Example: say we have new_col1=('A', 'max'), new_col2=('B', 'min'), new_col3=('A', 'min')
+            # Our result will have columns ordered like so: [new_col1, new_col3, new_col2]. We need
```
sfc-gh-yzou, line 618: why we would have new_col1, new_col3, and new_col2, i assume if you have a mapping entered, the order should be expected?
Also can we try to do the indexing correctly at query compiler groupby_agg layer, since we do the ordering preserving all at that layer today, and we want to have most of the functionality happen at query compiler layer with the modin upstreaming work also

src/snowflake/snowpark/modin/pandas/utils.py
```diff
@@ -568,17 +570,43 @@ def extract_validate_and_try_convert_named_aggs_from_kwargs(
...
 
-            if value[0] in named_aggs:
-                if not isinstance(named_aggs[value[0]], list):
-                    named_aggs[value[0]] = [named_aggs[value[0]]]
-                named_aggs[value[0]] += [
-                    AggFuncWithLabel(func=value[1], pandas_label=key)
-                ]
+            # This function converts our named aggregations dictionary from a mapping of
+            # new_label -> tuple[column_name, agg_func] to a mapping of
+            # column_name -> tuple[agg_func, new_label] in order to process
```
sfc-gh-yzou, line 575: Not sure we need to do this here? it seems that the difference here is that when the aggfunc is in the format with new_label -> tuple[column_name, agg_func], the new_label is treated as new row label in regular aggregation, and new column label in groupby agg, we should be able to handle this fairly straightforward at query compiler layer, right?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4409,19 +4441,142 @@ def generate_agg_qc(
...
-                        column: AggFuncInfo(
-                            func=agg_func if column in cols else "min",
-                            is_dummy_agg=column not in cols,
-                        )
-                        for column in column_to_agg_func.keys()
-                    }
-                    single_agg_func_query_compilers.append(
-                        generate_agg_qc(
-                            col_single_agg_func_map, get_pandas_aggr_func_name(agg_func)
+                if uses_named_aggs_df:
```
sfc-gh-yzou, line 4444: the should squeeze is never going to be true when uses_named_aggs_df is true, is that correct?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4135,12 +4136,57 @@ def agg(
...
         # by snowflake engine.
-        if not check_is_aggregation_supported_in_snowflake(func, kwargs, axis):
-            ErrorMessage.not_implemented(
-                f"Aggregate function {func} with parameters "
-                + ", ".join([f"{k}={v}" for k, v in kwargs.items()])
-                + " not supported yet in Snowpark pandas."
-            )
+        # If we are using Named Aggregations, we need to do our supported check slightly differently.
+        uses_named_aggs_df = (
+            is_dict_like(func)
```
sfc-gh-yzou, line 4141: can you give an example here about how the data structure format look like when checking if it whether uses_named_aggs_df?

src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -703,17 +704,28 @@ def aggregate(
...
+        if func is None:
+            if axis == 1:
+                raise ValueError(
+                    "`func` must not be `None` when `axis=1`. Named aggregations are not supported with `axis=1`."
+                )
+            func = extract_validate_and_try_convert_named_aggs_from_kwargs(
+                self, allow_duplication=False, axis=axis, is_from_agg=True, **kwargs
+            )
+        else:
+            func = validate_and_try_convert_agg_func_arg_func_to_str(
```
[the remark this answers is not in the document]
sfc-gh-joshi: Yes, though our frontend implementation of `aggregate` already diverges pretty heavily from the modin version. We might want to use the extension API to overwrite `aggregate` when we start the migration.

src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -703,17 +704,28 @@ def aggregate(
...
+                axis=axis,
+            )
 
         # This is to stay consistent with pandas result format, when the func is single
         # aggregation function in format of callable or str, reduce the result dimension to
         # convert dataframe to series, or convert series to scalar.
-        need_reduce_dimension = (
+        # Note: When named aggregations are used, the result is not reduced, even if there
+        # is only a single function.
+        need_reduce_dimension = func is not None and (
```
sfc-gh-joshi, line 728: When would `func` be None here? Doesn't the above argument validation always set `func` to a non-null value?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 - Added partial support for `DataFrame.pivot_table` with no `index` parameter, as well as for `margins` parameter.
 - Added appropriate error messages for is_permanent/anonymous udf/sproc registration to make it more clear that those features are not yet supported.
+- Added support for named aggregations in `DataFrame.aggregate` and `Series.aggregate`.
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: We support NamedAgg + the format x=('A', 'sum') which is a named aggregation but not a pd.NamedAgg - should I still change it?

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-joshi: So in your example (`df.groupby('by_col').agg(x=('A', 'min'), y=('B', 'max'), z=('A', 'max'))`), native pandas output has columns `x, y, z` in that order? 

Even though reindexing the columns should be cheap, I think it would be better to change the logic of our NamedAgg parsing, or just return an additional list with the appropriate column order. If I'm understanding you correctly, for `df.agg` this would affect the order of the row index in the output frame, and reindexing in that case is more expensive, so I'd say it's worth generalizing the ordering logic.

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
```diff
@@ -826,6 +826,66 @@ def aggregate_with_ordered_dataframe(
     return agg_ordered_dataframe
 
 
+def convert_named_agg_func_art_to_include_snowflake_quoted_identifier(
```
sfc-gh-joshi, line 829:
```suggestion
def convert_named_agg_func_arg_to_include_snowflake_quoted_identifier(
```
typo?

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
```diff
@@ -826,6 +826,66 @@ def aggregate_with_ordered_dataframe(
...
+    internal_frame: InternalFrame,
+    agg_func: dict[Hashable, dict[Hashable, AggFuncWithLabel]],
```
sfc-gh-joshi, line 831: Is `agg_func` the result of `extract_validate_and_try_convert_named_aggs_from_kwargs`? The name of this function is a little misleading since it's not actually a `NamedAgg` object at this point.

tests/integ/modin/frame/test_aggregate.py
sfc-gh-joshi, line 1: Did you add any tests where the output names specified in the NamedAgg require us to reindex to get the correct order?

sfc-gh-joshi (commented):
I reviewed most of the frontend/argument parsing logic, and haven't looked at the query compiler changes yet since I left a few questions about column ordering that I'd like to understand first.

src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -703,17 +704,28 @@ def aggregate(
...
-        func = validate_and_try_convert_agg_func_arg_func_to_str(
-            agg_func=func,
-            obj=self,
-            allow_duplication=False,
-            axis=axis,
-        )
+        if func is None:
+            if axis == 1:
+                raise ValueError(
+                    "`func` must not be `None` when `axis=1`. Named aggregations are not supported with `axis=1`."
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: This error isn't the same - groupby_agg isn't supported (at all) when axis=1 by us, while neither we nor pandas support df.agg with axis=1 and named aggregations - I'm not sure it makes sense to move the error, since we'd still need the error checking in groupby_agg anyways.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-rdurrani: For `df.agg` I handle the logic of passing order along (which ends up making the code more complex) since I want to avoid the reindexing. I'm thinking the additional code complexity doesn't super make sense for groupby_agg though, since the reindex on the column side is much much cheaper.

src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -703,17 +704,28 @@ def aggregate(
...
+                )
+            func = extract_validate_and_try_convert_named_aggs_from_kwargs(
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: That's a good catch - I'll actually combine the logic and make the correct error message!

src/snowflake/snowpark/modin/pandas/groupby.py
```diff
@@ -573,15 +573,21 @@ def aggregate(
             )  # pragma: no cover
         if func is None:
             func = extract_validate_and_try_convert_named_aggs_from_kwargs(
-                obj=self, allow_duplication=True, axis=self._axis, **kwargs
+                obj=self,
+                allow_duplication=True,
+                axis=self._axis,
+                is_from_agg=False,
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: Ah sorry - this argument is to check if its from df/series.agg vs. df/series.groupby_agg - we parse the arguments a little differently depending on whether its from agg directly or from groupby_agg, due to the ordering issue described above.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-rdurrani: Sounds good - added a comment explaining why this happens, and moved the reindexing to the qc layer!

src/snowflake/snowpark/modin/pandas/utils.py
sfc-gh-rdurrani: The issue is the ordering change. Since the order of the aggregations is not preserved by this dictionary, we need to reorder at the end of the operation. We can do so cheaply for groupby_agg because the new aggregations are the names of the columns - we can just swap the order of the data_column_pandas_labels + data_column_snowflake_quoted_identifiers before making the InternalFrame, and have the correct ordering reflected. For the agg though, the new labels are the index, so reordering is expensive, and so we need a different format of the map that preserves the order of the named aggregations.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: Yup - will add a comment explaining this.

src/snowflake/snowpark/modin/pandas/utils.py
sfc-gh-rdurrani: Can add a code comment to clarify.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-joshi: You need to determine the order of the columns/index either way, right? I think it'd be better to share the logic for computing this order when you parse the NamedAgg, but make the decision of whether to use that ordering immediately when computing aggregations (like for `df.agg`) or to reindex the columns afterwards (like what you're suggesting for `groupby.agg`) to the caller.

src/snowflake/snowpark/modin/pandas/base.py
sfc-gh-rdurrani: That is a good catch - let me fix that!

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
sfc-gh-rdurrani: Good catch, thank you!

tests/integ/modin/frame/test_aggregate.py
sfc-gh-rdurrani: yup!

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-rdurrani: Resolved offline - passing the order explicitly and then reordering the query compilers before concatenation.

src/snowflake/snowpark/modin/pandas/utils.py
sfc-gh-rdurrani: Resolved via passing order explicitly, and reordering query compilers prior to concatenation.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2675,6 +2672,40 @@ def convert_func_to_agg_func_info(
...
             col_agg_op.agg_snowflake_quoted_identifier for col_agg_op in agg_col_ops
         ]
+
+        # The ordering of the named aggregations is changed by us when we process
+        # the agg_kwargs into the func dict (named aggregations on the same
+        # column are moved to be contiguous, see groupby.py::aggregate for an
+        # example). We need to check if the order of the output columns is correct,
+        # and if not, reorder them.
+        if uses_named_aggs:
+            correct_ordering = list(agg_kwargs.keys())
```
sfc-gh-joshi, line 2682: Are the vanity arguments `_formatted_named_kwargs` and `_correct_aggregation_order` not passed at this point?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2675,6 +2672,40 @@ def convert_func_to_agg_func_info(
...
+                            )
+                        ]
+                    )
+                )
+                new_data_column_pandas_labels = list(
+                    data_column_label_to_quoted_identifier[0]
+                )
+                new_data_column_quoted_identifier = list(
+                    data_column_label_to_quoted_identifier[1]
+                )
```
sfc-gh-joshi, line 2708:
```suggestion
                new_data_column_pandas_labels, new_data_column_quoted_identifier = list(
                    zip(
                        *[
                            pair
                            for column_label in correct_ordering
                            for pair in filter(
                                lambda pair: pair[0] == column_label,
                                data_column_label_to_quoted_identifier,
                            )
                        ]
                    )
                )
```

IMO tuple assignment is easier to read.

You could also consider directly sorting `data_column_label_to_quoted_identifier` by key= each pair's index in `correct_ordering`, but I'm not sure if that would make the code harder to read or not.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4418,19 +4442,114 @@ def generate_agg_qc(
...
+                    ) -> dict[PandasLabelToSnowflakeIdentifierPair, AggFuncInfo]:
+                        """
+                        Helper function to produce the agg func map for a single aggregation, including dummy aggregations.
+
+                        Notes:
+                        Adds dummy aggregations for all columns that will be aggregated over.
+                        """
+                        col_single_agg_func_map = {}
+                        for c in column_to_agg_func.keys():
+                            if c == identifier_pair:
```
sfc-gh-joshi, line 4460: Why not `if identifier_pair in column_to_agg_func` + `col_single_agg_func_map[identifier_pair] = (...)`?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4418,19 +4442,114 @@ def generate_agg_qc(
...
+                    # in the correct order, which we recieve from the frontend, so that when the concatenation occurs,
+                    # the final QueryCompiler is ordered correctly.
+                    index_label_to_generated_qcs: dict[
+                        Hashable, list["SnowflakeQueryCompiler"]
+                    ] = {}
+                    for (
+                        identifier_pair,
+                        agg_func_with_label,
+                    ) in column_to_agg_func.items():
+                        if isinstance(agg_func_with_label, list):
```
sfc-gh-joshi, line 4484: Can you convert agg_func_with_label to a one-element list if it's a scalar to avoid the duplicated code here?

tests/integ/modin/frame/test_aggregate.py
```diff
@@ -162,6 +172,99 @@ def test_agg_dup_col(numeric_native_df, func, expected_union_count):
         eval_snowpark_pandas_result(snow_df, numeric_native_df, func)
 
 
+def test_named_agg_dup_col_pandas_fails_snowpark_pandas_supports(numeric_native_df):
```
sfc-gh-joshi, line 175: Can you split this test up or parametrize it so one failure doesn't cause the whole rest of the test to be skipped?

tests/integ/modin/frame/test_aggregate.py
```diff
@@ -192,6 +295,35 @@ def test_agg_with_all_missing(func, expected_union_count):
         eval_snowpark_pandas_result(snow_missing_df, missing_df, func)
 
 
+# For some reason, this aggregation causes pandas to fail.
+# This ticket: https://github.com/pandas-dev/pandas/issues/58810
```
sfc-gh-joshi, line 299: Do you know what exactly is causing this bug? Is it something in the values of the frame, or is it the names of the input/output columns?

tests/integ/modin/groupby/test_groupby_named_agg.py
```diff
@@ -55,3 +63,74 @@ def test_valid_func_with_named_agg_errors(basic_snowpark_pandas_df):
         assert_exception_equal=False,  # There is a difference in our errors.
         expect_exception_type=TypeError,
     )
+
+
+# Query count is 2 in below test due to line 65, basic_snowpark_pandas_df.to_pandas()
```
sfc-gh-joshi, line 68: Line number is variable, so you should omit it.
```suggestion
# Query count is 2 in below test due to basic_snowpark_pandas_df.to_pandas()
```

sfc-gh-joshi: Is it possible to change the fixture to a native pandas frame so we don't do this extra to_pandas()?

tests/integ/modin/groupby/test_groupby_named_agg.py
```diff
@@ -55,3 +63,74 @@ def test_valid_func_with_named_agg_errors(basic_snowpark_pandas_df):
...
+    eval_snowpark_pandas_result(
+        basic_snowpark_pandas_df,
+        basic_snowpark_pandas_df.to_pandas(),
+        lambda df: df.groupby("col1").agg(
+            new_col1=("col1", min), new_col2=("col2", min), new_col3=("col1", max)
+        ),
+    )
+
+
+# Query count is 2 in below test due to line 65, basic_snowpark_pandas_df.to_pandas()
```
sfc-gh-joshi, line 81:
```suggestion
# Query count is 2 in below test due to basic_snowpark_pandas_df.to_pandas()
```

tests/integ/modin/groupby/test_groupby_named_agg.py
```diff
@@ -55,3 +63,74 @@ def test_valid_func_with_named_agg_errors(basic_snowpark_pandas_df):
...
+        columns=["new_col1", "new_col1", "new_col2", "new_col3", "new_col3"],
+        index=pd.Index([3, 5, 6, 15, 16, 17], name="col4"),
+    )
+    snow_df = basic_snowpark_pandas_df.groupby("col4").agg(
+        new_col1=("col1", min), new_col2=("col2", min), new_col3=("col1", max)
+    )
+    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(snow_df, result_df)
+
+
+# Query count is 2 in below test due to line 65, basic_snowpark_pandas_df.to_pandas()
```
sfc-gh-joshi, line 112:
```suggestion
# Query count is 2 in below test due to basic_snowpark_pandas_df.to_pandas()
```

tests/integ/modin/groupby/test_groupby_named_agg.py
```diff
@@ -55,3 +63,74 @@ def test_valid_func_with_named_agg_errors(basic_snowpark_pandas_df):
...
+        lambda df: df.groupby("col1").agg(
+            new_col1=("col1", min), new_col2=("col2", min), new_col3=("col1", max)
+        ),
+    )
+
+
+# Query count is 2 in below test due to line 65, basic_snowpark_pandas_df.to_pandas()
+# That produces the native pandas DataFrame to test against.
+@sql_count_checker(query_count=2)
+def test_named_agg_output_column_order_with_dup_columns(basic_snowpark_pandas_df):
```
sfc-gh-joshi, line 84: Do you know what happens if you specify the same kwarg twice, like `df.agg(new_col1=("col1", min), new_col1=("col2", min))`? Is this worth adding a negative test for?

tests/integ/modin/groupby/test_groupby_series.py
```diff
@@ -76,10 +80,41 @@ def test_groupby_agg_series(agg_func, sort):
     index.names = ["grp_col"]
     series = pd.Series([3.5, 1.2, 4.3, 2.0, 1.8], index=index)
 
+    def perform_groupby(se):
+        se = se.groupby(by="grp_col", sort=sort)
+        if isinstance(agg_func, dict):
```
sfc-gh-joshi, line 85: Can `agg_func` also be passed as a dict?

sfc-gh-joshi (commented):
Thanks! I left a few more minor comments/suggestions.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: This codepath is in qc.groupby_agg - those vanity arguments are passed along the df/series.agg codepath (so they hit qc.agg) - so those arguments are not there here. (Those vanity arguments are actually passed as a surrogate to passing agg_kwargs as is done here)

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: I'm not sure I follow this comment?

sfc-gh-joshi: Never mind, I think I misunderstood the purpose of this function. I thought it was only generating a single column corresponding to `identifier_pair`.

I didn't realize that unlike df.agg w/ list or dict arguments, which combine common aggregations together, the named aggregation code path gives every single argument its own 1-hot row.

```
>>> df = pd.DataFrame([[1, 2, 3],
                   [4, 5, 6],
                   [7, 8, 9],
                   [np.nan, np.nan, np.nan]],
                  columns=['A', 'B', 'C'])
>>> df.agg(x=("A", min), y=("B", max), z=("A", max))
     A    B
x  1.0  NaN
y  NaN  8.0
z  7.0  NaN
>>> df.agg({"A": [min, max], "B": max})
       A    B
min  1.0  NaN
max  7.0  8.0
```

tests/integ/modin/frame/test_aggregate.py
sfc-gh-rdurrani: Not sure honestly - inserting a breakpoint into pandas, looks like their agg is missing an aggregation:
```python
-> result = reconstruct_and_relabel_result(result, func, **kwargs)
(Pdb) result
     nan  na  values
min  NaN NaN     NaN
sum  NaN NaN    10.0
```
so when it tries to relabel the result, its execting axis length 3 probably, and errors out? Can't replicate it with differently named columns:
```python
In [5]: df = native_pd.DataFrame([[1, 2, 3], [4,5,6]], columns=list('aba'))

In [6]: df
Out[6]:
   a  b  a
0  1  2  3
1  4  5  6

In [7]: df.columns = list('abc')

In [8]: df.agg(x=('a', 'sum'), y=('b', 'sum'))
> /Users/rdurrani/.miniconda3/envs/snowpark/lib/python3.10/site-packages/pandas/core/frame.py(10138)aggregate()
-> return result
(Pdb) c
Out[8]:
     a    b
x  5.0  NaN
y  NaN  7.0

In [9]: df.agg(x=('a', 'sum'), y=('b', 'sum'), z=('c', 'sum'))
> /Users/rdurrani/.miniconda3/envs/snowpark/lib/python3.10/site-packages/pandas/core/frame.py(10138)aggregate()
-> return result
(Pdb) c
Out[9]:
     a    b    c
x  5.0  NaN  NaN
y  NaN  7.0  NaN
z  NaN  NaN  9.0

In [10]: df.agg(x=('a', 'sum'), y=('b', 'min'), z=('c', 'sum'))
> /Users/rdurrani/.miniconda3/envs/snowpark/lib/python3.10/site-packages/pandas/core/frame.py(10138)aggregate()
-> return result
(Pdb) c
Out[10]:
     a    b    c
x  5.0  NaN  NaN
y  NaN  2.0  NaN
z  NaN  NaN  9.0
```

tests/integ/modin/groupby/test_groupby_named_agg.py
sfc-gh-rdurrani: Its a syntax error:
```python
In [5]: df.agg(new_col1=("col1", min), new_col1=("col2", min))
  Cell In[5], line 1
    df.agg(new_col1=("col1", min), new_col1=("col2", min))
                                   ^
SyntaxError: keyword argument repeated: new_col1


In [6]: pd.DataFrame(df).agg(new_col1=("col1", min), new_col1=("col2", min))
  Cell In[6], line 1
    pd.DataFrame(df).agg(new_col1=("col1", min), new_col1=("col2", min))
                                                 ^
SyntaxError: keyword argument repeated: new_col1
```
so I don't think it makes sense adding a test for it?

tests/integ/modin/groupby/test_groupby_series.py
sfc-gh-rdurrani: Good catch - seems like it can't?
```python
In [10]: series.to_pandas().groupby(level=0).agg({"x": "min"})
SpecificationError                        Traceback (most recent call last)
Cell In[10], line 1
----> 1 series.to_pandas().groupby(level=0).agg({"x": "min"})

File ~/.miniconda3/envs/snowpark/lib/python3.10/site-packages/pandas/core/groupby/generic.py:257, in SeriesGroupBy.aggregate(self, func, engine, engine_kwargs, *args, **kwargs)
    255 kwargs["engine"] = engine
    256 kwargs["engine_kwargs"] = engine_kwargs
--> 257 ret = self._aggregate_multiple_funcs(func, *args, **kwargs)
    258 if relabeling:
    259     # columns is not narrowed by mypy from relabeling flag
    260     assert columns is not None  # for mypy

File ~/.miniconda3/envs/snowpark/lib/python3.10/site-packages/pandas/core/groupby/generic.py:335, in SeriesGroupBy._aggregate_multiple_funcs(self, arg, *args, **kwargs)
    332 if isinstance(arg, dict):
    333     if self.as_index:
    334         # GH 15931
--> 335         raise SpecificationError("nested renamer is not supported")
    336     else:
    337         # GH#50684 - This accidentally worked in 1.x
    338         msg = (
    339             "Passing a dictionary to SeriesGroupBy.agg is deprecated "
    340             "and will raise in a future version of pandas. Pass a list "
    341             "of aggregations instead."
    342         )

SpecificationError: nested renamer is not supported
```

tests/integ/modin/groupby/test_groupby_named_agg.py
sfc-gh-joshi: Makes sense.

sfc-gh-joshi (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4424,19 +4442,101 @@ def generate_agg_qc(
...
-                            is_dummy_agg=column not in cols,
+                if uses_named_aggs:
+                    # If this is true, then we are dealing with agg with NamedAggregations.
+
+                    def generate_single_agg_column_func_map(
+                        identifier_pair: PandasLabelToSnowflakeIdentifierPair,
+                        agg_func: AggFuncWithLabel,
+                    ) -> dict[PandasLabelToSnowflakeIdentifierPair, AggFuncInfo]:
+                        """
+                        Helper function to produce the agg func map for a single aggregation, including dummy aggregations.
```
sfc-gh-yzou, line 4453: Can you explain why you need this function here? it seems that with named agg, duplicated name is not allowed, and each named agg can only have one function. You should be able to concatenate them without the need the dummy agg? do you have an example/test explains why you need to do it like this?

src/snowflake/snowpark/modin/pandas/base.py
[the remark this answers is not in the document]
sfc-gh-yzou: I double checked, in pandas 2.0, groupby with axis = 1 is not supported anymore, i got error  NotImplementedError: axis other than 0 is not supported

I think we can keep this special for agg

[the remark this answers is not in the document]
sfc-gh-yzou: However, since we intended to put those functionality support decision for the particular backend, can we move this check to the query compiler agg to have it at the same place with the other not implemented check

tests/integ/modin/frame/test_aggregate.py
```diff
@@ -162,6 +172,95 @@ def test_agg_dup_col(numeric_native_df, func, expected_union_count):
...
+            result_df = result_df.sort_index()
+
+        with SqlCounter(query_count=1, union_count=2):
+            assert_snowpark_pandas_equal_to_pandas(
+                snow_df.agg(**kwargs),
+                result_df,
+            )
+
+    def test_multiple_agg_on_only_dup_columns(self, numeric_native_df):
+        # rename to have duplicated column
```
sfc-gh-yzou, line 232: is there a test with duplicated result lable like x = ('A', 'min'), x = ('B', 'max'), that should also error even if the no duplication in original df.

CHANGELOG.md
```diff
@@ -15,12 +15,17 @@
...
+
+- Fixed a bug that causes output of GroupBy.aggregate's columns to be ordered incorrectly.
+
 #### Improvements
 
 - Added partial support for `DataFrame.pivot_table` with no `index` parameter, as well as for `margins` parameter.
 - Aligned error experience when calling udf and sprocs.
 - Added appropriate error messages for is_permanent/anonymous udf/sproc registration to make it more clear that those features are not yet supported.
 - Updated the signature of `DataFrame.shift`/`Series.shift`/`DataFrameGroupBy.shift`/`SeriesGroupBy.shift` to match pandas 2.2.1. Snowpark pandas does not yet support the newly-added `suffix` argument, or sequence values of `periods`.
+- Added support for named aggregations in `DataFrame.aggregate` and `Series.aggregate`.
```
sfc-gh-yzou, line 28: can you add for axis = 0 here

src/snowflake/snowpark/modin/pandas/base.py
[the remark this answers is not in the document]
sfc-gh-rdurrani: This is a pandas error message parity check - so shouldn't it be in the frontend? For context, pandas does one of two things for named agg with axis=1:
1. If we call `agg` with index values that exist, we get a KeyError: col_name
2. If we call `agg` with index values that do not exist, we get a Columns do not exist error.
I've opened an issue [here](https://github.com/pandas-dev/pandas/issues/58807) for the pandas folks to standardize the error message, but based off of that, it seems that df.agg(axis=1, **named_aggs) is unsupported as a feature by pandas. Therefore, this isn't a backend specific decision, and should remain at the front end?

src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -768,7 +790,16 @@ def aggregate(
...
         # TypeError: got an unexpected keyword argument 'skipna'
         if is_dict_like(func):
+            order_of_aggregations = list(kwargs.keys())
+            formatted_kwargs = ", ".join(
+                [f"{key}={value}" for key, value in kwargs.items()]
+            )
             kwargs.clear()
+            # Used to make error message formatting a little cleaner
+            # when using named aggregations.
+            kwargs["_formatted_named_kwargs"] = formatted_kwargs
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: in groupby, the `agg_kwargs` are passed directly to the query compiler, so we can do the formatting there. Here the kwargs are cleared and not passed to the QC, so we'd have to parse them out of the agg func - which means that they may not be in order in the error message, and the code to parse them looks a little messier.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: the dummy aggs are necessary for when different aggs are on different columns - e.g. say we have columns A, B, C, and aggregations x=('A', 'min'), y=('C', 'max'). When we do the aggregations for 'A', we need to pass in the dictionary {('A', '"A"'): AggFuncInfo('min', 'x'), ('C', '"C"'): AggFuncInfo('min', 'y', is_dummy_agg=True)} - this function is just a helper function to loop through all columns being aggregated on, and add dummy_aggs for the columns that the current aggregation is on, the same as the dict comprehension in line 4530 - I could make this a dict comprehension too, or change line 4530 to use this function - do you have a preference for which to do?

tests/integ/modin/frame/test_aggregate.py
sfc-gh-rdurrani: That is a syntax error in python, so it should be intercepted by the python layer. See relevant comment here: https://github.com/snowflakedb/snowpark-python/pull/1652#discussion_r1612427237. To summarize, I'm not sure it makes sense to add a negative test for a python syntax error.

CHANGELOG.md
sfc-gh-rdurrani: Good catch, thank you!

src/snowflake/snowpark/modin/pandas/base.py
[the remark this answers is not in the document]
sfc-gh-yzou: That is fine, I think i was mis-reading the error, i thought it is a case that is not intended to support by us, which is suitable to decide at the query compiler with other not implemented error. However, that seems something that is not intended to support anyway, which should be fine at frontend layer.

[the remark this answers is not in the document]
sfc-gh-rdurrani: Sounds good! Marking as resolved!

CHANGELOG.md
```diff
@@ -15,6 +31,10 @@
 
 - Added `DataFrame.cache_result` and `Series.cache_result` methods for users to persist DataFrames and Series to a temporary table lasting the duration of the session to improve latency of subsequent operations.
 
+#### Bug Fixes
+
+- Fixed a bug that causes output of GroupBy.aggregate's columns to be ordered incorrectly.
```
sfc-gh-yzou, line 36: move this above to new release

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: resolved offline.

sfc-gh-yzou (approved):

PR #1657, opened 2024-05-22
SNOW-1432019 Calculate subtree query complexity

sfc-gh-aalam:

   

   Fixes SNOW-1432019

  see doc: https://docs.google.com/document/d/1IS8qyNmWecF_Lej723hlqXXKUVceATnfeBziyWbFlh0/edit


src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -80,6 +96,18 @@ def sql(self) -> str:
         )
         return f"{self.pretty_name}({children_sql})"
 
+    @property
```
sfc-gh-jdu, line 99: Can we add some comments here or in complexity_stat.py to brief talk about the main idea of complexity score calculation (e.g., use # of columns involved as a proxy, etc.)?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -348,6 +393,11 @@ def sql_in_subquery(self) -> str:
     def schema_query(self) -> str:
         return self.sql_query
 
+    @property
+    def individual_complexity_stat(self) -> Counter[str]:
+        # SELECT * FROM entity
+        return Counter({ComplexityStat.COLUMN.value: 1})
```
sfc-gh-jdu, line 399: `*` is regarded as one column instead of all columns?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -403,6 +453,18 @@ def query_params(self) -> Optional[Sequence[Any]]:
     def schema_query(self) -> str:
         return self._schema_query
 
+    @property
+    def individual_complexity_stat(self) -> Counter[str]:
+        if self.pre_actions:
+            # having pre-actions implies we have a non-select query followed by a
```
sfc-gh-jdu, line 459:
```suggestion
            # Currently having pre-actions implies we have a non-select query followed by a
```

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -658,6 +720,45 @@ def schema_query(self) -> str:
     def children_plan_nodes(self) -> List[Union["Selectable", SnowflakePlan]]:
         return [self.from_]
 
+    @property
```
sfc-gh-jdu, line 723: Should we override it for SelectSnowflakePlan by reading complexity score from snowflake_plan directly?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -58,6 +106,17 @@ def __init__(
         self.data = data
         self.schema_query = schema_query
 
+    @property
+    def individual_complexity_stat(self) -> Counter[str]:
+        # select $1, ..., $m FROM VALUES (r11, r12, ..., r1m), (rn1, ...., rnm)
+        # TODO: use ARRAY_BIND_THRESHOLD
+        return Counter(
```
sfc-gh-jdu, line 113: ah then the complexity of create_dataframe will be very high compared with select from a table. If the table is large, then at least execution of selecting from a table will be more "complex"?

src/snowflake/snowpark/_internal/analyzer/unary_plan_node.py
```diff
@@ -47,13 +85,37 @@ def __init__(
...
+            estimate += Counter({ComplexityStat.GROUP_BY.value: 1}) + sum(
+                (expr.cumulative_complexity_stat for expr in self.grouping_expressions),
+                Counter(),
+            )
+        else:
+            # LIMIT 1
+            estimate += Counter({ComplexityStat.LOW_IMPACT.value: 1})
+
+        get_complexity_stat = (
+            lambda expr: expr.cumulative_complexity_stat
```
sfc-gh-jdu, line 102:
```suggestion
            getattr(expr, "cumulative_complexity_stat", Counter({ComplexityStat.COLUMN.value: 1}))
```
and no function is needed?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -656,7 +656,7 @@ def get_result_and_metadata(
 
     def get_result_query_id(self, plan: SnowflakePlan, **kwargs) -> str:
         # get the iterator such that the data is not fetched
-        result_set, _ = self.get_result_set(plan, to_iter=True, **kwargs)
+        result_set, _ = self.get_result_set(plan, ignore_results=True, **kwargs)
```
sfc-gh-jdu, line 659: this might be in another PR

sfc-gh-jdu (commented):
Neat!

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-aalam: that's right. even if data created from `create_dataframe` may be smaller than a table and thus be less expensive in terms of execution, this complexity is only concerned with compiler complexity since we won't be able to estimate execution complexity without more data.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: yes. in terms of compiling complexity `*` is easier than `col1`, ... `col100`. So we need to reflect that

src/snowflake/snowpark/_internal/analyzer/binary_expression.py
```diff
@@ -2,8 +2,23 @@
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
 
+import sys
 from typing import AbstractSet, Optional
 
+# collections.Counter does not pass type checker. Changes with appropriate type hints were made in 3.9+
+if sys.version_info <= (3, 9):
```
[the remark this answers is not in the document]
sfc-gh-aalam: I can move it to a util file - that's a good idea.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: yeah. this was a mistake. thanks for catching it

src/snowflake/snowpark/_internal/analyzer/unary_plan_node.py
sfc-gh-aalam: yep. updated

src/snowflake/snowpark/_internal/analyzer/expression.py
sfc-gh-aalam: added

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: yep. added

src/snowflake/snowpark/_internal/analyzer/materialization_utils.py
```diff
@@ -0,0 +1,46 @@
...
+    KT = typing.TypeVar("KT")
+
+    class Counter(collections.Counter, typing.Counter[KT]):
+        pass
+
+else:
+    from collections import Counter  # noqa
+
+
+class ComplexityStat(Enum):
```
sfc-gh-yzou, line 22: I left a comment about maybe just change this to PlanNodeCategory to be more general, and instead of having LOW_IMPACT, we can just have others,  etc

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -927,7 +927,7 @@ def do_resolve_with_resolved_children(
                 )
 
         if isinstance(logical_plan, UnresolvedRelation):
-            return self.plan_builder.table(logical_plan.name)
+            return self.plan_builder.table(logical_plan.name, logical_plan)
```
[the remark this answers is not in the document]
sfc-gh-aalam: if you look around the code, logical_plan is passed in other places but was not added here. If we are only concerned about generating sql, this is not a problem but not passing logical plan will not add it as a child to SnowflakePlan later on. This created a missing child node and unaccounted complexity stat.

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
```diff
@@ -69,7 +84,10 @@ def __init__(self, left: LogicalPlan, right: LogicalPlan) -> None:
 
 
 class SetOperation(BinaryNode):
-    pass
+    @property
+    def individual_complexity_stat(self) -> Counter[str]:
+        # (left) operator (right)
+        return Counter({ComplexityStat.SET_OPERATION.value: 1})
```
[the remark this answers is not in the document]
sfc-gh-aalam: I added a single value in my initial implementation but some expressions generate sql in multiple categories like `Interval`, `TableFunctionJoin`, `Lateral`. What I can change here is to do something like this

```
# default case
class Expression
  @property 
  def individual_complexity_class(self) -> Optional[Enum]:
    return None
    
  @property 
  def individual_complexity_stat(self) -> Counter[str]:
    return Counter({self.individual_complexity_class.value: 1}) if self.individual_complexity_class else Counter()

# single class case
class SetOperation:
  @property
  def individual_complexity_class(self) -> Optional[Enum]:
    return ComplexityStat.SET_OPERATOR

# multiple case
class TableFunctionJoin:
  @property
  def individual_complexity_stat(self) -> Counter[str]:
    return {...}
```

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
```diff
@@ -187,3 +205,23 @@ def __init__(
     @property
     def sql(self) -> str:
         return self.join_type.sql
+
+    @property
+    def individual_complexity_stat(self) -> Counter[str]:
+        # SELECT * FROM (left) AS left_alias join_type_sql JOIN (right) AS right_alias match_cond, using_cond, join_cond
+        estimate = Counter({ComplexityStat.JOIN.value: 1})
+        if isinstance(self.join_type, UsingJoin) and self.join_type.using_columns:
+            estimate += Counter(
```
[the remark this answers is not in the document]
sfc-gh-aalam: this is the self component. A join statement can be constructed in different ways. Based on how analyzer generates the final sql, we calculate the contribution. You can see check out analyzer code here: https://github.com/snowflakedb/snowpark-python/blob/4ff6d42c2780113e8fc8c6de168429d369025cb5/src/snowflake/snowpark/_internal/analyzer/analyzer.py#L845-L868

and here: https://github.com/snowflakedb/snowpark-python/blob/4ff6d42c2780113e8fc8c6de168429d369025cb5/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py#L714-L741

src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -80,6 +85,24 @@ def sql(self) -> str:
...
+    def individual_complexity_stat(self) -> Counter[str]:
+        """Returns the individual contribution of the expression node towards the overall
+        compilation complexity of the generated sql.
+        """
+        return Counter()
+
+    @cached_property
+    def cumulative_complexity_stat(self) -> Counter[str]:
+        """Returns the aggregate sum complexity statistic from the subtree rooted at this
```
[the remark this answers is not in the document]
sfc-gh-aalam: we will not be updating trees at expression level but at logical node plan level. You can see the mechanism on how the update the values for logical plan in `@cumulative_complexity_stat.setter`

src/snowflake/snowpark/_internal/server_connection.py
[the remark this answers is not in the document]
sfc-gh-aalam: ignore it. it was added by mistake

src/snowflake/snowpark/_internal/analyzer/materialization_utils.py
sfc-gh-aalam: good point

src/snowflake/snowpark/_internal/analyzer/binary_expression.py
```diff
@@ -2,8 +2,23 @@
...
+    import collections
+    import typing
+
+    KT = typing.TypeVar("KT")
+
+    class Counter(collections.Counter, typing.Counter[KT]):
```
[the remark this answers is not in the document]
sfc-gh-aalam: same as simple map but it also lets you do map1 + map2
example:
map1 = {a: 1, b: 2}
map2 = {b:1, c:3}
map1 + map2 = {a: 1, b: 3, c: 3}

src/snowflake/snowpark/_internal/analyzer/complexity_stat.py
```diff
@@ -0,0 +1,24 @@
...
+    UNPIVOT = "unpivot"
+    WINDOW = "window"
+    GROUP_BY = "group_by"
+    PARTITION_BY = "partition_by"
+    CASE_WHEN = "case_when"
+    LITERAL = "literal"
+    COLUMN = "column"
+    FUNCTION = "function"
+    IN = "in"
+    LOW_IMPACT = "low_impact"
```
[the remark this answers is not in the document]
sfc-gh-aalam: so you want to have both "LOW_IMPACT" and "OTHERS"?

src/snowflake/snowpark/_internal/analyzer/complexity_stat.py
```diff
@@ -0,0 +1,24 @@
...
+    SET_OPERATION = "set_operation"  # UNION, EXCEPT, INTERSECT, UNION ALL
+    SAMPLE = "sample"
+    PIVOT = "pivot"
+    UNPIVOT = "unpivot"
+    WINDOW = "window"
+    GROUP_BY = "group_by"
+    PARTITION_BY = "partition_by"
+    CASE_WHEN = "case_when"
+    LITERAL = "literal"
+    COLUMN = "column"
```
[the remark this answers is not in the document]
sfc-gh-aalam: done

src/snowflake/snowpark/_internal/analyzer/expression.py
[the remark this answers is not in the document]
sfc-gh-aalam: added a simple unit test to show this

tests/integ/scala/test_async_job_suite.py
```diff
@@ -349,13 +351,21 @@ def test_async_batch_insert(session):
     reason="TODO(SNOW-932722): Cancel query is not allowed in stored proc",
 )
 def test_async_is_running_and_cancel(session):
-    async_job = session.sql("select SYSTEM$WAIT(3)").collect_nowait()
+    # creating a sproc here because describe query on SYSTEM$WAIT()
```
sfc-gh-jdu, line 354: the commit is checked in accidentally?

sfc-gh-jdu (approved):

tests/integ/scala/test_async_job_suite.py
sfc-gh-aalam: no. this is another one of the case where describe query is launched. in this case, internal describe query on `SELECT$WAIT` actually blocks the query

sfc-gh-jdu: I see, but I mean should it be in another PR?

src/snowflake/snowpark/_internal/analyzer/expression.py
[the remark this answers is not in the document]
sfc-gh-yzou: i see that this is marked as cached property for expression, but not for other node. However, isn't expression also part of the logical plan tree. Theoretically, we should be able to manipulate the tree at whatever level, right?

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
```diff
@@ -187,3 +194,23 @@ def __init__(
...
+            self.join_condition.cumulative_complexity_stat
+            if self.join_condition
+            else Counter()
+        )
+        stat += (
+            self.match_condition.cumulative_complexity_stat
+            if self.match_condition
+            else Counter()
+        )
+        return stat
```
sfc-gh-yzou, line 216: why join node doesn't have an accumulated complexity here?

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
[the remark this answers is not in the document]
sfc-gh-yzou: i think I am still confused here. The join seems a LogicalPlan node, if we do not plan to do it at the plan node level,  where do we plan to calculate the accumulated complexity for join then?

tests/integ/scala/test_async_job_suite.py
sfc-gh-yzou: + 1 this change seems irrelevant to this pr

src/snowflake/snowpark/_internal/analyzer/binary_expression.py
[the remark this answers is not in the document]
sfc-gh-yzou: hmm, if map1 + map2 is the only difference, can we simply use map then? Right now the returned type is Counter[str] which is not straight forward to understand with a str, for example, why complexity is represented as a string? to reduce unnecessary confusion in the code, let's use more obvious data structure for representation

src/snowflake/snowpark/_internal/analyzer/complexity_stat.py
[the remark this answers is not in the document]
sfc-gh-yzou: i think we can have both, then we have a category for each node in the plan tree node, it would be easy for us to calculate the total number of nodes in the tree

sfc-gh-yzou:
@sfc-gh-aalam thanks for accommodate the comments. the overall flow looks good to me, but i think i still have some major questions that i am kind of confused? especially the individual_complexity seems not so useful to me, maybe let's go over the major questions on Monday's meeting very quickly

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
sfc-gh-aalam: addressed offline

tests/integ/scala/test_async_job_suite.py
sfc-gh-aalam: I will revert this change and fix SelectSQL implementation so we do not trigger analyze attributes for it.

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
```diff
@@ -187,3 +194,23 @@ def __init__(
     @property
     def sql(self) -> str:
         return self.join_type.sql
+
+    @property
+    def individual_complexity_stat(self) -> Counter[str]:
```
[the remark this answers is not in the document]
sfc-gh-aalam: I am renaming this to `individual_node_complexity` and `cumulative_node_complexity`. since the return value is a dictionary, adding `score` could also be misleading.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-yzou: @sfc-gh-aalam i think @sfc-gh-jdu got a good point here, and i recall we typically just use variable binding for this, and if data is coming from client side, i assume it could be small, maybe we can simply count the column, and have a VALUE category, and set 1 for value

src/snowflake/snowpark/_internal/analyzer/expression.py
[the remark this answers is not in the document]
sfc-gh-aalam: updated so expression and logical node plan complexity can both be updated

src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -83,6 +88,39 @@ def sql(self) -> str:
     def __str__(self) -> str:
         return self.pretty_name
 
+    @property
+    def plan_node_category(self) -> PlanNodeCategory:
+        return PlanNodeCategory.OTHERS
+
+    @property
+    def individual_node_complexity(self) -> Dict[str, int]:
+        """Returns the individual contribution of the expression node towards the overall
```
sfc-gh-yzou, line 97: so what have been little bit confusing is how you overwrite the individual_node_complexity and  cumulative_node_complexity, for example, in join, i see you overwritten the individual_node_complexity where you were also counting the complexity of the join_cond etc. However, when i look at star expression it seems we overwritten the cumulative_node_complexity by accumulating all expressions.

So this leaves the definition of those two function kind of vague, how about this, let's say individual_node_complexity is the complexity of all nodes except the children. the cumulative_node_complexity for Expression class is the sum of individual complexity and all children's complexity, idealy all class that extends the Expression just need to overwrite the individual complexity correctly. Also let's leave a comment here to make it clear.
Or you can define it in the opposite way, where individual complexity simply represent the complexity of the current node, like what kind of node category it belongs to. in other words, the join should also overwrite the accumulative complexity. 
Either way on your preference, but let's make it consist, and make the function definition clear with comment

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -39,12 +71,28 @@ def __init__(self, start: int, end: int, step: int, num_slices: int = 1) -> None
...
 
+    @property
+    def individual_node_complexity(self) -> Dict[str, int]:
+        # SELECT ( ROW_NUMBER()  OVER ( ORDER BY  SEQ8() ) -  1 ) * (step) + (start) AS id FROM ( TABLE (GENERATOR(ROWCOUNT => count)))
+        return {
+            PlanNodeCategory.WINDOW.value: 1,
+            PlanNodeCategory.ORDER_BY.value: 1,
+            PlanNodeCategory.LITERAL.value: 3,  # step, start, count
+            PlanNodeCategory.COLUMN.value: 1,  # id column
+            PlanNodeCategory.LOW_IMPACT.value: 2,  # ROW_NUMBER, GENERATOR
```
sfc-gh-yzou, line 82: shoudn't row_number and generator counted as function? since those are snowflake functions

src/snowflake/snowpark/_internal/analyzer/unary_plan_node.py
```diff
@@ -89,18 +188,50 @@ def __init__(
         super().__init__(child)
         self.column_map = column_map
 
+    @property
+    def individual_node_complexity(self) -> Dict[str, int]:
+        # SELECT * RENAME (before AS after, ...) FROM child
+        return {
+            PlanNodeCategory.COLUMN.value: 1 + 2 * len(self.column_map),
```
sfc-gh-yzou, line 195: why it has a 2* here? shouldn't before as after counted as 1 column?

src/snowflake/snowpark/_internal/analyzer/unary_plan_node.py
```diff
@@ -89,18 +188,50 @@ def __init__(
...
+            PlanNodeCategory.LOW_IMPACT.value: 1 + len(self.column_map),
+        }
```
sfc-gh-yzou, line 197: what is the 1 value for is that for the alias name?

tests/integ/test_query_plan_analysis.py
```diff
@@ -0,0 +1,543 @@
...
+    df3 = df.agg(avg("a"), avg(col("b") + lit(1)).as_("avg_b"))
+    df4 = df.group_by(["a", "b"]).agg(avg("c"))
+
+    #  SELECT avg("A") AS "AVG(A)" FROM ( SELECT  *  FROM sample_table) LIMIT 1
+    assert_df_subtree_query_complexity(
+        df1,
+        {
+            PlanNodeCategory.COLUMN.value: 3,
+            PlanNodeCategory.LOW_IMPACT.value: 1,
+            PlanNodeCategory.FUNCTION.value: 1,
```
sfc-gh-yzou, line 209: oh, where does the function comes from? is that from aliasing function?

src/snowflake/snowpark/_internal/analyzer/unary_plan_node.py
sfc-gh-aalam: for each col1 AS col2, we are counting complexity as col1.complexity + col2.complexity. Which makes 2* len(column_map) in this case

tests/integ/test_query_plan_analysis.py
sfc-gh-aalam: it comes from AVG

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-aalam: yep, we should. I'll fix it

src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -83,6 +88,39 @@ def sql(self) -> str:
     def __str__(self) -> str:
         return self.pretty_name
 
+    @property
+    def plan_node_category(self) -> PlanNodeCategory:
+        return PlanNodeCategory.OTHERS
```
sfc-gh-aalam, line 93: this should be join

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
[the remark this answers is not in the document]
sfc-gh-aalam: I earlier had stat as a short hand for statistic

src/snowflake/snowpark/_internal/analyzer/expression.py
sfc-gh-aalam: done

src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -83,6 +88,34 @@ def sql(self) -> str:
...
+    def individual_node_complexity(self) -> Dict[PlanNodeCategory, int]:
+        """Returns the individual contribution of the expression node towards the overall
+        compilation complexity of the generated sql.
+        """
+        return {self.plan_node_category: 1}
+
+    @property
+    def cumulative_node_complexity(self) -> Dict[PlanNodeCategory, int]:
+        """Returns the aggregate sum complexity statistic from the subtree rooted at this
+        expression node. Statistic of current node is included in the final aggregate.
```
sfc-gh-yzou, line 105: add a comment here to mention that the node complexity is the sum of its individual complexity and children complicity, please make sure overwrite the individual_node_complexity properly for new nodes to get the cumulative complexity correct.

sfc-gh-yzou (approved):

src/snowflake/snowpark/_internal/analyzer/expression.py
sfc-gh-aalam: added

PR #1670, opened 2024-05-23
SNOW-1445536: Remove StringMethods and DatetimeProperties

sfc-gh-joshi:

   

   Fixes SNOW-1445536

This PR removes the StringMethods and DatetimeProperties classes and pushes appropriate NotImplementedErrors to the query compiler. The only class remaining in series_utils.py is CategoryMethods, which we leave behind since some methods in the modin implementation do not use the query compiler. We could remove this by overriding `Series.cat` to error out immediately instead of returning the `CategoryMethods` proxy object.


src/snowflake/snowpark/modin/pandas/series_utils.py
```diff
@@ -101,572 +78,3 @@ def as_ordered(self, inplace=False):
...
-    @property
-    def daysinmonth(self):
-        return Series(query_compiler=self._query_compiler.dt_property("daysinmonth"))
-
-    @property
-    def days_in_month(self):
-        return Series(query_compiler=self._query_compiler.dt_property("days_in_month"))
-
-    @property
-    def tz(self) -> "tzinfo | None":
```
sfc-gh-vbudati, line 578: Some of the methods like `tz` here have an implementation - why not keep the implementation? Also if you are removing it, can you also update the documentation/supported methods rst files? Thanks!

sfc-gh-joshi: We still support the method, we're just using OSS modin's `DatetimeProperties` class for the frontend instead. Unlike `StringMethods`, our existing non-implementation for most datetime methods error at the query compiler layer instead of the frontend.

https://github.com/modin-project/modin/blob/31771d7bb9d1dc36fb26ec49dc7798ccfea366b6/modin/pandas/series_utils.py#L696C5-L696C11

sfc-gh-vbudati: Ah I see, that makes sense. Thanks!

sfc-gh-vbudati (approved):

sfc-gh-dpetersohn (approved):
🎉

sfc-gh-joshi:
It turns out I had to add some extra boilerplate to get the Modin classes to use the Snowpark pandas Series instead of the OSS version, and I also had to stub dt_* methods in the query compiler to get the proper error message (I must've missed this in CI prior).

You can see the changes in the last commit: https://github.com/snowflakedb/snowpark-python/commit/02b9e488c8a0d21e84bb4bc52984eb642ee7f241

PR #1671, opened 2024-05-24
SNOW-1445178: Bypass SQL Counter for stored sproc regression tests

sfc-gh-lmukhopadhyay:

   

   Fixes SNOW-1445178

### DIFF
--- a/tests/integ/modin/sql_counter.py
+++ b/tests/integ/modin/sql_counter.py
@@ -16,6 +16,7 @@ from pandas._typing import Scalar
 
 from snowflake.snowpark import QueryRecord
 from snowflake.snowpark.session import Session
+from tests.utils import IS_IN_STORED_PROC
 
 UPDATED_SUFFIX = "updated"
 ORIGINAL_SUFFIX = "original"
@@ -363,6 +364,9 @@ def sql_count_checker(
         high_count_expected=high_count_expected,
         high_count_reason=high_count_reason,
     )
+
+    if IS_IN_STORED_PROC:
+        return
 
     all_args = inspect.getargvalues(inspect.currentframe())
     count_kwargs = {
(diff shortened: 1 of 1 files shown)


sfc-gh-nkrishna (commented):
I'm a bit confused by this change, can you provide some more context @sfc-gh-lmukhopadhyay?

sfc-gh-mvashishtha (approved):
@sfc-gh-nkrishna there's an explanation [here](https://github.com/snowflakedb/snowpark-python/blob/51951f03c801e317e350f45b3a8a69fc30235c1b/tests/integ/modin/sql_counter.py#L368-L370).

We are trying to run the test suite with Snowpark pandas in a stored procedure to check that Snowpark pandas works in stored procedures.

sfc-gh-lmukhopadhyay:
> I'm a bit confused by this change, can you provide some more context @sfc-gh-lmukhopadhyay?

Yes sure, in [sql_counter.py](https://github.com/snowflakedb/snowpark-python/blob/main/tests/integ/modin/sql_counter.py#L216) we use pytest-assume but that package isn't available in conda. So when running modin integration tests in stored procedures with the Snowfort regression framework, I run into the error `Package 'pytest-assume' is not supported`. To get around this, I'm bypassing the SQL counter check if a test is in a stored proc instead of removing our dependency on pytest-assume.

tests/integ/modin/sql_counter.py
```diff
@@ -364,6 +365,12 @@ def sql_count_checker(
         high_count_reason=high_count_reason,
     )
 
+    # Bypassing sql counter since
+    #   1. it is an unnecessary metric for tests running in stored proc
```
sfc-gh-yzou, line 369: i think you can even move this above, at the top of this call. Have you manually verified that works ?

sfc-gh-lmukhopadhyay: Yes, I tested this with integ test `tests/integ/modin/frame/test_head_tail.py` which ran successfully. I'll update the Snowfort POC with this test after this change

sfc-gh-yzou (approved):

PR #1674, opened 2024-05-24
Basic select API working

sfc-gh-heshah:
- `Column`s are created with no reference to the parent `Dataframe` in the current Snowpark API
- Following this pattern when generating the AST (invoking `proto` IR entity constructors directly)
- `ast` parameter added to `Column` constructor to enable `Column` methods to create and push-down ASTs
- `ast` property added to `Column` class to enable backwards compatibility with `col._expression: Expression` using `SpColumnSqlExpr` as a backup

Current test case output AST
```
client_version {
  major: 42
}
client_language {
  python_language {
    version {
      major: 3
      minor: 8
    }
  }
}
body {
  assign {
    uid: 1
    var_id {
      bitfield1: 1
    }
    expr {
      sp_table {
        table: "test_table"
      }
    }
  }
}
body {
  assign {
    uid: 2
    var_id {
      bitfield1: 2
    }
    expr {
      sp_dataframe_filter {
        df {
          sp_dataframe_ref {
            id {
              bitfield1: 1
            }
          }
        }
        condition {
          sp_column_sql_expr {
            sql: "STR LIKE \'%e%\'"
          }
        }
      }
    }
  }
}
body {
  assign {
    uid: 3
    var_id {
      bitfield1: 3
    }
    expr {
      sp_dataframe_select__columns {
        df {
          sp_dataframe_ref {
            id {
              bitfield1: 2
            }
          }
        }
        cols {
          sp_column {
            name: "\"A\""
          }
        }
        cols {
          sp_column_sql_expr {
            sql: "\"STR\""
          }
        }
        cols {
          sp_column {
            name: "\"B\""
          }
        }
        cols {
          sp_column_sql_expr {
            sql: "\"STR\""
          }
        }
        cols {
          sp_column_sql_expr {
            sql: "ADD(\"A\", \"B\")"
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -79,6 +79,7 @@ from snowflake.snowpark.types import (
     TimestampType,
 )
 from snowflake.snowpark.window import Window, WindowSpec
+import snowflake.snowpark._internal.proto.ast_pb2 as proto
 
 # Python 3.8 needs to use typing.Iterable because collections.abc.Iterable is not subscriptable
 # Python 3.9 can use both
@@ -226,16 +227,22 @@ class Column:
     """
 
     def __init__(
-        self, expr1: Union[str, Expression], expr2: Optional[str] = None
+        self, expr1: Union[str, Expression], expr2: Optional[str] = None, ast: Optional[proto.SpColumnExpr] = None
     ) -> None:
+        self._ast = proto.SpColumnExpr() if ast is None else ast
+
         if expr2 is not None:
             if isinstance(expr1, str) and isinstance(expr2, str):
                 if expr2 == "*":
                     self._expression = Star([], df_alias=expr1)
                 else:
+                    col_name = quote_name(expr2)
                     self._expression = UnresolvedAttribute(
-                        quote_name(expr2), df_alias=expr1
+                        col_name, df_alias=expr1
                     )
+                    # TODO: figure out why basic aliasing is breaking above
+                    self._ast.sp_column_alias.col.sp_column.name = col_name
+                    self._ast.sp_column_alias.name = expr1
             else:
                 raise ValueError(
                     "When Column constructor gets two arguments, both need to be <str>"
@@ -244,11 +251,21 @@ class Column:
             if expr1 == "*":
                 self._expression = Star([])
             else:
-                self._expression = UnresolvedAttribute(quote_name(expr1))
+                col_name = quote_name(expr1)
+                self._expression = UnresolvedAttribute(col_name)
+                self._ast.sp_column.name = col_name
+
         elif isinstance(expr1, Expression):
             self._expression = expr1
         else:  # pragma: no cover
             raise TypeError("Column constructor only accepts str or expression.")
+
+    @property
+    def ast(self) -> proto.SpColumnExpr:
+        # TODO: remove hack to generate SQL from Expression class
+        if self._ast.WhichOneof("variant") is None:
+            self._ast.sp_column_sql_expr.sql = self._expression.sql
+        return self._ast
 
     def __getitem__(self, field: Union[str, int]) -> "Column":
         """Accesses an element of ARRAY column by ordinal position, or an element of OBJECT column by key."""
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -7,6 +7,7 @@ import sys
 from typing import Optional, Union
 
 import snowflake.snowpark
+import snowflake.snowpark._internal.proto.ast_pb2 as proto
 from snowflake.snowpark._internal.analyzer.binary_expression import (
     Add,
     And,
@@ -79,7 +80,6 @@ from snowflake.snowpark.types import (
     TimestampType,
 )
 from snowflake.snowpark.window import Window, WindowSpec
-import snowflake.snowpark._internal.proto.ast_pb2 as proto
 
 # Python 3.8 needs to use typing.Iterable because collections.abc.Iterable is not subscriptable
 # Python 3.9 can use both
@@ -227,7 +227,10 @@ class Column:
     """
 
     def __init__(
-        self, expr1: Union[str, Expression], expr2: Optional[str] = None, ast: Optional[proto.SpColumnExpr] = None
+        self,
+        expr1: Union[str, Expression],
+        expr2: Optional[str] = None,
+        ast: Optional[proto.SpColumnExpr] = None,
     ) -> None:
         self._ast = proto.SpColumnExpr() if ast is None else ast
 
@@ -237,9 +240,7 @@ class Column:
                     self._expression = Star([], df_alias=expr1)
                 else:
                     col_name = quote_name(expr2)
-                    self._expression = UnresolvedAttribute(
-                        col_name, df_alias=expr1
-                    )
+                    self._expression = UnresolvedAttribute(col_name, df_alias=expr1)
                     # TODO: figure out why basic aliasing is breaking above
                     self._ast.sp_column_alias.col.sp_column.name = col_name
                     self._ast.sp_column_alias.name = expr1
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -232,25 +232,29 @@ class Column:
         expr2: Optional[str] = None,
         ast: Optional[proto.SpColumnExpr] = None,
     ) -> None:
-        self._ast = proto.SpColumnExpr() if ast is None else ast
+        if ast is not None:
+            self._ast = ast
 
         if expr2 is not None:
+            # TODO: Find correct entities in AST for df_aliasing
             if isinstance(expr1, str) and isinstance(expr2, str):
                 if expr2 == "*":
                     self._expression = Star([], df_alias=expr1)
                 else:
-                    col_name = quote_name(expr2)
-                    self._expression = UnresolvedAttribute(col_name, df_alias=expr1)
-                    # TODO: figure out why basic aliasing is breaking above
-                    self._ast.sp_column_alias.col.sp_column.name = col_name
-                    self._ast.sp_column_alias.name = expr1
+                    self._expression = UnresolvedAttribute(
+                        quote_name(expr2), df_alias=expr1
+                    )
             else:
                 raise ValueError(
                     "When Column constructor gets two arguments, both need to be <str>"
                 )
         elif isinstance(expr1, str):
+            if ast is None:
+                self._ast = proto.SpColumnExpr()
             if expr1 == "*":
                 self._expression = Star([])
+                # TODO: Determine whether SpColumnSqlExpr is correct entity here
+                self._ast.sp_column_sql_expr.sql = "*"
             else:
                 col_name = quote_name(expr1)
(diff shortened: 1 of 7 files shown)


tests/thin-client/steel-thread.py
```diff
@@ -2,6 +2,7 @@
 from snowflake.snowpark.functions import col, contains
 
 session = Session.builder.create()
-df = session.table('test_table')
-df = df.filter("STR LIKE '%e%'")
-df.show()
+df = session.table('test_table').filter("STR LIKE '%e%'")
```
sfc-gh-azwiegincew, line 5: Maybe add this as a separate "test" or use cmdline args to choose the test.

src/snowflake/snowpark/dataframe.py
```diff
@@ -517,6 +517,7 @@ def __init__(
                              referenced in subsequent dataframe expressions.
         """
         self._session = session
+        self._ast_expr = ast_stmt.expr
```
sfc-gh-azwiegincew, line 520: There's no reason to retain the AST expression.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1125,6 +1126,11 @@ def select(
         exprs = parse_positional_args_to_list(*cols)
         if not exprs:
             raise ValueError("The input of select() cannot be empty")
+        
+        stmt = self._session._ast_batch.assign()
+        ast = stmt.expr
+        ast.sp_dataframe_select__columns.df.sp_dataframe_ref.id.bitfield1 = self._ast_id
+        ast.sp_dataframe_select__columns.variadic = False
```
sfc-gh-azwiegincew, line 1133: `variadic` is true for some forms – the logic seems to be in `parse_positional_args_to_list`. We have to capture this detail to correctly represent the unparsed form in query history.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1133,8 +1139,11 @@ def select(
         for e in exprs:
             if isinstance(e, Column):
                 names.append(e._named())
+                ast.sp_dataframe_select__columns.cols.append(e.ast)
             elif isinstance(e, str):
-                names.append(Column(e)._named())
+                col = Column(e)
+                names.append(col._named())
+                ast.sp_dataframe_select__columns.cols.append(col.ast)
```
sfc-gh-azwiegincew, line 1146: This is inefficient. Since the user passes in a string in this case, you're free to construct a new column expr from scratch without the inefficient copying.

This will naturally get refactored due to the previous comment though.

If this were to result in cleaner code, feel free to reimplement the logic, since in Phase 1 we will remove a lot of the existing complexity (we're preserving it for now so that this codebase can be used for Phase 0).

src/snowflake/snowpark/column.py
```diff
@@ -226,16 +227,22 @@ class Column:
...
             if isinstance(expr1, str) and isinstance(expr2, str):
                 if expr2 == "*":
                     self._expression = Star([], df_alias=expr1)
                 else:
+                    col_name = quote_name(expr2)
                     self._expression = UnresolvedAttribute(
-                        quote_name(expr2), df_alias=expr1
+                        col_name, df_alias=expr1
                     )
+                    # TODO: figure out why basic aliasing is breaking above
```
sfc-gh-azwiegincew, line 243: What's the problem?

src/snowflake/snowpark/column.py
```diff
@@ -226,16 +227,22 @@ class Column:
     """
 
     def __init__(
-        self, expr1: Union[str, Expression], expr2: Optional[str] = None
+        self, expr1: Union[str, Expression], expr2: Optional[str] = None, ast: Optional[proto.SpColumnExpr] = None
     ) -> None:
+        self._ast = proto.SpColumnExpr() if ast is None else ast
```
sfc-gh-azwiegincew, line 232: This logic seems wrong. You should defer the construction of the AST to the specific cases below.

src/snowflake/snowpark/column.py
```diff
@@ -244,12 +251,22 @@ def __init__(
...
+                self._expression = UnresolvedAttribute(col_name)
+                self._ast.sp_column.name = col_name
+
         elif isinstance(expr1, Expression):
             self._expression = expr1
         else:  # pragma: no cover
             raise TypeError("Column constructor only accepts str or expression.")
 
+    @property
+    def ast(self) -> proto.SpColumnExpr:
```
sfc-gh-azwiegincew, line 264: This shouldn't be necessary. This logic should happen at construction time for the appropriate case. We shouldn't be introducing imperative mutations with a hard-to-follow control flow in an API without mutable state like this one. This is why I'm suggesting having clear logic in the constructor that immediately constructs the required value without mutations spread across multiple branches and functions.

tests/thin-client/steel-thread.py
```diff
@@ -2,6 +2,7 @@
...
+# col = df.col("STR") # TODO: determine whether this should create and assign statement or not
+df2 = df.select("A", df.col("STR"), col("B"), df.STR, col("A") + col("B"))
+print(session._ast_batch._request)
```
sfc-gh-azwiegincew, line 8: Could you attach the output to the PR description please? I wanted to double-check the result.

sfc-gh-azwiegincew (changes requested):

src/snowflake/snowpark/dataframe.py
sfc-gh-azwiegincew: More info at https://protobuf.dev/reference/python/python-generated/#repeated-message-fields

src/snowflake/snowpark/column.py
sfc-gh-heshah: Would it be okay to still store a constructed `_ast` member and retrieve it later?

I'm facing an issue with deciding how to handle expressions of the form `df.select(df.col("A"), col("B"))`, where the provided `Column` instances are constructed before `df.select` has an opportunity to construct the necessary `SpColumnExpr` instance and populate the AST fields.

More generally any Dataframe API that accepts previously constructed Column instances will face this issue, hence the current solution. I agree with the point you made for line 1137 though as that is a clear opportunity to avoid copying from a potentially complex message into the Session AstBatch.

sfc-gh-azwiegincew: Yes, I think you're right and we have to store a pre-constructed value in `_ast`. This property shouldn't be necessary though. Maybe a helper function to copy `_ast` into whatever builder is passed in?

sfc-gh-heshah: Sounds good, would the goal of the helper function be to replace `protobuf.message.CopyFrom(col._ast)` and `append(col._ast)` when constructing the AST one level up from the constructor?

I was thinking as long as `_ast` is ensured to be an instance of `SpColumnExpr` a helper function might not be necessary, unless there is another purpose I'm missing, please let me know!

sfc-gh-azwiegincew: Yes. Definitely better without a helper if the code seems simple.

sfc-gh-heshah:
AST output from `tests/thin-client/select_outputs.py` after changes:
```
client_version {
  major: 42
}
client_language {
  python_language {
    version {
      major: 3
      minor: 8
    }
  }
}
body {
  assign {
    uid: 1
    var_id {
      bitfield1: 1
    }
    expr {
      sp_table {
        table: "test_table"
      }
    }
  }
}
body {
  assign {
    uid: 2
    var_id {
      bitfield1: 2
    }
    expr {
      sp_dataframe_filter {
        df {
          sp_dataframe_ref {
            id {
              bitfield1: 1
            }
          }
        }
        condition {
          sp_column_sql_expr {
            sql: "STR LIKE \'%e%\'"
          }
        }
      }
    }
  }
}
body {
  assign {
    uid: 3
    var_id {
      bitfield1: 3
    }
    expr {
      sp_dataframe_select__columns {
        df {
          sp_dataframe_ref {
            id {
              bitfield1: 2
            }
          }
        }
        cols {
          sp_column {
            name: "\"A\""
          }
        }
        cols {
          sp_dataframe_col {
            df {
              sp_dataframe_ref {
                id {
                  bitfield1: 2
                }
              }
            }
            col_name: "\"STR\""
          }
        }
        cols {
          sp_column {
            name: "\"B\""
          }
        }
        cols {
          sp_dataframe_col {
            df {
              sp_dataframe_ref {
                id {
                  bitfield1: 2
                }
              }
            }
            col_name: "\"STR\""
          }
        }
        variadic: true
      }
    }
  }
}
```

src/snowflake/snowpark/column.py
```diff
@@ -226,9 +227,16 @@ class Column:
     """
 
     def __init__(
-        self, expr1: Union[str, Expression], expr2: Optional[str] = None
+        self,
+        expr1: Union[str, Expression],
+        expr2: Optional[str] = None,
+        ast: Optional[proto.SpColumnExpr] = None,
     ) -> None:
+        if ast is not None:
```
sfc-gh-azwiegincew, line 235: This should be done without a guard. It's better to rely on a known slot being set to `None` rather than the absence of the key in the dictionary.

src/snowflake/snowpark/column.py
```diff
@@ -241,10 +249,17 @@ def __init__(
                     "When Column constructor gets two arguments, both need to be <str>"
                 )
         elif isinstance(expr1, str):
+            if ast is None:
+                self._ast = proto.SpColumnExpr()
             if expr1 == "*":
                 self._expression = Star([])
+                # TODO: Determine whether SpColumnSqlExpr is correct entity here
```
sfc-gh-azwiegincew, line 256: This seems correct.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1126,15 +1131,27 @@ def select(
         if not exprs:
             raise ValueError("The input of select() cannot be empty")
 
+        stmt = self._session._ast_batch.assign()
+        ast = stmt.expr
+        ast.sp_dataframe_select__columns.df.sp_dataframe_ref.id.bitfield1 = self._ast_id
+        ast.sp_dataframe_select__columns.variadic = (len(cols) > 1 or isinstance(cols[0], (list, tuple, set)))
```
sfc-gh-azwiegincew, line 1137: This is incorrect. If a single list is passed, we want `variadic = False`.

sfc-gh-azwiegincew (changes requested):

src/snowflake/snowpark/column.py
```diff
@@ -249,16 +248,18 @@ def __init__(
...
                 self._expression = Star([])
-                # TODO: Determine whether SpColumnSqlExpr is correct entity here
-                self._ast.sp_column_sql_expr.sql = "*"
             else:
-                col_name = quote_name(expr1)
-                self._expression = UnresolvedAttribute(col_name)
-                self._ast.sp_column.name = col_name
+                self._expression = UnresolvedAttribute(quote_name(expr1))
+
+            # some repitition here, but _expression logic will be eliminated eventually
```
sfc-gh-azwiegincew, line 256: Repetition

sfc-gh-azwiegincew (approved):
LG!

PR #1679, opened 2024-05-24
SNOW-1370035: [Local Testing] Improve consistency of udf input/output

sfc-gh-jrose:

   

   Fixes SNOW-1370035 and SNOW-1374027

   This PR improves local testing udf experience in a few ways when it comes to variant data.

- Variant data can contain just about anything, but often when passed to a udf it is coerced to certain types. I've added a mechanism to the udf call logic that coerces various types to the expected types.
- Variant Nones are passed to udfs as SqlNullWrapper objects. These are not currently defined or used in our repo so I've added a coercion for them in this PR as well.
- I've merged the logic for calculating udf inputs with the logic that validates input types. This resulted in a different error experience that could be considered a BCR, but the replaced error hasn't been released yet so I don't think it's a problem.
- UDFs that expect variant input, but receive other datatypes as input will have those datatypes coerced which matches the behaviour of live mode.

### DIFF
--- a/src/snowflake/snowpark/mock/_plan.py
+++ b/src/snowflake/snowpark/mock/_plan.py
@@ -39,7 +39,11 @@ from snowflake.snowpark._internal.analyzer.window_expression import (
     UnboundedPreceding,
     WindowExpression,
 )
-from snowflake.snowpark.mock._udf_utils import types_are_compatible
+from snowflake.snowpark.mock._udf_utils import (
+    coerce_variant_input,
+    remove_null_wrapper,
+    types_are_compatible,
+)
 from snowflake.snowpark.mock._util import get_fully_qualified_name
 from snowflake.snowpark.mock._window_utils import (
     EntireWindowIndexer,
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/mock/_udf_utils.py
```diff
@@ -11,6 +13,33 @@
...
+    if isinstance(value, SqlNullWrapper):
+        return None
+    return value
+
+
+def coerce_variant_input(value):
+    input_type = type(value)
+    if input_type in VARIANT_INPUT_MAPPING:
+        value = VARIANT_INPUT_MAPPING[input_type](value)
+    return value
```
sfc-gh-aling, line 41: I'm curious what about other types like list/dict, they are kept as the original value?

src/snowflake/snowpark/mock/_udf_utils.py
```diff
@@ -11,6 +13,33 @@
 )
 from snowflake.snowpark.types import NullType, _NumericType
 
+VARIANT_INPUT_MAPPING = {
+    bytes: lambda x: x.decode("utf-8"),
```
sfc-gh-stan, line 17: 🫨 bytes are automatically decoded into str when passed as a variant argument into UDFs?

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -536,24 +540,36 @@ def cleanup_imports():
...
-        actual_types = [col.datatype for col in function_input.sf_types.values()]
-        if len(actual_types) != len(udf._input_types):
-            raise SnowparkLocalTestingException(
-                f"Expected {len(udf._input_types)} arguments, but received {len(actual_types)}"
-            )
 
-        for expected, actual in zip(udf._input_types, actual_types):
-            if not types_are_compatible(expected, actual):
+            # SNOW-929218: Once proper type coercion is supported use that instead.
+            if isinstance(expected_type, VariantType) and not isinstance(
```
sfc-gh-aling, line 557: this is beyond the scope of this PR, does we support MapType/ArrayType in UDF and if so do they need to coerce like VariantType?

tests/integ/test_udf.py
```diff
@@ -1308,23 +1304,24 @@ def variant_get_data_type(v):
...
-            Row("<class 'float'>"),
-            Row("<class 'str'>"),
-            Row("<class 'bool'>"),
-            Row("<class 'list'>"),
-            Row("<class 'dict'>"),
-        ],
-    )
+    if not local_testing_mode:
+        # dynamic typing on one single column
+        df = session.sql(
```
sfc-gh-stan, line 1309: Can we rewrite this part into `df.select(parse_json(col1))`?

src/snowflake/snowpark/mock/_udf_utils.py
sfc-gh-jrose: That's what it looks like [here](https://github.com/snowflakedb/snowpark-python/blob/main/tests/integ/test_udf.py#L1246).

src/snowflake/snowpark/mock/_plan.py
sfc-gh-jrose: We do support them, but I don't think they need coercion in a similar way to Variant type. I don't think server side would try to implicitly cast an integer to a map in the same way it would an integer into a variant for example.

src/snowflake/snowpark/mock/_udf_utils.py
sfc-gh-aling: nvm I saw the test that `list` and `dict` are kept

sfc-gh-aling (approved):

sfc-gh-stan (approved):

PR #1685, opened 2024-05-28
SNOW-1437407: [Local Testing] Fix convert_timezone argument order

sfc-gh-jrose:

   

   Fixes SNOW-1437407

   The original mock implementation for convert_timezone assumed that args would be passed in the same order as they are provided to the non-mocked version. Upon further inspection I now see that the args are passed with sql semantics instead. I've updated the mock function to set local variable based on the sql semantics instead and added a test to show that it can be called with all three args.

I've also refactored most of the logic in the function to use pandas.DataFrame.apply instead of converting everything into a list and using an iterator.


src/snowflake/snowpark/mock/_functions.py
```diff
@@ -1700,42 +1700,53 @@ def mock_initcap(values: ColumnEmulator, delimiters: ColumnEmulator):
...
+            # Using dateutil because it uses iana timezones while pytz would use Olson tzdb.
+            source_time = source_time.replace(tzinfo=dateutil.tz.gettz(source_timezone))
+
+        if source_time.tzinfo is None:
+            source_time = LocalTimezone.replace_tz(source_time)
+
+        result = source_time.astimezone(dateutil.tz.gettz(target_timezone))
+
+        if return_type == TimestampTimeZone.NTZ:
+            result = result.replace(tzinfo=None)
```
sfc-gh-aling, line 1745: this logic seems new, is this also fixing a bug that previously we don't remove tzinfo TimestampNTZ type?

tests/integ/scala/test_function_suite.py
```diff
@@ -3835,6 +3835,52 @@ def test_convert_timezone(session, local_testing_mode):
...
+def test_convert_timezone_neg(session):
+    df = TestData.datetime_primitives1(session)
+    with pytest.raises(SnowparkSQLException):
+        df.select(
+            convert_timezone(lit("UTC"), "timestamp_tz", lit("US/Eastern"))
+        ).collect()
+
+
+def test_convert_timezone_nulls(session):
+    null_df = session.create_dataframe([[None]]).to_df("timestamp")
```
sfc-gh-aling, line 3847: dumb question, do we have test for non-None two args input for convert_timezone?

sfc-gh-jrose: Yes, but I've left that one unchanged.

src/snowflake/snowpark/mock/_functions.py
sfc-gh-jrose: That is correct.

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -1700,42 +1700,53 @@ def mock_initcap(values: ColumnEmulator, delimiters: ColumnEmulator):
 
 @patch("convert_timezone")
 def mock_convert_timezone(
-    target_timezone: ColumnEmulator,
-    source_time: ColumnEmulator,
-    source_timezone: Optional[ColumnEmulator] = None,
+    *args: ColumnEmulator,
```
sfc-gh-stan, line 1703: Question: what's the motivation behind this change?

sfc-gh-jrose: Excerpt from functions.py:
```
    if source_timezone is None:
        return call_builtin("convert_timezone", target_tz, source_time_to_convert)
    return call_builtin(
        "convert_timezone", source_tz, target_tz, source_time_to_convert
    )
```
When convert_timezone is called with 2 parameters they are in a certain order, but when called with 3 the additional parameter is added to the front instead. In practice it seems that the mocked version always gets three parameters, but the ordering is different so I don't think it makes as much sense to name those parameters.

sfc-gh-stan (approved):

sfc-gh-aling (approved):

PR #1696, opened 2024-05-29
Column ops implementation first draft

sfc-gh-heshah:
Added draft implementations for most column expressions (excluding those without corresponding AST entities, ISIN, OVER, and string expression based functions missing IR entities).


sfc-gh-oplaton (commented):
Superficially, the changes look good, with one observation: The functionality in `src/snowflake/snowpark/column.py` is very repetitive, to the point where I probably would have missed anything broken.

I think it would make sense for at least some of the functionality (e.g. binary operators) to be further encapsulated into a higher-order function.

For example, `__eq__` and `__ne__` could become:

```python
def __eq__(self, other):
    return self.bin_op_impl(lambda ast: ast.sp_column_equal_to, EqualTo, other)

def __ne__(self, other):
    return self.bin_op_impl(lambda ast: ast.sp_column_not_equal, NotEqualTo, other)

def bin_op_impl(self, property, comparator, other):
    ast = proto.SpColumnExpr()
    property(ast).lhs.CopyFrom(self._ast)
    Column._fill_ast(property(ast).rhs, other)
    return Column(comparator(self._expression, Column._to_expr(other)), ast = ast)
```

sfc-gh-oplaton:
I think the change looks good. Thanks for the updated approach! Of course, I'd recommend getting additional input from Leonhard and/or Arthur, as I might have missed some of the finer points of the design.

sfc-gh-oplaton (approved):

sfc-gh-heshah:
Sounds good, thank you for the suggestion! 

Originally I was thinking the change you mentioned should correspond to further abstracting the IR for `SpColumnExpr` operations as Arthur suggested on Slack, but I think this minimizes the code changes required if the AST definitions are changed.

Along those same lines I though there were other repetitive AST building patterns that could be abstracted into a static `Column._create_ast` method which I added. This should further reduce dependence on the exact protobuf definition, and minimize changes necessary in the code in case the AST is evolved further, but please let me know if its a bit overboard.

src/snowflake/snowpark/column.py
```diff
@@ -833,6 +881,22 @@ def _named(self) -> NamedExpression:
...
+    @staticmethod
+    def _create_ast(property: Optional[Callable] = None,
+                    assign_fields: Dict[str, Any] = {},
+                    copy_messages: Dict[str, Any] = {},
+                    fill_col_asts: Dict[str, ColumnOrLiteral] = {}
+        ) -> proto.SpColumnExpr:
+        ast = proto.SpColumnExpr()
+        if property is not None:
+            for attr, value in assign_fields.items():
+                setattr(property(ast), attr, value)
```
sfc-gh-oplaton, line 893: This isn't critical, but I wanted to make it explicit:

The point of using a callable to access object properties (`lambda o: o.some_property`) was to avoid using `getattr`/`setattr`, but they're equivalent, and there's no strict technical reason why callables are superior to `getattr`/`setattr`, just a personal preference (it looks more like code than stringly-typed code; very superficial preference).

That said, the callable pattern is more verbose at the call site, especially when using it repeatedly.

In summary, I think it would make the code less readable to have both the callable pattern and `getattr`/`setattr`. Pick one that you prefer (I don't feel strongly either way - callable looks more like code, but `getattr`/`setattr` will result in terser code) and be uniform. Uniformity matters a lot for ease of maintenance.

sfc-gh-oplaton: Aside: I assume performance isn't a consideration here. There might be a difference between using a callable and `getattr`/`setattr`, but I didn't look into it, and I don't have an intuition here. If this is going to run billions of times, maybe writing a short benchmark and comparing would also help (if you can do it in <15 minutes, this doesn't need to become a side project of its own).

src/snowflake/snowpark/column.py
```diff
@@ -226,6 +228,7 @@ class Column:
     This class has methods for the most frequently used column transformations and operators. Module :mod:`snowflake.snowpark.functions` defines many functions to transform columns.
     """
 
+    # NOTE: For now assume Expression instances can be safely ignored when building AST
```
sfc-gh-lspiegelberg, line 231: can you add an example that's problematic here? Not sure what to do to resolve this note.

src/snowflake/snowpark/column.py
```diff
@@ -235,14 +238,27 @@ def __init__(
...
         if expr2 is not None:
-            # TODO: Find correct entities in AST for df_aliasing
             if isinstance(expr1, str) and isinstance(expr2, str):
                 if expr2 == "*":
                     self._expression = Star([], df_alias=expr1)
                 else:
                     self._expression = UnresolvedAttribute(
                         quote_name(expr2), df_alias=expr1
                     )
+
```
sfc-gh-lspiegelberg, line 248: What would the alias field here be? is this for `expr.as_(alias)`? Can we maybe add an example what would need to get captured here?

src/snowflake/snowpark/column.py
```diff
@@ -255,11 +271,16 @@ def __init__(
 
             # some repetition here, but _expression logic will be eliminated eventually
             if self._ast is None:
-                self._ast = proto.SpColumnExpr()
                 if expr1 == "*":
-                    self._ast.sp_column_sql_expr.sql = "*"
+                    self._ast = Column._create_ast(
```
sfc-gh-lspiegelberg, line 275: As soon as https://github.com/snowflakedb/snowpark-python/pull/1701 is in , can we add tests for this?

src/snowflake/snowpark/column.py
```diff
@@ -269,79 +290,106 @@ def __init__(
...
+                property = lambda ast: ast.sp_column_apply__int,
+                copy_messages = {"col": self._ast},
+                assign_fields = {"idx": field},
+            )
+            return Column(SubfieldInt(self._expression, field), ast = ast)
         else:
             raise TypeError(f"Unexpected item type: {type(field)}")
 
     # overload operators
+    def _bin_op_impl(self, property: Callable, operator: BinaryExpression, other: ColumnOrLiteral) -> "Column":
```
sfc-gh-lspiegelberg, line 310: pls. add docstrings to this

src/snowflake/snowpark/column.py
```diff
@@ -269,79 +290,106 @@ def __init__(
...
+            copy_messages = {"rhs": self._ast},
+            fill_col_asts = {"lhs": other}
+        )
+        return Column(operator(Column._to_expr(other), self._expression), ast = ast)
+    
     def __eq__(self, other: Union[ColumnOrLiteral, Expression]) -> "Column":
         """Equal to."""
-        right = Column._to_expr(other)
-        return Column(EqualTo(self._expression, right))
+        return self._bin_op_impl(lambda ast: ast.sp_column_equal_to, EqualTo, other)
```
sfc-gh-lspiegelberg, line 328: big fan of the refactor here, so much cleaner!

src/snowflake/snowpark/column.py
```diff
@@ -269,79 +290,106 @@ def __init__(
...
         """Divide."""
-        return Column(Divide(self._expression, Column._to_expr(other)))
+        return self._bin_op_impl(lambda ast: ast.sp_column_divide, Divide, other)
 
     def __rtruediv__(self, other: Union[ColumnOrLiteral, Expression]) -> "Column":
-        return Column(Divide(Column._to_expr(other), self._expression))
+        return self._bin_op_rimpl(lambda ast: ast.sp_column_divide, Divide, other)
 
     def __mod__(self, other: Union[ColumnOrLiteral, Expression]) -> "Column":
         """Reminder."""
```
sfc-gh-lspiegelberg, line 379: typo: Remainder

src/snowflake/snowpark/column.py
```diff
@@ -461,65 +510,91 @@ def between(
...
+        return self._bin_op_rimpl(lambda ast: ast.sp_column_or, Or, other)      # pragma: no cover
 
     def __invert__(self) -> "Column":
         """Unary not."""
+        # TODO: Need an SpColumnNot IR entity
+        # return self._unary_op_impl(lambda ast: ast.sp_column_not, Not)
         return Column(Not(self._expression))
 
     def _cast(self, to: Union[str, DataType], try_: bool = False) -> "Column":
+        # TODO: Update SpColumnCast IR entity with new field "try_", then uncomment
```
sfc-gh-lspiegelberg, line 587: how difficult would this be? I would be also to defer this work to another PR here. Ok, to have only simple column expr supported for now.

src/snowflake/snowpark/column.py
```diff
@@ -591,11 +699,18 @@ def regexp(self, pattern: ColumnOrLiteralStr) -> "Column":
         :meth:`rlike` is an alias of :meth:`regexp`.
 
         """
+        # TODO: Should SpColumnRegexp.pattern be an Expr in the IR, could sbe SpColumnExpr as in SpColumnLike?
```
sfc-gh-lspiegelberg, line 702: There seem to be a few TODOs which require IR changes, maybe collect in doc, and then go through in meeting? We prob. can resolve those quickly.

src/snowflake/snowpark/column.py
```diff
@@ -734,21 +892,65 @@ def within_group(
...
         """
-        return Column(
-            WithinGroup(
-                self._expression,
-                [
-                    _to_col_if_str(col, "within_group")._expression
-                    for col in parse_positional_args_to_list(*cols)
-                ],
-            )
+        prop = lambda ast: ast.sp_column_within_group
```
sfc-gh-lspiegelberg, line 895: The code here looks long - can we add a comment to explain the logic?

src/snowflake/snowpark/column.py
```diff
@@ -734,21 +892,65 @@ def within_group(
...
+        return Column(WithinGroup(self._expression, order_by_cols), ast = ast)
+
     def _named(self) -> NamedExpression:
         if isinstance(self._expression, NamedExpression):
             return self._expression
         else:
             return UnresolvedAlias(self._expression)
+    
+    @staticmethod
+    def _create_ast(property: Optional[Callable] = None,
```
sfc-gh-lspiegelberg, line 926: This looks like a super helpful function. Can we add a docstring to explain it? This seems to be used throughout - so a good idea to have a detailed docstring and also whichever asserts/checks are needed within it.

src/snowflake/snowpark/column.py
sfc-gh-lspiegelberg: I don't think we need to be concerned with performance here. I would pick whichever code is easier to read/less to write. Having this function here is super useful throughout the code base!

src/snowflake/snowpark/column.py
```diff
@@ -734,21 +892,65 @@ def within_group(
...
+            for attr, messg in copy_messages.items():
+                getattr(property(ast), attr).CopyFrom(messg)
+            for attr, other in fill_col_asts.items():
+                Column._fill_ast(getattr(property(ast), attr), other)
+        return ast
+    
+    @classmethod
+    def _fill_ast(cls, ast: proto.SpColumnExpr, value: ColumnOrLiteral) -> None:
+        """
+        Copy from a Column object's AST, or copy a literal value into an AST expression.
```
sfc-gh-lspiegelberg, line 944: please explain params, use Google Style (you can configure in PyCharm) to explain params.

src/snowflake/snowpark/types.py
```diff
@@ -40,12 +41,17 @@ def __repr__(self):
 
     def is_primitive(self):
         return True
-
+    
+    def fill_ast(self, ast: proto.SpDataType) -> None:
```
sfc-gh-lspiegelberg, line 45: docstring

src/snowflake/snowpark/types.py
```diff
@@ -40,12 +41,17 @@ def __repr__(self):
...
+        raise NotImplementedError(f"{self.__class__.__name__} does not have an equivalent SpDataType yet")
```
sfc-gh-lspiegelberg, line 46: this is a base method I assume? Maybe let's not use NotImplementedError, but instead ValueError and what to implement.

sfc-gh-lspiegelberg: you can also use https://docs.python.org/3/library/abc.html

src/snowflake/snowpark/types.py
```diff
@@ -441,27 +485,39 @@ def __setitem__(self, key, value):
...
-
-    pass
+    def fill_ast(self, ast: proto.SpDataType) -> None:
+        ast.sp_geography_type = True
 
 
 class GeometryType(DataType):
     """Geometry data type. This maps to the GEOMETRY data type in Snowflake."""
-
+    # TODO: Create SpGeometryType IR enttiy, then uncomment
```
sfc-gh-lspiegelberg, line 514: typo: enttiy -> entity

src/snowflake/snowpark/types.py
```diff
@@ -115,6 +121,11 @@ def __hash__(self):
         if self.length == StringType._MAX_LENGTH:
             return StringType().__hash__()
         return super().__hash__()
+        
+    def fill_ast(self, ast: proto.SpDataType) -> None:
+        # TODO: Create SpStringType IR entity with field "length: Int", and update
```
sfc-gh-lspiegelberg, line 126: instead of all the TODOs as comment, could also use NotImplementedError with a comment on what to do. You could also define an error message

`ErrorMessage.missing_ir('SpStringType IR needs field "length: Int"')` 

that would help a lot!

sfc-gh-lspiegelberg (commented):
Great work! Left a few comments, I think it would be nice to have some testing added here.

src/snowflake/snowpark/column.py
sfc-gh-oplaton: If you're sure performance won't matter, I recommend using `getattr`/`setattr` everywhere. Way fewer lexical tokens at each call site.

sfc-gh-heshah: Just wrote a short benchmark using `timeit`, wasn't too much work
```python
import timeit

setup = """
import snowflake.snowpark._internal.proto.ast_pb2 as proto
test = proto.SpColumnExpr()
prop = lambda ast: ast.sp_column_apply__string
"""

getcode1 = """
t = prop(test)
"""

getcode2 = """
t = getattr(test, "sp_column_apply__string")
"""

print("Callable", timeit.timeit(setup=setup, stmt=getcode1))
print("getattr", timeit.timeit(setup=setup, stmt=getcode2))

setcode1 = """
test.sp_column_apply__string.field = "test"
"""

setcode2 = """
setattr(test.sp_column_apply__string, "field", "test")
"""

print(".field direct", timeit.timeit(setup=setup, stmt=setcode1))
print("setattr", timeit.timeit(setup=setup, stmt=setcode2))
```

The results I got for 1 million runs of each line were:
```
Callable:      0.04975545800000003
getattr:       0.040397291999999974
.field direct: 0.10505558299999995
setattr:       0.12402554099999996
```

Overall it seems `getattr` is actually faster, but setting the field directly is faster. I do think a refactor with using `getattr` will be cleaner so I will proceed with that.

src/snowflake/snowpark/column.py
```diff
@@ -461,65 +510,91 @@ def between(
...
+    def _unary_op_impl(self, property: Callable, operator: UnaryExpression) -> "Column":
+        ast = Column._create_ast(
+            property = property, 
+            copy_messages = {"col": self._ast},
+        )
+        return Column(operator(self._expression), ast = ast)
 
     def __neg__(self) -> "Column":
         """Unary minus."""
+        # TODO: Need SpColumnNeg IR entity
```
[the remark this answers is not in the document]
sfc-gh-azwiegincew: Yes. I'll spend some quality time with Jira today.

src/snowflake/snowpark/types.py
sfc-gh-heshah: I agree, raising an error on the client side would probably be the best approach to raise the alarm on any Snowpark API elements with no representation in the IR.

I was afraid of blocking the client from working while we evaluate the IR changes, but I can turn the `TODO`s into `NotImplementedError` if that is okay.

Please see the pull request I created addressing all of the `TODO`s here: https://github.com/snowflakedb/snowflake/pull/183083

src/snowflake/snowpark/types.py
sfc-gh-heshah: There are some intermediary types like `_AtomicType(DataType)` that should not be required to implement `fill_ast` I think so I will raise a `ValueError` instead.

src/snowflake/snowpark/column.py
sfc-gh-heshah: This won't be an issue once all API functions that return `Column` instances also populate the `ast` parameter in the constructor.

For example currently because the `isin` method is not implemented to build the AST before returning, doing `df.filter(col("A").isin(1, 2, 3) & col("B"))` would fail since the boolean operator `&` would try to construct an AST using that of the new `col("A").isin(1, 2, 3)` column (which we currently don't fill if the only thing provided in the `Column` constructor is an `expr1` of type `Expression`.

src/snowflake/snowpark/column.py
sfc-gh-heshah: The alias field would be something that the user provides to the `DataFrame.alias(self, name: str)` method.

The Snowpark API then allows users to create columns that reference this DataFrame instance by its alias using a line like `col(<df_alias>, <column_name>)`

In the IR we'll need the alias to resolve which `DataFrame` instance the user is referring to, but there might be a better way to go about enabling this functionality.

src/snowflake/snowpark/column.py
sfc-gh-heshah: Not very difficult, I have already made the changes and compiled the new protobuf in the PR in the monorepo with the modifications necessary: https://github.com/snowflakedb/snowflake/pull/183083

However, I will get rid of these `TODO`s in favor of a flexible approach in the `_create_ast` method which can silently skip IR entities and fields or raise an error depending on a parameter.

src/snowflake/snowpark/_internal/proto/update-from-devvm.sh
```diff
@@ -1,19 +1,21 @@
...
 # Host devvm
 #	  HostName sdp-devvm-<ldap user>
 
 set -euxo pipefail
+SCRIPT_DIR=$(dirname "$0")
 
-# Step 1: Built python proto file from scratch via bazel
-ssh devvm "bash -c 'cd Snowflake/trunk;bazel build //Snowpark:py_proto'"
+# Step 1: Build the python proto file from scratch via bazel
+ssh $1 "bash -c 'cd Snowflake/trunk;bazel build //Snowpark:py_proto'"
```
sfc-gh-lspiegelberg, line 15: maybe some checks on `$1`? we could also make this optional, i.e. specify a variable `HOST` and set it to $1 if available, print help if not given etc.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+import re
+import sys
+
+from functools import reduce
+from pathlib import Path
+from typing import Any, List, Optional
+
+import snowflake.snowpark._internal.proto.ast_pb2 as proto
+
+NoneType = type(None)
```
sfc-gh-lspiegelberg, line 13: nit: maybe make this variables private to the module (i.e., by prefixing with `_` and then exporting only the public interfaces?)

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    str: "string_val",
+    bytearray: "binary_val",
+    bytes: "binary_val",
+    decimal.Decimal: "big_decimal_val",
+    datetime.date: "date_val",
+    datetime.time: "python_time_val",
+    datetime.datetime: "python_timestamp_val",
+}
+def infer_const_ast(obj: Any, ast: proto.Expr) -> None:
+    """Infer the Const AST expression from obj, and populate the provided ast.Expr() instance"""
```
sfc-gh-lspiegelberg, line 28: let's try to use Google format here

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    float: "float64_val",
+    str: "string_val",
+    bytearray: "binary_val",
+    bytes: "binary_val",
+    decimal.Decimal: "big_decimal_val",
+    datetime.date: "date_val",
+    datetime.time: "python_time_val",
+    datetime.datetime: "python_timestamp_val",
+}
+def infer_const_ast(obj: Any, ast: proto.Expr) -> None:
```
sfc-gh-lspiegelberg, line 27: nit: maybe rename to fill_const_ast? I think I saw you define another fill API somewhere.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    elif isinstance(obj, list):
+        for v in obj:
+            infer_const_ast(v, ast.list_val.vs.add())
+    
+    elif isinstance(obj, tuple):
+        for v in obj:
+            infer_const_ast(v, ast.tuple_val.vs.add())
+    
+    else:
+        raise TypeError("not supported type: %s" % type(obj))
```
sfc-gh-lspiegelberg, line 80: NotImplementedError?

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def get_non_snowpark_stack_frame() -> inspect.FrameInfo:
+    idx = 0
```
sfc-gh-lspiegelberg, line 84: pls. add docstring

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    snowpark_path = Path(__file__).parents[1]
+    while snowpark_path in Path(curr_frame.filename).parents:
+        idx += 1
+        curr_frame = call_stack[idx]
+    return curr_frame
+
+
+# TODO: Instead of regexp, grab assign statments using Python ast library
+RE_SYMBOL_NAME = re.compile(r'^\s*([a-zA-Z_]\w*)\s*=.*$', re.DOTALL)
+def get_symbol() -> Optional[str]:
```
sfc-gh-lspiegelberg, line 96: docstring

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    curr_frame = call_stack[idx]
+    snowpark_path = Path(__file__).parents[1]
+    while snowpark_path in Path(curr_frame.filename).parents:
+        idx += 1
+        curr_frame = call_stack[idx]
+    return curr_frame
+
+
+# TODO: Instead of regexp, grab assign statments using Python ast library
+RE_SYMBOL_NAME = re.compile(r'^\s*([a-zA-Z_]\w*)\s*=.*$', re.DOTALL)
```
sfc-gh-lspiegelberg, line 95: you prob. can move this constant within the get_symbol function

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+def get_symbol() -> Optional[str]:
+    code = get_non_snowpark_stack_frame().code_context
+    if code is not None:
+        for line in code:
+            match = RE_SYMBOL_NAME.fullmatch(line)
+            if match is not None:
+                return match.group(1)
+
+
+def fill_src_position(ast: proto.SrcPosition) -> None:
```
sfc-gh-lspiegelberg, line 105: docstring

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    curr_frame = get_non_snowpark_stack_frame()
+
+    ast.file = curr_frame.filename
+    ast.start_line = curr_frame.lineno
+    
+    if sys.version_info[1] >= 11:
```
sfc-gh-lspiegelberg, line 111: should we fill for the older Python versions with None or so?

src/snowflake/snowpark/_internal/proto/update-from-devvm.sh
```diff
@@ -1,19 +1,21 @@
 #!/usr/bin/env bash
-# This is a little helper script to build on the devvm the python protobuf file and copy it over to this directory.
-# Run script from within this directory.
-# For this script to succeed, need to configure SSH devvm host according to
+# This is a little helper script to build the python protobuf file and copy it over to this directory.
+# Run script from within this directory, and provide the Host as the argument.
+# Cloud workspace ssh is set up via ~/.local/share/sfcli/ssh_config (or simply provide your workspace ID)
+# For this script to succeed with the DevVM, you need to configure the SSH devvm host according to
```
sfc-gh-lspiegelberg, line 5: Can you maybe add an example of what commands to invoke with workspace / devvm?

src/snowflake/snowpark/column.py
sfc-gh-lspiegelberg: prob. would  be good to include your explanation as comment here

src/snowflake/snowpark/column.py
```diff
@@ -363,6 +431,7 @@ def __round__(self, n=None):
     def __hash__(self):
         return hash(self._expression)
 
+    # TODO: Implement AST generation for in_ / isin (relies on MultipleExpression and ScalarSubquery)
```
sfc-gh-lspiegelberg, line 434: Let's file a JIRA under https://snowflakecomputing.atlassian.net/browse/SNOW-1473690, then TODO SNOW-12345

src/snowflake/snowpark/column.py
```diff
@@ -461,65 +530,90 @@ def between(
...
-        return Column(BitwiseOr(Column._to_expr(other), self._expression))
+        return self._bin_op_rimpl("bit_or", BitwiseOr, other)
 
     def bitxor(self, other: Union[ColumnOrLiteral, Expression]) -> "Column":
         """Bitwise xor."""
-        return Column(BitwiseXor(Column._to_expr(other), self._expression))
+        return self._bin_op_rimpl("bit_xor", BitwiseXor, other)
+
+    def _unary_op_impl(self, property: str, operator: UnaryExpression, attr: str = "col") -> "Column":
+        ast = Column._create_ast(
```
sfc-gh-lspiegelberg, line 555: docstring

src/snowflake/snowpark/types.py
```diff
@@ -41,12 +42,26 @@ def __repr__(self):
...
+    def fill_ast(self, ast: proto.SpDataType) -> None:
+        """Populates the provided SpDataType instance's fields with the values corresponding to this DataType's instance
+
+        Args:
+            ast (proto.SpDataType): A provided (previously created) instance of an SpDataType IR entity
+
+        Raises:
+            ValueError: If corresponding SpDataType IR entity is not available, raise an error
+        """
+        raise ValueError(
```
sfc-gh-lspiegelberg, line 54: why not NotImplementedError?

src/snowflake/snowpark/types.py
```diff
@@ -41,12 +42,26 @@ def __repr__(self):
...
+            f"{self.__class__.__name__} has not implemented this method to fill the SpDataType IR entity correctly"
+        )
+
 
 # Data types
 class NullType(DataType):
     """Represents a null type."""
 
-    pass
+    def fill_ast(self, ast: proto.SpDataType) -> None:
```
sfc-gh-lspiegelberg, line 63: should these be private methods as well? `_fill_ast`?

tests/ast/run-unparser.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
```
sfc-gh-lspiegelberg, line 1: I think @sfc-gh-azwiegincew recently updated this, pls reflect!

sfc-gh-lspiegelberg (commented):
some nits, great work! Let's get this in.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+import sys
+
+from functools import reduce
+from pathlib import Path
+from typing import Any, List, Optional
+
+import snowflake.snowpark._internal.proto.ast_pb2 as proto
+
+NoneType = type(None)
+PYTHON_TO_AST_CONST_MAPPINGS = {
```
sfc-gh-azwiegincew, line 14: empty line to separate 13 and 14

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    bytes: "binary_val",
+    decimal.Decimal: "big_decimal_val",
+    datetime.date: "date_val",
+    datetime.time: "python_time_val",
+    datetime.datetime: "python_timestamp_val",
+}
+def infer_const_ast(obj: Any, ast: proto.Expr) -> None:
+    """Infer the Const AST expression from obj, and populate the provided ast.Expr() instance"""
+    if obj is None:
+        fill_src_position(ast.null_val.src)
```
sfc-gh-azwiegincew, line 30: should we not call `fill_src_position()` in other cases?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-azwiegincew: I'm not a huge fan of `PYTHON_TO_AST_CONST_MAPPINGS`. It's not reused anywhere, and you have complications such as the bytearray case. I would suggest removing this and just adding the cases directly in your one function that uses this global.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-azwiegincew: Should be `get_first_non_snowpark_stack_frame`. Otherwise it sounds like any stack frame will do.

src/snowflake/snowpark/dataframe.py
```diff
@@ -85,6 +85,11 @@
     Unpivot,
     ViewType,
 )
+from snowflake.snowpark._internal.ast_utils import (
+    setattr_if_not_none,
```
sfc-gh-azwiegincew, line 89: sort

tests/ast/data/col_alias.test
```diff
@@ -0,0 +1,20 @@
+## TEST CASE
```
sfc-gh-azwiegincew, line 1: empty line after this for symmetry with EXPECTED OUTPUT – in all the new tests

tests/ast/data/col_alias.test
```diff
@@ -0,0 +1,20 @@
...
+from snowflake.snowpark.functions import col
```
sfc-gh-azwiegincew, line 2: I would suggest adding `import snowflake.snowpark.functions` in the test driver and using `functions.col` in your tests.

sfc-gh-azwiegincew: Or because `col` is so special (i.e. common), adding that import as well (keeping .functions for other functions in the future).

tests/ast/run-unparser.sh
sfc-gh-azwiegincew: Yes, this file is now deleted.

tests/ast/test_ast_driver.py
```diff
@@ -23,29 +25,33 @@ class TestCase:
 
 def parse_file(file):
     """Parses a test case file."""
-    with open(file, "r", encoding="utf-8") as f:
+    with open(file, encoding="utf-8") as f:
```
sfc-gh-azwiegincew, line 28: Why did you remove `r`?

sfc-gh-azwiegincew (approved):
Gorgeous tests! Just a few nits.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-heshah: For now this was just to be able to create the `NullVal` message. Being conservative with usage just to be safe, but it can be called on every message extending the `Expr` trait. Should I do so for all consts?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-heshah: For this I tried to follow the error raised in the `type_utils.infer_type` method here: https://github.com/snowflakedb/snowpark-python/blob/d5e9f04741ece0ce851538da4e93dfadd454334e/src/snowflake/snowpark/_internal/type_utils.py#L434

I think a `TypeError` would be more informative as other types seem to be explicitly not supported.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-heshah: The protobuf message created will default to `None` for these fields if not provided. Trying to assign them with `None` will raise an error due to the type mismatch.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: let's call it on all consts. You prob. can get rid off the if here

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: I see, this here is tricky one:

Basically, I'd say we should add NotImplementedError if we are not feature complete. However, if we think the above list is feature complete ok to throw TypeError.

sfc-gh-vbudati: I think the TypeError implies that no other types will be supported in the future/ not expected to be supported. I would use the NotImplementedError as well

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,176 @@
...
+    call_stack = inspect.stack()
+    curr_frame = call_stack[idx]
+    snowpark_path = Path(__file__).parents[1]
+    while snowpark_path in Path(curr_frame.filename).parents:
+        idx += 1
+        curr_frame = call_stack[idx]
+    return curr_frame
+
+
+# TODO: Instead of regexp, grab assign statments using Python ast library
```
sfc-gh-lspiegelberg, line 136: can we add a JIRA?

src/snowflake/snowpark/_internal/proto/ast_pb2.py
```diff
@@ -11,582 +11,589 @@
 _sym_db = _symbol_database.Default()
```
sfc-gh-lspiegelberg, line 12: has this been updated with the most recent one from the server / mono repo branch?

src/snowflake/snowpark/_internal/proto/update-from-devvm.sh
sfc-gh-lspiegelberg: thanks for expanding the comment here! This should be fairly easy to use now.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -724,7 +724,7 @@ def ast_query(self, request_id__ast) -> Any:
         return self._conn._rest.request(
             f"/queries/v1/query-request?requestId={request_id}", req
         )
```
sfc-gh-lspiegelberg, line 726: remove change here (use `git restore -s origin src/snowflake/snowpark/_internal/server_connection.py` or so)

src/snowflake/snowpark/column.py
sfc-gh-lspiegelberg: can we add the example?

sfc-gh-lspiegelberg: is there a TODO post phase 0?

tests/ast/run-unparser.sh
sfc-gh-lspiegelberg: pls delete this file

sfc-gh-lspiegelberg (commented):
one more pass and then this should be good to go

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -662,7 +662,6 @@ def get_result_query_id(self, plan: SnowflakePlan, **kwargs) -> str:
         result_set, _ = self.get_result_set(plan, to_iter=True, **kwargs)
         return result_set["sfqid"]
 
-
```
sfc-gh-vbudati, line 665: accident?

src/snowflake/snowpark/column.py
```diff
@@ -461,65 +530,104 @@ def between(
...
+        return self._bin_op_rimpl("bit_xor", BitwiseXor, other)
+
+    def _unary_op_impl(
+        self, property: str, operator: UnaryExpression, attr: str = "col"
+    ) -> "Column":
+        """Unary operation implementation for Columns with automatic AST logging
+
+        Args:
+            property (str): IR entity representing the unary operation in the Column AST
+            operator (UnaryExpression): Snowpark unary operator to buidl the Expression instance
```
sfc-gh-vbudati, line 563:
```suggestion
            operator (UnaryExpression): Snowpark unary operator to build the Expression instance
```
nit

sfc-gh-vbudati (approved):
Looks great! I think it should be good to merge once you delete the file Arthur and Leonhard mentioned - tests/ast/run-unparser.sh

src/snowflake/snowpark/mock/_connection.py
sfc-gh-heshah: I think this was a side effect of running the precommit fix_lint

src/snowflake/snowpark/_internal/proto/ast_pb2.py
sfc-gh-heshah: Yes it is the most recent one

sfc-gh-lspiegelberg (approved):

PR #1699, opened 2024-05-29
[SNOW-1013918]: Fix pivot_table on empty DataFrames

sfc-gh-rdurrani:

   

   Fixes SNOW-1013918

This PR does the following:
1. Expand support for empty DataFrames in `pivot_table`.
2. Fix a bug where if index + columns cover all columns in the DataFrame (e.g. df has cols a, b, c, d, and columns=['a', 'b'] and index=['c', 'd']) `values` would incorrectly be set to `index` instead of being empty.


src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
```diff
@@ -194,6 +194,20 @@ def pivot_helper(
...
 
+    if pivot_aggr_groupings is None:
+        # When pivot_aggr_groupings is None, there are no `values` to compute on. In that case, we simply return
+        # a DataFrame with no columns, whose index is the result of grouping by the index columns.
+        ordered_dataframe = ordered_dataframe.group_by(
+            groupby_snowflake_quoted_identifiers
+        )
+        return InternalFrame.create(
+            ordered_dataframe=ordered_dataframe,
+            data_column_pandas_index_names=[None] + columns,
```
sfc-gh-yzou, line 205: why there is an extra [None] here?

src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
```diff
@@ -194,6 +194,20 @@ def pivot_helper(
     if ordered_dataframe.queries.get("post_actions"):
         ordered_dataframe = cache_result(ordered_dataframe)
 
+    if pivot_aggr_groupings is None:
+        # When pivot_aggr_groupings is None, there are no `values` to compute on. In that case, we simply return
+        # a DataFrame with no columns, whose index is the result of grouping by the index columns.
+        ordered_dataframe = ordered_dataframe.group_by(
```
sfc-gh-yzou, line 200: you can probably just call get_distinct_rows in groupby_utils.py here, it does pretty much the same thing with a dropping of extra columns, just to unify the same usage here.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6411,8 +6412,8 @@ def pivot_table(
             # groupby (index) nor pivot columns as the aggregation columns.  For example, a dataframe with
             # index=['A','B'], data=['C','E'] and if 'A' is used in groupby, and 'C' used as pivot, then 'E' would be
             # used as the values column, and unused index column 'B' would be dropped.
-            full_columns_and_index = (
-                list(columns) if columns else [] + list(index) if index else []
+            full_columns_and_index = (list(columns) if columns else []) + (
+                list(index) if index else []
```
sfc-gh-yzou, line 6416: I don't see actual difference here, is that just a reformat?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6421,43 +6422,76 @@ def pivot_table(
...
+                (isinstance(values, list) and len(values) > 1),
+                index,
+            )
+        except SnowparkSQLException as e:
+            # Error Code 1146 corresponds to the Snowflake Exception when
+            # a dynamic pivot is called and there are no pivot values and no
+            # groupby columns specified. If we hit this error, that means that
+            # we have attempted a pivot on an empty DataFrame, so we catch
+            # the exception and return an empty DataFrame.
+            if e.sql_error_code == 1146:
```
sfc-gh-yzou, line 6474: For this behavior, we do rely on the fact that a describing query is triggered for the underneath pivot in ordered_dataframe, right? can you add a comment here to be clear

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6421,43 +6422,76 @@ def pivot_table(
...
+                from snowflake.snowpark.modin.pandas.utils import from_pandas
+
+                return from_pandas(
+                    native_pd.DataFrame(
+                        index=self.index, columns=self.columns
+                    ).pivot_table(
```
sfc-gh-yzou, line 6480: hmm, why do we need all those? shouldn't we just return an empty df?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6506,6 +6541,18 @@ def pivot_table(
                         margins_name,
                     )
                 )
+        elif (
+            margins
+            and len(pivot_qc.columns) == 0
+            and len(pivot_qc.columns.names) != (len(columns) + 1)
```
sfc-gh-yzou, line 6547: what is the last condition? can you put a comment here about what are we doing with this branch, and why we need that? From the code, it seems that we are trying to increase the level of column index, but not clear to me why, an example could help to understand this

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6474,8 +6508,9 @@ def pivot_table(
         # Add margins if specified, note this will also add the row position since the margin row needs to be fixed
         # as the last row of the dataframe.  If no margins, then we order by the group by columns.
         # The final condition checks to see if there are any columns in the pivot result. If there are no columns,
-        # this means that we pivoted on an empty table - in that case, we can skip adding margins, since the result
-        # will still be an empty DataFrame (but we will have increased the join and union count) for no reason.
+        # this means that we pivoted on an empty table - in that case, we can skip adding margins, and just add
+        # an extra layer to the columns Index, since the result will still be an empty DataFrame (but we
```
sfc-gh-yzou, line 6512: is this updated comment try to explain what the elif branch added? if yes, can you move the comment below?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: No - the index and the columns of the DataFrame have names and levels that depend on the inputs to `pivot_table`. Mocking the logic up here would increase the complexity of the code, so I figured it would be easier to just rely on pandas to perform the logic of setting metadata correctly.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: I added parentheses - the error was because order of operations was different - the parentheses make order of operations explicit and correct.

CHANGELOG.md
```diff
@@ -20,10 +20,12 @@
 - Added partial support for `DataFrame.pivot_table` with no `index` parameter, as well as for `margins` parameter.
 - Updated the signature of `DataFrame.shift`/`Series.shift`/`DataFrameGroupBy.shift`/`SeriesGroupBy.shift` to match pandas 2.2.1. Snowpark pandas does not yet support the newly-added `suffix` argument, or sequence values of `periods`.
 - Re-added support for `Series.str.split`.
+- Expand support for empty DataFrames in `DataFrame.pivot_table`.
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: good catch - I did it in another PR and for some reason when making this PR I thought that the new release notes for 1.19.0 from that PR were here as well!

src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
sfc-gh-rdurrani: I'm not actually certain myself - it seems that when there are no values columns, pandas adds an additional level to the index for the columns - I suspect its adding the level for the aggregation function?

src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
sfc-gh-rdurrani: `get_distinct_rows` will also group by the pivot columns though, right? I looked at the method, and it seems that it groups by the projected columns, but here we only want to group by the groupby_snowflake_quoted_identifiers.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: Sure - (we rely on the fact that we eagerly grab the pivot values in dynamic pivot, and so if there are no pivot values and no groupby columns we throw an exception eagerly from the server side) - will add a comment to that effect!

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+
+### Snowpark pandas API Updates
+
+#### New Features
+
+#### Bug Fixes
+
+#### Improvements
+
+- Expand support for empty DataFrames in `DataFrame.pivot_table`.
```
sfc-gh-yzou, line 17: this description here is too vague, can you add more details about the "empty dataframe" case you are talking about? do you mean handle pivot on empty dataframe with no rows case to match native pandas behavior?

src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
sfc-gh-yzou: oops, sorry, i meant get_groups_for_ordered_dataframe in the groupby_utils.py

src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
sfc-gh-yzou: It seems that when there are no values, all rest of the columns that are not index + columns are used as values columns, and the index becomes multiindex, and all levels of the original columns are retained, which means the new data_column_pandas_index_names should be the original data_column_pandas_index_names + columns. you can try it with following example 
```
df = pd.DataFrame({'foo': ['one', 'one', 'one', 'two', 'two',
                           'three'],
                   'bar': ['A', 'B', 'C', 'A', 'B', 'C'],
                   'baz': [1, 2, 3, 4, 5, 6],
                   'zoo': [1, 2, 3, 1, 3, 5]})
df.columns.names = ['column']
df.pivot(index = 'bar', columns='foo')
```

This confirms with the original column index that is single level, you probably want to check if it the original column index is multi-level, and see if it follows the same reason.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: where did we add this? and please also add the fact that we are relying on the fact that we eagerly trigger the describing query to grab pivot result column, if that is not the case, no eager grabbing of pivot value will be triggerred

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: I see. i want to double confirm that once we reached here because the dataframe has 0 rows, is that correct?

src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
sfc-gh-rdurrani: Sorry - to clarify, this is handling the case where there are no columns that can be used for values - e.g. in the above example, this codepath would only be hit if the call to `pivot_table` looked like this:
```python
df.pivot_table(index=['bar', 'baz'], columns=['foo', 'zoo'])
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: > where did we add this?
The feature I described above is a part of the dynamic pivot implementation.

If you mean where is the comment, it should be added right above here?

The error is raised before we even attempt to materialize anything client side, as the compilation of a dynamic pivot query involves eagerly materializing the pivot values, and erroring out if no groupby columns are specified and no pivot values are found.

CHANGELOG.md
sfc-gh-rdurrani: Yup - I'll go ahead and update this!

src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
sfc-gh-yzou: I see, but i think for this part pivot and pivot_table behaves the same, you can use df.pivot_table(index = 'bar', columns='foo') with the above example for verify. i believe that is just a label expansion which retains the original name with expanded level

sfc-gh-yzou: i think it is still the same, i you try above example with the pivot_table operation, then do df2.columns, you see the following result
```
MultiIndex([], names=['column', 'foo', 'zoo'])
```

sfc-gh-rdurrani: Ah I see - seems like the name that is added should then be the current names of the index. Will add.

sfc-gh-rdurrani: Also confirmed that this is the case with MultiIndex:
```python
In [6]: df.columns = pd.MultiIndex.from_tuples([('a', 1), ('a', 2), ('b', 1), ('b', 2)])

In [7]: df
Out[7]:
       a     b
       1  2  1  2
0    one  A  1  1
1    one  B  2  2
2    one  C  3  3
3    two  A  4  1
4    two  B  5  3
5  three  C  6  5

In [8]: df.columns.names = ['c1', 'c2']

In [9]: df
Out[9]:
c1      a     b
c2      1  2  1  2
0     one  A  1  1
1     one  B  2  2
2     one  C  3  3
3     two  A  4  1
4     two  B  5  3
5   three  C  6  5

In [10]: df.columns
Out[10]:
MultiIndex([('a', 1),
            ('a', 2),
            ('b', 1),
            ('b', 2)],
           names=['c1', 'c2'])

In [11]: df.pivot_table(index=[('a', 1), ('a', 2)], columns=[('b', 1), ('b', 2)]).columns
Out[11]: MultiIndex([], names=['c1', 'c2', ('b', 1), ('b', 2)])
```

sfc-gh-rdurrani: While adding the logic + tests for this, I decided to add tests to check that pivot_table works correctly with named columns when not empty (basically added a `named_columns` parameter to some other tests), and noticed that this causes a some other tests to fail (results from running all tests under the pivot directory):

`14 failed, 243 passed, 7 skipped, 26 xfailed, 4 xpassed, 164 warnings`

I'm thinking it makes sense to add the tests here (with an xfail) and fix the other failures not related to empty dataframes in a future PR. Thoughts?

sfc-gh-rdurrani: Also, side note: I've added tests for empty pivot table with multi-indexes to check that the multi index names are carried over to the result; however, we don't support pivot table with multi-index currently, so I've skipped the tests. I mostly just added them so that if we do add multi index support, we won't forget to turn on these tests and verify that they pass, but I'm happy to remove them for now? Do we prefer skips (which don't significantly increase CI time) or xfails with raises=NotImplementedError (which will make it so that these tests catch when we add multiindex support, but also add time to CI to run a bunch of tests we know will fail)?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: sorry, i was asking where did we put the comment,  i don't think i see the comment before. 

I think my point is that our this does require the fact that we eagerly triggered a metadata query (describe) for pivot, which will trigger a compilation on the server side, so people can be careful when they are replying on catching this error. However, i don't think we will be able to get rid of the metadata query in any short term, which should be fine.

tests/integ/modin/pivot/test_pivot_single.py
```diff
@@ -240,9 +243,186 @@ def test_pivot_on_inline_data_using_temp_table():
...
+        (("d", 4), [("a", 1), ("b", 2), ("c", 3)]),
+        ([("d", 4)], ("a", 1)),
+        ([("d", 4)], [("a", 1)]),
+        ([("d", 4)], [("a", 1), ("b", 2)]),
+        ([("d", 4)], [("a", 1), ("b", 2), ("c", 3)]),
+    ],
+)
+@pytest.mark.parametrize("margins", [True, False])
+@pytest.mark.parametrize("named_columns", [True, False])
+def test_pivot_empty_frame_snow_1013918_multiindex(
```
sfc-gh-yzou, line 353: it is little bit wired that we are adding test that are skipped, was it because mutliIndex is not supported by snowpark pandas yet, and what you are adding there should also works for multiIndex? Can you create a ticket for the case that currently is not supported yet, and add the ticket number in the reason

sfc-gh-yzou (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: I see - so should I update the comment or is it good as is?

tests/integ/modin/pivot/test_pivot_single.py
sfc-gh-rdurrani: Sure - the test is skipped because multiindex isn't supported yet, but I want to make sure we catch this case when we do implement multiindex. Will look for (or create) a ticket for this, and include in the description!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: i think would be good to add that in comment so that future developer is aware of this.

src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py
```diff
@@ -342,7 +398,16 @@ def pivot_helper(
     # Generate the data column pandas index names
     if not isinstance(columns, list):
         columns = [columns]
-    columns = [None] * len(pivot_aggr_groupings[0].prefix_label) + columns
+    if len(pivot_aggr_groupings[0].prefix_label) != 0:
+        columns = (
```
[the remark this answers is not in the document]
sfc-gh-yzou: maybe also give an examples about how the label look like here

sfc-gh-helmeleegy (approved):

PR #1701, opened 2024-05-29
[feature branch] AST test infra

sfc-gh-azwiegincew:
## AST Tests

This driver enables testing of the AST generation that will be used in the server-side Snowpark implementation, starting with Phase 0.

All generated AST should be tested using this mechanism. To add a test, create a new file under `tests/ast/data`. Files look like the following example. The test driver sets up the session and looks at the accumulated lazy values in the resulting environment.

N.B. No eager evaluation is permitted, as any intermediate batches will not be observed. This can easily be changed if necessary, however.

```python
## TEST CASE

df = session.table("test_table")
df = df.filter("STR LIKE '%e%'")

## EXPECTED OUTPUT

res1 = session.table('test_table')

res2 = res1.filter('STR LIKE '%e%'')
```

To generate the expected output the first time the test is run, or when the AST generation changes, run:
```bash
pytest --update-expectations tests/ast
```


tests/ast/test_ast_driver.py
```diff
@@ -0,0 +1,94 @@
...
+
+from dataclasses import dataclass
+import os
+import pathlib
+import pytest
+import subprocess
+
+TEST_DIR = pathlib.Path(__file__).parent / "data"
+
+UNPARSER_PATH="/Users/azwiegincew/src/snowflake/Snowpark/unparser/run.sh"
```
sfc-gh-azwiegincew, line 13: WIP

sfc-gh-lspiegelberg:
Can we maybe avoid checking in the .jar file? It's 22MB which will make any changesets large...

src/snowflake/snowpark/dataframe.py
```diff
@@ -1133,7 +1133,8 @@ def select(
 
         stmt = self._session._ast_batch.assign()
         ast = stmt.expr
-        ast.sp_dataframe_select__columns.df.sp_dataframe_ref.id.bitfield1 = self._ast_id
+        # TODO: remove the None guard below once we generate the correct AST.
```
sfc-gh-lspiegelberg, line 1136: what is 666 about? I assume just some random id for now?

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -795,6 +795,9 @@ def get_result_query_id(self, plan: SnowflakePlan, **kwargs) -> str:
         # get the iterator such that the data is not fetched
         result_set, _ = self.get_result_set(plan, to_iter=True, **kwargs)
         return result_set["sfqid"]
+    
+    def create_coprocessor(self):
```
sfc-gh-lspiegelberg, line 799: would we add in the future any functionality to this? Or is this a simple stub? In any case maybe a short comment to  address either 1.) what needs to get added here or 2.) that this should not be removed, but left because of ... would help :)

tests/ast/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+N.B. No eager evaluation is permitted, as any intermediate batches will not be observed. This can easily be changed if necessary, however.
+
+```python
+## TEST CASE
+
+df = session.table("test_table")
+df = df.filter("STR LIKE '%e%'")
+
+## EXPECTED OUTPUT
```
sfc-gh-lspiegelberg, line 15: could we make the expected output agnostic to empty lines? Feels like we may have a lot of troubleshooting with empty lines, and I don't think having a mismatch in empty lines warrants a test failure.

sfc-gh-lspiegelberg (approved):
lgtm, only concern is the 22MB file. Can we avoid checking that in?

src/snowflake/snowpark/mock/_connection.py
sfc-gh-azwiegincew: Done.

tests/ast/README.md
sfc-gh-azwiegincew: There's a _super_ easy way to reset these test expectations if the output changes. I don't think doing this is necessary and I would argue that having the exact correct output is cleaner than not.

PR #1703, opened 2024-05-29
Use pytest mark to run subset of integ modin tests in Snowfort

sfc-gh-lmukhopadhyay:

   

   Fixes SNOW-1447077

   We are using pytest.mark to annotate tests that are selected to be run in stored procedures with `short_regress` or `precommit`. The annotations are used to run the tests server side in the Snowfort regression framework either as part of the snowflake/main merge gate (short_regress) or the precommit run. For example to trigger the short_regress tests, the server side Snowfort test runs `pytest.main(['filedir', '-m', 'short_regress')`.

More details can be found in the design doc [here](https://docs.google.com/document/d/1OGJbvu9LiSli2VBcY0vZoBAJ2L9nyigl0PR9vauIGUY/edit#heading=h.7p62svfd1gs7).


tox.ini
```diff
@@ -196,6 +196,8 @@ markers =
     timeout: tests that need a timeout time
     localtest: local tests
     scala: scala tests
+    short_regress: short_regress tests
+    bptp: bptp tests
```
sfc-gh-azhan, line 200:
```suggestion
    precommit: precommit tests
```

tox.ini
```diff
@@ -196,6 +196,8 @@ markers =
     timeout: tests that need a timeout time
     localtest: local tests
     scala: scala tests
+    short_regress: short_regress tests
```
sfc-gh-azhan, line 199: should we rename it to be more specific? e.g., `modin_sp_short_regress` and `modin_sp_precommit`?

sfc-gh-lmukhopadhyay: Yeah that's a good idea!

sfc-gh-yzou:
@sfc-gh-lmukhopadhyay can you update the description of this pr to include how the mark will be used when we run the tests? basically how server side uses these marks?

sfc-gh-azhan (approved):

sfc-gh-jrose (approved):

tests/integ/modin/frame/test_apply.py
```diff
@@ -603,6 +603,7 @@ def g(v):
 ]
 
 
+@pytest.mark.modin_sp_short_regress
```
sfc-gh-yzou, line 606: this test seems simply raises an NotImplementedError, should we at least get one test point that runs?

tests/unit/modin/test_internal_frame.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, sp_df: OrderedDataFrame, internal: InternalFrame) -> None:
         self.internal_frame: InternalFrame = internal
 
 
+@pytest.mark.modin_sp_short_regress
```
sfc-gh-yzou, line 40: is it possible to simply mark the whole suite to run as short_regress?

tox.ini
```diff
@@ -196,6 +196,8 @@ markers =
     timeout: tests that need a timeout time
     localtest: local tests
     scala: scala tests
+    modin_sp_short_regress: modin short_regress tests run in sproc
```
sfc-gh-yzou, line 199: add a comment here to make it clear that those marker are used at server side to decide which tests to run, any change to the name will require a server change also, please be careful with any changes here

tests/integ/modin/frame/test_apply.py
```diff
@@ -603,6 +603,7 @@ def g(v):
...
 @pytest.mark.parametrize("data, apply_func", TRANSFORM_DATA_FUNC_MAP)
 @sql_count_checker(query_count=0)
```
sfc-gh-yzou, line 608: Can you also get one from groupby apply?

tests/integ/modin/frame/test_apply.py
sfc-gh-azhan: right this one does not need to be in.

tests/integ/modin/frame/test_apply.py
sfc-gh-azhan: Now groupby apply and some apply tests are failing. We will add it in the coming PRs if we can fix those tests.

tests/unit/modin/test_internal_frame.py
sfc-gh-azhan: Yes, but right now we try to be clear this is only used by modin. Today python API tests are defined in the snowfort code.

tox.ini
sfc-gh-azhan: Is this obvious? If marker changed, it definitely will affect server side.

tests/unit/modin/test_internal_frame.py
sfc-gh-yzou: I think my question is that is it possible to mark the whole suite with one marker instead of marking it one after another? Not about the name, we can keep modin in the name

tox.ini
sfc-gh-yzou: I don't think it is obvious, this is client side change, it is just used to run the tests, can be at both client and server. If  any change is made at client side, server side test will not automatically run,  which may not be captured immediately. 
It would be better if we can add a comment here to make it more clear to all developer in Snowpark python

tests/integ/modin/frame/test_apply.py
sfc-gh-yzou: Yes, @sfc-gh-lmukhopadhyay mentioned about this to me, she will tickets for the each category of test failure for tracking

tests/unit/modin/test_internal_frame.py
sfc-gh-lmukhopadhyay: Yes should be possible, we can add this in the next marker PR

tox.ini
sfc-gh-azhan: I see, we can add more comments to make it clear.

Issue #1590, opened 2024-05-15
SNOW-1427770: [Local testing] Table merge fails when update clause exists but no update takes place


mraraujo:
1. What version of Python are you using?

3.11.6

2. What operating system and processor architecture are you using?

Linux-6.5.0-28-generic-x86_64-with-glibc2.38

3. What are the component versions in the environment (`pip freeze`)?

Snowpark version 1.16.0

4. What did you do?
```
from snowflake.snowpark import Session
from snowflake.snowpark.functions import when_matched, when_not_matched

session = Session.builder.config('local_testing', True).create()

target = session.create_dataframe([[1, "a"]], schema=['id', 'name'])
target.write.save_as_table('target', mode='overwrite')
target = session.table('target')

source = session.create_dataframe([[2, "b"]], schema=['id', 'name'])

target.merge(source, target['id'] == source['id'], [when_matched().update({'name': source['name']}), when_not_matched().insert({'id': source['id'], 'name': source['name']})])
```

5. What did you expect to see?

The target table should have one additional row with id 2.
Instead, we get an exception:

```
Traceback (most recent call last):
  File "/home/maraujo/Desktop/test.py", line 12, in <module>
    target.merge(source, target['id'] == source['id'], [when_matched().update({'name': source['name']}), when_not_matched().insert({'id': source['id'], 'name': source['name']})])
  File "/home/maraujo/venvs/semantec-backend/lib/python3.11/site-packages/snowflake/snowpark/table.py", line 677, in merge
    _get_merge_result(
  File "/home/maraujo/venvs/semantec-backend/lib/python3.11/site-packages/snowflake/snowpark/table.py", line 253, in _get_merge_result
    rows_updated = int(rows[0][idx])
                       ~~~~~~~^^^^^
  File "/home/maraujo/venvs/semantec-backend/lib/python3.11/site-packages/snowflake/snowpark/row.py", line 142, in __getitem__
    return super().__getitem__(item)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: tuple index out of range
```

Note that nothing fails if the source dataframe also provides an update. E.g., changing the source to this:
```
source = session.create_dataframe([[1, "c"], [2, "b"]], schema=['id', 'name'])
```

succeeds.

6. Can you set logging to DEBUG and collect the logs?
```
2024-05-15 18:11:59,423 - MainThread session.py:500 - __init__() - INFO - Snowpark Session information: 
"version" : 1.16.0,
[...]

PR #1719, opened 2024-05-31
SNOW-1427770 local testing table merge fails when update clause exists but no update takes place

sfc-gh-aling:

   

   Fixes SNOW-NNNNNNN

fix https://github.com/snowflakedb/snowpark-python/issues/1590

the live connection, whether the result is updated/inserted/deleted is decided by whether the corresponding clause shows up, not by whether there is actual data update.

align the local testing behavior with live connection.


sfc-gh-jrose (approved):

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -1493,11 +1497,11 @@ def outer_join(base_df):
 
         # Generate metadata result
         res = []
-        if inserted_rows:
+        if inserted:
```
sfc-gh-stan, line 1500: I don't understand the changes 🤔  is there a case when `inserted` is False but `inserted_row_idx` is non empty?

sfc-gh-aling: it's the case that users specify update/insert/delete ops, but there is no actual data update -- `inserted_row_idx/updated_row_idx/deleted_row_idx` being empty.

this can illustrated with the added test cas + code in https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/table.py#L667-L685.

the test code containing condition `when_matched().delete()` in live connection returns `[Row(number of rows inserted=1, number of rows deleted=0)]` while in local testing it was returning [Row(1)] missing the 0 rows deleted part.

sfc-gh-stan: I see, in that case I prefer we do something like: 
```
if any(isintance(clause, UpdateMergeExpression) for clause in ``source_plan.clauses`):
    res.append(...)
```
or changing the variable from `updated` to `update_clause_specified` to distinguish between the case when there are rows updated vs when there are no rows updated but an update clause is specified.

sfc-gh-aling: I changed the var name, thanks for the suggestion

sfc-gh-stan (approved):

PR #1721, opened 2024-05-31
Skip SQL count check in test

sfc-gh-jdu:

   

   Fixes [SNOW-1435354](https://snowflakecomputing.atlassian.net/browse/SNOW-1435354)

   We skip sql count check in some scenarios (e.g., cte tests, stored proc regression tests) by adding a pytest argument

[SNOW-1435354]: https://snowflakecomputing.atlassian.net/browse/SNOW-1435354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integ/modin/conftest.py
+++ b/tests/integ/modin/conftest.py
@@ -45,6 +45,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--generate_pandas_api_coverage", action="store_true", default=False
     )
+    parser.addoption("--skip_sql_count_check", action="store_true", default=False)
 
 
 @pytest.fixture(scope="session", autouse=True)
@@ -52,6 +53,17 @@ def setup_pandas_api_coverage_generator(pytestconfig):
     enable_coverage = pytestconfig.getoption("generate_pandas_api_coverage")
     if enable_coverage:
         PandasAPICoverageGenerator()
+
+
+SKIP_SQL_COUNT_CHECK = False
+
+
+@pytest.fixture(scope="session", autouse=True)
+def setup_skip_sql_count_check(pytestconfig):
+    skip = pytestconfig.getoption("skip_sql_count_check")
+    if skip:
+        global SKIP_SQL_COUNT_CHECK
+        SKIP_SQL_COUNT_CHECK = True
 
 
 @pytest.fixture(scope="function")
--- a/tests/integ/modin/sql_counter.py
+++ b/tests/integ/modin/sql_counter.py
@@ -122,7 +122,11 @@ class SqlCounter:
         from tests.integ.modin.conftest import SKIP_SQL_COUNT_CHECK
 
         self._queries: list[QueryRecord] = []
-        self._no_check = no_check or SKIP_SQL_COUNT_CHECK
+
+        # Bypassing sql counter since
+        #   1. it is an unnecessary metric for tests running in stored procedures
+        #   2. pytest-assume package is not available in conda
+        self._no_check = no_check or IS_IN_STORED_PROC or SKIP_SQL_COUNT_CHECK
 
         # Save any expected sql counts initialized at start up.
         self._expected_sql_counts = {}
@@ -359,12 +363,6 @@ def sql_count_checker(
     *args,
     **kwargs,
 ):
-    # Bypassing sql counter since
-    #   1. it is an unnecessary metric for tests running in stored procedures
-    #   2. pytest-assume package is not available in conda
-    if IS_IN_STORED_PROC:
-        return
-
     """SqlCounter decorator that automatically validates the sql counts when test finishes."""
     sql_counter = SqlCounter(
         no_check=no_check,
--- a/tests/integ/modin/sql_counter.py
+++ b/tests/integ/modin/sql_counter.py
@@ -119,8 +119,10 @@ class SqlCounter:
         high_count_reason=None,
         **kwargs,
     ) -> "SqlCounter":
+        from tests.integ.modin.conftest import SKIP_SQL_COUNT_CHECK
+
         self._queries: list[QueryRecord] = []
-        self._no_check = no_check
+        self._no_check = no_check or SKIP_SQL_COUNT_CHECK
 
         # Save any expected sql counts initialized at start up.
         self._expected_sql_counts = {}


tests/integ/modin/conftest.py
```diff
@@ -45,6 +45,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--generate_pandas_api_coverage", action="store_true", default=False
     )
+    parser.addoption("--skip_sql_count_check", action="store_true", default=False)
```
sfc-gh-azhan, line 48: can we have a test for this?

sfc-gh-jdu: hmm what's the best way to test it? I'll use this argument very soon in #1626

sfc-gh-azhan (approved):

tests/integ/modin/sql_counter.py
```diff
@@ -119,8 +119,10 @@ def __init__(
         high_count_reason=None,
         **kwargs,
     ) -> "SqlCounter":
+        from tests.integ.modin.conftest import SKIP_SQL_COUNT_CHECK
+
```
sfc-gh-yzou, line 123: @sfc-gh-lmukhopadhyay I thought you add a skip for stored sproc here also, where did it go? i want to unify this two changes so that we just use the same way to control skip of sql counter check

tests/integ/modin/conftest.py
```diff
@@ -54,6 +55,17 @@ def setup_pandas_api_coverage_generator(pytestconfig):
         PandasAPICoverageGenerator()
 
 
+SKIP_SQL_COUNT_CHECK = False
+
```
sfc-gh-yzou, line 59: oh, does that needs to be a global variable? was that how we handle sql simplifier before?

tests/integ/modin/sql_counter.py
sfc-gh-lmukhopadhyay: I added it to [sql_count_checker decorator](https://github.com/snowflakedb/snowpark-python/pull/1721/files#diff-745c288d4dfae7e02324d3c313319c2a5757068eaa059e8a9dd08698e458b7d9R365) only, I was going to add it to the SqlCounter object instead today but it'd be better if we can reuse this

sfc-gh-yzou: yes, let's use the same thing. However, i am not sure if the change here will work for the stored sproc skip, because the stored sproc needs to skip any calling of pytest assume.

@sfc-gh-jdu can you take a look and unify the two changes? you can either revert the other change @sfc-gh-lmukhopadhyay has add, or move your change to where @sfc-gh-lmukhopadhyay has add. just keep in mind that the stored sproc skip needs to make sure no calling of method that requries pytest assume.

sfc-gh-evandenberg (approved):
LGTM

tests/integ/modin/sql_counter.py
```diff
@@ -119,8 +119,10 @@ def __init__(
...
         self._queries: list[QueryRecord] = []
-        self._no_check = no_check
+        self._no_check = no_check or SKIP_SQL_COUNT_CHECK
```
sfc-gh-lmukhopadhyay, line 125: @sfc-gh-jdu if you also set self._no_check for the condition IS_IN_STORED_PROC also, I believe it should skip the Sql Counter and thus pytest assume

tests/integ/modin/sql_counter.py
sfc-gh-jdu: I have to manually add this flag here https://github.com/snowflakedb/snowpark-python/pull/1626/files#diff-01c804bbf883a36af718bc99184689135c2eef596fba42a5f4f5ded2e025ef16R271, which means there should be another PR to add it in regression test file.

tests/integ/modin/conftest.py
sfc-gh-jdu: we also use a flag for sql simplifier, but it's implemented as a fixture. However, sql count check don't have access to a fixture, so I just use a global variable. It's indeed a global variable because I access it using `global` keyword later.

tests/integ/modin/sql_counter.py
sfc-gh-jdu: yeah this is a good idea

sfc-gh-helmeleegy (approved):
LGTM, thanks!

tests/integ/modin/sql_counter.py
sfc-gh-jdu: done

tests/integ/modin/sql_counter.py
sfc-gh-yzou: hmm, i am not sure if that is going to work for the stored sproc, I believe this line https://github.com/snowflakedb/snowpark-python/pull/1721/files#diff-745c288d4dfae7e02324d3c313319c2a5757068eaa059e8a9dd08698e458b7d9R384 requires the pytest_assume package.

if you set no_check here, will you be able to skip that line? also i don't think it is necessary to check IS_IN_STORED_PROC, if the flag can be used, we can probably change in our regression test run to make sure this flag is set to false.

sfc-gh-jdu: I see, no I don't think we can skip this line. Then we still need to add the flag in regression test @sfc-gh-lmukhopadhyay. Let me revert this change

sfc-gh-yzou: @sfc-gh-jdu sorry, what i mean is do you need this flag to be set inside the sql_counter initializer, can you simply do it at the same place as where the stored sproc is skipped?

sfc-gh-jdu: that place only skips `sql_count_check` decorator, but it seems not skip SqlCounter if it is created directly

sfc-gh-jdu: There should be another PR to add that flag in regression test file if you don't want to check IS_IN_STORED_PROC here, though it works

sfc-gh-lmukhopadhyay: This looks good to me, thanks for reverting the stored_sproc change!

sfc-gh-lmukhopadhyay (approved):

tests/integ/modin/sql_counter.py
sfc-gh-yzou: The current change looks good. @sfc-gh-lmukhopadhyay @sfc-gh-azhan let's follow up to simply reuse the SKIP_SQL_COUNT_CHECK in the stored sproc test infra, so that we do not have to do a special check for IS_IN_STORED_PROC. @sfc-gh-jdu can you create a ticket for this?

sfc-gh-yzou (approved):

sfc-gh-dpetersohn (approved):
LGTM

tests/integ/modin/sql_counter.py
sfc-gh-jdu: https://snowflakecomputing.atlassian.net/browse/SNOW-1462284 thank y'all!

sfc-gh-azhan: we can just pass the flag as a pytest argument when we need it. cc @sfc-gh-lmukhopadhyay

PR #1724, opened 2024-06-01
SNOW-1231747: Add Support for Series.str.__getitem__

sfc-gh-helmeleegy:

   

   Fixes SNOW-1231747

   Add Support for Series.str.__getitem__.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12184,6 +12184,33 @@ def _get_regex_params(self, flags: int = 0) -> str:
...
+            Index to retrieve data from.
+
+        Returns
+        -------
+        SnowflakeQueryCompiler representing result of the string operation.
+        """
+        import typing
+
+        if is_scalar(key):
+            if key is not None and not isinstance(key, int):
```
sfc-gh-dpetersohn, line 12203: `None` is not an instance of `int`
```suggestion
            if not isinstance(key, int):
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12184,6 +12184,33 @@ def _get_regex_params(self, flags: int = 0) -> str:
...
+                ErrorMessage.not_implemented(
+                    "Snowpark pandas string indexing doesn't yet support non-numeric keys"
+                )
+            return self.str_get(typing.cast(int, key))
+        else:
+            assert isinstance(key, slice)
```
sfc-gh-dpetersohn, line 12209: We might want a more descriptive error message here if the user passes in something wrong.

sfc-gh-helmeleegy: Good point. I also added a special case for bad (non-scalar and non-slice) keys to make sure we follow pandas behavior there. This way the assert should always pass. Still added a message though.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: I think we need the check here - just to make sure we handle a None key in the same way pandas does. It does not raise an error, but rather return None for all values in the series. The same goes for all other bad keys (non-scalar and non-slice). We also have to cast because None is not an instance of int.

sfc-gh-dpetersohn: I see what you mean, I thought that `series.str[None]` threw an error.

sfc-gh-helmeleegy:
@sfc-gh-dpetersohn @sfc-gh-jkew this should be ready for (re-)review.

sfc-gh-dpetersohn (approved):
LGTM

sfc-gh-jkew (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12184,6 +12184,36 @@ def _get_regex_params(self, flags: int = 0) -> str:
...
+        Parameters
+        ----------
+        key : scalar or slice
+            Index to retrieve data from.
+
+        Returns
+        -------
+        SnowflakeQueryCompiler representing result of the string operation.
+        """
+        import typing
```
sfc-gh-mvashishtha, line 12206: note: this comment is irrelevant if we can do `int()` instead of `typing.cast(int,)`.

```suggestion
```

Importing inside a function looks odd. Can we move this [here](https://github.com/snowflakedb/snowpark-python/blob/e7813c14fa195e0992e0c59f713bd8dc5f4b4d7e/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py#L11)?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12184,6 +12184,36 @@ def _get_regex_params(self, flags: int = 0) -> str:
...
+
+        if not is_scalar(key) and not isinstance(key, slice):
+            # Follow pandas behavior; all values will be None.
+            key = None
+        if is_scalar(key):
+            if key is not None and not isinstance(key, int):
+                ErrorMessage.not_implemented(
+                    "Snowpark pandas string indexing doesn't yet support non-numeric keys"
+                )
+            return self.str_get(typing.cast(int, key))
```
sfc-gh-mvashishtha, line 12216:
```suggestion
            return self.str_get(int(key))
```

Does this work?

sfc-gh-mvashishtha (approved):
@sfc-gh-helmeleegy I'm adding two minor comments. If it seems to you that they don't need to be addressing, I'm okay with merging this PR.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Nop. It doesn't work for None.
```
In [1]: int(None)
TypeError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 int(None)

TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Done.

PR #1727, opened 2024-06-03
[IR] Steel-thread example for pandas iloc get

sfc-gh-vbudati:
This is the steel-thread for Phase 0 pandas. The goal is to have iloc get working. I tried to cover all the basic row + columns cases but only test for scalar rows and columns.


src/snowflake/snowpark/modin/pandas/indexing.py
```diff
@@ -1049,14 +1049,34 @@ def __getitem__(
 
         original_row_loc = row_loc  # keep a copy for error message
 
+        # IR changes! Add iloc get nodes to AST.
+        stmt = self.df._get_session()._ast_batch.assign()
+        ast = stmt.expr
+        ast.pd_dataframe_iloc_get.df.pd_dataframe_ref.id.bitfield1 = (
```
sfc-gh-azwiegincew, line 1055: remove the parens

src/snowflake/snowpark/modin/pandas/indexing.py
```diff
@@ -1049,14 +1049,34 @@ def __getitem__(
...
+            self.df._get_ast_id()
+        )
+
+        # The row and column locations can be a scalar, slice, list-like, or Series object - record all of these
+        # separately for AST.
```
sfc-gh-azwiegincew, line 1060: I don't think you should do this here. I would have a reusable function `ast_expr_from_python_val(ast_expr, val)` that attempts to map an arbitrary Python value to an AST expression, following the types in the AST. You would invoke it for every value of type `Expr` in the AST. It will map all the built-ins (functions, scalars, lists, maps, slices, etc.) plus emit `Ref` nodes for dataframes, series, and indexes.

Once you have this fn, you can just blindly pass `key` to it.

sfc-gh-azwiegincew (changes requested):

src/snowflake/snowpark/modin/pandas/indexing.py
sfc-gh-vbudati: I think the linter did this when running pre-commit, I can try to get rid of it if preferred

src/snowflake/snowpark/modin/pandas/indexing.py
sfc-gh-vbudati: I think I see what you mean. Is this supposed to happen on the client side/`indexing.py` or in the TcmSession logic?

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -20,14 +25,27 @@ def expr_to_dataframe_expr(expr):
     return dfe
 
 
+def ast_expr_from_python_val(expr, val):
+    if isscalar(val):
+        expr.scalar = val
+    elif isinstance(val, DataFrame):
+        expr.dataframe_expr = val
```
sfc-gh-azwiegincew, line 32: this should be a `Ref`. we don't nest DF exprs.

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -20,14 +25,27 @@ def expr_to_dataframe_expr(expr):
...
+    elif isinstance(val, Series):
+        expr.series_expr = val
+    elif isinstance(val, slice):
```
sfc-gh-azwiegincew, line 35: slice and lists should go before dataframe and series, since they're Python values.

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -20,14 +25,27 @@ def expr_to_dataframe_expr(expr):
     return dfe
 
 
+def ast_expr_from_python_val(expr, val):
+    if isscalar(val):
+        expr.scalar = val
```
sfc-gh-azwiegincew, line 30: You need to wrap in type-specific wrappers, like `Int32Val`, etc. See const.ir.

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -20,14 +25,27 @@ def expr_to_dataframe_expr(expr):
...
+
 class AstBatch:
     def __init__(self, session):
         self._session = session
         self._id_gen = itertools.count(start=1)
         self._init_batch()
         # TODO: extended version from the branch snowpark-ir.
 
-    def assign(self, symbol = None):
+    def assign(self, symbol=None):
```
sfc-gh-azwiegincew, line 48: Did the linter do this? I see spaces around this use case of `=` in most of our Python code.

sfc-gh-azwiegincew (changes requested):

src/snowflake/snowpark/_internal/ast.py
sfc-gh-vbudati: not sure, I'll fix this

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -25,17 +28,48 @@ def expr_to_dataframe_expr(expr):
     return dfe
 
 
+# Map from python type to its corresponding IR field. IR fields below all have the 'v' attribute.
```
sfc-gh-azwiegincew, line 31: It's technically an "IR entity". These look like fields in the protobuf builder.

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -25,17 +28,48 @@ def expr_to_dataframe_expr(expr):
...
+TYPE_TO_IR_TYPE_NAME = {
+    bytes: "binary_val",
+    bool: "bool_val",
+    int32: "int_32_val",
+    int64: "int_64_val",
+    float64: "float_64_val",
+    Decimal: "big_decimal_val",
```
sfc-gh-azwiegincew, line 38: should be sorted

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -25,17 +28,48 @@ def expr_to_dataframe_expr(expr):
...
-        expr.slice_expr = val
-    elif is_list_like(val):
-        expr.array_expr = val
+    """
+    Converts a Python value to an IR expression.
+    This IR expression is set to an attribute of `expr`.
+
+    Parameters
+    ----------
+    expr : IR expression object
```
sfc-gh-azwiegincew, line 53: IR Entity protobuf builder

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -25,17 +28,48 @@ def expr_to_dataframe_expr(expr):
...
+    val : Python value that needs to be converted to IR expression.
+    """
+    if val is None:
+        expr.none_val = val
+    val_type = type(val)
+    if val_type not in TYPE_TO_IR_TYPE_NAME:
+        if isinstance(val, Callable):
+            expr.fn_val.params = signature(val).parameters
+            expr.fn_val.body = val
+        elif isinstance(val, Series):
```
sfc-gh-azwiegincew, line 63: Series should be next to Dataframe (and probably after list).

It would be really helpful if we maintained a canonical order for this list of types, which reoccurs in a few places.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-vbudati: I think this is the linter

sfc-gh-azwiegincew:
This now looks complete – just needs a test in tests/ast.

tests/thin-client/modin-steel-thread.py
```diff
@@ -0,0 +1,16 @@
...
+import snowflake.snowpark.modin.plugin  # noqa: F401
+from snowflake.snowpark import Session
+
+session = Session.builder.create()
+df = pd.DataFrame(
+    [[1, 2, 3, 4, 5, 6, 7], [8, 9, 10, 11, 12, 13, 14], [15, 16, 17, 18, 19, 20, 21]],
+    columns=["A", "B", "C", "D", "E", "F", "G"],
+)
+result = df.iloc[2, 2]
+print(result)
```
sfc-gh-vbudati, line 16: This test fails right now - I am not sure how to display `result`. It does not produce any other output.

src/snowflake/snowpark/modin/pandas/dataframe.py
```diff
@@ -158,6 +159,11 @@ def __init__(
 
         self._siblings = []
 
+        if not ast_stmt:
```
sfc-gh-azwiegincew, line 162: In column.py we're using if ... is None; we should use a consistent pattern for this check – whatever is the canonical approach in Python. It seems that `is None` makes more sense?

src/snowflake/snowpark/modin/pandas/dataframe.py
```diff
@@ -158,6 +159,11 @@ def __init__(
...
+            ast_stmt = pd.session._ast_batch.assign()
```
sfc-gh-azwiegincew, line 163: I don't really understand the point of creating an `assign` node like this. If a statement isn't actually supplied, how a "default" one won't really work. Is this because of the way that Pandas interacts with this code?

tests/ast/data/pd-iloc.test
```diff
@@ -0,0 +1,17 @@
...
+df = pd.DataFrame(
+    [[1, 2, 3, 4, 5, 6, 7], [8, 9, 10, 11, 12, 13, 14], [15, 16, 17, 18, 19, 20, 21]],
+    columns=["A", "B", "C", "D", "E", "F", "G"],
+)
+result = df.iloc[2, 2]
+print(result)
+
+## EXPECTED OUTPUT
+
+17
```
sfc-gh-azwiegincew, line 17: Does this work? The test driver here runs the unparser, so we should be seeing something that looks like the original source.

tests/thin-client/modin-steel-thread.py
sfc-gh-azwiegincew: Does it work with the main branch? I assume you're running without Phase 1 enabled?

sfc-gh-azwiegincew (changes requested):
LG; some minor nits. The change request is about the AST test.

PR #1735, opened 2024-06-04
[feature branch] Phase 0 client

sfc-gh-azwiegincew:
As of this diff, we continue to pursue Phase 0 and Phase 1 client on top of the same codebase to minimize development overhead. `ServerConnection.phase1_enabled()` can be used to branch based on the phase.

Phase 0 is the default and Phase 1 is enabled via the env var `SNOWPARK_PHASE_1=1`.

This diff makes Phase 0 behave as it's going to appear in the actual design. Phase 1 is still a total hack which issues a parallel request just to test the TCM. We will need to make the Phase 1 code path more real soon as well – in order to enable a proper end-to-end test.

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -696,6 +696,7 @@ class ServerConnection:
         )
 
     def create_coprocessor(self) -> None:
+        """Invoked during session initialization to create a server-side coprocessor."""
         id = str(uuid.uuid4())
         print(f"create cp start rid={id}")
         self._conn._rest.request(
@@ -706,6 +707,8 @@ class ServerConnection:
         time.sleep(5)  # TODO: need to send a response once the TCM is created.
         print("create cp complete")
 
+    # TODO: This function is currently invoked to prototype Phase 1 while maintaining most of the code
+    # targeting Phase 0. This function will likely disappear once Phase 1 is placed in a separate branch.
     def ast_query(self, request_id__ast) -> Any:
         (request_id, ast) = request_id__ast
         req = {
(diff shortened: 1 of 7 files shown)


sfc-gh-oplaton (approved):

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -715,6 +718,9 @@ def ast_query(self, request_id__ast) -> Any:
         return self._conn._rest.request(
             f"/queries/v1/query-request?requestId={request_id}", req
         )
+    
+    def phase1_enabled(self) -> bool:
```
sfc-gh-lspiegelberg, line 722: nit: `is_phase1_enabled`

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -800,6 +800,9 @@ def create_coprocessor(self):
         # It's not necessary to mock this call.
         pass
 
+    def phase1_enabled(self):
```
sfc-gh-lspiegelberg, line 803: nit: can we add a comment that mock conneciton is not relevant here? Therefore always return False?

tests/ast/test_ast_driver.py
```diff
@@ -60,10 +58,11 @@ def render(ast_base64: str) -> str:
     """Uses the unparser to render the AST."""
     assert pytest.unparser_jar, "A valid Unparser JAR path must be supplied either via --unparser-jar=<path> or the environment variable SNOWPARK_UNPARSER_JAR"
     res = subprocess.run([
-        UNPARSER_SCRIPT,
```
sfc-gh-lspiegelberg, line 63: which java version do we require? Maybe good to have a note somewhere, as this is invoked client-side.

sfc-gh-lspiegelberg (approved):
lgtm!

tests/ast/test_ast_driver.py
sfc-gh-azwiegincew: I believe Scala requires JDK 3. I don't think there's any danger here, although it's possible one of the dependencies is stricter. Let's see what happens when we try to pass CI/CD.

PR #1736, opened 2024-06-05
CI: ignore some not implemented methods code coverage

sfc-gh-azhan:
CI: ignore some not implemented methods code coverage

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/docstrings/series_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/docstrings/series_utils.py
@@ -7,13 +7,13 @@
 
 class StringMethods:
     def casefold():
-        pass
+        pass  # pragma: no cover
 
     def cat():
-        pass
+        pass  # pragma: no cover
 
     def decode():
-        pass
+        pass  # pragma: no cover
 
     def split():
         """
@@ -104,7 +104,7 @@ class StringMethods:
         """
 
     def rsplit():
-        pass
+        pass  # pragma: no cover
 
     def get():
         """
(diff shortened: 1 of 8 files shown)


src/snowflake/snowpark/modin/plugin/docstrings/series_utils.py
```diff
@@ -7,13 +7,13 @@
 
 class StringMethods:
     def casefold():
-        pass
+        pass  # pragma: no cover
```
sfc-gh-mvashishtha, line 10: Should we skip coverage for the entire classes instead, as [here](https://github.com/snowflakedb/snowpark-python/blob/903eee29597687ac64d4041c903c106d7a3e84a8/src/snowflake/snowpark/modin/plugin/docstrings/base.py#L388)?

sfc-gh-helmeleegy: +1. This is what we do for all other object types, not just base. So better (and easier) to stay consistent.

sfc-gh-azhan: Yes, we skipped all in tox.ini. PTAL

tox.ini
```diff
@@ -1,16 +1,18 @@
 [coverage:report]
 skip_covered = False
 show_missing = True
-# TODO: SNOW-1453027 remove omit
+# TODO: SNOW-1453027 remove omit for config and core
 omit = */src/snowflake/snowpark/modin/config/*
        */src/snowflake/snowpark/modin/core/*
+       */src/snowflake/snowpark/modin/plugin/docstrings/*
```
sfc-gh-azhan, line 7: @sfc-gh-helmeleegy we can just skip it here.

sfc-gh-helmeleegy: Got it.

sfc-gh-helmeleegy (approved):
LGTM, thanks for taking carte of this!

sfc-gh-nkumar (approved):

PR #1741, opened 2024-06-05
SNOW-1461192: Fix StringMethods and DatetimeProperties docs

sfc-gh-joshi:

   

   Fixes SNOW-1461192

SNOW-1445536 removed our copies of the `StringMethods` and `DatetimeProperties` classes in favor of their upstream modin versions. After this PR, doctests (run via `pytest src/snowflake/snowpark/modin/plugin/docstrings/series_utils.py`) still passed, but the `__doc__` attributes of these classes' methods were not correctly updated since we did not use `modin.config.DocModule` (only `snowflake.snowpark.modin.config.DocModule`).

This PR sets modin's `DocModule` attribute so the docstrings are appropriately overridden, and changes some module initialization order to accommodate this change.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/__init__.py
+++ b/src/snowflake/snowpark/modin/plugin/__init__.py
@@ -51,18 +51,23 @@ warnings.warn(
     stacklevel=1,
 )
 
-# We need this import here to prevent circular dependency issues, since snowflake.snowpark.modin.pandas
-# currently imports some internal utilities from snowflake.snowpark.modin.plugin. Test cases will
-# import snowflake.snowpark.modin.plugin before snowflake.snowpark.modin.pandas, so in order to prevent
-# circular dependencies from manifesting, apparently snowflake.snowpark.modin.pandas needs to
-# be imported first.
-# These imports also all need to occur after modin + pandas dependencies are validated.
-from snowflake.snowpark.modin import pandas  # isort: skip  # noqa: E402,F401
+from modin.config import DocModule as ModinDocModule  # isort: skip  # noqa: E402
 from snowflake.snowpark.modin.config import DocModule  # isort: skip  # noqa: E402
 from snowflake.snowpark.modin.plugin import docstrings  # isort: skip  # noqa: E402
 
+ModinDocModule.put(docstrings.__name__)
 DocModule.put(docstrings.__name__)
 
+# We need to import snowflake.snowpark.modin.pandas here to prevent circular dependency issues, since
+# snowflake.snowpark.modin.pandas currently imports some internal utilities from
+# snowflake.snowpark.modin.plugin. Test cases will import snowflake.snowpark.modin.plugin before
+# snowflake.snowpark.modin.pandas, so in order to prevent circular dependencies from manifesting,
+# apparently snowflake.snowpark.modin.pandas needs to be imported first.
+# Furthermore, this initialization needs to occur after dependency versions validation and
+# after documentation overrides, as ModinDocModule.put will produce a call to `importlib.reload`
+# that will overwrite our extensions.
+# See https://github.com/modin-project/modin/issues/7122
+from snowflake.snowpark.modin import pandas  # isort: skip  # noqa: E402,F401
 
 # Don't warn the user about our internal usage of private preview pivot
 # features. The user should have already been warned that Snowpark pandas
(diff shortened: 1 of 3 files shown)


sfc-gh-mvashishtha (approved):
Thanks!

sfc-gh-azhan (commented):
We are not showing any implemented APIs from src/snowflake/snowpark/modin/plugin/docstrings/series_utils.py in our doc. Can you help to fix it too?

sfc-gh-joshi:
@sfc-gh-azhan It seems like this is nontrivial to do since sphinx isn't currently autogenerating docstrings for those Series.str or Series.dt. Since this isn't a regression (as far as I can tell, the docs haven't been generated since before we migrated from the snowpandas repo), I'll follow up with it later (SNOW-1464571).

sfc-gh-azhan (approved):

sfc-gh-dpetersohn (approved):
LGTM!

sfc-gh-helmeleegy (approved):
LGTM. Thanks for the fix!

sfc-gh-joshi:
I'm having some struggles getting doctests to work properly with this new import order. Running individual doctests (for example, `pytest pytest src/snowflake/snowpark/modin/pandas/base.py::snowpark.modin.pandas.base.BasePandasDataset.all`) has no issues, but running the whole suite of doctests like CI does (`pytest -m doctest src/snowflake/snowpark/modin/ tests/unit/modin -x`) seems to cause an `importlib.reload` somewhere that overwrites our loaded modin extensions. Moreover, running just `pytest -m doctest src/snowflake/snowpark/modin/` (not specifying the unit tests) causes no tests to be run at all, so it's very confusing what order the various conftest + module initialization modules are run in.

I'll continue investigating to see if there's a fix for this, otherwise we might have to find a more primitive workaround to avoid incurring `importlib.reload`. This may be a bug that needs to be fixed upstream.

sfc-gh-yzou:
@sfc-gh-joshi does that mean if customer runs importlib.reload in their notebook, it could potentially cause problem to their workload also?

sfc-gh-joshi:
@sfc-gh-yzou Yes, but that's kind of an unavoidable architectural consequence of the extension plugin system. If a user calls `importlib.reload(modin)` or `importlib.reload(modin.pandas)` (or on some other submodule of `modin`), then the reloaded version of modin will have the dictionaries tracking our registered extension methods erased. Calling `importlib.reload(snowflake.snowpark.modin.plugin)` afterwards would probably fix this, though I haven't tested this interaction.

From what I can tell, the issue here is occurring because when we call modin's `DocModule.put`, it calls `importlib.reload` [here](https://github.com/modin-project/modin/blob/af5ed0628a4a609f70becd29fd081a290ba981a0/modin/config/envvars.py#L791-L802) in order to rerun all of modin's `_inherit_docstrings` annotations with the newly specified docstring classes as parents. In a REPL or in our other tests, this isn't an issue since we deliberately run `import snowflake.snowpark.modin.pandas` for the first time only _after_ `DocModule.put` is called, which means the module reload occurs before any extensions are loaded. However, it seems like something in the way our tox doctest environment is set up breaks this order, and the reload occurs _after_ extensions are loaded.

sfc-gh-joshi:
The fix I've settled on is to manually call `modin.utils._inherit_docstrings` for `StringMethods` and `DatetimeProperties`. Since these are the only two classes that we pull directly from upstream modin, I've elected to hardcode this for the time being; we can revisit a more robust solution as we remove more frontend classes from our fork. This was the only relatively simple solution I tried that worked, and since the cause is `importlib.reload` calls from upstream modin, I think it is much better to keep this simple workaround until `importlib.reload` calls are removed upstream.

I've asked @sfc-gh-mvashishtha to merge a fix for the offending issue (https://github.com/modin-project/modin/issues/7138) some time this week, which will hopefully become available in the next modin release. Until then, we should continue to directly call `modin.utils._inherit_docstrings` as a workaround.

src/snowflake/snowpark/modin/plugin/__init__.py
```diff
@@ -64,6 +64,22 @@
 DocModule.put(docstrings.__name__)
 
 
+# We cannot call ModinDocModule.put directly because it will produce a call to `importlib.reload`
+# that will overwrite our extensions. We instead directly call the _inherit_docstrings annotation
+# See https://github.com/modin-project/modin/issues/7122
+import modin.utils  # type: ignore[import]  # isort: skip  # noqa: E402
+import modin.pandas.series_utils  # type: ignore[import]  # isort: skip  # noqa: E402
+
+modin.utils._inherit_docstrings(
```
sfc-gh-mvashishtha, line 73: good solution!

sfc-gh-mvashishtha (approved):

PR #1742, opened 2024-06-05
SNOW-1019479: Fix describe with duplicate column names

sfc-gh-joshi:

   

   Fixes SNOW-1019479

This PR fixes a bug with `df.describe()` if `df` had columns with the same name but different dtypes. Since the method computes separate query compilers for the statistics on `object`, numeric, and `datetime` columns and concats them all together, if two columns shared a label but had different dtypes, it would try to concat them together and either error or produce an incorrect result. This PR works around this by relabeling the columns with ordinal values before performing any aggregations, and restoring the labels after concatenation.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -9999,7 +9999,18 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         # max            NaN      3.0    NaN
         sorted_percentiles = sorted(percentiles)
         dtypes = self.dtypes
-        query_compiler = self
+        # If we operate on the original frame's labels, then if two columns have the same name but
+        # different dtypes, the JOIN behavior of SnowflakeQueryCompiler.concat will produce incorrect
+        # results. For example, if a frame has two columns named "a", one with `object` dtype and the
+        # other `int64`, the `object` column would have a query compiler for the computed `top`/`freq`,
+        # while the numeric column would have a query compiler for `std` and other numeric aggregations.
+        # We want our final `concat` call that combines query compilers to insert NULL values for
+        # uncomputed statistics, but since `concat` tries to match labels, it would incorrectly try
+        # to combine the `top`/`freq`/`std`/(other numeric) rows into a single column.
+        # To circumvent this, we relabel all columns with a simple integer index, and restore the
+        # correct labels at the very end after `concat`.
+        original_columns = self.columns
+        query_compiler = self.set_columns(list(range(len(self.columns))))
         internal_frame = query_compiler._modin_frame
         # Compute count for all columns regardless of dtype
         query_compilers_to_concat = [
@@ -10323,10 +10334,14 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         assert (
             len(query_compilers_to_concat) > 1
         ), "must have more than one QC to concat"
-        return query_compilers_to_concat[0].concat(
-            other=query_compilers_to_concat[1:],
-            axis=0,
-            join="outer",
+        return (
+            query_compilers_to_concat[0].concat(
+                other=query_compilers_to_concat[1:],
+                axis=0,
+                join="outer",
+            )
+            # Restore the original pandas labels
+            .set_columns(original_columns)
         )
 
     def sample(
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -10001,14 +10001,57 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         dtypes = self.dtypes
         # If we operate on the original frame's labels, then if two columns have the same name but
         # different dtypes, the JOIN behavior of SnowflakeQueryCompiler.concat will produce incorrect
-        # results. For example, if a frame has two columns named "a", one with `object` dtype and the
-        # other `int64`, the `object` column would have a query compiler for the computed `top`/`freq`,
-        # while the numeric column would have a query compiler for `std` and other numeric aggregations.
-        # We want our final `concat` call that combines query compilers to insert NULL values for
-        # uncomputed statistics, but since `concat` tries to match labels, it would incorrectly try
-        # to combine the `top`/`freq`/`std`/(other numeric) rows into a single column.
+        # results. For example, consider the following dataframe, where an `object` column and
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9999,7 +9999,18 @@ def describe(
         # max            NaN      3.0    NaN
         sorted_percentiles = sorted(percentiles)
         dtypes = self.dtypes
-        query_compiler = self
+        # If we operate on the original frame's labels, then if two columns have the same name but
+        # different dtypes, the JOIN behavior of SnowflakeQueryCompiler.concat will produce incorrect
+        # results. For example, if a frame has two columns named "a", one with `object` dtype and the
+        # other `int64`, the `object` column would have a query compiler for the computed `top`/`freq`,
+        # while the numeric column would have a query compiler for `std` and other numeric aggregations.
```
sfc-gh-yzou, line 10006: hmm, is that because the other non-numeric columns got dropped during std ? if you have two df with both labels [a, a, c], the concat should be able to take care of it correctly.

tests/integ/modin/frame/test_describe.py
```diff
@@ -308,26 +308,13 @@ def test_describe_multiindex(index, columns, include, expected_union_count):
...
 
 
-DUP_COL_FAIL_REASON = "SNOW-1019479: describe on frames with mixed object/number columns with the same name fails"
-
-
 @pytest.mark.parametrize(
     "include, exclude, expected_union_count",
     [
         (None, None, 7),
-        pytest.param(
```
sfc-gh-yzou, line 318: do we have a test that one column or more columns is just all None ?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9999,7 +9999,18 @@ def describe(
...
+        # We want our final `concat` call that combines query compilers to insert NULL values for
+        # uncomputed statistics, but since `concat` tries to match labels, it would incorrectly try
+        # to combine the `top`/`freq`/`std`/(other numeric) rows into a single column.
+        # To circumvent this, we relabel all columns with a simple integer index, and restore the
```
sfc-gh-helmeleegy, line 10010: This looks like a simple yet effective solution. I agree with Yun that a small example would help here.

sfc-gh-helmeleegy (approved):
LGTM. Nice fix!

sfc-gh-dpetersohn (approved):

tests/integ/modin/frame/test_describe.py
sfc-gh-joshi: A few tests have parameters that have all-None columns in the data (`test_describe_numeric_only`, `test_describe_obj_only`), but I didn't include any specifically for this test. Do you think it's worth adding?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: Yes, the numeric aggregations will drop non-numeric columns, so previously if you had labels `["a", "a", "c"]` but both "a" columns were numeric, there were issues. There were only problems when there was a mix of object and numeric aggregations, since we had to drop columns with inappropriate dtypes before performing the aggregations.

I've updated the comment example to be more concrete.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9999,7 +9999,61 @@ def describe(
         # max            NaN      3.0    NaN
         sorted_percentiles = sorted(percentiles)
         dtypes = self.dtypes
-        query_compiler = self
+        # If we operate on the original frame's labels, then if two columns have the same name but
+        # different dtypes, the JOIN behavior of SnowflakeQueryCompiler.concat will produce incorrect
```
sfc-gh-yzou, line 10003: I want to make sure my understanding is correct, it is not actual an problem when the both columns have different dtypes, it is when one is numeric, and another is non-numeric, because the applicable function on the columns are different, one is top/freq, another one is min/max etc. If there are both numeric, but just different type, i imaging it should be ok, because the min, max df will have both columns in the query compiler, and concat should be able to take care of it correctly.

tests/integ/modin/frame/test_describe.py
sfc-gh-yzou: i think what i would be really interested is see what happens if a we have a column that have float numbers, but configured as object type, and if a column have mixed float, int and None marked as object what happens there.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9999,7 +9999,61 @@ def describe(
...
+        #              |  top | NaN |  'x' | NaN |  'j' |
+        #              +------+-----+------+-----+------+
+        #              | freq | NaN |   2  | NaN |   2  |
+        #              +------+-----+------+-----+------+
+        #              |  min |  1  | None |  3  | None |
+        #              +------+-----+------+-----+------+
+        #              |  max |  3  | None |  5  | None |
+        #              +------+-----+------+-----+------+
+        original_columns = self.columns
+        query_compiler = self.set_columns(list(range(len(self.columns))))
```
sfc-gh-yzou, line 10056: The fix here looks fine to me! I think i mainly want to make sure i got the issue here correctly

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: Yes, your understanding is correct. We have test cases verifying this as well.

tests/integ/modin/frame/test_describe.py
sfc-gh-joshi: The columns included in each set of aggregations are determined by the dtype of the column, not the type of the objects themselves. A column of float or mixed numeric/NULL values with `object` dtype would be treated the same as one with only non-numeric ones (top/freq are computed, std/min/max/etc. are not).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: can you update the comment then, it is kind if misleading

sfc-gh-yzou (approved):

PR #1744, opened 2024-06-06
SNOW-1460922: [Local Testing] Refactor mocked function calls to be more flexible

sfc-gh-jrose:

   

   Fixes SNOW-1460922

  This change refactors how mock functions are registered and called. The main goal of the change is to remove all special cases from the `handle_function_expression` and to instead provide API flexibility so that functions that would require similar special cases in the future can instead set appropriate flags so that they are handled correctly.

Previous special cases and the new way to handle them include:

- `count` and `sum` requiring a special check to see if the expression expects distinct values or not. Now when calling patch you can specify an alternate function to be used when a distinct result is desired. eg. `@patch("sum", distinct=mock_sum_distinct)`
- `row_number` and future window functions required their row index when computing. Now when calling the pass_row_index and the row index will be passed in via a required kwarg.
- Previously for functions like `current_time` that don't take an input and return a constant output we would need to guess that the constant return was intended to be the same value for all rows. Now you can specify the `pass_column_index` parameter to patch and use that to calculate the correct number of values to generate. I've left the previous guessing logic in just in case a customer had already implemented a constant return function, but it will only be used if the new index parameters are not used.
- I've removed the special case for `array_agg`. This function was not actually implemented and only referenced by a single test that only tested that the special case for `array_agg` worked. 

For an example of the type of function this PR will enable in the future see here:
https://github.com/snowflakedb/snowpark-python/pull/1745


tests/mock/test_functions.py
```diff
@@ -289,3 +294,22 @@ def test_to_char_is_row_index_agnostic(session):
     assert df.filter(col("a") > 3).select(to_char(col("a")), col("b")).collect() == [
         Row("5", 6)
     ]
+
+
+@pytest.mark.skipif(
+    "not config.getoption('local_testing_mode', default=True)",
+    reason="Only test local testing code in local testing mode.",
+)
+def test_function_mock_and_call_neg(session):
```
sfc-gh-yuwang, line 303: Hi, I do not have a tons of knowledge here, but as far as I see it, there is no tests for register() and unregister() function right? Can we add some unit tests for new functions in this PR?

sfc-gh-aalam (approved):
looks good

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -67,6 +66,120 @@
...
+    ) -> None:
+        self.name = name
+        self.impl = func_implementation
+        self.distinct = distinct or self
+        self._pass_row_index = pass_row_index
+        self._pass_column_index = pass_column_index
+        self._pass_input_data = pass_input_data
+
+    def _check_constant_result(self, input_data, args, result):
+        # This function helps automaticallly fill a column with a constant value in certain
```
sfc-gh-stan, line 141: NIT: Use doctstring instead of multiple one line comments

sfc-gh-stan (approved):
Code looks good to me, thank you for proposing and implementing this interface! I belive this makes supporting rank related window functions much easier. 
We should probably add a section in the documentation regarding the new way of patching: https://docs.snowflake.com/en/developer-guide/snowpark/python/testing-locally#known-issues /

src/snowflake/snowpark/mock/_functions.py
```diff
@@ -67,6 +66,119 @@
...
+    def __init__(self) -> None:
+        self._registry = dict()
+
+    @classmethod
+    def get_or_create(cls):
+        if cls._instance is None:
+            cls._instance = MockedFunctionRegistry()
+        return cls._instance
+
+    def get_function(self, func: Union[Expression, str]):
```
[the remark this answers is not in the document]
sfc-gh-stan: Also return type?

PR #1752, opened 2024-06-10
Index: decorator to handle lazy and non-lazy (columns) case

sfc-gh-samdesai1:

   

   Fixes SNOW-NNNNNNN

 create instance of non-lazy index for df.columns and add a decorator to support methods for lazy and non-lazy index


src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -389,6 +389,7 @@ def data_columns_index(self) -> "pd.Index":
                 # otherwise, when labels are tuples (e.g., [("A", "a"), ("B", "b")]),
                 # a MultiIndex will be created incorrectly
                 tupleize_cols=False,
+                convert_to_lazy=False,
```
sfc-gh-azhan, line 392: can you also fix the comment above line 375:
"""
Returns Snowpark pandas Index object for column index (df.columns).
Note this object is still hold an internal pandas index (i.e., not lazy) to avoid unnecessary pulling data from Snowflake.
"""

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -82,11 +82,14 @@ def __init__(
         copy: bool = False,
         name: object = None,
         tupleize_cols: bool = True,
+        convert_to_lazy: bool = True,
```
sfc-gh-azhan, line 85: can we add comments for all init arguments ?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +106,41 @@ def __init__(
                 tupleize_cols=tupleize_cols,
             )
 
+    def to_pandas_warning(self, func_name: str) -> None:
+        """
+        Helper method to notify users if they are using a method that currently calls to_pandas()
+        """
+        WarningMessage.single_warning(
+            f"The method {func_name} currently calls to_pandas() and materializes data. In future updates, this method will be lazily evaluated"
```
sfc-gh-azhan, line 114:
```suggestion
            f"The Index method {func_name} currently calls to_pandas() and materializes data. In future updates, this method will be lazily evaluated"
```

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +106,41 @@ def __init__(
...
+        )
```
sfc-gh-nkrishna, line 115: Can we move this method be part of the WarningMessage class in warning_message.py instead of here?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +106,41 @@ def __init__(
...
+
+    def is_lazy_check(func: Any) -> Any:
```
sfc-gh-azhan, line 117: add comment to describe what this function does

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +106,41 @@ def __init__(
...
+        def check_lazy(*args: Any, **kwargs: Any) -> Any:
+            func_name = func.__name__
+            if args[0].is_lazy:
+                returned_value = func(*args, **kwargs)
+                args[0].to_pandas_warning(func_name)
```
sfc-gh-azhan, line 122: We should not warn here and keep the original warning in each method. Once we implement one method, we can remove the warning there. If you put warning here, how do we remove it one by one?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +106,41 @@ def __init__(
...
+                return returned_value
+            else:
+                native_index = args[0]._index
+                native_func = getattr(native_index, func_name)
+                if not callable(native_func):
+                    return native_func
```
sfc-gh-azhan, line 128: add comment with an example

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +106,41 @@ def __init__(
...
+                args = args[1:]
```
sfc-gh-azhan, line 129: add comment why did this

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +106,41 @@ def __init__(
...
+                returned_value = native_func(*args, **kwargs)
+                if isinstance(returned_value, native_pd.Index):
```
sfc-gh-azhan, line 131: add comment why did this and also give an example

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -213,7 +253,38 @@ def has_duplicates(self) -> bool:
         """
         return not self.is_unique
 
+    @is_lazy_check
+    def unique(self, level: Hashable | None = None) -> Index:
```
sfc-gh-azhan, line 257: why add a new method in this PR?

sfc-gh-azhan (changes requested):

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: even the native pandas ones?

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: ok makes sense, I will change it

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: I am not sure, but this method was giving me an error [here](https://github.com/snowflakedb/snowpark-python/blob/64c5b848b6242b3361223a0e17d812eee3dcde08/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py#L5413) without the method

sfc-gh-azhan: can you run `test_concat_verify_integrity_axis1_negative1` in debug mode and set a breakpoint there and figure out why it does not fail before but now?

sfc-gh-samdesai1: Yep, the problem happens because of the original __getitem__ method sometimes slicing an index and returning a pandas index when it should be returning a modin index. Thus, `qc.columns[qc.columns.duplicated()]` from the link above was returning a pandas index method, which does have a unique attribute, so the tests were passing. When I added the decorator, I also added functionality to convert native pandas index outputs to modin indexes, which is why the test is failing now. As a result, I think we actually do need this method and I will keep it here unless you want me to do something different

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -372,9 +372,8 @@ def is_unnamed_series(self) -> bool:
     @property
     def data_columns_index(self) -> "pd.Index":
         """
-        Returns pandas Index object for column index (df.columns).
-        We can't do the same thing for df.index here because it requires pulling
-        the data from snowflake and filing a query to snowflake.
+        Returns Snowpark pandas Index object for column index (df.columns).
+        Note this object is still hold an internal pandas index (i.e., not lazy) to avoid unnecessary pulling data from Snowflake.
```
sfc-gh-nkrishna, line 376:
```suggestion
        Note this object will still hold an internal pandas index (i.e., not lazy) to avoid unnecessary pulling data from Snowflake.
```

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -82,11 +45,52 @@ def __init__(
...
+        tupleize_cols : bool (default: True)
+            When True, attempt to create a MultiIndex if possible.
+        convert_to_lazy : bool (default: True)
+            When True, create a lazy index object, otherwise, create an index object that locally saves a pandas index
+            We set convert_to_lazy as False when calling df.columns
+
+        Notes
+        -----
+        An Index instance can **only** contain hashable objects.
+        An Index instance *can not* hold numpy float16 dtype.
```
sfc-gh-nkrishna, line 75:
```suggestion
        An Index instance *cannot* hold numpy float16 dtype.
```

sfc-gh-nkrishna (commented):
Two more minor grammar nits

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -82,11 +45,52 @@ def __init__(
...
+            inferred from `data`.
+            See the :ref:`user guide <basics.dtypes>` for more usages.
+        copy : bool, default False
+            Copy input data.
+        name : object
+            Name to be stored in the index.
+        tupleize_cols : bool (default: True)
+            When True, attempt to create a MultiIndex if possible.
+        convert_to_lazy : bool (default: True)
+            When True, create a lazy index object, otherwise, create an index object that locally saves a pandas index
```
sfc-gh-azhan, line 69: When True, create a lazy index object from a local `data` input, otherwise, create an index object that saves a pandas index locally.
We only set convert_to_lazy as False to avoid pulling data back and forth from Snowflake, e.g., when calling df.columns, the column data should always be kept locally.

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -82,11 +45,52 @@ def __init__(
...
+        Index(['a', 'b', 'c'], dtype='object')
+
+        >>> pd.Index([1, 2, 3], dtype="uint8")
+        Index([1, 2, 3], dtype='uint8')
+        """
         from snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler import (
             SnowflakeQueryCompiler,
         )
 
+        self.is_lazy = convert_to_lazy
```
sfc-gh-azhan, line 92: nit: move to line 94

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +107,49 @@ def __init__(
...
+                # If the index is not lazy, get the cached native index and call the function
+                native_index = args[0]._index
+                native_func = getattr(native_index, func_name)
+
+                # If the function is a property, we will get a non-callable, so we just return it
+                # Examples of this are values or dtype
+                if not callable(native_func):
+                    return native_func
+
+                # Remove the first argument in args, because it is self and we don't need it
```
sfc-gh-azhan, line 132:
```suggestion
                # Remove the first argument in args, because it is `self` and we don't need it
```

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +107,49 @@ def __init__(
...
+                args = args[1:]
+                returned_value = native_func(*args, **kwargs)
+
+                # If we return a native Index, we need to convert this to a modin index
+                # Examples of this are astype and copy
```
sfc-gh-azhan, line 137:
```suggestion
                # Examples of this are `astype` and `copy`
```

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +107,49 @@ def __init__(
...
+                # If the function is a property, we will get a non-callable, so we just return it
+                # Examples of this are values or dtype
+                if not callable(native_func):
+                    return native_func
+
+                # Remove the first argument in args, because it is self and we don't need it
+                args = args[1:]
+                returned_value = native_func(*args, **kwargs)
+
+                # If we return a native Index, we need to convert this to a modin index
```
sfc-gh-azhan, line 136:
```suggestion
                # If we return a native Index, we need to convert this to a modin index but keep it locally.
```

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -103,6 +107,49 @@ def __init__(
...
+                # Examples of this are astype and copy
+                if isinstance(returned_value, native_pd.Index):
+                    returned_value = Index(returned_value, convert_to_lazy=False)
+                # Some methods also return a tuple with an Index, so convert that tuples first item to an index
+                # Examples of this are _get_indexer_strict and sort_values
```
sfc-gh-azhan, line 141:
```suggestion
                # Examples of this are `_get_indexer_strict` and `sort_values`
```

sfc-gh-nkrishna (approved):
LGTM pending Andong's changes

sfc-gh-vbudati (approved):
Looks great, thanks Samedh!

sfc-gh-azhan (approved):

PR #1757, opened 2024-06-11
Add Index.T

sfc-gh-vbudati:

   

   Fixes SNOW-1458140

   Implemented `Index.T` which by definition just returns `self`. Added one simple test.

### DIFF
--- a/tests/integ/modin/index/test_transpose.py
+++ b/tests/integ/modin/index/test_transpose.py
@@ -21,5 +21,4 @@ NATIVE_INDEX_TEST_DATA = [
 @pytest.mark.parametrize("native_index", NATIVE_INDEX_TEST_DATA)
 def test_index_transpose(native_index):
     snow_index = pd.Index(native_index)
-    t_index = snow_index.T
-    assert snow_index.equals(t_index) is True
+    assert snow_index.equals(snow_index.T)
--- a/tests/integ/modin/index/test_transpose.py
+++ b/tests/integ/modin/index/test_transpose.py
@@ -1,0 +1,25 @@
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
+
+import modin.pandas as pd
+import pandas as native_pd
+import pytest
+
+import snowflake.snowpark.modin.plugin  # noqa: F401
+from tests.integ.modin.sql_counter import sql_count_checker
+
+NATIVE_INDEX_TEST_DATA = [
+    native_pd.Index([], dtype="object"),
+    native_pd.Index([1, 2, 3]),
+    native_pd.Index([["a", "b"], [1, 2]]),
+]
+
+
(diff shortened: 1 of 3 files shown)


tests/integ/modin/index/test_transpose.py
```diff
@@ -0,0 +1,25 @@
...
+    native_pd.Index([["a", "b"], [1, 2]]),
+]
+
+
+@sql_count_checker(query_count=0)
+@pytest.mark.parametrize("native_index", NATIVE_INDEX_TEST_DATA)
+def test_index_transpose(native_index):
+    snow_index = pd.Index(native_index)
+    t_index = snow_index.T
+    assert snow_index.equals(t_index) is True
```
sfc-gh-nkrishna, line 25: Small suggestion simplifying this:
```suggestion
    assert snow_index.equals(snow_index.T)
```

tests/integ/modin/index/test_transpose.py
```diff
@@ -0,0 +1,25 @@
...
+from tests.integ.modin.sql_counter import sql_count_checker
+
+NATIVE_INDEX_TEST_DATA = [
+    native_pd.Index([], dtype="object"),
+    native_pd.Index([1, 2, 3]),
+    native_pd.Index([["a", "b"], [1, 2]]),
+]
+
+
+@sql_count_checker(query_count=0)
```
sfc-gh-joshi, line 20: Why are there no queries being issued? Do we not need to materialize any data for the equality check?

sfc-gh-nkrishna: I think it's b/c we are returning self so the equality is just comparing the object hash and doesn't actually need to materialize the result

sfc-gh-dpetersohn (approved):
LGTM

tests/integ/modin/index/test_transpose.py
sfc-gh-joshi: I think we should force a value comparison rather than just comparing the references with something like `assert pd.Index(native_index).equals(pd.Index(native_index).T)` instead. I know our objects are immutable, but I can imagine a hypothetical incorrect implementation of `T` that mutates the underlying query compiler but leaves the reference to the wrapping `Index` object the same, so we should make sure we're actually materializing the data for this comparison.

sfc-gh-samdesai1: This will issue a query or two when my implementation is merged because we need to call `to_pandas()` to check equality and that will require a query. It does not require one now because we store a native pandas index locally

sfc-gh-vbudati: Sure, I can make that change

sfc-gh-joshi: That makes sense, LGTM in that case.

sfc-gh-joshi (approved):

sfc-gh-nkrishna (approved):

PR #1761, opened 2024-06-12
SNOW-1445342: Support `DataFrame.expanding/rolling` and `Series.expanding/rolling` for aggregations `count`, `sum`, `min`, `max`, `mean`, `std`, and `var` with `axis=0`

sfc-gh-nkrishna:

   

   SNOW-1445321, SNOW-1445327, SNOW-1445331, SNOW-1445333, SNOW-1445338, SNOW-1445341, SNOW-1445342, SNOW-1445344

   This PR adds support for `DataFrame.expanding` and `Series.expanding` for aggregations `count`, `sum`, `min`, `max`, `mean`, `std`, and `var` with `axis=0`. It also adds support for `DataFrame.rolling` and `Series.rolling` for aggregation `count`.


src/snowflake/snowpark/modin/pandas/window.py
```diff
@@ -121,7 +121,7 @@ def __init__(
     ) -> None:
         # TODO: SNOW-1063358: Modin upgrade - modin.pandas.window.Rolling
         # Raise ValueError when invalid parameter values/combinations
-        if (isinstance(window, int) and window <= 0) or window is None:
+        if (isinstance(window, int) and window < 0) or window is None:
```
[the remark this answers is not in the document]
sfc-gh-dpetersohn: +1. Would you open an accompanying issue in Modin and link here?

sfc-gh-dpetersohn (commented):
Look great!

src/snowflake/snowpark/modin/pandas/window.py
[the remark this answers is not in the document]
sfc-gh-nkrishna: @sfc-gh-dpetersohn This isn't a Modin issue (Modin handles these cases correctly), but Modin just doesn't do this error checking in the frontend layer: https://github.com/modin-project/modin/blob/main/modin/pandas/window.py#L99. I added these checks at the frontend for Snowpark pandas, should I file this as a feature request or leave it be?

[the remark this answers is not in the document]
sfc-gh-yzou: Your change is about expanding, could you move things related to rolling to a different PR. If that is just a case that Snowpark pandas does not handle yet, I don't think it should stay at frontend layer (the error message also doesn't match the check), should be at the snowpark pandas backend layer. Also, what happens when rolling is 0, is it just require a different implementation to handle it?

src/snowflake/snowpark/modin/plugin/_internal/window_utils.py
```diff
@@ -101,3 +100,37 @@ def check_is_rolling_window_supported_by_snowflake(
...
+    expanding_kwargs: keyword arguments passed to expanding. The expanding keywords handled in the
+        function contains:
+        min_periods: int, default 1.
+            Minimum number of observations in window required to have a value; otherwise, result is np.nan.
+        axis: int or str, default 0
+            If 0 or 'index', roll across the rows.
+            If 1 or 'columns', roll across the columns.
+            For Series this parameter is unused and defaults to 0.
+        method: str {‘single’, ‘table’}, default ‘single’
+            **This parameter is ignored in Snowpark pandas since the execution engine will always be Snowflake.**
```
sfc-gh-yzou, line 122: do we raise a warning for this if customer uses a method that is not table? and do we mention this in the generated doc?

src/snowflake/snowpark/modin/plugin/_internal/window_utils.py
```diff
@@ -101,3 +100,37 @@ def check_is_rolling_window_supported_by_snowflake(
...
+
+    Returns
+    -------
+    bool
+        Whether operations can be executed with snowflake sql engine.
```
sfc-gh-yzou, line 127: this comments seems wrong, this function seems also raises error, and the return value is None here, please update the comment correctly, and update the fucntion to check_and_raise _error if the intention is to also raise an error

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10797,29 +10811,43 @@ def _rolling_agg(
...
+        # Handle case where min_periods = None
+        min_periods = 0 if min_periods is None else min_periods
         # Perform Aggregation over the window_expr
         new_frame = frame.update_snowflake_quoted_identifiers_with_expressions(
             {
                 quoted_identifier: iff(
-                    count(col(quoted_identifier)).over(window_expr) >= min_periods,
+                    count(col(row_position_quoted_identifier)).over(window_expr)
+                    >= min_periods
+                    if agg_func == "count"
```
sfc-gh-yzou, line 10845: can you  add some explanation here about why we need those special case here?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10797,29 +10811,43 @@ def _rolling_agg(
...
         frame = query_compiler._modin_frame.ensure_row_position_column()
         row_position_quoted_identifier = frame.row_position_snowflake_quoted_identifier
         if center:
             # -(window // 2) is equivalent to window // 2 PRECEDING
             rows_between_start = -(window // 2)  # type: ignore
             rows_between_end = (window - 1) // 2  # type: ignore
         else:
-            # 1 - window is equivalent to window - 1 PRECEDING
-            rows_between_start = 1 - window  # type: ignore
+            if window_func == "rolling":
```
sfc-gh-yzou, line 10826: instead of using string, can you simply define and enum to make all usage defnition centralized

docs/source/modin/supported/dataframe_supported.rst
```diff
@@ -169,7 +169,7 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``ewm``                     | N                               |                                  |                                                    |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``expanding``               | N                               |                                  |                                                    |
+| ``expanding``               | P                               |                                  | ``N`` if ``axis = 1``                              |
```
sfc-gh-yzou, line 172: you also need to add those to the window.rst to make sure doc gen will update those functions.

src/snowflake/snowpark/modin/plugin/_internal/window_utils.py
```diff
@@ -9,7 +9,8 @@
 
 from snowflake.snowpark.modin.plugin.utils.error_message import ErrorMessage
 
-IMPLEMENTED_ROLLING_AGG_FUNCS = ["sum", "mean", "var", "std", "min", "max"]
+IMPLEMENTED_ROLLING_AGG_FUNCS = ["count", "sum", "mean", "var", "std", "min", "max"]
```
sfc-gh-yzou, line 12: This two seems exactly the same list, can we just use one here?

sfc-gh-yzou: also those parameter seems only used by test, if that is the case, could you please move it to test folder

src/snowflake/snowpark/modin/pandas/window.py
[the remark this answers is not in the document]
sfc-gh-nkrishna: Rolling/expanding are very similar (expanding is cumulative and rolling is not) so I don't think it needs two separate PRs as I'm also implementing rolling.count in this PR (it's very similar implementation to expanding.count).

The error message at the frontend layer matches the errors that pandas/Modin throw as well when rolling/expanding are called with invalid parameters, so I think having it at the frontend layer is a good thing/beneficial.

Rolling is 0 is a special case, I haven't specifically looked into how to implement it but it does not work with the current implementation. I'm sure the current implementation is extensible but out of scope for this PR.

src/snowflake/snowpark/modin/plugin/_internal/window_utils.py
sfc-gh-nkrishna: No we shouldn't use one list b/c there are a bunch of window functions, EWM/Rolling/Expanding/Weighted Window -- keeping them separate b/c we may invest more in implementing some over others. For now, just getting rolling/expanding to parity but going to be focusing more on expanding in future since that has more customer/SE requests.

sfc-gh-nkrishna: Yeah moving it to the test file makes sense, will do

src/snowflake/snowpark/modin/plugin/_internal/window_utils.py
sfc-gh-nkrishna: Ah good point, can add the warning being raised at the frontend layer since it applies for all APIs -- I don't think we need to mention warnings like this in the generated doc though

docs/source/modin/supported/dataframe_supported.rst
sfc-gh-nkrishna: Good catch, will do

src/snowflake/snowpark/modin/plugin/_internal/window_utils.py
sfc-gh-nkrishna: Good catch, will update.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Added.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Yep done.

src/snowflake/snowpark/modin/pandas/window.py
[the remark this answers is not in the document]
sfc-gh-yzou: The are similar, but not the same, and doesn't seem matching the purpose of your current pr, your current pr intended to add support for new function, and make the review little bit difficult. for example, if you put the change here, i would expect you also add test to check this new case also, but seems not.

Also as where the check should happen, snowpark pandas specific check should happen at backend layer, or the overwritten layer, this folder is target for removed. From what you have described, when min_period = 0, it works for modin, it just doesn't work for Snowpark pandas, and we can potentially extend our backend to handle it. Then at backend should be fine, and i belive you already did the check at backend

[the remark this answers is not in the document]
sfc-gh-nkrishna: This isn't a Snowpark pandas specific check (that happens in the backend), this is a pandas check (all pandas cases where window < 0 results in this error).

The error message is `ValueError("window must be an integer 0 or greater")` the `=` part of the comparison was just a typo being fixed here.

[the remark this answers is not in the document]
sfc-gh-nkrishna: I added the window=0 negative test (checking it throws a NotImplementedError) to `test_rolling_params_unsupported`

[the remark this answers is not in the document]
sfc-gh-yzou: i see, i might read the message in an opposite way, then I think we should open an modin issue to upstreaming the frontend check changes.

For the test, there seems a kind of behavior change here, before this case raises a ValueError, now it start throwing an not implemented error.

Also, please add this change to CHAGNELOG for the window = 0 case

src/snowflake/snowpark/modin/plugin/_internal/window_utils.py
```diff
@@ -5,18 +5,35 @@
...
+    Attributes:
+        EXPANDING (str): Represents the expanding window.
+        ROLLING (str): Represents the rolling window.
+        WEIGHTED (str): Represents the weighted window.
+        EXPONENTIAL (str): Represents the exponential weighted window.
+    """
+
+    EXPANDING = "expanding"
+    ROLLING = "rolling"
+    WEIGHTED = "weighted"
```
sfc-gh-yzou, line 27: i do not see the WEIGHTED and  EXPONENTIAL is used today. is it for some other window that we might support later? if yes, let's add them later to be clear

sfc-gh-nkrishna: Yes they are for later, I can remove them

src/snowflake/snowpark/modin/pandas/window.py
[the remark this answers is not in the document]
sfc-gh-nkrishna: Added it as a bug fix: `Fixed a bug in `DataFrame.rolling` and `Series.rolling` so `window=0` now throws `NotImplementedError` instead of `ValueError``

docs/source/modin/supported/dataframe_supported.rst
sfc-gh-helmeleegy, line 172: Can we match this documentation with that of rolling?

sfc-gh-nkrishna: Good point, done

sfc-gh-helmeleegy (approved):
LGTM. This is a great push to our API coverage efforts. Thank you!

PR #1763, opened 2024-06-12
SNOW-1476363: Add `groupby().size()` pandas API

sfc-gh-dpetersohn:

   

   Fixes SNOW-1476363

   Add the pandas API `groupby().size()` to Snowpark pandas.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -3640,6 +3640,78 @@ def groupby_shift(
...
+            SnowflakeQueryCompiler: The result of groupby_size()
+        """
+        if not is_list_like(by):
+            by = [by]
+        # We reset index twice to ensure we perform the count aggregation on the row
+        # positions (which cannot be null).
+        result = (
+            self.reset_index(drop=True)
+            .reset_index(drop=False, names=MODIN_UNNAMED_SERIES_LABEL)
+            .take_2d_labels(slice(None), [MODIN_UNNAMED_SERIES_LABEL] + by)
```
sfc-gh-mvashishtha, line 3701: I think `[MODIN_UNNAMED_SERIES_LABEL] + by` becomes ambiguous when `by` includes `MODIN_UNNAMED_SERIES_LABEL`, for example if we are grouping an unnamed series by itself, like `s.groupby(s).size()`. We should be raising `NotImplementedError: Snowpark pandas GroupBy.sum does not yet support pd.Grouper, axis == 1, by != None and level != None, by containing any non-pandas hashable labels, or unsupported aggregation parameters` when grouping by a dataframe or series, but `groupby.size` on a series is giving me `ValueError: cannot insert __reduced__, already exists`:

```python
df = pd.Series([1, 2, 3])
df.groupby(df).size()
```

tests/integ/modin/groupby/test_groupby_size.py
```diff
@@ -0,0 +1,77 @@
...
+                    np.arange(2, dtype="int32") // 3,
+                    [np.nan],
+                    np.arange(2, dtype="int16") // 3,
+                    [np.nan],
+                ]
+            ),
+        }
+    )
+    pandas_df = snowpark_pandas_df.to_pandas()
+    with SqlCounter(query_count=1 if as_index else 2):
```
sfc-gh-mvashishtha, line 72: What causes the extra query when `as_index=False`?

docs/source/modin/supported/groupby_supported.rst
```diff
@@ -154,7 +154,7 @@ Computations/descriptive stats
 | ``shift``                   | P                               | ``Y`` if ``axis = 0``, ``freq`` is None,           |
 |                             |                                 | ``level`` is None, and ``by`` is in the columns    |
 +-----------------------------+---------------------------------+----------------------------------------------------+
-| ``size``                    | N                               |                                                    |
+| ``size``                    | Y                               |                                                    |
```
sfc-gh-azhan, line 157: please update docs/source/modin/groupby.rst too

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 - Added support for `Series.dt.dayofweek`, `Series.dt.day_of_week`, `Series.dt.dayofyear`, and `Series.dt.day_of_year`.
 - Added support for `Series.str.__getitem__` (`Series.str[...]`).
 - Added support for `Series.str.lstrip` and `Series.str.rstrip`.
+- Added support for `DataFrameGroupby.size`.
```
sfc-gh-azhan, line 36: should we support SeriesGroupby.size too?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-dpetersohn: Good point, I added the error checking to make sure we're throwing the right error.

tests/integ/modin/groupby/test_groupby_size.py
sfc-gh-dpetersohn: Looks like there's an extra `count(1)` query being issued.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-dpetersohn: I also changed the name to something unique so there's no conflict

CHANGELOG.md
sfc-gh-dpetersohn: I wasn't planning to, but it wasn't too much more effort 😄

tests/integ/modin/groupby/test_groupby_size.py
sfc-gh-nkrishna: I think we have this same behavior (extra count query) for reset_index, so I believe it is to be expected

tests/integ/modin/groupby/test_groupby_size.py
```diff
@@ -0,0 +1,91 @@
...
+                    [np.nan],
+                    np.arange(2, dtype="int32") // 3,
+                    [np.nan],
+                    np.arange(2, dtype="int16") // 3,
+                    [np.nan],
+                ]
+            ),
+        }
+    )
+    pandas_df = snowpark_pandas_df.to_pandas()
```
sfc-gh-nkrishna, line 70: Surprised this doesn't issue an extra query

sfc-gh-nkrishna (approved):
LGTM!

src/snowflake/snowpark/modin/pandas/groupby.py
```diff
@@ -736,8 +736,30 @@ def all(self, skipna=True):
...
 
     def size(self):
-        # TODO: SNOW-1063349: Modin upgrade - modin.pandas.groupby.DataFrameGroupBy functions
-        ErrorMessage.method_not_implemented_error(name="size", class_="GroupBy")
+        if self._axis == 1:
+            return DataFrameGroupBy(
+                self._df.T.iloc[:, [0]],
+                self._by,
+                0,
+                drop=self._drop,
```
sfc-gh-yzou, line 744: In our previous discussion when talking about upstreaming, i think we agreed that the frontend will do basic parameter check,  conversion, and result correction. 
if that is the case, then it is probably better to move this just to query backend, you can call transform on it. If we are worry about axis = 1, I think that is a parameter scheduled to be deprecated anyway, we can improve our error message to ask customer to do the transformation before they call size also.

src/snowflake/snowpark/modin/pandas/groupby.py
```diff
@@ -736,8 +736,30 @@ def all(self, skipna=True):
...
+                idx_name=self._idx_name,
+                **self._kwargs,
+            ).size()
+        result = self._wrap_aggregation(
+            type(self._query_compiler).groupby_size,
+            numeric_only=False,
+        )
+        if not isinstance(result, Series):
+            result = result.squeeze(axis=1)
+        if not self._kwargs.get("as_index") and not isinstance(result, Series):
```
sfc-gh-yzou, line 754: so are we checking for the case where result is a dataframe and as index is true? maybe use positive check to make things more clear?

tests/integ/modin/groupby/test_groupby_size.py
sfc-gh-dpetersohn: It's coming from the `squeeze(axis=1)` call here: https://github.com/snowflakedb/snowpark-python/pull/1763/files#diff-1d632233293b5d0c0f50ad639e2415f3d32851fd41d59f2295c38e29ff0e3707R753, but it shouldn't, `axis=1` does not need the `len` of the dataframe). We should fix that bug in another PR.

src/snowflake/snowpark/modin/pandas/groupby.py
```diff
@@ -1190,6 +1212,9 @@ def is_monotonic_increasing(self):
             name="is_monotonic_increasing", class_="GroupBy"
         )
 
+    def size(self):
+        return super().size().rename(self._df.columns[-1])
```
sfc-gh-yzou, line 1216: I thought we already handled the name in this part 
```
elif isinstance(self._df, Series):
            result.name = self._df.name
```

tests/integ/modin/groupby/test_groupby_size.py
sfc-gh-mvashishtha: I think this is `columnarize` in the `Series` constructor again. We need to kill it: https://github.com/modin-project/modin/issues/6111

sfc-gh-mvashishtha (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -3640,6 +3640,90 @@ def groupby_shift(
...
+            .groupby_agg(
+                by,
+                "count",
+                axis,
+                groupby_kwargs,
+                (),
+                {},
+            )
+        )
+        if not groupby_kwargs.get("as_index", True):
```
sfc-gh-yzou, line 3720: if you simply pass the correct name to reset_index, will we need to do rename here? or there are some other situtation we are worried about?

tests/integ/modin/groupby/test_groupby_size.py
sfc-gh-dpetersohn: https://github.com/snowflakedb/snowpark-python/pull/1767

sfc-gh-dpetersohn: That fixes the issue, it's not columnarize in this case 😄

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: This code is pulled directly from Modin OSS. The primary goal is to make as few changes as possible that need to be upstreamed. I don't think we should worry about `axis=1` use case.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: Something wacky is happening in the frontend and the `isinstance` check always evaluates to `False`. It is likely that Snowpark pandas is doing a `reset_index` before the `groupby` so this is needed currently. It can be removed once we fix `Series.groupby` to match Modin.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: All the code here was copied from Modin OSS. Syntax changes like this I would prefer to hold off on until they are changed in Modin first.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-dpetersohn: We need the new column name to be unique (there may be a `"size"` or a `MODIN_UNNAMED_SERIES_LABEL` named column already in the data). I'll add a comment there.

tests/integ/modin/groupby/test_groupby_size.py
sfc-gh-dpetersohn: I put the counter checks below this call. I didn't use the decorator in this case.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-yzou: i think our primary goal is to make sure both Snowpark pandas and modin aligned on the frontend based on our previous architecture discussion, and the second goal is to minimize the work.

If we are not worried about axis = 1, how about remove this axis = 1 handling at frontend, and the backend check should throw an error for axis = 1 now.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-yzou: is there a ticket for this? if we do not have ticket to track this, it is just going to be forget.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-yzou: this seems checking the frontend self._df, Do you mean we did a reset_index at frontend? that is weird, do you know where we did this?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: if that is the case, it is also possible that we will have conflict with positions_col_name, right? then may be generate the column name by deconflict among the existing pandas labels with a random suffix, we can add a simple utility function under utils.py, we have one for quoted_identifiier generation generate_snowflake_quoted_identifiers_helper, but this one doesn't require any quotes there, might be easy to just get another one to operate at pandas label level

sfc-gh-yzou: or another simple way is that you can simply unquote the identifier of the index column, which is guaranteed to be unique among the column, but may not be that obvious on purpose when read the code

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: SNOW-1063349 is how we are tracking the groupby methods aligning with Modin. This will get handled when we align with Modin.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: I couldn't find where in a cursory look, but I didn't want to block the customer on debugging this. It will get resolved once we align with Modin OSS.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-dpetersohn: >if that is the case, it is also possible that we will have conflict with positions_col_name, right?

Not really, unless the customer has a column named "__TEMP_POS_NAME__". I think this is fine as is, and if we want to refine the unique logic we should do it in a subsequent PR to unblock.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: Removed axis=1, but I also put back our TODO since it's not aligned anymore.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: Would you be able to open the ticket on Modin with what you propose? I'm not sure I fully understand your question.

sfc-gh-yzou: Oh no, all i am proposing is just do check like isinstance(result, DataFrame),instead of not isinstance(result, Series), this just make it clear that I am handling dataframe case, or are we also intend to handle other case?

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-yzou: Our current release is already out, the next release will come in next 1 or 2 weeks, so i don't see a reason to hurry for this. I don't want to relay on align with Modin OSS for this, it seems still far away, if there is something wrong we did, we can fix it right away, so it doesn't cause us other problem when we implement other methods

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: I would at least append a random suffix to be safe, you do not know what user will do, sometimes it is not the name they add, it could be caused due to our backend processing where we are quoting and unquoting the identifiers. And can you create a ticket to follow this up?

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: This is not the only request from the customer, and I would like for our team to be in the habit of shipping quickly. This works, debugging this will not add any value to the user, and fixing this is planned for the future. I prefer not to block the PR on something like this, since it may force a big change that affects every other groupby method. IMO it's better to do that when we are migrating code.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: This is where to file the issue in Modin: https://github.com/modin-project/modin/issues/new. Since it's your suggestion, I think it's best if you file it.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-dpetersohn: SNOW-1478730

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-yzou: We can ship quickly. However, ship quickly doesn't really mean ship hacky code without knowing the root cause, this will make our debugging difficult if we run into issue, and made our code base worse and worse. As far as we know the root cause, we can have follow up tickets to do it right after, they migration doesn't seem going to happen anytime soon also.

sfc-gh-dpetersohn: In the interest of unblocking this, I have root caused the issue and filed a ticket: https://snowflakecomputing.atlassian.net/browse/SNOW-1478924

The migration is in progress, so let's not belittle the progress being made there.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-dpetersohn: I also added a random 6 digit hex code at the end of the temporary name. I'll mark this as resolved and please unresolve it if there's something else that needs to be done.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-yzou: I see. so that is basically a problem if we are calling groupby getitem with size, groupby(['A', 'B']).['A'].size(), however, for regular groupby('A').size() or series.groupby().size() should still working fine. Could you please leave a comment here in the code with the ticket number about why are we doing the correction here?

This could potentially impact other APIs also, we probably should mark it as a bug. Furthermore, this may not be something we want to follow modin's implementation, there seems a reason for us to choose iloc in particular.

sfc-gh-yzou (approved):

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-dpetersohn: Sounds great, will add a TODO there.

PR #1764, opened 2024-06-12
SNOW-1476313: Wait for result for AsyncJob.result(no_result)

sfc-gh-yixie:

   

   Fixes SNOW-1476313

   Wait for result even if result isn't needed.


src/snowflake/snowpark/async_job.py
```diff
@@ -355,6 +357,25 @@ def result(
                 )
 
         if async_result_type == _AsyncResultType.NO_RESULT:
+            # The following section is copied from python connector.
+            # Later we should expose it from python connector and reuse it.
+            retry_pattern_pos = 0
+            while True:
```
sfc-gh-lspiegelberg, line 363: would maybe use a timeout / retry count instead of endless loop. This looks like an antipattern to me.

sfc-gh-yixie: This is to polling a query result so we may not want to timeout from the client. It's not like a connection.

sfc-gh-yuwang (approved):
LGTM

sfc-gh-jrose (approved):

PR #1773, opened 2024-06-13
Fill colum expr src positions, raise explicit errors on missing AST

sfc-gh-heshah:
- Adding src position information to Column AST building APIs
- Adding a flag in `ast_utils.py` to signal whether to fail loudly on missing AST fields
  - Generate useful error message in Column APIs if any Snowpark API generates a column and does not log its AST

### DIFF
--- a/src/snowflake/snowpark/_internal/ast_utils.py
+++ b/src/snowflake/snowpark/_internal/ast_utils.py
@@ -13,6 +13,7 @@ from typing import Any, List, Optional
 
 import snowflake.snowpark._internal.proto.ast_pb2 as proto
 
+FAIL_ON_MISSING_AST = True
 
 def build_const_from_python_val(obj: Any, ast: proto.Expr) -> None:
     """Infer the Const AST expression from obj, and populate the provided ast.Expr() instance
--- a/src/snowflake/snowpark/_internal/ast_utils.py
+++ b/src/snowflake/snowpark/_internal/ast_utils.py
@@ -9,13 +9,14 @@ import re
 import sys
 from functools import reduce
 from pathlib import Path
-from typing import Any, List, Optional
+from typing import Any, Optional
 
 import snowflake.snowpark._internal.proto.ast_pb2 as proto
 
 # This flag causes an explicit error to be raised if any Snowpark object instance is missing an AST or field, when this
 # AST or field is required to populate the AST field of a different Snowpark object instance.
 FAIL_ON_MISSING_AST = True
+
 
 def build_const_from_python_val(obj: Any, ast: proto.Expr) -> None:
     """Infer the Const AST expression from obj, and populate the provided ast.Expr() instance
--- a/src/snowflake/snowpark/_internal/ast_utils.py
+++ b/src/snowflake/snowpark/_internal/ast_utils.py
@@ -13,6 +13,8 @@ from typing import Any, List, Optional
 
 import snowflake.snowpark._internal.proto.ast_pb2 as proto
 
+# This flag causes an explicit error to be raised if any Snowpark object instance is missing an AST or field, when this
+# AST or field is required to populate the AST field of a different Snowpark object instance.
 FAIL_ON_MISSING_AST = True
 
 def build_const_from_python_val(obj: Any, ast: proto.Expr) -> None:
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -993,6 +993,7 @@ class Column:
         ast = proto.Expr()
         if property is not None:
             prop_ast = getattr(ast, property)
+            fill_src_position(prop_ast.src)
             for attr, value in assign_fields.items():
                 setattr_if_not_none(prop_ast, attr, value)
             for attr, value in assign_opt_fields.items():
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -3,7 +3,9 @@
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
 
+import inspect
 import sys
+from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
 import snowflake.snowpark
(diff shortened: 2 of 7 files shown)


src/snowflake/snowpark/column.py
```diff
@@ -990,12 +994,14 @@ def _create_ast(
...
         if property is not None:
             prop_ast = getattr(ast, property)
+            fill_src_position(prop_ast.src)
             for attr, value in assign_fields.items():
                 setattr_if_not_none(prop_ast, attr, value)
             for attr, value in assign_opt_fields.items():
                 setattr_if_not_none(getattr(prop_ast, attr), "value", value)
             for attr, messg in copy_messages.items():
-                getattr(prop_ast, attr).CopyFrom(messg)
+                if messg is not None:
```
sfc-gh-lspiegelberg, line 1003: This error here actually motivated me to look into IR for functions right now :) Not sure we should merge it, I think I'd prefer to throw a NotImplementedError here with a good message which API is responsible for this. Do you think we can make an edit for this?

src/snowflake/snowpark/column.py
```diff
@@ -990,12 +994,14 @@ def _create_ast(
         ast = proto.Expr()
         if property is not None:
             prop_ast = getattr(ast, property)
+            fill_src_position(prop_ast.src)
             for attr, value in assign_fields.items():
                 setattr_if_not_none(prop_ast, attr, value)
             for attr, value in assign_opt_fields.items():
                 setattr_if_not_none(getattr(prop_ast, attr), "value", value)
             for attr, messg in copy_messages.items():
```
sfc-gh-azwiegincew, line 1002: The standard abbreviation for "message" is "msg". "Messg" is so long, we might as well fully spell it out. I'd prefer `msg`.

src/snowflake/snowpark/column.py
```diff
@@ -990,12 +994,14 @@ def _create_ast(
...
-                getattr(prop_ast, attr).CopyFrom(messg)
+                if messg is not None:
+                    getattr(prop_ast, attr).CopyFrom(messg)
```
sfc-gh-azwiegincew, line 1004: `CopyFrom` is a code smell in protobuf. I suspect that the idea of this helper is flawed – why not just create protobuf directly instead of calling this fn?

sfc-gh-azwiegincew (changes requested):

src/snowflake/snowpark/column.py
sfc-gh-heshah: This line is mainly utilized to copy the AST stored in existing column instances to the `col` field of the newly created operator AST.

Since the Column API allows arbitrary columns to be constructed independently of DataFrames or Sessions, we aren't guaranteed to have a top level AST (`Assign` or `Eval`) that we can always continue building down the chain of Column API calls, hence the current need to use `CopyFrom`.

This might be worth discussing and I have some preliminary discussion points in the original Column Expressions design doc I can share.

sfc-gh-azwiegincew: Oh, I see. That's reasonable.

However, I still suspect that having this helper is counterproductive. Maybe let's discuss tomorrow in person?

src/snowflake/snowpark/column.py
sfc-gh-heshah: I tried adding some logic which should be able to raise an error which is granular to the line level where the usage of a Snowpark function API that does not perform AST logging is detected. Would this be sufficient?

It doesn't seem straightforward to get the exact API that doesn't log its AST.

sfc-gh-heshah: I'll leave an example of the error message below as well with the code snippet
```python
from snowflake.snowpark.functions import mean, stddev, col

df = session.table("test_table")

result = df.select((mean(df["STR"]) - 2 * stddev(df["STR"]) < col('STR')) & (col('STR') < mean(df["STR"]) + 2 * stddev(df["STR"])))
print(result.queries)
```

```
NotImplementedError: Calling Column API "__rmul__" which supports AST logging, from File "/Users/heshah/hemit/scratch.py", line 7, potentially in <module>
        result = df.select((mean(df["STR"]) - 2 * stddev(df["STR"]) < col('STR')) & (col('STR') < mean(df["STR"]) + 2 * stddev(df["STR"])))
        An API from snowflake.snowpark.functions used above has not yet implemented AST logging.
```

src/snowflake/snowpark/column.py
sfc-gh-heshah: Follow up from in-person discussion with Arthur: This helper function complicates the logging of the AST at the cost of saving a few lines of code within each Column API. To make the logging logic easier to follow in each Column API it would be better to directly create and fill the 1:1 AST entity in each function itself at the cost of verbosity. (This kind of helper method would be better suited to the alternative of exposing a Builder interface to the client code to create/fill the AST as discussed previously.)

src/snowflake/snowpark/dataframe.py
```diff
@@ -1095,7 +1096,11 @@ def columns(self) -> List[str]:
     def col(self, col_name: str) -> Column:
         """Returns a reference to a column in the DataFrame."""
         col_expr_ast = proto.Expr()
-        col_expr_ast.sp_dataframe_col.df.sp_dataframe_ref.id.bitfield1 = self._ast_id
+        if self._ast_id is None and FAIL_ON_MISSING_AST:
+            print(self._explain_string())
```
sfc-gh-lspiegelberg, line 1100: let's use logging instead of print

tests/ast/update-unparser.sh
```diff
@@ -31,10 +31,18 @@ JAR_HOME="$REMOTE_HOME/$UNPARSER_DIR/target/scala-2.13/unparser-assembly-0.1.jar
 LOCAL_TARGET_FILENAME="unparser-assembly-0.1.jar"
 
 
-# Step 1: Build the Unparser JAR using sbt if it does not exist
-ssh $HOST "bash -c 'cd Snowflake/trunk;bazel build //Snowpark/unparser:unparser'"
-JAR_HOME=$(ssh $HOST "bash -c 'cd Snowflake/trunk;bazel cquery --output=files //Snowpark/unparser:unparser.jar'")
-JAR_HOME=$REMOTE_HOME/Snowflake/trunk/$JAR_HOME
+# Step 1: Build the Unparser JAR using bazel or sbt if it does not exist
```
sfc-gh-lspiegelberg, line 34: I fixed this with a PR. Need to invoke for now both bazel + sbt :/

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -13,6 +13,7 @@
 
 import snowflake.snowpark._internal.proto.ast_pb2 as proto
 
+FAIL_ON_MISSING_AST = True
```
sfc-gh-lspiegelberg, line 16: add docstring to document what this does/which effect it has.

src/snowflake/snowpark/column.py
```diff
@@ -993,12 +997,24 @@ def _create_ast(
         ast = proto.Expr()
         if property is not None:
```
sfc-gh-azwiegincew, line 998: So where do we stand? Are you going to remove _create_ast()?

sfc-gh-heshah: Yes, we will remove `_create_ast()`, but we'll push this refactor to a separate PR in favor of narrowing this one's scope to just more informative error messages.

sfc-gh-lspiegelberg (approved):
lgtm, let's update the description of the PR when merging to reflect that this PR only does:
1.) raise error
2.) set src position for column expr

PR #1776, opened 2024-06-13
SNOW-1480718: Support Series.str.translate

sfc-gh-joshi:

   

   Fixes SNOW-1480718

This PR implements [Series.str.translate](https://pandas.pydata.org/docs/reference/api/pandas.Series.str.translate.html#pandas.Series.str.translate) using SQL TRANSLATE.

Unlike the native pandas version, we cannot perform one-to-many character mappings with SQL TRANSLATE. I considered using chained `REPLACE` calls to mimic this behavior, but that would cause discrepancies in situations where one REPLACE mapping generates entries that match the replacement keys of a subsequent REPLACE calls. As such, for these use cases it is recommended for users to manually chain `Series.str.replace` calls as necessary.

Slack thread discussing future work on a more robust multi-character approach using REPLACE with sentinel values: https://snowflake.slack.com/archives/C04HF38JFAQ/p1718388095381469


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13295,8 +13296,69 @@ def str_rstrip(self, to_strip: Union[str, None] = None) -> "SnowflakeQueryCompil
...
+            if len(key) != 1:
+                # Mimic error from str.maketrans
+                raise ValueError(
+                    f"Invalid mapping key '{key}'. String keys in translate table must be of length 1."
+                )
+            if value is not None and len(value) > 1:
+                raise ValueError(
+                    f"Invalid mapping value '{value}' for key '{key}'. Snowpark pandas currently only "
+                    "supports unicode ordinals or 1-codepoint strings as values in str.translate mappings. "
+                    "Consider using Series.str.replace to replace multiple characters."
```
sfc-gh-helmeleegy, line 13343: Why not automatically use Snowflake's `translate` for all single character replacement, and then wrap that with a `replace` for each one-charachter-to-many-characters mappings as specified in `table`? This will result in a chain of 0 or more nested `replace` call with (or without) an inner most call to `translate`.

This way the user gets the exact pandas semantic without having to make any code changes or learn any new concepts.

sfc-gh-joshi: I considered this option, but this causes problems if the calls to SQL REPLACE end up interfering with each other. 
Take this example:

```python
native_pd.Series("ab").str.translate(str.maketrans({"a": "AA", "A": "bad", "b": "a"}))
```
The expected output "AAa". The chained TRANSLATE/REPLACE approach would generate a SQL query similar to this:

```sql
WITH t1 AS (SELECT 'ab' AS col0)
SELECT
    REPLACE(
        REPLACE(
            TRANSLATE(col0, 'b', 'a'),
            'a',
            'AA'
        ),
        'A',
        'bad'
    )
FROM T1
;
```
This outputs `badbadbadbad` because of the sequential nature of the string operations. We could try to introduce some cycle detection on the input table and generate this nested REPLACE/TRANSLATE sequence if no second-order replacements would occur, but that felt like too much client-side work to me. If you think doing such a check is worth it to support the general case, then I can try and add it in this PR.

sfc-gh-helmeleegy: Yes, I was thinking about this case too. `Series.str.translate` happens in a single pass, so no concerns about interference. I believe Snowflake's `translate` will also happen in a single pass. The issue comes from the multiple passes introduced by the multiple calls to `replace`. (Note that this will still be an issue for users if they attempted to make multiple calls to `replace` on their own.) 

Also, note that this is an issue that we can detect by analyzing the input `table`. In particular, after applying the initial translate operation, then the output string will be made up of the 

if the keys in `table` mapping to multi-character values overlap with any of the values in `table`. The reason is that those keys get replaced at a later stage after some other replacement/translation operations have already taken place. So if they don't overlap with the outputs of those operations, then we don't run the risk of replacing the same character twice or more, which is unlike pandas behavior.

My suggestion is to treat this case as a corner case. Whenever we detect it, we just throw a NotImplementedError. If needed we can come back to revisit it and implement a good algorithm for it. The only thing we need to be careful about for now is to make sure that our tests have good coverage for all the variants of this case, where we should raise a NotImplemetedError.

sfc-gh-helmeleegy: In your example, the input `table` can be detected as unsupported since both `a` and `A` are among the keys mapped to multi-character values, and also overlap with some values.

sfc-gh-helmeleegy: I also agree that if we don't have a cycle, then even in the presence of overlaps, we might still be able to come up with an optimal sequence for the nested calls that can guarantee a single replacement per input character. I just think that we can wait on this for now.

sfc-gh-joshi: Sure, that makes sense. I was thinking about sticking to our "thin client" philosophy of minimizing this sort of pre-processing, but it should be easy enough to come up with a topological sort on `table` to at least detect the infeasible case double replacement for now.

Do you have any suggestions on how to word the error message for this? Describing a "cycle in input `table`" might be inscrutable to some users, but I'm not sure how else to succintly communicate why a replacement would fail.

tests/integ/modin/strings/test_strings.py
```diff
@@ -578,3 +579,113 @@ def test_get_with_dict_label(key, expected_result):
...
+            # Inputs:
+            # - "head shaking horizontally" = 1F642 + 200D + 2194 + FE0F
+            # - "heart on fire" = 2764 + FE0F + 200D + 1F525
+            # - "judge" = 1F9D1 + 200D + 2696 + FE0F
+            # Outputs:
+            # - "head shaking vertically" = 1F642 + 200D + 2195 + FE0F
+            # - "mending heart" = 2764 + FE0F + 200D + 1FA79
+            # - "health worker" = 1F91D1 + 200D + 2605 + FE0F
+            ["🙂‍↔️", "❤️‍🔥", "🧑‍⚖️"],
+            {
```
sfc-gh-mvashishtha, line 636: it looks like this doesn't change anything in pandas:

```python
import pandas as pd

s = pd.Series(["🙂‍↔️", "❤️‍🔥", "🧑‍⚖️"])
assert s.replace({0x2194: 0x2195,
                0x1F525: 0x1FA79,
                0x2696: 0x2605,
            }).equals(s)
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: Doing a topological sort anywhere seems like overkill when we can raise an error saying we don't support this feature yet. I suggest we wait for customer feedback to try implementing multi-character replace.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13295,8 +13296,69 @@ def str_rstrip(self, to_strip: Union[str, None] = None) -> "SnowflakeQueryCompil
...
+        # Snowflake SQL TRANSLATE:
+        #   TRANSLATE(<subject>, <sourceAlphabet>, <targetAlphabet>)
+        # Characters in the <sourceAlphabet> string are mapped to the corresponding entry in <targetAlphabet>.
+        # If <sourceAlphabet> is longer than <targetAlphabet>, then the trailing characters of <sourceAlphabet>
+        # are removed from the input string.
+        #
+        # Because TRANSLATE only supports 1-to-1 character mappings, any entries with multi-character
+        # values must be handled by REPLACE instead. 1-character keys are always invalid.
+        single_char_pairs = {}
+        multi_char_pairs = {}
```
sfc-gh-mvashishtha, line 13326:
```suggestion
```

I think this is unused

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13295,8 +13296,69 @@ def str_rstrip(self, to_strip: Union[str, None] = None) -> "SnowflakeQueryCompil
...
+            if value is not None and len(value) > 1:
+                raise ValueError(
+                    f"Invalid mapping value '{value}' for key '{key}'. Snowpark pandas currently only "
+                    "supports unicode ordinals or 1-codepoint strings as values in str.translate mappings. "
+                    "Consider using Series.str.replace to replace multiple characters."
+                )
+            if value is None or len(value) == 0:
+                none_keys.add(key)
+            elif len(value) > 1:
+                multi_char_pairs[key] = value
```
sfc-gh-mvashishtha, line 13348: I think this is unused. Not sure whether we ever hit this `elif`.

sfc-gh-mvashishtha (commented):
some minor comments

tests/integ/modin/strings/test_strings.py
sfc-gh-joshi: I don't think `replace` takes dicts as input. This works for me:
```
>>> s.str.translate({
...                 0x2194: 0x2195,
...                 0x1F525: 0x1FA79,
...                 0x2696: 0x2605,
...             })
0    🙂‍↕️
1    ❤️‍🩹
2    🧑‍★️
```
It does turn out that I specified one of the hex codes wrong (should be 0x2695 for "Staff of Aesculapius" instead of 0x2605 for "Black Star").

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: As discussed on slack, we've decided to just go with the 1-char direct mapping approach for now. That said, the motivating example for this function wanted to translate an encoding mapping individual characters to multi-digit numerals, so it would be nice to support the multiple-character case ASAP.

sfc-gh-helmeleegy: Yes, it will be nice if we can support multi-character mappings ASAP. The new [proposal](https://snowflake.slack.com/archives/C04HF38JFAQ/p1718388095381469) does not require a topological sort.

sfc-gh-helmeleegy (approved):
LGTM. Thanks, Jonathan! Can you please update the PR description to better describe the plan for supporting multi-characters in the future? You can give a short description and point to the slack thread for more details.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13935,8 +13936,66 @@ def str_rstrip(self, to_strip: Union[str, None] = None) -> "SnowflakeQueryCompil
...
+                key = chr(key)
+            if isinstance(value, int):
+                value = chr(value)
+            if len(key) != 1:
+                # Mimic error from str.maketrans
+                raise ValueError(
+                    f"Invalid mapping key '{key}'. String keys in translate table must be of length 1."
+                )
+            if value is not None and len(value) > 1:
+                raise ValueError(
```
sfc-gh-mvashishtha, line 13979: Should this be `NotImplementedError`?

sfc-gh-joshi: Good call, fixed.

sfc-gh-mvashishtha (approved):
nice work!

PR #1782, opened 2024-06-14
SNOW-1445839: Add support for nlargest and nsmallest

sfc-gh-nkumar:
SNOW-1445839: Add support for DataFrame.nlargest
SNOW-1445840: Add support for DataFrame.nsmallest
SNOW-1445733: Add support for Series.nlargest
SNOW-1445734: Add support for Series.nsmallest
   

- Add support for nlargest and nsmallest methods on DataFrame and Series. Raise NotimplementedError when keep='all'
- Copy updated frontend code from modin repo.


src/snowflake/snowpark/modin/pandas/series.py
```diff
@@ -1490,22 +1490,32 @@ def ne(self, other, level=None, fill_value=None, axis=0):  # noqa: PR01, RT01, D
...
         # TODO: SNOW-1063347: Modin upgrade - modin.pandas.Series functions
-        return self._default_to_pandas(pandas.Series.nlargest, n=n, keep=keep)
+        if len(self._query_compiler.columns) == 0:
+            # pandas returns empty series when requested largest/smallest from empty series
+            return self.__constructor__(data=[], dtype=float)
+        return Series(
+            query_compiler=self._query_compiler.nlargest(
+                n=n, columns=self.name, keep=keep
+            )
+        )
```
sfc-gh-nkumar, line 1505: This is copied from modin repo.

src/snowflake/snowpark/modin/pandas/series.py
```diff
@@ -1490,22 +1490,32 @@ def ne(self, other, level=None, fill_value=None, axis=0):  # noqa: PR01, RT01, D
...
         """
         # TODO: SNOW-1063347: Modin upgrade - modin.pandas.Series functions
+        if len(self._query_compiler.columns) == 0:
+            # pandas returns empty series when requested largest/smallest from empty series
+            return self.__constructor__(data=[], dtype=float)
         return self.__constructor__(
-            query_compiler=self._query_compiler.nsmallest(n=n, keep=keep)
+            query_compiler=self._query_compiler.nsmallest(
+                n=n, columns=self.name, keep=keep
+            )
```
sfc-gh-nkumar, line 1518: This is copied from modin repo.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13948,6 +13948,127 @@ def dt_strftime(self, date_format: str) -> None:
...
+              If True return smallest n values otherwise return largest n values.
+
+        Returns:
+            SnowflakeQueryCompiler
+        """
+        if keep not in ("first", "last", "all"):
+            raise ValueError('keep must be either "first", "last" or "all"')
+        if keep == "all":
+            method_name = "nsmallest" if ascending else "nlargest"
+            ErrorMessage.not_implemented(
```
sfc-gh-yzou, line 13975: Do we have a ticket to track the implementation? One straightforward ways is probably do a join based on the last row

tests/integ/modin/frame/test_nsmallest.py
sfc-gh-mvashishtha, line 1: Let's consolidate the `nsmallest` and `nlargest` tests into a single file with a test class that we parametrize by `method` for `['nsmallest', 'nlargest']`.

tests/integ/modin/series/test_nlargest.py
sfc-gh-mvashishtha, line 1: ditto about consolidating the nlargest and nsmallest tests

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13948,6 +13948,127 @@ def dt_strftime(self, date_format: str) -> None:
...
+            internal_frame.get_snowflake_quoted_identifiers_group_by_pandas_labels(
+                columns, include_index=False
+            )
+        )
+        ordering_columns = []
+        for label, identifiers in zip(columns, matched_identifiers):
+            if len(identifiers) == 0:
+                raise KeyError(label)
+            if len(identifiers) > 1:
+                raise ValueError(f"The column label '{label} is not unique.")
```
sfc-gh-mvashishtha, line 14009: do we have a test for this case?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13948,6 +13948,127 @@ def dt_strftime(self, date_format: str) -> None:
...
+        if keep not in ("first", "last", "all"):
+            raise ValueError('keep must be either "first", "last" or "all"')
+        if keep == "all":
+            method_name = "nsmallest" if ascending else "nlargest"
+            ErrorMessage.not_implemented(
+                f"Snowpark pandas method '{method_name}' doesn't yet support parameter keep='all'"
+            )
+        # Special case handling for unnamed series. 'columns' passed from frontend layer
+        # will be None, replace it with MODIN_UNNAMED_SERIES_LABEL.
+        if columns is None:
```
sfc-gh-yzou, line 13980: So if column is none, it means we want to get all columns, regardless it is series or dataframe, right? if it is dataframe, we can put columns = None even with dataframe, right?

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
```diff
@@ -1892,11 +1892,222 @@ def ne():
     def nlargest():
         """
         Return the first `n` rows ordered by `columns` in descending order.
+
+        Return the first `n` rows with the largest values in `columns`, in
+        descending order. The columns that are not specified are returned as
+        well, but not used for ordering.
+
+        This method is equivalent to
+        ``df.sort_values(columns, ascending=False).head(n)``, but more
```
sfc-gh-mvashishtha, line 1901: is it more performant? I would expect the SQL to be about the same for both.

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
```diff
@@ -1892,11 +1892,222 @@ def ne():
...
     def nsmallest():
         """
         Return the first `n` rows ordered by `columns` in ascending order.
+
+        Return the first `n` rows with the smallest values in `columns`, in
+        ascending order. The columns that are not specified are returned as
+        well, but not used for ordering.
+
+        This method is equivalent to
+        ``df.sort_values(columns, ascending=True).head(n)``, but more
```
sfc-gh-mvashishtha, line 2013: ditto about performance

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
```diff
@@ -1892,11 +1892,222 @@ def ne():
...
+
+        When using ``keep='last'``, ties are resolved in reverse order:
+
+        >>> df.nsmallest(3, 'population', keep='last')
+                  population  GDP alpha-2
+        Anguilla       11300  311      AI
+        Tuvalu         11300   38      TV
+        Nauru         337000  182      NR
+
+        When using ``keep='all'``, the number of element kept can go beyond ``n``
```
sfc-gh-mvashishtha, line 2082: Is there a missing period here after `n`?

src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -1598,11 +1598,189 @@ def ne():
...
+
+        See Also
+        --------
+        Series.nlargest: Get the `n` largest elements.
+        Series.sort_values: Sort Series by values.
+        Series.head: Return the first `n` rows.
+
+        Notes
+        -----
+        Faster than ``.sort_values().head(n)`` for small `n` relative to
```
sfc-gh-mvashishtha, line 1722: ditto

sfc-gh-mvashishtha (commented):
some minor comments

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13948,6 +13948,127 @@ def dt_strftime(self, date_format: str) -> None:
...
+        # Native pandas returns empty dataframe if n is negative. Set it to zero to
+        # provide same behavior.
+        n = max(0, n)
+        # Native pandas returns empty dataframe if 'columns' is a empty array. Set n
+        # to zero to provide same behavior.
+        if len(columns) == 0:
+            n = 0
+
+        # We need row position to break tie for duplicate values.
+        internal_frame = self._modin_frame.ensure_row_position_column()
```
sfc-gh-yzou, line 13997: do you need the row position column? or just the ordering column? If we do not need the row position column, we can just use the current ordering columns

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13948,6 +13948,127 @@ def dt_strftime(self, date_format: str) -> None:
...
+            )
+        )
+        ordering_columns = []
+        for label, identifiers in zip(columns, matched_identifiers):
+            if len(identifiers) == 0:
+                raise KeyError(label)
+            if len(identifiers) > 1:
+                raise ValueError(f"The column label '{label} is not unique.")
+            ordering_columns.append(
+                OrderingColumn(identifiers[0], ascending, na_last=True)
```
sfc-gh-yzou, line 14011: it here any column type that is not sortable in snowflake? for example, variant column etc?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: good point. we can work with existing ordering columns.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: good catch! added now.

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
sfc-gh-nkumar: right it's not. This got copied from pandas doc. Removed now.

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
sfc-gh-nkumar: removed.

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
sfc-gh-nkumar: Added.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: Is suggestion here to add type check?

sfc-gh-mvashishtha (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: Oh, no. let's do not do special type check, i am just curious about what happens. i think it is important for us to understand potential consequence

docs/source/modin/supported/dataframe_supported.rst
```diff
@@ -279,13 +279,13 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``ne``                      | P                               | ``level``                        |                                                    |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``nlargest``                | N                               |                                  |                                                    |
+| ``nlargest``                | P                               |                                  | ``N`` if ``keep == "all"``                         |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
```
sfc-gh-yzou, line 283: forgot to mention that before, you also need to make sure you update dataframe.rst, series.rst to make sure those two APIs occur in the generated API doc

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: Oh, what i was referring to is for example I can call the following 
```
df.nlargest(3, columns=None)
```
pandas raises an keyError. Here I believe you simply want to check if self._modin_frame.is_unnamed_series():
  columns = [MODIN_UNNAMED_SERIES_LABEL]
the logic seems simple and straightforward to understand. And maybe you can also add the above example as a test point

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13948,6 +13948,132 @@ def dt_strftime(self, date_format: str) -> None:
...
+        # Native pandas returns empty dataframe if 'columns' is a empty array. Set n
+        # to zero to provide same behavior.
+        if len(columns) == 0:
+            n = 0
+
+        internal_frame = self._modin_frame
+        # Map pandas labels to snowflake identifiers.
+        matched_identifiers = (
+            internal_frame.get_snowflake_quoted_identifiers_group_by_pandas_labels(
+                columns, include_index=False
```
sfc-gh-yzou, line 14001: want to check that we have test that order by the index columns, and case have duplicated columns among index and regular data columns for example index = ['A', 'B'], columns = ['C', 'B', 'E']?

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
sfc-gh-yzou: In fact, depends on how the implementation look like, it might be the case, for example i would expect snowflake top n performs better than order by <> limit n.  however, based on the current implementation, difficult to tell. Safe to do not mention this for now

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: Ah didn't know this method existed. Makes sense to use this.

> And maybe you can also add the above example as a test point

We already have a  test case for this in `test_nlargest_nsmallest_invalid_label`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: Are you sure variant is not sortable? Have you verified this?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: TBH given that multi-index is deprioritized. I am not sure if it best use of time to write tests for these edge cases related to multi-index.

docs/source/modin/supported/dataframe_supported.rst
sfc-gh-nkumar: Done

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: i don't think we even need multiIndex for this test, it is just test overlap between index and regular columns. 
Also if we do not intend to support multiIndex here, please make sure we raises an error if multiIndex is involved.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: No, I haven't, that is why i am asking here? did we have test for columns like variant, for example which have both string, int, timestamp etc. i think I just want to know what kind of column type we have tried, and have we noticed any behavior difference between Snowpark pandas and Pandas. 
When I look at the test case, it seems we only tried very basic numeric type.

sfc-gh-yzou: actually, i tried on a string column for native pandas, it raise TypeError. i don't think we need to have exact the same behavior, but we probably should have a test to indicate such behavior difference.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: Will add test to unblock.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: I think we need to more precise in code review comments. It's very confusing, first it seems like we aksing for type check then it seems we want to know what happens to unsortable types (which don't exist in snowflake) now we want to tests for string type. What exactly is ask here to unblock this PR?

sfc-gh-yzou: I think my ask here is what happens if user is calling nsmallest or nlargest on a column with non-numeric types, which we don't even seem to have a test case for it.  Is there any behavior difference between Snowpark pandas and Pandas? My first though is what happens if the column has a mixed data type, will it error out in Snowflake? then I tried out with just string type, it seems there is definitely a behavior difference between native pandas and Snowpark pandas.
 For example, at least i am sure sort on string type is definitely working in snowflake, but pandas doesn't allow such case, so pandas probably won't allow calling this function on columns with mixed dtype also, or maybe even variant type with the same value type. However, those case may or may not work with snowpark pandas which mainly relies on how snowflake works.

sfc-gh-yzou: i think i got the answer here already could you please maybe leave a comment here about such behavior difference between pandas and snowpark pandas, or maybe have a test case to indicate the behavior difference here.

sfc-gh-nkumar: Added test to unblock.

sfc-gh-yzou (approved):

tests/integ/modin/frame/test_nlargest_nsmallest.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+@sql_count_checker(query_count=1)
+def test_nlargest_nsmallest_large_n(snow_df, native_df, method):
+    eval_snowpark_pandas_result(
+        snow_df, native_df, lambda df: getattr(df, method)(100, "A")
+    )
+
+
+@sql_count_checker(query_count=5, join_count=1)
```
sfc-gh-helmeleegy, line 54: Comment to explain high query count?

sfc-gh-helmeleegy (approved):
Overall LGTM. Thanks!

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
 - Added support for `DataFrame.expanding` and `Series.expanding` for aggregations `count`, `sum`, `min`, `max`, `mean`, `std`, and `var` with `axis=0`.
 - Added support for `DataFrame.rolling` and `Series.rolling` for aggregation `count` with `axis=0`.
 - Added support for `Series.str.match`.
+- Added support for `DataFrame.nlargest`, `DataFrame.nsmallest`, `Series.nlargest` and `Series.nsmallest`
```
sfc-gh-helmeleegy, line 52:
```suggestion
- Added support for `DataFrame.nlargest`, `DataFrame.nsmallest`, `Series.nlargest` and `Series.nsmallest`.
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -14033,6 +14033,130 @@ def dt_strftime(self, date_format: str) -> None:
...
+            # TODO SNOW-1483214: Add support for keep='all'
+            ErrorMessage.not_implemented(
+                f"Snowpark pandas method '{method_name}' doesn't yet support parameter keep='all'"
+            )
+        # Special case handling for unnamed series. 'columns' passed from frontend layer
+        # will be None, replace it with MODIN_UNNAMED_SERIES_LABEL.
+        if self._modin_frame.is_unnamed_series():
+            columns = MODIN_UNNAMED_SERIES_LABEL
+
+        if not isinstance(columns, list):
```
devin-petersohn, line 14069:
```suggestion
        if not is_list_like(columns):
```

devin-petersohn (commented):
looks great, one minor issue

sfc-gh-dpetersohn (approved):
lgtm

PR #1785, opened 2024-06-15
Index rename, set_names, name and names

sfc-gh-samdesai1:

   

   Fixes SNOW-1458122

   Implements `Index.name`, `Index.names`, `Index.set_names` and `Index.rename`


src/snowflake/snowpark/modin/plugin/_internal/type_utils.py
```diff
@@ -92,7 +92,7 @@
     (np.half, FloatType()),
     (np.float16, FloatType()),
     (np.float64, DoubleType()),
-    (np.float_, DoubleType()),
+    (np.float_, DoubleType()),  # type: ignore
```
sfc-gh-vbudati, line 95: why is `# type: ignore` here and below?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -600,14 +599,19 @@ def set_names(
...
+        if self.is_lazy:
+            if not inplace:
+                return Index(
+                    self._query_compiler.set_index_names(names),
+                )
+            else:
+                self._set_names(names)
+                return None
+        else:
+            return self.to_pandas().set_names(names, level=level, inplace=inplace)
```
sfc-gh-vbudati, line 614: `_set_names` has logic to handle the non-lazy version of setting values without calling `to_pandas` is it possible to reuse that functionality or does that handle a different case?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1346,8 +1350,7 @@ def reindex(self) -> None:
         """
         # TODO: SNOW-1458121 implement reindex
 
-    @index_not_implemented()
-    def rename(self) -> None:
+    def rename(self, name: Hashable = None, inplace: bool = False) -> Self | None:
         """
```
sfc-gh-vbudati, line 1354: add examples in the docstring for rename

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: That handles a different case, that has no return, so it can be reused for the inplace method. However, when we are not in place, we need to create a new index, so we use the query compiler method `set_index_names`, which creates a new query compiler to pass to the index constructor

PR #1787, opened 2024-06-17
SNOW-1486524: Upgrade Modin version to 0.28.3 (pins numpy<2)

sfc-gh-dpetersohn:

   

   Fixes SNOW-1486524

Upgrade Modin version to 0.28.3 (pins numpy<2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
 - Added support for `DataFrame.expanding` and `Series.expanding` for aggregations `count`, `sum`, `min`, `max`, `mean`, `std`, and `var` with `axis=0`.
 - Added support for `DataFrame.rolling` and `Series.rolling` for aggregation `count` with `axis=0`.
 - Added support for `Series.str.match`.
+- Upgrade supported Modin version to 0.28.3
 
 #### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -35,6 +35,7 @@
 MODIN_REQUIREMENTS = [
     *PANDAS_REQUIREMENTS,
     f"modin{MODIN_DEPENDENCY_VERSION}",
+    "pandas==2.2.1",
```
sfc-gh-nkrishna, line 38: If Modin installs pandas as well, I don't believe we need the 2.2.1 pin here?

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
 - Added support for `DataFrame.expanding` and `Series.expanding` for aggregations `count`, `sum`, `min`, `max`, `mean`, `std`, and `var` with `axis=0`.
 - Added support for `DataFrame.rolling` and `Series.rolling` for aggregation `count` with `axis=0`.
 - Added support for `Series.str.match`.
+- Upgrade supported Modin version to 0.28.3
```
sfc-gh-nkrishna, line 52: I would say this doesn't need a CHANGELOG

sfc-gh-nkrishna (commented):
Are there any docs that need updating?

setup.py
sfc-gh-joshi: Modin 0.28.1 was a release we specially requested, which pins the patch version 2.2.1; modin 0.28.3 only pins minor version 2.2.x. I don't think 2.2.3 (the latest available version) has any breaking changes, but it's not available in snowflake anaconda and might cause problems if we don't pin here.

PR #1789, opened 2024-06-17
SNOW-1454714 Make modin QA tests green

sfc-gh-azhan:

   

   Fixes SNOW-1454714

The most change here is to change assertion for integer columns to not verify its physical length, e.g., in8 vs. int64. The physical length of integers are managed by Snowflake engine internally and can change overtime and also won't be treated as a behavior change.


tests/integ/modin/test_qcut.py
```diff
@@ -278,46 +278,3 @@ def test_qcut_invalid_quantiles_negative(q):
...
     with pytest.raises(expected_exception, match=expected_message):
         snow_s.quantile(q)
-
-
-@sql_count_checker(query_count=7)
-def test_qcut_two_columns():
-    # reported by Mats Stewall, applying qcut twice leads to exploding SQL query.
-    # attempt finding a remedy
-
-    DATA_PATH = "SNOWFLAKE_SAMPLE_DATA.TPCH_SF1"
```
sfc-gh-azhan, line 288: remove this test because no such data in QA

tests/integ/modin/frame/test_fillna.py
```diff
@@ -574,6 +574,9 @@ def test_df_fillna_method_reindexed_df_reordered_columns(method):
 
 
 @sql_count_checker(query_count=0)
+@pytest.mark.xfail(
+    reason="TODO SNOW-1489309 investigate why it starts to work now on qa"
```
sfc-gh-nkrishna, line 578: I'm assuming it doesn't work now on qa?

```suggestion
    reason="TODO SNOW-1489309 investigate failure on qa"
```

sfc-gh-nkrishna (approved):
One comment, but LGTM otherwise

tests/integ/modin/frame/test_fillna.py
sfc-gh-azhan: aha, it's a negative test so the feature works now but the test fails.

sfc-gh-joshi: Do you know what the new output is? Is it what we expect, or is something in the VARIANT column getting cast that wasn't before?

sfc-gh-azhan: Didn't go too much detail. I will investigate this diff right after.

sfc-gh-azhan: What I know is it does not raise exception any more.

sfc-gh-dpetersohn (approved):
LGTM

PR #1790, opened 2024-06-17
[feature branch] AST: DataFrame.to_df() + cleanup

sfc-gh-azwiegincew:
I added coverage for `DataFrame.to_df()`.

Cleaned up a little bit while I was at it:
 * Renamed fill_const_ast to build_const_from_python_val.
 * Renamed fill_src_position to set_src_position.
 * Introduced the helper fn parse_positional_args_to_list_variadic based on parse_positional_args_to_list.
 * Introduced DataFrame.set_ast_ref to avoid direct bitfield manipulation in many other places.
 * Because DataFrame.select() is used within the implementation of other functions, I created a new parameter `_ast_stmt` to suppress local generation of the AST.
 * Test assertions now emit Base64 for help in debugging.


src/snowflake/snowpark/dataframe.py
```diff
@@ -1095,6 +1112,7 @@ def select(
             Union[ColumnOrName, TableFunctionCall],
             Iterable[Union[ColumnOrName, TableFunctionCall]],
         ],
+        _ast_stmt = None,
```
sfc-gh-oplaton, line 1115: Add type annotation?

sfc-gh-oplaton (approved):

PR #1794, opened 2024-06-17
[SERVER-SIDE] Add test infrastructure to track all requests have dataframeAST field.

sfc-gh-lspiegelberg:
Vendors snowflake vcrpy from https://github.com/Snowflake-Labs/snowflake-vcrpy (could not get install working, therefore vendoring it) with custom Snowflake changes to track requests in vendored urllib3 within the snowflake python connector.

Add decorator `check_ast_encode_invoked` (applied with `autouse=True` to all tests) which checks that every query send contains `dataframeAst` property for phase 0, and errors out together with traceback information whenever tests need to be fixed / APIs are missing that need to be encoded within the AST.

You can skip the `vendored` folder when reviewing.


tests/integ/conftest.py
```diff
@@ -252,3 +293,170 @@ def temp_schema(connection, session, local_testing_mode) -> None:
...
+        request_id = kwargs["requestId"]
+
+        self.request_ids.append(request_id)
+        self.queries.append(query_record)
+        self.tracebacks.append(formatted_lines)
+
+
+# Disable phase0 Snowpark IR check by setting autouse=False,
+# by default enabled for all tests (except modin/) to check that Snowpark python requests contain
+# the dataframeAST field holding the encoded representation.
```
sfc-gh-azwiegincew, line 346: dataframeAst. I think the standard terminology in JSON is "property" instead of "field".

sfc-gh-azwiegincew (approved):

tests/integ/conftest.py
sfc-gh-lspiegelberg: let me update the comment in that case!

PR #1796, opened 2024-06-17
[SERVER-SIDE] Remove temporarily Modin tests as Modin is incompatible with Python 3.8

sfc-gh-lspiegelberg:
Remove tests so test suite runs under Python 3.8.


sfc-gh-azwiegincew (approved):

devin-petersohn:
Is it not possible to upgrade to Python 3.9? It's EOL in ~4 months.

sfc-gh-lspiegelberg:
> Is it not possible to upgrade to Python 3.9? Python 3.8 EOL in ~4 months.
> 
> edit: make it clear what is eol

I'd actually like to use directly python 3.11. The obstacle is less here client-side, but more server-side as the toolchains right now work with python 3.8 only. Upgrading to 3.11 requires probably a lift of 2-4 weeks, therefore to not stall development we're currently using python 3.8. Before merging the server-side snowpark branch back, we can find a path forward to resolve the python version issue.

PR #1799, opened 2024-06-18
SNOW-1489244: Add test for old numpy aliases

sfc-gh-joshi:

   

   Fixes SNOW-1489244


tests/integ/modin/test_old_numpy_aliases.py
```diff
@@ -0,0 +1,19 @@
...
+
+import modin.pandas as pd
+import numpy as np
+
+import snowflake.snowpark.modin.plugin  # noqa: F401
+from tests.integ.modin.sql_counter import sql_count_checker
+
+
+@sql_count_checker(query_count=0)
+def test_old_numpy_aliases():
```
sfc-gh-yzou, line 13: can you separate this test into two 
1) check the nan compatibility, where we want to check if s.isnan() still gives the correct result
2) check the float_ dtype compatibility, we want to create a series with dype(np.float__), and check if the mapping to snowpark type is correct, and the type mapped back is correct, like the test we had before

tests/integ/modin/test_old_numpy_aliases.py
```diff
@@ -0,0 +1,19 @@
...
+    # Test to ensure that the np.float_ and np.NaN aliases (removed in numpy 2.x)
+    # remain valid for use in Snowpark pandas, which is still on numpy 1.26.
+    s = pd.Series([0.1, np.NaN])
```
sfc-gh-yzou, line 16: also another question, right now this runs with our modin test env, it will pass now. i assume we will need to move that test suite to a separate env once we move on to new pandas version and numpy version

sfc-gh-joshi: I'm not sure if we'll be able to keep backwards compatibility once we're on a version of modin that supports numpy 2.x. @sfc-gh-dpetersohn would you know the answer to this?

sfc-gh-yzou: Sorry, by backward compatibility i meant our code should continue work with numpy 2.0 and numpy 1.46.x, which I believe you already resolved this issue now. 
However, this test will error out once our test env is only set up for numpy 2.0. In order for those test to continue running, we will need to setup a test env with numpy 1.46.x, similar as what you did for the python 3.8 test

tests/integ/modin/test_old_numpy_aliases.py
sfc-gh-joshi: done

sfc-gh-yzou (approved):

sfc-gh-aalam (approved):

PR #1801, opened 2024-06-18
SNOW-1489207: [Local Testing] Fix astype precision error

sfc-gh-jrose:

   

   Fixes SNOW-1489207

   This PR updates the logic of a utility function `fix_drift_between_column_sf_type_and_dtype`. Previously it cast DecimalType object to int64 pandas columns, but this is incorrect in many cases because DecimalTypes can contain decimal precision data. I've also changed the cast to ignore errors so that cases where the conversion fails can still attempt comparison with the original datatypes. I think this is more in line with the logic in live mode, but may still have edge cases that differ. I'll open a ticker to investigate further.


tests/unit/test_expression_dependent_columns.py
```diff
@@ -204,7 +204,6 @@ def test_cast():
 @pytest.mark.parametrize(
     "expression_class",
     [
-        BinaryExpression,
```
sfc-gh-jrose, line 207: BinaryExpression is already tested further down in this parameter list.

tests/integ/scala/test_column_suite.py
```diff
@@ -92,7 +92,7 @@ def test_alias(session):
     )
 
 
-def test_equal_and_not_equal(session):
+def test_equal_and_not_equal(session, local_testing_mode):
```
sfc-gh-stan, line 95: `local_testing_mode` seems to be unused?

sfc-gh-stan (approved):

sfc-gh-jdu (approved):

PR #1804, opened 2024-06-18
SNOW-1489361: Add support for GroupBy.all/any

sfc-gh-joshi:

   

   Fixes SNOW-1489361

This PR adds support for GroupBy.all and GroupBy.any.


sfc-gh-nkrishna (approved):
LGTM, might be worth doing a run on Jenkins (or manual) just to check that you have removed `any/all` from the unsupported APIs list if it is there

sfc-gh-joshi:
> LGTM, might be worth doing a run on Jenkins (or manual) just to check that you have removed any/all from the unsupported APIs list if it is there

Good catch, I forgot to do that.

CHANGELOG.md
```diff
@@ -45,11 +45,12 @@
 - Added support for `Series.dt.dayofweek`, `Series.dt.day_of_week`, `Series.dt.dayofyear`, and `Series.dt.day_of_year`.
 - Added support for `Series.str.__getitem__` (`Series.str[...]`).
 - Added support for `Series.str.lstrip` and `Series.str.rstrip`.
-- Added support for `DataFrameGroupby.size` and `SeriesGroupby.size`.
+- Added support for `DataFrameGroupBy.size` and `SeriesGroupBy.size`.
```
sfc-gh-dpetersohn, line 48: 💯

sfc-gh-dpetersohn (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2735,6 +2733,27 @@ def groupby_agg(
...
+                    [
+                        label
+                        for label in query_compiler._modin_frame.data_columns_index
+                        if label not in by_list
+                    ]
+                )
+                == 0
+            )
+            if not empty_columns and not all(
+                is_bool_dtype(t) or is_integer_dtype(t)
```
sfc-gh-yzou, line 2748: Let's do not do specialized type check here to be consistent among all aggregations. we simply let the aggregation goes through, and let snowflake engine decides whether it is valid to apply the aggregation or not. One of the very typical use case is variant, if customer has a variant column with all numeric type, some aggregation will continue be valid there, furthermore, the variant column can be produced by snowpark pandas during some operaiton like apply

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2017,33 +2017,32 @@ def _bool_reduce_helper(
 
         empty_value: bool
             The value returned for an empty dataframe.
-        reduce_op: {"and", "or"}
-            The name of the boolean operation to apply.
+        agg_func: {"all", "any"}
```
sfc-gh-yzou, line 2020: will you be able to move this function to aggregation_utils to centralize all aggregation related helper and utils? you cam do this as a follow up pr

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: Sure, I can remove it for now. The only concern I have is that in the future we might want to emit different SQL for non-numeric/boolean columns. Right now, we're using the `BOOLAND_AGG`/`BOOLOR_AGG` SQL functions on columns to support numbers/bools, but it's made clear in [documentation](https://docs.snowflake.com/en/sql-reference/functions/booland_agg#usage-notes) that these functions are not meant to be used with string types. Given that Python's notion of "truthiness" for strings is different from SQL's, if we want to support cases like `pd.Series(["abc", "", None]).all()` (which is at least moderately common in pandas), I think it would make sense to emit different SQL.

PR #1807, opened 2024-06-20
Index values, item, tolist, to_series, to_frame

sfc-gh-samdesai1:

   

   Fixes SNOW-1458117

   Implements `Index.values`, `Index.item()`, `Index.tolist()`, `Index.to_series()`, `Index.to_frame()`


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1679,6 +1679,31 @@ def get_index_names(self, axis: int = 0) -> list[Hashable]:
             else self._modin_frame.data_column_pandas_index_names
         )
 
+    def get_index_values(self) -> AnyArrayLike:
```
sfc-gh-azhan, line 1682: can you reuse `to_numpy` which is used by df.values?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1679,6 +1679,31 @@ def get_index_names(self, axis: int = 0) -> list[Hashable]:
...
+        """
+        Return an array representing the data in the Index.
+        """
+        return snowpark_to_pandas_helper(
+            self._modin_frame.ordered_dataframe.select(
+                self._modin_frame.index_column_snowflake_quoted_identifiers
+            )
+        ).values.flatten()
+
+    def get_index_item(self) -> Hashable:
```
sfc-gh-azhan, line 1692: can we reuse `take_2d_positional` which is used by df indexing, e.g., df.iloc[0]

tests/integ/modin/index/test_index_methods.py
```diff
@@ -197,3 +218,138 @@ def test_df_index_shape(native_df):
...
+
+@pytest.mark.parametrize("native_df", TEST_DFS)
+@sql_count_checker(query_count=1)
+def test_df_index_to_list(native_df):
+    snow_df = pd.DataFrame(native_df)
+    assert_equal(native_df.index.to_list(), snow_df.index.to_list())
+    assert_equal(native_df.columns.to_list(), snow_df.columns.to_list())
+
+
+@sql_count_checker(query_count=6)
```
sfc-gh-azhan, line 268: the query count is too high, why?

tests/integ/modin/index/test_index_methods.py
```diff
@@ -197,3 +218,138 @@ def test_df_index_shape(native_df):
...
+            snow_df.columns.to_series(index=range(len(snow_df.columns)), name=True),
+            check_dtype=False,
+            check_index_type=False,
+        )
+
+
+@pytest.mark.parametrize("native_index", NATIVE_INDEX_TEST_DATA)
+def test_index_to_frame(native_index):
+    snow_index = pd.Index(native_index)
+    with SqlCounter(query_count=5):
```
sfc-gh-azhan, line 314: too high qc

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-samdesai1: no, because that method does not get index values correctly

sfc-gh-azhan: why?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1492,10 +1493,14 @@ def item(self) -> None:
         ValueError
             If the data is not length = 1.
         """
-        # TODO: SNOW-1458117 implement item
+        if len(self) == 1:
+            return self.tolist()[0]
```
sfc-gh-azhan, line 1497: No this is not efficient which will pull the whole index locally.

sfc-gh-samdesai1: yep but the whole index is only one value in this case, that should be fine for us right?

tests/integ/modin/frame/test_setitem.py
```diff
@@ -351,22 +351,29 @@ def func_insert_new_column(df, column):
 
         # convert to snow objects when testing for Snowpark pandas
         if isinstance(df, pd.DataFrame):
-            if isinstance(column, native_pd.Index):
+            if isinstance(column, native_pd.Index) and not isinstance(
+                column, native_pd.DatetimeIndex
+            ):
                 column = pd.Index(column)
-            elif isinstance(column, native_pd.Series):
+            if isinstance(column, native_pd.Series):
```
sfc-gh-vbudati, line 358: why did you change the elif to if?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -688,7 +688,7 @@ def to_numpy(
         # here it's overwritten to actually perform numpy conversion, i.e. return an actual numpy object
         return self.to_pandas().to_numpy(dtype=dtype, na_value=na_value, **kwargs)
 
-    def repartition(self, axis: Any = None) -> "SnowflakeQueryCompiler":
+    def repartition(self, axis: Optional[Any] = None) -> "SnowflakeQueryCompiler":
```
sfc-gh-yzou, line 691: why is this change here? if not need by the current pr, please revert the change

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: was it because of the numpy data format? then can we just call to_pandas on this ?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -43,10 +43,10 @@ class Index:
     def __init__(
         self,
         # TODO: SNOW-1481037 : Fix typehints for index constructor, set_query_compiler and set_local_index
-        data: ArrayLike | Any = None,
+        data: ArrayLike | Any | None = None,
```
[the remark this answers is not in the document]
sfc-gh-samdesai1: The linter seems to automatically be changing some of this stuff, but I can change it back if we need to

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: This API is a bit weird, it is only supposed to work when the length of the index is 1. Native pandas also raises a value error when the index length is not 1

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1492,10 +1493,14 @@ def item(self) -> None:
...
+        if len(self) == 1:
+            return self.tolist()[0]
+        raise ValueError("can only convert an array of size 1 to a Python scalar")
 
-    @index_not_implemented()
-    def to_series(self) -> None:
+    # TODO: SNOW-1481037 : Fix typehints for index constructor, set_query_compiler and set_local_index
+    def to_series(
+        self, index: Index | None = None, name: Hashable | None = None
+    ) -> Any:
```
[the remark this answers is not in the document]
sfc-gh-samdesai1: There is an issue with that, thats why I added the TODO there

docs/source/modin/supported/index_supported.rst
```diff
@@ -140,17 +140,17 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``astype``                  | P                               |                                  |                                                    |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``item``                    | N                               |                                  |                                                    |
+| ``item``                    | Y                               |                                  |                                                    |
```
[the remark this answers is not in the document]
sfc-gh-samdesai1: Yep, they are in indexing.rst

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-samdesai1: I've removed this now

tests/integ/modin/frame/test_setitem.py
sfc-gh-samdesai1: Putting this back now, it was just a typo

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1519,10 +1524,19 @@ def to_series(self) -> None:
...
+        pandas_index = self.to_pandas()
+        if index is None:
+            index = pandas_index
+
+        if name is None:
+            name = self.name
+
+        return Series(data=pandas_index, index=index, name=name)
+
+    # TODO: SNOW-1481037 : Fix typehints
```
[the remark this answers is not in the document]
sfc-gh-samdesai1: Yes the goal is to fix the typehints in a new PR. I would like to fix them in the current PRs but it is more challenging than it looks because of circular imports and the linter removing any object names in quotes

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -201,13 +201,31 @@ def check_lazy(*args: Any, **kwargs: Any) -> Any:
                     returned_value = Index(returned_value, convert_to_lazy=False)
                 # Some methods also return a tuple with a pandas Index, so convert the tuple's first item to a modin Index
                 # Examples of this are `_get_indexer_strict` and `sort_values`
+
+                # We want to return a native pandas series for value counts because our loc/iloc functionality relies on it
+                # If we returned a snowpark pandas series, this would result in querying data many times
+                elif func_name == "value_counts":
```
sfc-gh-azhan, line 207: I'd like to learn more about this. Ideally, we should try to avoid this case.

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -281,8 +299,12 @@ def values(self) -> ArrayLike:
         >>> idx.values
         array([1, 2, 3])
         """
-        # TODO: SNOW-1458117 implement values
-        return self.to_pandas().values
+        return (
+            self._query_compiler.reset_index()
+            .drop(columns=self._query_compiler.columns)
```
sfc-gh-azhan, line 304: Should we move this drop columns to the constructor so we can guarantee all index's df are df with empty data columns?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1520,10 +1545,26 @@ def to_series(self) -> None:
...
         """
-        # TODO: SNOW-1458117 implement to_series
+        from snowflake.snowpark.modin.pandas import Series
 
-    @index_not_implemented()
-    def to_frame(self) -> None:
+        if name is None:
+            name = self.name
+        new_qc = (
+            self._query_compiler.drop(columns=self._query_compiler.columns)
```
sfc-gh-azhan, line 1553: move to constructor

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1520,10 +1545,26 @@ def to_series(self) -> None:
...
+            .reset_index()
+            .set_index(["index"], drop=False)
```
sfc-gh-azhan, line 1555: nice found!

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1520,10 +1545,26 @@ def to_series(self) -> None:
...
+            .set_index_names([None])
+        )
+        if index is not None:
+            new_qc = new_qc.set_index_from_series(Series(index)._query_compiler)
```
sfc-gh-azhan, line 1559: we should have a way to set_index from index. Can you double check? If not, please put a todo with https://snowflakecomputing.atlassian.net/browse/SNOW-1458137, we should make sure implement it there.

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1546,7 +1587,23 @@ def to_frame(self) -> None:
         Index.to_series : Convert an Index to a Series.
         Series.to_frame : Convert Series to DataFrame.
         """
-        # TODO: SNOW-1458117 implement to_frame
+        from snowflake.snowpark.modin.pandas import DataFrame
+
+        if name is None:
+            name = self.name
+        new_qc = self._query_compiler.drop(
```
sfc-gh-azhan, line 1594: drop it when we construct it.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1511,7 +1511,7 @@ def index(self) -> Union["pd.Index", native_pd.MultiIndex]:
         if self.is_multiindex():
             return self._modin_frame.index_columns_pandas_index
         else:
-            return pd.Index(self)
+            return pd.Index(self, tupleize_cols=False)
```
sfc-gh-azhan, line 1514: why?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4983,7 +4983,8 @@ def set_index_from_columns(
             keys
         ):
             # Error checking for missing labels is already done in frontend layer.
-            index_column_snowflake_quoted_identifiers.append(ids[0])
+            if ids:
```
sfc-gh-azhan, line 4986: see "# Error checking for missing labels is already done in frontend layer."
what's the error did you see here? And why it happens now?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1520,10 +1545,26 @@ def to_series(self) -> None:
         Index.to_frame : Convert an Index to a DataFrame.
         Series.to_frame : Convert Series to DataFrame.
         """
-        # TODO: SNOW-1458117 implement to_series
+        from snowflake.snowpark.modin.pandas import Series
 
-    @index_not_implemented()
-    def to_frame(self) -> None:
+        if name is None:
```
sfc-gh-yzou, line 1550: the to_series and to_frame seems done in a very similar way, where we can do 
1) reset the index to to make the original column a data column, and it will also create a default index that is row_number
2) set_index if needed, the index can be a given index or original index for series, and the original index or None for dataframe
3) correct the data column name as post processing, for series, we need to set the result series name, for dataframe, need to rename the column if name is given.
Also, you might want to pay attention to the final name, it seems if the original index has no name, and no name is given, pandas gives a number label 0 as pandas label for the data column.
Step 1 and 2 can be extracted as a common function used by both series and dataframe

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1688,8 +1745,7 @@ def tolist(self) -> list:
         >>> idx.to_list()
         [1, 2, 3]
         """
-        # TODO: SNOW-1458117 implement tolist
-        return self.to_pandas().tolist()
+        return self.values.tolist()
```
sfc-gh-yzou, line 1748: the original implementation seems fine to me, what is the reason to update this to values?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -281,8 +299,12 @@ def values(self) -> ArrayLike:
         >>> idx.values
         array([1, 2, 3])
         """
-        # TODO: SNOW-1458117 implement values
-        return self.to_pandas().values
+        return (
+            self._query_compiler.reset_index()
+            .drop(columns=self._query_compiler.columns)
+            .to_numpy()
+            .reshape(-1)
```
sfc-gh-yzou, line 306: what is the reshape of -1 doing here? can you add a comment?

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-yzou, line 304: it seems a lot of API is trying to do drop and reset_index here, might be useful to extract a common utility here like _convert_index_column_to_data_column, and used for all functions

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-yzou, line 207: if we don't do this here, what happens?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou, line 1514: why do we need tupleize_cols=False here?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou, line 4986: Not sure if I got this part, this seems indicating there could be case that ids is None or empty, however, how does this related to your change ?

tests/integ/modin/index/test_index_methods.py
```diff
@@ -212,3 +219,173 @@ def test_df_index_shape(native_df):
...
+        native_index.to_series(index=range(10, 10 + len(native_index)), name="name"),
+        snow_index.to_series(index=range(10, 10 + len(snow_index)), name="name"),
+        check_index_type=False,
+    )
+
+
+@sql_count_checker(query_count=4, join_count=1)
+@pytest.mark.parametrize("native_df", TEST_DFS)
+def test_df_index_columns_to_series(native_df):
+    snow_df = pd.DataFrame(native_df)
```
sfc-gh-yzou, line 315: do we have any test case whose index actually has a name?

tests/integ/modin/index/test_index_methods.py
```diff
@@ -212,3 +219,173 @@ def test_df_index_shape(native_df):
...
+    assert_series_equal(
+        native_df.index.to_series(),
+        snow_df.index.to_series(),
+        check_dtype=False,
+        check_index_type=False,
+    )
+    assert_series_equal(
```
sfc-gh-yzou, line 322: Could you separate the test case that uses the original index and have a new index? where no join count should be needed when original index used

tests/integ/modin/index/test_index_methods.py
```diff
@@ -212,3 +219,173 @@ def test_df_index_shape(native_df):
...
+        native_df.columns.to_series(),
+        snow_df.columns.to_series(),
+        check_dtype=False,
+        check_index_type=False,
```
sfc-gh-yzou, line 326: why are we skipping the index type check? also, for this tests I don't expect there should be any dtype change, you should be able to enable dtype check also

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: It should not be too bad because length is a cached property? Maybe we can discuss this more

src/snowflake/snowpark/modin/plugin/_internal/index.py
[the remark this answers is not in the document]
sfc-gh-samdesai1: The linter removes any quoted typehints and I cannot directly import Series there because of circular import errors

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-samdesai1: In the case where the index is lists, we do not want to create a multi-index

sfc-gh-samdesai1: In the case where the index is lists, we do not want to create a multi-index

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-samdesai1: Sorry, that was some experimentation, will remove it now

sfc-gh-samdesai1: Sorry, that was some experimentation, will remove it now

tests/integ/modin/index/test_index_methods.py
sfc-gh-samdesai1: I did not add that because we do not fully support index name api now

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: It's one query either way, but I wanted to use existing methods rather than calling to_pandas()

tests/integ/modin/index/test_index_methods.py
sfc-gh-samdesai1: We don't fully support index types, if we don't skip it, we look to check index inferred type which is not implemented. The dtype situation is a bit weird, for empty snowpark pandas df, the dtype for columns is object, but for native df it is int64. I'll ping on slack to discuss this

tests/integ/modin/index/test_index_methods.py
sfc-gh-samdesai1: yep makes sense

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: Yep sure. See [here](https://github.com/snowflakedb/snowpark-python/blob/68a2051b8fa5a5b8f436487cff2c3e6610d8eaaa/src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py#L778)

sfc-gh-samdesai1: Some of the loc methods go to 18 queries because we query data each time, see [here](https://github.com/snowflakedb/snowpark-python/blob/68a2051b8fa5a5b8f436487cff2c3e6610d8eaaa/src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py#L778) for the details

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: Yep I will do that

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: The reshape simply puts it into a "row vector" like native pandas. It changes shape from (3,) to (3, 1) for example

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: Ok makes sense, this will be a pretty small method though

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: I don't think I can do step 2 in a common function because we want to set index pretty differently in both cases

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: Yep

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: a series does not have a set_index method. For us to change a series index, we can do `ser.index = my_index`

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: yep

CHANGELOG.md
```diff
@@ -68,9 +68,10 @@
 - `pd.read_csv` reads using the native pandas CSV parser, then uploads data to snowflake using parquet. This enables most of the parameters supported by `read_csv` including date parsing and numeric conversions. Uploading via parquet is roughly twice as fast as uploading via CSV.
 - Initial work to support an Index directly in Snowpark pandas. Support for Index as a first-class component of Snowpark pandas is coming soon.
 - Added lazy index constructor and support for `len`, `shape`, `size`, `empty`, `to_pandas()` and `names`. 
-- Added support for `Index.copy()`
 - For `df.index`, Snowpark pandas creates a lazy index object. 
 - For `df.columns`, Snowpark pandas supports a non-lazy version of an Index since the data is already stored locally
+- Added support for `Index.copy()`
```
sfc-gh-azhan, line 73: should not move this line

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-azhan: I would prefer updating the code in indexing.py to call `to_pandas` there. Or you could do `return pd.Index(value_counts, convert_to_lazy=False)`. The point is to make sure the `value_counts` always return a pd.Index.

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-azhan: I believe both `to_series` and `to_frame` are almost the same right? We only need to handle name differently

CHANGELOG.md
sfc-gh-samdesai1: wouldn't it be better if the APIs mentioned were all at the bottom?

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: No, `to_frame` sets the index as the existing index or a default index, it does not allow users to input new index data. `to_series` allows users to input index data, that is why there is one join for tests that include passing an index

CHANGELOG.md
sfc-gh-azhan: Then why not combine into one line?

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: `value_counts` is supposed to return a `Series`, since the method is unimplemented and the existing functionality relies on it beiing a native_series, I chose not to make any changes there in this PR. I was thinking that this would be something to do when we implement `value_counts`, and we have it return a snowpark pandas series

CHANGELOG.md
sfc-gh-samdesai1: Because they are separate PRs, but I can combine if you think that is better

sfc-gh-azhan: There are two options:
1. keep it as is for what happened out side of this PR. Because when we release, we will visit this and reorganize it later
2. If you want to update both make sure update them correctly. E.g., line 74 does not say what support are they? Are they the APIs for series, df, or index?

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-azhan: oh, right, We should keep it consistent to return pd.Series. So how about add to_pandas in the indexing.py?

CHANGELOG.md
sfc-gh-samdesai1: Ok I will update it to be better

CHANGELOG.md
```diff
@@ -68,9 +68,10 @@
...
+- Added support for `Index.values`, `Index.item()`, `Index.tolist()`, `Index.to_series()` and `Index.to_frame()`
```
sfc-gh-azhan, line 74: since you have touched the previous one, how to combine them to 
```
Added support for Index APIs: `copy`, `values`, `item`, `tolist`, `to_series` and `to_frame`
```

CHANGELOG.md
sfc-gh-nkrishna: IMO this should be combined into one line, the release notes are customer facing so condensing it is better. You could even say something like "Added support for the following `Index` APIs: `values`, `item()`, etc.

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1493,10 +1509,26 @@ def item(self) -> None:
...
+        item = self._query_compiler.take_2d_positional(
+            index=slice(0, 2), columns=[0]
+        ).index.to_pandas()
+        if len(item) == 1:
+            return item[0]
+        raise ValueError("can only convert an array of size 1 to a Python scalar")
 
-    @index_not_implemented()
-    def to_series(self) -> None:
+    def _convert_index_column_to_data_column(self) -> Any:
```
sfc-gh-azhan, line 1519: Is this really necessary? Since it is just one line.

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1493,10 +1509,26 @@ def item(self) -> None:
         ValueError
             If the data is not length = 1.
         """
-        # TODO: SNOW-1458117 implement item
+        item = self._query_compiler.take_2d_positional(
```
sfc-gh-azhan, line 1512: add a comment why we are doing this

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -257,7 +270,6 @@ def to_pandas(self) -> native_pd.Index:
         return self._index
 
     @property
-    @is_lazy_check
     def values(self) -> ArrayLike:
         """
         Return an array representing the data in the Index.
```
sfc-gh-vbudati, line 275: did you mean to not implement values? If you're not implementing it, add the TODO back

sfc-gh-samdesai1: No, calling to_pandas() is implementing it, since we need to materialize the values anyways

CHANGELOG.md
```diff
@@ -68,9 +68,9 @@
 - `pd.read_csv` reads using the native pandas CSV parser, then uploads data to snowflake using parquet. This enables most of the parameters supported by `read_csv` including date parsing and numeric conversions. Uploading via parquet is roughly twice as fast as uploading via CSV.
 - Initial work to support an Index directly in Snowpark pandas. Support for Index as a first-class component of Snowpark pandas is coming soon.
 - Added lazy index constructor and support for `len`, `shape`, `size`, `empty`, `to_pandas()` and `names`. 
-- Added support for `Index.copy()`
```
sfc-gh-yzou, line 71: can we keep this here, there is no need to make all series to stay together in a change log, a change log is just a log for changes

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1493,10 +1504,25 @@ def item(self) -> None:
         ValueError
             If the data is not length = 1.
         """
-        # TODO: SNOW-1458117 implement item
+        # slice the first two elements of the index and materialize them
+        item = self._query_compiler.take_2d_positional(
+            index=slice(2), columns=[0]
```
sfc-gh-yzou, line 1509: typically, index column should have no columns, under that case, will columns[0] error out?  if you just put columns = [], would that just work?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1520,10 +1546,28 @@ def to_series(self) -> None:
...
+            new_qc = new_qc.set_index_from_series(Series(index)._query_compiler)
+
+        # create series and set the name
+        ser = Series(query_compiler=new_qc)
+        ser.name = name
+        return ser
+
+    @is_lazy_check
+    # TODO: SNOW-1481037 : Fix typehints
+    def to_frame(self, index: bool = True, name: Hashable | None = None) -> Any:
```
sfc-gh-yzou, line 1570: you mentioned that you were using Any because it got reverted, right? if you quote your return type like "DataFrame" does it automatically revers?

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1546,7 +1590,29 @@ def to_frame(self) -> None:
         Index.to_series : Convert an Index to a Series.
         Series.to_frame : Convert Series to DataFrame.
         """
-        # TODO: SNOW-1458117 implement to_frame
+        from snowflake.snowpark.modin.pandas import DataFrame
+
+        # save the index name and the given name to be the index name and column name of the new df respectively
+        index_name = self.name
```
sfc-gh-yzou, line 1596: it seems that we don't really need the variable index_name here

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1546,7 +1590,29 @@ def to_frame(self) -> None:
...
+        column_name = name
+
+        # if we do not have an index name and the given name is also None, we default to a dataframe column name of 0
+        if name is None and index_name is None:
+            column_name = 0
+        # if the index has a name, we set the column name as this name
+        elif name is None:
+            column_name = index_name
+
+        # Do a reset index to create a data column, filling it with the values of self and assigning it column name
```
sfc-gh-yzou, line 1606: "create a data column" -> "convert the index column to a data column" and the index becomes the pandas default index (row position).

sfc-gh-yzou: also add an concrete example as part of the comment to be clear

tests/integ/modin/index/test_index_methods.py
```diff
@@ -212,3 +219,179 @@ def test_df_index_shape(native_df):
...
+    )
+    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+        snow_df.columns.to_frame(),
+        native_df.columns.to_frame(),
+        check_index_type=False,
+        check_column_type=False,
+    )
+
+    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+        snow_df.index.to_frame(index=False, name=1),
```
sfc-gh-yzou, line 387: can you put index value, and name value as parameters, similar as how you put native df? also, do we need to test over all dfferent dfs, or shall we test with different NATIVE_INDEX_TEST_DATA

tests/integ/modin/index/test_index_methods.py
```diff
@@ -212,3 +219,179 @@ def test_df_index_shape(native_df):
...
+    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+        snow_index.to_frame(),
+        native_index.to_frame(),
+        check_index_type=False,
+        check_column_type=False,
+    )
+
+    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(
+        snow_index.to_frame(index=False, name="name"),
+        native_index.to_frame(index=False, name="name"),
```
sfc-gh-yzou, line 363: try to put index and name as test parameters to make each test point clear

tests/integ/modin/index/test_index_methods.py
```diff
@@ -212,3 +219,179 @@ def test_df_index_shape(native_df):
...
+            check_index_type=False,
+        )
+
+
+@sql_count_checker(query_count=4, join_count=1)
+@pytest.mark.parametrize("native_df", TEST_DFS)
+def test_df_index_columns_to_series(native_df):
+    snow_df = pd.DataFrame(native_df)
+
+    # Snowpark pandas sets the dtype of df.columns for an empty df as object and native pandas sets it as int,
```
sfc-gh-yzou, line 316: try to put index and name as parameter for the test point

src/snowflake/snowpark/modin/plugin/_internal/index.py
```diff
@@ -1510,7 +1505,7 @@ def item(self) -> None:
             If the data is not length = 1.
         """
         item = self._query_compiler.take_2d_positional(
-            index=slice(0, 2), columns=[0]
```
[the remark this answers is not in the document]
sfc-gh-samdesai1: What do you mean? this gets the first two elements of the index and if we only have one element, we know item() is valid so we return it. It removes the need to check length and only issues one query

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: they both work, even though the index never has columns

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: yes it automatically removes quotes and then causes errors

src/snowflake/snowpark/modin/plugin/_internal/index.py
sfc-gh-samdesai1: got it

tests/integ/modin/index/test_index_methods.py
sfc-gh-samdesai1: this is a bit of a challenge because the index parameter needs to have the correct length

sfc-gh-yzou (approved):

tests/integ/modin/index/test_index_methods.py
sfc-gh-yzou: i am also fine with one test point like this for now, if it is too much work to do the parameterization

sfc-gh-samdesai1: No, I can do the parametrization, I just was not sure how we wanted to do it

sfc-gh-vbudati (approved):

PR #1810, opened 2024-06-20
[SNOW-1491328] Support functions.py for AST/Snowpark IR.

sfc-gh-lspiegelberg:
[SNOW-1491328](https://snowflakecomputing.atlassian.net/browse/SNOW-1491328)

Support AST encoding for functions (up to table functions, UDFx) in `functions.py`.

Details:
- Fix `update_from_devvm.sh` script to invoke both bazel builds required.
- Add `check=True` to `test_ast_driver.py` to error out when subprocess fails.
- Removes cls method `._fill_ast` from `Column` class and refactors into free method within `ast_utils.py`.
- Introduces new `build_fn_apply` helper function to create `ApplyExpr` in IR.

[SNOW-1491328]: https://snowflakecomputing.atlassian.net/browse/SNOW-1491328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -133,6 +145,56 @@ def build_const_from_python_val(obj: Any, ast: proto.Expr) -> None:
...
+        else:
+            pos_arg = proto.Expr()
+            build_const_from_python_val(arg, pos_arg)
+            expr.pos_args.append(pos_arg)
+
+    for name, arg in kwargs.items():
+        kwarg = proto.Tuple_String_Expr()
+        kwarg._1 = name
+        if isinstance(arg, proto.Expr):
+            kwarg._2.CopyFrom(arg)
```
sfc-gh-oplaton, line 189: I'm curious about this line vs line 176.

At line 176, the code does `expr.pos_args.append(arg)` (uses `arg` as is).

Here, the code does `kwarg._2.CopyFrom(arg)`.

Is there a semantic difference between using the object reference as-is and cloning the object in depth? Would `CopyFrom` work equally well on line 176? (`expr.pos_args.append(Expr().CopyFrom(arg))` or some equivalent). Would `kwarg._2 = arg` work equally well here?

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -192,3 +254,139 @@ def set_src_position(ast: proto.SrcPosition) -> None:
...
+
+    Raises:
+        TypeError: An SpColumnExpr can only be populated from another SpColumnExpr or a valid Literal type
+    """
+    if isinstance(value, snowflake.snowpark.Column):
+        return ast.CopyFrom(value._ast)
+    elif isinstance(value, VALID_PYTHON_TYPES_FOR_LITERAL_VALUE):
+        build_const_from_python_val(value, ast)
+    elif isinstance(value, Expression):
+        pass  # TODO: clean this up
```
sfc-gh-oplaton, line 280: Would it make sense for this to be some kind of `...` or `raise NotImplementedError()`?

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -192,3 +254,139 @@ def set_src_position(ast: proto.SrcPosition) -> None:
...
+        assign_fields (Dict[str, Any], optional): Subtype fields with well known protobuf types that support direct assignment. Defaults to {}.
+        copy_messages (Dict[str, Any], optional): Subtype message fields which must be copied into (do not support assignment). Defaults to {}.
+        fill_expr_asts (Dict[str, ColumnOrLiteral], optional): Subtype Expr fields that must be filled explicitly from a ColumnOrLiteral type. Defaults to {}.
+
+    Returns:
+        proto.SpColumnExpr: Returns fully populated SpColumnExpr AST from given arguments
+    """
+
+    ast = proto.Expr()
+    if property is not None:
```
sfc-gh-oplaton, line 305: Style nit:

```python
if property is None:
    return ast
```

src/snowflake/snowpark/functions.py
```diff
@@ -567,7 +592,13 @@ def bround(col: ColumnOrName, scale: Union[Column, int]) -> Column:
     """
     col = _to_col_if_str(col, "bround")
     scale = _to_col_if_lit(scale, "bround")
-    return call_builtin("ROUND", col, scale, lit("HALF_TO_EVEN"))
+
+    ast = proto.Expr()
+    build_fn_apply(ast, "bround", col, scale)
+
+    col = call_builtin("ROUND", col, scale, lit("HALF_TO_EVEN"))
```
sfc-gh-oplaton, line 599: Nit: Capitalization.

I assume identifier canonicalization happens at the server, so there's not much point in being strict about this at the client. Is that reasonable?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: I think line 176 is preferred, however I couldn't get line 189 working with a similar different syntax. `kwarg._2 = arg` gives an error, that's why I used the `CopyFrom`

src/snowflake/snowpark/functions.py
sfc-gh-lspiegelberg: left it as in the original snowpark python code. Could be lowercased. For now, added a comment.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: Thank you for clarifying. I still have memories from working on systems that differentiated between using a subtree in multiple places, `ShallowClone` and `DeepClone` 😃

sfc-gh-oplaton (approved):

sfc-gh-azwiegincew (approved):

PR #1817, opened 2024-06-21
[SERVER-SIDE CI] Temporary fix for unparser fat jar

sfc-gh-lspiegelberg:
Fixes issue of bazel not producing a fat jar in `scala_binary` by calling both the bazel build and the sbt build. This is a temporary workaround till we fix in bazel the rules to emit a fat jar.


sfc-gh-vbudati (approved):
Thanks Leonhard!!

PR #1818, opened 2024-06-21
Support reading excel files in snowpandas. As seen in telemetry.

sfc-gh-jkew:


src/snowflake/snowpark/modin/core/execution/dispatching/factories/baseio.py
```diff
@@ -279,66 +279,8 @@ def read_clipboard(cls, sep=r"\s+", **kwargs):  # pragma: no cover # noqa: PR01
...
-            date_parser=date_parser,
-            thousands=thousands,
-            comment=comment,
-            skipfooter=skipfooter,
-            convert_float=convert_float,
-            mangle_dupe_cols=mangle_dupe_cols,
-            na_filter=na_filter,
-            **kwds,
-        )
+    def read_excel(cls, **kwargs):  # noqa: PR01
```
sfc-gh-dpetersohn, line 282: What was wrong with the old implementation here?

sfc-gh-jkew: Pandas would produce an error if it received an argument it didn't understand. I forget exactly which one it was here; but essentially there are two paths here:
1. Follow the read_csv approach and add all possible arguments for the current version of pandas ( and no more ). We need to do this for read_csv because we have both a server side and client side solution.
2. Just pass through kwargs into pandas and read the data frame.

In this case I went with (2) to keep the maintenance lighter; but I could be convinced of (1).

src/snowflake/snowpark/modin/core/execution/dispatching/factories/baseio.py
```diff
@@ -279,66 +279,8 @@ def read_clipboard(cls, sep=r"\s+", **kwargs):  # pragma: no cover # noqa: PR01
...
+        intermediate = pandas.read_excel(**kwargs)
```
sfc-gh-lspiegelberg, line 283: Should we capture the error in case openpyxl/reader is missing, and suggest to the user a package that is part of our sproc/conda universe?

docs/source/modin/supported/general_supported.rst
```diff
@@ -64,6 +64,12 @@ Data manipulations
 |                             |                                 |                                  | performance. You can force the use of the Snowflake|
 |                             |                                 |                                  | parser with ``engine=snowflake``                   |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
+| ``read_excel``              | D                               |                                  | Uses local pandas to read excel files, using the   |
```
sfc-gh-lspiegelberg, line 67: nit: I always got some push-back on using the phrase "local". There seems to be a preference at Snowflake to use "client" instead. If it is not too much to ask, maybe would recommend reformulating here a bit. I also think 

>> Please refer to the native pandas documentation for more details.

could be either 1.) replaced with a link or 2.) left out when we throw a good error message.

sfc-gh-lspiegelberg (approved):
lgtm, left some nits.

tests/utils.py
```diff
@@ -1287,6 +1287,10 @@ def test_file_csv_quotes_special(self):
     def test_file_csv_special_format(self):
         return os.path.join(self.resources_path, "testCSVspecialFormat.csv")
 
+    @property
+    def test_file_excel(self):
+        return os.path.join(self.resources_path, "test_excel.xlsx")
```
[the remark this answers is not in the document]
sfc-gh-jkew: It's not worth testing really. We would be testing native pandas, not our own code.

src/snowflake/snowpark/modin/core/execution/dispatching/factories/baseio.py
sfc-gh-jkew: Most of the engines that are supported are in our conda universe (other than the open document ones), but I think you are right, we could have a slightly better error messaging here than pandas itself.

tests/utils.py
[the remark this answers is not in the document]
sfc-gh-lspiegelberg: makes sense, maybe good to have a comment then. Basically stating let's leave correctness to pandas here. This test is simply to ensure coverage.

src/snowflake/snowpark/modin/core/execution/dispatching/factories/baseio.py
sfc-gh-lspiegelberg: That'd be awesome!

docs/source/modin/supported/general_supported.rst
```diff
@@ -64,6 +64,12 @@ Data manipulations
 |                             |                                 |                                  | performance. You can force the use of the Snowflake|
 |                             |                                 |                                  | parser with ``engine=snowflake``                   |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
+| ``read_excel``              | D                               |                                  | Uses local pandas to read excel files, using the   | 
+|                             |                                 |                                  | engine specified by the pandas. You will need to   |
+|                             |                                 |                                  | separately install a supported excel reader such   |
+|                             |                                 |                                  | openpyxl. Please refer to the native pandas        |
```
sfc-gh-helmeleegy, line 70:
```suggestion
|                             |                                 |                                  | as openpyxl. Please refer to the native pandas     | 
```

docs/source/modin/supported/general_supported.rst
sfc-gh-helmeleegy, line 67: The definition of `D` above this table says:
```
``D`` stands for defaults to single node pandas execution via UDF/Sproc
```
I think we can just say `Y` here instead - since all the parameters are supported.

sfc-gh-helmeleegy (approved):
LGTM, just a couple of nits. I'm also interested in learning if there is any advantage in passing an explicit list of arguments vs just kwargs. I cannot think of one.

sfc-gh-jrose (approved):

PR #1831, opened 2024-06-25
Fix server-side-snowpark linter errors [SNOW-1491327]

sfc-gh-oplaton:

   

   Prerequisite for SNOW-1491327

   Fix flake8 issues in AST-related modules.


tests/ast/test_ast_driver.py
```diff
@@ -23,19 +23,19 @@ class TestCase:
 
 def parse_file(file):
     """Parses a test case file."""
-    with open(file, "r", encoding="utf-8") as f:
+    with open(file, encoding="utf-8") as f:
```
sfc-gh-heshah, line 26: I assume its the tox precommit hooks which are removing the "r", Arthur pointed it out in previous PR for me. By default `open` does use the read mode, but I assume we want to be explicit.

So far I've been adding it back manually after running `tox fix-lint`, but this isn't a great solution. Perhaps we can remove it though if its common enough for the formatters to remove it by default?

sfc-gh-oplaton: I think we should configure our tools to do what we want them to. Fighting against the tools is a losing battle.

I prefer explicit and readable code, too, up to the point where some things become too verbose, or restate the obvious. I don't have a strong preference here.

sfc-gh-azwiegincew: I agree we should fix this, but I think it's rational to just leave a TODO for now, since we clearly have bigger fish to fry.

sfc-gh-azwiegincew (approved):

PR #1837, opened 2024-06-25
[SNOW-1502893]: Add support for `pd.crosstab`

sfc-gh-rdurrani:

   

   Fixes SNOW-1502893

Add support for pd.crosstab.


src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -1786,20 +1785,134 @@ def crosstab(
...
+        kwargs = {"aggfunc": aggfunc}
+
+    table = df.pivot_table(
+        "__dummy__",
+        index=unique_rownames,
+        columns=unique_colnames,
+        margins=margins,
+        margins_name=margins_name,
+        dropna=dropna,
+        # observed=dropna,
```
sfc-gh-helmeleegy, line 1839: remove

docs/source/modin/supported/general_supported.rst
```diff
@@ -18,7 +18,7 @@ Data manipulations
 | ``concat``                  | P                               | ``levels`` is not supported,     |                                                    |
 |                             |                                 | ``copy`` is ignored              |                                                    |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``crosstab``                | N                               |                                  |                                                    |
+| ``crosstab``                | Y                               |                                  |                                                    |
```
sfc-gh-helmeleegy, line 21: I think it should be `P` and inherit all the parameter limitations from `pivot_table`.

src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -1786,20 +1785,134 @@ def crosstab(
     """
     Compute a simple cross tabulation of two (or more) factors.
```
sfc-gh-helmeleegy, line 1786: Full doc string in the plugin please.

src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -1786,20 +1785,134 @@ def crosstab(
...
+    df = DataFrame(data, index=common_idx)
+
+    if values is None:
+        df["__dummy__"] = 0
+        kwargs = {"aggfunc": "count"}
+    else:
+        df["__dummy__"] = values
+        kwargs = {"aggfunc": aggfunc}
+
+    table = df.pivot_table(
```
sfc-gh-helmeleegy, line 1832: Can we also pass the calling method name (`crosstab` in this case) so that the NotImplementedError messages that would be raised inside `pivot_table` show the correct method name that was originally called by the user?

sfc-gh-helmeleegy: (For less confusion and easier debuggability)

sfc-gh-helmeleegy:
Tests?

docs/source/modin/supported/general_supported.rst
sfc-gh-rdurrani: In this case, the limitations from pivot_table do not apply, since we do not accept the unsupported arguments from pivot_table, and index must be specified, so margins is fully supported.

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-rdurrani: I"m not certain that that's possible? Unless we catch the not implemented error and edit it? There really shouldn't be any NotImplementedError though, since all of the arguments of crosstab are supported.

src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -1848,20 +1847,229 @@ def crosstab(
...
+
+                # Normalize core
+                f = normalizers[normalize]
+                table = f(table)
+                table = table.fillna(0).reindex(index=table_index)
+
+            elif normalize == "all":
+                # Normalize core
+                f = normalizers[normalize]
+                table = f(table.iloc[:, :-1]) * 2.0
```
sfc-gh-rdurrani, line 2059: This line can introduce numerical instability/rounding errors, but the alternative (which is what pandas does: f(table.iloc[:-1, :-1]) and then concat with index_margin) takes the counts up to:

* expected_query_count=7, actual_query_count=8
* expected_join_count=41, actual_join_count=177
* expected_union_count=22, actual_union_count=123

so I think that its probably ok? Would appreciate feedback!

tests/integ/modin/crosstab/test_crosstab.py
```diff
@@ -0,0 +1,767 @@
...
+                dropna=dropna,
+            )
+            assert_snowpark_pandas_equal_to_pandas(snow_df, native_df)
+
+    @pytest.mark.parametrize("normalize", [0, 1, True, "all", "index", "columns"])
+    def test_normalize(self, dropna, normalize):
+        query_count = 4 if normalize not in (0, "index") else 3
+        join_count = 0 if normalize not in (0, "index") else 3
+        if not dropna:
+            join_count = 9 if normalize in (0, "index") else 4
```
sfc-gh-mvashishtha, line 538:
```suggestion
        query_count = 3 if normalize in (0, "index") else 4
        if dropna:
	    join_count = 3 if normalize in (0, "index") else 0
        else:
            join_count = 9 if normalize in (0, "index") else 4
```

nit: reduce `not`s and only assign `join_count` once.

tests/integ/modin/crosstab/test_crosstab.py
```diff
@@ -0,0 +1,767 @@
...
+                "shiny",
+                "dull",
+                "shiny",
+                "shiny",
+                "shiny",
+            ],
+            dtype=object,
+        )
+        if normalize is True:
+            sql_counts = counts["all"]
```
sfc-gh-mvashishtha, line 664: nit: can we do `counts[True] = counts['all']` at the beginning of the function instead?

tests/integ/modin/crosstab/test_crosstab.py
```diff
@@ -0,0 +1,767 @@
...
+            snow_df = pd.crosstab(
+                a,
+                [b, c],
+                rownames=["a"],
+                colnames=["b", "c"],
+                normalize=normalize,
+                margins=True,
+                dropna=dropna,
+            )
+            assert_snowpark_pandas_equal_to_pandas(snow_df, native_df)
```
sfc-gh-mvashishtha, line 690:
```suggestion
            eval_snowpark_pandas_result(
                pd,
                native_pd,
                lambda module: module.crosstab(
                    a,
                    [b, c],
                    rownames=["a"],
                    colnames=["b", "c"],
                    normalize=normalize,
                    margins=True,
                    dropna=dropna,
                )
            )
```

tests/integ/modin/crosstab/test_crosstab.py
```diff
@@ -0,0 +1,767 @@
...
+                    "beef",
+                    "fish",
+                    "beef",
+                ],
+                "age": [7, 2, 8, 5, 9, 3, 6, 1],
+            }
+        )
+
+        with SqlCounter(query_count=query_count, join_count=join_count):
+            native_df_result = native_pd.crosstab(
```
sfc-gh-mvashishtha, line 714: again, can use `eval_snowpark_pandas_result(pd, native_pd)`

sfc-gh-mvashishtha (commented):
Leaving minor comments on just the test file.

src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -801,7 +801,6 @@ def aggregate(
         # TypeError: got an unexpected keyword argument 'skipna'
         if is_dict_like(func) and not uses_named_kwargs:
             kwargs.clear()
-
```
sfc-gh-mvashishtha, line 804: nit: revert

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-mvashishtha: There can be unsupported aggregation functions, like `median` (we should test that, by the way). I think that it's ok if the error message mentions `pivot_table`. I think the user will be able to understand that error message, and our telemetry will tell us that the original function was `crosstab`.

sfc-gh-mvashishtha (commented):
I've reviewed all the files

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-helmeleegy: I don't think it will be clear to the user, especially if it's part of a chain of operations. The chain may even include an actual `pivot_table` operation that is well supported. The error message in this case will be very confusing.

sfc-gh-mvashishtha: This is a potential problem with many cases where Snowpark pandas methods call other Snowpark pandas methods. In my opinion it's not worth fixing.

sfc-gh-rdurrani: > There can be unsupported aggregation functions, like median (we should test that, by the way). I think that it's ok if the error message mentions pivot_table. I think the user will be able to understand that error message, and our telemetry will tell us that the original function was crosstab.

Good catch - completely slipped my mind! Will add a test for unsupported aggs!

tests/integ/modin/crosstab/test_crosstab.py
sfc-gh-rdurrani: I did that earlier, but it was leading to the wrong value being fetched out of the dictionary - I suspect it has something to do with setting `True` as a key, and having other truthy values as keys as well, but I'm not certain!

src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -1994,21 +1992,310 @@ def crosstab(
...
+        # of indices.)
+        df = dfs[0]
+        for right in dfs[1:]:
+            df = df.merge(right, left_index=True, right_index=True)
+        if len(arrays) > 0:
+            index = df.index
+            right_df = pd.concat(arrays, axis=1)
+            # Increases query count by 1, but necessary for error checking.
+            index_length = len(df)
+            if index_length != array_lengths[0]:
```
sfc-gh-rdurrani, line 2188: This additional query is caused by a change in our set_index behavior - do folks think this is ok for error checking purposes?

src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -1994,21 +1992,310 @@ def crosstab(
...
+    Returns
+    -------
+    DataFrame
+        Cross tabulation of the data.
+
+    Notes
+    -----
+
+    Raises NotImplementedError if aggfunc is not one of "count", "mean", "min", "max", or "sum", or
+    margins is True, normalize is True or all, and values is passed.
```
sfc-gh-rdurrani, line 2038: I'm not supporting this right now, because I don't think that this is a crucial customer use case, but have created a follow up ticket here: https://snowflakecomputing.atlassian.net/browse/SNOW-1642451!

sfc-gh-rdurrani: To clarify, I don't think this is worth blocking the PR, but am happy to implement if folks think it is worth it!

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-mvashishtha: I think it's okay to leave this for now.

sfc-gh-mvashishtha (approved):

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-helmeleegy: You're right. We indeed frequently have Snowpark pandas methods call other Snowpark pandas methods. We did take care of passing the calling method name as an argument in some cases to customize the error messages accordingly. However, I'm fine with not worrying about this for the purpose of this PR, and possibly handling all similar cases in the code base in one pass at some point.

sfc-gh-helmeleegy (approved):
LGTM. Thanks, Rehan!

tests/integ/modin/crosstab/test_crosstab.py
```diff
@@ -844,6 +844,172 @@ def eval_func(lib):
                 eval_func,
             )
 
+    @pytest.mark.parametrize("aggfunc", ["count", "mean", "min", "max", "sum"])
+    def test_margins_and_values(self, dropna, aggfunc):
+        a = np.array(
+            [
+                "foo",
```
sfc-gh-mvashishtha, line 851: I think we are using this array ~10 times in this file. Can we make `a` a fixture (note we prefer fixtures since we don't want test cases to mutate them)?

Also can we reduce duplication for other highly reused inputs?

sfc-gh-helmeleegy:
> Tests?

This is already covered.

sfc-gh-mvashishtha (approved):

sfc-gh-mvashishtha (approved):

sfc-gh-mvashishtha (approved):

PR #1838, opened 2024-06-25
Fix Index typehints and move class

sfc-gh-samdesai1:

   

   Fixes SNOW-1481037

   Fix some of the typehints in the index class to correctly identify return and input types


sfc-gh-vbudati (approved):
Thanks Samedh!

sfc-gh-yzou (approved):
@sfc-gh-samdesai1 please try to generate the api doc and see if the type are set correctly

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -42,8 +46,7 @@
 class Index:
     def __init__(
         self,
-        # TODO: SNOW-1481037 : Fix typehints for index constructor, set_query_compiler and set_local_index
-        data: ArrayLike | Any = None,
+        data: ArrayLike | SnowflakeQueryCompiler = None,
```
sfc-gh-joshi, line 49:
```suggestion
        data: ArrayLike | SnowflakeQueryCompiler | None = None,
```
You should add the extra `None` annotation to make clear this is optional.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -109,8 +112,7 @@ def __init__(
 
     def set_query_compiler(
         self,
-        # TODO: SNOW-1481037 : Fix typehints for index constructor, set_query_compiler and set_local_index
-        data: ArrayLike | Any = None,
+        data: ArrayLike | SnowflakeQueryCompiler = None,
```
sfc-gh-joshi, line 115:
```suggestion
        data: ArrayLike | SnowflakeQueryCompiler | None = None,
```

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -140,8 +137,7 @@ def set_query_compiler(
 
     def set_local_index(
         self,
-        # TODO: SNOW-1481037 : Fix typehints for index constructor, set_query_compiler and set_local_index
-        data: ArrayLike | Any = None,
+        data: ArrayLike | SnowflakeQueryCompiler = None,
```
sfc-gh-joshi, line 140:
```suggestion
        data: ArrayLike | SnowflakeQueryCompiler | None = None,
```

sfc-gh-joshi (changes requested):
A couple of nits regarding optional types. A couple more things:

Can you fix the type annotations on these three methods?
- `is_lazy_check`: just making it go `Callable -> Callable` is fine, as long as it's clear that this is an annotation. This also should be moved out of the class to the top level of the file, since it's neither an instance nor class method (first argument isn't `self`/`cls`).
- `duplicated` (currently returns Any)
- `str` (return type should be `native_pd.core.strings.accessor.StringMethods` rather than `str`)

I also noticed there's a function that uses `Self` in the return type; we should explicitly use `Index` instead. If you want to use `Self`, you should replace all the other type hint references to `Index` as well, and change the import to come from the native `typing` module rather than `pandas._typing`.

sfc-gh-joshi (approved):

PR #1821, opened 2024-06-24
SNOW-1445832: Added support for `DataFrame.stack`

sfc-gh-nkrishna:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-1445832

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.

3. Please describe how your code solves the related issue.

   This PR adds support for `DataFrame.stack`

PR #1848, opened 2024-06-27
[SNOW-1445834, SNOW-1445726]: Implemented `DataFrame.unstack` and `Series.unstack`

sfc-gh-nkrishna:

   

   Fixes SNOW-1445834, SNOW-1445726

   This PR implements `DataFrame.unstack` and `Series.unstack`.


src/snowflake/snowpark/modin/pandas/dataframe.py
```diff
@@ -1794,23 +1794,31 @@ def nsmallest(self, n, columns, keep="first"):  # noqa: PR01, RT01, D200
...
         """
         Pivot a level of the (necessarily hierarchical) index labels.
         """
         # TODO: SNOW-1063346: Modin upgrade - modin.pandas.DataFrame functions
-        if not isinstance(self.index, pandas.MultiIndex) or (
-            isinstance(self.index, pandas.MultiIndex)
-            and is_list_like(level)
-            and len(level) == self.index.nlevels
+        # This ensures that non-pandas MultiIndex objects are caught.
+        is_multiindex = len(self.index.names) > 1
```
sfc-gh-nkrishna, line 1808: This check in Modin results in a query being issued to pull the index -- I wonder if we deviate from Modin temporarily and push this down to the query compiler as we have to do another index check there (and we can reduce to 2 queries in total).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15028,3 +15028,103 @@ def stack(
...
+                .set_index_names([None])
+            )
+
+            from snowflake.snowpark.modin.pandas.dataframe import DataFrame
+
+            qc = (
+                DataFrame(query_compiler=qc)
+                .rename_axis([None] * len(qc.columns), axis=1)
+                ._query_compiler
+            )
```
sfc-gh-nkrishna, line 15104: Seems like the `rename_axis` implementation is handled by the frontend layer and not pushed down to the query compiler -- couldn't find a better way of handling this. Wonder if folks have other ideas -- I can add a comment if that's helpful.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15028,3 +15028,103 @@ def stack(
...
+            qc = (
+                self.reset_index()
+                .melt(
+                    id_vars=index_names,
+                    value_vars=self.columns,
+                    var_name="index_second_level",
+                    value_name=MODIN_UNNAMED_SERIES_LABEL,
+                    ignore_index=False,
+                )
+                .replace(to_replace=UNPIVOT_NULL_REPLACE_VALUE, value=np.nan)
```
sfc-gh-jkew, line 15123: I would think unpivot_utils would handle swapping back the UNPIVOT_NULL_REPLACE_VALUE. We generally want to keep references to UNPIVOT_NULL_REPLACE_VALUE inside of unpivot_utils because it should be replaced once the sql parameter is set to be the default.

src/snowflake/snowpark/modin/pandas/dataframe.py
devin-petersohn: There is a query compiler method supported in OSS Modin that is `self._query_compiler.has_multiindex` that you could use here. Example in Modin: 

https://github.com/modin-project/modin/blob/759d548814a6ac224e83e7531cf98e20b13d85cb/modin/pandas/dataframe.py#L3227

src/snowflake/snowpark/modin/pandas/series.py
```diff
@@ -1535,19 +1535,35 @@ def set_axis(
...
-        )
+        # We can't unstack a Series object, if we don't have a MultiIndex.
+        if len(self.index.names) > 1:
+            result = DataFrame(
+                query_compiler=self._query_compiler.unstack(
+                    level, fill_value, sort, is_series_output=False
+                )
+            )
+        else:
+            raise ValueError(  # pragma: no cover
```
sfc-gh-yzou, line 1558: I think we can simply push all implementation details to backend, the multiIndex check with error, and also the drop level implementation etc. That is what we evetually want to get at the end also, where frontend only responsible for basic input parameter check, simple parameter reduction, and final result format (dataframe, series, inplace) handlization

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: Typically, once we come to backend, we try not to go back to the frontend object. Here seems we just want to reset the column index name field to all None, you can simply copy the internal frame with data_column_index_names field set to all None at this moment, and we should implement rename_axis correctly with Snowpark pandas

tests/integ/modin/frame/test_unstack.py
```diff
@@ -0,0 +1,49 @@
...
+import pandas as native_pd
+import pytest
+
+from tests.integ.modin.sql_counter import sql_count_checker
+from tests.integ.modin.utils import eval_snowpark_pandas_result
+
+
+@sql_count_checker(query_count=1)
+def test_unstack_input_no_multiindex():
+    # Only 1 Query as input dataframe to unstack is not a multi-index,
```
sfc-gh-yzou, line 16: I see a lot of our tests are trying to reset the names of index, do we have any test that originally have a name configured for the name fields?

tests/integ/modin/frame/test_unstack.py
```diff
@@ -0,0 +1,49 @@
...
+    # so calls df.reset_index().melt()
+    index = native_pd.MultiIndex.from_tuples(
+        [("one", "a"), ("one", "b"), ("two", "a"), ("two", "b")]
+    )
+    # Note we call unstack below to create a dataframe without a multiindex before
+    # calling unstack again
+    native_df = native_pd.Series(np.arange(1.0, 5.0), index=index).unstack(level=0)
```
sfc-gh-yzou, line 23: since unpviot have issue with missing values, do we have any test that have missing value there?

tests/integ/modin/frame/test_unstack.py
```diff
@@ -0,0 +1,49 @@
...
+    snow_df = pd.DataFrame(native_df)
+    eval_snowpark_pandas_result(snow_df, native_df, lambda df: df.unstack())
```
sfc-gh-yzou, line 25: Let's also add a test whose df have more than one column, any may be a column with a different data type like string

tests/integ/modin/frame/test_unstack.py
sfc-gh-yzou, line 23: the level parameter also takes string as input, which is probably the level name, let's add test  for that also

sfc-gh-nkrishna: Yes we test unpivot with missing values as part of the `df.stack()` PR: https://github.com/snowflakedb/snowpark-python/pull/1821/files. I don't think it's worth adding a duplicate of that test here since it's a very similar implementation with calling melt then unpivot (just without the `sort_rows_by_column_values`).

tests/integ/modin/frame/test_unstack.py
sfc-gh-nkrishna: Yep good idea.

src/snowflake/snowpark/modin/pandas/series.py
sfc-gh-nkrishna: I think we should leave the frontend checks as-is since they are raising errors that pandas raises as well, so no need to go to the query compiler. It will continue to be as consistent as possible with open-source Modin -- I got the query counts down by using `self._query_compiler.has_multiindex`

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: This is an artifact of using `melt` -- we add tests for this in the `stack` PR which leverages `melt` here: https://github.com/snowflakedb/snowpark-python/pull/1821/files

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Given we are interacting with the query compiler the whole way through, creating an InternalFrame that is a copy of all the attributes since it's immutable and just setting `data_column_index_names` to [None] and recreating the query compiler seems like a less clean implementation to me. I added a comment around what this is doing.

tests/integ/modin/frame/test_unstack.py
sfc-gh-nkrishna: It is noted in the docs that non-integer support for level is not available yet, I also added a NotImplementedError and a test

src/snowflake/snowpark/modin/pandas/dataframe.py
```diff
@@ -1796,23 +1796,33 @@ def nsmallest(self, n, columns, keep="first"):  # noqa: PR01, RT01, D200
...
         """
         Pivot a level of the (necessarily hierarchical) index labels.
         """
         # TODO: SNOW-1063346: Modin upgrade - modin.pandas.DataFrame functions
-        if not isinstance(self.index, pandas.MultiIndex) or (
-            isinstance(self.index, pandas.MultiIndex)
-            and is_list_like(level)
-            and len(level) == self.index.nlevels
+        # This ensures that non-pandas MultiIndex objects are caught.
+        num_index_levels = self._query_compiler._modin_frame.num_index_levels()
```
devin-petersohn, line 1810: I'm not sure this is safe, since we don't force a `_modin_frame` on the `_query_compiler`

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
                 "Snowpark pandas doesn't support multiindex columns in stack API"
             )
 
-        index_names = ["index"]
+        index_names = self.get_index_names()
```
sfc-gh-nkrishna, line 15004: This code change is very similar to unstack with the non multiindex case and fixes a bug with setting index name. There is an associated test added as well

src/snowflake/snowpark/modin/pandas/dataframe.py
sfc-gh-nkrishna: Ah okay let me change this back then, thanks!

devin-petersohn: I recommend using and implementing `self._query_compiler.has_multiindex()`

sfc-gh-nkrishna: Yep changed to use has_multiindex

devin-petersohn (approved):
LGTM

sfc-gh-nkrishna:
@devin-petersohn I think you used your personal account to approve

sfc-gh-dpetersohn (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
...
+            ]
+            vals = [
+                c for c in self.columns if c not in index_cols and c not in pivot_cols
+            ]
+            qc = self.reset_index().pivot_table(
+                columns=pivot_cols,
+                index=index_cols,
+                values=vals,
+                aggfunc="min",
+                fill_value=None,
```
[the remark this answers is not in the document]
sfc-gh-yzou: Also, do we have any test for the fill_value? I did a quick glance of the tests, but i don't think I found one

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
...
+                )
+        else:
+            # N.B. normally non-MultiIndex cases would throw an error in pandas
+            # for series objects. However, pandas seems to handle this for DFs
+            # by doing a transposed stack! Strange behavior, but since we throw
+            # an error in the modin API layer for Series without MultiIndex indices,
+            # we can assume that this must be a DataFrame. This assumption might
+            # change in the future, so this code should be changed along with it.
+            if has_index_names_none:
+                # Index name defaults to "index" after reset_index() operation if index name is None
```
sfc-gh-yzou, line 15169: is the behavior of unstack is the same as stack() when it is single index for dataframe? if yes, can we just call stack here directly?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
...
+                margins_name="All",
+                observed=False,
+                sort=sort,
+            )
+
+            # Set the correct index names based on if index name is None
+            output_index_names = qc.get_index_names()
+            output_index_names_replace_level_with_none = [
+                None
+                if isinstance(output_index_names[i], str)
```
sfc-gh-yzou, line 15137: here it seems because the None columns is set to other names, and here we are trying to reset it back. However, those kind of name based operation could be error prone, since level_* is a pretty common used default name in pandas, and the dataframe feed into this operation could be a dataframe that contains those labels already.

I think we can keep the original index names and columns names, and derive the final name  based on the original name, probably with positions, order is something will be maintained during the operations.

Also, we can simply recreate the internal frame here to update the names, which seems much  simplier to do here

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
...
+                .melt(
+                    id_vars=index_cols,  # type: ignore
+                    value_vars=self.columns,
+                    var_name=col_label,
+                    value_name=MODIN_UNNAMED_SERIES_LABEL,
+                    ignore_index=False,
+                )
+                .replace(to_replace=UNPIVOT_NULL_REPLACE_VALUE, value=np.nan)
+                .set_index_from_columns([col_label] + index_cols)  # type: ignore
+            )
```
sfc-gh-yzou, line 15188: also, i tried the single index unstack with some np.nan value inside, df.unstack(fill_value=0.5) seems not filling the values in pandas

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
...
+            qc = (
+                self.reset_index()
+                .melt(
+                    id_vars=index_cols,  # type: ignore
+                    value_vars=self.columns,
+                    var_name=col_label,
+                    value_name=MODIN_UNNAMED_SERIES_LABEL,
+                    ignore_index=False,
+                )
+                .replace(to_replace=UNPIVOT_NULL_REPLACE_VALUE, value=np.nan)
```
sfc-gh-yzou, line 15186: here i think i still have the question about why we need to do such replace, can you give me some more details about this, if it is a bug, we should have ticket to fix it

sfc-gh-nkrishna: Repro from what I am seeing:

```
native_df = native_pd.DataFrame(data=[[0, np.nan], [np.nan, 3]], index=["cat", "dog"], columns=["weight", "height"])
snow_df = pd.DataFrame(native_df)
snow_df.reset_index().melt(id_vars="index", value_vars=['weight', 'height'], var_name="added_column", value_name="__reduced__", ignore_index=False)
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: No there are some similarities, but we don't use `sort_rows_by_column_values` and the index behavior is different.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
...
             )
             .sort_rows_by_column_values(
-                columns=index_names,  # type: ignore
+                columns=index_cols,  # type: ignore
                 ascending=[True],
                 kind="stable",
                 na_position="last",
                 ignore_index=False,
             )
             .replace(to_replace=UNPIVOT_NULL_REPLACE_VALUE, value=np.nan)
```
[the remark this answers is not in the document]
sfc-gh-nkrishna: Responded in another comment, resolving this one.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
...
+                # Index name defaults to "level_{i}" after reset_index() operation
+                # if index name is None
+                index_cols = [
+                    f"level_{i}" if index_names[i] is None else index_names[i]
+                    for i in range(len(index_names))
+                ]
+            else:
+                index_cols = index_names  # type: ignore
+
+            pivot_cols = [index_cols[lev] for lev in level]  # type: ignore
```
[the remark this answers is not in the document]
sfc-gh-nkrishna: This still works for level -1 (see the tests), it will simply grab the last column in index_cols. For a two-level MultiIndex, -1 has the same behavior as level 1.

src/snowflake/snowpark/modin/pandas/dataframe.py
```diff
@@ -1796,23 +1796,32 @@ def nsmallest(self, n, columns, keep="first"):  # noqa: PR01, RT01, D200
...
         # TODO: SNOW-1063346: Modin upgrade - modin.pandas.DataFrame functions
-        if not isinstance(self.index, pandas.MultiIndex) or (
-            isinstance(self.index, pandas.MultiIndex)
-            and is_list_like(level)
-            and len(level) == self.index.nlevels
+        # This ensures that non-pandas MultiIndex objects are caught.
+        is_multiindex = self._query_compiler.is_multiindex()
+
+        if not is_multiindex or (
+            is_multiindex and is_list_like(level) and len(level) == self.index.nlevels
```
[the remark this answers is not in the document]
sfc-gh-nkrishna: Good point, done.

src/snowflake/snowpark/modin/pandas/series.py
sfc-gh-nkrishna: The case where we use `droplevel` case is only if the input is a Series, so how do we push it to the query compiler layer?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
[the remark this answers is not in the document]
sfc-gh-nkrishna: If you pass `fill_value` here, the `pivot_table` implementation still just calls `fillna` internally. I think it's better to keep the `fillna` to handle both the implementations (if you `pivot`, else you use `melt`).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15001,35 +15001,204 @@ def stack(
...
+        if has_index_names_none:
+            # Index name defaults to "index" after reset_index() operation if index name is None
+            index_cols = ["index"]
+        else:
+            index_cols = index_names  # type: ignore
+
         # Stack is equivalent to doing df.melt() with index reset, sorting the values, then setting the index
         # Note that we always use sort_rows_by_column_values even if sort is False
+        col_label = (
+            "index_second_level" if self.columns.name is None else self.columns.name
```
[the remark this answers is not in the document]
sfc-gh-nkrishna: Yes exactly, this gets fed to `melt` via the `var_name` parameter so we control this name.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Filed JIRA with the above example to repro and added a comment in the code.

src/snowflake/snowpark/modin/pandas/series.py
sfc-gh-nkrishna: Resolved offline, passing `is_series_input` to the QC layer.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15066,35 +15066,200 @@ def stack(
...
+            Replace NaN with this value if the unstack produces missing values.
+
+        sort : bool, default True
+            Sort the level(s) in the resulting MultiIndex columns.
+
+        is_series_input : bool, default False
+            Whether the input is a Series, in which case we call `droplevel`
+        """
+        if not isinstance(level, int):
+            ErrorMessage.not_implemented(
```
sfc-gh-yzou, line 15153: do we have a ticket to support level as a string or list? especially with string, seems something could be used with multiIndex commonly, and we do handle len(self.index_names)> 1. 
we have function parse_levels_to_integer_levels and get_pandas_labels_for_levels and get_snowflake_identifiers_for_levels to help handle the level parameters in general

sfc-gh-nkrishna: There is a ticket I filed for it -- https://snowflakecomputing.atlassian.net/browse/SNOW-1558364. I added a TODO here and will do this as a follow-up to this PR. I'll try to support multiindex column names in the same follow-up PR as well.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15066,35 +15066,200 @@ def stack(
...
+        # Check to see if we have a MultiIndex, if we do, make sure we remove
+        # the appropriate level(s), and we pivot accordingly.
+        if len(index_names) > 1:
+            index_col_names_after_reset_index = [
+                f"level_{i}" if index_names[i] is None else index_names[i]
+                for i in range(len(index_names))
+            ]
+            index_cols = index_col_names_after_reset_index
+
+            pivot_cols = [index_cols[lev] for lev in level]  # type: ignore
```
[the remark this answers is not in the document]
sfc-gh-nkrishna: For some reason it still complains even when I added a type here I think b/c of how the indexing into index_cols works.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15799,34 +15801,208 @@ def stack(
...
+            output_index_names_replace_level_with_none = [
+                None
+                if output_index_names[i] in column_names_to_reset_to_none
+                else output_index_names[i]
+                for i in range(len(output_index_names))
+            ]
+            qc = qc.set_index_names(output_index_names_replace_level_with_none)
+            # Set the unnamed column values back to None
+            output_column_names = qc.columns.names
+            output_column_names_replace_level_with_none = [
```
sfc-gh-yzou, line 15912: is the reset of column needed, it seems that the rest of the original index column is used as the pivot columns, and the result columns name are values of from the pivot column, the original column names shouldn't be in any of the result column of pivot

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15799,34 +15801,208 @@ def stack(
...
+
+        index_names = self.get_index_names()
+
+        # Check to see if we have a MultiIndex, if we do, make sure we remove
+        # the appropriate level(s), and we pivot accordingly.
+        if len(index_names) > 1:
+            # Resetting the index keeps the index columns as the first n data columns
+            qc = self.reset_index()
+            index_cols = qc._modin_frame.data_column_pandas_labels[0 : len(index_names)]
+            pivot_cols = [index_cols[lev] for lev in level]  # type: ignore
```
sfc-gh-yzou, line 15867: thanks for updating this! I think we can make this part even simpiler, it seems that you want to separate the reseted index columns into pivot columns and index columns, and you want to record the index column whose original label that is 
you propbably just do it in one look
```
for i in range(len(index_cols)):
     if i in level:
           pivot_cols.append(index_cols(i))
     else:
            res_index_cols.appen(index_cols(i))
            res_index_none.append(index_names(i) is None)
```

sfc-gh-yzou: then res_index_none can just be directly used to check which index column we want to set it back to None

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15799,34 +15801,208 @@ def stack(
...
+        qc = qc.melt(
+            id_vars=index_cols,  # type: ignore
+            value_vars=self.columns,
+            var_name=col_label,
+            value_name=MODIN_UNNAMED_SERIES_LABEL,
+            ignore_index=False,
+        )
+
+        if operation == StackOperation.STACK:
+            # Only sort rows by column values in case of 'stack'
```
sfc-gh-yzou, line 15979: so if it is unstack, what is the order her? can you left a comment here?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15799,34 +15801,208 @@ def stack(
...
+        if operation == StackOperation.STACK:
+            qc = qc.set_index_from_columns(index_cols + [col_label])  # type: ignore
         else:
-            return qc
+            qc = qc.set_index_from_columns([col_label] + index_cols)  # type: ignore
+
+        # Set the original unnamed index and column values back to None
+        output_index_names = qc.get_index_names()
+        output_index_names = [
+            None
```
sfc-gh-yzou, line 15999: i felt you don't even need to record none here, it seems that you just want to reset the columns back to its orignal name, in other words,
orignial_col_label = None if self.columns.name is None else self.columns.name
so if operation == StackOperation.STACK:
        index_names = original_index_cols + [original_col_label]
    else:
       index_names = [origianl_col_label] + original_index_columns

is that what we want?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15799,34 +15801,208 @@ def stack(
...
+
+        if is_series_input and qc.columns.nlevels > 1:
+            # If input is Series and output is MultiIndex, drop the top level of the MultiIndex
+            qc = qc.set_columns(qc.columns.droplevel())
+        return qc
+
+    def _stack_helper(
+        self,
+        index_names: list[Hashable],
+        operation: StackOperation,
```
sfc-gh-yzou, line 15934: do we need to pass index_names here? isn't it just self.get_index_names().? if that is the case, let's direct get the index names, otherwise, we can make mistake by passing wrong index names in here

tests/integ/modin/frame/test_unstack.py
```diff
@@ -0,0 +1,89 @@
...
+import modin.pandas as pd
+import numpy as np
+import pandas as native_pd
+import pytest
+
+from tests.integ.modin.sql_counter import sql_count_checker
+from tests.integ.modin.utils import eval_snowpark_pandas_result
+
+
+@pytest.mark.parametrize(
```
sfc-gh-yzou, line 14: add a test whose index is more than two levels, so we cover the case when the result is multi-level

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Added a comment saying it's just the row position order

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Refactored to use only one loop

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: We still need this code as some of the names we are replacing with None could end up in the column names, see the case below:
```
    index = pd.MultiIndex.from_product(
        iterables=[[2, 1], ["a", "b"]], names=["hello", None]
    )
    native_df = native_pd.DataFrame(np.random.randn(4), index=index, columns=["A"])
    snow_df = pd.DataFrame(native_df)
    snow_df.unstack()
```

```
snow =                 A          
level_1         a         b
hello                      
1       -0.407852  0.038916
2        2.405507  0.374415
expected_pandas =               A          
              a         b
hello                    
1     -0.407852  0.038916
2      2.405507  0.374415
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: It's not just the column name, we still need to track the index names that correspond to None and reset those as well.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Done

tests/integ/modin/frame/test_unstack.py
sfc-gh-nkrishna: `test_unstack_multiindex` in the `frame` folder does result in a MultiIndex output already.

sfc-gh-yzou (approved):

PR #1858, opened 2024-06-29
[SNOW-1482588] Eliminate cast expression for numeric value

sfc-gh-yzou:

   

   Fixes SNOW-1482588

Today, we generate an extra cast expression for all literal expression to make sure the expression generated is casted to an expected type. However, such cast introduces following problems:
1) In general expression, the final dtype should be decided by snowflake, and user should cast the value explicitly if needed
2) extra cast expression increases expression depth which causes extra overhead in both compilation and execution.  In our experiment with IN expression, the difference can be dramatic.

In this pr, the following changes are applied:
1) For binary expression, multiExpression (used in column in), InExpression, avoid generating sql value with casting for numerical types. For example: 'A' > 1 :: INT will become 'A' > 1, and 'A' in (1 :: INT, 2 :: INT) will be come 'A' in (1, 2). We only do that for expressions, so for column select with value expression, they still follows our python to snowflake mapping.
2) rename to_sql_avoid_offset to to_sql_try_avoid cast to be more clear. 
3) added parameter _eliminate_numeric_sql_value_cast_enabled to control the change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Added distributed tracing using open telemetry APIs for table stored procedure function in `DataFrame`:
   - _execute_and_get_query_id
 - Allow `df.plot()` and `series.plot()` to be called, materializing the data into the local client
-- Added session parameter _eliminate_numeric_sql_value_cast_enabled to allow generating sql value without casting for numeric values when not needed. The value is default False for now.
+- Avoid casting for numeric sql values for binary expression, multiExpression and InExpression. The change is controlled by session parameter _eliminate_numeric_sql_value_cast_enabled.
 
 ### Snowpark pandas API Updates
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Added distributed tracing using open telemetry APIs for table stored procedure function in `DataFrame`:
   - _execute_and_get_query_id
 - Allow `df.plot()` and `series.plot()` to be called, materializing the data into the local client
-- Avoid casting for numeric sql values for binary expression, multiExpression and InExpression. This is a performance improvement especially for InExpression, and can be enabled through session.eliminate_numeric_sql_value_cast_enabled = True.
+- Improves performance for binary column expression and df._in by avoiding unnecessary cast for numeric values. This optimization can be enabled through session.eliminate_numeric_sql_value_cast_enabled = True.
 
 ### Snowpark pandas API Updates
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Added distributed tracing using open telemetry APIs for table stored procedure function in `DataFrame`:
   - _execute_and_get_query_id
 - Allow `df.plot()` and `series.plot()` to be called, materializing the data into the local client
+- Added session parameter _eliminate_numeric_sql_value_cast_enabled to allow generating sql value without casting for numeric values when not needed. The value is default False for now.
 
 ### Snowpark pandas API Updates
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Added distributed tracing using open telemetry APIs for table stored procedure function in `DataFrame`:
   - _execute_and_get_query_id
 - Allow `df.plot()` and `series.plot()` to be called, materializing the data into the local client
-- Avoid casting for numeric sql values for binary expression, multiExpression and InExpression. The change is controlled by session parameter _eliminate_numeric_sql_value_cast_enabled.
+- Avoid casting for numeric sql values for binary expression, multiExpression and InExpression. This is a performance improvement especially for InExpression, and can be enabled through session.eliminate_numeric_sql_value_cast_enabled = True.
 
 ### Snowpark pandas API Updates
 
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -337,7 +337,11 @@ class Analyzer:
             return expr.sql
 
         if isinstance(expr, Literal):
-            sql = to_sql(expr.value, expr.datatype, eliminate_numeric_sql_value_cast_enabled=self.session._eliminate_numeric_sql_value_cast_enabled)
+            sql = to_sql(
+                expr.value,
+                expr.datatype,
+                eliminate_numeric_sql_value_cast_enabled=self.session._eliminate_numeric_sql_value_cast_enabled,
+            )
             if parse_local_name:
                 sql = sql.upper()
             return sql
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -337,7 +337,7 @@ class Analyzer:
             return expr.sql
 
         if isinstance(expr, Literal):
-            sql = to_sql(expr.value, expr.datatype)
+            sql = to_sql(expr.value, expr.datatype, eliminate_numeric_sql_value_cast_enabled=self.session._eliminate_numeric_sql_value_cast_enabled)
             if parse_local_name:
                 sql = sql.upper()
             return sql
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -48,9 +48,9 @@ from snowflake.snowpark._internal.analyzer.binary_expression import (
 )
 from snowflake.snowpark._internal.analyzer.binary_plan_node import Join, SetOperation
 from snowflake.snowpark._internal.analyzer.datatype_mapper import (
+    numeric_to_sql_without_cast,
     str_to_sql,
     to_sql,
-    to_sql_without_cast,
 )
 from snowflake.snowpark._internal.analyzer.expression import (
     Attribute,
@@ -255,30 +255,46 @@ class Analyzer:
             )
 
         if isinstance(expr, MultipleExpression):
-            return block_expression(
-                [
-                    self.analyze(
+            block_expressions = []
+            for expression in expr.expressions:
+                if self.session.eliminate_numeric_sql_value_cast_enabled:
+                    resolved_expr = self.to_sql_try_avoid_cast(
                         expression,
                         df_aliased_col_name_to_real_col_name,
                         parse_local_name,
                     )
-                    for expression in expr.expressions
-                ]
-            )
+                else:
+                    resolved_expr = self.analyze(
+                        expression,
+                        df_aliased_col_name_to_real_col_name,
+                        parse_local_name,
+                    )
+
+                block_expressions.append(resolved_expr)
+            return block_expression(block_expressions)
 
         if isinstance(expr, InExpression):
+            in_values = []
+            for expression in expr.values:
+                if self.session.eliminate_numeric_sql_value_cast_enabled:
+                    in_value = self.to_sql_try_avoid_cast(
+                        expression,
+                        df_aliased_col_name_to_real_col_name,
+                        parse_local_name,
+                    )
+                else:
+                    in_value = self.analyze(
+                        expression,
+                        df_aliased_col_name_to_real_col_name,
+                        parse_local_name,
+                    )
+
+                in_values.append(in_value)
             return in_expression(
                 self.analyze(
                     expr.columns, df_aliased_col_name_to_real_col_name, parse_local_name
                 ),
-                [
-                    self.analyze(
-                        expression,
-                        df_aliased_col_name_to_real_col_name,
-                        parse_local_name,
-                    )
-                    for expression in expr.values
-                ],
+                in_values,
             )
 
         if isinstance(expr, GroupingSet):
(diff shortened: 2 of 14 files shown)


tests/integ/test_column_names.py
```diff
@@ -648,49 +763,89 @@ def test_binary_expression(session):
...
-        == get_metadata_names(session, df2)[0]
-        == '"(""\'INF\'"" = \'INF\' :: FLOAT)"'
+
+    # local testing mode is not handling INF correctly today, skip the result
+    # value check when it is local testing mode
+    expected_rows = None if local_testing_mode else [Row(True)]
+    verify_column_result(
+        session,
+        df2,
+        ['"(""\'INF\'"" = \'INF\' :: FLOAT)"'],
```
sfc-gh-yzou, line 846: @sfc-gh-aling @sfc-gh-stan This case seems returning False for local testing mode, is that expected?

tests/integ/test_column_names.py
```diff
@@ -648,49 +763,89 @@ def test_binary_expression(session):
...
+        ],
+        [
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            LongType() if not local_testing_mode else BooleanType(),
```
sfc-gh-yzou, line 807: @sfc-gh-aling @sfc-gh-stan The bitwise operator seems returning Boolean type for local testing mode here, is that expected?

tests/integ/test_column_names.py
```diff
@@ -254,49 +297,66 @@ def test_rank_related_function_expression(session):
...
-    assert (
-        [x.name for x in df2._output]
-        == df2.columns
-        == get_metadata_names(session, df2)
-        == [
+
+    expected_dtypes = (
+        [StringType(), LongType(), BooleanType(), ArrayType(LongType())]
+        if local_testing_mode
+        else [StringType(), LongType(), BooleanType(), ArrayType(StringType())]
```
sfc-gh-yzou, line 347: @sfc-gh-jdu for this case,  it actually seems should be ArrayType(LongType), but without local testing mode, we seems getting ArrayType(StringType()), do you know if that is what we expected today?

tests/integ/test_column_names.py
sfc-gh-stan: Can you elaborate please? What's not being handled correctly?

tests/integ/test_column_names.py
sfc-gh-jdu: Yes, because inner type of an array is not enforced, because it's not a structured type. So the server will just return StringType

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - Added distributed tracing using open telemetry APIs for table stored procedure function in `DataFrame`:
   - _execute_and_get_query_id
+- Avoid casting for numeric sql values for binary expression, multiExpression and InExpression. The change is controlled by session parameter _eliminate_numeric_sql_value_cast_enabled.
```
sfc-gh-jdu, line 11: We can mention performance improvement for binary column operations and dataframe._in method (users don't know about internal expressions). Also, we should mention the public parameter, like "this performance optimization can be enabled by setting `session.eliminate_numeric_sql_value_cast_enabled=True`"

tests/integ/test_column_names.py
sfc-gh-stan: No it's not. I know why though. This is likely the `ColumnEmulator` doesn't have a numeric `dtype` when the bitwise operation happened. If the dtype is `object`, the return value will be `boolean`. That's why we have https://github.com/snowflakedb/snowpark-python/blob/8a4276387e09947d55139fac6a1d6e3590f3f891/src/snowflake/snowpark/mock/_util.py#L215

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -702,16 +722,22 @@ def window_frame_boundary(self, offset: str) -> str:
...
             return offset
 
-    def to_sql_avoid_offset(
+    def to_sql_try_avoid_cast(
         self,
         expr: Expression,
         df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
         parse_local_name: bool = False,
     ) -> str:
+        """
```
sfc-gh-jdu, line 731: 👍

tests/integ/scala/test_literal_suite.py
```diff
@@ -30,6 +30,7 @@
     reason="SNOW-1362917: Schema inference not fully aligned for local testing mode.",
 )
 def test_literal_basic_types(session):
+    session._eliminate_numeric_sql_value_cast_enabled = False
```
sfc-gh-jdu, line 33: remove?

src/snowflake/snowpark/session.py
```diff
@@ -569,6 +579,10 @@ def cte_optimization_enabled(self) -> bool:
         """
         return self._cte_optimization_enabled
 
+    @property
```
sfc-gh-jdu, line 582: we should have a set function?

tests/integ/test_column_names.py
sfc-gh-yzou: Hi @sfc-gh-stan sorry for the late reply. For this test case, it has a dataframe with one single column named 'INF', and have value 'inf', which is  counted as np.inf, and the test is simply checking if df1["'INF'"] == math.inf, which should be True, Snowpark python seems returning the correct result, but with local testing the result seems False.

tests/integ/test_column_names.py
sfc-gh-yzou: Oh, do we have a plan to fix that? is here a ticket to track this ?

sfc-gh-stan: I don't think this is a known issue. @sfc-gh-aling can you help unblock Yun here?

tests/integ/test_column_names.py
sfc-gh-stan: No worries, what is `df1["'INF'"]` returning for local testing then? It should also be `np.inf`.

tests/integ/test_column_names.py
sfc-gh-yzou: I don't think i am necessary blocking here, I can create a ticket for you to look into this if this is not a known issue, and left a comment on the test

tests/integ/scala/test_literal_suite.py
sfc-gh-yzou: rmoved

src/snowflake/snowpark/session.py
sfc-gh-yzou: yes, added, aslo added a telemetry on variable setting

tests/integ/test_column_names.py
sfc-gh-yzou: Ok, i think i know what is happening there. when we create a dataframe like this 
```
session.create_dataframe([["inf"]], schema=["'INF'"])
```
the column is actually mapped to string column, in snowflake, when comparing with 'inf' = 'INF'::FLOAT, it was able to do property casting, and return True. However, in local testing i believe it is not able to do such casting, in python, string to numeric value is not castable. That is the behavior difference between python and snowflake

tests/integ/test_column_names.py
sfc-gh-yzou: I see, so that is an expected behavior with snowpark on snowflake then.

tests/integ/test_session.py
```diff
@@ -681,6 +682,31 @@ def test_cte_optimization_enabled_on_session(db_parameters):
         assert new_session2.cte_optimization_enabled is True
 
 
+@pytest.mark.skipif(IS_IN_STORED_PROC, reason="Can't create a session in SP")
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
```
sfc-gh-stan, line 687: This should be an xfail instead. We distinguish between xfail and skipif such that xfail is for tests that will not be supported in Local Testing and skipif is for tests that are blocked on feature or bug fix in development.

tests/integ/test_simplifier_suite.py
```diff
@@ -57,6 +57,21 @@ def skip(pytestconfig):
         )
 
 
+paramList = [True, False]
+
+
+@pytest.fixture(params=paramList)
+def setup_reduce_cast(request, session):
```
sfc-gh-stan, line 64: Can this be done using pytest.parameterize?

tests/integ/test_column_names.py
sfc-gh-stan: Let's create a followup ticket for this. cc: @sfc-gh-jrose @sfc-gh-aling

tests/integ/test_column_names.py
sfc-gh-stan: This is [coercion](https://docs.snowflake.com/en/sql-reference/data-type-conversion).
How did the test work before then?

sfc-gh-yzou: The test doesn't check any result before, I have updated the whole test suite to check the result type and result in this change, and start seeing  mismatches here. It seems that the local testing is probably lacking some test coverage for special values.

CHANGELOG.md
sfc-gh-jdu: It seems the first part is not addressed

tests/integ/test_column_names.py
```diff
@@ -607,28 +690,60 @@ def test_binary_expression(session):
...
+        ],
+        [
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            BooleanType(),
+            LongType() if not local_testing_mode else BooleanType(),
```
sfc-gh-stan, line 735: Same issue as below

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -216,9 +228,22 @@ def schema_expression(data_type: DataType, is_nullable: bool) -> str:
...
+    Generate the sql str for numeric datatype without cast expression. One exception
+    is for float nan and inf, where a cast is always required for Snowflake to be able
+    to handle it correctly.
+    """
     if value is None:
         return "NULL"
-    if isinstance(datatype, StringType):
-        return f"'{value}'"
+
+    if not (isinstance(datatype, _NumericType)):
```
sfc-gh-jdu, line 240:
```suggestion
    if not isinstance(datatype, _NumericType):
```

src/snowflake/snowpark/session.py
```diff
@@ -631,6 +645,21 @@ def cte_optimization_enabled(self, value: bool) -> None:
             )
         self._cte_optimization_enabled = value
 
+    @eliminate_numeric_sql_value_cast_enabled.setter
+    @experimental_parameter(version="1.20.0")
+    def eliminate_numeric_sql_value_cast_enabled(self, value: bool) -> None:
+        """Set the value for eliminate_numeric_sql_value_cast_enabled"""
+
+        if value in [True, False]:
```
sfc-gh-jdu, line 653: You can only send telemetry when it is false for now

tests/integ/test_column_names.py
sfc-gh-yzou: created a ticket here https://snowflakecomputing.atlassian.net/browse/SNOW-1524633.  @sfc-gh-aling i assigned it to your for now, feel free to triage it.

tests/integ/test_simplifier_suite.py
```diff
@@ -1196,17 +1225,23 @@ def test_chained_sort(session):
...
     df1 = session.create_dataframe([[1, -2], [3, -4]], schema=["a", "b"])
 
     session.sql_simplifier_enabled = True
     df2 = session.create_dataframe([[1, -2], [3, -4]], schema=["a", "b"])
 
+    # replace '$POSTFIX' with the expected integer literal postfix to get the final expected simplified query
+    integer_literal_postfix = (
+        "" if session.eliminate_numeric_sql_value_cast_enabled else " :: INT"
+    )
+    simplified_query = simplified_query.replace("$POSTFIX", integer_literal_postfix)
```
sfc-gh-jdu, line 1243: you can use https://docs.python.org/3/library/stdtypes.html#str.format instead of replacing strings

tests/integ/test_column_names.py
sfc-gh-stan: I created [SNOW-1524637](https://snowflakecomputing.atlassian.net/browse/SNOW-1524637), please put this ticket number wherever you see this issue happen.

sfc-gh-jdu (approved):
except changelog and local testing change (I didn't review), other parts LGTM

tests/integ/test_column_names.py
sfc-gh-yzou: added comment with ticket number

tests/integ/test_column_names.py
sfc-gh-yzou: also added the ticket number in comment

CHANGELOG.md
sfc-gh-yzou: sorry, updated to refect the user facing expressions

tests/integ/test_session.py
sfc-gh-yzou: updated to xfail, i think I was copying this from other similar tests.

tests/integ/test_simplifier_suite.py
sfc-gh-yzou: yes, but i will need to add a finally to reset the parameter back at the end of each tests, this setup function captures all steps, easier for reuse the code, would prefer to use the fixture here since more than one test point is involved

tests/integ/test_column_names.py
sfc-gh-yzou: added comment with ticket number

src/snowflake/snowpark/session.py
sfc-gh-yzou: i actually do want to send telemetry for both values whenever user configures it, the telemetry method I added for this parameter is able to deal with both True and False.

tests/integ/test_simplifier_suite.py
sfc-gh-yzou: updated

sfc-gh-stan (approved):
Local Testing test changes LGTM

sfc-gh-aling (approved):

PR #1863, opened 2024-07-01
SNOW-1458125 Implement Index.astype

sfc-gh-vbudati:

   

   Fixes SNOW-1458125

   `Index.astype` converts the `dtype` of an Index object from one type to another. Below are some examples:
   
```py
>>> idx = pd.Index([10, 20, 30, 40], dtype=int)
>>> idx.astype(float)
Index([10.0, 20.0, 30.0, 40.0], dtype='float64')

>>> idx = pd.Index(["a", "b", "c"], dtype=str)
>>> idx.astype("object")
Index(['a', 'b', 'c'], dtype='object')

>>> idx = pd.Index([], dtype=object)
>>> idx.astype(int)
Index([], dtype='int64')

>>> idx = pd.Index([], dtype=int)
>>> idx.astype("O")
Index([], dtype='object')
```

Known issues:
Performing the following does not work as expected:
```py
>>> idx = pd.Index([10, 20, 30, 40], dtype=int)
>>> idx.astype(object)
Index([10, 20, 30, 40], dtype='int64') # <--- supposed to be object
```

This is most likely due to a known error with `apply` - `apply` is called during `to_pandas`, if all data in a variant column is of the same type, `apply` "infers" the type and incorrectly converts it, in this case, to int.
Tracked by tickets: SNOW-1480906 and SNOW-1514565.


CHANGELOG.md
```diff
@@ -103,11 +103,10 @@
...
 - Initial work to support an `pd.Index` directly in Snowpark pandas. Support for `pd.Index` as a first-class component of Snowpark pandas is coming soon.
-- Added a lazy index constructor and support for `len`, `shape`, `size`, `empty`, `to_pandas()` and `names`. For `df.index`, Snowpark pandas creates a lazy index object.
+- Added a lazy index constructor and support for `len`, `shape`, `size`, `empty`, `to_pandas`, and `names`. For `df.index`, Snowpark pandas creates a lazy index object.
 - For `df.index`, Snowpark pandas creates a lazy index object. 
 - For `df.columns`, Snowpark pandas supports a non-lazy version of an `Index` since the data is already stored locally.
-- Added support for `Index.copy()`
-- Added support for Index APIs: `dtype`, `values`, `item()`, `tolist()`, `to_series()` and `to_frame()`
+- Added support for Index APIs: `dtype`, `values`, `item`, `tolist`, `to_series`, `to_frame`, `copy`, and `as_type`.
```
sfc-gh-yzou, line 109: You don't need to merge the two items, this is change log, also served with a log purpose.

src/snowflake/snowpark/modin/pandas/series.py
```diff
@@ -2683,7 +2683,7 @@ def _create_or_update_from_compiler(self, new_query_compiler, inplace=False):
         new_query_compiler : PandasQueryCompiler
             QueryCompiler to use to manage the data.
         inplace : bool, default: False
-            Whether or not to perform update or creation inplace.
```
sfc-gh-yzou, line 2686: this seems an irrelevant change, can we revert it from this PR?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7865,6 +7865,7 @@ def astype(
         self,
         col_dtypes_map: dict[str, Union[dtype, ExtensionDtype]],
         errors: Literal["raise", "ignore"] = "raise",
+        is_index: bool = False,
```
sfc-gh-yzou, line 7868: maybe call it include_index to indicate wether to include the index for dtypes?

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -468,8 +469,7 @@ def shape(self) -> tuple:
         """
         return (len(self),)
 
-    @is_lazy_check
-    def astype(self, dtype: Any, copy: bool = True) -> Index:
+    def astype(self, dtype: str | type | ExtensionDtype, copy: bool = True) -> Index:
```
sfc-gh-yzou, line 472: did we double check that the generated doc have the correct Index class link?

sfc-gh-yzou, line 472: seems we removed the is_lazy_check for this API, is it now always lazy? feels we probably still need the is_lazy_check here, otherwise, what happens if the index is not lazily stored?

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -501,12 +501,20 @@ def astype(self, dtype: Any, copy: bool = True) -> Index:
...
+        if self.dtype == dtype:
+            # Ensure that self.astype(self.dtype) is self
+            return self.copy() if copy else self
+
+        col_dtypes = {
+            column: pandas_dtype(dtype)
+            for column in self._query_compiler.get_index_names()
+        }
+        new_query_compiler = self._query_compiler.astype(col_dtypes, is_index=True)
+        self.set_query_compiler(new_query_compiler)
```
sfc-gh-yzou, line 516: is astype an inplace operations? or it returns a new object? self.set_query_compiler sounds like we are replacing the query compiler of the current object? However, if we are returning a new object, we should create a new Index class object here, not self?

tests/integ/modin/index/test_index_methods.py
```diff
@@ -37,13 +37,46 @@
 ]
 
 
+@pytest.mark.parametrize(
+    "index, type",
+    [
+        (native_pd.Index([1, 2, 3, 4], dtype=int), float),
+        (native_pd.Index(["a", "b", "c"], dtype=str), "object"),
```
sfc-gh-yzou, line 44: can you add some more case here to test:
1) string with '1', '2', '3' to int/float
2) boolean to string
3) object with 1, 2, 3 to int

tests/integ/modin/index/test_index_methods.py
```diff
@@ -37,13 +37,46 @@
...
+    ],
+)
+@sql_count_checker(query_count=1)
+def test_index_astype(index, type):
+    snow_index = pd.Index(index)
+    native_index = index.astype(type)
+    snow_index = snow_index.astype(type)
+    assert_index_equal(snow_index, native_index)
+
+
```
sfc-gh-yzou, line 56: also add a test to test the invalid cases to check the behavior, for example, string to int

tests/integ/modin/index/test_index_methods.py
```diff
@@ -37,13 +37,46 @@
...
+    "index, type",
+    [
+        (native_pd.Index([1, 2, 3, 4], dtype=int), float),
+        (native_pd.Index(["a", "b", "c"], dtype=str), "object"),
+        (native_pd.Index([], dtype=object), int),
+        (native_pd.Index([], dtype=int), "O"),
+    ],
+)
+@sql_count_checker(query_count=1)
+def test_index_astype(index, type):
```
sfc-gh-yzou, line 50: also have an index with name, to make sure the index name is handled correctly

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -501,12 +501,20 @@ def astype(self, dtype: Any, copy: bool = True) -> Index:
...
-        )
+        if dtype is not None:
+            dtype = pandas_dtype(dtype)
+
+        if self.dtype == dtype:
+            # Ensure that self.astype(self.dtype) is self
+            return self.copy() if copy else self
+
+        col_dtypes = {
+            column: pandas_dtype(dtype)
```
sfc-gh-nkumar, line 512: nit: We already converted to pandas_dtype above. This can just be `column: dtype`

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-vbudati: I named it `is_index` because it is not doing index columns along with data columns, it's only doing index columns. Only the Index object uses this.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: Yup! Works as expected

sfc-gh-vbudati: Hmm that's a good point, the non-lazy functionality still exists. I added it back.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: Good catch, I fixed that. This is not an inplace operation.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -501,12 +508,18 @@ def astype(self, dtype: Any, copy: bool = True) -> Index:
         >>> idx.astype('float')
         Index([1.0, 2.0, 3.0], dtype='float64')
         """
-        WarningMessage.index_to_pandas_warning("astype")
-        return Index(
-            self.to_pandas().astype(dtype=dtype, copy=copy),
-            dtype=dtype,
-            convert_to_lazy=self.is_lazy,
-        )
+        if dtype is not None:
```
sfc-gh-nkumar, line 511: Discussed offline: We might need some tests to ensure this also works for non-lazy index. Otherwise Looks good!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7878,6 +7879,9 @@ def astype(
             Control raising of exceptions on invalid data for provided dtype.
             - raise : allow exceptions to be raised
             - ignore : suppress exceptions. On error return original object.
+        is_index : bool, default=False
+            Whether an Index object is invoking astype. Index objects use only index columns;
```
sfc-gh-yzou, line 7883: add more comments like if true, only index columns are involved during operations, and it only works for single index now .

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7888,16 +7892,23 @@ def astype(
             ErrorMessage.not_implemented(
                 f"Snowpark pandas astype API doesn't yet support errors == '{errors}'"
             )
-        col_dtypes_curr = {
-            k: v for k, v in self.dtypes.to_dict().items() if k in col_dtypes_map
-        }
+
+        if is_index:
+            col_dtypes_curr = {
+                column: self.index.dtype for column in self.get_index_names()
```
sfc-gh-yzou, line 7898: this implemention only works for single index, let's do:
if is_index:
   if self.is_multiindex() is True:
         raise NotImplementedError

tests/integ/modin/index/test_astype.py
```diff
@@ -0,0 +1,204 @@
...
+            'Failed to cast variant value "a" to FIXED',
+        ),
+        (
+            native_pd.Index([True, "b", "c", 1, 2, None, 4], dtype="O"),
+            bool,
+            'Failed to cast variant value "b" to BOOLEAN',
+        ),
+    ],
+)
+@sql_count_checker(query_count=0)
```
sfc-gh-yzou, line 132: Thanks a lot for adding those tests!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7888,16 +7892,23 @@ def astype(
...
+                k: v for k, v in self.dtypes.to_dict().items() if k in col_dtypes_map
+            }
 
         astype_mapping = {}
         id_to_sf_type_map = self._modin_frame.quoted_identifier_to_snowflake_type()
         labels = list(col_dtypes_map.keys())
         col_ids = (
             self._modin_frame.get_snowflake_quoted_identifiers_group_by_pandas_labels(
-                labels, include_index=False
+                labels,
```
sfc-gh-yzou, line 7910: so this relies on the fact that the query compiler for index object only contains index columns, otherwise, it will take both index and data column into account.
That was also why I was suggesting to change name from is_index to include_index. The two set of code added conflicts little bit on the purpose the code above only takes care of index, the implementation here takes both index and data column.

To make the implementation in this function more consistent, you can make the purpose clear to do either one of the following:
1) either apply astype on data column or index column only. If you do this, update get_snowflake_quoted_identifiers_group_by_pandas_labels to also take an parameter include_data_column, and make the function only look into index columns by setting include_data_column = False
2) apply astype on the internalframe with include_index as an option. if you decide to go with this option, then udpate the code for col_dtypes_curr to include both data column and  index column in the dict.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -108,6 +109,7 @@ def __init__(
         name: object = None,
         tupleize_cols: bool = True,
         convert_to_lazy: bool = True,
+        query_compiler: SnowflakeQueryCompiler | None = None,
```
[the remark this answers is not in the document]
sfc-gh-vbudati: ah that's a good point, I missed that. I'll fix that.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-vbudati: I chose to go with option #2 here, what you suggested makes sense. Thanks Yun!!

sfc-gh-nkumar (approved):

sfc-gh-yzou (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7981,16 +7988,28 @@ def astype(
             ErrorMessage.not_implemented(
                 f"Snowpark pandas astype API doesn't yet support errors == '{errors}'"
             )
-        col_dtypes_curr = {
-            k: v for k, v in self.dtypes.to_dict().items() if k in col_dtypes_map
-        }
+
+        col_dtypes_curr = {}
+        if include_index:
```
sfc-gh-yzou, line 7993: @sfc-gh-vbudati since you haven't merged this pr yet, how about let's just do consistent on all implemention on index API, have a different query compiler API for the index like index_astype, turn this funciton into internal function to use

sfc-gh-vbudati: done!

PR #1865, opened 2024-07-01
Only raise explicit errors for Snowpark non-internal API calls

sfc-gh-heshah:
Ensures all tests pass on main branch


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -15,7 +15,7 @@
 
 # This flag causes an explicit error to be raised if any Snowpark object instance is missing an AST or field, when this
 # AST or field is required to populate the AST field of a different Snowpark object instance.
-FAIL_ON_MISSING_AST = True
+FAIL_ON_MISSING_AST = False
```
sfc-gh-azwiegincew, line 18: This isn't the right fix. There are cases (such as `to_df`), where we legitimately don't pass any column AST because all these API invocations are internal to some higher level API (such as `to_df`).

sfc-gh-azwiegincew (approved):

PR #1866, opened 2024-07-01
SNOW-1372268 - np.where<cond>, scalar, scalar produces wrong output

sfc-gh-jkew:
SNOW-1372268 - np.where(<cond>, scalar, 0) produces wrong output
df.where(<cond>, 0) produces NaN values

Fixed the underlying issue in df.where; and simplified some of the numpy adapter code.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8665,7 +8665,7 @@ def where(
                 snowflake_quoted_identifier
             )
             # TODO (SNOW-904421): Other value can fail to cast in snowflake if not compatible type
-            if other_value:
+            if other_value or other_value == 0:
```
sfc-gh-nkumar, line 8668: Could we have same issue if other_value is `False`? Shall we check for `other_value is not None`?

sfc-gh-nkumar (commented):
Looks good!
I think we can further improve the condition to check for `not None` to fix this for other literal which are Falsy like `False`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: Good idea. False worked already because False == 0, but "is not None" is obviously much clearer.

sfc-gh-dpetersohn (approved):

sfc-gh-nkrishna (approved):

PR #1867, opened 2024-07-01
Fix flakiness in opentelemetry test

sfc-gh-yuwang:

   

   Fixes SNOW-NNNNNNN


sfc-gh-yuwang:
All the unit test is covered by integ test so it is safe to move all of them to integ test. The combine coverage test is a proof.

tests/unit/test_open_telemetry.py
```diff
@@ -109,366 +98,6 @@ def minus_udf(x: int, y: int) -> int:
     dict_exporter.clear()
 
 
-def test_register_udaf_from_file(dict_exporter):
-    test_file = os.path.normpath(
-        os.path.join(
-            os.path.dirname(__file__),
```
sfc-gh-yzou, line 115: it seems we removed all those tests? do we have a plan to add it back? or those tests are not necessary? 
Also, curious, what is causing the flakyness?

sfc-gh-yuwang: I put a comment above, all these tests are covered in integ tests, so it should be safe to remove them

sfc-gh-yuwang: The flakiness is caused by race condition I believe. When testing opentelemetry, I had to initialize a exporter. The exporter is like a global resource that cannot be re-initialized. Which makes all the tests accessing one resource. I am still looking if there is a better way to solve this issue

sfc-gh-yzou: I see. could you update the pr description with what are we doing in this pr, and why the test could be removed safely?

sfc-gh-yzou (approved):

tests/conftest.py
```diff
@@ -95,3 +95,42 @@ def clear_warning_dict():
     # clear the warning dict so that warnings from one test don't affect
     # warnings from other tests.
     warning_dict.clear()
+
+
+def attr_to_dict(span):
```
sfc-gh-aalam, line 100: these functions look like they belong to test utils instead of conftest

sfc-gh-aalam (commented):
I see that you are using new functions to check for correctness don't see anything specifically handling the race condition you mentioned. Can you point me to that?

tests/utils.py
```diff
@@ -1453,3 +1453,42 @@ class TypeMap(NamedTuple):
...
+                return False
+            else:
+                continue
+        if expected_answer[answer_name] != result[answer_name]:
+            return False
+    return True
+
+
+def check_answers(results, expected_answer):
+    results = span_extractor(results)
```
sfc-gh-yzou, line 1489: that seems used very specific for span check, can you rename the function to something like check_tracing_result or check_telemetry_result to be more clear, and add some comment for the function to describe what it is used for?

tests/utils.py
```diff
@@ -1453,3 +1453,42 @@ class TypeMap(NamedTuple):
...
     TypeMap("geometry", "geometry", GeometryType()),
 ]
+
+
+def attr_to_dict(span):
+    dict_attr = {}
+    for name in span.attributes:
+        dict_attr[name] = span.attributes[name]
+    dict_attr["code.filepath"] = os.path.basename(dict_attr["code.filepath"])
+    dict_attr["status_code"] = span.status.status_code
```
sfc-gh-yzou, line 1463: can we add type hint for those functions?

tests/utils.py
```diff
@@ -1453,3 +1453,42 @@ class TypeMap(NamedTuple):
...
+
+def span_extractor(dict_exporter):
+    spans = []
+    raw_spans = dict_exporter.get_finished_spans()
+    for span in raw_spans:
+        spans.append((span.name, attr_to_dict(span), span))
+    return spans
+
+
+def check_single_answer(result, expected_answer):
```
sfc-gh-yzou, line 1476: similar, check_single_answer -> check_single_span_result etc

sfc-gh-yzou: if nobody else is calling check_single_answer, there is probably no need to have a separate function for this

sfc-gh-yzou (commented):
Hi @sfc-gh-yuwang i left some minor comment about code part, but i don't think i have the expertise to review the fix for flakyness of the test. i will let @sfc-gh-aalam or someone else who knows this part better to make the decision

tests/utils.py
sfc-gh-yuwang: I will add docstrings and type hints.
This is a helper function for check_answer, it is not meant to directly be used in tests, so I have to keep this

sfc-gh-jdu (approved):

sfc-gh-aalam (approved):

PR #1870, opened 2024-07-01
SNOW-1018027 - Limited support for df.plot and series.plot through to_pandas

sfc-gh-jkew:
Super simple plot support. As seen in telemetry.
Includes docs-only changes to the Modin extensions functions to support @property accessors.
Does not include df.hist or other plotting methods in DFGroupBy. That will be a separate commit.


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - Added distributed tracing using open telemetry APIs for table stored procedure function in `DataFrame`:
   - _execute_and_get_query_id
+- Allow df.plot() and series.plot() to be called, materializing the data into the local client
```
sfc-gh-lspiegelberg, line 11: nit: trailing . at end of sentence.
Also I'd suggest putting ticks ` around the APIs in the log, so it looks prettier. I.e.,

`df.plot()` vs. df.plot()

docs/source/modin/supported/series_supported.rst
```diff
@@ -296,6 +296,8 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``pipe``                    | N                               |                                  |                                                    |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
+| ``plot``                    | D                               |                                  |                                                    |
```
sfc-gh-lspiegelberg, line 299: same here, I'd also maybe add to best practices a section that says:

If you want to plot large amounts of data, consider first minimizing the data being plotted. I.e., precompute aggregates/groupbys before plotting together with an example.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
```diff
@@ -57,3 +61,49 @@ def memory_usage(self, index: bool = True, deep: bool = False) -> Any:
...
+    sort_columns=False,
+    **kwargs,
+):  # noqa: PR01, RT01, D200
+    """
+    Make plots of ``DataFrame``. Materializes data into memory and uses the
+    existing pandas PlotAccessor
+    """
+    # TODO: SNOW-1063346: Modin upgrade - modin.pandas.DataFrame functions
+    WarningMessage.single_warning(
+        "DataFrame.plot materializes data to the local machine for plotting"
```
sfc-gh-lspiegelberg, line 107: Pls add . at end of sentence.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
```diff
@@ -57,3 +61,49 @@ def memory_usage(self, index: bool = True, deep: bool = False) -> Any:
...
+    )
+    return self._to_pandas().plot
```
sfc-gh-lspiegelberg, line 109: For my understanding, the decorator will call this function with all of the args?

tests/integ/modin/test_plot.py
```diff
@@ -0,0 +1,62 @@
...
+
+import io
+
+import modin.pandas as pd
+import pandas as npd
+
+species = ["adelie"] * 3 + ["chinstrap"] * 3 + ["gentoo"] * 3
+measurements = ["bill_length", "flipper_length", "bill_depth"] * 3
+values = [37.3, 187.1, 17.7, 46.6, 191.7, 17.6, 45.5, 212.7, 14.2]
+
```
sfc-gh-lspiegelberg, line 13: Can we add a test where to plot a Series/DataFrame is passed?

According to https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.html it seems to allow in `*args` a dataframe (and likely series) object. Just to make sure the frontend handles a necessary conversion if needed. 

It may also be that the documentation here is not accurate, or *args has self keyword folded. Prob. this scenario can be easily ticked off.

sfc-gh-lspiegelberg (approved):
some nits, one question re test scenario. Else, lgtm!

sfc-gh-dpetersohn (approved):
Overall LGTM!

tests/integ/modin/test_plot.py
sfc-gh-jkew: The docs are not accurate. The semantics of these accessors is a little odd.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
```diff
@@ -57,3 +61,49 @@ def memory_usage(self, index: bool = True, deep: bool = False) -> Any:
...
+    sort_columns=False,
+    **kwargs,
+):  # noqa: PR01, RT01, D200
+    """
+    Make plots of ``DataFrame``. Materializes data into memory and uses the
+    existing pandas PlotAccessor
+    """
+    # TODO: SNOW-1063346: Modin upgrade - modin.pandas.DataFrame functions
+    WarningMessage.single_warning(
+        "DataFrame.plot materializes data to the local machine for plotting."
```
sfc-gh-lmukhopadhyay, line 107: How likely is it that users would run into OOM issues when converting to_pandas? Should we add a more detailed warning if kind = line or scatter, such as "We recommend plotting a sample of the data to avoid memory issues"?

sfc-gh-lmukhopadhyay (approved):
LGTM, just 1 question!

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
sfc-gh-jkew: Not likely. Most users are using small data sets anyways. I think we can do a couple of things in the future though:
1. Have an implementation of to_pandas which can use a shared memory, copy on write cache, which invalidates itself when we make a change to the data frame
2. When the data is large, /then/ we print a warning
3. Introduce progress bars to make it clearer that there is a remote execution/environment in play

docs/source/modin/supported/series_supported.rst
sfc-gh-jkew: I think users are smart enough to know this. I actually think they know this better than we do ;)

sfc-gh-joshi (approved):
LGTM! Thanks.

PR #1875, opened 2024-07-03
SNOW-1458132 Implement Index.unique and Index.nunique

sfc-gh-vbudati:

   

   Fixes SNOW-1458132

   Implemented `Index.unique` and `Index.nunique`. 

`unique` returns an Index object containing all the unique values in the order in which they were found. With Index, a level can be specified to see where to pull the unique values from; this applies only for MultiIndex objects, which are not yet supported in Snowpark pandas. The groupby aggregation functionality in the query compiler is reused for this.

```py
>>> idx = pd.Index(["a", "b", 1, 2, None, None, "a", 2, 2])
>>> idx.unique()
Index(['a', 'b', 1, 2, None], dtype='object')
```

`nunique` returns the number of unique elements in a given Index object. The result returned will always be a scalar value since Index objects are single dimensional, this also applied to MultiIndex objects.

```py
>>> idx = pd.Index(["a", "b", 1, 2, None, None, "a", 2, 2])
>>> idx.nunique()
4
```


src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1392,8 +1414,24 @@ def nunique(self) -> None:
...
+        3    7
+        4    7
+        dtype: int64
+
+        >>> s.nunique()
+        4
+        """
+        return self._reduce_dimension(
+            self._query_compiler.nunique(axis=0, dropna=dropna, **{"is_index": True})
+        )
```
sfc-gh-mvashishtha, line 1434:
```suggestion
        return self._query_compiler.reset_index().nunique(axis=0, dropna=dropna).to_pandas().iloc[0,0]
```

I think you can implement this like so and revert some of the query compiler and Index changes:

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -415,17 +418,21 @@ def unique(self, level: Hashable | None = None) -> Index:
...
-        # TODO: SNOW-1458132 implement unique
-        WarningMessage.index_to_pandas_warning("unique")
-        return Index(self.to_pandas().unique(level=level))
+        # TODO: SNOW-1514782 MultiIndex - support level for Index.unique.
+        if level is not None:
+            ErrorMessage.not_implemented(
+                "level is not yet supported for Index.unique. It is used with "
+                "MultiIndex objects, which will be supported in the future."
+            )
+        return Index(data=self._query_compiler.unique(is_index=True, level=0))
```
sfc-gh-mvashishtha, line 435:
```suggestion
        result = Index(data=self._query_compiler.reset_index(names=['index']).unique().set_index(['index']))
        result.name = self.name
        return result
```

I think you can implement `unique()` this way without changing the query compiler `unique()` method:

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -415,17 +418,21 @@ def unique(self, level: Hashable | None = None) -> Index:
...
         >>> idx = pd.Index([1, 1, 2, 3, 3])
         >>> idx.unique()
         Index([1, 2, 3], dtype='int64')
         """
-        # TODO: SNOW-1458132 implement unique
-        WarningMessage.index_to_pandas_warning("unique")
-        return Index(self.to_pandas().unique(level=level))
+        # TODO: SNOW-1514782 MultiIndex - support level for Index.unique.
+        if level is not None:
+            ErrorMessage.not_implemented(
```
sfc-gh-mvashishtha, line 431: I think our coverage check will require you to add a test for `level`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7408,29 +7424,55 @@ def nunique(
         # Result is basically a series with the column labels as index and the distinct count as values
         # for each data column
         # frame holds rows with nunique values, but result must be a series so transpose single row
-        return self._nunique_columns(dropna).transpose_single_row()
+        return self._nunique_columns(
```
sfc-gh-yzou, line 7427: since you have extended the other apis to include is_index parameter, why not stay consistent on all API extension on this?

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: I got rid of the level check here since Yun mentioned that this logic is already taken care of [here](https://github.com/snowflakedb/snowpark-python/pull/1875/files/2f2f924670b5f9fbd446c1825c83a43619e19c8a#r1669002320).

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: I'm keeping the original logic I have here

tests/integ/modin/index/test_unique.py
```diff
@@ -0,0 +1,39 @@
...
+@sql_count_checker(query_count=1)
+def test_index_unique(native_index):
+    snow_index = pd.Index(native_index)
+    assert_index_equal(snow_index.unique(), native_index.unique())
+
+
+@sql_count_checker(query_count=2)
+def test_index_unique_nan_vs_none_negative():
+    native_index = native_pd.Index(["a", "b", 1, 2, np.nan, None, "a", 2])
+    snow_index = pd.Index(native_index)
```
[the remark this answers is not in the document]
sfc-gh-vbudati: I tried adding a test like this, PTAL!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7408,29 +7424,55 @@ def nunique(
...
         Return query compiler with unique values.
         """
-
-        assert 1 == len(
-            self._modin_frame.data_column_snowflake_quoted_identifiers
-        ), "unique can be only applied to 1-D DataFrame (Series)"
+        # With groupby, specify either one of `level` or `by`, not both.
+        if is_index is True:
+            # TODO: SNOW-1514782 MultiIndex - support level for Index.unique.
+            assert level is not None
```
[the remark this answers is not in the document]
sfc-gh-vbudati: For regular index, it's always 0. For MultiIndex it can be >= 0 which is why I have it like this.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7408,29 +7424,55 @@ def nunique(
...
-        empty.
+        Parameters
+        ----------
+        is_index : bool, optional
+            Set `is_index` to True when an Index object's query compiler is invoking this method.
+            When True, use index columns to compute unique instead of data columns.
+        level : int, optional
+            When `is_index` is True, this specifies which level of the MultiIndex to use when
+            performing unique. For single-dimensional Index objects, `level` defaults to 0.
+            This functionality is not fully implemented and currently only works for `level=0`.
```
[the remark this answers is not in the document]
sfc-gh-vbudati: should I get rid of references to MultiIndex?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-vbudati: I'm no longer implementing nunique as a special case

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -415,17 +415,20 @@ def unique(self, level: Hashable | None = None) -> Index:
...
         --------
         >>> idx = pd.Index([1, 1, 2, 3, 3])
         >>> idx.unique()
         Index([1, 2, 3], dtype='int64')
         """
-        # TODO: SNOW-1458132 implement unique
-        WarningMessage.index_to_pandas_warning("unique")
-        return Index(self.to_pandas().unique(level=level))
+        # TODO: SNOW-1514782 MultiIndex - support level for Index.unique.
+        return Index(
```
sfc-gh-mvashishtha, line 427: let's do this so we don't have to change the existing query compiler interface, and we're not using `reset_index()`:

```suggestion
        return Index(
            data=self._query_compiler.groupby_agg(
                by=self._query_compiler.get_index_names(axis=0), agg_func={}, axis=0, groupby_kwargs={"sort": False, "as_index": True, "dropna": False}, agg_args=[], agg_kwargs={}))
```

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1392,8 +1395,27 @@ def nunique(self) -> None:
...
+
+        >>> s.nunique()
+        4
+        """
+        return (
+            self._query_compiler.reset_index()
+            .nunique(axis=0, dropna=dropna)
+            .to_pandas()
+            .iloc[0, 0]
+        )
```
sfc-gh-mvashishtha, line 1418: @sfc-gh-yzou is correct that we should also avoid `reset_index()` here to reduce the number of `row_number` calls. We can do this, except we should probably abstract the `modin_frame` method calls into a separate method in the query compiler:


```suggestion
        from snowflake.snowpark.functions import col        
        return (
            SnowflakeQueryCompiler(self._query_compiler._modin_frame.append_column('index', col(self._query_compiler._modin_frame.index_column_snowflake_quoted_identifiers[0])))
            .nunique(axis=0, dropna=dropna)
            .to_pandas()
            .iloc[0, 0]
        )
```

tests/integ/modin/index/test_unique.py
```diff
@@ -33,7 +33,36 @@ def test_index_unique_nan_vs_none_negative():
...
+        {
+            "B": [10, 20, 30, 40, 50],
+            "A": [11, 22, 33, 44, 55],
+            "C": [60, 70, 80, 90, 100],
+        },
+        index=native_pd.Index([5, 4, 3, 2, 1], name="A"),
+    )
+    snow_df = pd.DataFrame(native_df)
+
+    with pytest.raises(AssertionError):
```
sfc-gh-mvashishtha, line 54: it's clearer to move this context manager around the `assert_index_equal` alone.

tests/integ/modin/index/test_unique.py
```diff
@@ -33,7 +33,36 @@ def test_index_unique_nan_vs_none_negative():
...
+        # Assert that even though the index column and a data column have
+        # the same name "A", the correct column is picked for both cases and
+        # different results are produced.
+        snow_index = snow_df.index
+        snow_df_a = snow_df.A
+        with SqlCounter(query_count=1):
```
sfc-gh-mvashishtha, line 60: For clarity, we should remove the `SqlCounter`.

tests/integ/modin/index/test_unique.py
```diff
@@ -33,7 +33,36 @@ def test_index_unique_nan_vs_none_negative():
...
+            assert_index_equal(snow_index.unique(), snow_df_a.unique())
+
+    with SqlCounter(query_count=1):
+        assert_index_equal(snow_df.index.unique(), native_df.index.unique())
+    # TODO: SNOW-1524901: snow_df.A.unique() does not produce the correct result here.
```
sfc-gh-mvashishtha, line 65: this should be a separate test case in `series/test_unique.py` marked as `pytest.mark.xfail(strict=True)`

tests/integ/modin/index/test_unique.py
```diff
@@ -33,7 +33,36 @@ def test_index_unique_nan_vs_none_negative():
...
+            "C": [60, 70, 80, 90, 100],
+        },
+        index=native_pd.Index([5, 4, 3, 2, 1], name="A"),
+    )
+    snow_df = pd.DataFrame(native_df)
+
+    with pytest.raises(AssertionError):
+        # Assert that even though the index column and a data column have
+        # the same name "A", the correct column is picked for both cases and
+        # different results are produced.
```
sfc-gh-mvashishtha, line 57: I don't understand the phrase "the correct column is picked for both cases and different results are produced." Which column? Picked for what? Which results? Why are they different? Is there a clearer way to describe what you're testing for here?

sfc-gh-vbudati: I tried to make it clearer. Lmk if this is still confusing

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha, line 1: Can we revert the changes to this file now?

tests/integ/modin/index/test_unique.py
sfc-gh-mvashishtha: Thank you!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-vbudati: good catch, done! Thanks for all the reviews Mahesh!

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1392,8 +1400,38 @@ def nunique(self) -> None:
...
+        1    3
+        2    5
+        3    7
+        4    7
+        dtype: int64
+
+        >>> s.nunique()
+        4
+        """
+        from snowflake.snowpark.functions import col
```
sfc-gh-mvashishtha, line 1418: nit: Can we move this import to the top of the file?

sfc-gh-mvashishtha (approved):

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1392,8 +1401,36 @@ def nunique(self) -> None:
...
+        3    7
+        4    7
+        dtype: int64
+
+        >>> s.nunique()
+        4
+        """
+        return (
+            SnowflakeQueryCompiler(
+                self._query_compiler._modin_frame.append_column(
```
sfc-gh-yzou, line 1421: at frontend layer, we are only suppose to access the query_compiler layer not the  _modin_frame layer. it seems we are just trying to get a data column that is the same as index column, then reuse the nunique, it is almost similar as doing reset_index, or insert a column to the index dataframe.

However, is the purpose to try avoid extend the nunique interface at query compiler, i don't think it worth it, nobody says the nuique implementaiton at query backend can only be applied on data columns. Or you can just add another  index_nunique function at query compiler, that reuses some funcitonality of the query compiler

sfc-gh-vbudati: I think the reason it's done here is to try to have as few changes as possible between upstream modin and this code base. I'll make this a new query compiler function.

sfc-gh-vbudati: Is this better?

sfc-gh-evandenberg (approved):
LGTM

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -415,17 +415,25 @@ def unique(self, level: Hashable | None = None) -> Index:
...
         Examples
         --------
         >>> idx = pd.Index([1, 1, 2, 3, 3])
         >>> idx.unique()
         Index([1, 2, 3], dtype='int64')
         """
-        # TODO: SNOW-1458132 implement unique
-        WarningMessage.index_to_pandas_warning("unique")
-        return Index(self.to_pandas().unique(level=level))
+        # TODO: SNOW-1514782 MultiIndex - support level for Index.unique.
```
sfc-gh-yzou, line 426: don't you need to do a check here that if (level is not None) and (level != 0 or level != -1), and raise the following Error, 
IndexError: Too many levels: Index has only 1 level, -2 is not a valid level number

The ticket doesn't seem at the right place.

sfc-gh-yzou: also add a test for this error case

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7460,6 +7460,22 @@ def make_nunique(identifier: str, dropna: bool) -> SnowparkColumn:
...
+    def nunique_index(self, dropna: bool) -> "SnowflakeQueryCompiler":
+        """
+        Return number of unique elements in an Index object.
+
+        Returns
+        -------
+        SnowflakeQueryCompiler where the first row in the first column contains the number
+        of unique elements.
+        """
+        return SnowflakeQueryCompiler(
```
sfc-gh-yzou, line 7472: that is better, but if you are already at the query compiler layer, why not extend the _nunique_columns(dropna) function to handle the index columns, and i don't think you need any transpose here also, there seems no need to append any new columns there

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-yzou: I am not i don't think that change is related to upstream modin, since this is completely new to modin. if you are talking about compiler interface, i can accept this argument, but then the way to resolve this is to have two different compiler interface there

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7460,6 +7460,22 @@ def make_nunique(identifier: str, dropna: bool) -> SnowparkColumn:
         )
         return SnowflakeQueryCompiler(frame)
 
+    def nunique_index(self, dropna: bool) -> "SnowflakeQueryCompiler":
```
sfc-gh-yzou, line 7463: also, if you have a function nunique_index here, you don't need to return a SnowflakeQueryCompiler anymore, you can just return a scaler, and move the evaluation to this function.

sfc-gh-yzou (approved):

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-mvashishtha: When we upstream this change to Modin, I recommend adding a method to add the index column as a data column, and not extending the interface of `unique`. That way, we can minimize the query compiler changes while also not accessing the private `_modin_frame`.

Generally we would like to avoid extending the query compiler interface with features that are only useful for indexes.

PR #1877, opened 2024-07-03
[feature branch] SNOW-1489960:Index accessors + NA

sfc-gh-azwiegincew:


src/snowflake/snowpark/dataframe.py
```diff
@@ -1390,20 +1391,24 @@ def filter(self, expr: ColumnOrSqlExpr) -> "DataFrame":
...
-        ast = stmt.expr.sp_dataframe_filter
-        self.set_ast_ref(ast.df)
-        set_src_position(ast.src)
-        if isinstance(expr, Column):
-            pass  # TODO
-        elif isinstance(expr, str):
-            ast.condition.sp_column_sql_expr.sql = expr
+        if _ast_stmt is None:
+            stmt = self._session._ast_batch.assign()
+            ast = stmt.expr.sp_dataframe_filter
```
sfc-gh-lspiegelberg, line 1401: should we use here the `with_src_position` helper as well?

sfc-gh-azwiegincew: Yes. However, I'd say this is outside of the scope of this diff (i.e. we need to update that in quite a few places). This code had to be made conditional to support nested invocations.

tests/ast/data/functions.test
```diff
@@ -1155,7 +1155,7 @@ res277 = res112.select(when(col("a") % 2 == 0, lit("even")))
 
 res278 = res112.select(iff(col("a") % 2 == 0, lit("even"), lit("odd")))
 
-res279 = res112.select(in_(Empty$, Empty$), in_([col("A"), col("B"), col("A")], Empty$), in_([col("col1"), col("col2")], [[1, "a"], [2, "b"]]))
+res279 = res112.select(in_(Add a handler for Empty$ in Unparser.expr(), Add a handler for Empty$ in Unparser.expr()), in_([col("A"), col("B"), col("A")], Add a handler for Empty$ in Unparser.expr()), in_([col("col1"), col("col2")], [[1, "a"], [2, "b"]]))
```
sfc-gh-oplaton, line 1158: I presume this is intentional, since this matches the current implementation. Is that correct?

sfc-gh-oplaton (approved):

tests/ast/data/functions.test
sfc-gh-azwiegincew: Yes. To be clear to everyone else who might be looking at this, I changed the catch-all handler (which we expect to assert when we're done) to generate an extended string instead of just the name of the AST entity that isn't being handled. It's clearly beyond the scope of this PR to fix the underlying issue.

PR #1890, opened 2024-07-09
SNOW-982770: [Local Testing] Add support for rank related features.

sfc-gh-jrose:

   

   Fixes SNOW-982770

   This PR does a few things to add support for rank related features to local testing.

- Adds some metadata to ColumnEmultors so that they can retain information about how they have been sorted. Arguably this could be made a private variable as I would not expect users to need direct access to this information outside of mocking their own functions.
- Modifies windowed functions to retain calculated expression columns that are used for ordering when passing to a mocked rank function.
- Adds the mocked rank functions and expands test examples to include more branches in the local testing logic.


src/snowflake/snowpark/mock/_snowflake_data_type.py
```diff
@@ -344,6 +349,7 @@ def __init__(
             {} if not sf_types_by_col_index else sf_types_by_col_index
         )
         self._null_rows_idxs_map = {}
+        self.sorted_by = []
```
sfc-gh-aling, line 352: how is this information used?

tests/integ/scala/test_function_suite.py
```diff
@@ -4484,29 +4484,67 @@ def test_iff(session, local_testing_mode):
...
 
-@pytest.mark.skipif(
-    "config.getoption('local_testing_mode', default=False)",
-    reason="cume_dist is not yet supported in local testing mode.",
-)
 def test_cume_dist(session):
     Utils.check_answer(
         TestData.xyz(session).select(
-            cume_dist().over(Window.partition_by(col("X")).order_by(col("Y")))
+            "X", "Y", cume_dist().over(Window.partition_by(col("X")).order_by(col("Y")))
```
sfc-gh-aling, line 4490: I see some changes in the test, like adding "X", "Y" in the select statement, change sort from `False` to `True` in the check_answer.

what's the purpose of the change here? is the previous test not working in the implementation or it's for adding more test coverage here?

src/snowflake/snowpark/mock/_snowflake_data_type.py
sfc-gh-jrose: This metadata allows the `handle_order_by_clause` to annotate a TableEmulator with the columns that it hsa been sorted by. That is useful because some of the rank functions need that information to break ties if the data is sorted.

sfc-gh-aalam (approved):

src/snowflake/snowpark/mock/_snowflake_data_type.py
sfc-gh-aling: oh yes, I saw that piece of information used by the _rank method

tests/integ/scala/test_function_suite.py
sfc-gh-aling: thanks, now I understand why cols X and Y are introduced.

btw does the server guarantee the same ordering or can return different orders. wondering if we should mention this in the limitation or known gap section of local testing framework.

sfc-gh-aling (approved):

tests/integ/scala/test_function_suite.py
sfc-gh-jrose: I believe the server order is consistent, not sure if it has a guarentee.

PR #1892, opened 2024-07-09
SNOW-1527038 Fix value_counts bug where the index is not sorted properly

sfc-gh-vbudati:

   

   Fixes SNOW-1527038

`Series.value_counts` is a method that counts the occurrence of each value in the given Series. There are several parameters to it, the only one relevant here is `sort`.

When `sort=True`, the result is supposed to display the value counts in descending order. When two values have the same count, maintain the ordering from the original Series. For instance, in the example below 3 appears before 2 and therefore the result is ordered as such.
```py
>>> pd.Series([1, 3, 1, 1, 1, 3, 1, 1, 1, 2, 2, 2, 3]).value_counts()
1    7
3    3
2    3
Name: count, dtype: int64
```

In Snowpark pandas, the descending order behavior is followed. However, the original order of the Series is not maintained when tie breaking is necessary. The test with which I found the bug is below. Notice how the positions of 5 and 6 are swapped - 5 appears before 6 but Snowpark pandas displays 6 first.
```py
>>> native_series = native_pd.Series([1, 2, 3, 1, 2, 3, 5, 6, 7, 8, 4, 4, 5, 6, 7, 1, 2, 1, 2, 3, 4, 3, 4, 5, 6, 7])
>>> native_series.value_counts()
1    4
2    4
3    4
4    4
5    3 <--- 
6    3 <---
7    3
8    1
Name: count, dtype: int64

>>> snow_series = pd.Series(native_series)
>>> snow_series.value_counts()
1    4
2    4
3    4
4    4
6    3 <---
5    3 <---
7    3
8    1
Name: count, dtype: int64
```

Internally, a groupby aggregation COUNT is performed on the data column in Series. Specifying `sort=False` in the `groupby_agg` method should provide a result with the required order - when `sort=False`, the original order in the Series is maintained. However in `value_counts`, after the groupby aggregation is performed, another sort is performed on the result `ordered_dataframe` which gets rid of the "row number ordering".

To fix this bug, after the groupby aggregation is performed, ensure that the row position column exists in the internal frame and explicitly specify sorting by row number.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -10177,7 +10177,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
             by=by,
             agg_func={COUNT_LABEL: "count"},
             axis=0,
-            groupby_kwargs={"dropna": dropna},
+            groupby_kwargs={"dropna": dropna, "sort": False},
             agg_args=(),
             agg_kwargs={},
         )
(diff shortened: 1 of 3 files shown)


sfc-gh-rdurrani (approved):
LGTM!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10196,13 +10196,21 @@ def value_counts(
             count_identifier = internal_frame.data_column_snowflake_quoted_identifiers[
                 0
             ]
+        internal_frame = internal_frame.ensure_row_position_column()
 
         # When sort=True, sort by the frequency (count column);
         # otherwise, respect the original order (use the original ordering columns)
         ordered_dataframe = internal_frame.ordered_dataframe
         if sort:
+            # When sort=True, sort the value counts in descending order but maintain the
```
sfc-gh-joshi, line 10205: nit: sort of counts is determined by `ascending` flag

sfc-gh-joshi: can you also add a comment mentioning that the first specified `OrderingColumn` is the last sort to be applied?

sfc-gh-joshi (approved):

PR #1906, opened 2024-07-11
SNOW-1462014: Fix Ability to disable OTel span creation even if opentelemetry is installed

sfc-gh-yuwang:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Improvements
 - Added support server side string size limitations.
-- Added support for `session.enable_span_record` and `session.disable_span_record`that user can decide whether record open telemetry span. The following APIs create spans:
+- Added support for environment variable `SNOWPARK_SPAN_RECORD` that user can decide whether record open telemetry span. Recording is enabled when `SNOWPARK_SPAN_RECORD` is not set or set to `enable`, it is disabled when `SNOWPARK_SPAN_RECORD` set to `disable`. The following APIs create spans:
   - `DataFrame.collect`
   - `DataFrame.collect_nowait`
   - `DataFrame.to_pandas`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Improvements
 - Added support server side string size limitations.
-- Added support that user can decide whether record open telemetry span.
+- Added support for `functions.enable_open_telemetry` and `functions.disable_open_telemetry`that user can decide whether record open telemetry span.
 
 #### Bug Fixes
 - Fixed a bug where SQL generated for selecting `*` column has an incorrect subquery.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Improvements
 - Added support server side string size limitations.
-- Added support for `functions.enable_open_telemetry` and `functions.disable_open_telemetry`that user can decide whether record open telemetry span.
+- Added support for `functions.enable_span_record` and `functions.disable_span_record`that user can decide whether record open telemetry span.
 
 #### Bug Fixes
 - Fixed a bug where SQL generated for selecting `*` column has an incorrect subquery.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,20 @@
 
 #### Improvements
 - Added support server side string size limitations.
-- Added support for `functions.enable_span_record` and `functions.disable_span_record`that user can decide whether record open telemetry span.
+- Added support for `session.enable_span_record` and `session.disable_span_record`that user can decide whether record open telemetry span. The following APIs create spans:
+  - `DataFrame.collect`
+  - `DataFrame.collect_nowait`
+  - `DataFrame.to_pandas`
+  - `DataFrame.count`
+  - `DataFrame.show`
+  - `DataFrame._execute_and_get_query_id`
+  - `DataFrameWriter.save_as_table`
+  - `UDFRegistration.register`
+  - `UDFRegistration.register_from_file`
+  - `UDAFRegistration.register`
+  - `UDAFRegistration.register_from_file`
+  - `UDTFRegistration.register`
+  - `UDTFRegistration.register_from_file`
 
 #### Bug Fixes
 - Fixed a bug where SQL generated for selecting `*` column has an incorrect subquery.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Improvements
 - Added support server side string size limitations.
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/functions.py
```diff
@@ -8536,3 +8536,15 @@ def locate(expr1: str, expr2: ColumnOrName, start_pos: int = 1) -> Column:
     _substr = lit(expr1)
     _str = _to_col_if_str(expr2, "locate")
     return builtin("charindex")(_substr, _str, lit(start_pos))
+
+
+def enable_open_telemetry():
```
sfc-gh-enie, line 8541: small nit: should this say `enable_tracing()`? opentelemetry is also logs and metrics, I think the intention is only to control tracing, correct?

sfc-gh-yuwang: What about `enable_span_record()`, I think this is more specific about what the function do

sfc-gh-enie: Yeah makes sense, as long as it's scoped to tracing concepts.

sfc-gh-enie: cc @sfc-gh-bli Should we have similar API on Java side as well?

src/snowflake/snowpark/_internal/open_telemetry.py
```diff
@@ -25,6 +25,7 @@
 registration_modules = ["udf.py", "udtf.py", "udaf.py", "stored_procedure.py"]
 # this parameter make sure no error when open telemetry is not installed
 open_telemetry_found = True
+open_telemetry_enabled = True
```
sfc-gh-enie, line 28: nit: keep this consistent with `enable_span_record`

sfc-gh-enie (approved):

src/snowflake/snowpark/functions.py
```diff
@@ -8536,3 +8536,23 @@ def locate(expr1: str, expr2: ColumnOrName, start_pos: int = 1) -> Column:
     _substr = lit(expr1)
     _str = _to_col_if_str(expr2, "locate")
     return builtin("charindex")(_substr, _str, lit(start_pos))
+
+
+def enable_span_record() -> None:
+    """
+    This function enable open telemetry span recording in action functions. If opentelemetry libraries exist in the environment, span will be recorded.
```
sfc-gh-jfreeberg, line 8543:
```suggestion
    This function enables Open Telemetry instrumentation. If Open Telemetry libraries exist in the environment, spans will be recorded.
    
    The following APIs create spans:
    - DataFrame.count()
    - DataFrame.to_pandas()
    - DataFrame.collect()
    - DataFrameWriter.copy_into_location()
    - DataFrameWriter.save_as_table()
    - UDFRegistration.register_from_file()
    - UDFRegistration.register()
    - StoredProcedureRegistration.register()
    - StoredProcedureRegistration.register_from_file()
```

We need a place to document which Snowpark APIs emit spans, can we use this method? Seems appropriate.

sfc-gh-jfreeberg: @sfc-gh-yuwang  fyi

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 #### Improvements
 - Added support server side string size limitations.
+- Added support for `functions.enable_span_record` and `functions.disable_span_record`that user can decide whether record open telemetry span.
```
sfc-gh-yixie, line 9: `functions` might not be the good place for them.

CHANGELOG.md
```diff
@@ -6,6 +6,20 @@
 
 #### Improvements
 - Added support server side string size limitations.
+- Added support for environment variable `SNOWPARK_SPAN_RECORD` that user can decide whether record open telemetry span. Recording is enabled when `SNOWPARK_SPAN_RECORD` is not set or set to `enable`, it is disabled when `SNOWPARK_SPAN_RECORD` set to `disable`. The following APIs create spans:
+  - `DataFrame.collect`
+  - `DataFrame.collect_nowait`
+  - `DataFrame.to_pandas`
+  - `DataFrame.count`
+  - `DataFrame.show`
+  - `DataFrame._execute_and_get_query_id`
```
sfc-gh-yixie, line 15: This is an internal API. We shouldn't mention it in the change log.

CHANGELOG.md
```diff
@@ -6,6 +6,20 @@
 
 #### Improvements
 - Added support server side string size limitations.
+- Added support for environment variable `SNOWPARK_SPAN_RECORD` that user can decide whether record open telemetry span. Recording is enabled when `SNOWPARK_SPAN_RECORD` is not set or set to `enable`, it is disabled when `SNOWPARK_SPAN_RECORD` set to `disable`. The following APIs create spans:
```
sfc-gh-yixie, line 9: Better align with python connector connections toml and other SDKs for this name (SNOWPARK_SPAN_RECORD). We don't want users to config Snowflfake's SDKs one by one. A single one for all of them is better.

sfc-gh-yixie: `SNOWFLAKE_...` may be better. See here https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect

CHANGELOG.md
sfc-gh-yixie: @sfc-gh-yuwang I don't see it any more in the code. Is it removed?

src/snowflake/snowpark/_internal/open_telemetry.py
```diff
@@ -36,7 +36,10 @@
 def open_telemetry_context_manager(func, dataframe):
 
     # trace when required package is installed
-    if open_telemetry_found:
+    if (
+        open_telemetry_found
+        and os.environ.get("SNOWPARK_SPAN_RECORD", "enable") == "enable"
```
sfc-gh-yixie, line 41: Two questions:
Is `os.env.get("abc")` this case insensitive?
Do we use true/false for other env variables or enable/disable?

sfc-gh-yuwang: 1. yes, it is case sensitive
2. seems like we don't have a tradition for this because it only allow string, but I can change it to "True"/"False"

sfc-gh-jfreeberg:
Ellen did some digging and found this: https://opentelemetry.io/docs/zero-code/python/configuration/#disabling-specific-instrumentations

It looks like there's a standard env var for this: `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS`. So we should probably configure our package to just respect this config: *The environment variable can be set to a comma-separated list of instrumentations entry point names to exclude from instrumentation. Most of the time the entry point name is the same as the package name and it is set in the project.entry-points.opentelemetry_instrumentor table in the package pyproject.toml file.*

PR #1908, opened 2024-07-11
[SNOW-1491300] Support Dataframe.{distinct,drop_duplicates} in Snowpark IR

sfc-gh-lspiegelberg:
Support `Dataframe.{distinct,drop_duplicates}` AST encoding. Fixes `update-from-devvm.sh` file to retrieve both `ast_pb2.py` and `scalapb_pb2.py` generated protobuf files from devvm/ws.


sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/dataframe.py
```diff
@@ -1191,12 +1194,14 @@ def select(
...
-                if ast:
+                if not _suppress_ast and ast:
                     ast.cols.append(e._ast)
 
             elif isinstance(e, str):
-                col_expr_ast = ast.cols.add() if ast else proto.Expr()
-                fill_ast_for_column(col_expr_ast, e, None)
+                col_expr_ast = None
+                if not _suppress_ast:
+                    col_expr_ast = ast.cols.add() if ast else proto.Expr()
```
sfc-gh-azwiegincew, line 1203: variadic = True

src/snowflake/snowpark/dataframe.py
```diff
@@ -1175,14 +1176,16 @@ def select(
...
-            stmt = self._session._ast_batch.assign()
-            ast = with_src_position(stmt.expr.sp_dataframe_select__columns, stmt)
-            self.set_ast_ref(ast.df)
-            ast.variadic = is_variadic
-        else:
-            stmt = _ast_stmt
-            ast = None
+        stmt = None
+        if not _suppress_ast:
+            if _ast_stmt is None:
```
sfc-gh-heshah, line 1181: Minor nit: I found it easier to add an `_emit_ast` parameter instead and change the first condition to `if _ast_stmt is None and _emit_ast`. This avoids the double negative and the additional changes to the logic in `select`, `filter`, and `distinct` later.

Either approach accomplishes the same thing though, so it's up to our preference for how we want to suppress AST generation internally, since this will likely be needed across many additional APIs.

sfc-gh-oplaton (approved):

src/snowflake/snowpark/dataframe.py
sfc-gh-lspiegelberg: I have no strong preference over `_supress_ast` or `emit_ast`. However, we should keep one. I used `_supress_ast` to keep the naming similar to existing APIs (e.g., table).

PR #1909, opened 2024-07-11
SNOW-1527038 Fix Jenkins failure, explain `Series.value_counts` behavior 

sfc-gh-vbudati:

   

   Fixes SNOW-1527038

This PR is here since there were some issues with running the `series/test_value_counts.py::test_series_value_counts` test. When run locally, the Snowpark pandas output matches native pandas. However, when run on Jenkins, the output differs.

Jenkins runs to refer to: prod runs 604, 605, and 606 (this run has clearer console output so that you can see the difference).

On closer inspection, the Snowpark pandas output is deterministic - native pandas' output is the one that differed on both platforms. 

There are two parameters to `Series.value_counts` that control the ordering: `sort` and `ascending`. `sort` sorts the value counts when True. `ascending` specifies whether this count is going to be sorted in ascending order or descending order. 

In the event of a tie, native pandas is supposed to respect the original ordering in the Series. This is done with the help of Python hashmaps. However, the hashmap's behavior can differ based on Python versions and platforms, making the result of `value_counts` in native Pandas non-deterministic. Snowpark pandas does not have this determinism issue and respects the original ordering in the Series.

Refer to pandas GitHub issue: https://github.com/pandas-dev/pandas/issues/15833

This PR is to make this behavior clear in the Series documentation and in the test where this discrepancy arises.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/docstrings/series.py
+++ b/src/snowflake/snowpark/modin/plugin/docstrings/series.py
@@ -2912,15 +2912,16 @@ class Series:
             frequencies of the unique values. Being different from native pandas,
             Snowpark pandas will return a Series with `decimal.Decimal` values.
         sort : bool, default True
-            Sort by the count when True. Preserve the order of the data when False.
+            Sort by frequencies when True. Preserve the order of the data when False.
             When there is a tie between counts, the order is still deterministic, but
             may be different from the result from native pandas. Snowpark pandas will
             always respect the order of insertion during ties. Native pandas is not
-            deterministic since the original order/order of insertion is based on the
-            Python hashmap which may produce different results on different versions.
+            deterministic when `sort=True` since the original order/order of insertion
+            is based on the Python hashmap which may produce different results on
+            different versions.
             Refer to: https://github.com/pandas-dev/pandas/issues/15833
         ascending : bool, default False
-            Whether to sort the count in ascending order or descending order.
+            Whether to sort the frequencies in ascending order or descending order.
         bins : int, optional
             Rather than count values, group them into half-open bins,
             a convenience for ``pd.cut``, only works with numeric data.
--- a/src/snowflake/snowpark/modin/plugin/docstrings/series.py
+++ b/src/snowflake/snowpark/modin/plugin/docstrings/series.py
@@ -2912,11 +2912,15 @@ class Series:
             frequencies of the unique values. Being different from native pandas,
             Snowpark pandas will return a Series with `decimal.Decimal` values.
         sort : bool, default True
-            Sort by frequencies when True. Preserve the order of the data when False.
+            Sort by the count when True. Preserve the order of the data when False.
             When there is a tie between counts, the order is still deterministic, but
-            may be different from the result from native pandas.
+            may be different from the result from native pandas. Snowpark pandas will
+            always respect the order of insertion during ties. Native pandas is not
+            deterministic since the original order/order of insertion is based on the
+            Python hashmap which may produce different results on different versions.
+            Refer to: https://github.com/pandas-dev/pandas/issues/15833
         ascending : bool, default False
-            Sort in ascending order.
+            Whether to sort the count in ascending order or descending order.
         bins : int, optional
             Rather than count values, group them into half-open bins,
             a convenience for ``pd.cut``, only works with numeric data.
--- a/src/snowflake/snowpark/modin/plugin/docstrings/series.py
+++ b/src/snowflake/snowpark/modin/plugin/docstrings/series.py
@@ -2913,12 +2913,12 @@ class Series:
             Snowpark pandas will return a Series with `decimal.Decimal` values.
         sort : bool, default True
             Sort by frequencies when True. Preserve the order of the data when False.
-            When there is a tie between counts, the order is still deterministic, but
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -2912,11 +2912,15 @@ def value_counts():
             frequencies of the unique values. Being different from native pandas,
             Snowpark pandas will return a Series with `decimal.Decimal` values.
         sort : bool, default True
-            Sort by frequencies when True. Preserve the order of the data when False.
+            Sort by the count when True. Preserve the order of the data when False.
```
sfc-gh-yzou, line 2915: Let's also use frequencies here to stay consistent with pandas.

I think we want to add that "original data order is preserved when there are equal frequency" for sort = True part

src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -2912,11 +2912,15 @@ def value_counts():
...
             When there is a tie between counts, the order is still deterministic, but
-            may be different from the result from native pandas.
+            may be different from the result from native pandas. Snowpark pandas will
+            always respect the order of insertion during ties. Native pandas is not
+            deterministic since the original order/order of insertion is based on the
+            Python hashmap which may produce different results on different versions.
+            Refer to: https://github.com/pandas-dev/pandas/issues/15833
         ascending : bool, default False
-            Sort in ascending order.
+            Whether to sort the count in ascending order or descending order.
```
sfc-gh-yzou, line 2923: let's use frequencies here also

src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -2912,11 +2912,15 @@ def value_counts():
             frequencies of the unique values. Being different from native pandas,
             Snowpark pandas will return a Series with `decimal.Decimal` values.
         sort : bool, default True
-            Sort by frequencies when True. Preserve the order of the data when False.
+            Sort by the count when True. Preserve the order of the data when False.
             When there is a tie between counts, the order is still deterministic, but
-            may be different from the result from native pandas.
+            may be different from the result from native pandas. Snowpark pandas will
+            always respect the order of insertion during ties. Native pandas is not
+            deterministic since the original order/order of insertion is based on the
```
sfc-gh-yzou, line 2919: add "not deterministic when sort = True"

tests/integ/modin/series/test_value_counts.py
```diff
@@ -129,6 +129,15 @@ def test_value_counts_bins():
...
 def test_series_value_counts(native_series, normalize, sort, ascending, dropna):
+    # If `sort=True`, sort by the count. If `sort=False`, maintain the original ordering.
+    # `ascending` controls whether to sort the count in ascending or descending order.
+
+    # When there is a tie between counts, the order is still deterministic, but
+    # may be different from the result from native pandas. Snowpark pandas will
+    # always respect the order of insertion during ties. Native pandas is not
+    # deterministic since the original order/order of insertion is based on the
+    # Python hashmap which may produce different results on different versions.
+    # Refer to: https://github.com/pandas-dev/pandas/issues/15833
```
sfc-gh-yzou, line 140: hmm, don't we need to update the test to mitigate the jekins failure? update the expected result to check the actual snowpark pandas result use assert_frame_equal to compare with an expected result

src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -2914,9 +2914,14 @@ def value_counts():
         sort : bool, default True
             Sort by frequencies when True. Preserve the order of the data when False.
             When there is a tie between counts, the order is still deterministic, but
```
sfc-gh-yzou, line 2916: add  "where the order in the original data is preserved" after "the order is still deterministic"

tests/integ/modin/series/test_value_counts.py
```diff
@@ -137,3 +180,123 @@ def test_series_value_counts(native_series, normalize, sort, ascending, dropna):
...
+            True,
+            True,
+            False,
+        ),
+    ],
+)
+@sql_count_checker(query_count=1)
+def test_series_value_counts_non_deterministic_pandas_behavior(
+    native_series, expected, normalize, sort, ascending, dropna
+):
```
sfc-gh-yzou, line 285: @sfc-gh-vbudati sorry, i am little bit confused by this change here. i see you now added a new test here to directly check the result. Then how does this fix the other test point test_series_value_counts? the other test seems going to continue fail inside jenkins or even github under some cases, right?
The purpose here is to make sure out test are stable under out env, right? so we need to fix the test_series_value_counts with correct result check, right ?

sfc-gh-vbudati: Sorry, I forgot to remove the original faulty tests - I just did that. Only tests that pass should be here. I only explicitly compare the results for the tests that have Jenkins failures - otherwise I'd have to do 80 comparisons.

sfc-gh-yzou (approved):

sfc-gh-lmukhopadhyay (approved):
LGTM!

PR #1910, opened 2024-07-12
SNOW-1342243: Call getOrCreate at package import time so that sproc user does not need to

sfc-gh-helmeleegy:

   

   Fixes SNOW-1342243

   Call getOrCreate at package import time so that sproc user does not need to.


sfc-gh-jrose (commented):
This change seems dangerous to me. Here are some scenarios that might break:

- What happens when a user does not have credentials configured before importing this package.
- What happens if the user ignores the default session and creates their own, then calls something that references [get_active_session](https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/session.py#L205)?
- What happens if a user imports this package while using local testing mode?

src/snowflake/snowpark/modin/pandas/__init__.py
```diff
@@ -346,3 +346,7 @@ def __getattr__(name: str) -> Any:
 modin.core.dataframe.algebra.default2pandas.default.DefaultMethod.register = (
     snowflake.snowpark.modin.core.dataframe.algebra.default2pandas.default.DefaultMethod.register
 )
+
+# Call getOrCreate() at import time on behalf of user so that in a stored procedure,
+# the user does not need to do it explicitly.
+snowflake.snowpark.Session.SessionBuilder().getOrCreate()
```
sfc-gh-yzou, line 352: @sfc-gh-helmeleegy for snowpark pandas, we by design to ask the user to explicitly set up the session to stay consistent with sowpark usage, and we do not require customer to provide connection files , so getOrCreate() could fail if user does not have a connection file. 

With stored sproc, i thought the session object is automatically created, and snowaprk pandas have the capablity to get the current active session for usage.  For some reason, that seems not the case. @sfc-gh-sfan why it is happening in such way?

sfc-gh-yzou: actually @sfc-gh-sfan I have a question, do you have a code pointer on how the session creation is handled in stored sproc today?

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@
 #### Bug Fixes
 - Fixed an issue when using np.where and df.where when the scalar 'other' is the literal 0.
 
+#### Improvements
+- Call getOrCreate at package import time so that sproc user does not need to.
```
sfc-gh-mvashishtha, line 31: I would rewrite this as "Make importing Snowpark pandas automatically create a Snowpark session in stored procedures."

tests/integ/test_stored_procedure.py
```diff
@@ -239,6 +239,20 @@ def sp_pow(session_, x, y):
     assert pow_sp(2, 10, session=session) == 1024
 
 
+def test_basic_snowpark_pandas_stored_procedure(session):
+    def return1(session_):
```
sfc-gh-mvashishtha, line 243: I thought the goal was to test that `session` is not required.

src/snowflake/snowpark/modin/pandas/__init__.py
```diff
@@ -346,3 +348,8 @@ def __getattr__(name: str) -> Any:
 modin.core.dataframe.algebra.default2pandas.default.DefaultMethod.register = (
     snowflake.snowpark.modin.core.dataframe.algebra.default2pandas.default.DefaultMethod.register
 )
+
+# Call getOrCreate() at import time on behalf of user so that in a stored procedure,
+# the user does not need to do it explicitly.
+if is_in_stored_procedure():
+    snowflake.snowpark.Session.SessionBuilder().getOrCreate()
```
sfc-gh-mvashishtha, line 355: If it's possible to create multiple sessions in a stored procedure, then there's an edge case where the user creates multiple Snowpark sessions in the stored procedure and then imports snowpark pandas. Before this change, that would work, but now, they'll get an [error](https://github.com/snowflakedb/snowpark-python/blob/d80a1950c385ae6204e0e1bfc3d52b5e49f5a85c/src/snowflake/snowpark/session.py#L210).

sfc-gh-mvashishtha, line 355: Should we add a TODO to stop doing this once Snowpark always creates the session automatically, with a link to the jira ticket for that?

sfc-gh-yzou: I am not aware that snowpark has plan to create the session automatically. cc @sfc-gh-jdu do we have any plan for this?

The problem here is more for stored sproc, in stored sproc, user is suppose to use the session created by stored sproc which uses specific connection configurations, because the one used by users are not usable within stored sproc. 
Originally, we thought stored sproc will always create the session, but seems it is only creating the session when the session parameter passed. The fix/workaround here seems make sure if snowpark pandas is used in stored sproc, make sure we create the session.
However, since session parameter is a required parameter for stored sproc before, it seems it is almost always creating a session object. I think we can always create a session object in snowpark python stored sproc, i don't see a reason why user will not use a session object with snowpark python stored sproc. cc @sfc-gh-sfan

sfc-gh-aalam: I don't understand the need for this. Inside a python stored proc environment, we already have access to a session object. Why can't we use that?

src/snowflake/snowpark/modin/pandas/__init__.py
sfc-gh-helmeleegy: After making the session argument optional in the proc's handler, the session object was not being created anymore in the case when the session argument was not specified. That's because it was not being passed along with the other arguments to the handler anymore.

@sfc-gh-zyao has a [draft PR](https://github.com/snowflakedb/snowflake/pull/193694) for pre-creating session objects even for session-less handlers, which seems promising. Until this becomes available, we will need to either pass the session argument, or create it at the sproc side using `getOrCreate`. Currently, it's left to the user to create it inside the body of the handler.

This PR attempts to automatically create the session object at package import time (only for sprocs) so that users do not need to worry about creating it themselves.

CHANGELOG.md
sfc-gh-helmeleegy: Good suggestion. Updated.

tests/integ/test_stored_procedure.py
sfc-gh-helmeleegy: Before this PR, the Snowpark python API would still require having a session argument for the sproc's handler. I updated the PR to allow for optional session argument even when using the Snowpark python API and added tests for both cases (with and without the session argument).

@sfc-gh-sfan would be great if you can confirm that the update to Snowpark python does make sense.

src/snowflake/snowpark/modin/pandas/__init__.py
sfc-gh-helmeleegy: @sfc-gh-aalam As pointed out in another thread here. Since we made the session argument optional (at the server/xp side), the session object was not being created anymore unless it was explicitly specified in the sproc's handler. Otherwise, it won't be created. This [draft PR](https://github.com/snowflakedb/snowflake/pull/193694) should make sure it's created in both cases.

@sfc-gh-mvashishtha I guess if we want, we can try-catch this scenario and suppress that error if it was raised - to make sure we keep the current behavior?

sfc-gh-helmeleegy: @sfc-gh-yzou Please refer to the other conversation threads.

sfc-gh-helmeleegy:
> This change seems dangerous to me. Here are some scenarios that might break:
> 
> * What happens when a user does not have credentials configured before importing this package.
> * What happens if the user ignores the default session and creates their own, then calls something that references [get_active_session](https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/session.py#L205)?
> * What happens if a user imports this package while using local testing mode?

These are all good points @sfc-gh-jrose. Do you think that they will be addressed after this other [PR](https://github.com/snowflakedb/snowflake/pull/193694) is merged? In this case, calling `getOrCreate` will actually get rather than `create`. Also, is there a way to detect if we are in a location testing mode, such that we can skip this workaround?

src/snowflake/snowpark/modin/pandas/__init__.py
sfc-gh-yzou: Yeah, i think the direction @sfc-gh-zyao is the right direction. Our original intention is to make the argument optional, not making the session object creation optional. I was talking to @sfc-gh-zyao , he mentioned he will work on this to get the fix in soon. I think we can just wait for this fix, instead of adding the workaround here, since it also took long time for the package to be released in stored sproc also.

sfc-gh-jrose:
> These are all good points @sfc-gh-jrose. Do you think that they will be addressed after this other [PR](https://github.com/snowflakedb/snowflake/pull/193694) is merged? In this case, calling `getOrCreate` will actually get rather than `create`. Also, is there a way to detect if we are in a location testing mode, such that we can skip this workaround?

With the code now gated by `is_in_stored_procedure` I don't think local testing is a concern.

PR #1911, opened 2024-07-12
SNOW-1533740: use SnowflakeCreateTable for cache result

sfc-gh-aalam:

   

   Fixes SNOW-1533740


sfc-gh-jdu (approved):
Can we add a test to check the executed query from cache_result?

tests/integ/scala/test_snowflake_plan_suite.py
```diff
@@ -208,15 +219,20 @@ def test_create_scoped_temp_table(session):
...
+                None,
+                None,
                 df._plan,
                 use_scoped_temp_objects=True,
                 is_generated=False,
             )
             .queries[0]
             .sql
-            == f' CREATE  TEMPORARY  TABLE {temp_table_name}("NUM" BIGINT, "STR" STRING(8))'
+            == f' CREATE  TEMPORARY  TABLE {temp_table_name} If  NOT  EXISTS ("NUM" BIGINT, "STR" STRING(8))'
```
sfc-gh-aalam, line 235: this is not a breaking change because 
1. in case of `cache_result`, we never called with with `is_generated=False`
2. when `is_generated=False` i.e. table creation is explicitely requested by users, then for save mode append, we add `if not exists` in the old behavior.

src/snowflake/snowpark/dataframe.py
```diff
@@ -3960,14 +3962,18 @@ def cache_result(
...
-            create_temp_table = self._session._analyzer.plan_builder.create_temp_table(
-                temp_table_name,
-                self._plan,
-                use_scoped_temp_objects=self._session._use_scoped_temp_objects,
-                is_generated=True,
+            df = self._with_plan(
+                SnowflakeCreateTable(
+                    [temp_table_name],
+                    None,
+                    SaveMode.APPEND,
```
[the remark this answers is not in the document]
sfc-gh-aalam: I added another branch to run a `get_create_and_insert_plan` when `is_generated=True`. When `is_generated=True` save mode is simply ignored. Would you suggest I just use `None` save mode here?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -723,9 +724,12 @@ def save_as_table(
         clustering_keys: Iterable[str],
         comment: Optional[str],
         child: SnowflakePlan,
+        source_plan: Optional[LogicalPlan],
+        use_scoped_temp_objects: bool,
+        is_generated: bool,
```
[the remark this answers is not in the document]
sfc-gh-aalam: will do

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -723,9 +724,12 @@ def save_as_table(
         clustering_keys: Iterable[str],
         comment: Optional[str],
         child: SnowflakePlan,
+        source_plan: Optional[LogicalPlan],
+        use_scoped_temp_objects: bool,
```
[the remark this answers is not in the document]
sfc-gh-aalam: it is an internal feature to reduce scope of temp tables inside a stored procedure not exposed to clients outside. see: https://docs.google.com/document/d/1Gjtw4AYhxlQaQu9z8fzJT5sfrPBYuHs6lL7KDSmixJA/edit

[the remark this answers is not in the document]
sfc-gh-aalam: outside of a stored proc, SCOPED keyword is a NOOP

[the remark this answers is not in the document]
sfc-gh-yzou: I see.

src/snowflake/snowpark/dataframe.py
```diff
@@ -3959,14 +3960,18 @@ def cache_result(
...
         else:
-            create_temp_table = self._session._analyzer.plan_builder.create_temp_table(
-                temp_table_name,
-                self._plan,
-                use_scoped_temp_objects=self._session._use_scoped_temp_objects,
-                is_generated=True,
+            df = self._with_plan(
+                SnowflakeCreateTable(
+                    [temp_table_name],
+                    None,
```
sfc-gh-yzou, line 3966: The SaveMode we passed here seems None, but when I look at the class init method for SnowflakeCreateTable, it seems it is not optional, which i think is correct, and we should have a save mode here. The SaveMode for SnowflakeCreateTable seems ERROR_IF_EXIST here, Let's make sure we have a SaveMode here. Also in the save_as_table function, we can add a check that only ERROR_IF_EXIST is supported for internally generated table. Then under  SaveMode == ERROR_IF_EXIST, you can add you if branch there to be clear

tests/integ/scala/test_snowflake_plan_suite.py
```diff
@@ -228,9 +229,15 @@ def test_create_scoped_temp_table(session):
         df = session.table(table_name)
         temp_table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
         assert (
-            session._plan_builder.create_temp_table(
-                temp_table_name,
+            session._plan_builder.save_as_table(
+                [temp_table_name],
+                None,
+                SaveMode.APPEND,
```
sfc-gh-yzou, line 235: This shouldn't be APPEND mode, right?

tests/integ/scala/test_snowflake_plan_suite.py
```diff
@@ -250,15 +263,21 @@ def test_create_scoped_temp_table(session):
...
+                None,
                 df._plan,
+                None,
                 use_scoped_temp_objects=True,
                 is_generated=False,
             )
             .queries[0]
             .sql
-            == f' CREATE  TEMPORARY  TABLE {temp_table_name}("NUM" BIGINT, "STR" STRING(8))'
+            == f' CREATE  TEMPORARY  TABLE {temp_table_name} If  NOT  EXISTS ("NUM" BIGINT, "STR" STRING(8))'
```
sfc-gh-yzou, line 280: if we do no go through append, do we still have this issue here

tests/integ/test_dataframe.py
```diff
@@ -1462,6 +1462,24 @@ def test_df_col(session):
     assert isinstance(c._expression, Star)
 
 
+@pytest.mark.xfail(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="Session.sql and DataFrame.cache_result not supported",
```
sfc-gh-yzou, line 1467: @sfc-gh-aling is that the state of today for cache_result?

tests/integ/scala/test_snowflake_plan_suite.py
```diff
@@ -239,9 +246,15 @@ def test_create_scoped_temp_table(session):
             == f' CREATE  SCOPED TEMPORARY  TABLE {temp_table_name}("NUM" BIGINT, "STR" STRING(8))'
         )
         assert (
-            session._plan_builder.create_temp_table(
-                temp_table_name,
+            session._plan_builder.save_as_table(
+                [temp_table_name],
+                None,
+                SaveMode.APPEND,
```
sfc-gh-yzou, line 252: same for teh mode here

tests/integ/scala/test_snowflake_plan_suite.py
sfc-gh-aalam: if we do not go through append, we will not have a create temp table statement. instead we will have CTAS statement.

tests/integ/test_dataframe.py
sfc-gh-aalam: there are two limitations in local testing which is why we cannot enable this test for local testing.

1. `session.sql` is not supported. We use the sql query to test that this sql is present in query history.
2. even if we replace `session.sql` with `session.create_dataframe` to create the test dataframe, query listeners are not supported in local testing since we do not build sql queries in local testing mode.

since this test is specifically testing the queries generated by cache_result, it should be fine to skip this test for local testing.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -741,9 +742,12 @@ def save_as_table(
...
+        is_generated: bool,  # true if the table is generated internally
     ) -> SnowflakePlan:
         full_table_name = ".".join(table_name)
```
sfc-gh-yzou, line 749: can we add a check at the very beginning that if is_generated is true, the SaveMode today must be ERROR_IF_EXISTS. Otherwise it is kind of wired that why other mode doesn't support is_generated.

sfc-gh-yzou: and also add a test that when using other mode with is_generated, it errors out today

tests/integ/test_dataframe.py
sfc-gh-yzou: From the above description, it seems the reason describe is not accurate, where the reason seems because session.sql, and session.query_history is not supported with local testing today, not because of cache_result.
How about let's create a dataframe use session.create_dataframe here, then if the local testing is able to support query_history, they can turn this test on. Also, update the reason to reflect the actual reason.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: I was wondering if we make is_generated compatible with ERROR_IF_EXISTS and APPEND. is_generated + APPEND will be used to run create and insert plan, and is_generated + ERROR_IF_EXISTS will be used to create CTAS plan

sfc-gh-yzou (approved):

PR #1914, opened 2024-07-15
SNOW-1491329 Support window functions in Snowpark IR

sfc-gh-lspiegelberg:
Support window functions/expressions in Snowpark IR.


src/snowflake/snowpark/window.py
```diff
@@ -225,7 +249,19 @@ def order_by(
                 elif isinstance(e._expression, Expression):
                     order_spec.append(SortOrder(e._expression, Ascending()))
 
-        return WindowSpec(self.partition_spec, order_spec, self.frame)
+        ast = proto.SpWindowSpecExpr()
+        window_ast = with_src_position(ast.sp_window_spec_order_by)
+        window_ast.wnd.CopyFrom(self._ast)
+        for e in exprs:
```
sfc-gh-lspiegelberg, line 255: there's some code duplication here, could use helper function. For now kept as is to not introduce complexity.

tests/ast/data/windows.test
```diff
@@ -0,0 +1,71 @@
...
+
+df15 = df.select(avg("value").over(Window.rangeBetween(Window.currentRow, Window.unboundedPreceding)))
+
+## EXPECTED OUTPUT
+
+res1 = session.table("test_table")
+
+res2 = res1.select(col("NUM").as_("key"), col("STR").as_("value"))
+
+res3 = res2.select(avg(col("value")).over(Window.partition_by(col("value")).order_by(col("key")).rows_between(0, 2)).as_("window1"), avg(col("value")).over(Window.order_by(col("key").desc()).range_between(-9223372036854775807, 9223372036854775807)).as_("window2"))
```
sfc-gh-lspiegelberg, line 47: Note here that the constants `Window.CURRENT_ROW` are internally labels for integers. I left a comment here https://github.com/snowflakedb/snowflake/pull/194828/files#r1679926588 to discuss strategies.

src/snowflake/snowpark/column.py
```diff
@@ -923,14 +923,19 @@ def name(self, alias: str, variant_is_as: bool = None) -> "Column":
...
         Returns a window frame, based on the specified :class:`~snowflake.snowpark.window.WindowSpec`.
         """
         if not window:
             window = Window._spec()
-        return window._with_aggregate(self._expression)
+
+        ast = proto.Expr()
+        expr = with_src_position(ast.sp_column_over)
+        expr.col.CopyFrom(self._ast)
+        expr.window_spec.CopyFrom(window._ast if window else WindowSpec())
```
sfc-gh-azwiegincew, line 936: Now that WindowSpecs are exprs, we no longer need to use CopyFrom. Each WindowSpec can generate its own Assign() and here we can use `SpWindowSpecRef(id)`. Alternatively, if you want to keep this model, then the AST should go back to using a `data` decl describing a sum type.

sfc-gh-azwiegincew (changes requested):

src/snowflake/snowpark/column.py
sfc-gh-lspiegelberg: Hi, following up on:

> Now that WindowSpecs are exprs, we no longer need to use CopyFrom. Each WindowSpec can generate its own Assign() and here we can use SpWindowSpecRef(id). Alternatively, if you want to keep this model, then the AST should go back to using a data decl describing a sum type.

this would mean for
```
window1 = Window.partition_by("value").order_by("key").rows_between(Window.CURRENT_ROW, 2)
df.select(avg("value").over(window1))
```
we would encode this as
```
Assign(uid=1, var_id=1, expr=SpWindowSpecRowBetween(wnd=SpWindowSpecOrderBy(wnd=SpWindowSpecPartitionBy(wnd=SpWindowSpec(),...), ...) , ...))

Assign(..., expr=SpDataFrameSelect(..., expr=SpColumnOver(SpWindowSpecRef(id=1)))
```
?

Or do you have in mind here
```
Assign(uid=1000, var_id=1, expr=SpWindowSpec())
Assign(uid=1001, var_id=2, expr=SpWindowSpecPartitionBy(wnd=SpWindowSpecRef(id=1), ...))
...
```
?

sfc-gh-lspiegelberg: discussed offline: Solution is to keep it aligned with `Column`, however `SpWindowSpec` should not be an `Expr` but rather inherit the SrcPosition trait. Will make changes accordingly.

sfc-gh-azwiegincew (approved):

PR #1916, opened 2024-07-16
SNOW-1458130 Minor documentation fix for `Series.sort_values`

sfc-gh-vbudati:

   

   Fixes SNOW-1458130

   Added missing parameter `key` to the respective column and also add details about how sorting works.


docs/source/modin/supported/series_supported.rst
```diff
@@ -379,7 +379,8 @@ Methods
 | ``sort_index``              | P                               |                                  | ``N`` if given the ``key`` param,                  |
 |                             |                                 |                                  | or MultiIndex                                      |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``sort_values``             | P                               |                                  | ``N`` if given the ``key`` parameter               |
+| ``sort_values``             | P                               | ``key``                          | Snowpark pandas currently uses stable sort when    |
```
sfc-gh-helmeleegy, line 382: What about DataFrame.sort_values?

sfc-gh-vbudati: Should be the same for DataFrame.sort_values, did not think of that. I updated the documentation for that as well.

sfc-gh-helmeleegy (approved):
LGTM. Thanks!

docs/source/modin/supported/series_supported.rst
sfc-gh-helmeleegy: Great.

docs/source/modin/supported/dataframe_supported.rst
```diff
@@ -390,7 +390,9 @@ Methods
 | ``sort_index``              | P                               |                                  | ``N`` if given the ``key`` param. ``N`` if         |
 |                             |                                 |                                  | ``axis == 1``, or MultiIndex.                      |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``sort_values``             | P                               |                                  | ``N`` if given the ``key`` param or ``axis == 1``  |
+| ``sort_values``             | P                               | ``key``                          | ``N`` if given the ``key`` param or ``axis == 1``. |
+|                             |                                 |                                  | Snowpark pandas currently uses stable sort when    |
```
sfc-gh-mvashishtha, line 394: Some corrections:

- "when sorting the index values" doesn't seem to apply since this is about `DataFrame`. Anyway, I would delete this fragment because it's redundant.
- This description makes it sound like "stable sort" is an algorithm, but it's a description of an algorithm.
- We should mention that Snowpark pandas ignores the `kind` parameter completely. We currently warn `choice of sort algorithm 'quicksort' is ignored. sort kind must be 'stable' or None`.

So I suggest something like 

`The ``kind`` paramater has no effect. Snowpark pandas always uses a stable sort algorithm, while pandas by default does not.`

docs/source/modin/supported/series_supported.rst
sfc-gh-mvashishtha, line 382: ditto

sfc-gh-mvashishtha (approved):
thank you!

PR #1928, opened 2024-07-17
[SNOW-1533733] Add execution_queries property for SnowflakePlan to get the final query will be sent over to server for execution

sfc-gh-yzou:
SNOW-1533733

   - [x] This is a pure refactoring change, relying on the existing testing to be finished.

With https://docs.google.com/document/d/1DKFgGOHSTuI8gvHOaU2Xj_wzfFvMSPuKKtLgDxfvNeA/edit?pli=1, we are separating queries used for final execution and queries for intermediate process, and the queries for final execution will have optimizations applied.

In this pr, we add a new property execution_queries for the SnowflakePlan, which is used to get the final query will be sent over to server for execution, the returned exuecution_queries will have optimization applied if the optimization is enabeld.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -250,6 +250,14 @@ class SnowflakePlan(LogicalPlan):
 
     @property
     def execution_queries(self) -> Dict["PlanQueryType", List["Query"]]:
+        """
+        Get the list of queries that will be sent over to server evaluation. The queries
+        have optimizations applied of optimizations are enabled.
+
+        Returns
+        -------
+        A mapping between the PlanQueryType and the list of Queries corresponds to the type.
+        """
         # apply optimizations
         final_plan = self.replace_repeated_subquery_with_cte()
         return {
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -254,7 +254,7 @@ class SnowflakePlan(LogicalPlan):
         final_plan = self.replace_repeated_subquery_with_cte()
         return {
             PlanQueryType.QUERIES: final_plan.queries,
-            PlanQueryType.POST_ACTIONS: final_plan.post_actions
+            PlanQueryType.POST_ACTIONS: final_plan.post_actions,
         }
 
     @property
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -247,6 +248,23 @@ def __eq__(self, other: "SnowflakePlan") -> bool:
     def __hash__(self) -> int:
         return hash(self._id) if self._id else super().__hash__()
 
+    @property
```
sfc-gh-aling, line 251: nit: do you consider `cache_property` for slight perf improvement? what if every call needs a re-calculation then it doesn't apply

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -561,21 +562,23 @@ def get_result_set(
     ]:
         action_id = plan.session._generate_new_action_id()
         # potentially optimize the query using CTEs
-        plan = plan.replace_repeated_subquery_with_cte()
+        # plan = plan.replace_repeated_subquery_with_cte()
```
sfc-gh-aling, line 565: I guess we don't need this comment
```suggestion
```

sfc-gh-aling (approved):
local testing change LGTM

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: oh, that can not be a cache_property here, because optimizations will be applied based on configurations, and the queries generated can be different based on the optimization configuration at the time when this property is called. 
For example:
session.cte_optimizationz_enabled = False
queries1 = df._plan.execution_queries
session.cte_optimizationz_enabled = True
queries2= df._plan.execution_queries

queries1 and queries2 are expected to be different here. If we use cached_property, queries1 and queries2 will be the same, which will not be correct.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-yzou: removed

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aling: I see, thanks for the explanation!

sfc-gh-aalam (approved):

sfc-gh-jdu (approved):

PR #1930, opened 2024-07-17
Dataframe: Natural join, cross join, join

sfc-gh-oplaton:

   

   Fixes SNOW-1491303

   Implement AST generation for `Dataframe.join`, `Dataframe.cross_join` and `Dataframe.natural_join`


src/snowflake/snowpark/dataframe.py
```diff
@@ -2245,14 +2256,30 @@ def natural_join(
...
+        join_type = create_join_type(kwargs.get("join_type") or how or "inner")
         join_plan = Join(
             self._plan,
             right._plan,
-            NaturalJoin(create_join_type(join_type or "inner")),
+            NaturalJoin(join_type),
             None,
             None,
         )
+
```
sfc-gh-lspiegelberg, line 2267: somewhere here `self._ast` seems to be missing.  I.e., with
```
    def set_ast_ref(self, sp_dataframe_expr_builder: Any) -> None:
        """
        Given a field builder expression of the AST type SpDataframeExpr, points the builder to reference this dataframe.
        """
        # TODO: remove the None guard below once we generate the correct AST.
        if self._ast_id is not None:
            sp_dataframe_expr_builder.sp_dataframe_ref.id.bitfield1 = self._ast_id
```
I would expect to have the id of `self` being carried over to `ast` in some form. Put simply, intuitively the logic here should be something like
```
ast.lhs = self._ast
ast.rhs = other._ast
```
maybe I am missing something?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2245,14 +2256,30 @@ def natural_join(
...
+            ast.join_type.sp_join_type__left_outer = True
+        elif isinstance(join_type, RightOuter):
+            ast.join_type.sp_join_type__right_outer = True
+        elif isinstance(join_type, FullOuter):
+            ast.join_type.sp_join_type__full_outer = True
+        else:
+            raise ValueError(f"Unsupported join type {join_type}")
+
         if self._select_statement:
             select_plan = self._session._analyzer.create_select_statement(
```
sfc-gh-azwiegincew, line 2284: I think the problem is here. Can you check that the return value has your AST?

sfc-gh-azwiegincew: My suspicion is that you need to call `create_selet_statement({current args}, stmt)`

sfc-gh-oplaton: Hmm. This location is in `Dataframe.natural_join`, which currently passes the tests. I'm having issues with `Dataframe.join`, where I actually made a change similar in spirit to this suggestion.

sfc-gh-oplaton: As we discovered in person, the problem was with the `Dataframe.show()` calls.

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: Debugging with Arthur, we found that my test case included calls to `Dataframe.show()` because I lifted the tests directly from documentation examples.

tests/ast/data/DataFrame.join.inner.predicate_rsuffix.test
```diff
@@ -0,0 +1,17 @@
...
+
+df3 = df1.join(df2, (df1["num"] < df2["num"]) & (df1["str"] == df2["str"]), rsuffix="_2").select('*')
+
+## EXPECTED OUTPUT
+
+res1 = session.table("test_table")
+
+res2 = session.table("test_table")
+
+res3 = res1.join(res2, how=(res1["num"] < res2["num"] & res1["str"] == res2["str"]), on="inner")
```
sfc-gh-vbudati, line 15: I think this is wrong -- `how` and `on` are switched?

sfc-gh-oplaton: Good catch, thank you! This was a bug in the unparser. Updated the unparser changes and the expectation tests.

tests/ast/data/DataFrame.join.inner.column.test
```diff
@@ -0,0 +1,17 @@
...
+
+df3 = df1.join(df2, "num").select(df1.num.alias("n1"), df1.str, df2.str)
+
+## EXPECTED OUTPUT
+
+res1 = session.table("test_table")
+
+res2 = session.table("test_table")
+
+res3 = res1.join(res2, on=("num"), how="inner")
```
sfc-gh-azwiegincew, line 15: We should avoid generating the parentheses around the `on` expression. Would you mind creating a task or just adding a TODO, please?

sfc-gh-azwiegincew (approved):

tests/ast/data/DataFrame.join.inner.column.test
sfc-gh-oplaton: Resolved in https://github.com/snowflakedb/snowflake/pull/197145

PR #1935, opened 2024-07-17
[SNOW-1458121]: Add support for `Index.reindex`.

sfc-gh-rdurrani:

   

   Fixes SNOW-1458121


src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1356,12 +1362,67 @@ def reindex(self) -> None:
         ValueError
             If non-unique index and ``method`` or ``limit`` passed.
 
+        Notes
+        -----
+        If the index is lazy, ``method=nearest`` is not supported.
+
+        If the index is lazy, and duplicate values are present,
```
sfc-gh-rdurrani, line 1369: We could match pandas behavior here by adding another column that indicates if there are duplicate values in the existing index, but I figured that that would add additional computation that isn't super necessary, although happy to change this if folks disagree?

tests/integ/modin/index/test_reindex.py
```diff
@@ -0,0 +1,269 @@
...
+import pytest
+
+import snowflake.snowpark.modin.plugin  # noqa: F401
+from snowflake.snowpark.exceptions import SnowparkSQLException
+from tests.integ.modin.sql_counter import SqlCounter, sql_count_checker
+from tests.integ.modin.utils import assert_index_equal
+
+
+def assert_reindex_result_equal(
+    snow_result: tuple[pd.Index, np.ndarray], native_result: tuple[pd.Index, np.ndarray]
```
sfc-gh-mvashishtha, line 19: should the first element of `native_result` be `native_pd.Index`?

tests/integ/modin/index/test_reindex.py
```diff
@@ -0,0 +1,269 @@
...
+    excp_match: str,
+) -> None:
+    with pytest.raises(excp_type, match=excp_match):
+        func(native_idx)
+
+    with pytest.raises(excp_type, match=excp_match):
+        func(snow_idx)
+
+
+def create_datetime_index(snow_series: pd.Series) -> pd.Index:
```
sfc-gh-mvashishtha, line 39: Ideally we should not be manipulating Snowpark pandas internals in this test. I think we can delete this method and use `pd.Index(snow_series)` everywhere we need a datetime index.

tests/integ/modin/index/test_reindex.py
```diff
@@ -0,0 +1,269 @@
...
+                data_column_snowflake_quoted_identifiers=[],
+                index_column_pandas_labels=[None],
+                index_column_snowflake_quoted_identifiers=modin_frame.data_column_snowflake_quoted_identifiers,
+            )
+        )
+    )
+
+
+@sql_count_checker(query_count=0)
+def test_reindex_invalid_limit_parameter():
```
sfc-gh-mvashishtha, line 61: nit: I would remove `reindex` from all these test method names. The file is called `test_reindex` already. I'd also remove `reindex_index` from some of the method names that include that fragment.

tests/integ/modin/index/test_reindex.py
```diff
@@ -0,0 +1,269 @@
...
+        with SqlCounter(query_count=2, join_count=2):
+            native_result = native_idx.reindex(
+                list("CADEFG"), method=method, limit=limit
+            )
+            snow_result = snow_idx.reindex(list("CADEFG"), method=method, limit=limit)
+            assert_reindex_result_equal(snow_result, native_result)
+
+
+@sql_count_checker(query_count=2, join_count=2)
+def test_reindex_index_ordered_index_unordered_new_index():
```
sfc-gh-mvashishtha, line 147:
```suggestion
def test_sorted_index_reindexed_to_sorted_new_index():
```

tests/integ/modin/index/test_reindex.py
```diff
@@ -0,0 +1,269 @@
...
+            )
+        )
+    )
+
+
+@sql_count_checker(query_count=0)
+def test_reindex_invalid_limit_parameter():
+    native_idx = native_pd.Index(list("ABC"))
+    snow_idx = pd.Index(native_idx)
+    assert_reindex_exceptions_match(
```
sfc-gh-mvashishtha, line 64: here and below, does `eval_snowpark_pandas_result` work? If not, can we extend it or create a separate test method for `Index`? That will reduce code duplication and reduce the likelihood that we make an error in trying to apply the same function to both a Snowpark pandas index and the corresponding native pandas index.

tests/integ/modin/index/test_reindex.py
```diff
@@ -0,0 +1,269 @@
...
+        native_pd.date_range("12/29/2008", periods=10, freq="D")
+    )
+    snow_result = snow_date_index.reindex(
+        pd.date_range("12/29/2008", periods=10, freq="D")
+    )
+    assert_reindex_result_equal(snow_result, native_result)
+
+
+@sql_count_checker(query_count=0)
+def test_reindex_index_non_overlapping_different_types_index_negative():
```
sfc-gh-mvashishtha, line 203: does this exception match pandas's?

docs/source/modin/supported/index_supported.rst
```diff
@@ -109,7 +109,8 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``max``                     | N                               |                                  |                                                    |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``reindex``                 | N                               |                                  |                                                    |
+| ``reindex``                 | P                               |                                  | ``N`` if the Index is MultiIndex and lazy, or      |
```
sfc-gh-mvashishtha, line 112: Should we also add a note that while pandas will raise an error for reindexing an index with duplicate values, Snowpark pandas will not and will instead do a left join between the two indexes?

sfc-gh-mvashishtha (commented):
Leaving comments on just the tests and the documentation for now.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2245,7 +2245,64 @@ def _reindex_axis_0(
         modin_frame = self._modin_frame
         method = kwargs.get("method", None)
         fill_value = kwargs.get("fill_value", np.nan)  # type: ignore[arg-type]
+        limit = kwargs.get("limit", None)
         _filter_column_snowflake_quoted_id = None
+        is_index = kwargs.get("_is_index", False)
+        if is_index:
+            modin_frame = modin_frame.ensure_row_position_column()
+            row_position_column = modin_frame.row_position_snowflake_quoted_identifier
+            modin_frame = InternalFrame.create(
```
sfc-gh-mvashishtha, line 2254: can't we use `append_column` to add the row position column as well?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2279,6 +2336,33 @@ def _reindex_axis_0(
                 modin_frame.data_column_snowflake_quoted_identifiers
             )
         )
+        if is_index:
+            # We want to remove the monotonic_increasing and monotonic_decreasing columns so that they will not be filled here.
```
sfc-gh-mvashishtha, line 2340: I'm confused about what the internal frame we're creating here is used for if it doesn't have the `monotonic_increasing` or `monotonic_decreasing` columns. Could you add a comment explaining what it's used for and how it relates to `new_qc`?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2376,7 +2459,72 @@ def _reindex_axis_0(
                             index_column_snowflake_quoted_identifiers=new_qc._modin_frame.index_column_snowflake_quoted_identifiers,
                         )
                     )
-
+        if is_index:
+            modin_frame = new_qc._modin_frame
+            # We need to get the new quoted identifier after filling happens.
+            row_position_column = modin_frame.data_column_snowflake_quoted_identifiers[
+                modin_frame.data_column_pandas_labels.index("indices")
```
sfc-gh-mvashishtha, line 2466: Again, I'd prefer to access `"indices"` by its position (is that `data_column_pandas_labels[-1]`?) rather than by value.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2376,7 +2459,72 @@ def _reindex_axis_0(
...
+                    monotonic_decreasing_snowflake_quoted_id,
+                    monotonic_increasing_snowflake_quoted_id,
+                ]
+            ).to_pandas()
+            monotonic_decreasing = materialized_frame[
+                monotonic_decreasing_snowflake_quoted_id.strip('"')
+            ]
+            monotonic_increasing = materialized_frame[
+                monotonic_increasing_snowflake_quoted_id.strip('"')
+            ]
```
sfc-gh-mvashishtha, line 2502:
```suggestion
            monotonic_decreasing = materialized_frame.iloc[:, 1]
            monotonic_increasing = materialized_frame.iloc[:, 2]
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2376,7 +2459,72 @@ def _reindex_axis_0(
...
+                    )
+                    and any_overlap
+                ):
+                    method_str = {"bfill": "backfill", "ffill": "pad"}.get(
+                        method, method  # type: ignore[call-overload]
+                    )
+                    raise ValueError(
+                        f"limit argument for '{method_str}' method only well-defined if index and target are monotonic"
+                    )
+            return new_qc, materialized_frame[row_position_column.strip('"')].values  # type: ignore[return-value]
```
sfc-gh-mvashishtha, line 2527:
```suggestion
            return new_qc, materialized_frame.iloc[:, 0].values  # type: ignore[return-value]
```

sfc-gh-mvashishtha (commented):
I went through all the files

tests/integ/modin/index/test_reindex.py
sfc-gh-rdurrani: This name is wrong, since the new index is unsorted.

tests/integ/modin/index/test_reindex.py
sfc-gh-rdurrani: Makes sense - for the ones its expected to error, I can use eval_snowpark_pandas_result - for the ones where it returns, I'll still use my helper method, but I'll change it to use assert_index_equals from the utils so there's less code duplication! (The helper method is necessary since the result of reindex is a tuple containing an index and an indexing array). To resolve the function issue, I'll pass in the function to the function as a lambda, similar to eval_snowpark_pandas_result!

tests/integ/modin/index/test_reindex.py
sfc-gh-rdurrani: This is an exception on our end since we're trying to join columns of different types - pandas doesn't error out, and instead returns the correct result:
```python
In [3]: pd.date_range("1/1/2010", periods=6, freq="D").reindex(list("CAB"))
Out[3]: (Index(['C', 'A', 'B'], dtype='object'), array([-1, -1, -1]))
```

docs/source/modin/supported/index_supported.rst
sfc-gh-rdurrani: I added that note to the docstring - usually behavior change notes go in the docstring?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: append_column will add a duplicate physical row order column - I'd rather not add the additional column to the physical object.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: We need the monotonic increasing and decreasing columns physically in the data in order to use it later on, but we don't want them filled by the fillna column. To keep them around after the join, they need to be in the data columns for the internal frame though, as the join purges non active columns from the current view of the snowpark dataframe (still physically present, but no longer accessible from the current view). We basically just need them to check at the very end if the original data was monotonic increasing or decreasing. Will go ahead and add this explanation to the comment as well!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2279,6 +2336,33 @@ def _reindex_axis_0(
...
+            data_column_pandas_labels.remove("_index_lag_col")
+            data_column_pandas_labels.remove("_is_monotonic_decreasing")
+            data_column_pandas_labels.remove("_is_monotonic_increasing")
+            data_column_snowflake_quoted_identifiers.remove(lag_col_snowflake_quoted_id)
+            data_column_snowflake_quoted_identifiers.remove(
+                monotonic_decreasing_snowflake_quoted_id
+            )
+            data_column_snowflake_quoted_identifiers.remove(
+                monotonic_increasing_snowflake_quoted_id
+            )
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: We can't remove them positionally, since we don't know if they're in the last position - if the user called reindex with filling, then we also added another column to the frame - the column that lets us know if there are any na values present in the original data so that we do not fill them. We could add additional logic here to figure out what the indices are, but I figured that that would be confusing, and explicitly removing by name is clearer and less likely to cause confusion to others reading the code. As for the _index_lag_col, I can add some label de-duplication?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: Again, feels safer to index by name just in case.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: Okay, I think the comments help clarify what is happening now.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: Oh, I didn't see that we sometimes add `dummy_reindex_column_for_fill`. I'm okay with accessing by label here in that case to reduce code complexity.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: Resolved [here](https://github.com/snowflakedb/snowpark-python/pull/1935/files#r1687118143).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: In this case we know exactly which column we're selecting at each position, so we should use positions instead of making assumptions about how we quote the identifiers.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: Ditto from [here](https://github.com/snowflakedb/snowpark-python/pull/1935/files#r1684761263): I think in this case we really should use positions.

tests/integ/modin/index/test_reindex.py
sfc-gh-mvashishtha: Marking this as unresolved since I don't think you've changed the test names yet.

tests/integ/modin/index/test_reindex.py
sfc-gh-mvashishtha: At commit c5c80043c49f455fbd79816dc36055b9eefb76eb, `pd.Index(pd.date_range("1/1/2010", periods=6, freq="D"))` is giving me a lazy `snowflake.snowpark.modin.plugin.extensions.index.Index` 

Alternatively we can do `pd.date_range("1/1/2010", periods=6, freq="D").to_frame().set_index(0)`, which seems to use fewer queries. Note that the name is 0 because for reasons I don't understand (maybe a pandas bug), `native_pd.Series().to_frame()` and `pd.Series().to_frame()` give the column name 0.

tests/integ/modin/index/test_reindex.py
sfc-gh-mvashishtha: That's correct, so can we do this?

```suggestion
def test_unsorted_index_reindexed_to_sorted_new_index():
```

I couldn't parse the current function name, and I think you mean to talk about sortedness, not orderedness, because `Index` is always ordered.

tests/integ/modin/index/test_reindex.py
sfc-gh-mvashishtha: Could you file an issue and link it here?

tests/integ/modin/index/test_reindex.py
sfc-gh-rdurrani: Hmm - seems like the bug was fixed! I'll go ahead and remove this workaround!

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1356,12 +1362,70 @@ def reindex(self) -> None:
...
+                "limit argument only valid if doing pad, backfill or nearest reindexing"
+            )
+        kwargs = {
+            "method": method,
+            "level": level,
+            "limit": limit,
+            "tolerance": tolerance,
+            "_is_index": True,
+        }
+        try:
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: This actually isn't a bug - it matches pandas behavior (although pandas keeps the objects as tuples rather than lists):

```python
In [2]: a = native_pd.MultiIndex.from_tuples([(1, 2), (3, 4)])

In [3]: pd.Index(a)
Out[3]: Index([[1, 2], [3, 4]], dtype='object')

In [4]: pd.Index(a).levels
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 pd.Index(a).levels

File ~/Documents/snowpark-python/src/snowflake/snowpark/modin/plugin/extensions/index.py:234, in Index.__getattr__(self, key)
    228     if hasattr(native_index, key):
    229         # Any methods that not supported by the current Index.py but exist in a
    230         # native pandas index object should raise a not implemented error for now.
    231         raise ErrorMessage.not_implemented(
    232             f"Index.{key} is not yet implemented"
    233         )
--> 234 raise err

File ~/Documents/snowpark-python/src/snowflake/snowpark/modin/plugin/extensions/index.py:224, in Index.__getattr__(self, key)
    206 """
    207 Return item identified by `key`.
    208
   (...)
    221 of current scope that are not implemented.
    222 """
    223 try:
--> 224     return object.__getattribute__(self, key)
    225 except AttributeError as err:
    226     if not key.startswith("_"):

AttributeError: 'Index' object has no attribute 'levels'

In [5]: pd.Index(a)._query_compiler
Out[5]: <snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler.SnowflakeQueryCompiler at 0x14a6f51b0>

In [6]: pd.Index(a)._query_compiler._modin_frame
Out[6]: InternalFrame(ordered_dataframe=<snowflake.snowpark.modin.plugin._internal.ordered_dataframe.OrderedDataFrame object at 0x14a6f5660>, label_to_snowflake_quoted_identifier=(LabelIdentifierPair(label=(None,), snowflake_quoted_identifier='"__index__"'),), num_index_columns=1, data_column_index_names=((None,),))

In [7]: pd.Index(a)._query_compiler._modin_frame.ordered_dataframe
Out[7]: <snowflake.snowpark.modin.plugin._internal.ordered_dataframe.OrderedDataFrame at 0x14a6e64a0>

In [8]: pd.Index(a)._query_compiler._modin_frame.ordered_dataframe.to_pandas()
Out[8]:
         __index__  __row_position__
0  [\n  1,\n  2\n]                 0
1  [\n  3,\n  4\n]                 1

In [9]: native_pd.Index(a)
Out[9]: Index([(1, 2), (3, 4)], dtype='object')

In [10]: native_pd.Index(a).levels
AttributeError                            Traceback (most recent call last)
Cell In[10], line 1
----> 1 native_pd.Index(a).levels

AttributeError: 'Index' object has no attribute 'levels'
```
I could remove the try/except, but I figured that this was a neater Exception to throw? I could also add a check that checks if the type of the index column is an ArrayType() and then throw an Exception without entering the QC if that works better?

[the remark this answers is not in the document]
sfc-gh-mvashishtha: this looks better

sfc-gh-mvashishtha (approved):
LGTM, thanks for responding to my comments!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2237,7 +2237,55 @@ def _reindex_axis_0(
...
+            modin_frame = modin_frame.append_column(
+                "_is_monotonic_decreasing",
+                coalesce(
+                    min_(
+                        col(index_col_snowflake_quoted_id)
+                        < col(lag_col_snowflake_quoted_id)
+                    ).over(),
+                    pandas_lit(False),
+                ),
+            )
```
sfc-gh-nkrishna, line 2272: I think it would be clearer to put this code into the `is_monotic_increasing/decreasing` functions in `index.py` -- that way we can also basically implement those in this PR since the work is being done here anyways.

sfc-gh-nkrishna: Discussed offline: Rehan is going to create a helper method that can later be used in monotonically increasing/decreasing functions once those are implemented in the future.

sfc-gh-nkrishna (approved):
LGTM!

PR #1937, opened 2024-07-19
[SNOW-1541087] Allow deep copy of SnowflakePlan and Selectable

sfc-gh-yzou:
SNOW-1541087

The SnowflakePlan and Selectable has analyzer and session associated with it, which can not be copied today. In order to enable the deepcopy of those node, we overwrites the __deepcopy__ method to to copy the reference of the analyzer and session.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -214,6 +214,31 @@ class Selectable(LogicalPlan, ABC):
     def __hash__(self) -> int:
         return hash(self._id) if self._id else super().__hash__()
 
+    def __deepcopy__(self, memodict={}) -> "Selectable":  # noqa: B006
+        copied_selectable = Selectable(
+            # analyze contains session object, which cannot be deep copied. Be-careful when
+            # using the deep copied selectable entry.
+            analyzer=self.analyzer,
+            api_calls=deepcopy(self._api_calls),
+        )
+        copied_selectable.deepcopy_fields(self)
+
+        return copied_selectable
+
+    def deepcopy_fields(self, from_obj: "Selectable") -> None:  # noqa: B006
+        self.pre_actions = deepcopy(from_obj.pre_actions)
+        self.post_actions = deepcopy(from_obj.post_actions)
+        self.flatten_disabled = from_obj.flatten_disabled
+        self._column_states = deepcopy(from_obj._column_states)
+        self._snowflake_plan = deepcopy(from_obj._snowflake_plan)
+        self.expr_to_alias = deepcopy(from_obj.expr_to_alias)
+        self.df_aliased_col_name_to_real_col_name = deepcopy(
+            from_obj.df_aliased_col_name_to_real_col_name
+        )
+        self._cumulative_node_complexity = deepcopy(
+            from_obj._cumulative_node_complexity
+        )
+
     @property
     @abstractmethod
     def sql_query(self) -> str:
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -181,6 +181,27 @@ def __setitem__(self, col_name: str, col_state: ColumnState) -> None:
...
+def _deepcopy_selectable_fields(
+    from_selectable: "Selectable", to_selectable: "Selectable"
+) -> None:
+    """
+    Make a deep copy of the fields from the from_selectable to the to_selectable
+    """
+    to_selectable.pre_actions = deepcopy(from_selectable.pre_actions)
+    to_selectable.post_actions = deepcopy(from_selectable.post_actions)
+    to_selectable.flatten_disabled = from_selectable.flatten_disabled
+    to_selectable._column_states = deepcopy(from_selectable._column_states)
```
sfc-gh-aalam, line 193: do we test if we trigger a `get_attributes` call here?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -420,6 +420,32 @@ def __copy__(self) -> "SnowflakePlan":
                 placeholder_query=self.placeholder_query,
             )
 
+    def __deepcopy__(self, memodict={}) -> "SnowflakePlan":  # noqa: B006
+        copied_plan = SnowflakePlan(
+            queries=copy.deepcopy(self.queries) if self.queries else [],
```
sfc-gh-aalam, line 425: SnowflakePlan also has some cached_properties like `attributes` which run a describe query once and then reuse the result from subsequent calls. During this deepcopy, we do not borrow those results over. I'm wondering if that can cause any issues down the line.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yzou: With integ test, i don't see an easy way to see if the describing call is triggered because the describing call doesn't show up in the query history. However, I have manually verified that the describing call is not triggered during the deepcopy process

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: The attribute result is get from the schema call, during the copy, we make sure the schema queries are copied correctly, so the worst case is that a if the attribute call is called on the copied plan, a describing call will be triggered.  In our test, we do check the schema query is copied correctly. 
Theoretically, all those properties should be able to be reconstructed correctly if all fields of the class is copied correctly. 
We can also make a copy of all the cached_property if exists, which is probably not necessary now, since the deep copy is currently only used by our new compilation stage, and none of our optimization step should require the attribute property.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -409,6 +436,17 @@ def __init__(
             self._schema_query = sql
             self._query_param = params
 
+    def __deepcopy__(self, memodict={}) -> "SelectSQL":  # noqa: B006
+        copied = SelectSQL(
```
sfc-gh-jdu, line 440: Will it trigger an extra describe call, when convert_to_select is true and the original sql is a non-select sql?

sfc-gh-jdu (commented):
Have you run it with disabling sql simplifier?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yzou: That is a good point. Yes, it will when convert_to_select is True, updated the implementation to use false first, and then copy values over

sfc-gh-yzou:
@sfc-gh-jdu i have tried to run the new test suite with both disable_sql_simplifier True and False, all passed. triggering a daily action here https://github.com/snowflakedb/snowpark-python/actions/runs/10050143904

sfc-gh-aalam (approved):

sfc-gh-jdu (approved):

PR #1940, opened 2024-07-19
SNOW-1545615: Change snowflake.snowpark.modin.pandas to modin.pandas in DF/Series doc generation

sfc-gh-joshi:

   

   Fixes SNOW-1545615

This PR adds extra customization to allow sphinx documentation to use `modin.pandas` as the defined module rather than `snowflake.snowpark.modin.pandas`, in anticipation of the removal of the vendored copies of our DataFrame and Series classes.

Currently, our rst files only reference `modin.pandas` for the `StringMethods` and `DatetimeProperties` classes, which are taken from upstream modin. The other `currentmodule `directives referencing `snowflake.snowpark.modin.pandas` would indirectly trigger an import of `snowflake.snowpark.modin.plugin`, which implicitly initialized the extension and set the correct docstrings. Now, with the absence of this `currentmodule` directives in `Series.rst`, we need to properly initialize the plugin to set the modin namespace correctly from within `conf.py`.

Unfortunately, this creates complications in documentation generation:
1. Sphinx's `autosummary-generate` script, responsible for generating stub .rst files (those found in `docs/source/pandas_api/`) recognizes `Series.str` and `Series.dt` as property objects rather than classes when trying to access properties like `Series.str.capitalize`, and errors out when trying to import them. It is unclear why this was not previously the case.
    - **Solution:** Within the `setup` function of `conf.py`, which runs before `autosummary-generate`, we replace `Series.str`/`dt` with the actual `StringMethods`/`DatetimeProperties` classes, allowing sphinx to correctly access their namespaces. However...
2. After monkeypatching `str`/`dt` to fix accesses to `Series.str.*` and `Series.dt.*`, sphinx now cannot document the top-level properties `Series.str` and `Series.dt`.
    - **Solution:** We create a custom template and directive (`modin_accessor.rst`, `automodinaccessor`) that controls name resolution for these top-level properties.

The change to the signature of pd.read_excel is necessary to circumvent an ambiguous reference warning, as for some reason sphinx cannot resolve whether the `DataFrame` in the return type is a `modin.pandas.DataFrame`, `snowflake.snowpark.modin.pandas.DataFrame`, or `snowpark.DataFrame`.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -148,9 +148,17 @@ class ModinAccessorDocumenter(PropertyDocumenter):
     priority = 0.55
 
     def import_object(self, raiseerror=False):
+        # Set `self.object` and related fields after importing the object, since sphinx has difficulty
+        # trying to import the top-level Series.str and Series.dt objects.
+        # Returns True if the object was successfully imported.
+        # See definition on parent classes:
+        # https://github.com/sphinx-doc/sphinx/blob/907d27dc6506c542c11a7dd16b560eb4be7da5fc/sphinx/ext/autodoc/__init__.py#L2714
+        # https://github.com/sphinx-doc/sphinx/blob/907d27dc6506c542c11a7dd16b560eb4be7da5fc/sphinx/ext/autodoc/__init__.py#L400
         import modin.pandas as pd
         self.module = pd
         self.parent = pd.Series
+        # objpath is an array like ["Series", "str"]
+        # object_name should be the name of the property (in this case "str")
         self.object_name = self.objpath[-1]
         self.object = getattr(pd.Series, self.object_name)
         self.isclassmethod = False
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -144,8 +144,8 @@ class ModinAccessorDocumenter(PropertyDocumenter):
     objtype = "modinaccessor"
     directivetype = "attribute"
 
-    # lower priority than the other custom attribute documenter (defined below)
-    priority = 0.55
+    # lower priority than the default PropertyDocumenter so it is not chosen for normal properties
+    priority = 0.6
 
     def import_object(self, raiseerror=False):
         # Set `self.object` and related fields after importing the object, since sphinx has difficulty
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -123,13 +123,46 @@ from sphinx.ext.autodoc import (  # isort:skip
     AttributeDocumenter,
     Documenter,
     MethodDocumenter,
+    PropertyDocumenter,
 )
 from sphinx.ext.autosummary import Autosummary  # isort:skip
 
 
+class ModinAccessorDocumenter(PropertyDocumenter):
+    """
+    Generates documentation for properties of Modin objects like Series.str and Series.dt that
+    are themselves accessor classes.
+    This class is necessary because we need to monkeypatch the Series.str/dt property objects
+    with the actual classes in order for autosummary-generate to produce stubs for them.
+
+    This class is not responsible for properties of those accessors like Series.str.capitalize.
+
+    See sphinx source for PropertyDocumenter:
+    https://github.com/sphinx-doc/sphinx/blob/907d27dc6506c542c11a7dd16b560eb4be7da5fc/sphinx/ext/autodoc/__init__.py#L2691
+    """
+
+    objtype = "modinaccessor"
+    directivetype = "attribute"
+
+    # lower priority than the other custom attribute documenter (defined below)
+    priority = 0.55
+
+    def import_object(self, raiseerror=False):
+        import modin.pandas as pd
+        self.module = pd
+        self.parent = pd.Series
+        self.object_name = self.objpath[-1]
+        self.object = getattr(pd.Series, self.object_name)
+        self.isclassmethod = False
+        return True
+
+
 class ModinAccessorLevelDocumenter(Documenter):
     """
-    Performs name resolution and formatting for modin Accessor classes like Series.str and Series.dt.
+    Performs name resolution and formatting for properties of Modin Accessor classes like
+    Series.str.capitalize and Series.dt.date.
+
+    This class is not responsible for the top-level object like Series.str or Series.dt.
     """
 
     def format_name(self):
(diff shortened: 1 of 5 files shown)


sfc-gh-jrose (approved):

docs/source/conf.py
```diff
@@ -123,13 +123,46 @@
...
+
+    See sphinx source for PropertyDocumenter:
+    https://github.com/sphinx-doc/sphinx/blob/907d27dc6506c542c11a7dd16b560eb4be7da5fc/sphinx/ext/autodoc/__init__.py#L2691
+    """
+
+    objtype = "modinaccessor"
+    directivetype = "attribute"
+
+    # lower priority than the other custom attribute documenter (defined below)
+    priority = 0.55
```
sfc-gh-yzou, line 148: what is this priority used for? why do we need this?

docs/source/conf.py
```diff
@@ -123,13 +123,46 @@
     AttributeDocumenter,
     Documenter,
     MethodDocumenter,
+    PropertyDocumenter,
 )
 from sphinx.ext.autosummary import Autosummary  # isort:skip
 
 
+class ModinAccessorDocumenter(PropertyDocumenter):
+    """
```
sfc-gh-yzou, line 132: If i understand correctly, this is used for generating the doc for the class its-self like DateTime property, and the class below is used for generating documentation for the methods inside the class like dt.year etc, is that correct?

docs/source/conf.py
```diff
@@ -123,13 +123,46 @@
...
+
+    objtype = "modinaccessor"
+    directivetype = "attribute"
+
+    # lower priority than the other custom attribute documenter (defined below)
+    priority = 0.55
+
+    def import_object(self, raiseerror=False):
+        import modin.pandas as pd
```
sfc-gh-yzou, line 151: can you add comment here about what are we doing? it is not obvious here, for example why self.object_name = self.objpath[-1]?

docs/source/conf.py
```diff
@@ -123,13 +123,46 @@
...
+        self.module = pd
+        self.parent = pd.Series
+        self.object_name = self.objpath[-1]
+        self.object = getattr(pd.Series, self.object_name)
+        self.isclassmethod = False
+        return True
```
sfc-gh-yzou, line 157: do you need to return any value here?

docs/source/conf.py
```diff
@@ -243,9 +276,33 @@ def process_modin_accessors(args):
...
+    # classes can take effect) will report an error like the following for every child of Series.str
+    # and Series.dt:
+    #
+    # WARNING: [autosummary] failed to import modin.pandas.Series.str.slice.
+    # Possible hints:
+    # * AttributeError: 'property' object has no attribute 'slice'
+    # * ImportError: 
+    # * ModuleNotFoundError: No module named 'modin.pandas.Series'
+    #
+    # Because we're replacing the `property` object, we also need to set the __doc__ of the new
```
sfc-gh-yzou, line 293: what do we mean here by "replacing the `property` object"

src/snowflake/snowpark/modin/pandas/io.py
```diff
@@ -514,7 +514,7 @@ def read_excel(
     storage_options: StorageOptions = None,
     dtype_backend: DtypeBackend | NoDefault = no_default,
     engine_kwargs: dict | None = None,
-) -> DataFrame | dict[IntStrT, DataFrame]:  # pragma: no cover
+) -> pd.DataFrame | dict[IntStrT, pd.DataFrame]:  # pragma: no cover
```
sfc-gh-yzou, line 517: a quick question, if we do not add pd here, what happens? does it mean no link will be generated for the DataFrame class?

docs/source/conf.py
sfc-gh-joshi: I'm not sure why, but for some reason sphinx decides to check every single `Documenter` subclass with the specified `directivetype` (in this case "attribute") to determine which to use when documenting an attribute. The priority value (1 for the built-in default one) determines which class is used.

docs/source/conf.py
sfc-gh-joshi: Yes.

src/snowflake/snowpark/modin/pandas/io.py
sfc-gh-joshi: This error gets produced when trying to build docs:

```
/Users/joshi/code/snowpandas/src/snowflake/snowpark/modin/pandas/io.py:docstring of snowflake.snowpark.modin.pandas.io.read_excel:: WARNING: more than one target found for cross-reference 'DataFrame': modin.pandas.DataFrame, snowflake.snowpark.modin.pandas.dataframe.DataFrame, snowflake.snowpark.DataFrame, snowflake.snowpark.dataframe.DataFrame
```

After changing it to `pd.DataFrame`, no link is generated for the class, but the warning at least is no longer raised.

docs/source/conf.py
sfc-gh-joshi: Originally, `Series.str` is a python `property`, which has its __doc__ attribute set to the expected docstring. We're replacing it with the `StringMethods`/`DatetimeProperties` class instead, which has a different docstring.

docs/source/conf.py
sfc-gh-joshi: The parent method returns `True` on import success/false on failure, and sets `object`, `object_name`, etc. as side effects.

docs/source/conf.py
sfc-gh-joshi: added a comment

docs/source/conf.py
sfc-gh-yzou: i think I am still kind of fuzzy here, so if you do not specify a priority here, what will happen? can you give me an example here?

sfc-gh-joshi: If no priority is specified, then this custom class will be used when attempting to document any class's attributes. When I try removing the priority field and rebuilding, it attempts to use this class to load documentation for `DataFrame.columns`, and errors out because this class is hard-coded to work specifically for Series.

Since we want to use the default sphinx class in as many places as possible, we set this priority to be below that of the default, and below that used for sub-properties like `Series.str.capitalize`. This way, this custom class is only used when a method's template explicitly specifies the `modinaccessor::` directive.

src/snowflake/snowpark/modin/pandas/io.py
sfc-gh-yzou: is it the warning that here are conflict definition of DataFrame, but I thought we have addressed this problem

sfc-gh-yzou (approved):

docs/source/conf.py
```diff
@@ -123,13 +123,54 @@
...
     Documenter,
     MethodDocumenter,
+    PropertyDocumenter,
 )
 from sphinx.ext.autosummary import Autosummary  # isort:skip
 
 
+class ModinAccessorDocumenter(PropertyDocumenter):
+    """
+    Generates documentation for properties of Modin objects like Series.str and Series.dt that
```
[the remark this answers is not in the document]
sfc-gh-joshi: added

docs/source/conf.py
```diff
@@ -243,9 +284,33 @@ def process_modin_accessors(args):
 
 
 def setup(app):
+    # Make sure modin.pandas namespace is properly set up
+    import modin.pandas as pd
+    import snowflake.snowpark.modin.plugin
+    # Monkeypatch dt/str to make sure their children are resolvable by autosummary-generate.
+    # Without this monkeypatch, the autosummary-generate (which runs before any custom documenter
```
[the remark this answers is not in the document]
sfc-gh-joshi: added

src/snowflake/snowpark/modin/pandas/io.py
sfc-gh-joshi: I think sphinx previously couldn't see the `modin.pandas.DataFrame` class, but now it can. This should become unnecessary once we remove dataframe.py.

sfc-gh-rdurrani (approved):
LGTM!

PR #1942, opened 2024-07-19
SNOW-1491251 Column CaseWhen

sfc-gh-heshah:
Build CaseWhen expression ASTs. Note that `functions.when` is the only method for Snowpark Python users to create a CaseWhen column expression, meaning the first `when` call should always have both a condition and value.


tests/ast/data/case_when.test
```diff
@@ -0,0 +1,50 @@
...
+df = df.select(when(col("A").between(37, 42), col("B")).otherwise(col("C")))
+
+df = df.select(when(col("A").equal_nan() , "one").otherwise("other"))
+
+df = df.select(when(col("a").is_null(), lit(1)).when(col("a") == 1, lit(2)).otherwise(lit(3)))
+
+df = df.select(when(col("a").contains("foo"), "foo").otherwise("null").when(col("a").contains("bar"), "bar"))
+
+df = df.select(when(col("a") & col("b"), "both").otherwise("other"))
+
```
sfc-gh-lspiegelberg, line 18: let's add TODO comment here with JIRA to fix this up in case.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -281,6 +282,49 @@ def with_src_position(
     return expr_ast
 
 
+def _fill_ast_with_column(
```
sfc-gh-azwiegincew, line 285: Need a better fn name. One idea would be `set_column_ast`, but you might come up with something better. The param `ast` should probably be called `builder` or sth to indicate that's the value that gets set, otherwise, especially given the current name of the function, the reader gets the opposite idea.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -281,6 +282,49 @@ def with_src_position(
...
+    """
+    if value._ast is None and FAIL_ON_MISSING_AST:
+        raise NotImplementedError(
+            f"Column({value._expression})._ast is None due to the use of a Snowpark API which does not support AST logging yet."
+        )
+    elif value._ast is not None:
+        ast.CopyFrom(value._ast)
+
+
+def _fill_ast_with_snowpark_column_or_sql_expr(
```
sfc-gh-azwiegincew, line 305: dito

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-heshah: I was thinking of following your pattern with the renamed `build_const_from_python_val` and renaming the `ast` parameters to `expr_builder`. Does that sound reasonable? The functions would be called
- `build_expr_from_snowpark_column`
- `build_expr_from_snowpark_column_or_sql_str`

If this pattern is more informative we should probably also rename `_fill_ast_with_snowpark_column_or_literal` to `build_expr_from_snowpark_column_or_python_val`, (potentially also `build_const_from_python_val` to `build_expr_from_python_val`  and change the parameter name in `build_const_from_python_val` from `ast` to `expr_builder` as well)

sfc-gh-heshah: An additional note here is that the type hints defined at the bottom of `type_utils.py` provide a good indication of the "build" methods we'll need and could potentially guide the naming scheme as well.
```python
# Type hints
ColumnOrName = Union["snowflake.snowpark.column.Column", str]
ColumnOrLiteralStr = Union["snowflake.snowpark.column.Column", str]
ColumnOrSqlExpr = Union["snowflake.snowpark.column.Column", str]
LiteralType = Union[VALID_PYTHON_TYPES_FOR_LITERAL_VALUE]
ColumnOrLiteral = Union["snowflake.snowpark.column.Column", LiteralType]
```

sfc-gh-azwiegincew: Yeah, these names sound great.

tests/ast/data/case_when.test
```diff
@@ -0,0 +1,51 @@
...
+df = df.select(when(col("A") == 1, "one").when(col("A") <= 2, "two").otherwise("other"))
+
+df = df.select(when(col("A").between(37, 42), col("B")).otherwise(col("C")))
+
+df = df.select(when(col("A").equal_nan() , "one").otherwise("other"))
+
+df = df.select(when(col("a").is_null(), lit(1)).when(col("a") == 1, lit(2)).otherwise(lit(3)))
+
+# TODO: SNOW-1546550
```
sfc-gh-lspiegelberg, line 15: nit: add short description, i.e. `TODO:  SNOW-1546550 fix multiple otherwise calls.`

sfc-gh-lspiegelberg (approved):
lgtm

PR #1944, opened 2024-07-19
SNOW-1545096, SNOW-1546452: Add support for Series.dt.microsecond/nanosecond

sfc-gh-helmeleegy:

   

   Fixes SNOW-1545096, SNOW-1546452 

   Add support for Series.dt.microsecond/nanosecond.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10086,6 +10087,8 @@ def dt_property(self, property_name: str) -> "SnowflakeQueryCompiler":
             # depend on the Snowflake session's WEEK_START parameter. Subtract
             # 1 to match pandas semantics.
             "dayofweek": (lambda column: builtin("dayofweekiso")(col(column)) - 1),
+            "microsecond": (lambda column: floor(date_part("ns", col(column)) / 1000)),
+            "nanosecond": (lambda column: date_part("ns", col(column)) % 1000),
```
sfc-gh-joshi, line 10091: why do you need to take the `% 1000` here but not for the microsecond calculation?

sfc-gh-evandenberg (approved):
LGTM

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Because the number of nanoseconds should exclude whole microseconds, while the number of microseconds should not exclude whole milliseconds. (I guess because `Series.dt.millisecond` does not exist in pandas.) This is why the formulae look a bit different. But they are consistent with pandas behavior.

sfc-gh-joshi (approved):

PR #1951, opened 2024-07-22
SNOW-1491251 Implement functions.in_ and Column.in_ AST generation, with DataFrames

sfc-gh-heshah:
[SNOW-1491251](https://snowflakecomputing.atlassian.net/browse/SNOW-1491251)

Includes some modifications to `snowpark.functions.in_`. Note that `snowpark.functions.in_` is meant to be used for multiple columns where `snowpark.Column.in_` is meant to be used for a single column. Their AST patterns are similar, but we use `SpColumnInSeq` to handle `Column.in_` and the `ApplyExpr` IR entity to handle `functions.in_`.

The modifications also fix some faulty output from the Unparser for the `functions.in_` test

Also fixes a bug with `DataFrame._session._conn._suppress_not_implemented_error` usage

[SNOW-1491251]: https://snowflakecomputing.atlassian.net/browse/SNOW-1491251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/ast/data/col_in_.test
```diff
@@ -0,0 +1,19 @@
...
+
+df = df.select(col("NUM").in_([1, 2, 3]))
+
+## EXPECTED OUTPUT
+
+res1 = session.table("test_table")
+
+res2 = res1.select(col("NUM").in_(1, 2, 3))
+
+res3 = res2.select(col("NUM").in_(res2))
```
sfc-gh-lspiegelberg, line 17: Is it possible to use multiple dataframes here? Or only a single one? If the latter, maybe a quick validation/error would be good to have. If not, let's have a test case.

sfc-gh-lspiegelberg (approved):
one nit, plta. Else, lgtm.

tests/ast/data/col_in_.test
sfc-gh-heshah: Seems like only one DataFrame is allowed as a parameter. If multiple are used, the existing Snowpark code has parameter validation which raises the following exception, which confirms this as well.
```
TypeError: '<class 'snowflake.snowpark._internal.analyzer.expression.ScalarSubquery'>' is not supported for the values parameter of the function in(). You must either specify a sequence of literals or a DataFrame that represents a subquery.
```

PR #1953, opened 2024-07-22
[SNOW-1545219]Query Generator for plan compilation

sfc-gh-yzou:

SNOW-1545219

For the query generator during plan compilation, we 
1) allow skipping build of schema query during SnowflakePlan builder
2) overwrite do_resolve_with_resolved_children to allow some node that requires special processing during the plan build.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -566,6 +566,10 @@ class SelectSnowflakePlan(Selectable):
 
     @property
     def schema_query(self) -> str:
+        assert (
+            self.snowflake_plan.schema_query is not None
+        ), "No schema query found for the SelectSnowflakePlan"
+
         return self.snowflake_plan.schema_query
 
     @property
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -1143,7 +1143,7 @@ class SelectTableFunction(Selectable):
         return self._snowflake_plan.placeholder_query
 
     @property
-    def schema_query(self) -> str:
+    def schema_query(self) -> Optional[str]:
         return self._snowflake_plan.schema_query
 
     @property
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -298,7 +298,7 @@ class Selectable(LogicalPlan, ABC):
             query = Query(self.sql_query, params=self.query_params)
             queries = [*self.pre_actions, query] if self.pre_actions else [query]
             if skip_schema_query is True:
-                schema_query = None
+                schema_query = ""
             else:
                 schema_query = self.schema_query
             self._snowflake_plan = SnowflakePlan(
@@ -566,10 +566,6 @@ class SelectSnowflakePlan(Selectable):
 
     @property
     def schema_query(self) -> str:
-        assert (
-            self.snowflake_plan.schema_query is not None
-        ), "No schema query found for the SelectSnowflakePlan"
-
         return self.snowflake_plan.schema_query
 
     @property
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -291,12 +291,19 @@ class Selectable(LogicalPlan, ABC):
     @property
     def snowflake_plan(self):
         """Convert to a SnowflakePlan"""
+        return self.get_snowflake_plan(skip_schema_query=False)
+
+    def get_snowflake_plan(self, skip_schema_query):
         if self._snowflake_plan is None:
             query = Query(self.sql_query, params=self.query_params)
             queries = [*self.pre_actions, query] if self.pre_actions else [query]
+            if skip_schema_query is True:
+                schema_query = None
+            else:
+                schema_query = self.schema_query
             self._snowflake_plan = SnowflakePlan(
                 queries,
-                self.schema_query,
+                schema_query,
                 post_actions=self.post_actions,
                 session=self.analyzer.session,
                 expr_to_alias=self.expr_to_alias,
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -293,14 +293,11 @@ class Selectable(LogicalPlan, ABC):
         """Convert to a SnowflakePlan"""
         return self.get_snowflake_plan(skip_schema_query=False)
 
-    def get_snowflake_plan(self, skip_schema_query):
+    def get_snowflake_plan(self, skip_schema_query) -> SnowflakePlan:
         if self._snowflake_plan is None:
             query = Query(self.sql_query, params=self.query_params)
             queries = [*self.pre_actions, query] if self.pre_actions else [query]
-            if skip_schema_query is True:
-                schema_query = ""
-            else:
-                schema_query = self.schema_query
+            schema_query = None if skip_schema_query else self.schema_query
             self._snowflake_plan = SnowflakePlan(
                 queries,
                 schema_query,
@@ -565,7 +562,7 @@ class SelectSnowflakePlan(Selectable):
         return self._snowflake_plan._id
 
     @property
-    def schema_query(self) -> str:
+    def schema_query(self) -> Optional[str]:
         return self.snowflake_plan.schema_query
 
     @property
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -472,8 +472,13 @@ class SnowflakePlan(LogicalPlan):
 
 
 class SnowflakePlanBuilder:
-    def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
+    def __init__(self, session: "snowflake.snowpark.session.Session", skip_schema_query: bool = False) -> None:
         self.session = session
+        # Whether skip the schema query build. If true, no the schema_query associated
+        # with the resolved plan will be None.
+        # This option is currently only expected to be used for the query generator applied
+        # on the optimized plan.
+        self._skip_schema_query = skip_schema_query
 
     @SnowflakePlan.Decorator.wrap_exception
     def build(
@@ -493,9 +498,14 @@ class SnowflakePlanBuilder:
                 params=select_child.queries[-1].params,
             )
         ]
-        new_schema_query = (
-            schema_query if schema_query else sql_generator(child.schema_query)
-        )
+
+        if self._skip_schema_query is True:
+            new_schema_query = None
+        else:
+            new_schema_query = (
+                schema_query if schema_query else sql_generator(child.schema_query)
+            )
+
         placeholder_query = (
             sql_generator(select_child._id)
             if self.session._cte_optimization_enabled and select_child._id is not None
@@ -540,10 +550,12 @@ class SnowflakePlanBuilder:
                 )
             ]
         )
-
-        left_schema_query = schema_value_statement(select_left.attributes)
-        right_schema_query = schema_value_statement(select_right.attributes)
-        schema_query = sql_generator(left_schema_query, right_schema_query)
+        if self._skip_schema_query is True:
+            schema_query = None
+        else:
+            left_schema_query = schema_value_statement(select_left.attributes)
+            right_schema_query = schema_value_statement(select_right.attributes)
+            schema_query = sql_generator(left_schema_query, right_schema_query)
         placeholder_query = (
             sql_generator(select_left._id, select_right._id)
             if self.session._cte_optimization_enabled
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -57,7 +57,6 @@ from snowflake.snowpark._internal.analyzer.analyzer_utils import (
     drop_table_if_exists_statement,
     file_operation_statement,
     filter_statement,
-    get_full_table_name,
     insert_into_statement,
     join_statement,
     join_table_function_statement,
@@ -790,7 +789,7 @@ class SnowflakePlanBuilder:
                 "Internally generated tables must be called with mode ERROR_IF_EXISTS"
             )
 
-        full_table_name = get_full_table_name(table_name)
+        full_table_name = ".".join(table_name)
         is_temp_table_type = table_type in TEMPORARY_STRING_SET
         # here get the column definition from the child attributes. In certain cases we have
         # the attributes set to ($1, VariantType()) which cannot be used as valid column name
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -345,7 +345,7 @@ class SnowflakePlan(LogicalPlan):
                     pre_queries.append(query)
             # when self.schema_query is None, that means no schema query is propogated during
             # the process, there is no need to update the schema query.
-            if (self.schema_query is not None) and (plan.schema_query is not None):
+            if (new_schema_query is not None) and (plan.schema_query is not None):
                 new_schema_query = new_schema_query.replace(
                     plan.queries[-1].sql, plan.schema_query
                 )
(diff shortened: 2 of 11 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -489,46 +509,6 @@ def build(
             placeholder_query=placeholder_query,
         )
 
-    @SnowflakePlan.Decorator.wrap_exception
-    def build_from_multiple_queries(
```
sfc-gh-yzou, line 493: @sfc-gh-aalam this function seems not used anymore after you remove the create_temp_table, I removed it here because it was complaining i don't have test coverage for this code.

sfc-gh-aalam: thanks

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -267,12 +267,19 @@ def api_calls(self, value: Optional[List[Dict[str, Any]]]) -> None:
     @property
     def snowflake_plan(self):
         """Convert to a SnowflakePlan"""
+        return self.get_snowflake_plan(skip_schema_query=False)
+
+    def get_snowflake_plan(self, skip_schema_query):
         if self._snowflake_plan is None:
             query = Query(self.sql_query, params=self.query_params)
             queries = [*self.pre_actions, query] if self.pre_actions else [query]
+            if skip_schema_query is True:
```
sfc-gh-jdu, line 276:
```suggestion
            schema_query = self.schema_query if skip_schema_query else None
```
I feel None is better, the same below

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -267,12 +267,19 @@ def api_calls(self, value: Optional[List[Dict[str, Any]]]) -> None:
     @property
     def snowflake_plan(self):
         """Convert to a SnowflakePlan"""
+        return self.get_snowflake_plan(skip_schema_query=False)
+
+    def get_snowflake_plan(self, skip_schema_query):
```
sfc-gh-jdu, line 272: type hint

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -467,9 +482,14 @@ def build(
...
             )
         ]
-        new_schema_query = (
-            schema_query if schema_query else sql_generator(child.schema_query)
-        )
+
+        if self._skip_schema_query is True:
+            new_schema_query = ""
+        else:
+            new_schema_query = (
```
sfc-gh-jdu, line 489:
```suggestion
            new_schema_query = schema_query or sql_generator(child.schema_query)
```

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -765,6 +750,7 @@ def save_as_table(
         source_plan: Optional[LogicalPlan],
         use_scoped_temp_objects: bool,
         is_generated: bool,  # true if the table is generated internally
+        child_attribute: List[Attribute],
```
sfc-gh-aalam, line 753: nit: lets call it `child_attributes`

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -632,7 +614,10 @@ def large_local_relation_plan(
         )
         select_stmt = project_statement([], temp_table_name)
         drop_table_stmt = drop_table_if_exists_statement(temp_table_name)
-        schema_query = schema_query or schema_value_statement(attributes)
+        if self._skip_schema_query is True:
```
sfc-gh-aalam, line 617: nit
```suggestion
        if self._skip_schema_query:
```

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(
+        self,
+        session: Session,
+        table_create_child_attribute_map: Dict[str, List[Attribute]],
+    ) -> None:
+        super().__init__(session)
+        # overwrite the plan_builder initiated in the super to skip the building of schema query
+        self.plan_builder = SnowflakePlanBuilder(self.session, skip_schema_query=True)
+        # map between the Snowflake tabel created and its child attributes
```
sfc-gh-jdu, line 39:
```suggestion
        # map between the Snowflake table created and its child attributes
```

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -0,0 +1,114 @@
...
+        self.table_create_child_attribute_map: Dict[
```
sfc-gh-jdu, line 40: this is only used for save_as_table right? If yes, can you add a comment?

src/snowflake/snowpark/_internal/compiler/plan_compiler.py
```diff
@@ -12,16 +12,48 @@
...
+        current_session = self._plan.session
+        return (
+            self._plan.source_plan is not None
+        ) and current_session.cte_optimization_enabled
+
     def compile(self) -> Dict[PlanQueryType, List[Query]]:
-        # apply optimizations
-        final_plan = self._plan.replace_repeated_subquery_with_cte()
+        final_plan = self._plan
+        if self.should_apply_optimizations():
```
sfc-gh-aalam, line 52: shall we rename this to `should_apply_cte_optimization`. I'm assuming in the future it looks like

```
if should_apply_optimization_a():
    # apply optimization A
if should_apply_optimization_b():
    # apply optimization B
...
````

File: src/snowflake/snowpark/_internal/compiler/utils.py [hunk 10]
```diff
@@ -0,0 +1,34 @@
...
+
+
+def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
+    """
+    Helper function to construct the query generator for a given valid SnowflakePlan.
+    """
+    table_create_child_attribute_map: Dict[str, List[Attribute]] = {}
+    if plan.source_plan is not None and isinstance(
+        plan.source_plan, SnowflakeCreateTable
+    ):
```
sfc-gh-aalam, line 23: df.collect(), df.show(), df.count() will not have a `SnowflakeCreateTable` node that the top. I believe it only happens when we have df.write... or df.cache result where this assumption will hold

tests/integ/compiler/test_query_generator.py
```diff
@@ -0,0 +1,226 @@
...
+    logical_plans = [snowflake_plan.source_plan, df_res._plan.source_plan]
+    generated_queries = query_generator.generate_queries(logical_plans)
+    result_queries = [
+        query.sql.lstrip() for query in generated_queries[PlanQueryType.QUERIES]
+    ]
+    result_post_actions = [
+        query.sql.lstrip() for query in generated_queries[PlanQueryType.POST_ACTIONS]
+    ]
+    assert result_queries == expected_queries
+    assert result_post_actions == expected_post_actions
```
sfc-gh-jdu, line 226: can we add a test for 1) non-select query with session.sql 2) another type of dataframe operation with multiple queries, like https://github.com/snowflakedb/snowpark-python/blob/50275a7e0c93035759fd95f2563f18d99ae46954/tests/integ/test_dataframe.py#L2995 3) dataframe reader https://github.com/snowflakedb/snowpark-python/blob/50275a7e0c93035759fd95f2563f18d99ae46954/tests/integ/test_cte.py#L389-L390

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-aalam: reading your tests, it seems your assumption here is only for the case when table is being written and you need attributes. then I think this assumption is fair

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: updated

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: updated

src/snowflake/snowpark/_internal/compiler/plan_compiler.py
sfc-gh-yzou: eventually it will look like:
```
if self.should_apply_optimizations():
     pre-process()
     if optimization_a_enabled:
          optimization_A()
     if optimization_b_enabled:
         optimization_B()
    ....
    if optimization_triggered:
        generate_query
```

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: the  table_create_child_attribute_map is only needed to handle SnowflakeCreateTable, here we are extracting the table_create_child_attribute_map if root node is SnowflakeCreateTable, and our assumption is that the SnowflakeCreateTable must be a root node of a logical plan, which is true today. 
Added more comment on top of the if

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yzou: updated

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yzou: added

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: updated

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-yzou: updated

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-yzou: it is used to resolve SnowflakeCreateTable node, which is used for both save_as_table and cache_result. Or maybe you are referring to the save_as_table function in planBuilder. I added in the comment that it is used to resolve the SnowflakeCreateTable node

tests/integ/compiler/test_query_generator.py
sfc-gh-yzou: added

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -472,8 +482,18 @@ def add_aliases(self, to_add: Dict) -> None:
...
 class SnowflakePlanBuilder:
-    def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
+    def __init__(
+        self,
+        session: "snowflake.snowpark.session.Session",
+        skip_schema_query: bool = False,
+    ) -> None:
         self.session = session
+        # Whether skip the schema query build. If true, no the schema_query associated
+        # with the resolved plan will be None.
```
sfc-gh-aalam, line 492:
```suggestion
        # Whether skip the schema query build. If true, the schema_query associated
        # with the resolved plan will be None.
```
nit

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        create_table_node = plan.source_plan
+        # resolve the node child to get the child attribute that is needed for later code
+        # generation. Typically, the query attached to the create_table_node is already a
+        # resolved plan, and the resolve will be a no-op.
+        # NOTE that here we rely on the fact that the SnowflakeCreateTable node is the root
+        # of a source plan. Test will fail if that assumption is broken.
+        resolved_child = plan.session._analyzer.resolve(create_table_node.query)
+        full_table_name = ".".join(create_table_node.table_name)
```
sfc-gh-aalam, line 35: I see `".".join(create_table_node.table_name)` in multiple places. let's make it a `SnowflakeCreateTable` node property

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+from snowflake.snowpark._internal.analyzer.expression import Attribute
+from snowflake.snowpark._internal.analyzer.snowflake_plan import SnowflakePlan
+from snowflake.snowpark._internal.analyzer.snowflake_plan_node import (
+    SnowflakeCreateTable,
+)
+from snowflake.snowpark._internal.compiler.query_generator import QueryGenerator
+
+
+def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
```
sfc-gh-aalam, line 16: Is this util for test only? Right now it is only populating a single entry in the dictionary `table_create_child_attribute_map`. I'm assuming in the future, large query breakdown will have its corresponding entries in the dictionary but I don't see a way of populating using a single plan as input

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -360,6 +367,9 @@ def with_subqueries(self, subquery_plans: List["SnowflakePlan"]) -> "SnowflakePl
 
     @cached_property
     def attributes(self) -> List[Attribute]:
+        assert (
```
sfc-gh-jdu, line 370: @sfc-gh-aalam though highly unlikely, is it possible that schema_query can be None or `""` before? Just for the risk here, maybe just leave it as a comment instead of assertion?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -493,9 +513,15 @@ def build(
...
             )
         ]
-        new_schema_query = (
-            schema_query if schema_query else sql_generator(child.schema_query)
-        )
+
+        if self._skip_schema_query:
+            new_schema_query = None
+        else:
+            assert (
```
sfc-gh-jdu, line 520: same here

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -150,13 +150,17 @@ def __init__(
...
-        query: Optional[LogicalPlan],
+        query: LogicalPlan,
         table_type: str = "",
         clustering_exprs: Optional[Iterable[Expression]] = None,
         comment: Optional[str] = None,
         is_generated: bool = False,
     ) -> None:
         super().__init__()
+
+        assert (
```
sfc-gh-jdu, line 161: same here, I'll avoid assertion

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: @sfc-gh-jdu I don't think that is possible, if you look at the implementation for analyze_attributes https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/analyzer/schema_utils.py#L71
when schema_query is None, it will raise an error when it does sql.strip().lower(). if it is "", it will error out when we call session._get_result_attributes(sql). 
With this change, we make the None case more obvious.

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: it is also used in the save_as_table function, since the tablenames are passed as argument, i added a utility function get_full_table_name to unify all those implementations.

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: i think the question is probably not how do you propagate this map, but first do you need to propagate this map? If yes, then we can not skip schema query build, otherwise. Otherwise, we can always redesign the save_as_table interface to make sure things are optional

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: updated

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-yzou: After the refactoring, the SnowflakeCreateTable always attached with a df._plan, and we rely on this fact to make sure we can perform optimization on the child, i want to make sure such assumption is not broken in the future.

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: as the offline discussion, the property only needs to be propagated if the table is created on top of file reading and when use insert statement. For the usage of large query breakdown, the breakdown shouldn't happen right after file reading, and there should be no need to propogate this property, we can redesign the interface of save_as_table when dealing with the large query breakdown.

sfc-gh-aalam: If it is only the top level node that needs to save the attributes, then why not save a single list of attributes instead of a map which is only going to have a single entry?

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: yes, updated to have one entry

PR #1963, opened 2024-07-23
SNOW-1491302 Add AST generation in Dataframe With Col Methods

sfc-gh-heshah:
SNOW-1491302

AST generation and tests added for `Dataframe.{with_column, with_columns, rename, with_column_renamed}`

There are also bug fixes for previously added AST generation from other APIs which were not handling  `Column` inputs (simple fixes to use the correct helper functions)


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -346,6 +346,28 @@ def build_expr_from_snowpark_column_or_python_val(
...
         raise TypeError(f"{type(value)} is not a valid type for Column or literal AST.")
+    
+
+def build_expr_from_snowpark_column_or_table_fn(
+    expr_builder: proto.Expr, value: Union["snowflake.snowpark.Column", "snowflake.snowpark.table_function.TableFunctionCall"]
+) -> None:
+    """Copy from a Column object's AST, or TableFunctionCall object's AST, into an AST expression.
+
+    Args:
+        expr_builder (proto.Expr): A previously created Expr() or SpColumnExpr() IR entity intance to be filled
```
sfc-gh-lspiegelberg, line 357: should the type hint be `Union[proto.Expr, proto.SpColumnExpr]`?

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -346,6 +346,28 @@ def build_expr_from_snowpark_column_or_python_val(
         pass
     else:
         raise TypeError(f"{type(value)} is not a valid type for Column or literal AST.")
+    
+
+def build_expr_from_snowpark_column_or_table_fn(
```
sfc-gh-lspiegelberg, line 351: why not directly use if and the ` build_expr_from_snowpark_column(expr_builder, value)` helper? Why do we need to introduce a new helper function here?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-heshah: Actually the description of the argument is incorrect here, my mistake. I'm noticing this error has propagated down across the helper methods so I'll clean that up.

Just to be clear I think we moved to represent `Column` instances within `Expr` IR entities to accommodate both `SpColumnExpr` variants and values from `const.ir`. So all `expr_builder` arguments should be only `proto.Expr` instances.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-heshah: This is to accommodate arguments to `with_column(s)` which only allow Snowpark `Column` instances or `TableFunctionCall` instances in the `values` parameter.

I created this helper in anticipation of the fact that the `TableFunctionCall` ASTs will be generated in the same pattern as `Column` ASTs and stored in an internal `_ast` field. For now we don't support that so the `TableFunctionCall` case will raise a `NotImplementedError`, but later it'll be simple call to `CopyFrom`.

src/snowflake/snowpark/dataframe.py
```diff
@@ -4332,19 +4358,31 @@ def rename(
         rename_map = {k: v for k, v in zip(normalized_name_list, rename_list)}
         rename_plan = Rename(rename_map, self._plan)
 
+        # AST
+        for col, new_name in col_or_mapper.items():
+            kv_tuple_ast = expr.col_or_mapper.seq_map_val.kvs.add()
+            build_expr_from_snowpark_column_or_python_val(kv_tuple_ast.vs.add(), col)
```
sfc-gh-azwiegincew, line 4364: The order of arguments seems to be inconsistent between these 2 fns.

sfc-gh-heshah: Yeah, I believe the `build_expr_from_python_val` should have the `expr_builder` argument come first as well (this feels more intuitive and all of the other helper methods also have the arguments with the `proto.Expr` arg come first).

I can make a cleanup TODO or complete the change in this PR if the above sounds good. Its been irking me for a while :)

src/snowflake/snowpark/dataframe.py
```diff
@@ -3174,6 +3187,13 @@ def with_column(
             col_name: The name of the column to add or replace.
             col: The :class:`Column` or :class:`table_function.TableFunctionCall` with single column output to add or replace.
         """
+        if ast_stmt is None:
```
sfc-gh-lspiegelberg, line 3190: maybe let's add `_emit_ast` as arg?

src/snowflake/snowpark/dataframe.py
```diff
@@ -4308,8 +4338,15 @@ def rename(
             col_or_mapper: The old column instance or column name to be renamed, or the dictionary mapping from column instances or columns names to their new names (string)
             new_column: The new column name (string value), if a single old column is given
         """
+        # AST
```
sfc-gh-lspiegelberg, line 4341: `#AST.`

sfc-gh-lspiegelberg (approved):
couple nits, please address before merging.

PR #1964, opened 2024-07-23
SNOW-1552465: Push `result.name = None` from frontend to QC

sfc-gh-joshi:

   

   Fixes SNOW-1552465

When a Snowpark pandas aggregation returns a Series, we often use `transpose_single_row` in the query compiler to force the result to conform to the expected dimensions. After this operation, the column name is the string "NULL", which we fix using an `isinstance(result, Series)` check that guards `result.name = None`. This check does not exist in modin, and for consistency with other methods that return a Series, the name setting operation should be performed within the query compiler.

This PR modifies the query compiler `agg` and `nunique` methods to set column names to `MODIN_UNNAMED_SERIES_LABEL` after calling `transpose_single_row`, and removes the name setting operation in the corresponding frontend methods. I opted to change the columns after rather than within `transpose_single_row` in case other query compiler options would want to set the resulting column name to something different.

### DIFF
--- a/src/snowflake/snowpark/modin/pandas/base.py
+++ b/src/snowflake/snowpark/modin/pandas/base.py
@@ -2286,13 +2286,6 @@ class BasePandasDataset(metaclass=TelemetryMeta):
             **kwargs,
         )
         result_qc = self._reduce_dimension(result_qc)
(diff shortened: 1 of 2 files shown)


sfc-gh-evandenberg (approved):
LGTM

Issue #1967, opened 2024-07-23
SNOW-1552772: FutureWarning from Pandas about indexing in `mock._pandas_util`


NickFishr:
1. What version of Python are you using?
   ```
   Python 3.11.0rc1 (main, Aug 12 2022, 10:02:14) [GCC 11.2.0]
   ```

2. What are the Snowpark Python and pandas versions in the environment?
   ```shell
   pandas==2.2.1
   snowflake-snowpark-python==1.20.0
   ```

3. What did you do?

   Created a local testing session and tried to create some Snowpark dataframes from Pandas ones.

   ```python
   import pandas as pd
   from snowflake.snowpark import Session

   session = Session.builder.config("local_testing", True).create()

   pandas_df = pd.DataFrame({"Timestamp": [pd.to_datetime(1490195805, unit="s")]})
   df = session.create_dataframe(pandas_df)

   pandas_df = pd.DataFrame([[1.0]], columns=["A"], dtype=pd.Float64Dtype())
   df = session.create_dataframe(pandas_df)
   ```

5. What did you expect to see?

   No warnings.

   Instead Pandas threw the following warnings:
   ```
   /tmp/snowpark-python/src/snowflake/snowpark/mock/_pandas_util.py:96: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
     if isinstance(data.dtypes[col_idx], pd.DatetimeTZDtype):
   /tmp/snowpark-python/src/snowflake/snowpark/mock/_functions.py:811: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
     data = row[0]
   /tmp/snowpark-python/src/snowflake/snowpark/mock/_pandas_util.py:82: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
     if isinstance(data.dtypes[col_idx], PANDAS_INTEGER_TYPES)
[...]

PR #1968, opened 2024-07-23
Fix Pandas FutureWarning about integer indexing

NickFishr:

   Fixes SNOW-NNNNNNN

   Github Issue #1967 

   Pandas prefers integer indexing with the `.iloc[]` accessor instead of implicitly assuming that an integer key is not a label but an index. It also issues a `FutureWarning` that in the future also integer keys will be treated as labels and not as integer index, which will most likely cause an error when this behavior is implemented.

   The former code used either used `data[x]` instead of `data.iloc[x]` or (maybe unintentionally) `data.iloc[x][y]` instead of `data.iloc[x, y]` and caused this warning.

   Not sure if an additional test is really necessary to check for warnings in the future, but I've added it anyway, please feel free to omit the test from the PR if you don't see it as necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@
 - Fixed a bug that Window Functions LEAD and LAG do not handle option `ignore_nulls` properly.
 - Fixed a bug where values were not populated into the result DataFrame during the insertion of table merge operation.
 
+#### Improvements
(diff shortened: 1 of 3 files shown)


NickFishr:

sfc-gh-aling (approved):

sfc-gh-jrose (approved):

PR #1970, opened 2024-07-24
(Readability) Replace all instances of suppress_ast with emit_ast to avoid negated logic

sfc-gh-oplaton:

   

   Fixes SNOW-0000000

Stylistic change - prefer direct logic instead of inverted logic.

Use `emit_ast` (default = `True`) everywhere instead of `suppress_ast` (default = `False`).

To get ASTs out of public APIs, do nothing. To prevent public APIs from emitting ASTs, provide `emit_ast = False`.

Name the argument `_emit_ast` on public APIs and `emit_ast` on internal APIs.

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -277,7 +277,7 @@ def _disambiguate(
             )
             for name in lhs_names
         ],
-        _suppress_ast=True,
+        _emit_ast=False,
     )
 
     rhs_remapped = rhs.select(
@@ -285,7 +285,7 @@ def _disambiguate(
             _alias_if_needed(rhs, name, rhs_prefix, rsuffix, common_col_names)
             for name in rhs_names
         ],
-        _suppress_ast=True,
+        _emit_ast=False,
     )
     return lhs_remapped, rhs_remapped
 
@@ -1131,7 +1131,7 @@ class DataFrame:
             Iterable[Union[ColumnOrName, TableFunctionCall]],
         ],
         _ast_stmt: proto.Assign = None,
-        _suppress_ast: bool = False,
+        _emit_ast: bool = True,
     ) -> "DataFrame":
         """Returns a new DataFrame with the specified Column expressions as output
         (similar to SELECT in SQL). Only the Columns specified as arguments will be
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -277,15 +277,15 @@ def _disambiguate(
             )
             for name in lhs_names
         ],
```
sfc-gh-lspiegelberg, line 279: this change makes sense to me, let's also check that all docstrings are updated with a unique description (that probably needs to get copied around). If you dislike copy&paste of the docstring, we could think of a decorator to inject a unique description. For mid-term, we may need to think about using a PIMPL pattern for the actual implementation. Most of the current Snowpark APIs are public, so adding a new parameter could be problematic.

sfc-gh-lspiegelberg (approved):
LGTM, let's check docstrings as well before merging.

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/session.py
```diff
@@ -1834,17 +1834,15 @@ def update_query_tag(self, tag: dict) -> None:
                     f"Expected query tag to be valid json. Current query tag: {tag_str}"
                 )
 
-    def table(
-        self, name: Union[str, Iterable[str]], suppress_ast: bool = False
-    ) -> Table:
+    def table(self, name: Union[str, Iterable[str]], emit_ast: bool = True) -> Table:
```
sfc-gh-lspiegelberg, line 1837: nit: table should be public API? so `_emit_ast` according to PR description?

src/snowflake/snowpark/table.py
```diff
@@ -273,9 +273,9 @@ def __init__(
         table_name: str,
         session: Optional["snowflake.snowpark.session.Session"] = None,
         ast_stmt: Optional[proto.Assign] = None,
-        suppress_ast: bool = False,
```
sfc-gh-lspiegelberg, line 276: nit: same here, should be public. tbh, I am ok to call it always `_emit_ast`. Will make things easier.

sfc-gh-oplaton: Changed all files to use `_emit_ast` uniformly.

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: Updated all docstrings to use the same description.

src/snowflake/snowpark/session.py
sfc-gh-oplaton: Done.

PR #1976, opened 2024-07-24
DRAFT: client-side types in Snowpark Python

sfc-gh-mvashishtha:


src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -688,7 +691,20 @@ def binary_operator_extractor(
                 expr.right, df_aliased_col_name_to_real_col_name, parse_local_name
             )
         if isinstance(expr, BinaryArithmeticExpression):
+            # TODO: it doesn't seem appropriate to rewrite the expression at this stage,
```
sfc-gh-lspiegelberg, line 694: You're probably already on it, but I feel the logic here should be something like:
```
if is_user_defined_type(expr.left.datatype) or is_user_defined_type(expr.right.datatype):
   gen = get_user_defined_binary_expr(expr) # <-- look up based on left/right datatype and type(expr)
   if not gen:
      raise NotImplementedError(f"No sql codegen available for {type(expr)}: {expr.left.datatype} x {expr.right.datatype}"
```

and then you can register code generators like this:

```
register_sql_binary_codegen(Subtract, TimestampType, TimestampType, lambda left_sql, right_sql: f'datediff("ns", {right_sql_expr}, {left_sql_expr})')
....
```

sfc-gh-mvashishtha: Done!

src/snowflake/snowpark/modin/pandas/__init__.py
```diff
@@ -346,3 +346,71 @@ def __getattr__(name: str) -> Any:
 modin.core.dataframe.algebra.default2pandas.default.DefaultMethod.register = (
     snowflake.snowpark.modin.core.dataframe.algebra.default2pandas.default.DefaultMethod.register
 )
+
```
sfc-gh-lspiegelberg, line 349: this looks great here @sfc-gh-mvashishtha! I can already see this abstraction being helpful.

src/snowflake/snowpark/modin/pandas/new_types_demo.py
```diff
@@ -0,0 +1,52 @@
+import modin.pandas as pd
+import snowflake.snowpark.modin.plugin
+import numpy as np
+import pandas as native_pd
+from snowflake.snowpark.session import Session; session = Session.builder.create()
+import logging; logging.getLogger("snowflake.snowpark").setLevel(logging.DEBUG)
+
```
sfc-gh-lspiegelberg, line 7: for demo purposes would recommend a jupyter noteboook, just easier as you can quickly modify code.

PR #1980, opened 2024-07-24
SNOW-1555397 Fix index.to_frame()'s wrong result column name

sfc-gh-azhan:

   

   Fixes SNOW-1555397

Fix the result column name for `Index.to_frame` when `name` is not specified.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/extensions/index.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/index.py
@@ -1662,26 +1662,22 @@ class Index:
         #       0               100
         #       1               200
         #       2               300
-        # If `name` is specified, use it as new column name; otherwise, set new column name to the original index name.
-        # Note there is one exception case: when the original index name is None, the new column name should be 0.
-        if name != native_pd._libs.lib.no_default:
-            new_col_name = name
-        else:
-            new_col_name = self._query_compiler.get_index_name()
-            if new_col_name is None:
-                new_col_name = 0
-
+        new_qc = self._query_compiler.reset_index()
         # if index is true, we want self to be in the index and data columns of the df,
         # so set the index as the data column and set the name of the index
         if index:
-            new_qc = self._query_compiler.reset_index()
-            new_qc = (
-                new_qc.set_index([new_qc.columns[0]], drop=False)
-                .set_columns([new_col_name])
-                .set_index_names([self._query_compiler.get_index_name()])
+            new_qc = new_qc.set_index([new_qc.columns[0]], drop=False).set_index_names(
+                [self.name]
             )
+        # If `name` is specified, use it as new column name; otherwise, set new column name to the original index name.
+        # Note there is one exception case: when the original index name is None, the new column name should be 0.
+        if name != lib.no_default:
+            new_col_name = name
         else:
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1633,17 +1662,26 @@ def to_frame(self, index: bool = True, name: Hashable | None = None) -> DataFram
         #       0               100
         #       1               200
         #       2               300
+        # If `name` is specified, use it as new column name; otherwise, set new column name to the original index name.
+        # Note there is one exception case: when the original index name is None, the new column name should be 0.
+        if name != native_pd._libs.lib.no_default:
```
sfc-gh-yzou, line 1667: can this check just be name != lib.no_default like what you added in the type hint

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1633,17 +1662,26 @@ def to_frame(self, index: bool = True, name: Hashable | None = None) -> DataFram
...
+            new_col_name = name
+        else:
+            new_col_name = self._query_compiler.get_index_name()
```
sfc-gh-yzou, line 1670: hmm, i only see we provide get_index_names() methods in the snowflake query compiler, is this added somewhere else?

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1633,17 +1662,26 @@ def to_frame(self, index: bool = True, name: Hashable | None = None) -> DataFram
...
             new_qc = (
                 new_qc.set_index([new_qc.columns[0]], drop=False)
-                .set_columns([name])
-                .set_index_names([self.name])
+                .set_columns([new_col_name])
+                .set_index_names([self._query_compiler.get_index_name()])
             )
         else:
-            new_qc = self._query_compiler.reset_index(names=[name])
+            new_qc = self._query_compiler.reset_index(names=[new_col_name])
```
sfc-gh-yzou, line 1684: i am looking at this code block,, why don't we just have 
```
new_qc = self._query_compiler.reset_index(names=[new_col_name])
if index:
   new_qc = new_qc.set_index([new_qc.columns[0]], drop=False).set_index_names([self._query_compiler.get_index_name()])
```

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-azhan: It was in query_compiler. We can just use self.name instead.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-azhan: Looks better. I updated the code and move set column name in the end because there is a corner case: 
```
pd.DataFrame([],index=pd.Index([[1, 2], [2, 3], [3, 4]])).reset_index(names=[None]).set_index([None], drop=False)
```
which generated different result than native one. We probably don't need to fix this since this should be a rare case.

sfc-gh-nkumar (approved):

sfc-gh-yzou (approved):

PR #1983, opened 2024-07-25
SNOW-1531493: Clean up temp table with threading

sfc-gh-jdu:

   

   Fixes SNOW-1531493

   -  [x] I am adding a new automated test(s) to verify correctness of my new code

   This PR moves all temp cleaning logic to a separate class, which will clean up temp tables once they are not reference anymore.


tests/integ/test_temp_table_cleanup.py
```diff
@@ -19,40 +24,61 @@
 ]
```
sfc-gh-jdu, line 25: any other tests you can think of?

tox.ini
```diff
@@ -8,7 +8,7 @@ omit = */src/snowflake/snowpark/modin/config/*
 [coverage:run]
 branch = true
 parallel = true
-concurrency = multiprocessing
+concurrency = multiprocessing, thread
```
sfc-gh-jdu, line 11: this is required for covering code run by thread

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -100,12 +100,12 @@ def __init__(
         # When `is_temp_table_for_cleanup` is True, it's a temp table
         # generated by Snowpark (currently only df.cache_result) under the hood
         # and users are not aware of it.
-        if is_temp_table_for_cleanup:
-            session._temp_table_ref_count_map[name] += 1
+        if is_temp_table_for_cleanup and session._temp_table_cleaner is not None:
+            session._temp_table_cleaner.inc_ref_count(name)
```
sfc-gh-yzou, line 104: can you move all  those to the temp_table_cleaner, you can have an interface 
```
table_ref_created(table: SnowflakeTable) {
    inc_ref_count(name)
    table._finalizer = weakref.finalize(dec_ref_count, name)
}
```

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+import threading
+from collections import defaultdict
+from queue import Queue
+from typing import TYPE_CHECKING, Dict
+
+if TYPE_CHECKING:
+    from snowflake.snowpark.session import Session  # pragma: no cover
+
+
+class TempTableCleaner:
```
sfc-gh-yzou, line 15: maybe just call this AutoManagedTableGarbageCollector, sounds fancier :)

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+    session.
+    """
+
+    def __init__(self, session: "Session") -> None:
+        self.session = session
+        # this dict maintains key-value pair from Snowpark-generated temp table fully-qualified name
+        # to its reference count for later temp table management
+        # this dict will still be maintained even if the cleaner is stopped (`stop()` is called)
+        self.ref_count_map: Dict[str, int] = defaultdict(int)
+        self.queue = Queue()
```
sfc-gh-yzou, line 33: can you add type hint for those variables also?

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+            table_name = self.queue.get()
+            if table_name is None:
+                # exit the thread loop
+                break
+            self.drop_table(table_name)
+
+    def drop_table(self, name: str) -> None:
+        common_log_text = f"temp table {name} in session {self.session.session_id}"
+        try:
+            # TODO SNOW-1556553: Remove this workaround once multi-threading of Snowpark session is supported
```
sfc-gh-yzou, line 58: let's add a debug log before we start dropping the table also

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+        return self.cleanup_thread is not None and self.cleanup_thread.is_alive()
+
+    def start(self) -> None:
+        if not self.is_alive():
+            self.cleanup_thread = threading.Thread(target=self.process_cleanup)
+            self.cleanup_thread.start()
+
+    def stop(self) -> None:
+        if self.is_alive():
+            self.queue.put(None)
```
sfc-gh-yzou, line 77: is that how we clear the queue? isn't there a method called clear or somethign like this?

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+            self.cleanup_thread.join()
```
sfc-gh-yzou, line 78: let's also clean up the ref_count_map

sfc-gh-yzou: That makes me start thinking of a scenario where if user start turning this process on and off, which may not working well.
for example, 
1) one ref of table1 is established
2) clean up turned off
3) an other table1 ref is established
4) turn clean up on, and one table1 ref is dropped
we might drop off the table1 incorrectly.

in python is there a way to discover all SnowflakeTable object to recover the ref state correctly? that is similar as service recover, where we need to recover the service state correctly when the service is back on, otherwise the service will not operate correctly.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -100,12 +100,12 @@ def __init__(
...
             # the finalizer will be triggered when it gets garbage collected
             # and this table will be dropped finally
             self._finalizer = weakref.finalize(
-                self, session._dec_temp_table_ref_count, name
+                self, session._temp_table_cleaner.dec_ref_count, name
```
sfc-gh-yzou, line 108: if we call the other interface table_ref_created, maybe let's call this table_ref_destroyed to be consistent, this two interfaces basically defines the actions needs to be take when those two events happens.

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+    def process_cleanup(self) -> None:
+        while True:
+            table_name = self.queue.get()
+            if table_name is None:
+                # exit the thread loop
+                break
+            self.drop_table(table_name)
+
+    def drop_table(self, name: str) -> None:
+        common_log_text = f"temp table {name} in session {self.session.session_id}"
```
sfc-gh-yzou, line 56: will we be able to collect the number of tables auto dropped as a telemetry?

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+        self.ref_count_map: Dict[str, int] = defaultdict(int)
+        self.queue = Queue()
+        self.cleanup_thread = None
+
+    def inc_ref_count(self, name: str) -> None:
+        self.ref_count_map[name] += 1
+
+    def dec_ref_count(self, name: str) -> None:
+        self.ref_count_map[name] -= 1
+        if self.ref_count_map[name] <= 0:
```
sfc-gh-yzou, line 41: what does it mean if self.ref_count_map[name] < 0: is that an expected case, if not, let's record an error, and have some information for investigation

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-yzou, line 78: so what happens when the clean up thread is stopped, and started, and during that time some table ref is dropped down. shouldn't we repick up clean up of those tables?

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+            self.ref_count_map.pop(name)
```
sfc-gh-yzou, line 42: shall we pop the table now, or pop it out of the map after the drop is successfully done?

src/snowflake/snowpark/_internal/temp_table_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+    def drop_table(self, name: str) -> None:
+        common_log_text = f"temp table {name} in session {self.session.session_id}"
+        try:
+            # TODO SNOW-1556553: Remove this workaround once multi-threading of Snowpark session is supported
+            with self.session._conn._conn.cursor() as cursor:
+                cursor.execute(f"drop table if exists {name}")
+            logging.debug(f"Cleanup Thread: Successfully dropped {common_log_text}")
+        except Exception:
+            logging.debug(
+                f"Cleanup Thread: Failed to drop {common_log_text}"
```
sfc-gh-yzou, line 64: shall we reinsert it in the queue for a retry if it is not because the table not exists?

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: I'll avoid GarbageCollector because this is too specific to language runtime. How about just calling TempTableManager, which is more extensible for future usage?

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: Yes, I think we can by just adding a counter in the class. I'd rather do it in another PR, perhaps considering other telemetry data we want to collect. Or you want me to do it in this PR?

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: with the current implementation, we will not be able to clean up those tables. I'm not sure whether it's really necessary because the above scenario is rare. But if we want to address this scenario, I'd like to 1) do not pop the table name from ref_count_map if it's zero, 2) make the cleanup thread run periodically to clean up all tables with zero count, instead of on-demand implementation with a queue, 3) only delete the table name in ref_count_map if it's dropped successfully.

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: It doesn't really matter for now, because `dec_ref_count` is only triggered during garbage collection once, so this table won't be visited again anyway. But if we want to consider [this case](https://github.com/snowflakedb/snowpark-python/pull/1983#discussion_r1693390798), we should pop it out only when the drop is successfully done.

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: I'd prefer not for now, because it's possible that we will run drop sql infinitely if it always fails. Though we can add something like retry and only run drop sql twice at most.

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-jdu: good call, I moved all logic to temp_table_cleaner now, which is much neater

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: nah, this is used to break the loop in cleanup thread and force it exit, then we can call thread.join() to stop this thread. Added a comment for this.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-jdu: table_ref_destroyed is not correct because after decrement if the reference count is not 0, the temp table won't be dropped. I'd prefer to keep the current name because it describes what it's exactly doing, and I also make it private now and won't be used outside of class, which might be confusing.

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: Good point, I added a log here

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-yzou: First of all, I am not sure if that scenario is rare or not, which seems an irrelevant question here. This is purely about the responsibility of the clean up thread, and the behavior of on thread start and stop. 

Based on the current implementation, it seems the current behavior is  
1) on stop, we basically will stop the clean up thread.
2) on start,  it will not start from where it left, so whatever table whose reference has reached to 0 during the time between stop and start will always be there.

If the clean up thread's responsibility is to clean up the table whose ref is 0 in the session, then we probably should resume the clean up. agree that we do not need to have a queue, but we may need some thread safe data structure, since you will have concurrent access to the map, since all concurrent write should be serialized by the main thread, this may not be bad.

if the cleanup thread's responsibility is to clean up the table whose ref becomes 0 after the thread starts, then we do not have to, which seems little bit incomplete to me.

I might leaning towards to the first behavior to get a compete table clean responsibility

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,93 @@
...
+    def start(self) -> None:
+        if not self.is_alive():
+            self.cleanup_thread = threading.Thread(target=self.process_cleanup)
+            self.cleanup_thread.start()
+
+    def stop(self) -> None:
+        if self.is_alive():
+            # None will break the loop in cleanup thread and force it exit
+            # then we can call thread.join() to stop this thread
+            self.queue.put(None)
```
sfc-gh-yzou, line 92: i am not sure if I get this part, were you relying on this code here
```
if table_name is None:
                # exit the thread loop
                break
```
so if the queue by the time when we call stop, still have some items, it will finish those items first until it reaches none? and until it reaches none, there can continue be some items added to the queue if some ref descrease happens? is there a way to force interrupt the thread in python to stop the clean up immediately. if user calls stop, that might means we are doing something wrong, we should stop immediately.

src/snowflake/snowpark/session.py
```diff
@@ -593,6 +594,8 @@ def close(self) -> None:
         finally:
             try:
                 self._conn.close()
+                if self._temp_table_auto_cleaner.is_alive():
```
sfc-gh-yzou, line 597: there is one thing that i am not so clear about the behavior, let's say user created more than one session, and one session didn't get used for long time, does it gets garbage collected()? with that extra clean up component, will it block the session to be garbage collected()? do you have an answer for this?

sfc-gh-jdu: https://github.com/snowflakedb/snowpark-python/blob/7c9df8446301c156248d76d45b67a5d0fa7b6726/src/snowflake/snowpark/session.py#L256 will be called before Session object gets garbage collected. So with `close`, this thread will also be terminated and won't interrupt  Session object's garbage collection

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: First, `stop` and `TempTableAutoCleaner` is not exposed to users so the only way users can "stop" cleaner is to set `session.auto_clean_up_temp_table_enabled = False`. Based on our previous experience of how users are using sql simplifier's parameter, it is more likely to be used:

```
session.parameter = True
... # dataframes code that needs sql simplifier
session.parameter = False
```

so with this context, I feel the cleaner should finish cleanup temp tables that in the queue. But if you prefer to force it exit, we can also do that.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-jdu: removed inc_ref_count, and change the name of this method to `_delete_ref_count`

src/snowflake/snowpark/_internal/temp_table_cleaner.py
sfc-gh-jdu: per our offline discussion, I changed the implementation to first behavior.

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,93 @@
...
+        # this dict will still be maintained even if the cleaner is stopped (`stop()` is called)
+        self.ref_count_map: Dict[str, int] = defaultdict(int)
+        self.queue: Queue = Queue()
+        self.cleanup_thread: Optional[threading.Thread] = None
+
+    def add(self, table: SnowflakeTable) -> None:
+        self._inc_ref_count(table.name)
+        # the finalizer will be triggered when it gets garbage collected
+        # and this table will be dropped finally
+        _ = weakref.finalize(table, self._dec_ref_count, table.name)
```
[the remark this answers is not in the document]
sfc-gh-jdu: see https://github.com/snowflakedb/snowpark-python/pull/1983#discussion_r1696099716

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+    session.
+    """
+
+    def __init__(self, session: "Session") -> None:
+        self.session = session
+        # this dict maintains key-value pair from Snowpark-generated temp table fully-qualified name
+        # to its reference count for later temp table management
+        # this dict will still be maintained even if the cleaner is stopped (`stop()` is called)
+        self.ref_count_map: Dict[str, int] = defaultdict(int)
+        self.queue: Queue = Queue()
```
sfc-gh-yzou, line 35: add a comment about what is this queue used for?

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+        self.cleanup_thread: Optional[Thread] = None
+        self.stop_event = Event()
```
sfc-gh-yzou, line 37: comment for stop event? that seems a single state, what is the benefit of using event, not just a single boolean flag ?

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+                f"Unexpected reference count {self.ref_count_map[name]} for table {name}"
+            )
+
+    def process_cleanup(self) -> None:
+        while not self.stop_event.is_set():
+            try:
+                # it's non-blocking after timeout and become interruptable with stop_event
+                # it will raise an `Empty` exception if queue is empty after timeout,
+                # then we catch this exception and avoid breaking loop
+                table_name = self.queue.get(timeout=1)
```
sfc-gh-yzou, line 62: The try catch here seems little bit wired, can you simply do a check fi the queue is empty before get? and if it is empty, just continue, or you can sleep for a second before continue

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+
+from snowflake.snowpark._internal.analyzer.snowflake_plan_node import SnowflakeTable
+
+if TYPE_CHECKING:
+    from snowflake.snowpark.session import Session  # pragma: no cover
+
+
+class TempTableAutoCleaner:
+    """
+    Clean up temporary tables created automatically by Snowpark at the session level.
```
sfc-gh-yzou, line 19: -> clean up *unused* temp tables

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+
+    This function identifies and removes any temporary tables that have been
```
sfc-gh-yzou, line 21: the comment sounds little bit wired here, you mean identifies the temp tables that has no reference anymore, right? can you re organize the comment here little bit to make it more clear

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+        logging.debug(f"Cleanup Thread: Ready to drop {common_log_text}")
+        try:
+            # TODO SNOW-1556553: Remove this workaround once multi-threading of Snowpark session is supported
+            with self.session._conn._conn.cursor() as cursor:
+                cursor.execute(f"drop table if exists {name}")
+            logging.debug(f"Cleanup Thread: Successfully dropped {common_log_text}")
+        except Exception:
+            logging.warning(
+                f"Cleanup Thread: Failed to drop {common_log_text}"
+            )  # pragma: no cover
```
sfc-gh-yzou, line 78: can we add the exception message in the log also?

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def is_alive(self) -> bool:
+        return self.cleanup_thread is not None and self.cleanup_thread.is_alive()
+
+    def start(self) -> None:
+        if not self.is_alive():
+            self.stop_event.clear()
```
sfc-gh-yzou, line 85: should we clear the stop event regardless the thread is alive or not?

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+            self.ref_count_map.pop(name)
+            # clean up
+            self.queue.put(name)
+        elif self.ref_count_map[name] < 0:
+            logging.debug(
+                f"Unexpected reference count {self.ref_count_map[name]} for table {name}"
+            )
+
+    def process_cleanup(self) -> None:
+        while not self.stop_event.is_set():
```
sfc-gh-yzou, line 57: is there a reason we are using stop_event here, instead of directly interrupt the thread?

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+        self.stop_event = Event()
+
+    def add(self, table: SnowflakeTable) -> None:
+        self.ref_count_map[table.name] += 1
+        # the finalizer will be triggered when it gets garbage collected
+        # and this table will be dropped finally
+        _ = weakref.finalize(table, self._delete_ref_count, table.name)
+
+    def _delete_ref_count(self, name: str) -> None:
+        self.ref_count_map[name] -= 1
```
sfc-gh-yzou, line 46: this function does more than just decrease the ref count, can you add more comment here about what this function is doing?

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: done (and I also change to set the stop event in `stop` regardless of being alive or not)

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: done

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,100 @@
...
+from snowflake.snowpark._internal.analyzer.snowflake_plan_node import SnowflakeTable
+
+if TYPE_CHECKING:
+    from snowflake.snowpark.session import Session  # pragma: no cover
+
+
+class TempTableAutoCleaner:
+    """
+    Automatically cleans up unused temporary tables created in the current session
+    when the DataFrame is no longer referenced (i.e., gets garbage collected).
```
sfc-gh-yzou, line 20: DataFrame -> temp table, this component doesn't use the dataframe concept at all.

sfc-gh-yzou (approved):

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -0,0 +1,95 @@
...
+            except Empty:
+                continue
+
+    def drop_table(self, name: str) -> None:
+        common_log_text = f"temp table {name} in session {self.session.session_id}"
+        logging.debug(f"Cleanup Thread: Ready to drop {common_log_text}")
+        try:
+            # TODO SNOW-1556553: Remove this workaround once multi-threading of Snowpark session is supported
+            with self.session._conn._conn.cursor() as cursor:
+                cursor.execute(f"drop table if exists {name}")
```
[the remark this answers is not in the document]
sfc-gh-jdu: nah it seems not easy to add it to query listener because it goes through ServerConnection. I changed the query as you suggested.

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: done

PR #1986, opened 2024-07-25
SNOW-1489371: Implement GroupBy.value_counts

sfc-gh-joshi:

   

   Fixes SNOW-1489371

This PR adds support for GroupBy.value_counts, accepting all parameters except `bin`, which we do not support for DataFrame/Series.value_counts.

Upstream modin defaults to pandas for both DataFrameGroupBy/SeriesGroupBy.value_counts, so some of these changes should be eventually upstreamed.

pandas has different behavior than what might be expected from documentation; this PR tries to align with existing behavior as much as possible. This is documented in this pandas issue: https://github.com/pandas-dev/pandas/issues/59307

1. When `normalize=True`, pandas sorts by the pre-normalization counts, leading to counterintuitive results. This only matters when `groupby` is called with `sort=False` and `value_counts` with `sort=True`. See test cases for an example.
2. pandas does not always respect the original order of data, depending on the configuration of sort flags in `groupby` and the `value_counts` call itself. The behaviors are as follows (copied from query compiler comments):
```
# pandas currently provides the following behaviors based on the different sort flags.
# These behaviors are not entirely consistent with documentation; see this issue for discussion:
# https://github.com/pandas-dev/pandas/issues/59307
#
# Example data (using pandas 2.2.1 behavior):
# >>> df = pd.DataFrame({"X": ["B", "A", "A", "B", "B", "B"], "Y": [4, 1, 3, -2, -1, -1]})
#
# 1. groupby(sort=True).value_counts(sort=True)
#   Sort on non-grouping columns, then sort on frequencies, then sort on grouping columns.
# >>> df.groupby("X", sort=True).value_counts(sort=True)
# X  Y
# A   1    1
#     3    1
# B  -1    2
#    -2    1
#     4    1
# Name: count, dtype: int64
#
# 2. groupby(sort=True).value_counts(sort=False)
#   Sort on non-grouping columns, then sort on grouping columns.
# >>> df.groupby("X", sort=True).value_counts(sort=True)
# X  Y
# X  Y
# A   1    1
#     3    1
# B  -2    1
#    -1    2
#     4    1
# Name: count, dtype: int64
#
# 3. groupby(sort=False).value_counts(sort=True)
#   Sort on frequencies.
# >>> df.groupby("X", sort=False).value_counts(sort=True)
# X  Y
# B  -1    2
#     4    1
# A   1    1
#     3    1
# B  -2    1
# Name: count, dtype: int64
#
# 4. groupby(sort=False).value_counts(sort=False)
#   Sort on nothing (entries match the order of the original frame).
# X  Y
# B   4    1
# A   1    1
#     3    1
# B  -2    1
#    -1    2
# Name: count, dtype: int64
#
# Lastly, when `normalize` is set with groupby(sort=False).value_counts(sort=True, normalize=True),
# pandas will sort by the pre-normalization counts rather than the resulting proportions. As this
# is an uncommon edge case, we cannot handle this using existing QC methods efficiently, so we just
# update our testing code to account for this.
# See comment 
```


docs/source/modin/supported/groupby_supported.rst
```diff
@@ -166,7 +166,7 @@ Computations/descriptive stats
 +-----------------------------+---------------------------------+----------------------------------------------------+
 | ``take``                    | N                               |                                                    |
 +-----------------------------+---------------------------------+----------------------------------------------------+
-| ``value_counts``            | N                               |                                                    |
+| ``value_counts``            | P                               | SeriesGroupBy does not implement ``bins``          |
```
sfc-gh-azhan, line 169:
```suggestion
| ``value_counts``            | P                               | ``N`` if ``bins`` is given for SeriesGroupBy          |
```

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@
 - Added support for `Index.is_boolean`, `Index.is_integer`, `Index.is_floating`, `Index.is_numeric`, and `Index.is_object`.
 - Added support for `DatetimeIndex.round`, `DatetimeIndex.floor` and `DatetimeIndex.ceil`.
 - Added support for `Series.dt.days_in_month` and `Series.dt.daysinmonth`.
+- Added support for `DataFrameGroupBy.value_counts` and `SeriesGroupBy.value_counts`.
```
sfc-gh-azhan, line 79: I found we have a `value_counts` method in general.py, do you support that too?

sfc-gh-joshi: I'm not sure why we don't support the general.py method actually, since that uses the existing `Series.value_counts` implementation. Looking at the comments in our code + the pandas website, it seems like `pd.value_counts` is actually undocumented: https://pandas.pydata.org/docs/reference/general_functions.html

It should be simple to implement, but I'll leave that for a separate PR.

sfc-gh-azhan: SG. Let's do it then.

sfc-gh-azhan (approved):

src/snowflake/snowpark/modin/pandas/groupby.py
```diff
@@ -188,13 +189,28 @@ def sem(self, ddof=1):
...
+            by=self._by,
+            axis=self._axis,
+            groupby_kwargs=self._kwargs,
+            subset=subset,
+            normalize=normalize,
+            sort=sort,
+            ascending=ascending,
+            dropna=dropna,
+        )
+        if self._as_index:
```
sfc-gh-vbudati, line 208: what sets `_as_index`?

src/snowflake/snowpark/modin/plugin/docstrings/groupby.py
```diff
@@ -203,7 +203,101 @@ def sem():
...
+            Don't include counts of rows that contain NA values.
+
+        Returns
+        -------
+        :class:`~snowflake.snowpark.modin.pandas.Series` or :class:`~snowflake.snowpark.modin.pandas.DataFrame`
+            Series if the groupby as_index is True, otherwise DataFrame.
+
+        Notes
+        -----
+        - If the groupby as_index is True then the returned Series will have a MultiIndex with one level per input column.
```
sfc-gh-vbudati, line 233: Add a note about this stuff as well:
```
Lastly, when `normalize` is set with groupby(sort=False).value_counts(sort=True, normalize=True),
# pandas will sort by the pre-normalization counts rather than the resulting proportions. This
# is an uncommon edge case and we cannot handle this using existing QC methods efficiently
```
and also mentioning the github issue would be useful I think

sfc-gh-vbudati (commented):
Great work, Jonathan! This seems like a monstrous task, thank you for having such detailed documentation about it. I left a couple comments and I think it's good to merge after.

src/snowflake/snowpark/modin/plugin/docstrings/groupby.py
sfc-gh-joshi: Good call, I added this to the docstring.

src/snowflake/snowpark/modin/pandas/groupby.py
sfc-gh-joshi: It's a property on the GroupBy frontend object ([here](https://github.com/snowflakedb/snowpark-python/blob/4bcd987af779e9281ff4b5bbdd37de32911ae299/src/snowflake/snowpark/modin/pandas/groupby.py#L1085)), which reads a kwarg set in GroupBy's constructor ([here](https://github.com/snowflakedb/snowpark-python/blob/4bcd987af779e9281ff4b5bbdd37de32911ae299/src/snowflake/snowpark/modin/pandas/groupby.py#L119-L124)).

sfc-gh-vbudati (approved):

PR #1987, opened 2024-07-26
SNOW-1557229: Add method to replace child for LogicalPlan and derivatives

sfc-gh-aalam:

   

   Fixes SNOW-1557229

   This PR add method `repalce_child` to `LogicalPlan`, `SnowflakePlan` and `Selectable` classes to replace and old child node with a new one. This method is used when rewriting plan tree during optimization stages.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -205,7 +205,7 @@ def _deepcopy_selectable_fields(
     # to avoid run into recursively copy problem, we do not copy the _snowflake_plan
     # field by default and let it rebuild when needed. As far as we have other fields
     # copied correctly, the plan can be recovered properly.
-    to_selectable._is_deep_copied = True
+    to_selectable._is_valid_for_replacement = True
 
 
 class Selectable(LogicalPlan, ABC):
@@ -351,25 +351,6 @@ class Selectable(LogicalPlan, ABC):
             if self.snowflake_plan.source_plan
             else []
         )
-
-    def replace_child(self, old_node, new_node) -> None:
-        """This method is called during optimization stage to cut plan tree at a certain node.
-
-        It must only be called on a deep copied plan node, otherwise it will raise an exception.
-        """
-        if not self._is_deep_copied:
-            raise ValueError(
-                "replace child can only be called on a deep copied plan node."
-            )
-
-        source_plan = self.snowflake_plan.source_plan
-        # if source_plan is same as self, then we don't have children nodes
-        if source_plan is None or source_plan == self:
-            raise ValueError(
-                "old_node to be replaced is not found in the children nodes."
-            )
-
-        source_plan.replace_child(old_node, new_node)
 
     @property
     def column_states(self) -> ColumnStateDict:
@@ -862,19 +843,6 @@ class SelectStatement(Selectable):
         )
         return complexity
 
-    def replace_child(self, old_node, new_node) -> None:
-        if not self._is_deep_copied:
-            raise ValueError(
-                "replace child can only be called on a deep copied plan node."
-            )
-
-        if self.from_ != old_node:
-            raise ValueError(
-                "old_node to be replaced is not found in the children nodes."
-            )
-
-        self.from_ = new_node
-
     def to_subqueryable(self) -> "Selectable":
         """When this SelectStatement's subquery is not subqueryable (can't be used in `from` clause of the sql),
         convert it to subqueryable and create a new SelectStatement with from_ being the new subqueryable。
@@ -1291,18 +1259,6 @@ class SetStatement(Selectable):
     def individual_node_complexity(self) -> Dict[PlanNodeCategory, int]:
         # we add #set_operands - 1 additional operators in sql query
         return {PlanNodeCategory.SET_OPERATION: len(self.set_operands) - 1}
-
-    def replace_child(self, old_node, new_node) -> None:
-        if not self._is_deep_copied:
-            raise ValueError(
-                "replace child can only be called on a deep copied plan node."
-            )
-
-        if old_node not in self._nodes:
-            raise ValueError(
-                "old_node to be replaced is not found in the children nodes."
-            )
-        self._nodes = [node if node != old_node else new_node for node in self._nodes]
 
 
 class DeriveColumnDependencyError(Exception):
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -232,6 +232,8 @@ class Selectable(LogicalPlan, ABC):
         self._cumulative_node_complexity: Optional[Dict[PlanNodeCategory, int]] = None
 
     def __eq__(self, other: "Selectable") -> bool:
+        if not isinstance(other, Selectable):
+            return False
         if self._id is not None and other._id is not None:
             return type(self) is type(other) and self._id == other._id
         else:
@@ -348,6 +350,17 @@ class Selectable(LogicalPlan, ABC):
             if self.snowflake_plan.source_plan
             else []
         )
+
+    def replace_child(self, old_node, new_node) -> None:
+        """Replaces a child node with a new node in the select statement from its children_plan_nodes."""
+        source_plan = self.snowflake_plan.source_plan
+        # if source_plan is same as self, then we don't have children nodes
+        if source_plan is None or source_plan == self:
+            raise ValueError(
+                "old_node to be replaced is not found in the children nodes."
+            )
+
+        source_plan.replace_child(old_node, new_node)
 
     @property
     def column_states(self) -> ColumnStateDict:
@@ -840,6 +853,14 @@ class SelectStatement(Selectable):
         )
         return complexity
 
+    def replace_child(self, old_node, new_node) -> None:
+        if self.from_ != old_node:
+            raise ValueError(
+                "old_node to be replaced is not found in the children nodes."
+            )
+
+        self.from_ = new_node
+
     def to_subqueryable(self) -> "Selectable":
         """When this SelectStatement's subquery is not subqueryable (can't be used in `from` clause of the sql),
         convert it to subqueryable and create a new SelectStatement with from_ being the new subqueryable。
@@ -1256,6 +1277,13 @@ class SetStatement(Selectable):
     def individual_node_complexity(self) -> Dict[PlanNodeCategory, int]:
         # we add #set_operands - 1 additional operators in sql query
         return {PlanNodeCategory.SET_OPERATION: len(self.set_operands) - 1}
+
+    def replace_child(self, old_node, new_node) -> None:
+        if old_node not in self._nodes:
+            raise ValueError(
+                "old_node to be replaced is not found in the children nodes."
+            )
+        self._nodes = [node if node != old_node else new_node for node in self._nodes]
 
 
 class DeriveColumnDependencyError(Exception):
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -60,7 +60,6 @@ from snowflake.snowpark._internal.analyzer.expression import (
 from snowflake.snowpark._internal.analyzer.schema_utils import analyze_attributes
 from snowflake.snowpark._internal.analyzer.snowflake_plan import Query, SnowflakePlan
 from snowflake.snowpark._internal.analyzer.snowflake_plan_node import (
-    LeafNode,
     LogicalPlan,
     SnowflakeTable,
 )
@@ -374,7 +373,7 @@ class Selectable(LogicalPlan, ABC):
         self._column_states = deepcopy(value)
 
 
-class SelectableEntity(Selectable, LeafNode):
+class SelectableEntity(Selectable):
     """Query from a table, view, or any other Snowflake objects.
     Mainly used by session.table().
     """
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -349,6 +352,25 @@ def children_plan_nodes(self) -> List[Union["Selectable", SnowflakePlan]]:
             else []
         )
 
+    def replace_child(self, old_node, new_node) -> None:
+        """This method is called during optimization stage to cut plan tree at a certain node.
+
+        It must only be called on a deep copied plan node, otherwise it will raise an exception.
+        """
+        if not self._is_deep_copied:
```
sfc-gh-yzou, line 360: i don't think we need this check, kind of wired. If user want to do an in-place update of child, they can still do it, the job here is to make sure the child update is down correctly.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -840,6 +862,19 @@ def individual_node_complexity(self) -> Dict[PlanNodeCategory, int]:
         )
         return complexity
 
+    def replace_child(self, old_node, new_node) -> None:
+        if not self._is_deep_copied:
```
sfc-gh-yzou, line 866: let's remove this deep copy check

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1257,6 +1292,18 @@ def individual_node_complexity(self) -> Dict[PlanNodeCategory, int]:
...
+            raise ValueError(
+                "replace child can only be called on a deep copied plan node."
+            )
+
+        if old_node not in self._nodes:
+            raise ValueError(
+                "old_node to be replaced is not found in the children nodes."
+            )
+        self._nodes = [node if node != old_node else new_node for node in self._nodes]
```
sfc-gh-yzou, line 1305: you will also need to update the operant selectable and the pre and post actions

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -60,6 +61,24 @@ def cumulative_node_complexity(self) -> Dict[PlanNodeCategory, int]:
     def cumulative_node_complexity(self, value: Dict[PlanNodeCategory, int]):
         self._cumulative_node_complexity = value
 
+    def replace_child(self, old_node, new_node) -> None:
```
sfc-gh-yzou, line 64: you will need to update other nodes like limit to make sure the other fields that is associated with child is updated correctly

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -32,6 +32,7 @@ class LogicalPlan:
     def __init__(self) -> None:
         self.children = []
```
sfc-gh-yzou, line 33: you also need to add update in the binary_plan_node.py to update the left and right correclty.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -291,6 +293,21 @@ def children_plan_nodes(self) -> List[Union["Selectable", "SnowflakePlan"]]:
         else:
             return []
 
+    def replace_child(self, old_node, new_node) -> None:
+        """This method is called during optimization stage to cut plan tree at a certain node.
```
sfc-gh-yzou, line 297: @sfc-gh-aalam here is something kind of bothers me, for regular logical plans like limit, snowflakeTables, the update seems fine to me, because the node is still valid. However, for resolved nodes like SnowflakePlan and Selectable, after the update the node will not be valid anymore, for example, the queries, post actions may change, provide such general method on the node its-self doesn't seem something safe, and when user will need to pay attention when use this method.

I am thinking shall we just move the replace child as one single utility function, and do it like
```
def replace_child(parent, child, new_child)
    if isinstance(parent, SelectStatement):
         xxxx
    if isintance(xxxx):
         xxxx 
```
and only use this during our compilation phase. The disadvantage is that if we added a new  logical node, we might miss the correct way for update if the developer didn't pay attention to it, but i think that issue can still remains as far as we already provided a default way to update the  node.

tests/unit/test_replace_child.py
```diff
@@ -0,0 +1,339 @@
...
+from snowflake.snowpark._internal.analyzer.table_function import TableFunctionExpression
+from snowflake.snowpark._internal.analyzer.unary_plan_node import Project
+
+old_plan = LogicalPlan()
+new_plan = LogicalPlan()
+irrelevant_plan = LogicalPlan()
+
+
+@pytest.mark.parametrize("using_snowflake_plan", [True, False])
+def test_logical_plan(using_snowflake_plan, mock_query):
```
sfc-gh-yzou, line 33: also test other logical plan types

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-jdu: I was talking to Afroz about it. When I was looking at this PR, I was not aware of that this method is only used to update the plan after deepcopy, so I feel in-place update is kinda dangerous, and it shouldn't be used. But if this is only used for compilation stage on plan that gets deep copied, then it's fine. So just to prevent developer from calling this function to update snowflake plan in other places accidentally, I feel a warning or such a check is necessary. If you think it's too much, maybe just a comment (like the one above in docstring) is enough.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: don't we plan to reset the snowflake plan after the plan tree has been restructured. The query generation should handle that step right?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-aalam: ohh right. this is going to be a much bigger update.

tests/unit/test_replace_child.py
sfc-gh-aalam: added

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-aalam: moved to utils.py and made the change there.

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -39,3 +51,64 @@ def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
         )
 
     return QueryGenerator(plan.session, snowflake_create_table_plan_info)
+
+
+def replace_child(
+    parent: PlanTreeNode, old_node: PlanTreeNode, new_node: Selectable
```
sfc-gh-yzou, line 57: old_node -> old_child to be more consistent

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -39,3 +51,64 @@ def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
...
+        replace_child(parent.source_plan, old_node, new_node)
+        return
+
+    if isinstance(parent, SelectStatement):
+        parent.from_ = new_node
+        return
+
+    if isinstance(parent, SetStatement):
+        parent._nodes = [
+            node if node != old_node else new_node for node in parent._nodes
```
sfc-gh-yzou, line 79: you also need to update the selectable of the operant of the setstatement

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -39,3 +51,64 @@ def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
...
+        assert parent.source_plan is not None
+        replace_child(parent.source_plan, old_node, new_node)
+        return
+
+    if isinstance(parent, SelectStatement):
+        parent.from_ = new_node
+        return
+
+    if isinstance(parent, SetStatement):
+        parent._nodes = [
```
sfc-gh-yzou, line 78: also if the new child is not selectable, we will have to create a selectable for it.

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -39,3 +51,64 @@ def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
...
+        parent.child = new_node
+        return
+
+    if isinstance(parent, BinaryNode):
+        parent.children = [
+            node if node != old_node else new_node for node in parent.children
+        ]
+        if parent.left == old_node:
+            parent.left = new_node
+        else:
```
sfc-gh-yzou, line 99: the left and right can be the one to be replaced if they happen to be the same node, so we can not do if else here

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -474,6 +479,9 @@ def __deepcopy__(self, memodict={}) -> "SnowflakePlan":  # noqa: B006
             # session object after deepcopy
             session=self.session,
         )
+        copied_plan._is_valid_for_replacement = True
+        if copied_source_plan:
```
sfc-gh-yzou, line 483: i am little bit worried that if the source plan is not in the format that one node is logical plan node, another node is snwoflake plan node, some node may not be marked correctly. However, let's keep it like this for now. we can remove this later if we run into problem

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-aalam: To convert a generic `PlanTreeNode` into a selectable, we would also have to add analyzer as an argument which I am not against. But I'm wondering if we can make the requirement for `new_child` more strict to only be a `SelectableEntity` since both in repeated subquery elimination and large query breakdown, we only replace a subtree with a `SelectableEntity`?

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -39,3 +58,86 @@ def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
...
+        return
+
+    if isinstance(parent, SetStatement):
+        parent._nodes = [
+            node if node != old_child else new_child for node in parent._nodes
+        ]
+        parent.set_operands = tuple(
+            operand
+            if operand.selectable != old_child
+            else SetOperand(new_child, operand.operator)
```
sfc-gh-yzou, line 95: instead of creating new one, we can direct update the selectable of the operand since it is an inplace update, in case other node is referring to the same operand

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -32,6 +32,9 @@ class LogicalPlan:
...
         self._cumulative_node_complexity: Optional[Dict[PlanNodeCategory, int]] = None
+        # This flag is used to determine if the current node is a valid candidate for
+        # replacement in plan tree during optimization stage using replace_child method.
```
sfc-gh-jdu, line 36: should we add a comment here for deepcopy?

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -3,17 +3,36 @@
...
+    TableMerge,
+    TableUpdate,
+)
+from snowflake.snowpark._internal.analyzer.unary_plan_node import UnaryNode
 from snowflake.snowpark._internal.compiler.query_generator import (
     QueryGenerator,
     SnowflakeCreateTablePlanInfo,
 )
 
+PlanTreeNode = Union[SnowflakePlan, Selectable, LogicalPlan]
```
sfc-gh-jdu, line 34: SnowflakePlan and Selectable are both subclasess of LogicalPlan

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -39,3 +58,86 @@ def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
         )
 
     return QueryGenerator(plan.session, snowflake_create_table_plan_info)
+
+
+def replace_child(
+    parent: PlanTreeNode, old_child: PlanTreeNode, new_child: SelectableEntity
```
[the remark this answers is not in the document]
sfc-gh-jdu: Why is the new_child SelectableEntity? Can't be other plan?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-aalam: added

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -39,3 +56,96 @@ def create_query_generator(plan: SnowflakePlan) -> QueryGenerator:
...
+    based on the parent node type.
+    """
+
+    def to_selectable(plan: LogicalPlan, analyzer: Analyzer) -> Selectable:
+        """Given a LogicalPlan, convert it to a Selectable."""
+        if isinstance(plan, Selectable):
+            return plan
+
+        snowflake_plan = analyzer.resolve(plan)
+        return SelectSnowflakePlan(snowflake_plan, analyzer=analyzer)
```
sfc-gh-aalam, line 80: @sfc-gh-yzou does this look right to you for handling CTE related nodes

sfc-gh-yzou (approved):

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: yes, and i think this handles selecable correctly also.

sfc-gh-jdu (approved):

PR #1994, opened 2024-07-29
[SNOW-1554591] Support pandas.Dataframe in session.create_dataframe

sfc-gh-lspiegelberg:
Adds support for `pandas.Dataframe` in `session.create_dataframe`. Effectively encodes in the IR information about the temporary table which stores the data of the `pandas.DataFrame` server-side.


src/snowflake/snowpark/session.py
```diff
@@ -2809,12 +2822,27 @@ def convert_row_to_list(
             and isinstance(origin_data, pandas.DataFrame)
             and isinstance(self._conn, MockServerConnection)
         ):
+            # MockServerConnection creates internally a table, and returns Table object (which inherits from Dataframe).
```
sfc-gh-vbudati, line 2825: nit
```suggestion
            # MockServerConnection internally creates a table, and returns Table object (which inherits from Dataframe).
```

tests/ast/data/Session.create_dataframe_from_pandas.test
```diff
@@ -0,0 +1,18 @@
...
+
+from snowflake.snowpark import Row
+
+from snowflake.snowpark.types import IntegerType, StringType, StructField
+
+import pandas as pd
+
+df = session.create_dataframe(pd.DataFrame([(1, 2, 3, 4)], columns=["a", "b", "c", "d"]))
+
+# Snowpark ignores provided schema for pandas Dataframe.
```
sfc-gh-vbudati, line 11: Would it be possible to also test for the warning with these tests?

sfc-gh-vbudati (approved):
Looks great!

sfc-gh-azwiegincew (approved):

tests/ast/data/Session.create_dataframe_from_pandas.test
sfc-gh-lspiegelberg: we could certainly add a test! I imagine this to come up in code review for merge back, so let's dicuss with Snowpark python team then on what we should do for this case.

PR #2002, opened 2024-07-30
SNOW-1458141 Implement Index.all and Index.any

sfc-gh-vbudati:

   

   Fixes SNOW-1458141

- Implemented `Index.all` and `Index.any`. 
- `all` should return True when all values are truthy, else False.
- `any` should return True when any value is truthy, else False.
- I reused the Series' implementation for these methods.
- Fixed SNOW-1017231.


tests/integ/modin/index/test_all_any.py
```diff
@@ -0,0 +1,66 @@
...
+    native_pd.Index([3, np.nan, 5], name="my_index"),
+    native_pd.Index([1.1, np.nan]),
+    native_pd.Index([5, None, 7]),
+    native_pd.Index(["a", "b", 1, 2]),
+    native_pd.Index(["a", "b", "c", "d"]),
+    native_pd.Index([5, None, 7]),
+]
+
+
+@pytest.mark.parametrize("native_index", NATIVE_INDEX_BOOL_INT_TEST_DATA)
```
sfc-gh-nkrishna, line 33: I think it would be better to parametrize any/all and just have 2 tests -- one for the negative test and one for the normal test since the code is pretty much replicated. For the negative even the error checking you can do with f-string

sfc-gh-nkrishna (approved):
LGTM (approving), just one suggestion

tests/integ/modin/index/test_all_any.py
sfc-gh-vbudati: done!

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -734,11 +733,32 @@ def all(self) -> None:
...
+        False, because 0 is considered False.
+
+        >>> pd.Index([0, 1, 2]).all()
+        False
         """
-        # TODO: SNOW-1458141 implement all
+        # If self is empty, the result for all is always True. This `self.empty` check results in one extra query.
+        # Passing an empty object through `query_compiler.all` results in the ValueError:
+        # The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
+        if self.empty:
```
sfc-gh-azhan, line 752: we can skip this check and just use to_series.all so we don't need this extra query right?

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -734,11 +733,32 @@ def all(self) -> None:
...
+            return True
+        return (
+            self.to_series()
+            ._query_compiler.all(axis=0, bool_only=None, skipna=None, **kwargs)
```
sfc-gh-azhan, line 756: why not just to_series().all(...)?

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -763,8 +783,28 @@ def any(self) -> None:
...
+
+        >>> index = pd.Index([0, 0, 0])
+        >>> index.any()
+        False
         """
-        # TODO: SNOW-1458141 implement any
+        # If self is empty, the result for any is always False. This `self.empty` check results in one extra query.
+        # Passing an empty object through `query_compiler.any` results in the ValueError:
+        # The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
+        if self.empty:
```
sfc-gh-azhan, line 800: similar suggestions to all()

docs/source/modin/supported/index_supported.rst
```diff
@@ -65,9 +65,9 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | DataFrame method            | Snowpark implemented? (Y/N/P/D) | Missing parameters               | Notes for current implementation                   |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``all``                     | N                               |                                  |                                                    |
+| ``all``                     | P                               |                                  | ``N`` for non-integer/boolean types.               |
```
sfc-gh-azhan, line 68:
```suggestion
| ``all``                     | P                               |                                  | ``N`` for non-numeric types.               |
```

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: That should work, I didn't think of that.

docs/source/modin/supported/index_supported.rst
sfc-gh-vbudati: I wrote non-integer/boolean because float types don't work

sfc-gh-azhan: why float does not work? It works in native pandas though.

sfc-gh-vbudati: It's because floats cannot be cast to boolean values in Snowflake (this is what the existing Series query compiler method does). I'm thinking maybe trying to round them or using some `IFF` statement might get around this? I'll try that out.

sfc-gh-azhan: if `pd.Series([1.1, 0.1, 3]).all()` does not work then we could fix it too.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: `to_series.all()` does not return the correct value - instead it raises the ValueError I mentioned in the comment above

sfc-gh-azhan: It sounds like a bug. Does `pd.Series([]).all()` works?

sfc-gh-azhan: I got a `NotImplementedError: Snowpark pandas all API doesn't yet support non-integer/boolean columns` error

sfc-gh-vbudati: Follow-up from in-person discussion - I fixed the bug when the Series is empty. Here's an old JIRA referring to it. https://snowflakecomputing.atlassian.net/browse/SNOW-1017231

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2102,9 +2102,18 @@ def _bool_reduce_helper(
         assert agg_func in ("all", "any")
 
         frame = self._modin_frame
+
+        # Even though it incurs an extra query, we must get the length of the index to prevent errors.
+        # Empty Series objects return True/False for all/any respectively, regardless of the Series dtype.
+        # Otherwise, Series([], dtype="object").all() would raise a NotImplementedError while
+        # Series([], dtype=bool).all() would work fine; in reality they both produce the same result, True.
+        empty_index = self.get_axis_len(axis=0) == 0
```
sfc-gh-vbudati, line 2110: I fixed the bug where empty Series worked incorrectly with all/any. None of this logic is needed anymore. We could get rid of `empty_index = self.get_axis_len(axis=0) == 0` but then empty Series of non-bool/int type would raise a NotImplementedError. However, this would reduce the query count everywhere by 1.

sfc-gh-vbudati: I got rid of this query - now `pd.Series([], dtype=bool)` works with all/any but `pd.Series([], dtype=float)` doesn't. One reason I did this is because it's inconvenient to modify lots of the existing test query counts where it's expecting only one query.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -708,8 +708,7 @@ def T(self, *args: Any, **kwargs: Any) -> Index:
         """
         return self
 
-    @index_not_implemented()
-    def all(self) -> None:
+    def all(self, *args, **kwargs) -> bool | ExtensionArray:
```
sfc-gh-azhan, line 711: can you add one sentence in the docstring below saying `*args, **kwargs` are added for compatibility but not used in Snowpark.

sfc-gh-azhan (approved):

PR #2004, opened 2024-07-30
[SNOW-1491295] Support DataFrame.agg in Snowpark IR

sfc-gh-lspiegelberg:
Support DataFrame.agg in Snowpark IR

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -1732,9 +1732,6 @@ class DataFrame:
 
         df = self.group_by(_emit_ast=False).agg(*exprs, _emit_ast=False)
 
-        # if _emit_ast:
-        #     stmt.var_id.bitfield1 = self._ast_id
-
         return df
 
     @df_to_relational_group_df_api_usage
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -1655,6 +1655,7 @@ class DataFrame:
     def agg(
         self,
         *exprs: Union[Column, Tuple[ColumnOrName, str], Dict[str, str]],
+        _emit_ast: bool = True,
     ) -> "DataFrame":
         """Aggregate the data in the DataFrame. Use this method if you don't need to
         group the data (:func:`group_by`).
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -1713,7 +1714,25 @@ def agg(
...
+
+        # AST.
+        if _emit_ast:
+            stmt = self._session._ast_batch.assign()
+            expr = with_src_position(stmt.expr.sp_dataframe_agg, stmt)
+            self.set_ast_ref(expr.df)
+            exprs, is_variadic = parse_positional_args_to_list_variadic(*exprs)
+            for e in exprs:
+                if isinstance(e, Column):
+                    build_expr_from_snowpark_column_or_col_name(
```
sfc-gh-azwiegincew, line 1726: Unless things have changed from the last time I called the function, `build_expr_from_python_val()` includes this case, so you don't need the if.

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/dataframe.py
sfc-gh-heshah: If moving parameter type checking to the server side is better / the plan, then please disregard my comments below.

`build_expr_from_python_val` should still include the `Column` case, but I think the use of this method for handling both constants and `Column`s should be limited to APIs which need this functionality due to highly flexible input parameters (like in `TableFunctionCall`s, and others I think)

In most cases where the argument is expected to be a `Column` or `str` representing the column name, eventually Snowpark raises an exception if this isn't true. It might be better to have this check as `isinstance(e, (Column, str))` just so we can raise errors early in the thin client later.

sfc-gh-azwiegincew: Isn't this the case here though? We are literally allowing e to be any of the types supported by the more flexible function, so why have the branch?

sfc-gh-heshah: Sorry, yes, that makes sense. The same argument typing is present in `RelationalGroupedDataFrame.agg`, but I was under the impression that I had done something similar for checking all three possible argument types. Looks like I just called `build_expr_from_python_val` myself 😅

sfc-gh-azwiegincew: We might to optionally pass the allowed set of types to `build_expr_from_python_val()` to take care of the issue at a more fundamental level. I'm assuming we'll significantly revamp validation when we go to Phase 1 and lose the existing client-side validation.

sfc-gh-lspiegelberg: makes sense, let me remove the if statement to simplify the logic then.

PR #2012, opened 2024-07-31
SNOW-1558877: Add Snowpark pandas DatetimeIndex class

sfc-gh-nkumar:

   

   Fixes SNOW-NNNNNNN

   - Add DatetimeIndex class (which is a subclass of pd.Index)
   - Use self.__constructor__ in index.py to return correct Snowpark pandas index type.
   TODO (changedoc)

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
@@ -35,7 +35,6 @@ from snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler import (
     SnowflakeQueryCompiler,
 )
 from snowflake.snowpark.modin.plugin.extensions.index import Index
-from snowflake.snowpark.modin.plugin.utils.warning_message import WarningMessage
 
 _CONSTRUCTOR_DEFAULTS = {
     "freq": lib.no_default,
@@ -131,6 +130,13 @@ class DatetimeIndex(Index):
         >>> idx
         DatetimeIndex(['2020-01-01 02:00:00-08:00', '2020-02-01 03:00:00-08:00'], dtype='datetime64[ns, America/Los_Angeles]', freq=None)
         """
+        if isinstance(data, SnowflakeQueryCompiler):
+            # Raise error if underlying type is not a TimestampType.
+            dtype = data.index_dtypes[0]
+            if not dtype == np.dtype("datetime64[ns]"):
+                raise ValueError(
+                    "DatetimeIndex can only be created from a query compiler with TimestampType."
+                )
         kwargs = {
             "freq": freq,
             "tz": tz,
@@ -143,19 +149,4 @@ class DatetimeIndex(Index):
             "copy": copy,
             "name": name,
         }
-        if isinstance(data, SnowflakeQueryCompiler):
-            # Raise error if underlying type is not a TimestampType.
-            dtype = data.index_dtypes[0]
-            if not dtype == np.dtype("datetime64[ns]"):
-                raise ValueError(
-                    "DatetimeIndex can only be created from a query compiler with TimestampType."
-                )
-            # Raise warning if `data` is query compiler with non-default arguments.
-            for arg_name, arg_value in kwargs.items():
-                if arg_value is not _CONSTRUCTOR_DEFAULTS[arg_name]:
-                    WarningMessage.ignored_argument(
-                        "DatetimeIndex",
-                        arg_name,
-                        "Non-default arguments are not supported for DatetimeIndex with query compiler.",
-                    )
-        self.set_query_compiler(data=data, **kwargs)
+        self._init_index(data, _CONSTRUCTOR_DEFAULTS, **kwargs)
--- a/src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
@@ -26,6 +26,7 @@ Module houses ``DatetimeIndex`` class, that is distributed version of
 
 from __future__ import annotations
 
+import numpy as np
 import pandas as native_pd
 from pandas._libs import lib
 from pandas._typing import ArrayLike, Dtype, Frequency, Hashable, TimeAmbiguous
@@ -35,7 +36,6 @@ from snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler import (
 )
 from snowflake.snowpark.modin.plugin.extensions.index import Index
 from snowflake.snowpark.modin.plugin.utils.warning_message import WarningMessage
-from snowflake.snowpark.types import TimestampType
 
 _CONSTRUCTOR_DEFAULTS = {
     "freq": lib.no_default,
@@ -145,10 +145,8 @@ class DatetimeIndex(Index):
         }
         if isinstance(data, SnowflakeQueryCompiler):
             # Raise error if underlying type is not a TimestampType.
-            datatype = data._modin_frame.quoted_identifier_to_snowflake_type()[
-                data._modin_frame.index_column_snowflake_quoted_identifiers[0]
-            ]
-            if not isinstance(datatype, TimestampType):
+            dtype = data.index_dtypes[0]
+            if not dtype == np.dtype("datetime64[ns]"):
                 raise ValueError(
                     "DatetimeIndex can only be created from a query compiler with TimestampType."
                 )
--- a/src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
@@ -1,0 +1,163 @@
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
+# Licensed to Modin Development Team under one or more contributor license agreements.
+# See the NOTICE file distributed with this work for additional information regarding
+# copyright ownership.  The Modin Development Team licenses this file to you under the
+# Apache License, Version 2.0 (the "License"); you may not use this file except in
+# compliance with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+# Code in this file may constitute partial or total reimplementation, or modification of
+# existing code originally distributed by the Modin project, under the Apache License,
+# Version 2.0.
+
+"""
+Module houses ``DatetimeIndex`` class, that is distributed version of
+``pandas.DatetimeIndex``.
+"""
+
+from __future__ import annotations
+
+import pandas as native_pd
+from pandas._libs import lib
+from pandas._typing import ArrayLike, Dtype, Frequency, Hashable, TimeAmbiguous
+
+from snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler import (
+    SnowflakeQueryCompiler,
+)
+from snowflake.snowpark.modin.plugin.extensions.index import Index
+from snowflake.snowpark.modin.plugin.utils.warning_message import WarningMessage
+from snowflake.snowpark.types import TimestampType
+
+_CONSTRUCTOR_DEFAULTS = {
+    "freq": lib.no_default,
+    "tz": lib.no_default,
+    "normalize": lib.no_default,
+    "closed": lib.no_default,
+    "ambiguous": "raise",
+    "dayfirst": False,
+    "yearfirst": False,
+    "dtype": None,
+    "copy": False,
+    "name": None,
+}
+
+
+class DatetimeIndex(Index):
+
+    # Equivalent index type in native pandas
+    _NATIVE_INDEX_TYPE = native_pd.DatetimeIndex
+
+    def __new__(cls, *args, **kwargs):
+        """
+        Create new instance of DatetimeIndex. This overrides behavior of Index.__new__.
+        Args:
+            *args: arguments.
+            **kwargs: keyword arguments.
+
+        Returns:
+            New instance of DatetimeIndex.
+        """
+        return object.__new__(cls)
+
+    def __init__(
+        self,
+        data: ArrayLike | SnowflakeQueryCompiler | None = None,
+        freq: Frequency | lib.NoDefault = _CONSTRUCTOR_DEFAULTS["freq"],
+        tz=_CONSTRUCTOR_DEFAULTS["tz"],
+        normalize: bool | lib.NoDefault = _CONSTRUCTOR_DEFAULTS["normalize"],
+        closed=_CONSTRUCTOR_DEFAULTS["closed"],
+        ambiguous: TimeAmbiguous = _CONSTRUCTOR_DEFAULTS["ambiguous"],
+        dayfirst: bool = _CONSTRUCTOR_DEFAULTS["dayfirst"],
+        yearfirst: bool = _CONSTRUCTOR_DEFAULTS["yearfirst"],
+        dtype: Dtype | None = _CONSTRUCTOR_DEFAULTS["dtype"],
+        copy: bool = _CONSTRUCTOR_DEFAULTS["copy"],
+        name: Hashable | None = _CONSTRUCTOR_DEFAULTS["name"],
+    ) -> None:
+        """
+        Immutable ndarray-like of datetime64 data.
+
+        Parameters
+        ----------
+        data : array-like (1-dimensional) or snowflake query compiler
+            Datetime-like data to construct index with.
+        freq : str or pandas offset object, optional
+            One of pandas date offset strings or corresponding objects. The string
+            'infer' can be passed in order to set the frequency of the index as the
+            inferred frequency upon creation.
+        tz : pytz.timezone or dateutil.tz.tzfile or datetime.tzinfo or str
+            Set the Timezone of the data.
+        normalize : bool, default False
+            Normalize start/end dates to midnight before generating date range.
+        closed : {'left', 'right'}, optional
+            Set whether to include `start` and `end` that are on the
+            boundary. The default includes boundary points on either end.
+        ambiguous : 'infer', bool-ndarray, 'NaT', default 'raise'
+            When clocks moved backward due to DST, ambiguous times may arise.
+            For example in Central European Time (UTC+01), when going from 03:00
+            DST to 02:00 non-DST, 02:30:00 local time occurs both at 00:30:00 UTC
+            and at 01:30:00 UTC. In such a situation, the `ambiguous` parameter
+            dictates how ambiguous times should be handled.
+
+            - 'infer' will attempt to infer fall dst-transition hours based on
+              order
+            - bool-ndarray where True signifies a DST time, False signifies a
+              non-DST time (note that this flag is only applicable for ambiguous
+              times)
+            - 'NaT' will return NaT where there are ambiguous times
+            - 'raise' will raise an AmbiguousTimeError if there are ambiguous times.
+        dayfirst : bool, default False
+            If True, parse dates in `data` with the day first order.
+        yearfirst : bool, default False
+            If True parse dates in `data` with the year first order.
+        dtype : numpy.dtype or DatetimeTZDtype or str, default None
+            Note that the only NumPy dtype allowed is `datetime64[ns]`.
+        copy : bool, default False
+            Make a copy of input ndarray.
+        name : label, default None
+            Name to be stored in the index.
+
+        Examples
+        --------
+        >>> idx = pd.DatetimeIndex(["1/1/2020 10:00:00+00:00", "2/1/2020 11:00:00+00:00"], tz="America/Los_Angeles")
+        >>> idx
+        DatetimeIndex(['2020-01-01 02:00:00-08:00', '2020-02-01 03:00:00-08:00'], dtype='datetime64[ns, America/Los_Angeles]', freq=None)
+        """
+        kwargs = {
+            "freq": freq,
+            "tz": tz,
+            "normalize": normalize,
+            "closed": closed,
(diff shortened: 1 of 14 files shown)


src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
```diff
@@ -0,0 +1,124 @@
...
+        tz=lib.no_default,
+        normalize: bool | lib.NoDefault = lib.no_default,
+        closed=lib.no_default,
+        ambiguous: TimeAmbiguous = "raise",
+        dayfirst: bool = False,
+        yearfirst: bool = False,
+        dtype: Dtype | None = None,
+        copy: bool = False,
+        name: Hashable | None = None,
+        convert_to_lazy: bool = True,
```
sfc-gh-yzou, line 57: @sfc-gh-azhan @sfc-gh-vbudati i am always confused by this convert_to_lazy parameter, is that parameter only used when data is not a SnowflakeQueryCompiler, or even if it is a SnowflakeQueryCompiler, we will convert it to local if convert_to_lazy is true?
And what is the benefit of keeping some index object as local, why not just always pushing it  to remote like what we did for dataframe or series? If the argument is if the data is already in local, then i think the same argument also holds if the given data is not a SnowflakeQueryCompiler

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
```diff
@@ -0,0 +1,124 @@
...
+        >>> idx
+        DatetimeIndex(['2020-01-01 02:00:00-08:00', '2020-02-01 03:00:00-08:00'], dtype='datetime64[ns, America/Los_Angeles]', freq=None)
+        """
+        kwargs = {
+            "freq": freq,
+            "tz": tz,
+            "normalize": normalize,
+            "closed": closed,
+            "ambiguous": ambiguous,
+            "dayfirst": dayfirst,
```
sfc-gh-yzou, line 114: I assume when it is lazy, all those parameter information will get lost, or even ignored if the input is already a snowflake query compiler.  maybe we can a warning about this?

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
```diff
@@ -0,0 +1,124 @@
...
+            "yearfirst": yearfirst,
+            "dtype": dtype,
+            "copy": copy,
+            "name": name,
+        }
+        self.is_lazy = convert_to_lazy
+        if self.is_lazy:
+            self.set_query_compiler(data=data, **kwargs)
```
sfc-gh-yzou, line 122: if the original data is already a query compiler, shall we aways call a to_timestamp to convert it to timestamp type to make the index valid?

tests/integ/modin/index/test_equals.py
```diff
@@ -10,36 +10,81 @@
 from tests.integ.modin.sql_counter import SqlCounter, sql_count_checker
 
 
+def to_snow_index(native_index: native_pd.Index, convert_to_lazy: bool):
+    """
```
sfc-gh-yzou, line 14: do we have any test that test what happens if the data for datetime index is query compiler and the type is not timestamp?

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -523,11 +512,23 @@ def astype(self, dtype: str | type | ExtensionDtype, copy: bool = True) -> Index
...
+        result_class = Index
+        if dtype == "datetime64[ns]":
+            # local import to avoid circular dependency.
+            from snowflake.snowpark.modin.plugin.extensions.datetime_index import (
+                DatetimeIndex,
+            )
+
+            result_class = DatetimeIndex
+
+        if not self.is_lazy:
```
sfc-gh-yzou, line 524: shouldn't is_lazy_check already handles this correctly

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-azhan: It is only used for `df.columns`. Given this can be used quite often, we take this as the only special case to avoid unnecessary round trip.

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-azhan: Agree, to simplify, we raise a warning or even a NotImplementedError if those arguments are given with data is qc.

sfc-gh-azhan: Actually, let's raise NotImplementedError for this case (qc + kwargs) for now and track it with a new ticket. This can be done by reusing `pd.to_datetime` and we can implement it in another PR.

Note that @sfc-gh-vbudati 's rename PR will refactor the init of Index to change qc to Dataframe/Series so we can track the original df/series. We will need to update here too in the future.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -523,11 +512,23 @@ def astype(self, dtype: str | type | ExtensionDtype, copy: bool = True) -> Index
             # Ensure that self.astype(self.dtype) is self
             return self.copy() if copy else self
 
+        result_class = Index
+        if dtype == "datetime64[ns]":
```
sfc-gh-azhan, line 516: not just this but `is_datetime64_any_dtype`

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1584,7 +1584,13 @@ def index(self) -> Union["pd.Index", native_pd.MultiIndex]:
         if self.is_multiindex():
             return self._modin_frame.index_columns_pandas_index
         else:
-            return pd.Index(self)
+            index_datatype = self._modin_frame.quoted_identifier_to_snowflake_type()[
+                self._modin_frame.index_column_snowflake_quoted_identifiers[0]
+            ]
+            if index_datatype == TimestampType:
```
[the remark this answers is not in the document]
sfc-gh-nkumar: Interesting. So python constructor can return object of different class type, TIL.

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-nkumar: I think we should just raise warning if data is query_compiler and any argument is not default.

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-nkumar: I think we should not do unnecessary casting. For correctness we can add check that query_compiler passed here must be of type datetime.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-nkumar: Ack

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-nkumar: Not in this case. is_lazy_check will end up creating `Index` when target dtype is datetime but we want `DatetimeIndex`

tests/integ/modin/index/test_equals.py
sfc-gh-nkumar: Good point. Let me add a test with assert check.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -440,7 +476,7 @@ def unique(self, level: Hashable | None = None) -> Index:
             raise IndexError(
                 f"Too many levels: Index has only 1 level, {level} is not a valid level number."
             )
-        return Index(
+        return self.__constructor__(
```
sfc-gh-azhan, line 479: CMIIW, since init or new will return DatetimeIndex if the type is datetime64 like, then we could always use Index(..) here right?

tests/integ/modin/index/test_index_methods.py
```diff
@@ -21,10 +22,24 @@
...
+    if isinstance(native_index, native_pd.DatetimeIndex):
+        return pd.DatetimeIndex(native_index)
+    return pd.Index(native_index)
+
+
 @sql_count_checker(query_count=2)
 @pytest.mark.parametrize("native_index", NATIVE_INDEX_TEST_DATA)
 def test_index_copy(native_index):
-    snow_index = pd.Index(native_index)
+    snow_index = to_snow_index(native_index)
```
sfc-gh-azhan, line 42: Can you double check `pd.Index(pd.DatetimeIndex([123]))` should still return a DatetimeIndex? 
Because `native_pd.Index(native_pd.DatetimeIndex([123]))` return DatetimeIndex.

So without any change, this test should work too.

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-yzou: Let's take that offline, there seems some problem for this usage

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -106,6 +108,57 @@ def check_lazy(*args: Any, **kwargs: Any) -> Any:
...
+        convert_to_lazy : bool (default: True)
+            When True, create a lazy index object from a local data input, otherwise, create an index object that saves a pandas index locally.
+            We only set convert_to_lazy as False to avoid pulling data back and forth from Snowflake, e.g., when calling df.columns, the column data should always be kept locally.
+        """
+        from snowflake.snowpark.modin.plugin.extensions.datetime_index import (
+            DatetimeIndex,
+        )
+
+        if isinstance(data, SnowflakeQueryCompiler):
+            datatype = data._modin_frame.quoted_identifier_to_snowflake_type()[
```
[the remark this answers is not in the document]
sfc-gh-nkumar: I added warning in DatetimeIndex. Is the recommendation here to add warning or assert check or ValueError?

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-nkumar: Yes, that's right.

tests/integ/modin/index/test_index_methods.py
sfc-gh-nkumar: Added test.

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
```diff
@@ -0,0 +1,166 @@
...
+            dtype = data.index_dtypes[0]
+            if not dtype == np.dtype("datetime64[ns]"):
+                raise ValueError(
+                    "DatetimeIndex can only be created from a query compiler with TimestampType."
+                )
+            # Raise warning if `data` is query compiler with non-default arguments.
+            for arg_name, arg_value in kwargs.items():
+                if arg_value is not _CONSTRUCTOR_DEFAULTS[arg_name]:
+                    WarningMessage.ignored_argument(
+                        "DatetimeIndex",
```
sfc-gh-yzou, line 158: i am little bit confused here about this warning message, it says we do support the default argument, what would be the behavior if user is providing a query compiler and also freq or tz argument?

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-yzou: seems we are going to clean up those local index behavior, let's left a comment here for clean up then

sfc-gh-yzou: you can put this ticket number here https://snowflakecomputing.atlassian.net/browse/SNOW-1573193

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -106,13 +114,62 @@ def check_lazy(*args: Any, **kwargs: Any) -> Any:
...
+            When True, attempt to create a MultiIndex if possible.
+        convert_to_lazy : bool (default: True)
+            When True, create a lazy index object from a local data input, otherwise, create an index object that saves a pandas index locally.
+            We only set convert_to_lazy as False to avoid pulling data back and forth from Snowflake, e.g., when calling df.columns, the column data should always be kept locally.
+        """
+        from snowflake.snowpark.modin.plugin.extensions.datetime_index import (
+            DatetimeIndex,
+        )
+
+        if isinstance(data, SnowflakeQueryCompiler):
```
sfc-gh-yzou, line 155: can you add a comment here that a DatetimeIndex object will be returned if the column/date have datetime type?

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-nkumar: > i think it would be better to have consistent behavior, we either error out for both cases or allow it for both cases.

I didn't understand what 2 cases you are referring to. Can you please provide examples?

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-nkumar: Ack

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-nkumar: Ack

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -92,17 +150,34 @@ def __init__(
...
+        self,
+        data: ArrayLike | SnowflakeQueryCompiler | None,
+        ctor_defaults: dict,
+        **kwargs: Any,
+    ):
+        if isinstance(data, SnowflakeQueryCompiler):
+            # Raise warning if `data` is query compiler with non-default arguments.
+            for arg_name, arg_value in kwargs.items():
+                if arg_value is not ctor_defaults[arg_name]:
+                    WarningMessage.ignored_argument(
```
sfc-gh-azhan, line 171: I think we should raise assertion error since qc data is only used by us not the user.

tests/integ/modin/index/test_equals.py
```diff
@@ -10,28 +10,77 @@
...
+        ),  # different tz
     ],
 )
 def test_index_equals(lhs, rhs, expected):
     with SqlCounter(query_count=2, join_count=2):
         native_result = native_pd.Index(lhs).equals(native_pd.Index(rhs))
         assert native_result == expected
 
-        snow_result = pd.Index(lhs).equals(pd.Index(rhs))
+        snow_result = to_snow_index(lhs).equals(pd.Index(rhs))
```
sfc-gh-azhan, line 80: we don't have to change this right? `pd.Index(pd.DatetimeIndex(...))` should be a `DatetimeIndex`

tests/integ/modin/index/test_index_methods.py
sfc-gh-azhan, line 42: same here

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-nkumar: Added assert

tests/integ/modin/index/test_equals.py
sfc-gh-nkumar: done

tests/integ/modin/index/test_index_methods.py
sfc-gh-nkumar: done

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-yzou: Right now, if we do pd.DatetimeIndex([1, 2, 3]), this works fine, because we call native_pd.DatetimeIndex([1, 2, 3]) with will convert the integer type to datetime to create a valid DatetimeIndex, and then upload the data to snowflake.
However, if we have an integer column "A" in snowflake, and we create a DateTimeIndex out of that column, it raises an error.
@sfc-gh-azhan i think we have discussed that offline, and we think if user is calling DateTimeIndex constructer on an integer or string column, it is like saying they want to convert the column to timestamp column to construct a valid DatetimeIndex, and we can follow up to make this work in the next pr.
@sfc-gh-nkumar for now can we raise a NotImplementedError instead of ValueError, and add a TODO to follow this up.

sfc-gh-azhan: More specifically, those cases are not well supported in this PR IMO:
`pd.DatetimeIndex(pd.Series([1,2,3]))` or `pd.DatetimeIndex(pd.Index([1,2,3]))`.
We could put those cases in next PR and reuse `pd.to_datetime()` for them.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -92,17 +150,31 @@ def __init__(
...
+    def _init_index(
+        self,
+        data: ArrayLike | SnowflakeQueryCompiler | None,
+        ctor_defaults: dict,
+        **kwargs: Any,
+    ):
+        if isinstance(data, SnowflakeQueryCompiler):
+            # Raise warning if `data` is query compiler with non-default arguments.
+            for arg_name, arg_value in kwargs.items():
+                assert (
```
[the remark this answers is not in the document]
sfc-gh-azhan: SnowflakeQueryCompiler are only used internally so I suggest using assertion.

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
```diff
@@ -0,0 +1,152 @@
...
+        --------
+        >>> idx = pd.DatetimeIndex(["1/1/2020 10:00:00+00:00", "2/1/2020 11:00:00+00:00"], tz="America/Los_Angeles")
+        >>> idx
+        DatetimeIndex(['2020-01-01 02:00:00-08:00', '2020-02-01 03:00:00-08:00'], dtype='datetime64[ns, America/Los_Angeles]', freq=None)
+        """
+        if isinstance(data, SnowflakeQueryCompiler):
+            # Raise error if underlying type is not a TimestampType.
+            current_dtype = data.index_dtypes[0]
+            if not current_dtype == np.dtype("datetime64[ns]"):
+                raise ValueError(
```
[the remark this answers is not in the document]
sfc-gh-azhan: SnowflakeQueryCompiler are only used internally so I suggest using assertion.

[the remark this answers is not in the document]
sfc-gh-azhan: SnowflakeQueryCompiler are only used internally so I suggest using assertion.

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-nkumar: First case is interesting because with Varnikas recent change series passed in will be treated as parent not actual data, we might have to rethink our design of passing parent to Index. 
I will add a failing test for second test with a follow up JIRA.

sfc-gh-nkumar: follow up JIRA https://snowflakecomputing.atlassian.net/browse/SNOW-1616989

tests/integ/modin/index/test_index_methods.py
```diff
@@ -23,6 +23,20 @@
 )
 
 
+def to_snow_index(native_index):
```
sfc-gh-azhan, line 26: can be removed.

tests/integ/modin/series/test_loc.py
```diff
@@ -1749,11 +1749,11 @@ def test_series_non_partial_string_indexing_cases(ops, error):
             eval_snowpark_pandas_result(snowpark_str, native_str, ops, check_freq=False)
 
 
-@sql_count_checker(query_count=2)
+@sql_count_checker(query_count=3)
```
sfc-gh-azhan, line 1752: do you know why now we have one extra query?

sfc-gh-nkumar: Fixed.

tests/integ/modin/index/test_index_methods.py
sfc-gh-nkumar: done

sfc-gh-azhan (commented):
Can you do a rebase and we can take a look again of the constructor?

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
```diff
@@ -0,0 +1,152 @@
...
+        name : label, default None
+            Name to be stored in the index.
+
+        Examples
+        --------
+        >>> idx = pd.DatetimeIndex(["1/1/2020 10:00:00+00:00", "2/1/2020 11:00:00+00:00"], tz="America/Los_Angeles")
+        >>> idx
+        DatetimeIndex(['2020-01-01 02:00:00-08:00', '2020-02-01 03:00:00-08:00'], dtype='datetime64[ns, America/Los_Angeles]', freq=None)
+        """
+        if isinstance(data, SnowflakeQueryCompiler):
```
sfc-gh-yzou, line 133: @sfc-gh-azhan as we discussed offline, we will remove the support of SnowflakeQueryCompiler, but use Snowpark series or Snowpark Dataframe on the interface to avoid confusion to users.   And the error should be NotImplementedError when it is Snowpark Series or Snowpark DataFrame.

For now it is ok, @sfc-gh-azhan can we get a ticket to track the work?

sfc-gh-yzou (approved):

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-azhan: https://snowflakecomputing.atlassian.net/browse/SNOW-1618623
I should be able to have a PR to review today.

sfc-gh-azhan (approved):
Discuss offline, some constructor argument refactoring will be done once this is in.

PR #2017, opened 2024-08-01
New script update-from-local.sh that supports local development; similar to update-from-devvm.sh

sfc-gh-azwiegincew:


src/snowflake/snowpark/_internal/proto/update-from-local.sh
```diff
@@ -0,0 +1,31 @@
...
+  then
+    echo "Using source root \"$1\""
+    SRC_ROOT=$1
+fi
+set -euxo pipefail
+
+SCRIPT_DIR=$(dirname "$0")
+
+# Step 1: Build the python proto file from scratch via bazel
+pushd $SRC_ROOT/Snowpark
```
sfc-gh-lspiegelberg, line 21:
```pushd ... > /dev/null && \
./make_ast.sh && popd
```
else, when make_ast.sh fails you're stuck in the wrong dir...

sfc-gh-azwiegincew: Isn't it the opposite? We want to popd in all cases, not just when make_ast succeeds.

src/snowflake/snowpark/_internal/proto/update-from-local.sh
```diff
@@ -0,0 +1,31 @@
...
+  then
+    echo "No source directory provided, defaulting to \"$SRC_ROOT\""
+elif [ -n "$1" ]
+  then
+    echo "Using source root \"$1\""
+    SRC_ROOT=$1
+fi
+set -euxo pipefail
+
+SCRIPT_DIR=$(dirname "$0")
```
sfc-gh-lspiegelberg, line 18: To execute relative, this is pattern
```
# execute relative to script
SCRIPT_LOCATION=$(dirname "$(readlink -f "$BASH_SOURCE")")
cd $SCRIPT_LOCATION
```

sfc-gh-lspiegelberg (approved):
couple nits, mainly for some weird bash edge cases.

PR #2023, opened 2024-08-01
SNOW-1570506 Fix `Series.argmax` and `Series.argmin`, add tests

sfc-gh-vbudati:

   

   Fixes SNOW-1570506

 Fixed the existing implementation of `Series.argmax` and `Series.argmin`. Previously, `np.isnan(result)` was not working since the wrong type of `result` was being provided. Instead, I changed this logic to directly see if result is some form of None/nan, and if it is return -1 (valid result does not exist).

```py
>>> s = pd.Series([13, 1, 2, None, 1], index=["A", "B", "C", "D", "E"])
>>> s.argmin(skipna=False)
-1
>>> s.argmin(skipna=True)
1
>>> s.argmax(skipna=False)
-1
>>> s.argmax(skipna=True)
0
```


sfc-gh-nkrishna (approved):
LGTM!

src/snowflake/snowpark/modin/pandas/series.py
```diff
@@ -781,25 +781,37 @@ def apply(
...
-        if np.isnan(result) or result is pandas.NA:
+        if self._query_compiler.has_multiindex():
+            # The index is a MultiIndex, current logic does not support this.
+            ErrorMessage.not_implemented(
+                "Series.argmax is not yet supported when the index is a MultiIndex."
+            )
+        result = self.reset_index(drop=True).idxmax(
+            axis=axis, skipna=skipna, *args, **kwargs
+        )
+        if not is_integer(result) and not result:  # if result=0, return 0 not -1
```
sfc-gh-evandenberg, line 797: Is this comment accurate?  It says if result=0 but I don't see that as a check here.

sfc-gh-evandenberg (approved):
LGTM with one question.

src/snowflake/snowpark/modin/pandas/series.py
sfc-gh-helmeleegy: The result=0 case is not captured by this condition, and so it's simply handled by the `return result` line below. What I'm not clear about is why one would think that if result=0, we return -1? Maybe remove this comment?

sfc-gh-helmeleegy, line 797: Why `and` and not `or` like the original implementation?

src/snowflake/snowpark/modin/pandas/series.py
```diff
@@ -781,25 +781,37 @@ def apply(
...
         """
         # TODO: SNOW-1063347: Modin upgrade - modin.pandas.Series functions
-        result = self.idxmax(axis=axis, skipna=skipna, *args, **kwargs)
-        if np.isnan(result) or result is pandas.NA:
+        if self._query_compiler.has_multiindex():
+            # The index is a MultiIndex, current logic does not support this.
+            ErrorMessage.not_implemented(
+                "Series.argmax is not yet supported when the index is a MultiIndex."
+            )
+        result = self.reset_index(drop=True).idxmax(
```
sfc-gh-helmeleegy, line 794: Given that we are calling `reset_index` here, it seems that `result` can never be None or non-int. So I'm wondering if the check below is ever satisfied.

sfc-gh-vbudati: result is None or a non-integer when idxmax does not find a matching value, so the check still needs to exist

src/snowflake/snowpark/modin/pandas/series.py
sfc-gh-vbudati: I think I should be able to reduce to just check `not is_integer(result)`

sfc-gh-vbudati: I'm not directly checking whether result is 0, rather checking if result is a valid integer. I should be able to reduce this check to just `not is_integer(result)` and it should work fine

sfc-gh-helmeleegy (approved):
Looks great, thanks Varnika!

PR #2028, opened 2024-08-03
SNOW-1558347: Fix PARSE_HEADER bug in session.read.csv

sfc-gh-jrose:

   

   Fixes SNOW-1558347

   Previously if a user specified `PARSE_HEADER = True` in an externally defined file format it would not be able to be used with `session.read.csv` without throwing an exception. This exception was caused due to how we infer csv schemas in the client. The client separates out csv loading into a schema inference step and a query step. Parsing the header is valid in the inference step, but should be skipped in the query step.

  This change modifies the logic to the read step so that a temporary file format is always created. The external file format is queried and merged with the local options giving the local options precedence. The local options are then modified as needed such as for the parse header issue above.

  I've added a test that provides a scenario that would have previously failed.

  These parts may be contentious:
  - `create_file_format_statement` is changed to only set the format if it wasn't specified in the options
  - `_merge_file_format_options` does some parsing of the options itself. I didn't see any other logic that currently handles that format. It should work for all currently existing formats, but it could break for future options if their formatting is significantly different.   
  - The new logic for `read_file` prefers the client specified options over the file_format options whereas previously the client format options would have been ignored. This is trivial to change, but I think the new style should make more sense to the user.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -1109,6 +1109,32 @@ class SnowflakePlanBuilder:
             source_plan,
         )
 
+    def _merge_file_format_options(
+        self, file_format_options: Dict[str, Any], options: Dict[str, str]
+    ) -> Dict[str, Any]:
+        """
+        Merges remotely defined file_format options with the local options set. This allows the client
+        to override any locally defined options that are incompatible with the file format.
+        """
+        if "FORMAT_NAME" not in options:
+            return file_format_options
+
+        type_map = {"String": str, "List": str, "Integer": int, "Boolean": bool}
+        new_options = {**file_format_options}
+        file_format = self.session.sql(
+            f"DESCRIBE FILE FORMAT {options['FORMAT_NAME']}"
+        ).collect()
+        for setting in file_format:
+            if (
+                setting.property_value == setting.property_default
+                or setting.property in new_options
+            ):
+                continue
+            new_options[setting.property] = type_map.get(setting.property_type, str)(
+                setting.property_value
+            )
+        return new_options
+
     def read_file(
         self,
         path: str,
@@ -1121,6 +1147,9 @@ class SnowflakePlanBuilder:
         metadata_schema: Optional[List[Attribute]] = None,
     ):
         format_type_options, copy_options = get_copy_into_table_options(options)
+        format_type_options = self._merge_file_format_options(
+            format_type_options, options
+        )
         pattern = options.get("PATTERN")
         # Can only infer the schema for parquet, orc and avro
         # csv and json in preview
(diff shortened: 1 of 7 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1039,6 +1039,45 @@ def create_or_replace_dynamic_table(
...
+
+        type_map = {
+            "String": str,
+            "List": process_list,
+            "Integer": int,
+            "Boolean": bool,
+        }
+        new_options = {**file_format_options}
+        file_format = self.session.sql(
+            f"DESCRIBE FILE FORMAT {options['FORMAT_NAME']}"
```
sfc-gh-jdu, line 1066: We may be careful about eagerly triggering such a query 1) it will be visible to users 2) maybe add a try/catch because it's still possible that it will fail.

sfc-gh-jrose: That's a good point. I think I can do this better in a way that maintains lazy evaluation. I'll refactor a bit.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1039,6 +1039,45 @@ def create_or_replace_dynamic_table(
...
+            if (
+                setting["property_value"] == setting["property_default"]
+                or setting["property"] in new_options
+            ):
+                continue
+
+            new_options[str(setting["property"])] = type_map.get(
+                str(setting["property_type"]), str
+            )(setting["property_value"])
+        return new_options
```
sfc-gh-aling, line 1079: I have some very basic questions:

1. you mentioned there are two steps -- infer schema and query, are both using the same temporary file format, or the temporary format is just for query

2. how does the code solve the parse_header issue, do we flip the value of parse_header during the query step? I don't see a special case of parse_header, if this is a generalized approach will this have side effect on other options?

3. instead of doing the fix for users, I saw doc about this parse_header, does it mean users are not supposed to set `PARSE_HEADER` when loading csv, it's their job to fix their file format?
```
This file format option is applied to the following actions only:
- Automatically detecting column definitions by using the INFER_SCHEMA function.
- Loading CSV data into separate columns by using the INFER_SCHEMA function and MATCH_BY_COLUMN_NAME copy option.
```

sfc-gh-jrose: 1. At the moment both the infer step and the query step will create their own temporary file format if needed. This PR does not force a temporary format for inference because that step does not seem to have any incompatibilities with the external file formats.
2. This change allows the existing logic [here](https://github.com/snowflakedb/snowpark-python/blob/7f99a294aa3bb8a4f533108095083d52ba174f32/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L1108-L1115) to be applied to an externally defined file format.
3. Our api only allows them to specify one file format for both inference and loading. Right now if they were to modify the file format to have parse_header be false they will not be able to load column names from the header at load time. This would is a poor experience for users.

sfc-gh-mvashishtha:
@sfc-gh-jrose are there any changes for Snowpark pandas reviewers to review here?

sfc-gh-jrose:
> @sfc-gh-jrose are there any changes for Snowpark pandas reviewers to review here?

No. I merged main incorrectly earlier which pulled in a snowpark pandas change. I've since removed that commit, but git still pulled in the group automatically.

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: Can you create a jira and add a todo here for doing it lazily in the future?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aling: thanks for the explanation!

sfc-gh-aling (approved):

PR #2034, opened 2024-08-06
DRAFT: New types: at InternalFrame by tracking types through SnowparkPandasColumn

sfc-gh-mvashishtha:

   

   Fixes SNOW-NNNNNNN


src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -60,6 +63,65 @@
 LEFT_PREFIX = "left"
 RIGHT_PREFIX = "right"
 
+def snowpark_pandas_col(name):
+    def to_type_and_snowpark_column(input_snowpark_pandas_columns_getter):
+        snowpark_label_to_type = input_snowpark_pandas_columns_getter()
+        return snowpark_label_to_type[name], SnowparkColumn(name)
+    return SnowparkPandasColumn(to_type_and_snowpark_column=to_type_and_snowpark_column)
+
+class SnowparkPandasColumn:
```
sfc-gh-yzou, line 72: I think the thought is can we replace SnowparkPandasColumn for all SnowparkColumn used in query_compiler and internal frame everywhere.
i think one problem is that for snowpark functions that takes a snowparkcolumn or name, how do they work. for the ones we will oeverwrite, that seems not a problem. For the ones that is not implemented, can they just fall back to snowpark column.

here it might be easier to provide SnowparkPandasColumn as an extension of Snowpark column, with an extra init parameter that is dtype.

PR #2038, opened 2024-08-06
[SNOW-1570952] Remove unnecessary workaround for query comparison

sfc-gh-yzou:
SNOW-1570952

Previously, we put an function for deduplication of queries to allow deduplication of same query text but different place_holder_id. However, when propagating those query, the same query comes from the same node, they will also have the same placeholder id, therefore, there is no need for this extra comparison function.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -1564,6 +1564,7 @@ class Query:
             self.sql == other.sql
             and self.query_id_place_holder == other.query_id_place_holder
             and self.is_ddl_on_temp_object == other.is_ddl_on_temp_object
+            and self.params == other.params
         )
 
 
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -613,25 +613,6 @@ def build_binary(
...
-        def _query_exists(current_query: Query, existing_queries: List[Query]) -> bool:
-            for existing_query in existing_queries:
-                if (
-                    (current_query.sql == existing_query.sql)
-                    and (
-                        current_query.is_ddl_on_temp_object
-                        == existing_query.is_ddl_on_temp_object
-                    )
-                    and (current_query.params == existing_query.params)
-                ):
```
sfc-gh-aalam, line 630: current `Query.__eq__` does not check for params equality. Is that still okay?

sfc-gh-yzou: i actually view this as a bug for query comparison, i just updated the query comparison to see if any test fails

sfc-gh-jdu (approved):

sfc-gh-aalam (approved):

PR #2039, opened 2024-08-06
SNOW-1615695: Push empty all/any processing down to query compiler

sfc-gh-joshi:

   

   Fixes SNOW-1615695

Snowpark pandas implements `df.any/all(axis=0)` via the BOOLAND_AGG and BOOLOR_AGG SQL functions. These functions return a single row of NULL if there are no elements, while pandas produces a vacuous True/False value. In the fixes for SNOW-1458141, these cases are handled in the frontend, but since the frontend changes don't make sense for other modin backends, they should be pushed down to the query compiler or lower.

This PR adds an IFNULL statement to wrap the result of BOOLAND_AGG/BOOLOR_AGG to handle the empty-row case without any additional queries. This returns the frontend all/any functions to match the modin implementations.

### DIFF
--- a/src/snowflake/snowpark/modin/pandas/base.py
+++ b/src/snowflake/snowpark/modin/pandas/base.py
@@ -925,7 +925,7 @@ class BasePandasDataset(metaclass=TelemetryMeta):
                 return data_for_compute.all(
                     axis=axis, bool_only=False, skipna=skipna, **kwargs
                 )
-            result = self._reduce_dimension(
+            return self._reduce_dimension(
                 self._query_compiler.all(
                     axis=axis, bool_only=bool_only, skipna=skipna, **kwargs
                 )
@@ -948,7 +948,7 @@ class BasePandasDataset(metaclass=TelemetryMeta):
                 return result.all(
                     axis=axis, bool_only=bool_only, skipna=skipna, **kwargs
                 )
-        return True if result is None else result
+            return result
 
     def any(self, axis=0, bool_only=None, skipna=True, **kwargs):
         """
@@ -969,7 +969,7 @@ class BasePandasDataset(metaclass=TelemetryMeta):
                 return data_for_compute.any(
                     axis=axis, bool_only=False, skipna=skipna, **kwargs
                 )
-            result = self._reduce_dimension(
+            return self._reduce_dimension(
                 self._query_compiler.any(
                     axis=axis, bool_only=bool_only, skipna=skipna, **kwargs
                 )
@@ -990,7 +990,7 @@ class BasePandasDataset(metaclass=TelemetryMeta):
                 return result.any(
                     axis=axis, bool_only=bool_only, skipna=skipna, **kwargs
                 )
-        return False if result is None else result
+            return result
 
     def apply(
         self,
--- a/src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
@@ -480,7 +480,7 @@ def get_snowflake_agg_func(
             return lambda col: column_quantile(col, interpolation, q)
         elif agg_func in ("all", "any"):
             # If there are no rows in the input frame, the function will also return NULL, which should
-            # instead by TRUE for "all" and FALSE for "any"."
+            # instead by TRUE for "all" and FALSE for "any".
             # Need to wrap column name in IDENTIFIER, or else the agg function will treat the name
             # as a string literal.
             default_value = bool(agg_func == "all")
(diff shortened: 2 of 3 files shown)


src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
```diff
@@ -478,6 +478,17 @@ def get_snowflake_agg_func(
                 # through the aggregate frontend in this manner is unsupported.
                 return None
             return lambda col: column_quantile(col, interpolation, q)
+        elif agg_func in ("all", "any"):
+            # If there are no rows in the input frame, the function will also return NULL, which should
+            # instead by TRUE for "all" and FALSE for "any"."
```
sfc-gh-vbudati, line 483:
```suggestion
            # instead by TRUE for "all" and FALSE for "any".
```
nit

sfc-gh-vbudati (approved):

src/snowflake/snowpark/modin/pandas/base.py
```diff
@@ -990,7 +990,7 @@ def any(self, axis=0, bool_only=None, skipna=True, **kwargs):
                 return result.any(
                     axis=axis, bool_only=bool_only, skipna=skipna, **kwargs
                 )
-        return False if result is None else result
```
sfc-gh-yzou, line 993: The original implementation returns a value when result is None, but now it always returns result, is there a behavior change?

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
```diff
@@ -478,6 +478,17 @@ def get_snowflake_agg_func(
...
             return lambda col: column_quantile(col, interpolation, q)
+        elif agg_func in ("all", "any"):
+            # If there are no rows in the input frame, the function will also return NULL, which should
+            # instead by TRUE for "all" and FALSE for "any".
+            # Need to wrap column name in IDENTIFIER, or else the agg function will treat the name
+            # as a string literal.
+            default_value = bool(agg_func == "all")
+            return lambda col: builtin("ifnull")(
+                # mypy refuses to acknowledge snowflake_agg_func is non-NULL here
+                snowflake_agg_func(builtin("identifier")(col)),  # type: ignore[misc]
```
sfc-gh-yzou, line 489: is the identifier the identifier for the final column? can we do the same as how other aggregation function is handled to use alias to generate the column name and also use the generate_identifier method to generate the new column name?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2129,22 +2129,21 @@ def _bool_reduce_helper(
...
-            # so we need to manually remove it
+            # If there are now rows (but there are columns), booland_agg/boolor_agg would return NULL.
+            # This behavior is handled within aggregation_utils to avoid an extra query.
             return self.agg(
                 agg_func,
                 axis=0,
                 args=[],
                 kwargs={"skipna": skipna},
-            ).set_columns([MODIN_UNNAMED_SERIES_LABEL])
+            )
```
sfc-gh-yzou, line 2146: why the column name is removed here?

src/snowflake/snowpark/modin/pandas/base.py
sfc-gh-joshi: Prior to #2002, it was not possible for the query compiler method to return None. In upstream Modin, it is also not possible for the result to be None, so this PR is going back to the behavior expected by the Modin frontend.

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
sfc-gh-joshi: This is the identifier of the input column (it basically expands to `BOOLAND_AGG(IDENTIFIER("column_name"))`. This previously was handled further down in `generate_aggregation_column`, but because we now need to wrap the aggregation result in an `IFNULL` I needed to move the logic here.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: In #1964 I added a call to `set_columns([MODIN_UNNAMED_SERIES_LABEL])` in the `agg` method to set the name after a transpose operation, so this call is now redundant.

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
sfc-gh-yzou: I see, can you add a comment above about how the correspnding sql look like?

sfc-gh-yzou (approved):

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
sfc-gh-joshi: added

PR #2040, opened 2024-08-06
[SNOW-1491309] Snowpark IR, support DataFrame.describe

sfc-gh-lspiegelberg:
Support DataFrame.describe for Snowpark IR. Adds preliminary stddev implementation to local testing API given this feature is missing at the moment.


src/snowflake/snowpark/mock/_functions.py
```diff
@@ -220,6 +220,39 @@ def mock_avg(column: ColumnEmulator) -> ColumnEmulator:
     return ColumnEmulator(data=[res], sf_type=ColumnType(res_type, False))
```
sfc-gh-lspiegelberg, line 222: used here basically the `avg` implementation and adjusted accordingly. Tests should be done as part of merge back

sfc-gh-oplaton (approved):

PR #2043, opened 2024-08-07
Fix opentelemetry test

sfc-gh-yuwang:

   

   snowpark daily precommit pipeline was broken because of opentelemetry test.
   There is no recent change of opentelemetry and the precommit test started to fail since https://github.com/snowflakedb/snowpark-python/pull/1987, the root cause is still mysterious to me, but I don't think we should spend a lot of time on digging it, so I moved them to integ test and removed redundant tests. I have confirmed that our code coverage is not affected.


sfc-gh-yuwang:
After consideration, I decided to move tests to integ test and remove redundant tests

sfc-gh-yzou:
I really doubt it is cause by the pr linked there, because that pr seems just adding a function that is not called by anyone. Also, the test actually seems flaky to me, I see sometimes it fails on windows, sometime on azure etc. if it is caused by that pr. If moving the test can make it stable, we can do that.

sfc-gh-yzou (approved):

sfc-gh-jrose (approved):

sfc-gh-yuwang:
> I really doubt it is cause by the pr linked there, because that pr seems just adding a function that is not called by anyone. Also, the test actually seems flaky to me, I see sometimes it fails on windows, sometime on azure etc. if it is caused by that pr. If moving the test can make it stable, we can do that.

I am also not sure what exactly caused it, the past daily precommit before that PR seems to be stable. I found it really hard to debug this flakiness because it only appears in daily precommit. So I just move all of them to integ test because that has always been stable, I also made sure that our code coverage does not drop

PR #2045, opened 2024-08-07
SNOW-1618623 lazy index: refactor qc as input for index constructor

sfc-gh-azhan:

   

   Fixes SNOW-1618623

lazy index: 

- refactor qc as input for index constructor and make it consistent with other init methos like dataframe and series.
- raise ValueError for `pd.Index(df)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@
 
 ### Behavior change
 - `Dataframe.columns` now returns native pandas Index object instead of Snowpark Index object.
+- Refactor and introduce `query_compiler` argument in `Index` constructor to create `Index` from query compiler.
 
 ## 1.20.0 (2024-07-17)
 
--- a/src/snowflake/snowpark/modin/plugin/extensions/index.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/index.py
@@ -63,11 +63,12 @@ class Index(metaclass=TelemetryMeta):
 
     def __new__(
         cls,
-        data: ArrayLike | SnowflakeQueryCompiler | None = None,
+        data: ArrayLike | native_pd.Index | Series | None = None,
         dtype: str | np.dtype | ExtensionDtype | None = _CONSTRUCTOR_DEFAULTS["dtype"],
         copy: bool = _CONSTRUCTOR_DEFAULTS["copy"],
         name: object = _CONSTRUCTOR_DEFAULTS["name"],
         tupleize_cols: bool = _CONSTRUCTOR_DEFAULTS["tupleize_cols"],
+        query_compiler: SnowflakeQueryCompiler = None,
     ) -> Index:
         """
         Override __new__ method to control new instance creation of Index.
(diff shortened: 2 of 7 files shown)


src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -97,12 +95,15 @@ def __init__(
...
         """
-        self._parent = data if isinstance(data, BasePandasDataset) else None
-        data = data._query_compiler if isinstance(data, BasePandasDataset) else data
-        if isinstance(data, SnowflakeQueryCompiler):
-            qc = data
+        if query_compiler:
+            self._query_compiler = query_compiler
+        elif isinstance(data, BasePandasDataset):
+            self._parent = data
+            self._query_compiler = data._query_compiler.drop(
```
sfc-gh-mvashishtha, line 102: I know this behavior existed before this PR, but it doesn't match pandas:

```python
df = pd.DataFrame([1, 2])
# in pandas, get ValueError: Index data must be 1-dimensional
# in Snowpark pandas, get Index([0, 1], dtype='int64')
print(pd.Index([1, 2]))
```

Is this behavior intentional? Do we document it?

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@
 
 ### Behavior change
 - `Dataframe.columns` now returns native pandas Index object instead of Snowpark Index object.
+- Refactor and introduce `query_compiler` argument in `Index` constructor to create `Index` from query compiler.
```
sfc-gh-mvashishtha, line 78: Users shouldn't be aware of the `query_compiler` argument to the `Index` constructor. Should we be noting this change at all in the release notes?

sfc-gh-azhan: why? the constructor is a public API though.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-azhan: I cannot follow the above examples they should both work.

sfc-gh-mvashishtha: Sorry, I edited my example. I meant constructing an index from the dataframe.

CHANGELOG.md
sfc-gh-mvashishtha: It's a public API, but we also don't want or expect users to construct indexes out of query compilers at all.

I would consider this change a refactor that's not worth documenting, but I'm okay with leaving the note here.

sfc-gh-azhan: since it's a public API, it's safer to document any changes related to it. I agree that we won't find any user to be affected by this.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-azhan: oh, that should be a bug. Good catch! Let me see if I can fix it here.

tests/integ/modin/index/test_index_methods.py
```diff
@@ -359,6 +359,9 @@ def test_index_parent():
     assert_frame_equal(snow_idx._parent, df)
     assert_index_equal(snow_idx, native_idx1)
 
+    with pytest.raises(ValueError):
+        pd.Index(df)
```
sfc-gh-mvashishtha, line 363: nit: This assertion seems unrelated to this test case, "Check whether the parent field in Index is updated properly." Can we move it to a separate test case?

sfc-gh-mvashishtha (approved):

sfc-gh-nkrishna (approved):
LGTM!

PR #2047, opened 2024-08-08
SNOW-1568022: [Local Testing] Add support for within_group and array_agg

sfc-gh-jrose:

   

   Fixes SNOW-1568022

   This PR implements the array_agg function and adds support for WithinGroup clauses.


src/snowflake/snowpark/mock/_functions.py
```diff
@@ -390,6 +390,15 @@ def mock_covar_pop(column1: ColumnEmulator, column2: ColumnEmulator) -> ColumnEm
     )
 
 
+@patch("array_agg")
+def mock_array_agg(column: ColumnEmulator, is_distinct: bool) -> ColumnEmulator:
+    columns_data = ColumnEmulator(column.unique()) if is_distinct else column
+    return ColumnEmulator(
+        data=[list(columns_data.dropna())],
+        sf_type=ColumnType(ArrayType(StringType()), False),
```
sfc-gh-aling, line 398: dumb question, why it's StringType not the data type of the input column?

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -259,6 +260,13 @@ def __init__(
         self.options = options
 
 
+def coerce_order_by_arguments(order_by: List[Union[Expression]]):
```
sfc-gh-aling, line 263: nit: for type hint, is it missing other types for Union?

tests/integ/scala/test_dataframe_aggregate_suite.py
```diff
@@ -1336,7 +1332,7 @@ def test_listagg_within_group(session):
     )
     Utils.check_answer(
         df.group_by("color")
-        .agg(listagg("length", ",").within_group(df.length.asc()))
```
sfc-gh-aling, line 1339: just curious why we did this change, and are they equivalent?

src/snowflake/snowpark/mock/_functions.py
sfc-gh-jrose: I can actually probably remove that. The old default was for it to be string type, but I think with some of my more recent changes it should be able to adapt without specifying. That way when structured type support is done rolling out it won't need to be updated her.

tests/integ/scala/test_dataframe_aggregate_suite.py
sfc-gh-jrose: The default sorting order is ascending so I wanted to make sure that was the case with this change. These two statements are equivalent in live mode.

sfc-gh-aling (approved):

PR #2048, opened 2024-08-08
Unify function handling; introduce support for table functions and fix existing issues

sfc-gh-oplaton:

   

   Fixes [SNOW-1509198](https://snowflakecomputing.atlassian.net/browse/SNOW-1509198)

Split `build_fn_apply` into a lower half, `build_fn_apply_args`, an upper half,
`build_builtin_fn_apply`, and implement additional helpers:

* `build_udf_apply` to support `call_udf`.
* `build_session_table_fn_apply` to support.
  `Session.table_function("literal_fn_name", <args...>)` calls.
* `build_table_fn_apply` to support all types of table function APIs:
    * Built-in functions such as `explode`, `explode_outer` and `flatten`.
    * `functions.call_table_function` which returns `TableFunctionCall`.
    * `Session.table_function` calls that take a `Callable` or
      `TableFunctionCall` as input.
    * `functions.table_function("fn_name")` which returns a `Callable`.

Capture the AST for 

* `functions.call_table_function`
* `functions.call_udf`, which used to be treated incorrectly as a built-in
  function
* `functions.explode_outer`
* `functions.explode`
* `functions.flatten`
* `functions.table_function`

Fix the type annotation for `Session.table_function`. It can take the
`Callable` returned from `functions.table_function` as `func_name`.

Implement `session.Session.table_function` correctly for all cases it's meant
to handle:

* The function is given by its literal name. This is a straightforward
  `ApplyExpr` of type `SessionTableFn`
* The function is given by a `Callable` or `TableFunctionCall` object. This is
  an `ApplyExpr` of type `TableFn -> CallType.SessionTableFn` where the
  function object is the first (and only) argument.

The change to `session.Session.table_function` removes duplicate code in
argument handling. The `build_fn_apply` helper already had the functionality
required by `Session.table_function` in a more general form.

Added test coverage to `Session.table_function.test` to take advantage of the
improved functionality in `session.Session.table_function`.

Fixed the test case for `call_udf`. UDFs round trip correctly now.

Handle literal function names uniformly in `ast_utils._set_fn_name`.

### Miscellaneous

The built-in table functions `explode`, `explode_outer` and `flatten` appear
incompatible with the local test mode. Local test coverage is disabled for
these functions in other parts of Snowpark. Did not have time to fully flesh
out a better approach to local testing for these functions. We might want to
ask for help on the broader team.

Disabled the `Session.create_dataframe_from_pandas` test because it was failing
in my environment.

I was hoping to complete the implementation of `TableFunctionCall` to fully
support `over()` and `alias`. Unfortunately, more challenges await in this
area, and I will not be able to complete the work before I leave on break.

Didn't make the corresponding changes to `session.Session.call` (stored
procedures). For symmetry, we will need to add a `build_stored_procedure_apply`
helper, and remove the argument handling code from `Session._call`, leaving
`build_fn_apply_args` as the source of truth for representing function
application arguments.

One more fix to time zone handling, to make this part more robust.

See also https://github.com/snowflakedb/snowflake/pull/201720

[SNOW-1509198]: https://snowflakecomputing.atlassian.net/browse/SNOW-1509198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -183,29 +183,101 @@ def build_proto_from_struct_type(
         ast_field.nullable = field.nullable
```
sfc-gh-lspiegelberg, line 184: this may be orthogonal to this PR potentially, but I am curious: How do you think about supporting decorators?

I.e.
```
f = udf(lambda x: x + 1, return_type=IntegerType())

@udf(return_type=StringType())
def foo(s):
  return f'Hello {s}'
```

Could we model those with the current changes as well, or how would additional work look like?

This does not need to be part of this PR, I am just curious whether the changes here could help support this use case as well or what we'd need to adapt.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -528,3 +600,14 @@ def snowpark_expression_to_ast(expr: Expression) -> proto.Expr:
         raise NotImplementedError(
             f"Snowpark expr {expr} of type {type(expr)} is an expression with missing AST or for which an AST can not be auto-generated."
         )
+
+
```
sfc-gh-lspiegelberg, line 604: nit: maybe move above first usage?

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -528,3 +600,14 @@ def snowpark_expression_to_ast(expr: Expression) -> proto.Expr:
...
+def _set_fn_name(name: Union[str, Iterable[str]], fn: proto.FnRefExpr) -> None:
+    if isinstance(name, str):
```
sfc-gh-lspiegelberg, line 606: nit: docstring

src/snowflake/snowpark/functions.py
```diff
@@ -867,7 +869,7 @@ def create_map(
     col = object_construct_keep_null(*cols)
 
     # Alias to create_map
```
sfc-gh-lspiegelberg, line 871: nit: should we add a helper for that, that uses internally _set_fn_name?

src/snowflake/snowpark/functions.py
```diff
@@ -1322,8 +1324,15 @@ def explode(col: ColumnOrName) -> "snowflake.snowpark.table_function.TableFuncti
         <BLANKLINE>
     """
     col = _to_col_if_str(col, "explode")
+    # AST
```
sfc-gh-lspiegelberg, line 1327: nit: AST.

src/snowflake/snowpark/functions.py
```diff
@@ -8234,7 +8278,16 @@ def table_function(function_name: str) -> Callable:
         >>> session.table_function(split_to_table(lit("split words to table"), lit(" ")).over()).collect()
         [Row(SEQ=1, INDEX=1, VALUE='split'), Row(SEQ=1, INDEX=2, VALUE='words'), Row(SEQ=1, INDEX=3, VALUE='to'), Row(SEQ=1, INDEX=4, VALUE='table')]
     """
-    return lambda *args, **kwargs: call_table_function(function_name, *args, **kwargs)
+    fn = lambda *args, **kwargs: call_table_function(  # noqa: E731
```
sfc-gh-lspiegelberg, line 8281: there is no comment in the original code,  but maybe we could add one to explain why we need to wrap yet in another lambda?

sfc-gh-lspiegelberg: (alternative was functools.partial here)

src/snowflake/snowpark/table_function.py
```diff
@@ -78,6 +78,7 @@ def over(
         Note that if this function is called but both ``partition_by`` and ``order_by`` are ``None``, the table function call will put all input rows into a single partition.
         If this function isn't called at all, the Snowflake database will use implicit partitioning.
         """
+        # TODO(oplaton): Extend over() to collect the AST.
```
sfc-gh-lspiegelberg, line 81: JIRA ticket?

sfc-gh-lspiegelberg: this may be similar to window functions, I think I implemented there some logic to help the over() case.

sfc-gh-lspiegelberg (approved):
LGTM, left a few nits. Nothing major standing out. Great work!

src/snowflake/snowpark/functions.py
sfc-gh-oplaton: I'm not really sure about the design of `functions.table_function`. It just seems like a way of making a Python callable for a function name, e.g.

```python
foo = table_function('foo')
foo(lit(1), lit('bar'))
```

The public API has other inconsistencies and confusing choices, as I mentioned in Slack, too. For example, the signature of `Session.table_function` is `func_name, *args, **kwargs`, but `func_name` can also be this callable, or a `TableFunctionCall`. `TableFunctionCall` in particular takes parameters.

```python
make_table = call_table_function('make_table', lit(42))
```

The meaning of these two is unambiguous:

```python
Session.table_function(make_table)
Session.table_function('make_table', lit(42))
```

What does this do?

```python
Session.table_function(make_table, lit(99))
```

src/snowflake/snowpark/table_function.py
sfc-gh-oplaton: Thanks for the pointer! Will examine the window `over()` implementation. Will file JIRA ticket out of band.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: This would be a good discussion topic. I haven't put any thought into it yet.

If I had to do this from scratch, my steps would be:

* Find out how both examples capture the functions for Snowflake execution. I assume some degree of Python reflection is somehow involved.
* In both examples, actually serializing the function is somewhat orthogonal (or abstracted away) from getting the AST.
* In both examples, the callable object is a Python object. The `@udf` decorator gets the function as a parameter, just like the `udf` call. Assuming a black-box abstraction that can serialize these callable objects, what is non-trivial about doing the work in a decorator?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/functions.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/functions.py
sfc-gh-oplaton: Done.

PR #2063, opened 2024-08-09
SNOW-1546396: Add table creation source for large query breakdown

sfc-gh-aalam:

   

   Fixes SNOW-1546396

   Introduce new class `TableCreationSource` to identify the source of table creation and allow column attributes to be empty.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -96,6 +96,7 @@ from snowflake.snowpark._internal.analyzer.snowflake_plan_node import (
     LogicalPlan,
     SaveMode,
     SnowflakeCreateTable,
+    TableCreationSource,
 )
 from snowflake.snowpark._internal.analyzer.unary_plan_node import (
     CreateDynamicTableCommand,
@@ -840,10 +841,21 @@ class SnowflakePlanBuilder:
         child: SnowflakePlan,
         source_plan: Optional[LogicalPlan],
         use_scoped_temp_objects: bool,
-        is_generated: bool,  # true if the table is generated internally
-        child_attributes: List[Attribute],
+        creation_source: TableCreationSource,
+        # child attributes will be none in the case of large query breakdown
+        # where we use ctas query to create the table which does not need to
+        # know the column metadata
+        child_attributes: Optional[List[Attribute]],
     ) -> SnowflakePlan:
+        is_generated = creation_source in (
+            TableCreationSource.CACHE_RESULT,
+            TableCreationSource.LARGE_QUERY_BREAKDOWN,
+        )
         if is_generated and mode != SaveMode.ERROR_IF_EXISTS:
+            # an internally generated save_as_table comes from two sources:
+            #   1. df.cache_result: plan is created with create table + insert statement
+            #   2. large query breakdown: plan is created using a CTAS statement
+            # For these cases, we must use mode ERROR_IF_EXISTS
             raise ValueError(
                 "Internally generated tables must be called with mode ERROR_IF_EXISTS"
             )
@@ -855,7 +867,7 @@ class SnowflakePlanBuilder:
         # in save as table. So we rename ${number} with COL{number}.
         hidden_column_pattern = r"\"\$(\d+)\""
         column_definition_with_hidden_columns = attribute_to_schema_string(
-            child_attributes
+            child_attributes or []
         )
         column_definition = re.sub(
             hidden_column_pattern,
@@ -976,8 +988,25 @@ class SnowflakePlanBuilder:
                 propagate_referenced_ctes=False,
             )
         elif mode == SaveMode.ERROR_IF_EXISTS:
-            if is_generated:
+            if creation_source == TableCreationSource.CACHE_RESULT:
                 return get_create_and_insert_plan(child, replace=False, error=True)
+
+            if creation_source == TableCreationSource.LARGE_QUERY_BREAKDOWN:
+                return self.build(
+                    lambda x: create_table_as_select_statement(
+                        full_table_name,
+                        x,
+                        column_definition,
+                        replace=False,
+                        table_type=table_type,
+                        clustering_key=clustering_keys,
+                        comment=comment,
+                    ),
+                    child,
+                    source_plan,
+                    is_ddl_on_temp_object=is_temp_table_type,
+                    propagate_referenced_ctes=False,
+                )
 
             return self.build(
                 lambda x: create_table_as_select_statement(
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
@@ -159,6 +159,12 @@ class SaveMode(Enum):
     TRUNCATE = "truncate"
 
 
+class TableCreationSource(Enum):
+    EXPLICIT_USER_COMMAND = "explicit_user_command"
+    CACHE_RESULT = "cache_result"
+    LARGE_QUERY_BREAKDOWN = "large_query_breakdown"
+
+
 class SnowflakeCreateTable(LogicalPlan):
     def __init__(
         self,
@@ -169,7 +175,7 @@ class SnowflakeCreateTable(LogicalPlan):
         table_type: str = "",
         clustering_exprs: Optional[Iterable[Expression]] = None,
         comment: Optional[str] = None,
-        is_generated: bool = False,
+        creation_source: TableCreationSource = TableCreationSource.EXPLICIT_USER_COMMAND,
     ) -> None:
         super().__init__()
 
@@ -184,7 +190,7 @@ class SnowflakeCreateTable(LogicalPlan):
         self.children.append(query)
         self.clustering_exprs = clustering_exprs or []
         self.comment = comment
-        self.is_generated = is_generated
+        self.creation_source = creation_source
 
     @property
     def individual_node_complexity(self) -> Dict[PlanNodeCategory, int]:
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer.py
@@ -991,7 +991,7 @@ class Analyzer:
                 resolved_child,
                 logical_plan,
                 self.session._use_scoped_temp_objects,
-                logical_plan.is_generated,
+                logical_plan.creation_source,
                 resolved_child.attributes,
             )
 
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -829,6 +829,11 @@ def create_table_as_select_statement(
     clustering_key: Optional[Iterable[str]] = None,
     comment: Optional[str] = None,
 ) -> str:
+    column_definition_sql = (
+        f"{LEFT_PARENTHESIS}{column_definition}{RIGHT_PARENTHESIS}"
+        if column_definition
+        else EMPTY_STRING
+    )
     cluster_by_clause = (
         (CLUSTER_BY + LEFT_PARENTHESIS + COMMA.join(clustering_key) + RIGHT_PARENTHESIS)
         if clustering_key
@@ -837,8 +842,8 @@ def create_table_as_select_statement(
     comment_sql = get_comment_sql(comment)
     return (
         f"{CREATE}{OR + REPLACE if replace else EMPTY_STRING} {table_type.upper()} {TABLE}"
-        f"{IF + NOT + EXISTS if not replace and not error else EMPTY_STRING}"
-        f" {table_name}{LEFT_PARENTHESIS}{column_definition}{RIGHT_PARENTHESIS}"
+        f"{IF + NOT + EXISTS if not replace and not error else EMPTY_STRING} "
+        f"{table_name}{column_definition_sql}"
         f"{cluster_by_clause} {comment_sql} {AS}{project_statement([], child)}"
     )
 
--- a/tests/integ/scala/test_snowflake_plan_suite.py
+++ b/tests/integ/scala/test_snowflake_plan_suite.py
@@ -15,7 +15,10 @@ from snowflake.snowpark._internal.analyzer.snowflake_plan import (
     Query,
     SnowflakePlan,
 )
-from snowflake.snowpark._internal.analyzer.snowflake_plan_node import SaveMode
+from snowflake.snowpark._internal.analyzer.snowflake_plan_node import (
+    SaveMode,
+    TableCreationSource,
+)
 from snowflake.snowpark._internal.utils import TempObjectType
 from snowflake.snowpark.functions import col, lit, table_function
 from snowflake.snowpark.session import Session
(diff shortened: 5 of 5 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -976,9 +988,26 @@ def get_create_and_insert_plan(child: SnowflakePlan, replace=False, error=True):
                 propagate_referenced_ctes=False,
             )
         elif mode == SaveMode.ERROR_IF_EXISTS:
-            if is_generated:
+            if creation_source == TableCreationSource.CACHE_RESULT:
                 return get_create_and_insert_plan(child, replace=False, error=True)
 
+            if creation_source == TableCreationSource.LARGE_QUERY_BREAKDOWN:
+                return self.build(
+                    lambda x: create_table_as_select_statement(
```
sfc-gh-jrose, line 996: Could this be done with partial instead?

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: yeah, we can. It will de-duplicate some lines in this function. Let me update that.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -169,7 +175,7 @@ def __init__(
         table_type: str = "",
         clustering_exprs: Optional[Iterable[Expression]] = None,
         comment: Optional[str] = None,
-        is_generated: bool = False,
+        creation_source: TableCreationSource = TableCreationSource.EXPLICIT_USER_COMMAND,
```
sfc-gh-jdu, line 178: I prefer not having a default value so developer will be asked to set it explicitly

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -855,7 +867,7 @@ def save_as_table(
         # in save as table. So we rename ${number} with COL{number}.
         hidden_column_pattern = r"\"\$(\d+)\""
         column_definition_with_hidden_columns = attribute_to_schema_string(
-            child_attributes
+            child_attributes or []
```
sfc-gh-yzou, line 870: i think I would prefer the column definition to be None also, which indicate clear that we do not have column definition

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -159,17 +159,23 @@ class SaveMode(Enum):
     TRUNCATE = "truncate"
 
 
+class TableCreationSource(Enum):
+    EXPLICIT_USER_COMMAND = "explicit_user_command"
```
sfc-gh-yzou, line 163: that doesn't really mean EXPLICIT_USER_COMMAND, right? sometimes, we may have our internal API call that may does this also, for example, to_snowpark_pandas. Maybe w can just say Others, and add a comment here that it will be coming from other save_as_table call

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -159,17 +159,23 @@ class SaveMode(Enum):
...
+    CACHE_RESULT = "cache_result"
```
sfc-gh-yzou, line 164: add comment here about what does those source mean, for example, CACHE_RESULT indicates table created by the df.cache_result API ect

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -840,10 +841,21 @@ def save_as_table(
         child: SnowflakePlan,
         source_plan: Optional[LogicalPlan],
         use_scoped_temp_objects: bool,
-        is_generated: bool,  # true if the table is generated internally
-        child_attributes: List[Attribute],
+        creation_source: TableCreationSource,
```
sfc-gh-yzou, line 844: add a comment for the creation_source, and why we need that

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: in fact, if the column definition is [], it should be an error state for general case, right?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -928,71 +959,22 @@ def get_create_and_insert_plan(child: SnowflakePlan, replace=False, error=True):
...
-                    clustering_key=clustering_keys,
-                    comment=comment,
-                ),
-                child,
-                source_plan,
-                is_ddl_on_temp_object=is_temp_table_type,
-                propagate_referenced_ctes=False,
-            )
+            if creation_source == TableCreationSource.LARGE_QUERY_BREAKDOWN:
+                return get_create_table_as_select_plan(child, replace=False, error=True)
```
sfc-gh-yzou, line 975: it seems we are using get_create_table_as_select_plan(child, replace=False, error=True) for both TableCreationSource.LARGE_QUERY_BREAKDOWN and the rest, is that right? if that is the case, it seems we don't really need the TableCreationSource here, is that right?

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -151,7 +151,7 @@ def do_resolve_with_resolved_children(
                 copied_resolved_child,
                 logical_plan,
                 self.session._use_scoped_temp_objects,
-                logical_plan.is_generated,
+                logical_plan.creation_source,
```
sfc-gh-yzou, line 154: don't you also need to update the child_attributes to pass this as None if not exist?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: actually, nvm. I think we still need the TableCreationSource, but you don't need that extra if here

sfc-gh-aalam: yeah. they are the same. I'll remove one

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-aalam: I was planning to make it part of large query breakdown. Would you say I add it in this one

sfc-gh-aalam: added in this PR

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: added

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-aalam: added

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -840,10 +841,35 @@ def save_as_table(
...
+            table_name: fully qualified table name
+            column_names: names of columns for the table
+            mode: APPEND, TRUNCATE, OVERWRITE, IGNORE, ERROR_IF_EXISTS
+            table_type: temporary, transient, or permanent
+            clustering_keys: list of clustering columns
+            comment: comment associated with the table
+            child: the SnowflakePlan that is being materialized into a table
+            source_plan: the source plan of the child
+            use_scoped_temp_objects: should we use scoped temp objects
+            creation_source: the source of the api that is creating the table
```
sfc-gh-yzou, line 859: large query break down is not an api, I think we can update the comment to be more clear and with more detail like
"""
the creator for the SnowflakeCreateTable node, today it can come from the cache result api, compilation transformations like large query breakdown, or other save_as_table call.
This parameter is used to identifier whether a table is  internally generated with cache_result or by the complilation tranformation, and some special check and handling needs to be applied to guarantee the correctness of generated query.
"""

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -840,10 +841,35 @@ def save_as_table(
...
+        """
+        is_generated = creation_source in (
+            TableCreationSource.CACHE_RESULT,
+            TableCreationSource.LARGE_QUERY_BREAKDOWN,
+        )
         if is_generated and mode != SaveMode.ERROR_IF_EXISTS:
+            # an internally generated save_as_table comes from two sources:
+            #   1. df.cache_result: plan is created with create table + insert statement
+            #   2. large query breakdown: plan is created using a CTAS statement
+            # For these cases, we must use mode ERROR_IF_EXISTS
```
sfc-gh-yzou, line 872: let's also add a check that if child_attribute is None, it can only be for the case when creation source is TableCreationSource.LARGE_QUERY_BREAKDOWN

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -121,16 +121,15 @@ def do_resolve_with_resolved_children(
...
-            # associated with the current query generator.
-            # TODO: this check need to be relaxed when large query breakdown with temp table
-            #       is implemented, because the child attributes are not necessary to create
-            #       the temp table
-            assert self._snowflake_create_table_plan_info is not None
-            assert (
-                self._snowflake_create_table_plan_info.table_name
+            child_attributes = None
+            if (
+                self._snowflake_create_table_plan_info
```
sfc-gh-yzou, line 126: let's be more strict about this part of code, since we have the logical_plan.creation_source here, let's do a check that if the creation_source is large query breakdown, set the child attribute to None, else, we can keep the original assertion and get the child attribute from the map.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -928,71 +978,19 @@ def get_create_and_insert_plan(child: SnowflakePlan, replace=False, error=True):
...
-                child,
-                source_plan,
-                is_ddl_on_temp_object=is_temp_table_type,
-                propagate_referenced_ctes=False,
-            )
+            return get_create_table_as_select_plan(child, replace=False, error=False)
+
         elif mode == SaveMode.ERROR_IF_EXISTS:
-            if is_generated:
+            if creation_source == TableCreationSource.CACHE_RESULT:
```
sfc-gh-yzou, line 990: let's add a comment here about why we use the create and insert plan for cache result, which for handling cache_result during transaction, for the other internally generated table by large query break down, since the optimization will be blocked for query with active transaction, we can safely create it with select statement

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -159,17 +159,30 @@ class SaveMode(Enum):
     TRUNCATE = "truncate"
 
 
+class TableCreationSource(Enum):
+    """The enum to indicate the source of a create table request.
+
+    CACHE_RESULT: when create table request comes from DataFrame.cache_result
+    LARGE_QUERY_BREAKDOWN: when request comes from large query breakdown optimization
```
sfc-gh-yzou, line 166: "request comes from " -> table is created by

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -121,16 +121,15 @@ def do_resolve_with_resolved_children(
...
+                and self._snowflake_create_table_plan_info.table_name
```
sfc-gh-yzou, line 127: also since we do that now, let's also update the create_query_generator function here to be more clear, we only need to extract the attribute when the creator_source is not large query breakdown. Even though we don't expect we will have actual usage of creating query generator when SnowflakeCreateTable node is coming from large query breakdown, this is more for code completness.

tests/integ/compiler/test_query_generator.py
```diff
@@ -145,6 +146,7 @@ def test_table_create(session, mode):
         column_names=None,
         mode=mode,
         query=df._plan,
+        creation_source=TableCreationSource.OTHERS,
```
sfc-gh-yzou, line 149: can we add a test that test with the TableCreationSource is LargeQueryBreakdown, you might have to a check of the result query with actual query text here

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-yzou: @sfc-gh-aalam Can we keep the original assertion check when the create source is not large query breakdown?

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-yzou: @sfc-gh-aalam how about this one? if we do not do this one, the create_query_compiler will try to extract the child_atrribute out of the child plan

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -840,32 +841,92 @@ def save_as_table(
...
+            mode: APPEND, TRUNCATE, OVERWRITE, IGNORE, ERROR_IF_EXISTS
+            table_type: temporary, transient, or permanent
+            clustering_keys: list of clustering columns
+            comment: comment associated with the table
+            child: the SnowflakePlan that is being materialized into a table
+            source_plan: the source plan of the child
+            use_scoped_temp_objects: should we use scoped temp objects
+            creation_source: the creator for the SnowflakeCreateTable node, today it can come from the
+                cache result api, compilation transformations like large query breakdown, or other like
+                save_as_table call. This parameter is used to identifier whether a table is internally
```
sfc-gh-yzou, line 861: identifier -> identify

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-yzou: sorry, what i mean is we can do 
```
if (logical_plan.creation_source != TableCreationSource.LARGE_QUERY_BREAKDOWN):
     assert self._snowflake_create_table_plan_info is not None
     assert (self._snowflake_create_table_plan_info.table_name==logical_plan.table_name)
     child_attributes = self._snowflake_create_table_plan_info.child_attributes
```

so that it explicitly errors out at the line where it needs the child attribute to be more clear. otherwise, i believe it will error out inside the save_as_table function with the check today

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -121,16 +122,23 @@ def do_resolve_with_resolved_children(
...
+            if (
+                logical_plan.creation_source
+                != TableCreationSource.LARGE_QUERY_BREAKDOWN
+                and self._snowflake_create_table_plan_info
+                and self._snowflake_create_table_plan_info.table_name
                 == logical_plan.table_name
-            )
+            ):
+                child_attributes = (
+                    self._snowflake_create_table_plan_info.child_attributes
```
sfc-gh-yzou, line 140: @sfc-gh-aalam how about the other comment for handling the creator in the create_query_generator here to only do the resolve if the creator is not large_query_breakdown https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/compiler/utils.py#L59

sfc-gh-yzou (approved):

PR #2064, opened 2024-08-09
SNOW-1546396: Fix update_resolvable_node to reset attributes for SelectStatement & SetStatement

sfc-gh-aalam:

   

   Fixes SNOW-1546396

### DIFF
--- a/src/snowflake/snowpark/_internal/compiler/utils.py
+++ b/src/snowflake/snowpark/_internal/compiler/utils.py
@@ -180,7 +180,11 @@ def update_resolvable_node(
     query_generator: QueryGenerator,
 ):
     """
-    Helper function to re-resolve the resolvable node and do an in-place update for cached fields.
+    Helper function to make an in-place update for a node that has had its child updated.
+    It works in two parts:
+      1. Re-resolve the proper fields based on the child.
+      2. Resets re-calculable fields such as _sql_query, _snowflake_plan, _cumulative_node_complexity.
+
     The re-resolve is only needed for SnowflakePlan node and Selectable node, because only those nodes
     are resolved node with sql query state.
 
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -193,6 +193,9 @@ def update_resolvable_node(
                           |
                         JOIN
     resolve_node(SelectSnowflakePlan, query_generator) will resolve both SelectSnowflakePlan and SnowflakePlan nodes.
+
+    This operation also resets the previously calculated cumulative_node_complexity for the node so a
```
sfc-gh-jdu, line 197: If we add "reset" operation in this function, this function name is not informational enough - so how about call it `update_resolvable_node_and_reset_complexity`?

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: i think we also reset other recalculable fields like _sql_query, _snowflake_plan, so this name may not be accurate also. maybe we can update the comment to mention that the update includes two parts:
1) Re-Resolve the proper fields based on the child
2) Clear up re-calculatable fields like _sql_query, _snowflake_plan, _cumulative_node_complexity ect

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -207,20 +210,48 @@ def update_resolvable_node(
             update_resolvable_node(node.source_plan, query_generator)
         resolve_and_update_snowflake_plan(node, query_generator)
 
-    elif isinstance(node, (SelectStatement, SetStatement)):
+    elif isinstance(node, SelectStatement):
+        # clean up the cached sql query and snowflake plan to allow
```
sfc-gh-yzou, line 214: do we need to reset that forSnowflakePlan node also?

sfc-gh-yzou: i think we can simply do it before if isinstance(node, SnowflakePlan) for all type of nodes, right?

sfc-gh-aalam: Snowflake plan actually copies it over from the new resolved snowflake plan which already sets it to None.

sfc-gh-yzou (approved):

PR #2068, opened 2024-08-12
Set the new DSL-level version number attribute for AST requests

sfc-gh-azwiegincew:
This avoids the needlessly complicated process of dealing with the semantic client version number on the server.

Also, set the proper client version number AND the full Python version.


src/snowflake/snowpark/_internal/ast.py
```diff
@@ -144,8 +144,9 @@ def flush(self) -> Tuple[str, str]:
     def _init_batch(self):
         self._request_id = uuid.uuid4()  # Generate a new unique ID.
         self._request = proto.Request()
-        self._request.client_version.major = 42
```
sfc-gh-lspiegelberg, line 147: should we remove this? what is the client_version here, the Snowpark client version (i.e., 1.20?)?

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -144,8 +144,9 @@ def flush(self) -> Tuple[str, str]:
...
+        self._request.client_version.major = 42  # TODO
         self._request.client_version.minor = 0
         (major, minor, micro, releaselevel, serial) = sys.version_info
         self._request.client_language.python_language.version.major = major
         self._request.client_language.python_language.version.minor = minor
```
sfc-gh-lspiegelberg, line 151: nit: would prefer to have the MAX_VERSION be publicly exposed with an interface/property.

sfc-gh-lspiegelberg (approved):
not sure about client_version, if possible - can we fix it in this PR as well? Else, lgtm.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-azwiegincew: I added a TODO here.

PR #2075, opened 2024-08-13
SNOW-1559329: Create a client-side Timedelta type.

sfc-gh-mvashishtha:
Fixes SNOW-1559329.

Support:
- create a dataframe column or a series out of data containing `pd.Timedelta` and/or `datetime.timedelta` objects
- convert the resulting dataframe or series to pandas


sfc-gh-nkrishna (commented):
Overall, looks good to me. Left some questions/minor comments. Nice work Mahesh!

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @staticmethod
+    @abstractmethod
+    def to_pandas(value: Any) -> Any:
+        """
+        Convert an object representing this type in Snowpark Python to the pandas representation.
+        """
+
+
+class TimedeltaType(SnowparkPandasType):
```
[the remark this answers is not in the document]
sfc-gh-mvashishtha: That's right.

src/snowflake/snowpark/modin/plugin/_internal/type_utils.py
```diff
@@ -250,6 +245,10 @@ def to_snowflake(
         if is_float_dtype(p):
             return DoubleType()
 
+        snowpark_pandas_type = get_snowpark_pandas_type_for_pandas_type(p)
+        if snowpark_pandas_type is not None:
+            return snowpark_pandas_type()
```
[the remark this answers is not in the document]
sfc-gh-mvashishtha: Yep, that's the goal.

sfc-gh-nkrishna (approved):

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1052,11 +1062,33 @@ def create_ordered_dataframe_from_pandas(
     # np.nan, pd.NA, pd.NaT and None all as missing value.
     isna_data = df.isna().to_numpy().tolist()
 
+    snowpark_pandas_types = []
```
sfc-gh-azhan, line 1065: I would suggest refactoring it in a different way. The name of this method `create_ordered_dataframe_from_pandas` is about to create the ODF only and there is nothing to do with `SnowparkPandasType`. Instead, we could have another helper method called `infer_snowpark_pandas_type(df) -> tuple[snowflake_type, struct_fields]`. Then we can use this struct_fields as an input to `create_ordered_dataframe_from_pandas`.

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -0,0 +1,139 @@
...
+        if isinstance(value, native_pd.Timedelta):
+            return value.value
+        return native_pd.Timedelta(value).value
+
+
+def rewrite_python_object_if_snowpark_pandas_type(value: Any) -> Any:
+    """
+    If a python object is an instance of a Snowpark pandas type, rewrite it into its Snowpark Python representation.
+    """
+    for cls, from_pandas in _python_type_to_from_pandas.items():
```
sfc-gh-azhan, line 127: Is it possible to get the class of the `value` first and just check whether it is in the dict?

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1071,16 +1103,19 @@ def create_ordered_dataframe_from_pandas(
                 else (
                     data[x][y].item()
                     if isinstance(data[x][y], np.generic)
-                    else data[x][y]
+                    else rewrite_python_object_if_snowpark_pandas_type(data[x][y])
```
sfc-gh-azhan, line 1106: how about call it `ensure_snowpark_python_type`?

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -672,6 +715,90 @@ def validate_no_duplicated_data_columns_mapped_for_labels(
...
+
+        Returns:
+            A list of Snowpark types for this frame's index columns.
+        """
+        return [
+            self.snowflake_quoted_identifier_to_data_type[v.snowflake_quoted_identifier]
+            for v in self.label_to_snowflake_quoted_identifier[: self.num_index_columns]
+        ]
+
+    def to_pandas(
```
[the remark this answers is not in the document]
sfc-gh-azhan: me too

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13241,6 +13222,10 @@ def create_lazy_type_functions(
             label_to_snowflake_quoted_identifier=label_to_snowflake_quoted_identifier,
             num_index_columns=new_frame.num_index_columns,
             data_column_index_names=new_frame.data_column_index_names,
+            snowflake_quoted_identifier_to_data_type={
```
sfc-gh-azhan, line 13225: why we need this change in this PR?

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -93,6 +96,11 @@ class InternalFrame:
     # The length of data_column_index_names equals to number of multiindex levels.
     # For a 3-level MultiIndex, the value can be like ['A', 'B', 'C']
     data_column_index_names: tuple[LabelTuple, ...]
+    # Map from snowflake identifier to cached Snowpark pandas data type.
+    # The type is None if we don't know the Snowpark data type.
+    # n.b. that we map to SnowparkPandasType rather than to DataType, because
+    # we don't want to try tracking regular Snowpark Python types at all.
+    snowflake_quoted_identifier_to_data_type: dict[str, Optional[SnowparkPandasType]]
```
sfc-gh-azhan, line 103:
```suggestion
    snowflake_quoted_identifier_to_type: dict[str, Optional[SnowparkPandasType]]
```

sfc-gh-azhan: we can be consistent that:
`type` is short for snowpark pandas type;
`pandas type` is the native pandas type;
`snowpark type` is the snowpark python type.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -163,6 +174,29 @@ def create(
                 from_pandas_label(name, num_levels=1)
                 for name in data_column_pandas_index_names
             ),
+            snowflake_quoted_identifier_to_data_type={
```
sfc-gh-azhan, line 177: let's add an assertion helper method for checking type inputs, e.g.,

- check the input type length
- check the values either is None or SnowparkPandasType

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -163,6 +174,29 @@ def create(
...
+                # Combine the data_column_types and index_column_types with
+                # their associated identifiers.
+                k: v
```
sfc-gh-azhan, line 180: nit: move them into a helper method called `create_snowflake_quoted_identifier_to_data_type`

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -102,8 +110,10 @@ def create(
         data_column_pandas_labels: list[Hashable],
         data_column_pandas_index_names: list[Hashable],
         data_column_snowflake_quoted_identifiers: list[str],
+        data_column_types: Optional[list[Optional[SnowparkPandasType]]] = None,
```
sfc-gh-evandenberg, line 113: Should default arguments be at the end since the index_column_... following arguments require values?

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -271,10 +305,18 @@ def data_column_snowflake_quoted_identifiers(self) -> list[str]:
         ]
 
     def quoted_identifier_to_snowflake_type(self) -> dict[str, DataType]:
```
sfc-gh-azhan, line 307: we probably should rename it to `quoted_identifier_to_type` since it may contain snowpark pandas type now.

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -0,0 +1,139 @@
...
+from typing import Any, Callable, Optional, Union
+
+import numpy as np
+import pandas as native_pd
+
+from snowflake.snowpark.modin.plugin.utils.warning_message import WarningMessage
+from snowflake.snowpark.types import DataType, LongType
+
+TIMEDELTA_WARNING_MESSAGE = (
+    "Snowpark pandas support for Timedelta is still under development."
```
sfc-gh-evandenberg, line 17: Is this something we've already promised to customers?  Otherwise would leave more generic.
```suggestion
    "Snowpark pandas support for Timedelta is not currently available."
```

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -0,0 +1,139 @@
...
+        Convert the Snowpark Python representation of Timedelta to native_pd.Timedelta.
+        """
+        return native_pd.Timedelta(value, unit="nanosecond")
+
+    @staticmethod
+    def from_pandas(value: Union[native_pd.Timedelta, datetime.timedelta]) -> int:
+        """
+        Convert a pandas representation of a Timedelta to its nanoseconds.
+        """
+        # `Timeedelta.value` converts Timedelta to nanoseconds.
```
sfc-gh-evandenberg, line 117:
```suggestion
        # `Timedelta.value` converts Timedelta to nanoseconds.
```

sfc-gh-evandenberg (approved):
Minor comments, LGTM

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -672,6 +715,90 @@ def validate_no_duplicated_data_columns_mapped_for_labels(
                 message += f" {user_frame_identifier}"
             raise ValueError(message)
 
+    def cached_data_column_snowpark_pandas_types(
```
sfc-gh-azhan, line 718: can be a cached property

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -672,6 +715,90 @@ def validate_no_duplicated_data_columns_mapped_for_labels(
...
+
+        Returns:
+            A list of Snowpark types for this frame's data columns.
+        """
+        return [
+            self.snowflake_quoted_identifier_to_data_type[v.snowflake_quoted_identifier]
+            for v in self.label_to_snowflake_quoted_identifier[self.num_index_columns :]
+        ]
+
+    def cached_index_column_snowpark_pandas_types(
```
sfc-gh-azhan, line 738: can be a cached property

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1226,6 +1261,7 @@ def snowpark_to_pandas_helper(
     ordered_dataframe: OrderedDataFrame,
     *,
     statement_params: Optional[dict[str, str]] = None,
+    type_overrides: Optional[list[DataType]] = None,
```
sfc-gh-azhan, line 1264: just call it `cached_snowpark_pandas_types` and this should be required right?

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -0,0 +1,139 @@
...
+    We represent Timedelta as the integer number of nanoseconds between the
+    two times.
+    """
+
+    snowpark_type = LongType()
+    pandas_type = np.dtype("timedelta64[ns]")
+    types_to_convert_with_from_pandas = [native_pd.Timedelta, datetime.timedelta]
+
+    def __init__(self) -> None:
+        WarningMessage.single_warning(TIMEDELTA_WARNING_MESSAGE)
```
sfc-gh-azhan, line 102: add a todo to remind us to remove this warning before GA. (We can use the doc jira number for this)

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+    If a python object is an instance of a Snowpark pandas type, rewrite it into its Snowpark Python representation.
+    """
+    for cls, from_pandas in _python_type_to_from_pandas.items():
+        if isinstance(value, cls):
+            return from_pandas(value)
+    return value
+
+
+def get_snowpark_pandas_type_for_pandas_type(
```
sfc-gh-azhan, line 133: This sounds like better to be a `SnowparkPandasType` class method, e.g., `SnowparkPandasType.from_pandas_type()`

sfc-gh-azhan (changes requested):
Looks great and let's resolve all comments before merge.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-mvashishtha: Good idea! Constructing each `StructField` requires the Snowflake quoted identifier, so `infer_snowpark_pandas_type` should return `DataType` but `StructField` objects. Done.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
```diff
@@ -93,6 +143,11 @@ class InternalFrame:
     # The length of data_column_index_names equals to number of multiindex levels.
     # For a 3-level MultiIndex, the value can be like ['A', 'B', 'C']
     data_column_index_names: tuple[LabelTuple, ...]
+    # Map from snowflake identifier to cached Snowpark pandas data type.
+    # The type is None if we don't know the Snowpark data type.
+    # n.b. that we map to SnowparkPandasType rather than to DataType, because
+    # we don't want to try tracking regular Snowpark Python types at all.
+    snowflake_quoted_identifier_to_type: dict[str, Optional[SnowparkPandasType]]
```
sfc-gh-nkumar, line 150: InternalFrame is an Immutable class. We should not use mutable dict here, that will break the guarantee that we have right now.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
sfc-gh-nkumar: Unfortunately we can not cache a mutable object. This will break the immutability guarantee of InternalFrame.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
sfc-gh-nkumar: Same as above

sfc-gh-nkumar (changes requested):

src/snowflake/snowpark/modin/plugin/_internal/frame.py
sfc-gh-azhan: https://stackoverflow.com/questions/11014262/how-to-create-an-immutable-dictionary-in-python

src/snowflake/snowpark/modin/plugin/_internal/frame.py
sfc-gh-azhan: InternalFrame is immutable so we can use cached property right?

sfc-gh-nkumar: We can only use it if return type is immutable. For example if return type is int/tuple we can cache it but for list/dict we can not cache it. 
Consider following scenario:
```
x = frame.cached_data_column_snowpark_pandas_types
x["abc"] = "xyz"
print(frame.cached_data_column_snowpark_pandas_types) # this will have abc -> xyz in output
```

sfc-gh-azhan: Oh, good lesson! How about return tuple instead of list.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
sfc-gh-mvashishtha: Switched to immutable `MappingProxyType `.

src/snowflake/snowpark/modin/plugin/_internal/frame.py
sfc-gh-mvashishtha: We could return tuple and have `cached_property`, but `InternalFrame.create()` takes a `list`, so I'd have to either change that argument to `Sequence`, or make  the places that construct `data_column_types` out of an existing `cached_data_column_snowpark_pandas_types` switch the type from tuple to list. The performance gains from `cached_property` are minimal so I think `property` is ok.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -517,9 +516,18 @@ def from_pandas(
                 ordered_dataframe=ordered_dataframe,
                 data_column_pandas_labels=original_column_labels,
                 data_column_pandas_index_names=original_column_index_names,
+                data_column_types=snowpark_types[
+                    len(index_snowflake_quoted_identifiers) : (
```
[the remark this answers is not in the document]
sfc-gh-azhan: We could change `infer_snowpark_types_from_pandas` to return `snowpark_pandas_data_column_types`, `snowpark_pandas_index_column_types`, and `snowpark_types` and directly use them here.

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -0,0 +1,141 @@
...
+                new_snowpark_python_type.pandas_type, existing_cls
+            ), f"Already registered Snowpark pandas type for class {cls} with {existing_cls}"
+        _pandas_type_to_snowpark_pandas_type[
+            new_snowpark_python_type.pandas_type
+        ] = new_snowpark_python_type
+
+        return new_snowpark_python_type
+
+
+class SnowparkPandasType(DataType, metaclass=SnowparkPandasTypeMetaclass):
```
sfc-gh-nkumar, line 71: Sorry I missed this earlier. This class also needs to be frozen dataclass (for immutability) as this is being used inside InternalFrame. Please take a look at OrderingColumn class for example.

sfc-gh-nkumar (approved):
We also have to convert SnowparkPandasType into immutable class. (PTAL at OrderingColumn for example). Otherwise looks good! Thanks

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
```diff
@@ -1034,9 +1034,9 @@ def generate_column_agg_info(
             The new index data column index names for the dataframe after aggregation
     """
 
-    quoted_identifier_to_snowflake_type: dict[
+    quoted_identifier_to_type: dict[
```
sfc-gh-azhan, line 1037: as we discussed offline, let's remove this renaming out of this PR. We can revisit this naming decision in future PRs.

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
sfc-gh-mvashishtha: Fixed.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: That doesn't make sense to me because `infer_snowpark_types_from_pandas` shouldn't have the context of how the pandas dataframe it gets was constructed (specifically, that it contains the original index columns, followed by data columns, then a row position column), whereas `from_pandas` should have that context. I prefer the current version.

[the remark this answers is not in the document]
sfc-gh-azhan: I see. I was confused because I didn't read the code above which preprocessed `df` to be the original df plus the index.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 #### New Features
 - Added support for `snowflake.snowpark.testing.assert_dataframe_equal` that is a util function to check the equality of two Snowpark DataFrames.
+- Added limited support for the `Timedelta` type.
```
sfc-gh-azhan, line 9:
```suggestion
- Added limited support for the `Timedelta` type, including creating `Timedelta` columns and `to_pandas`.
```

sfc-gh-azhan (approved):
Great start!

PR #2083, opened 2024-08-13
[SNOW-1491307] Session.write_pandas support for Snowpark IR

sfc-gh-lspiegelberg:
Support `session.write_pandas` for pandas Dataframes. Snowpark pandas Dataframes are out of scope for phase 0 and subsequently error. Fix local testing rest mock object by returning mock request with error message whenever called that will get propagated to user. This helps to more easily trace missing features within Mock/local testing feature.


tests/ast/data/session_write_pandas.test
```diff
@@ -0,0 +1,15 @@
...
+
+df = pd.DataFrame([[1, 2, 3], [4, 5, 6]])
+
+ans = session.write_pandas(df, "test_table")
+
+ans2 = session.write_pandas(df, "test", schema="a", database="b", chunk_size=7, compression="brotli", on_error="ignore", parallel=10, quote_identifiers=False, auto_create_table=True, create_temp_table=True, overwrite=True, table_type="temp", random=90)
+
+## EXPECTED OUTPUT
+
+res1 = session.write_pandas(pandas.DataFrame(<not shown>)), "test_table")
```
sfc-gh-azwiegincew, line 13: Strange – why is the symbol not getting captured here?

sfc-gh-azwiegincew (approved):

tests/ast/data/session_write_pandas.test
sfc-gh-lspiegelberg: Good catch! Looked into it, in the `with_src_position` statement I forgot the `stmt` as parameter. Fixed now, i.e.

```
with_src_position(stmt.expr.sp_write_pandas, stmt)
```

PR #1737, opened 2024-06-05
SNOW-929218 Add type coercion support for CaseWhen and Iff

sfc-gh-stan:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-929218

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

PR #2087, opened 2024-08-13
SNOW-1510073: [Local Testing] Add type coercion to udf expressions

sfc-gh-jrose:

   

   Fixes SNOW-1510073

  This PR replaces the udf logic to use type coercion logic introduced in #1737. I took a look at sprocs as well, but I believe there logic is already correct without a similar change. 

  I've also skipped some tests that rely on running modulo against a coerced integer column. This is due to a bug in coerced integers that prevents the modulo operator from working correctly. I'll open a separate PR to address that. See ticket SNOW-1625599 for more info.


tests/integ/test_udf.py
```diff
@@ -673,8 +674,8 @@ def plus4_then_mod5_with_import(x):
     session.add_import(stage_file)
     plus4_then_mod5_udf = udf(
         plus4_then_mod5_with_import,
-        return_type=IntegerType(),
-        input_types=[IntegerType()],
+        return_type=LongType(),
+        input_types=[LongType()],
```
sfc-gh-aalam, line 678: Is this expected? would this be considered bcr?

sfc-gh-aling: +1, I think we should avoid changing existing live test code.
if needed I think we can use the local testing mode fixture.

sfc-gh-jrose: The df that is generated below this definition will have an inferred type of LongType. These tests have been relying on the server side coercing the type to IntegerType. This breaks in local testing due to the bug I've described in SNOW-1625599. I could try to fix that bug before merging this change.

sfc-gh-aalam (approved):

sfc-gh-aling (approved):

PR #2095, opened 2024-08-15
SNOW-1618349: Add support for `pd.merge_asof`

sfc-gh-nkrishna:
Fixes SNOW-1618349

This PR adds initial support for `pd.merge_asof` -- parameters `by`, `left_by`, `right_by`, `left_index`, `right_index`, `suffixes`, and `tolerance` are not yet supported. Additionally `direction=nearest` is not yet supported but this can be done in a follow-up PR.


src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -394,26 +394,152 @@ def merge_ordered(
...
+    left_by: str | None = None,
+    right_by: str | None = None,
+    suffixes: Suffixes = ("_x", "_y"),
+    tolerance: int | Timedelta | None = None,
     allow_exact_matches: bool = True,
     direction: str = "backward",
 ) -> DataFrame:  # noqa: PR01, RT01, D200
     """
     Perform a merge by key distance.
+
```
sfc-gh-helmeleegy, line 416: Any idea why the docstrings for methods in `general.py` are not maintained under the `../docstrings/` directory?

sfc-gh-nkrishna: There's only about 20 functions, so since it's a small enough set having the docstrings for each one doesn't bloat the file too much. But for uniformity, I can refactor this in a follow-up PR. I'll file a JIRA for it.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,192 @@ def merge(
...
+        if (
+            by
+            or left_by
+            or right_by
+            or left_index
+            or right_index
+            or tolerance
+            or suffixes != ("_x", "_y")
+        ):
+            ErrorMessage.not_implemented(
```
sfc-gh-helmeleegy, line 6820: Can you add negative tests for those NotImplementedErrors?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,192 @@ def merge(
...
+                "Snowpark pandas merge_asof API does not currently support parameters "
```
sfc-gh-helmeleegy, line 6821: Nit:
```suggestion
                "Snowpark pandas merge_asof method does not currently support parameters "
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,192 @@ def merge(
...
+                + "'by', 'left_by', 'right_by, 'left_index', 'right_index', "
+                + "'suffixes', or 'tolerance'"
+            )
+        if direction not in ("backward", "forward"):
+            ErrorMessage.not_implemented(
+                "Snowpark pandas merge_asof API only supports directions 'forward' and 'backward'"
```
sfc-gh-helmeleegy, line 6827: Nit:
```suggestion
                "Snowpark pandas merge_asof method only supports directions 'forward' and 'backward'"
```

sfc-gh-helmeleegy (commented):
Looks good overall. Just a few nits. Thanks for the quick turnaround.

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-helmeleegy: I think it's more about minimizing the conflicts with Modin OSS. Thanks!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Done

sfc-gh-helmeleegy (approved):

src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -394,26 +394,158 @@ def merge_ordered(
...
+    >>> pd.merge_asof(trades, quotes, on="time")
+                         time   price  quantity     bid     ask
+    0 2016-05-25 13:30:00.023   51.95        75   51.95   51.96
+    1 2016-05-25 13:30:00.038   51.95       155   51.97   51.98
+    2 2016-05-25 13:30:00.048  720.77       100  720.50  720.93
+    3 2016-05-25 13:30:00.048  720.92       100  720.50  720.93
+    4 2016-05-25 13:30:00.048   98.00       100  720.50  720.93
     """
     # TODO: SNOW-1063345: Modin upgrade - modin.pandas functions in general.py
     if not isinstance(left, DataFrame):
```
sfc-gh-rdurrani, line 551: I think that this check may be wrong? We're checking that the left isn't of type `DataFrame`,  and then erroring out on the type of the right? We should probably be checking the type of the right as well?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,198 @@ def merge(
...
+            If True, allow matching with the same ‘on’ value (i.e. less-than-or-equal-to / greater-than-or-equal-to)
+            If False, don’t match the same ‘on’ value (i.e., strictly less-than / strictly greater-than).
+        direction : ‘backward’ (default), ‘forward’, or ‘nearest’
+            Whether to search for prior, subsequent, or closest matches.
+
+        Returns
+        -------
+        SnowflakeQueryCompiler
+        """
+        if (
```
sfc-gh-rdurrani, line 6811: Is `by` not supported by Snowpark? ASOF joins allow a by value, so could we support `by` in this PR?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,198 @@ def merge(
...
+        right_frame = right._modin_frame
+        # Project to Snowpark DataFrame in order to perform ASOF Join
+        left_snowpark_df = (
+            left_frame.ordered_dataframe.to_projected_snowpark_dataframe()
+        )
+        right_snowpark_df = (
+            right_frame.ordered_dataframe.to_projected_snowpark_dataframe()
+        )
+        if on:
+            on_quoted_identifier = quote_name_without_upper_casing(on)
```
sfc-gh-rdurrani, line 6840: This will break if the quoted identifier is different from the pandas name, right? (e.g. say we have a series of operations, and the quoted identifier becomes `"pandas_name_xai23"`, which happens sometimes when deduplicating columns) - could we instead use the mapping that the internal frame has from pandas label to quoted identifer to figure out the quoted identifier?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,198 @@ def merge(
...
+            lhs_expr = left_snowpark_df[on_quoted_identifier]
+            rhs_expr = right_snowpark_df[on_quoted_identifier]
+            order_by_expr = [
+                left_snowpark_df[on_quoted_identifier],
+                right_snowpark_df[on_quoted_identifier],
+            ]
+        else:
+            assert left_on and right_on
+            left_on_quoted_identifier = quote_name_without_upper_casing(left_on)
```
sfc-gh-rdurrani, line 6849: ditto

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,198 @@ def merge(
...
+        else:
+            # Output pandas labels consists of all data pandas labels in both frames
+            assert left_on and right_on
+            data_column_pandas_labels = (
+                left_frame.data_column_pandas_labels
+                + right_frame.data_column_pandas_labels
+            )
+
+        left_frame_num_valid_cols = len(left_frame.data_column_pandas_labels)
+        right_frame_num_valid_cols = len(right_frame.data_column_pandas_labels)
```
sfc-gh-rdurrani, line 6892: Can we move this into the join_utils file?

tests/integ/modin/test_merge_asof.py
```diff
@@ -0,0 +1,257 @@
...
+            "right_val": [1, 2, 3, 6, 7],
+            "another_col_right": [0, np.nan, 3, 6, 8],
+            "a": [1, 2, 3, 6, 7],
+        },
+    )
+    return df1, df2
+
+
+@pytest.fixture(scope="function")
+def left_right_native_df_left_right_by():
```
sfc-gh-rdurrani, line 53:
```suggestion
def left_right_native_df_left_right_on():
```

tests/integ/modin/test_merge_asof.py
```diff
@@ -0,0 +1,257 @@
...
+        direction=direction,
+    )
+    snow_output = pd.merge_asof(
+        left_snow_df,
+        right_snow_df,
+        on="a",
+        allow_exact_matches=allow_exact_matches,
+        direction=direction,
+    )
+    eval_snowpark_pandas_result(snow_output, native_output, lambda df: df)
```
sfc-gh-rdurrani, line 131: maybe use `assert_snowpark_pandas_equals...`?

tests/integ/modin/test_merge_asof.py
```diff
@@ -0,0 +1,257 @@
...
+        pd.merge_asof(left_snow_df, right_snow_df, on="time", direction="nearest")
+
+
+@sql_count_checker(query_count=0)
+def test_merge_asof_params_unsupported(left_right_timestamp_data):
+    left_native_df, right_native_df = left_right_timestamp_data
+    left_snow_df, right_snow_df = pd.DataFrame(left_native_df), pd.DataFrame(
+        right_native_df
+    )
+    with pytest.raises(
```
sfc-gh-rdurrani, line 220: shouldn't we have a separate pytest.raises for each one, as otherwise the first one will trip the raises?

tests/integ/modin/test_merge_asof.py
```diff
@@ -0,0 +1,257 @@
...
+                pd.Timestamp("2016-05-25 13:30:00.048"),
+            ],
+            "price": [51.95, 51.95, 720.77, 720.92, 98.0],
+            "quantity": [75, 155, 100, 100, 100],
+        }
+    )
+    return quotes, trades
+
+
+allow_exact_matches = pytest.mark.parametrize("allow_exact_matches", [True, False])
```
sfc-gh-rdurrani, line 105: nice touch!

sfc-gh-rdurrani (changes requested):
left some comments + requests

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: I don't see a way to input the `by` into the Snowpark join function. In future, we should really refactor `join_utils.join` to support `asof` join and ideally have a way to directly pass in `by`.

tests/integ/modin/test_merge_asof.py
sfc-gh-nkrishna: Good point, done.

tests/integ/modin/test_merge_asof.py
sfc-gh-nkrishna: Good catch, done.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Great point, done!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Done.

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-nkrishna: No I think this is correct. We are checking that the left value is a DataFrame, if it isn't we are saying that we can't merge it with the right value's type.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Which lines do you think should be moved into `join_utils`? I think most of this code is specific to asof join so isn't general purpose enough to warrant a function in `join_utils`, but I can make it one if you think it's better.

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-rdurrani: but shouldn't we be saying that since it isn't a DataFrame, we can't merge type(left)? (Purely semantic, but this is how the other error messages usually are, but maybe this is a weird pandas thing)?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: Had a feeling that was the case - please ignore this comment in that case!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: Ah I was thinking more perhaps adding the whole do the join + map to new labels to a function (either the general join function, or a specific asof join function) in join_utils.py so that way you don't have to do the mapping here. (Also potentially check if theres code in join_utils.py that you can reuse to get the mapping). Basically, have some function that returns a JoinOrAlignResult (as would happen if you did a traditional join).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6837,25 +6837,35 @@ def merge_asof(
...
         )
         if on:
-            on_quoted_identifier = quote_name_without_upper_casing(on)
-            lhs_expr = left_snowpark_df[on_quoted_identifier]
-            rhs_expr = right_snowpark_df[on_quoted_identifier]
-            order_by_expr = [
-                left_snowpark_df[on_quoted_identifier],
-                right_snowpark_df[on_quoted_identifier],
-            ]
+            left_on_quoted_identifier = (
```
sfc-gh-rdurrani, line 6840: What happens if multiple columns share the same label in vanilla pandas? I think grabbing the first column is good for us, but I'd be curious to see how pandas handles that.

sfc-gh-rdurrani (commented):
LGTM - left a couple quick q's tho!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Actually I made sure that we handle that case same as pandas, I can add a test for it

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-nkrishna: Fair point, changed it to `left` and added a similar check for `right` like you suggested.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: native pandas returns `ValueError: The column label 'A' is not unique.`

sfc-gh-nkrishna: Currently in the code, we handle the duplicates though. 

```
    A left_val   A  right_val   A
0   1        a  10          1  10
1   5        b  12          3  13
2  10        c  13          7  15
```

I think this behavior is okay without any additional warnings, but I could also add checks for duplicates. What do you think?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,208 @@ def merge(
...
+        if direction == "backward":
+            match_condition = (
+                (lhs_expr >= rhs_expr) if allow_exact_matches else (lhs_expr > rhs_expr)
+            )
+        else:
+            match_condition = (
+                (lhs_expr <= rhs_expr) if allow_exact_matches else (lhs_expr < rhs_expr)
+            )
+
+        joined_snowpark_df = left_snowpark_df.join(
```
sfc-gh-yzou, line 6878: snowpark pandas is not suppose to be directly operate on snowpark dataframe, all operation has to go through ordered dataframe, please make sure we extend our join_utils, and join operator at ordered dataframe properly
cc @sfc-gh-nkumar please also help review this pr to help make sure we extend the utility properly

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6744,6 +6745,208 @@ def merge(
...
+        left_index: bool = False,
+        right_index: bool = False,
+        by: Optional[Union[str, list[str]]] = None,
+        left_by: Optional[str] = None,
+        right_by: Optional[str] = None,
+        suffixes: Suffixes = ("_x", "_y"),
+        tolerance: Optional[Union[int, Timedelta]] = None,
+        allow_exact_matches: bool = True,
+        direction: str = "backward",
+    ) -> "SnowflakeQueryCompiler":
```
sfc-gh-yzou, line 6763: @sfc-gh-nkumar can you help review here to see if the merge implementation can be resued here for merge_asof

sfc-gh-yzou (changes requested):

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1176,15 +1188,29 @@ def join(
...
-        # If we are doing a cross join, `on` cannot be specified.
-        if how != "cross":
+        if how == "asof":
+            left_match_condition = Column(left_match_condition)
+            # Get the new mapped right match condition identifier
+            right_match_condition = Column(
+                right_identifiers_rename_map[right_match_condition]  # type: ignore
+            )
+            # ASOF Join requires the use of match_condition
             snowpark_dataframe = left_snowpark_dataframe_ref.snowpark_dataframe.join(
```
sfc-gh-rdurrani, line 1198: Shouldn't `by` be specified here as the `on` parameter?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6940,6 +6941,133 @@ def merge(
...
+            If False, don’t match the same ‘on’ value (i.e., strictly less-than / strictly greater-than).
+        direction : ‘backward’ (default), ‘forward’, or ‘nearest’
+            Whether to search for prior, subsequent, or closest matches.
+
+        Returns
+        -------
+        SnowflakeQueryCompiler
+        """
+        if (
+            by
```
sfc-gh-rdurrani, line 7008: Shouldn't we support `by` now?

sfc-gh-rdurrani (changes requested):
I think that we should add support for `by`, since it won't be a whole lot of extra code.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-nkrishna: I tried doing that, but there is some extra work required for `by/on` with ASOF Join. Specifically, I ended up getting this:

```
SnowparkSQLException: (1304): 01b67bf9-090b-0f68-000c-a905793aac47: 010010 (42601): SQL compilation error:
ON clause for ASOF JOIN must contain conjunctions of equality conditions only. Disjunctions are not allowed. Each side of an equality condition must only refer to either the left table or the right table. EQUAL_NULL(SNOWPARK_LEFT."ticker", SNOWPARK_RIGHT."ticker_ksr3") is invalid.
```

sfc-gh-rdurrani: Oh interesting - mind flagging that with the snowpark team? If that's the case, I agree that we don't need to support by here!

sfc-gh-rdurrani (approved):
LGTM!

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1052,6 +1052,9 @@ def join(
         right: "OrderedDataFrame",
         left_on_cols: Optional[list[str]] = None,
         right_on_cols: Optional[list[str]] = None,
+        left_match_condition: Optional[str] = None,
```
sfc-gh-yzou, line 1055: this seems for column identifier, let's be consistent on the name, and call it left_match_col, and right_match_col

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1052,6 +1052,9 @@ def join(
...
+        right_match_condition: Optional[str] = None,
+        comparator_match_condition: Optional[str] = None,
```
sfc-gh-yzou, line 1057: this seems just for match_comparator, let's rename it to be more accurate.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1068,17 +1071,25 @@ def join(
             right: The other OrderedDataFrame to join.
             left_on_cols: A list of column names from self OrderedDataFrame to be used for the join.
             right_on_cols: A list of column names from right OrderedDataFrame to be used for the join.
+            left_match_condition: Snowflake identifier to match condition on from 'left' frame.
+                Only applicable for 'asof' join.
+            right_match_condition: Snowflake identifier to match condition on from 'right' frame.
+                Only applicable for 'asof' join.
+            comparator_match_condition: {"__ge__", "__gt__", "__le__", "__lt__"}
```
sfc-gh-yzou, line 1078: instead of string, can be provide an enum for the operators that are supported

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1119,6 +1130,7 @@ def join(
             "right",
             "inner",
             "outer",
+            "asof",
```
sfc-gh-yzou, line 1133: It seems that when asof is used, the left_match_col, right_match_col and match comparator is required, i think we can add an assertion check here if how is asof, that  left_match_col, right_match_col and match comparator is not provided

sfc-gh-yzou: also, this condition here checks if the join can be skipped here. i don't think we can skip the asof join if the left on and right on is for row position column, because it depends on the match condition. however if match condition is on the row_position_column, and __ge__ and __lt__ is used as comparator, we could propbably skip the join here

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1176,15 +1188,29 @@ def join(
             eq = Column(left_col).equal_null(Column(right_col))
             on = eq if on is None else on & eq
 
-        # If we are doing a cross join, `on` cannot be specified.
-        if how != "cross":
+        if how == "asof":
+            left_match_condition = Column(left_match_condition)
+            # Get the new mapped right match condition identifier
+            right_match_condition = Column(
+                right_identifiers_rename_map[right_match_condition]  # type: ignore
```
sfc-gh-yzou, line 1195: how come we need a type ignore here?

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1119,6 +1130,7 @@ def join(
             "right",
```
sfc-gh-yzou, line 1130: also expand the check here https://github.com/snowflakedb/snowpark-python/pull/2095/files#diff-b123de92fc315a358428fa2a33390e78a0b8ee641420d7f0f67fe65dc58315daL1101, that the newly added left_macth_col and right_match_col also have to be part of the projected column

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-yzou: so there is only one match column is allowed for Pandas? what is the error we will see if more than on match column is provided?

src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -394,29 +394,163 @@ def merge_ordered(
...
-    tolerance=None,
+    by: str | list[str] | None = None,
+    left_by: str | None = None,
+    right_by: str | None = None,
+    suffixes: Suffixes = ("_x", "_y"),
+    tolerance: int | Timedelta | None = None,
     allow_exact_matches: bool = True,
     direction: str = "backward",
-) -> DataFrame:  # noqa: PR01, RT01, D200
+):  # noqa: PR01, RT01, D200
```
sfc-gh-yzou, line 413: why the return type is removed? and can the #noqa be removed here?

src/snowflake/snowpark/modin/pandas/general.py
```diff
@@ -394,29 +394,163 @@ def merge_ordered(
...
     """
     Perform a merge by key distance.
+
+    This is similar to a left-join except that we match on nearest key rather than equal keys.
+    Both DataFrames must be sorted by the key. For each row in the left DataFrame:
```
sfc-gh-yzou, line 418: the documentation should for pd modular should be provided in the same way as how other api is provided, right? otherwise, once @sfc-gh-joshi removes those file, the doc will be removed also. 
@sfc-gh-joshi can you help here to see how to configure the docstring for general.py?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6940,6 +6941,133 @@ def merge(
...
+            or left_by
+            or right_by
+            or left_index
+            or right_index
+            or tolerance
+            or suffixes != ("_x", "_y")
+        ):
+            ErrorMessage.not_implemented(
+                "Snowpark pandas merge_asof method does not currently support parameters "
+                + "'by', 'left_by', 'right_by', 'left_index', 'right_index', "
```
sfc-gh-yzou, line 7018: do we have a ticket to track the support for those, it seems the support of left_by, right_by, left_index, right_index, suffix is exactly the same as merge and join, most of the code can probably be resued

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6940,6 +6941,133 @@ def merge(
...
+        joined_frame, _ = join_utils.join(
+            left=left_frame,
+            right=right_frame,
+            how="asof",
+            left_on=[left_match_quoted_identifier],
+            right_on=[right_match_quoted_identifier],
+            left_match_condition=left_match_quoted_identifier,
+            right_match_condition=right_match_quoted_identifier,
+            comparator_match_condition=comparator_match_condition,
+            join_key_coalesce_config=[JoinKeyCoalesceConfig.LEFT] if on else None,
```
sfc-gh-yzou, line 7066: this seems behaves the same as merge, if the left_on column name and right_on column is the same, the result column will have only one column, otherwise it will be the same column. For example, if left_on='a', right_on='a', the result will only have one a, other wise it will have two columns.

sfc-gh-yzou: you should be able to reuse the following function to get the join keys
```
left_keys, right_keys = join_utils.get_join_keys(
                left._modin_frame,
                right._modin_frame,
                on,
                left_on,
                right_on,
                left_index,
                right_index,
            )
```
and the same logic for decising the colease config used in merge can probably also be reused
```
 coalesce_config = []
        for lkey, rkey in zip(left_keys, right_keys):
            if lkey == rkey or rkey in external_join_keys:
                coalesce_config.append(join_utils.JoinKeyCoalesceConfig.LEFT)
            elif lkey in external_join_keys:
                coalesce_config.append(join_utils.JoinKeyCoalesceConfig.RIGHT)
            else:
                coalesce_config.append(join_utils.JoinKeyCoalesceConfig.NONE)
```
we can probably add another utilty function in join_utils to get the coalesce_config

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6940,6 +6941,133 @@ def merge(
...
+                + "'suffixes', or 'tolerance'"
+            )
+        if direction not in ("backward", "forward"):
+            ErrorMessage.not_implemented(
+                "Snowpark pandas merge_asof method only supports directions 'forward' and 'backward'"
+            )
+
+        left_frame = self._modin_frame
+        right_frame = right._modin_frame
+        if on:
```
sfc-gh-yzou, line 7028: @sfc-gh-nkumar i believe most of the logic in the merge query_compiler method can be reused here, i am thinking may be we can have an _merge in snowflake_query_compiler that could take all common logic, or provide another util function in join_utils.  wdyt?
Since right now, the overlap is relatively small, i am fine if we want to do the reuse later when supporting other parameters, if we do that, @sfc-gh-nkrishna please make it clear that we want to do some refactor to resue the logic for merge as much as possible.

tests/integ/modin/test_merge_asof.py
```diff
@@ -0,0 +1,278 @@
...
+            "right_val": [1, 2, 3, 6, 7],
+            "another_col_right": [0, np.nan, 3, 6, 8],
+            "b": [1, 2, 3, 6, 7],
+        },
+    )
+    return df1, df2
+
+
+@pytest.fixture(scope="function")
+def left_right_timestamp_data():
```
sfc-gh-yzou, line 72: do we have a test to test the behavior difference when the on column is not numeric of timestamp type?

tests/integ/modin/test_merge_asof.py
```diff
@@ -0,0 +1,278 @@
...
+    quotes = native_pd.DataFrame(
+        {
+            "time": [
+                pd.Timestamp("2016-05-25 13:30:00.023"),
+                pd.Timestamp("2016-05-25 13:30:00.023"),
+                pd.Timestamp("2016-05-25 13:30:00.030"),
```
sfc-gh-yzou, line 78: does date type also work ?

sfc-gh-nkrishna: Yes, added a test

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-nkrishna: Good catch, added some error checking in the frontend with negative tests to match.

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-nkrishna: Good callout -- All the general `pd` APIs are done this way (e.g. `pivot`, `merge`), so I don't think this should block this PR but definitely something worth noting.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Added a TODO with a ticket.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Ditto added a TODO with a ticket.

tests/integ/modin/test_merge_asof.py
sfc-gh-nkrishna: Added a test for this:

pandas raises a ValueError (`Incompatible merge dtype, dtype\(\'O\'\) and dtype\(\'O\'\), both sides must have numeric dtype`)

Our code raises a SnowparkSQLException (`MATCH_CONDITION clause is invalid: The left and right side expressions must be numeric or timestamp expressions.`). 

The errors are quite close so I don't think the dtype check in the frontend is worth it. Thoughts?

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-nkrishna: I added an `assert right_match_col` earlier in the function, but mypy still complains ` Invalid index type "Optional[str]", expected type "str"`

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1052,6 +1053,9 @@ def join(
         right: "OrderedDataFrame",
         left_on_cols: Optional[list[str]] = None,
         right_on_cols: Optional[list[str]] = None,
+        left_match_col: Optional[str] = None,
+        right_match_col: Optional[str] = None,
+        match_comparator: Optional[Enum] = None,
```
sfc-gh-nkrishna, line 1058: Can't type as MatchComparator due to circular import issue

sfc-gh-yzou: you will be able to use MatchComparator with quote, like "MatchComparator" here, right?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: @sfc-gh-nkrishna i am little bit confused here, it seems the join_key_coalesce_config is still None if user is using left_on and right_on, and the result will contain both columns. 
You might be confused about the comment of colease config, it is not a logic inside get_join_keys, the get_join_keys helps figure out what are the correct join keys form left and right given all parameters related to key, but it doesn't figure out the coalesce_config, the coalesce_config logic I pasted above is actually from the merge configuration here https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py#L6876.
I was suggesting that we can extract this as a common util function, and call it for both merge, and this function here

tests/integ/modin/test_merge_asof.py
sfc-gh-yzou: No, let's do not do type check. My main purpose here is to make sure we understand the behavior difference, as far as we have a test to capture this behavior, we are good here.

tests/integ/modin/test_merge_asof.py
```diff
@@ -0,0 +1,388 @@
...
+        right_native_df,
+        left_on="a",
+        right_on="b",
+        allow_exact_matches=allow_exact_matches,
+        direction=direction,
+    )
+    snow_output = pd.merge_asof(
+        left_snow_df,
+        right_snow_df,
+        left_on="a",
```
sfc-gh-yzou, line 204: one test you want to add based on the implementation is probably when left_on = 'a', and right_on = 'a', which should have the same behavior as on = 'a'

src/snowflake/snowpark/modin/pandas/general.py
sfc-gh-yzou: Yeah. It seems we may not have such capability today yet. I am following up on this with @sfc-gh-joshi offline on this.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-yzou: you will need to add it right before this line

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: Ah that makes sense, thanks for the clarification!

sfc-gh-nkrishna:
@sfc-gh-yzou Could you take another pass? Addressed all comments

sfc-gh-yzou (approved):

PR #2096, opened 2024-08-15
[SNOW-1491253] Snowpark IR - server side support for DataFrame.stat functions.

sfc-gh-lspiegelberg:
Adds AST generation for `DataFrame.stat.{approx_quantile,corr,cov,crosstab,sample_by}`.

Other changes:
Existing functions `DataFrame.{col,sample,union_all,pivot}` get `_emit_ast` as new parameter to allow AST generation to be disabled on demand. The Snowpark `Column` class allows to disable AST generation with added `_emit_ast` now too. Adding In local testing mock functions `approx_percentile_accumulate,approx_percentile_estimate,covar_samp,corr_samp` , returning dummy values until properly implemented, to allow the mock server connection to successfully run the stats test.


tests/ast/data/DataFrame.stat.test
```diff
@@ -0,0 +1,91 @@
...
+
+fractions = {"Bob": 0.5, "Nico": 1.0}
+
+sample_df = df.stat.sample_by("name", fractions)
+
+## EXPECTED OUTPUT
+
+df = session.table("test_table")
+
+df2 = df.stat
```
sfc-gh-azwiegincew, line 35: Let's do the `na` approach here and not bind `stat`. This stat object is stateless, like na, unlike write.

sfc-gh-lspiegelberg: I already merged the server-side branch. We certainly can do the `na` approach here, however imho the more correct one should be similar to `write` as you can assign the `DataframeStatFunctions` object technically. If your concern is formatting, I believe your work with the inliner will resolve this. I can change everything similar to `na`, yet this means I need to reopen a server-branch, refactor, modify IR and modify this branch here. I estimate the work to be `~ 3h`. Do you think atm this a time investment worth it? If so sure - but I think we can always refactor this later if need be. Another option is open up a low priority JIRA ticket and do this refactor in another branch pair.

sfc-gh-azwiegincew: This unnecessarily complicates the AST, making our Phase 2 work a lot more complicated than it should be. This is a super simple fix, but feel free to merge as-is.

sfc-gh-azwiegincew (approved):

PR #2097, opened 2024-08-15
SNOW-1625830: Fix circular import when calling `to_snowpark_pandas` without initializing Snowpark pandas

sfc-gh-joshi:

   

   Fixes SNOW-1625830

Calling `to_snowpark_pandas` on a Snowpark Python DataFrame without first performing `import snowflake.snowpark.modin.plugin` currently raises an error (see attached JIRA for exact reproduction). This was not the case in previous releases, and this PR fixes internal imports such that performing this operation implicitly initializes Snowpark pandas.

After this PR, when `to_snowpark_pandas` is called without explicitly initializing Snowpark pandas, one of two things happens:
1. If modin is not installed, Snowpark pandas will surface the following error:
```
ModuleNotFoundError: Modin is not installed. Run `pip install "snowflake-snowpark-python[modin]"` to resolve.
```
This is the same error as if the user had tried to `import snowflake.snowpark.modin.plugin` without installing the modin dependency.

2. If modin is installed, Snowpark pandas will implicitly initialize Snowpark pandas by internally running `import snowflake.snowpark.modin.plugin`. If the user performs `import modin.pandas as pd` afterwards, the modin namespace will be set up with Snowpark pandas behavior.

### DIFF
--- a/tests/integ/test_df_to_snowpark_pandas.py
+++ b/tests/integ/test_df_to_snowpark_pandas.py
@@ -40,13 +40,7 @@ def test_to_snowpark_pandas_no_modin(session, tmp_table_basic):
     # Check if modin is installed (if so, we're running in Snowpark pandas; if not, we're just in Snowpark Python)
     try:
         import modin  # noqa: F401
-
-        modin_installed = True
     except ModuleNotFoundError:
-        modin_installed = False
-    if modin_installed:
-        snowpark_df.to_snowpark_pandas()  # should have no errors
-    else:
         # Current Snowpark Python installs pandas==2.2.2, but Snowpark pandas depends on modin
         # 0.28.1, which needs pandas==2.2.1. The pandas version check is currently performed
         # before Snowpark pandas checks whether modin is installed.
@@ -58,3 +52,5 @@ def test_to_snowpark_pandas_no_modin(session, tmp_table_basic):
             match="does not match the supported pandas version in Snowpark pandas",
         ):
             snowpark_df.to_snowpark_pandas()
+    else:
+        snowpark_df.to_snowpark_pandas()  # should have no errors
--- a/tests/integ/test_df_to_snowpark_pandas.py
+++ b/tests/integ/test_df_to_snowpark_pandas.py
@@ -10,6 +10,14 @@ import pytest
 from snowflake.snowpark._internal.utils import TempObjectType
 from tests.utils import Utils
 
+pytestmark = [
+    pytest.mark.xfail(
+        "config.getoption('local_testing_mode', default=False)",
+        reason="This is testing Snowpark pandas installation",
+        run=False,
+    )
+]
+
 
 @pytest.fixture(scope="module")
 def tmp_table_basic(session):
@@ -29,11 +37,14 @@ def tmp_table_basic(session):
 
 def test_to_snowpark_pandas_no_modin(session, tmp_table_basic):
     snowpark_df = session.table(tmp_table_basic)
-    # Check if modin is installed
+    # Check if modin is installed (if so, we're running in Snowpark pandas; if not, we're just in Snowpark Python)
     try:
         import modin  # noqa: F401
 
         snowpark_df.to_snowpark_pandas()  # should have no errors
     except ModuleNotFoundError:
-        with pytest.raises(ModuleNotFoundError, match="Modin is not installed."):
+        with pytest.raises(
+            ModuleNotFoundError,
+            match=r"(Modin is not installed)|(does not match the supported pandas version in Modin)",
+        ):
             snowpark_df.to_snowpark_pandas()
--- a/tests/integ/test_df_to_snowpark_pandas.py
+++ b/tests/integ/test_df_to_snowpark_pandas.py
@@ -44,7 +44,7 @@ def test_to_snowpark_pandas_no_modin(session, tmp_table_basic):
         snowpark_df.to_snowpark_pandas()  # should have no errors
     except ModuleNotFoundError:
         with pytest.raises(
-            ModuleNotFoundError,
-            match=r"(Modin is not installed)|(does not match the supported pandas version in Modin)",
+            (ModuleNotFoundError, RuntimeError),
+            match=r"(Modin is not installed)|(does not match the supported pandas version in Snowpark pandas)",
         ):
             snowpark_df.to_snowpark_pandas()
--- a/tests/integ/test_df_to_snowpark_pandas.py
+++ b/tests/integ/test_df_to_snowpark_pandas.py
@@ -41,10 +41,20 @@ def test_to_snowpark_pandas_no_modin(session, tmp_table_basic):
     try:
         import modin  # noqa: F401
 
-        snowpark_df.to_snowpark_pandas()  # should have no errors
+        modin_installed = True
     except ModuleNotFoundError:
+        modin_installed = False
+    if modin_installed:
+        snowpark_df.to_snowpark_pandas()  # should have no errors
+    else:
+        # Current Snowpark Python installs pandas==2.2.2, but Snowpark pandas depends on modin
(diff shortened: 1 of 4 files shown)


tests/integ/test_df_to_snowpark_pandas.py
```diff
@@ -0,0 +1,39 @@
...
+
+    try:
+        yield table_name
+    finally:
+        Utils.drop_table(session, table_name)
+
+
+def test_to_snowpark_pandas_no_modin(session, tmp_table_basic):
+    snowpark_df = session.table(tmp_table_basic)
+    # Check if modin is installed
```
sfc-gh-yzou, line 32: i am little bit confused here about the behavior, what happens under the following scenario before and after your change:
1: modin is not installed
2: modin is installed but user didn't do" import modin.pandas as pd; import snowflake.snowpark.modin.plugin"

i assume now under case 2, once we call snowpark_df.to_snowpark_pandas(), user will be able to perform snowpark pandas dataframe operations on top it just fine. what happens if user do the following after that?
```
import modin.pandas as pd
pd.read_snowflake(xx) 
```

sfc-gh-joshi: Yes, in case 2 any operations on `modin.pandas` would be with Snowpark pandas.

sfc-gh-yzou: Can you give me a brief description about the behavior before and after your change for the above 2 scenarios ?

sfc-gh-joshi: **Previous (release 1.18, any commit prior to 40a534e)**
1. modin not installed: Calling `to_snowpark_pandas_dataframe` raises `ModuleNotFoundError: Modin is not installed. Run `pip install "snowflake-snowpark-python[modin]"` to resolve.`
2. modin is installed: After calling `to_snowpark_pandas_dataframe`, the Snowpark pandas plugin is initialized and after running `import modin.pandas as pd`, users can perform Snowpark pandas operations as normal.

**Current (releases 1.19/1.20)**
1. modin not installed: Calling `to_snowpark_pandas_dataframe` raises `ModuleNotFoundError: No module named 'modin'. The error message is different because it bypasses the Snowpark pandas plugin initialization code which would validate the modin version, and instead directly runs code that tries to import Modin.
2. modin is installed: Calling `to_snowpark_pandas_dataframe` raises `ImportError: cannot import name '_create_operator_docstring' from partially initialized module 'snowflake.snowpark.modin.utils' (most likely due to a circular import) (/Users/joshi/code/snowpandas/src/snowflake/snowpark/modin/utils.py)`.


**This PR**
As in PR description, the behavior for both cases matches that in release 1.18.

tests/integ/test_df_to_snowpark_pandas.py
```diff
@@ -0,0 +1,39 @@
...
+    try:
+        import modin  # noqa: F401
```
sfc-gh-yzou, line 34: why do we import modin here ?

tests/integ/test_df_to_snowpark_pandas.py
sfc-gh-yzou: I see, if that is the case, this will run under the env where modin is not installed, right? 
Also, i see we have a test suite called test_df_to_snowpark_pandas, the test didn't import snowpark pandas, how come it didn't error out? did the modular got initiated for all integ test? can we have one test that mimic the user's behavior?

sfc-gh-yzou:
@sfc-gh-joshi it seems your test is failing, can you take a look? can you also start the snowpark python daily test for this change

tests/integ/test_df_to_snowpark_pandas.py
sfc-gh-joshi: It seems like I misunderstood how our test environments are set up; you're right that this test runs in the Snowpark Python test environment that does not have modin.

Like you said, the existing test under `tests/integ/modin/test_df_to_snowpark_pandas.py` run with the plugin properly initialized (this happens in `tests/integ/modin/conftest.py`). It looks like if we're to properly test both scenarios we'll need to add a couple new test environments:
(1: no modin installed) Putting the test outside the Snowpark pandas hierarchy (like the `tests/integ/test_df_to_snowpark_pandas.py` file I added here) more or less gives us the desired setup, but the test is currently failing due to a mismatch in pandas version (this environment is using 2.2.2 while modin needs 2.2.1).
(2: modin is installed) This test would have to run outside the Snowpark pandas hierarchy to avoid initializing the plugin, but also needs to be in an environment that has the modin dependency + correct pandas version installed.

Do you have any suggestions for alternative approaches, or should I proceed with adding these extra test environments? Since they'd only be executing 1 or 2 tests, I think it should be OK to add to precommit, but if we want to avoid bloat I can also put them in the daily job instead.

tests/integ/test_df_to_snowpark_pandas.py
sfc-gh-joshi: I misunderstood how our test environments are set up and I thought the `modin-*` CI jobs would also run Snowpark Python tests, so I added this to check if modin was present in the current environment. It seems like this is currently never run with modin installed, so we'd have to explicitly create separate environments to run this test (see my comment in the other thread).

tests/integ/test_df_to_snowpark_pandas.py
sfc-gh-yzou: I think here is what we can do, we already have test env that have modin installed (snowpark pandas) and env without modin installed (snowpark python), you can right the test to capture behavior on both env, and keep the test under snowpark integration fold, not modin folder. This test will automatically been run under the non-modin env, and then add this test point to snowparkpandasnotdoctest in tox.ini, it will run under the modin env.

For the 2.2.1 and 2.2.2 I felt the modin check should happen first before pandas, because if customer installs modin, it should automatically fix the pandas problem.

sfc-gh-joshi: @sfc-gh-yzou I made the change you suggested (keeping the test in top-level `tests/integ`, and adding this test to tox.ini to run in an environment with modin.

I made the test check for either a `ModuleNotFoundError` (modin not installed) or `RuntimeError` (mismatched pandas version); as in my above comment, we need to check pandas version before modin internals do because modin surfaces a different warning. We can revisit this behavior upstream later.

tests/integ/test_df_to_snowpark_pandas.py
sfc-gh-joshi: I changed the test setup such that this test runs in both Snowpark Python and Snowpark pandas CI. The `import modin` statement is used to check which environment we're in; apparently this try/except is the "idiomatic" way of checking for the presence of a package in Python.

sfc-gh-joshi:
Daily job run in progress: https://github.com/snowflakedb/snowpark-python/pull/2132

tests/integ/test_df_to_snowpark_pandas.py
```diff
@@ -0,0 +1,50 @@
...
+    try:
+        yield table_name
+    finally:
+        Utils.drop_table(session, table_name)
+
+
+def test_to_snowpark_pandas_no_modin(session, tmp_table_basic):
+    snowpark_df = session.table(tmp_table_basic)
+    # Check if modin is installed (if so, we're running in Snowpark pandas; if not, we're just in Snowpark Python)
+    try:
```
sfc-gh-mvashishtha, line 41: can we wrap just the `import modin` in the try/except? and put the `snowpark_df.to_snowpark_pandas()` in `else`?

sfc-gh-joshi: good point, fixed

sfc-gh-mvashishtha (approved):

sfc-gh-yzou (approved):

sfc-gh-helmeleegy (approved):
LGTM, thanks Jonathan!

PR #2103, opened 2024-08-15
SNOW-1628400: Keep nanoseconds in Timestamp scalars.

sfc-gh-mvashishtha:
Fixes SNOW-1628400


sfc-gh-helmeleegy (approved):
LGTM.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1491,7 +1493,10 @@ def convert_numpy_pandas_scalar_to_snowpark_literal(value: Any) -> LiteralType:
     if isinstance(value, np.generic):
         value = value.item()
     elif isinstance(value, native_pd.Timestamp):
-        value = value.to_pydatetime()
+        raise ValueError(
+            "Internal error: cannot represent Timestamp as a Snowpark literal. "
```
sfc-gh-jkew, line 1497: The message is a little confusing here. We cannot represent timestamp so make it a string and then a timestamp?

sfc-gh-jkew (approved):

PR #2104, opened 2024-08-15
[feature branch] Avoid creating an intermediate binding for the stateless object <df>.stat

sfc-gh-azwiegincew:


src/snowflake/snowpark/dataframe_stat_functions.py
```diff
@@ -92,14 +92,8 @@ def approx_quantile(
         if _emit_ast:
             # Add an assign node that applies SpDataframeStatsApproxQuantile() to the input, followed by its Eval.
```
sfc-gh-lspiegelberg, line 93: the constructor here prob. needs to be modified to remove self._ast_stmt (+ its parameter)

sfc-gh-lspiegelberg (approved):
lgtm

PR #2106, opened 2024-08-15
SNOW-1458122 Index name apis

sfc-gh-vbudati:

   

   Fixes SNOW-1458122, SNOW-1021837

   Implemented Index `name`, `names`, `rename`, and `set_names`. 

This behavior is slightly different from native pandas since not all references to an Index or a DataFrame/Series' index are kept track of - this would require a big redesign. Instead, an Index can keep track of its parent DataFrame/Series and modify them when necessary.

`_parent` keeps track of the parent Series/DataFrame and is set whenever `df.index` or `series.index` is called.

Related design doc: https://docs.google.com/document/d/1clzOiC0Wu9KtTf0MqFDWagPesvKF0rKEizfz5IZ43v4/edit

This PR also fixes bug SNOW-1021837 where the index column name was not being correctly updated after `reset_index`.


src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -132,7 +133,7 @@ def __init__(
 
         Parameters
         ----------
-        data : array-like (1-dimensional), pandas.Index, modin.pandas.Series, optional
+        data : array-like (1-dimensional), pandas.Index, modin.pandas.Series, SnowflakeQueryCompiler, optional
```
sfc-gh-azhan, line 136: this is a change made last week, qc will be given throw argument `query_compiler`.
```suggestion
        data : array-like (1-dimensional), pandas.Index, modin.pandas.Series, optional
```

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -172,11 +173,13 @@ def __init__(
 
     def _init_index(
         self,
-        data: ArrayLike | native_pd.Index | Series | None,
+        data: ArrayLike | native_pd.Index | Series | DataFrame | None,
```
sfc-gh-azhan, line 176:
```suggestion
        data: ArrayLike | native_pd.Index | Series | None,
```
Can not create index from dataframe

sfc-gh-azhan (changes requested):

tests/integ/modin/index/test_name.py
```diff
@@ -0,0 +1,291 @@
...
+        new_native_idx.names
+        == new_snow_idx.names
+        == ([None] if new_name is None else ["grade"])
+    )
+    assert native_df.index.name == snow_df.index.name == "score"
+
+
+@pytest.mark.parametrize("inplace", [True, False])
+@sql_count_checker(query_count=0)
+def test_index_rename_list(inplace):
```
sfc-gh-mvashishtha, line 245: This doesn't seem to be working end-to-end with `to_pandas`. Should we fix or raise an error before the user calls `to_pandas`?

```python
i = pd.Index([1, 2, 3])
i.name = ['name1']
i.to_pandas()
```

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1545,8 +1571,29 @@ def rename(self) -> None:
...
+        --------
+        >>> idx = pd.Index(['A', 'C', 'A', 'B'], name='score')
+        >>> idx.rename('grade', inplace=False)
+        Index(['A', 'C', 'A', 'B'], dtype='object', name='grade')
+        >>> idx.rename('grade', inplace=True)
+
+        Note
+        ----
+        Native pandas only allows hashable types for names. Snowpark pandas allows
+        name to be any scalar or list-like type. If a type is used for the name,
```
sfc-gh-mvashishtha, line 1585:
```suggestion
        name to be any scalar or list-like type. If a tuple is used for the name,
```

tests/integ/modin/index/test_name.py
```diff
@@ -0,0 +1,291 @@
...
+    # a MultiIndex.
+    native_idx = native_pd.Index(["A", "C", "A", "B"], name="score")
+    snow_idx = pd.Index(native_idx)
+    with pytest.raises(ValueError, match="Level must be None for non-MultiIndex"):
+        native_idx.set_names("grade", level=level)
+
+    # Verifying the results.
+    native_res = native_idx.set_names("grade")
+    snow_res = snow_idx.set_names("grade", level=level)
+    assert native_res.name == snow_res.name == "grade"
```
sfc-gh-mvashishtha, line 291: Could you please also add a test case for SNOW-1021837 and mention it in the PR description? I think your change will fix that bug.

tests/integ/modin/index/test_name.py
sfc-gh-vbudati: I don't think it's valid to set the name like that. I added a check to the `name` property to not accept non-hashable types as a name. The discrepancy I'm pointing out here is that Index.set_names in pandas accepts a list of names but rename does not (documentation says that it should) - so I made list-like names work with both APIs.
However, the name itself cannot be a list.

sfc-gh-mvashishtha (approved):

sfc-gh-azhan (approved):

PR #2107, opened 2024-08-15
[feature branch] Updated unparser expectations – expression inlining

sfc-gh-azwiegincew:


sfc-gh-lspiegelberg (approved):
lgtm, could we potentially rename `tests/ast/data/unparser.test` to something more informative? The distinct feature of this test seems calling `df` within a function?

PR #2074, opened 2024-08-12
SNOW-1491175 Remove set_src_position, overhaul with_src_position for better performance

sfc-gh-heshah:
[SNOW-1491175](https://snowflakecomputing.atlassian.net/browse/SNOW-1491175)

Remove all uses of `set_src_position` and `get_first_non_snowpark_stack_frame`, update the necessary test cases, and use the `inspect` module with better practices including
- Deleting the retrieved frame to prevent reference cycles and memory leaks
- Avoid capturing code context for every file along the stack
- Skip two frames out of the Snowpark library code to retrieve relevant user code immediately when possible
- Add comments discussing functionality in detail for future improvements

[SNOW-1491175]: https://snowflakecomputing.atlassian.net/browse/SNOW-1491175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2108, opened 2024-08-15
Server side snowpark caller frame depth

sfc-gh-heshah:


src/snowflake/snowpark/column.py
```diff
@@ -613,10 +613,10 @@ def bitxor(self, other: Union[ColumnOrLiteral, Expression]) -> "Column":
         build_expr_from_snowpark_column_or_python_val(ast.rhs, other)
         return Column(BitwiseXor(Column._to_expr(other), self._expression), ast=expr)
 
-    def __neg__(self) -> "Column":
+    def __neg__(self, _caller_frame_depth=2) -> "Column":
```
sfc-gh-azwiegincew, line 616: Is there ever a possibility of passing a different value here? If not (and I don't think there is), you should just pass the constant 2 in line 619. Same everywhere except where you actually have a use case for parametrization.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -51,120 +57,138 @@ def build_expr_from_python_val(expr_builder: proto.Expr, obj: Any) -> None:
...
+        big_decimal_val_ast.unscaled_value = unscaled_val.to_bytes(
             req_bytes, "big", signed=True
         )
-        expr_builder.big_decimal_val.scale = dec_tuple.exponent
+        big_decimal_val_ast.scale = dec_tuple.exponent
+        big_decimal_val_ast.scale = dec_tuple.exponent
 
     elif isinstance(obj, datetime.datetime):
-        set_src_position(expr_builder.python_timestamp_val.src)
+        python_timestamp_val_ast = with_src_position(expr_builder.python_timestamp_val, caller_frame_depth=caller_frame_depth)
```
sfc-gh-azwiegincew, line 118: IMO these names are too long for local variables given the simple context they're in (and therefore the lack of need for any kind of disambiguation). i would just them `ast`.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -346,84 +370,98 @@ def set_builtin_fn_alias(ast: proto.Expr, alias: str) -> None:
...
+            assert SNOWPARK_LIB_PATH in Path(last_snowpark_file).parents
+            first_non_snowpark_file = frame.f_code.co_filename
+            assert SNOWPARK_LIB_PATH not in Path(first_non_snowpark_file).parents
+
+        # Once we've stepped out of the snowpark package, we should be in the code of interest.
+        # However, the code of interest may execute in an environment that is not accessible via the filesystem.
+        # e.g. Jupyter notebooks, REPLs, calls to exec, etc.
+        filename = frame.f_code.co_filename if frame is not None else "<unknown>"
+        if frame is None or not Path(filename).is_file():
+            src.file = "<unknown>"
```
sfc-gh-azwiegincew, line 438: I'd leave the file as an empty string and let whoever renders the UI decide what to do.

src/snowflake/snowpark/column.py
```diff
@@ -691,82 +693,102 @@ def __invert__(self) -> "Column":
...
+        if emit_ast:
+            expr = proto.Expr()
+            if try_:
+                ast = with_src_position(expr.sp_column_try_cast, caller_frame_depth=caller_frame_depth)
+                ast.col.CopyFrom(self._ast)
+                to._fill_ast(ast.to)
+            else:
+                ast = with_src_position(expr.sp_column_cast, caller_frame_depth=caller_frame_depth)
+                ast.col.CopyFrom(self._ast)
+                to._fill_ast(ast.to)
```
sfc-gh-azwiegincew, line 715:
```
ast = with _src_position(exp.sp_column_try_cast if try_ else expr.sp_column_cast, ...)
ast.col.CopyFrom...
to._fill_ast...
```

src/snowflake/snowpark/column.py
sfc-gh-azwiegincew: Oh, I see you're actually using this – great!

sfc-gh-heshah:
Addressed the feedback here in #2074, as I'll probably be leaving this as a draft 😅

PR #2118, opened 2024-08-16
SNOW-1540820 Support index.identical

sfc-gh-azhan:

   

   Fixes SNOW-1540820 Support index.identical


src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1203,18 +1205,37 @@ def equals(self, other: Any) -> bool:
...
-    def identical(self) -> None:
+    def identical(self, other: Any) -> bool:
         """
         Similar to equals, but checks that object attributes and types are also equal.
 
         Returns
         -------
         bool
-            If two Index objects have equal elements and the same type True,
+            If two Index objects have equal elements and same type True,
```
sfc-gh-azhan, line 1215: those code are copied from pandas

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -16587,7 +16587,12 @@ def index_equals(self, other: "SnowflakeQueryCompiler") -> bool:
             )
         }
 
-        rows = join_result.result_frame.ordered_dataframe.agg(agg_exprs).collect()
+        try:
+            rows = join_result.result_frame.ordered_dataframe.agg(agg_exprs).collect()
+        except Exception:
```
sfc-gh-nkumar, line 16592: One potential issue we have here it might incorrectly return False for other type of errors se well for example network failures.
Can you please specifically check of exception where this should be False? Another option is to perform this type check before doing value check and return False if types don't align.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -1203,18 +1205,37 @@ def equals(self, other: Any) -> bool:
...
+        True
+
+        >>> idx1 = pd.Index(['1', '2', '3'], name="A")
+        >>> idx2 = pd.Index(['1', '2', '3'], name="B")
+        >>> idx2.identical(idx1)
+        False
         """
-        # TODO: SNOW-1458148 implement identical
+        return (
+            self.equals(other)
```
sfc-gh-nkumar, line 1231: nit: we can move lightweight checks before this `self.equals` which is a join query. This way we can avoid this if other checks don't pass.

tests/integ/modin/index/test_index_methods.py
```diff
@@ -400,3 +400,22 @@ def test_create_index_from_df_negative():
...
     with pytest.raises(ValueError):
         pd.DatetimeIndex(pd.DataFrame([[1, 2], [3, 4]]))
+
+
+@sql_count_checker(query_count=4, join_count=4)
+def test_identical():
+    i1 = pd.Index(["a", "b", "c"])
+    i2 = pd.Index(["a", "b", "c"])
+
+    assert i1.identical(i2)
```
sfc-gh-helmeleegy, line 410: why not compare to native pandas?

sfc-gh-azhan: this is the test copied from pandas too.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -16580,7 +16580,17 @@ def index_equals(self, other: "SnowflakeQueryCompiler") -> bool:
             )
         }
 
-        rows = join_result.result_frame.ordered_dataframe.agg(agg_exprs).collect()
+        try:
+            rows = join_result.result_frame.ordered_dataframe.agg(agg_exprs).collect()
+        except SnowparkSQLException as e:
```
sfc-gh-vbudati, line 16585: I'm not sure if there are different error messages for different type casts. I've seen something like:
```
SnowparkSQLException: (1304): 01b1d9a9-0507-c7de-000c-a9033d54baf7: 100037 (22018): Boolean value 'bad value' is not recognized
```
before (this is from Rehan's fillna type casting doc).
I'm not sure if this error can be raised and if there are similar type errors but with different text output. Thought it was worth pointing out.

sfc-gh-vbudati (commented):
Everything looks good, had one comment on the type casting error messages.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-azhan: good point, I think it's better to loose the condition here then.

sfc-gh-helmeleegy (approved):
LGTM, thanks Andong!

sfc-gh-vbudati (approved):

PR #2149, opened 2024-08-22
[SNOW-1636750] Add MockServerConnection check when suppressing results.

sfc-gh-lspiegelberg:
Fixes a few doctests, by making sure the `_suppress_not_implemented_error` check is only carried out for MockServerConnection. The regular ServerConnection should a) not have such a field, b) work for the APIs in general.

Other:
- Fix Snowpark python `WithinGroup` to receive `Expression` object instead of `Column`. Add check to make sure type hint is followed, so analyzer doesn't break internally.
- Fix APIs `sort, alias, agg, groupby, except_, natural_join, generator` to properly assign stmt ids.
- Fix TableFunction client-side where incorrect field was set (used `table_fn_call_type__builtin_fn`, but with current IR it is actually `table_fn_call_type__builtin`).

This PR reduces current doctest failures from
`= 58 failed, 342 passed, 1 skipped, 3451 deselected, 44 warnings in
322.36s (0:05:22) =`
to
`37 failed, 363 passed, 1 skipped, 3451 deselected, 44 warnings`


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -162,7 +162,10 @@ def build_expr_from_python_val(expr_builder: proto.Expr, obj: Any) -> None:
         ast = with_src_position(expr_builder.tuple_val)
         for v in obj:
             build_expr_from_python_val(ast.vs.add(), v)
-
+    elif isinstance(obj, snowflake.snowpark.dataframe.DataFrame):
+        ast = with_src_position(expr_builder.sp_dataframe_ref)
+        assert obj._ast_id is not None, "Dataframe must have valid id"
```
sfc-gh-azwiegincew, line 167: The message might be cryptic in some situations. I'd say something about missing AST.

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: fixed!

PR #2162, opened 2024-08-23
SNOW-1058583: Add support for iceberg tables to create_table and copy_into_table statements.

sfc-gh-jrose:

   

   Fixes SNOW-1058583

   This PR adds support for iceberg table parameters to save_as_table and saveAsTable. I also looked into copy_into_table, but that appears to work already without the `ICEBERG` keyword. I've passed the parameters through in this function as well in case it needs to create a table.
    The Python test environment has weak iceberg support so these tests may need to be disabled if they prove to be flaky.


tests/unit/test_analyzer_util_suite.py
```diff
@@ -353,3 +352,38 @@ def test_join_statement_negative():
...
+        catalog_sync="integration_name",
+        storage_serialization_policy="OPTIMIZED",
+    ) == (
+        " CREATE    ICEBERG  TABLE test_table(test_col varchar)  EXTERNAL_VOLUME  = 'example_volume' "
+        " CATALOG  = 'example_catalog'  BASE_LOCATION  = '/root'  CATALOG_SYNC  = 'integration_name'"
+        "  STORAGE_SERIALIZATION_POLICY  = 'OPTIMIZED' "
+    )
+
+
+def test_create_iceberg_table_as_select_statement():
```
sfc-gh-aalam, line 374: can we try to see if we can add a simple integration test as well. Maybe use ddl statement to check if the properties are set correctly? Might be harder since we are dealing with iceberg tables.

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -246,6 +268,12 @@ def save_as_table(
                 max_data_extension_time,
                 change_tracking,
                 copy_grants,
+                is_iceberg,
```
sfc-gh-aalam, line 271: for iceberg only params, can we raise a value error if those params are provided when iceberg table is not set to true?

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -129,6 +129,12 @@ def save_as_table(
         max_data_extension_time: Optional[int] = None,
         change_tracking: Optional[bool] = None,
         copy_grants: bool = False,
+        is_iceberg: bool = False,
```
sfc-gh-yixie, line 132: How about the `df.copy_into_table()`?

sfc-gh-jrose: I tested out copy_into_table statements and they work without the iceberg keyword even if you're copying in to an iceberg table, so I don't think we need to update our sql to include it.

sfc-gh-yixie: `copy_into_table` also automatically creates a table if it doesn't exist. So better also add a parameter.

src/snowflake/snowpark/dataframe.py
```diff
@@ -3026,6 +3026,12 @@ def copy_into_table(
         transformations: Optional[Iterable[ColumnOrName]] = None,
         format_type_options: Optional[Dict[str, Any]] = None,
         statement_params: Optional[Dict[str, str]] = None,
+        is_iceberg: bool = False,
+        external_volume: Optional[str] = None,
+        catalog: Optional[str] = None,
+        base_location: Optional[str] = None,
+        catalog_sync: Optional[str] = None,
+        storage_serialization_policy: Optional[str] = None,
```
sfc-gh-yixie, line 3034: I'm hesitating whether we should put these into a dictionary with parameter `iceberg_config` so we group them altogether. Same for save_as_table. What do you think?

tests/integ/scala/test_dataframe_writer_suite.py
```diff
@@ -152,6 +153,57 @@ def test_write_with_target_table_autoincrement(
...
+        pytest.skip("Test requires iceberg support.")
+
+    table_name = Utils.random_table_name()
+    df = session.create_dataframe(
+        [],
+        schema=StructType(
+            [
+                StructField("a", StringType()),
+                StructField("b", IntegerType()),
+            ]
```
sfc-gh-yixie, line 167: Does this work for ArrayType and MapType?

src/snowflake/snowpark/dataframe.py
sfc-gh-jrose: That seems alright to me. There isn't any precedent to follow from spark since our io is already quite different. That might be a little bit more boilerplate for the user, but these values aren't very variable so they could probably just make a constant dict of the settings and use it whenever there is a save_as_table or copy_into call.

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-jrose: Added this functionality.

src/snowflake/snowpark/dataframe.py
sfc-gh-jrose: I've refactored based on this suggestion.

tests/integ/scala/test_dataframe_writer_suite.py
sfc-gh-jrose: After some testing it looks like there's several edge cases that break for structured types. This change is large already so I will address those edge cases in a second PR.

tests/integ/scala/test_dataframe_copy_into.py
```diff
@@ -245,6 +245,36 @@ def test_copy_csv_basic(session, tmp_stage_name1, tmp_table_name):
...
+    test_file_on_stage = f"@{tmp_stage_name1}/{test_file_csv}"
+    df = session.read.schema(user_schema).csv(test_file_on_stage)
+
+    test_table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
+    df.copy_into_table(
+        test_table_name,
+        iceberg_config={
+            "external_volume": "PYTHON_CONNECTOR_ICEBERG_EXVOL",
+            "catalog": "SNOWFLAKE",
+            "base_location": "snowpark_python_tests",
```
sfc-gh-aalam, line 262: can we also support them in a case insensitive way. i.e.

```
iceberg_config={
            "external_volume": "PYTHON_CONNECTOR_ICEBERG_EXVOL",
            "CATALOG": "SNOWFLAKE",
            "BASE_LOCATION": "snowpark_python_tests"
}
```
should also work

sfc-gh-aalam (approved):
looks good. Just need to make options case insensitive

CHANGELOG.md
```diff
@@ -19,6 +19,19 @@
   - `max_data_extension_time`
   - `change_tracking`
   - `copy_grants`
+  - `iceberg_config` A dicitionary that can hold the follwoing iceberg configuration options:
```
sfc-gh-yixie, line 22: Nit: `following` instead of `follwoing`.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -778,25 +784,40 @@ def create_table_statement(
...
-        }
-    )
+    options = {
+        ENABLE_SCHEMA_EVOLUTION: enable_schema_evolution,
+        DATA_RETENTION_TIME_IN_DAYS: data_retention_time,
+        MAX_DATA_EXTENSION_TIME_IN_DAYS: max_data_extension_time,
+        CHANGE_TRACKING: change_tracking,
+    }
+
+    if iceberg_config is not None:
```
sfc-gh-yixie, line 802: Suggest also checking empty dict. `if iceberg_config and "external_volume" in  iceberg_config`.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -778,25 +784,40 @@ def create_table_statement(
...
+                ),
+            }
+        )
+
+    options_statement = get_options_statement(options)
     return (
         f"{CREATE}{(OR + REPLACE) if replace else EMPTY_STRING}"
         f" {(get_temp_type_for_object(use_scoped_temp_objects, is_generated) if table_type.lower() in TEMPORARY_STRING_SET else table_type).upper()} "
-        f"{TABLE}{table_name}{(IF + NOT + EXISTS) if not replace and not error else EMPTY_STRING}"
+        f"{ICEBERG if iceberg_config is not None else EMPTY_STRING}{TABLE}{table_name}{(IF + NOT + EXISTS) if not replace and not error else EMPTY_STRING}"
```
sfc-gh-yixie, line 820: Same comment as above.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -871,18 +893,31 @@ def create_table_as_select_statement(
...
-            CHANGE_TRACKING: change_tracking,
-        }
-    )
+    options = {
+        ENABLE_SCHEMA_EVOLUTION: enable_schema_evolution,
+        DATA_RETENTION_TIME_IN_DAYS: data_retention_time,
+        MAX_DATA_EXTENSION_TIME_IN_DAYS: max_data_extension_time,
+        CHANGE_TRACKING: change_tracking,
+    }
+    if iceberg_config is not None:
```
sfc-gh-yixie, line 902: same as above

src/snowflake/snowpark/dataframe.py
```diff
@@ -3079,6 +3080,19 @@ def copy_into_table(
             transformations: A list of column transformations.
             format_type_options: A dict that contains the ``formatTypeOptions`` of the ``COPY INTO <table>`` command.
             statement_params: Dictionary of statement level parameters to be set while executing this action.
+            iceberg_config: A dictionary that can contain the following iceberg confiration values:
```
sfc-gh-yixie, line 3083: configuration instead of confiration

sfc-gh-yixie (changes requested):

sfc-gh-yixie (approved):

PR #2165, opened 2024-08-26
[SNOW-1636729] Improve join/align performance by avoid unnecessary coalesce

sfc-gh-yzou:

   

   Fixes SNOW-1636729

Concatenating series from the same dataframe requires joins today, but theoretically there is no join needed since they come from the same dataframe.

The reason those join are there is because concat calls join_on_index to join two series at a time, the join is only optimized today if the join is performed on the row position column, but join_on_index tries to creates a new column using coalesce even if the two columns to coalesce is the same column. Therefore, in the next join, the two columns to join on can not be recognized anymore.

In this pr, we tries to maximize the optimization opportunity by keeping the original column whenever possible, for example, if the two columns to coalesce is actually the same column, no actual coalesce is required.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
@@ -343,42 +343,52 @@ def _create_internal_frame_with_join_or_align_result(
             right_col = result_helper.map_right_quoted_identifiers([origin_right_col])[
                 0
             ]
-            # Coalescing is only required for 'outer' or 'asof' joins or align.
-            # For 'inner' and 'left' join we use left join keys and for 'right' join we
-            # use right join keys.
-            # For 'left' and 'coalesce' align we use left join keys.
-            if how in ("asof", "outer"):
-                # Generate an expression equivalent of
-                # "COALESCE('left_col', 'right_col') as 'left_col'"
-                coalesce_column_identifier = (
-                    result_ordered_frame.generate_snowflake_quoted_identifiers(
-                        pandas_labels=[
-                            extract_pandas_label_from_snowflake_quoted_identifier(
-                                left_col
-                            )
-                        ],
-                    )[0]
-                )
-                coalesce_column_identifiers.append(coalesce_column_identifier)
-                coalesce_column_values.append(coalesce(left_col, right_col))
-                if origin_left_col_type == origin_right_col_type:
-                    coalesce_col_type = origin_left_col_type
-            elif how == "right":
-                # No coalescing required for 'right' join. Simply use right join key
-                # as output column.
-                coalesce_column_identifier = right_col
-                coalesce_col_type = origin_right_col_type
-            elif how in ("inner", "left", "coalesce"):
-                # No coalescing required for 'left' or 'inner' join and for 'left' or
-                # 'coalesce' align. Simply use left join key as output column.
+            # if the result ordering column has the same ordering columns as the original left ordering columns,
+            # that means the original left and right shares the same base, and no actual join is applied.
+            no_join_applied = (
+                result_ordered_frame.ordering_columns
+                == left.ordered_dataframe.ordering_columns
+            )
+            if no_join_applied and origin_left_col == origin_right_col:
                 coalesce_column_identifier = left_col
                 coalesce_col_type = origin_left_col_type
             else:
-                raise AssertionError(f"Unsupported join/align type {how}")
+                # Coalescing is only required for 'outer' or 'asof' joins or align.
+                # For 'inner' and 'left' join we use left join keys and for 'right' join we
+                # use right join keys.
+                # For 'left' and 'coalesce' align we use left join keys.
+                if how in ("asof", "outer"):
+                    # Generate an expression equivalent of
+                    # "COALESCE('left_col', 'right_col') as 'left_col'"
+                    coalesce_column_identifier = (
+                        result_ordered_frame.generate_snowflake_quoted_identifiers(
+                            pandas_labels=[
+                                extract_pandas_label_from_snowflake_quoted_identifier(
+                                    left_col
+                                )
+                            ],
+                        )[0]
+                    )
+                    coalesce_column_identifiers.append(coalesce_column_identifier)
+                    coalesce_column_values.append(coalesce(left_col, right_col))
+                    if origin_left_col_type == origin_right_col_type:
+                        coalesce_col_type = origin_left_col_type
+                elif how == "right":
+                    # No coalescing required for 'right' join. Simply use right join key
+                    # as output column.
+                    coalesce_column_identifier = right_col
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -319,42 +319,62 @@ def _create_internal_frame_with_join_or_align_result(
...
-            elif how == "right":
-                # No coalescing required for 'right' join. Simply use right join key
-                # as output column.
-                coalesce_column_identifier = right_col
-                coalesce_col_type = origin_right_col_type
-            elif how in ("inner", "left", "coalesce"):
-                # No coalescing required for 'left' or 'inner' join and for 'left' or
-                # 'coalesce' align. Simply use left join key as output column.
+            # if the result ordering column has the same ordering columns as the original left ordering columns,
+            # that means the original left and right shares the same base, and no actual join is applied.
```
sfc-gh-mvashishtha, line 323: I don't understand why this is true. should we explain more in the comment here?

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -319,42 +319,62 @@ def _create_internal_frame_with_join_or_align_result(
...
+            no_join_applied = (
+                result_ordered_frame.ordering_columns
+                == left.ordered_dataframe.ordering_columns
+            )
+            if no_join_applied and origin_left_col == origin_right_col:
```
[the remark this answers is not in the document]
sfc-gh-yzou: this should be the right place to do this optimization, because this is the function that has been called by all join and align methods, and the improvement will be applied to all calls automatically.

Seems we previously already did an optimization for one align scenario, if we do it here, we shouldn't need the logic there, let see if i can remove the change in the align

tests/integ/modin/binary/test_binary_op.py
```diff
@@ -2586,3 +2586,19 @@ def test_df_sub_series():
     eval_snowpark_pandas_result(
         snow_df, native_df, lambda df: df.sub(df["two"], axis="index"), inplace=True
     )
+
+
+@sql_count_checker(query_count=1, join_count=0)
+def test_binary_op_series_from_same_df():
```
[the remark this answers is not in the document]
sfc-gh-yzou: The test seems only have 1 align, this test more than 1 align, because typically the first align can already be optimized away without the optimization. let me add some comment here

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -1192,10 +1212,11 @@ def align(
     # join columns of left and right matches, then there is no need to coalesce the join
     # keys, simply inherent from left gives the correct result.
     # Retaining the original columns also helps avoid unnecessary join in later steps.
-    if (
-        how == "outer"
-        and aligned_ordered_frame.ordering_columns != left.ordering_columns
-    ):
+    # if (
+    #    how == "outer"
```
sfc-gh-yzou, line 1216: This seems a duplication with the current optimization. remove it for testing to see anything fails

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -1187,15 +1217,8 @@ def align(
     # NULL  NULL 2             NULL      4   e  2
     coalesce_key_config = None
     inherit_join_index = InheritJoinIndex.FROM_LEFT
-    # When it is `outer` align, we need to coalesce the align columns. However, if the
```
sfc-gh-yzou, line 1190: @sfc-gh-nkumar this optimization is now performed in general for all join and align

tests/integ/modin/test_concat.py
```diff
@@ -1063,3 +1063,31 @@ def test_concat_keys():
...
     assert_frame_equal(snow_df, native_df, check_dtype=False)
+
+
+@sql_count_checker(query_count=4, join_count=0)
+def test_concat_series_from_same_df(join):
+    num_cols = 4
+    select_data = [f'{i} as "{i}"' for i in range(num_cols)]
+    query = f"select {', '.join(select_data)}"
+
+    df = pd.read_snowflake(query)
```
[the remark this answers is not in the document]
sfc-gh-yzou: This is right now using read_snowflake for two reasons
1) the join optimization can only be applied when join is on row_position column, only when using read_snowflake, the default row_position column is created by snowpark pandas, and used the same as index column. When you start with a pandas dataframe, the index and row position column are actually two different columns when write back to snowpark, and the optimization can not be applied. Let me add some comment there.
2) this is the reported use case also.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -337,44 +351,60 @@ def _create_internal_frame_with_join_or_align_result(
...
+                coalesce_config = JoinKeyCoalesceConfig.LEFT
                 coalesce_column_identifier = left_col
                 coalesce_col_type = origin_left_col_type
             else:
-                raise AssertionError(f"Unsupported join/align type {how}")
+                # Coalescing is only required for 'outer' or 'asof' joins or align.
+                # For 'inner' and 'left' join we use left join keys and for 'right' join we
+                # use right join keys.
+                # For 'left' and 'coalesce' align we use left join keys.
+                if how in ("asof", "outer"):
```
sfc-gh-yzou, line 380: @sfc-gh-nkumar When i start looking at this part of code now, it feels little bit wired, it seems we are deciding how the column coalesce happens, we do not look into the coalesce configure, but the join type. It is good in the sense that it tries to reduce the extra logic caller need to check, but it is kind of confusing in the sense that we also have an coalesce configure parameter there

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-yzou: this is the behavior guaranteed by the ordering dataframe method today, for regular join/align, the result ordering column will always be combination of left and right ordering to provide the guarantee of uniquess. Only when optimization happens (when join is performed on dfs from the same snowpark dataframe on the same row_position column, left ordering column only is used.
I added more comment there

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkumar: Right, IIRC correctly logic is to ignore the coalesce config for join types where coalesce is not required. We can probably assert if coalesce config is provided for join types which do not require assert. But I don't feel strongly either way.

sfc-gh-nkumar (approved):

sfc-gh-mvashishtha (approved):
Thanks for addressing my comments! LGTM

PR #2059, opened 2024-08-08
SNOW-1119855: Push base unimplemented methods to query compiler when possible (1/?)

sfc-gh-joshi:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-1119855 (first of several PRs)

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.

3. Please describe how your code solves the related issue.

This PR prepares for the imminent removal of base.py from the Snowpark pandas codebase by pushing as many unimplemented methods down to the query compiler as possible. Some unimplemented methods must still be overridden because Modin uses other implemented query compiler/frontend methods in their implementation; these methods have been moved to `src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py`.

### The following methods are affected:

The below BasePandasDataset methods either have existing corresponding query compiler stubs that can be used directly, or default to pandas *at the query compiler layer*, which Snowpark pandas specifies to raise a NotImplementedError.
<details>
<summary><b>Pushed to the query compiler</b></summary>
<ul>
<li>align</li>
<li>at_time (via between_time)</li>
<li>between_time</li>
<li>clip</li>
<li>combine (via binary_op)</li>
<li>combine_first (via binary_op)</li>
<li>explode</li>
<li>infer_objects</li>
<li>kurt</li>
<li>kurtosis (via kurt)</li>
<li>mode</li>
<li>sem</li>
<li>tz_convert (via tz_convert)</li>
<li>tz_localize (via tz_localize)</li>
<li>__sizeof__</li>
</ul>
</details>

The below methods do not have corresponding query compiler methods, and either (1) would work in Snowpark pandas but haven't been tested/need too many queries, (2) use frontend/other query compiler methods that we cannot override, or (3) manipulates index/columns in ways that we may not support.
<details>
<summary><b>Moved to base_overrides.py</b></summary>
<li>asof: modin uses dropna + reindex</li>
<li>bool: method is deprecated in pandas but pretty trivial</li>
<li>droplevel: modin does manipulations on column/index objects</li>
[...]

PR #2167, opened 2024-08-26
SNOW-1119855: Remove modin/pandas/base.py (2/2)

sfc-gh-joshi:

   

   Fixes SNOW-1119855

This followup to #2059 removes our vendored copy of `base.py` altogether from the codebase. Many methods are still overridden in `snowflake/snowpark/modin/plugin/extensions/base_overrides.py`, with the reason for overriding each given inline in the comments. These methods are as follows:
<details>
<summary>list</summary>
<ul>
<li>agg/aggregate</li>
<li>_agg_helper</li>
<li>various aggregations (count, max, min, mean, median, std, sum, var)</li>
<li>_binary_op</li>
<li>_dropna</li>
<li>fillna</li>
<li>isin</li>
<li>quantile</li>
<li>_to_series_list</li>
<li>shift</li>
<li>skew</li>
<li>resample</li>
<li>expanding</li>
<li>rolling</li>
<li>indexer properties (iloc, loc, iat, at)</li>
<li>__getitem__</li>
<li>sort_values</li>
<li>where</li>
<li>mask</li>
<li>to_csv</li>
<li>sample</li>
<li>pct_change</li>
<li>astype</li>
<li>drop</li>
<li>__len__</li>
<li>set_axis</li>
<li>describe</li>
<li>diff</li>
<li>tail</li>
<li>idxmax</li>
<li>idxmin</li>
<li>unary operators (abs, __invert__, __neg__)</li>
<li>rename_axis</li>
<li>__array__ufunc__</li>
<li>reindex</li>
<li>_get_index</li>
<li>_set_index</li>
</ul>
</details>

Some of these differences can be upstreamed fairly easily, and we will work to upstream them once the updated modin build process for Snowflake becomes clearer. Some methods will require significantly more work to reconcile.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1462,6 +1460,12 @@ def set_columns(self, new_pandas_labels: Axes) -> "SnowflakeQueryCompiler":
         )
         return SnowflakeQueryCompiler(new_internal_frame)
 
+    # TODO SNOW-837664: add more tests for df.columns
```
sfc-gh-rdurrani, line 1463: Minor nit, but should we separate this out into two methods? i.e. def _get_columns(self), def _set_columns(self, labels), columns = property(_get_columns, _set_columns)? It's more verbose, but looks nicer, and is consistent with the rest of the codebase, but again, bot a blocker.

sfc-gh-rdurrani (commented):
LGTM overall - left a minor nit, and a minor question (just to confirm my understanding). Would approve, but I think we may need to add testing/annotations in order to get the coverage test passing!

src/snowflake/snowpark/modin/pandas/__init__.py
```diff
@@ -220,7 +217,6 @@ def __getattr__(name: str) -> Any:
     "date_range",
     "Index",
     "MultiIndex",
-    "Series",
```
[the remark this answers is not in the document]
sfc-gh-rdurrani: Sorry - to clarify, I'm assuming that this change is fixing a previous typo, since Series shows up twice in the list.

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -303,3 +424,1901 @@ def truncate(
...
+#    https://github.com/modin-project/modin/issues/7104
+# 4. Snowpark pandas uses different default arguments from modin. This occurs if some parameters are
+#    only partially supported (like `numeric_only=True` for `skew`), but this behavior should likewise
+#    be revisited.
+
+# `aggregate` for axis=1 is performed as a call to `BasePandasDataset.apply` in upstream Modin,
+# which is unacceptable for Snowpark pandas. Upstream Modin should be changed to allow the query
+# compiler or a different layer to control dispatch.
+@register_base_override("aggregate")
+def aggregate(
```
sfc-gh-yzou, line 450: I assume those are directly copied from the base.py. Or is there any difference compare with the original file? if yes, can you mark it here?

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -6,31 +6,152 @@
...
+from snowflake.snowpark.modin.plugin.utils.error_message import (
+    ErrorMessage,
+    base_not_implemented,
+)
+from snowflake.snowpark.modin.plugin.utils.warning_message import WarningMessage
+from snowflake.snowpark.modin.utils import validate_int_kwarg
+
+
+def register_base_override(method_name: str):
+    def decorator(base_method: Any):
```
sfc-gh-yzou, line 90: what is this function try to do here ? can yo add some comment?

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -6,31 +6,152 @@
...
+        if callable(base_method) and (
+            not method_name.startswith("_")
+            or (method_name in TELEMETRY_PRIVATE_METHODS)
```
sfc-gh-yzou, line 93: also, now since the base class is not declared with metaclass=TelemetryMeta, does that mean we will miss telemetry the comes from the base class? if that is the case, this could be a regression for telemetry cc @sfc-gh-azhan

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
```diff
@@ -218,3 +218,36 @@ def cache_result(self, inplace: bool = False) -> Optional[pd.Series]:
...
+    types : tuple
+        The types of the args.
+    args : tuple
+        The args to the func.
+    kwargs : dict
+        Additional keyword arguments.
+
+    Returns
+    -------
+    BasePandasDataset
```
sfc-gh-yzou, line 242: originally that seems in base but now it is duplicated in both series and dataframe, can this be in base_overrides to avoid duplication?

tests/unit/modin/modin/test_envvars.py
```diff
@@ -90,6 +90,32 @@ def test_custom_help(make_custom_envvar):
...
+    # TODO: once modin 0.31.0 is available, use the actual modin DocModule class
+    snowflake.snowpark.modin.utils._inherit_docstrings(
+        classes.BasePandasDataset,
+        overwrite_existing=True,
+    )(pd.base.BasePandasDataset)
+    DocModule.put("docs_module")
+
+
+DOC_OVERRIDE_XFAIL_REASON = (
+    "test docstring overrides currently cannot override real docstring overrides until "
```
sfc-gh-yzou, line 114: what does that mean?

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-azhan: @sfc-gh-joshi Let's add a test to make sure we have telemetry for those cases. E.g., `series.mean` which does not have a series mean method but only use the base's mean method.

sfc-gh-azhan (changes requested):

src/snowflake/snowpark/modin/pandas/__init__.py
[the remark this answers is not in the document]
sfc-gh-joshi: Yes, this fixes the previous typo.

src/snowflake/snowpark/modin/plugin/__init__.py
```diff
@@ -63,15 +63,22 @@
...
-    docstrings.series_utils.StringMethods,
-    overwrite_existing=True,
-)(modin.pandas.series_utils.StringMethods)
-
-modin.utils._inherit_docstrings(
-    docstrings.series_utils.CombinedDatetimelikeProperties,
-    overwrite_existing=True,
-)(modin.pandas.series_utils.DatetimeProperties)
+# TODO: https://github.com/modin-project/modin/issues/7113 and https://github.com/modin-project/modin/issues/7134
+# Upstream Modin has issues with certain docstring generation edge cases, so we should use our version instead
```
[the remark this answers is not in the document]
sfc-gh-joshi: These are fixed on main, but I don't think they're in any released version yet.

src/snowflake/snowpark/modin/pandas/__init__.py
```diff
@@ -318,8 +314,7 @@ def __getattr__(name: str) -> Any:
 # Manually re-export the members of the pd_extensions namespace, which are not declared in __all__.
 _EXTENSION_ATTRS = ["read_snowflake", "to_snowflake", "to_snowpark", "to_pandas"]
 # We also need to re-export native_pd.offsets, since modin.pandas doesn't re-export it.
-# snowflake.snowpark.pandas.base also needs to be re-exported to make docstring overrides for BasePandasDataset work.
-_ADDITIONAL_ATTRS = ["offsets", "base"]
```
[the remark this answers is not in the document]
sfc-gh-joshi: I think this import has actually been unnecessary for a while now--if I make this change on `main`, docstring overrides still work as expected (both in HTML output and via the `help` command).

This re-export (plus the `from snowflake.snowpark.modin.pandas import base` statement above) were previously necessary for Sphinx to build, but I think since after #1940 we stopped referencing `snowflake.snowpark.modin.pandas` in favor of `modin.pandas`, this stopped being necessary. Either way, it is no longer possible to do this re-export because Snowpark pandas is no longer declaring its own `modin.pandas.base` sub-module.

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-joshi: Good point. `mean` is actually overridden here, so it gets our telemetry annotation, but I've added a test for `copy`, which is inherited straight from `BasePandasDataset`. I had to add a little extra logic to `TelemetryMeta` to make it work, and I'll revisit it when I remove dataframe/series.py.

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-joshi: Upstream modin doesn't have an equivalent of `register_dataframe_override` for BasePandasDataset, so this method tries to simulate that behavior. I added a comment to that effect.

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-joshi: `iloc` is overridden within this file, and it already has an existing test in `test_telemetry`.

sfc-gh-joshi:
> LGTM overall - left a minor nit, and a minor question (just to confirm my understanding). Would approve, but I think we may need to add testing/annotations in order to get the coverage test passing!

@sfc-gh-rdurrani sounds good, I'll do a pass on coverage after this CI run.

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-azhan: nice!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -385,6 +385,8 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
     this class is best explained by looking at https://github.com/modin-project/modin/blob/a8be482e644519f2823668210cec5cf1564deb7e/modin/experimental/core/storage_formats/hdk/query_compiler.py
     """
 
+    lazy_execution = True
```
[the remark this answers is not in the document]
sfc-gh-yzou: So If I understand correctly, the lazy_execution is used to skip some of the checks that may require eager evaluation, but it may skip more checks than what we expected today. For example, for the astype, the column key check is skipped because it is lazy_execution, but we don't need to skip the check since our column labels are eagerly evaluated at client side.

Right now, we set lazy_execution check, and the check got skipped. However, in our original implementation, we actually have the check, did we overwrite the method in base_overrides, or is there a behavior change?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: sorry, where do we have _set_columns, I only see set_columns. why do we need so many set_columns? And i agree the function name should be consistent, either all with _, or both not

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1462,6 +1460,12 @@ def set_columns(self, new_pandas_labels: Axes) -> "SnowflakeQueryCompiler":
         )
         return SnowflakeQueryCompiler(new_internal_frame)
 
+    # TODO SNOW-837664: add more tests for df.columns
+    def _get_columns(self) -> native_pd.Index:
+        return self._modin_frame.data_columns_index
+
+    columns: native_pd.Index = property(_get_columns, set_columns)
```
sfc-gh-yzou, line 1467: so what would be the correct way to do this in the future, i think the same problem also applies to index. in general, i would think having get_columns, and set_columns two functions implemented at query compiler layer is very clear, and the frontend layer should call those method correctly

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10724,6 +10734,12 @@ def is_multiindex(self, *, axis: int = 0) -> bool:
         """
         return self._modin_frame.is_multiindex(axis=axis)
 
+    def abs(self) -> "SnowflakeQueryCompiler":
```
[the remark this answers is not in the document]
sfc-gh-yzou: i think that is fine for now.

Based on whatever we have discussed before for modin upstream, eventually we do want to having one function per method, and let the query compiler layer decide whether to use the same or different implementation, and the frontend shouldn't not decide when to default to pandas, it should be controlled by the backend

src/snowflake/snowpark/modin/plugin/utils/frontend_constants.py
```diff
@@ -0,0 +1,26 @@
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
+# Do not look up certain attributes in columns or index, as they're used for some
```
sfc-gh-yzou, line 5: you mentioned we want to take a look to see if we could unify this with the definition in modin, can we get a ticket to track this?

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-yzou: @sfc-gh-joshi Does that mean for a method that is just defined in base, if they are not overridden here with the register_base_override annotation, we will not have the telemetry. Till the end,  we will align most implementation in this file with modin, and do not have much override here, what happens then?

src/snowflake/snowpark/modin/pandas/series.py
```diff
@@ -108,7 +133,7 @@
     ],
     apilink="pandas.Series",
 )
-class Series(BasePandasDataset):
+class Series(BasePandasDataset, metaclass=TelemetryMeta):
```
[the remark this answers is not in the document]
sfc-gh-joshi: I think what I referred to was overriding the `__init__` method of `Series`, not the actual class declaration itself. If we were to override the class declaration, it wouldn't be much different from what we're currently doing.

My thinking is that when we remove Series/DF, we do something like this:

```
for attr_name, attr_value in Series.__dict__:
    register_series_override(attr_name)(snowpark_pandas_telemetry_method(attr_value))
```

which would overwrite the existing method with a telemetry-enabled version.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -535,17 +536,30 @@ def __new__(
                 snowflake.snowpark.modin.pandas.window.Rolling]:
                 The modified class with decorated methods.
         """
-        for attr_name, attr_value in attrs.items():
+        attr_dict = dict(attrs.items())
+        # If BasePandasDataset, defined exclusively by upstream modin, is a parent of this class,
+        # then apply the telemetry decorator to it.
+        # https://stackoverflow.com/a/71105206
```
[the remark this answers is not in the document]
sfc-gh-joshi: Yes, I just added this.

Apparently, the `attrs` argument passed to the metaclass only contains attributes that are directly defined on the class. For example, we define `abs` in `series.py`, so `abs` shows up in the old implementation of this metaclass, and gets the telemetry annotation added.

Methods that are declared on upstream Modin's `BasePandasDataset` and NOT redefined on Series/DF, such as `copy`, aren't visited by this method in our old implementation. As such, methods like `copy` wouldn't get the telemetry decorator. The new changes I added basically forcibly add telemetry to methods inherited from `BasePandasDataset`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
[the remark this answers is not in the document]
sfc-gh-joshi: You're right, but for this reason, we currently redefine `astype` in `base_overrides.py`, so it still checks the column labels as before. We can't just override all methods that use the `lazy_execution` flag because it's used in `__getattr__` ([here](https://github.com/modin-project/modin/blob/eac21a8463d7152b3ad48a8b80d6da5ac645be6a/modin/pandas/base.py#L4223)), and modin 0.28.1 will try to default to pandas if the dataframe is empty, and it checks the `lazy_execution` flag before performing this emptiness check.

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-joshi: > if they are not overridden here with the register_base_override annotation, we will not have the telemetry.

Yes, though for now, I've pushed a change to `TelemetryMeta` to force Series/DataFrame to apply the telemetry annotation to methods they inherit from BasePandasDataset.

> Till the end, we will align most implementation in this file with modin, and do not have much override here, what happens then?

As I mentioned above: one possible approach would be to use the plugin system to iterate over all of Series/DF's methods, and replace them with a telemetry-annotated version.

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
sfc-gh-azhan: we can define a `__array_function__` in a new file called `base_extension.py` or `extension_utils.py` and reuse the code right?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: It was initially `_set_columns` when Rehan first reviewed it. I've removed that method now, and it's just `get_columns` (newly added) and `set_columns` (existing) referenced in the property.

src/snowflake/snowpark/modin/plugin/__init__.py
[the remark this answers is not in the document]
sfc-gh-joshi: SNOW-1643979 (also added in code)

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
sfc-gh-joshi: > if you look at to_snowflake, to_snowpark, and to_pandas, the fact they are defined in both series and daframe are for documentation purpose, they need slightly different description and example. The implementation is pretty much all reused.

Makes sense, I'll separate it out.

> we can define a __array_function__ in a new file called base_extension.py or extension_utils.py and reuse the code right?

Added `base_extensions.py`.

src/snowflake/snowpark/modin/plugin/utils/frontend_constants.py
sfc-gh-joshi: SNOW-1643986 (also added to code)

sfc-gh-joshi:
@sfc-gh-yzou @sfc-gh-azhan I made some changes to telemetry to more closely aligned with what's necessary after we remove series.py/dataframe.py as well. In the previous version of the PR, I added some overrides in `TelemetryMeta` (used in our vendored `Series` and `DataFrame` classes to check methods defined on the parent `BasePandasDataset` as well, but this was causing issues with extension functions, and would have to change once we stop vendoring those frontend methods as well.

Instead, I've added a pass in `snowflake/snowpark/modin/pandas/__init__.py` that runs after extensions functions are initialized, which examines all non-extension functions defined by `BasePandasDataset` and replaces them with telemetry-annotated versions via the DF/Series override system. We will be able to continue using this method (with some slight changes) once we remove our vendored copies of dataframe.py and series.py.

sfc-gh-azhan (approved):
LGTM

sfc-gh-yzou (approved):

PR #2168, opened 2024-08-26
[SNOW-1638240] Support session.read, DataframeReader

sfc-gh-lspiegelberg:
Supports `session.read` with the `DataframeReader` class for Snowpark IR.


src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -367,11 +393,20 @@ def schema(self, schema: StructType) -> "DataFrameReader":
         Returns:
             a :class:`DataFrameReader` instance with the specified schema configuration for the data to be read.
         """
+
+        # AST.
+        if _emit_ast:
+            reader = proto.SpDataframeReader()
+            ast = with_src_position(reader.sp_dataframe_reader_schema)
+            ast.reader.CopyFrom(self._ast)
```
sfc-gh-vbudati, line 401: what exactly does the `reader` field represent?

sfc-gh-vbudati (approved):
Looks great, just had one question. Thanks Leonhard!

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-lspiegelberg: Basically when you call `session.read` this returns a `DataFrameReader` object. When we build the AST up, we encode this behavior. We need a separate entity, because (unfortunately) the reader object maintains internal state.

PR #2172, opened 2024-08-27
Fix query_generator for new compilation stage after optimization applied

sfc-gh-yzou:
SNOW-1641324

The test Enable CTE Optimization merge gate is currently failing because
1) the _projection_in_str for select statement is reset properly after updating, which causes the expr_alias is updated properly during query generation
2) query parameter is not propogated properly during query generation which causes variable binding query fail

### DIFF
--- a/src/snowflake/snowpark/_internal/compiler/utils.py
+++ b/src/snowflake/snowpark/_internal/compiler/utils.py
@@ -269,16 +269,16 @@ def update_resolvable_node(
         assert node.snowflake_plan is not None
         update_resolvable_node(node.snowflake_plan, query_generator)
         node.analyzer = query_generator
-        node.expr_to_alias.update(node._snowflake_plan.expr_to_alias)
-        node.df_aliased_col_name_to_real_col_name.update(
-            node._snowflake_plan.df_aliased_col_name_to_real_col_name
-        )
 
         node.pre_actions = node._snowflake_plan.queries[:-1]
         node.post_actions = node._snowflake_plan.post_actions
         node._api_calls = node._snowflake_plan.api_calls
 
         if isinstance(node, SelectSnowflakePlan):
+            node.expr_to_alias.update(node._snowflake_plan.expr_to_alias)
+            node.df_aliased_col_name_to_real_col_name.update(
+                node._snowflake_plan.df_aliased_col_name_to_real_col_name
+            )
             node._query_params = []
             for query in node._snowflake_plan.queries:
                 if query.params:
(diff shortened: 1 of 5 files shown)


sfc-gh-yzou:
daily job started here https://github.com/snowflakedb/snowpark-python/actions/runs/10583546111

tests/integ/test_cte.py
```diff
@@ -138,6 +138,67 @@ def test_binary(session, action):
...
+        2,
+        "b",
+        1,
+        "c",
+        3,
+        "d",
+    ]
+
+    df_res = df2.union(df1).union(df2).union(df1)
+    check_result(session, df_res, expect_cte_optimized=True)
```
sfc-gh-yzou, line 186: @sfc-gh-jdu i found for this case there is only one common node detected, but by just looking at the code, there should be two common node, right? is that currently expected, if not, we can get a ticket to further investigate it

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -266,13 +269,27 @@ def update_resolvable_node(
         assert node.snowflake_plan is not None
         update_resolvable_node(node.snowflake_plan, query_generator)
         node.analyzer = query_generator
+        node.expr_to_alias.update(node._snowflake_plan.expr_to_alias)
+        node.df_aliased_col_name_to_real_col_name.update(
+            node._snowflake_plan.df_aliased_col_name_to_real_col_name
+        )
```
sfc-gh-aalam, line 275: `SelectTableFunction` does not have these properties, right? can this be moved below with _query_params update

sfc-gh-jdu: is this the reason it failed before?

tests/integ/test_cte.py
sfc-gh-jdu: Yeah I think so, so which one of df1 and df2 are not converted to CTE?

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: no, it actually is not, the reason that fails, is because we didn't set the node._projection_in_str = None,  this part is more to complete the property propogation. 

I think we will need to get some test that requires expr_to_alias and df_aliased_col_name_to_real_col_name, so far i don't think we have any test test that, i have created a ticket here https://snowflakecomputing.atlassian.net/browse/SNOW-1641640

sfc-gh-yzou: @sfc-gh-aalam yes, moved

tests/integ/test_cte.py
sfc-gh-yzou: df1 is converted, but not df2

sfc-gh-aalam (approved):

sfc-gh-jdu (approved):

PR #2179, opened 2024-08-28
SNOW-1638049: Add support for pd.to_timedelta

sfc-gh-nkumar:
Fixes SNOW-1638049

Added support for pd.to_timedelta
- Supports lazy snowpark pandas objects Series and Index.
- does not support `errors` parameter.
- does not support converting from strings (on lazy objects).

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -270,6 +270,7 @@ from snowflake.snowpark.modin.plugin._internal.snowpark_pandas_types import (
 from snowflake.snowpark.modin.plugin._internal.timestamp_utils import (
     VALID_TO_DATETIME_DF_KEYS,
     DateTimeOrigin,
+    col_to_timedelta,
     generate_timestamp_col,
     raise_if_to_datetime_not_supported,
     to_snowflake_timestamp_format,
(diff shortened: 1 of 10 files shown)


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -6223,6 +6224,86 @@ def dataframe_to_datetime(
...
+            errors : {'ignore', 'raise', 'coerce'}, default 'raise'
+                - If 'raise', then invalid parsing will raise an exception.
+                - If 'coerce', then invalid parsing will be set as NaT.
+                - If 'ignore', then invalid parsing will return the input.
+            include_index: If true, also convert index columns to timedelta.
+
+        Returns:
+            A new query compiler with the data converted to timedelta.
+        """
+        if errors != "raise":
```
sfc-gh-nkrishna, line 6256: This logic is pretty similar to `astype`, can we just use that in this case or will that implementation not work for this case?

sfc-gh-azhan: IIRC, astype always raise right?

sfc-gh-azhan: But yes, let's try to reuse code as much as possible

sfc-gh-nkumar: to_timedelta is more generic as compared to astype. `to_timedelta` also takes `unit` as input and multiply accordingly to convert in nanoseconds. Not sure if there it's possible to reuse here.

sfc-gh-azhan (approved):

sfc-gh-nkrishna (approved):
LGTM!

PR #2186, opened 2024-08-28
[SNOW-1458134]: Add support for monotonic checks for Series and Index

sfc-gh-rdurrani:

   

   Fixes SNOW-1458134

Add support for Index and Series is_monotonic_increasing and is_monotonic_decreasing


docs/source/modin/supported/series_supported.rst
```diff
@@ -43,9 +43,9 @@ Attributes
 +-----------------------------+---------------------------------+----------------------------------------------------+
 | ``index``                   | Y                               |                                                    |
 +-----------------------------+---------------------------------+----------------------------------------------------+
-| ``is_monotonic_decreasing`` | N                               |                                                    |
+| ``is_monotonic_decreasing`` | Y                               |                                                    |
```
sfc-gh-vbudati, line 46: Add to `series.rst` as well

src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -3659,6 +3659,38 @@ def hasnans():
         Return True if there are any NaNs.
         """
 
+    @property
+    def is_monotonic_increasing():
+        """
```
sfc-gh-vbudati, line 3664: add "Returns" section in both docstrings

sfc-gh-vbudati: you also have the names for both switched

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -398,7 +398,6 @@ def values(self) -> ArrayLike:
         return self.to_pandas().values
 
     @property
-    @index_not_implemented()
     def is_monotonic_increasing(self) -> None:
```
sfc-gh-vbudati, line 401: change return type and add "Returns" section in docstring

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -411,11 +410,23 @@ def is_monotonic_increasing(self) -> None:
...
+        >>> pd.Index([1, 2, 3]).is_monotonic_increasing
+        True
+        >>> pd.Index([1, 2, 2]).is_monotonic_increasing
+        True
+        >>> pd.Index([1, 3, 2]).is_monotonic_increasing
+        False
         """
-        # TODO: SNOW-1458134 implement is_monotonic_increasing
+        return (
+            self._query_compiler.is_monotonic_increasing(index=True)
```
sfc-gh-vbudati, line 424: I think another thing you could have done is just implement `is_monotonic_increasing/decreasing` for Series and done
`return self.to_series().is_monotonic_increasing()`

src/snowflake/snowpark/modin/plugin/extensions/series_overrides.py
```diff
@@ -164,3 +164,29 @@ def plot(
 @series_not_implemented()
 def transform(self, func, axis=0, *args, **kwargs):  # noqa: PR01, RT01, D200
     pass  # pragma: no cover
+
+
+@register_series_accessor("is_monotonic_increasing")
+@property
+@snowpark_pandas_telemetry_method_decorator
+def is_monotonic_increasing(self):  # noqa: RT01, D200
```
sfc-gh-vbudati, line 172: Why implement the methods here instead of in `series.py`?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2272,9 +2272,105 @@ def reindex(
...
             return self._reindex_axis_1(labels=labels, **kwargs)
 
+    def is_monotonic_decreasing(self, index: bool) -> "SnowflakeQueryCompiler":
+        """
+        Returns a QueryCompiler containing only a column that checks for monotonically
+        decreasing values.
+
+        Parameters
+        ----------
+        index: bool
```
sfc-gh-vbudati, line 2282: If you implement the index version of this method as `index.to_series().is_monotonic_increasing/decreasing()` you wouldn't need the `index` parameter here and can shorten the logic here too. Most index methods convert the index to series and then call the series version of the method.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2272,9 +2272,105 @@ def reindex(
...
+                    monotonic_decreasing_snowflake_quoted_identifier
+                ],
+                index_column_pandas_labels=new_modin_frame.index_column_pandas_labels,
+                index_column_snowflake_quoted_identifiers=new_modin_frame.index_column_snowflake_quoted_identifiers,
+                data_column_types=None,
+                index_column_types=None,
+            )
+        )
+
+    def is_monotonic_increasing(self, index: bool) -> "SnowflakeQueryCompiler":
```
sfc-gh-vbudati, line 2323: A lot of the code is shared between `is_monotonic_increasing` and `is_monotonic_decreasing` here - you combine it into one method and decide whether it's increasing or decreasing through a param passed to it

sfc-gh-vbudati (commented):
Left some comments, the logic looks great!

docs/source/modin/supported/series_supported.rst
sfc-gh-joshi: and `index.rst`

src/snowflake/snowpark/modin/plugin/extensions/series_overrides.py
sfc-gh-joshi: +1, the implementation seems to match the upstream modin version so it should be fine to leave in there.

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -411,11 +410,23 @@ def is_monotonic_increasing(self) -> None:
...
+            .to_pandas()
```
sfc-gh-azhan, line 425: looks wrong to me: pull the whole data out alert!

sfc-gh-azhan (changes requested):

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-joshi: The result here is a scalar boolean, so I think doing this is fine (the `_reduce_dimension` call done by the series version is basically the same thing).

src/snowflake/snowpark/modin/plugin/docstrings/series.py
sfc-gh-rdurrani: Good catch, thank you!

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-rdurrani: Will change it to that!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: Done!

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-rdurrani: +1 to @sfc-gh-joshi - the output of this method is a single bool, so we need to materialize the data. I also have a small hack in the QC to make sure we're not materializing too much data - I call limit on the ordered_dataframe (since we only need one of the values, since every value in the row is the result), so we're only materializing a single ton (i.e. an object with one row and one column).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2273,9 +2273,82 @@ def reindex(
...
+        if increasing:
+            data_column_snowflake_quoted_identifiers.append(
+                monotonic_increasing_snowflake_quoted_identifier
+            )
+        else:
+            data_column_snowflake_quoted_identifiers.append(
+                monotonic_decreasing_snowflake_quoted_identifier
+            )
+        new_modin_frame = new_qc._modin_frame
+        return SnowflakeQueryCompiler(
```
sfc-gh-azhan, line 2334: why not just call ordered dataframe collect to get the bool here?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2273,9 +2273,82 @@ def reindex(
         else:
             return self._reindex_axis_1(labels=labels, **kwargs)
 
+    def is_monotonic_decreasing(self) -> "SnowflakeQueryCompiler":
```
sfc-gh-azhan, line 2276: why not directly return bool?

sfc-gh-vbudati (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: The upstream modin frontend expects a query compiler object. We could override this, but I think for now it's best to keep as aligned with upstream as possible and revisit it in OSS modin.

sfc-gh-azhan: This sounds like a bug to me. We should override this IMO.

sfc-gh-rdurrani: I'm personally not certain that we *should* override this, since we want the QC to be returning QCs from methods in general, but if we do decide to override this, I think it should be done on upstream modin first, as otherwise, we'd have to move this method to `series_override.py`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-rdurrani: Continued here: https://github.com/snowflakedb/snowpark-python/pull/2186#discussion_r1736668546

sfc-gh-joshi (approved):

sfc-gh-azhan (approved):
Discussed offline and agree to move forward with current solution.

PR #2193, opened 2024-08-29
SNOW-1637945: Add support for TimedeltaIndex attributes

sfc-gh-nkumar:
Fixes SNOW-1637945

Add support for TimedeltaIndex attributes `days`, `seconds`, `microseconds`, and `nanoseconds`.


sfc-gh-helmeleegy (approved):
LGTM.

sfc-gh-azhan (approved):

tests/integ/modin/index/test_timedelta_index_methods.py
```diff
@@ -54,12 +55,22 @@ def test_non_default_args(kwargs):
...
     msg = f"Snowpark pandas does not yet support the property TimedeltaIndex.{property}"
     with pytest.raises(NotImplementedError, match=msg):
         getattr(snow_index, property)
+
+
+@pytest.mark.parametrize("attr", ["days", "seconds", "microseconds", "nanoseconds"])
+@sql_count_checker(query_count=1)
+def test_timedelta_index_properties(attr):
+    native_index = native_pd.TimedeltaIndex(
+        ["1d", "1h", "60s", "1s", "800ms", "5us", "6ns", "1d 3s", "9m 15s 8us"]
```
sfc-gh-helmeleegy, line 71: Can we add a null here?

sfc-gh-helmeleegy (changes requested):
Looks good overall. I just think we need to test against null data too.

tests/integ/modin/index/test_timedelta_index_methods.py
sfc-gh-nkumar: done

sfc-gh-azhan (changes requested):
Sorry to block this now.
@sfc-gh-mvashishtha and I had a chat today about this. Ideally we should reuse the dt.properties code to do this, we should be able to support this too:
`pd.Series([1,2,3]).astype("timedelta64[ns]").dt.seconds`
cc @sfc-gh-helmeleegy

sfc-gh-nkumar:
> Sorry to block this now. @sfc-gh-mvashishtha and I had a chat today about this. Ideally we should reuse the dt.properties code to do this, we should be able to support this too: `pd.Series([1,2,3]).astype("timedelta64[ns]").dt.seconds` cc @sfc-gh-helmeleegy

yeah, that is the plan. We will also use this same method to implement Series.dt.* attributes relevant for timedelta column.

sfc-gh-azhan (approved):

sfc-gh-helmeleegy:
> > Sorry to block this now. @sfc-gh-mvashishtha and I had a chat today about this. Ideally we should reuse the dt.properties code to do this, we should be able to support this too: `pd.Series([1,2,3]).astype("timedelta64[ns]").dt.seconds` cc @sfc-gh-helmeleegy
> 
> yeah, that is the plan. We will also use this same method to implement Series.dt.* attributes relevant for timedelta column. There are two sets of properties. One applicable to datetime columns and one applicable to timedelta columns. dt_property implements all properties applicable to datetime column and timedelta_property implement all properteis applicable to timedelta column. In both the scenarios code is reused between Series.dt* and [Datetime/Timedelta]Index.* attributes.

That's right. This is the plan.

sfc-gh-helmeleegy (approved):

PR #2199, opened 2024-08-30
[SNOW-1514712] Support UDFs for Snowpark IR

sfc-gh-lspiegelberg:
Part I of supporting UDFs. This supports `udf` decorator.

Other:
- Fixes in `udf_utils.py` the case when in `get_opt_arg_defaults` the returned python `arg_spec = inspect.getfullargspec(target_func)` is `None`. (Yes, this actually happens).
- Fixes in `conftest.py` for the AST tests the initialization of the session. Previous init using constructors is incorrect, need to use builder pattern as session is not properly registered for `_get_active_session` else.
- Changes `call_udf` to more correctly become calling a builtin function.


src/snowflake/snowpark/_internal/ast.py
```diff
@@ -161,3 +171,14 @@ def _init_batch(self):
         self._request.client_language.python_language.version.label = releaselevel
 
         self._request.client_ast_version = CLIENT_AST_VERSION
+
+    def register_callable(self, func: Callable) -> int:
+        """tracks client-side an actual callable and returns an id."""
```
sfc-gh-azwiegincew, line 176: t => T
an id => its ID

tests/ast/data/col_udf.test
```diff
@@ -0,0 +1,50 @@
...
+
+# call_udf should be encoded as builtin applexpr. It only works for named (& permanently) registered udfs.
+
+add_two = udf(lambda x: x + 2, return_type=IntegerType(), input_types=[IntegerType()], name="add_two", replace=True)
+
+df.select(call_udf("add_two", col("A")).as_("a_Ans")).collect()
+
+param_udf = udf(lambda x, y: str(y + x**2), return_type=VariantType(), input_types=[IntegerType(), FloatType()],
+    name="param_udf", is_permanent=True, stage_location="@", imports=["numpy"], packages=["bla"],
+     replace=True, if_not_exists=False, parallel=8, max_batch_size=2, source_code_display=False,
```
sfc-gh-azwiegincew, line 21: bad indentation

PR #2202, opened 2024-08-30
SNOW-1640548, SNOW-1640549, SNOW-1640554, SNOW-1640562: Add support for Series.dt.days/seconds/microseconds/nanoseconds

sfc-gh-helmeleegy:

   

   Fixes SNOW-1640548, SNOW-1640549, SNOW-1640554, SNOW-1640562

  Add support for Series.dt.days/seconds/microseconds/nanoseconds.


src/snowflake/snowpark/modin/pandas/series.py
```diff
@@ -2420,7 +2424,9 @@ def dt(self):  # noqa: RT01, D200
         """
         # TODO: SNOW-1063347: Modin upgrade - modin.pandas.Series functions
         current_dtype = self.dtype
-        if not is_datetime64_any_dtype(current_dtype):
+        if not is_datetime64_any_dtype(current_dtype) and not is_timedelta64_dtype(
```
sfc-gh-nkumar, line 2427: We should also raise an error if user tries to access timedelta property on datetime column and vice-a-versa.

sfc-gh-nkumar (approved):
Looks good! One comment to improve error messages when accessing dt properties on incorrect type.

src/snowflake/snowpark/modin/pandas/series.py
sfc-gh-joshi: Should we be doing this type validation here, or just letting the query compiler or SQL engine take care of it?

sfc-gh-nkumar: Generally it's fine to let SQL engine take care of error handling. But timedelta type is special case as it only exists on client so it makes sense to raise type errors in client layer. modin frontend end vs query compiler, right now we are doing such checks in both places, so either works.

sfc-gh-helmeleegy: Good catch. Done.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10842,16 +10842,16 @@ def dt_freq(self) -> "SnowflakeQueryCompiler":
         return self.dt_property("freq")
 
     def dt_seconds(self) -> "SnowflakeQueryCompiler":
-        return self.dt_property("seconds")
+        return self.timedelta_property("seconds")
```
sfc-gh-azhan, line 10845: Why change this? why we call timedelta_property for dt_property now? What about the existing code in dt_property?

sfc-gh-helmeleegy: Because initially all properties were calling `dt_property` such that if they were not implemented, a NotImplementedError is raised. This was the case for `seconds` for example before this PR. Now we are distinguishing between datetime properties from timedelta properties, where the former is handled by `timedelta_property` and the latter is handled by `dt_property`.

sfc-gh-helmeleegy: Because initially all properties were calling `dt_property` such that if they were not implemented, a NotImplementedError is raised. This was the case for `seconds` for example before this PR. Now we are distinguishing between datetime properties from timedelta properties, where the former is handled by `timedelta_property` and the latter is handled by `dt_property`.

sfc-gh-azhan: But `dt_property` support `second`, I don't suggest but you could use similar way to support `seconds`. 
I'm a little worry about the code management here. For example `day` go to `dt_property` but `days` go to `timedelta_properties` but esentially they should be the same thing. 
CC @sfc-gh-nkumar Any idea how we can improve this?

sfc-gh-azhan (approved):

PR #2205, opened 2024-08-30
SNOW-1063347: Remove series.py

sfc-gh-joshi:

   

   Fixes SNOW-1063347

This PR removes series.py (following up #2167). As before, any preserved overrides are given a reason in code comments. The following methods have been added to `series_overrides.py`:
- __init__
- sub/subtract, and most other binary operators
- aggregate/agg
- apply
- map
- argmax/argmin
- case_when
- dropna
- duplicated
- sum/std/var
- cat/str/dt
- name
- empty
- fillna
- groupby
- info
- _qcut
- replace
- reset_index
- set_axis
- rename
- sort_values
- unstack
- where
- value_counts
- shift
- squeeze
- _prepare_inter_op
- _to_datetime
- to_list
- _create_or_update_from_compiler
- to_frame
- _to_pandas
- to_numpy
- __setitem__
- __repr__

Because of differences in `__getattr__` logic, we need to add `dt`, `str`, and `columns` to `_ATTRS_NO_LOOKUP` in this PR to prevent additional queries from being incurred. This will create slightly different behavior for attempting to access a named index element directly:
```
>>> s = pd.Series(["a", "b"], index=["str", "columns"])
>>> s.columns
AttributeError: 'Series' object has no attribute 'columns'
>>> s.to_pandas().columns  # DIFFERENT: treats as index access
'b'
>>> s.str
<modin.pandas.series_utils.StringMethods object at 0x135f57940>
>>> s.to_pandas().str  # SAME: gets accessor
<pandas.core.strings.accessor.StringMethods object at 0x135f93b50>
```


src/snowflake/snowpark/modin/pandas/__init__.py
```diff
@@ -165,13 +167,20 @@
 )
 
 # this must occur before overrides are applied
-_attrs_defined_on_modin_base = set(dir(modin.pandas.base.BasePandasDataset))
-_attrs_defined_on_series = set(
-    dir(Series)
-)  # TODO: SNOW-1063347 revisit when series.py is removed
-_attrs_defined_on_dataframe = set(
-    dir(DataFrame)
+_attrs_defined_on_modin_series = set()
```
sfc-gh-azhan, line 170: can we add some description before those code to clarify why we need those attrs related variables?

sfc-gh-joshi: added

sfc-gh-azhan (approved):

tests/unit/modin/test_unsupported.py
```diff
@@ -144,8 +144,6 @@ def test_unsupported_df(df_method, kwargs):
         ["prod", {}],
         ["ravel", {}],
         ["reorder_levels", {"order": ""}],
-        ["repeat", {"repeats": ""}],
```
sfc-gh-yzou, line 147: why those methods start becoming supported now? how are they supported now?

sfc-gh-yzou: if that is not a snowpark pandas claimed supported API, i thing we should overwrite in the extension to raise an unsupported error

tests/integ/modin/test_to_numpy.py
```diff
@@ -58,8 +58,9 @@ def test_to_numpy_basic(data, pandas_obj, func):
             assert_array_equal(df.values, native_df.values)
     if pandas_obj == "Series":
         with SqlCounter(query_count=1):
-            res = df.to_list()
-        expected_res = native_df.to_list()
+            res = df.to_numpy()
+        expected_res = native_df.to_numpy()
```
sfc-gh-yzou, line 62: even thought this test is wired that we have to_list inside to_numpy test, do we have test for to_list? that seems a supported API

tests/integ/modin/series/test_take.py
```diff
@@ -33,3 +34,10 @@ def test_series_take():
     expected = pd.Series([6], index=[2])
     with SqlCounter(query_count=2, join_count=2):
         assert_series_equal(actual, expected)
+
+
+@sql_count_checker(query_count=0)
+def test_series_take_axis_1():
```
sfc-gh-yzou, line 40: is that a new behavior or just missed tests?

tests/integ/modin/series/test_getattr.py
```diff
@@ -16,7 +16,8 @@
 @pytest.mark.parametrize(
     "name, expected_query_count, expected_join_count",
     [
-        ("a", 2, 2),
+        # Upstream Modin performs a check against self.index["a"] rather than self["a"], incurring an additional query.
+        ("a", 3, 2),
```
sfc-gh-yzou, line 20: that is counted as regression, can we overwrite the corresponding implementation

src/snowflake/snowpark/modin/plugin/__init__.py
```diff
@@ -56,6 +56,26 @@
 
 DocModule.put(docstrings.__name__)
 
+# Configure Modin engine so it detects our Snowflake I/O classes.
+# This is necessary to run even basic code like Series/DataFrame constructors,
+# as these factories define the `from_pandas` method for each Modin backend.
+
+from modin.config import Engine  # isort: skip  # noqa: E402
```
sfc-gh-yzou, line 63: i am little bit confused here why we suddenly start requiring set this up now? but not when we are removing the other files?

src/snowflake/snowpark/modin/plugin/docstrings/base.py
```diff
@@ -1491,7 +1491,7 @@ def iloc():
         With a scalar integer.
 
         >>> type(df.iloc[0])
-        <class 'snowflake.snowpark.modin.pandas.series.Series'>
+        <class 'modin.pandas.series.Series'>
```
sfc-gh-yzou, line 1494: i felt little bit wired to have test shown the whole class path in the example, is that the only place we have this? we can fix it like this for now, but maybe later we can have a separate pr to remove those in all examples

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12352,6 +12382,18 @@ def _quantiles_single_col(
 
         return SnowflakeQueryCompiler(internal_frame)
 
+    def mode(
+        self, axis: Axis = 0, numeric_only: bool = False, dropna: bool = True
```
sfc-gh-yzou, line 12386: can those method not implemented error be at the overwritten part? instead of query compiler like how other apis works today

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -757,6 +758,24 @@ def copy(self) -> "SnowflakeQueryCompiler":
         qc.snowpark_pandas_api_calls = self.snowpark_pandas_api_calls.copy()
         return qc
 
+    def to_list(self) -> list:
+        """
+        Return a native Python list of the values.
+
+        Only called if the frontend object was a Series.
+        """
+        return self.to_pandas().squeeze().to_list()
```
sfc-gh-yzou, line 767: this seems a wired implementation here, which seems tailored specifically for series. what is the modin implementation ? can we use the overwrite for this for now, and see how can we migrate over later?

src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -3277,14 +3277,6 @@ def take():
         3    2
         dtype: int64
 
-
-        For `Series` axis parameter is unused and defaults to 0.
-
-        >>> ser.take([0, 3], axis=1)
```
sfc-gh-yzou, line 3283: why this is removed now?

src/snowflake/snowpark/modin/pandas/indexing.py
```diff
@@ -1311,12 +1311,16 @@ def __setitem__(
...
         new_qc = self.qc.set_2d_positional(
-            row_loc._query_compiler
-            if isinstance(row_loc, BasePandasDataset)
-            else row_loc,
-            col_loc._query_compiler
-            if isinstance(col_loc, BasePandasDataset)
-            else col_loc,
+            (
+                row_loc._query_compiler
+                if isinstance(row_loc, BasePandasDataset)
```
sfc-gh-yzou, line 1316: what is the change we made here? is it just the ()?

tests/unit/modin/test_unsupported.py
sfc-gh-joshi: They are not newly supported, but they now fail at the query compiler layer rather than the frontend, and therefore need properties that are difficult to define on a mocked `SnowflakeQueryCompiler` object. They're now being checked in `tests/integ/modin/test_unimplemented.py`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: DataFrames don't have a `to_list` method. Upstream Modin does something similar:

```python
    def to_list(self):
        """
        Return a list of the values.

        These are each a scalar type, which is a Python scalar (for str, int, float) or a pandas scalar (for Timestamp/Timedelta/Interval/Period).

        Returns
        -------
        list
        """
        return SeriesDefault.register(pandas.Series.to_list)(self)
```

src/snowflake/snowpark/modin/plugin/__init__.py
sfc-gh-joshi: This was left over from before we decided to override `__init__`, since there were some method calls that needed to find the `from_pandas` method for the Snowflake backend. I'll remove this setup code since we're overriding `__init__` anyway.

src/snowflake/snowpark/modin/pandas/indexing.py
sfc-gh-joshi: Yes, it's just a parentheses (it was a linter artifact from when I used `pd.base.BasePandasDataset` instead of importing the name first). I've reverted this.

tests/integ/modin/test_to_numpy.py
sfc-gh-yzou: Then i don't think you need to change this test now, can you revert the changes here for now, add a ticket to refactor that test and add some tests for to_list

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: I actualy see an override of to_list in the series_override methods, which seems kind of saying this method is not called. Since this is just a refactoring pr, i would suggest let's just use override for now, and handle those implementation change in a separate pr

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-yzou: API not implemented errors today are all thrown at frontend api layer, so let's keep it the same everywhere, if we decide to do those not implemented error at query_compiler layer, let's do it the same for all API also

src/snowflake/snowpark/modin/plugin/docstrings/series.py
sfc-gh-yzou: I see, do you know if it is a new behavior introduced in 2.2.* compare with 1.5.3

sfc-gh-joshi: Yes, seems like it works in pandas 2.0.3 and fails in pandas 2.1.4.

tests/integ/modin/test_to_numpy.py
sfc-gh-joshi: reverted and filed SNOW-1652384

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: (discussed on slack) moved back to frontend overrides

src/snowflake/snowpark/modin/plugin/docstrings/series.py
sfc-gh-yzou: Can you add a entry in the change log to indicate this difference?

sfc-gh-yzou (approved):
@sfc-gh-joshi Thanks for taking care of the comment! Can you add one description in the changelog to indicate the change for take?

src/snowflake/snowpark/modin/plugin/docstrings/series.py
sfc-gh-joshi: done

PR #2213, opened 2024-09-03
[SNOW-1649172]: Fix `loc` set when setting DataFrame row with Series value

sfc-gh-rdurrani:

   

   Fixes SNOW-1649172

When doing `df.loc[x] = series`, an error occurs because series does not have the same number of columns as the dataframe being set. Instead, the Series should be transposed and set, regardless of whether it has an equal number of rows as the dataframe has columns.


src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -2215,6 +2218,36 @@ def set_frame_2d_labels(
     index_is_frame = isinstance(index, InternalFrame)
     item_is_frame = isinstance(item, InternalFrame)
     item_is_scalar = is_scalar(item)
+    original_index = index
+    # If `item` is from a Series (rather than a Dataframe), flip the series item values to apply them
```
sfc-gh-helmeleegy, line 2222: Do we need to update the examples in the docstring?

sfc-gh-vbudati: No, we don't need to. I added an example though

sfc-gh-helmeleegy: Sounds good. I'm not sure I can find it though.

src/snowflake/snowpark/modin/plugin/docstrings/base.py
```diff
@@ -1832,6 +1832,15 @@ def loc():
         viper               0       0
         sidewinder          0       0
 
+        Setting the values with a Series item.
```
sfc-gh-vbudati, line 1835: @sfc-gh-helmeleegy this is the example I added

tests/integ/modin/frame/test_iloc.py
```diff
@@ -3205,3 +3205,68 @@ def test_raise_set_cell_with_list_like_value_error():
...
+
+    eval_snowpark_pandas_result(
+        snow_df,
+        native_df,
+        ilocset,
+    )
+
+
+@sql_count_checker(query_count=1, join_count=3)
+def test_df_iloc_full_set_row_from_series_int_and_string_indexes():
```
sfc-gh-azhan, line 3255: you can combine this one into the previous one.

tests/integ/modin/frame/test_loc.py
```diff
@@ -4068,6 +4068,74 @@ def test_df_loc_get_with_timedelta_and_none_key():
...
+    def locset(df):
+        series = (
+            pd.Series([1, 4, 9], index=index)
+            if isinstance(df, pd.DataFrame)
+            else native_pd.Series([1, 4, 9], index=index)
+        )
+        if isinstance(df, pd.DataFrame):
+            df.loc[:] = series
+        else:
+            if index == [0, 1, 2]:
```
sfc-gh-azhan, line 4111: you'd better just compare the result with the expected result. These steps here can be confusing.

src/snowflake/snowpark/modin/pandas/indexing.py
```diff
@@ -1039,6 +1039,9 @@ def __setitem__(
         )
         if item_is_2d_array:
             item = pd.DataFrame(item)
+        frame_is_df_and_item_is_series = isinstance(item, pd.Series) and isinstance(
```
sfc-gh-azhan, line 1042: what happens for item is a Index or list? Are they matching with pandas? Can you verify too?

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -2213,6 +2216,36 @@ def set_frame_2d_labels(
...
+    # across columns rather than rows.
+    if frame_is_df_and_item_is_series and (columns == slice(None) or len(columns) > 1):  # type: ignore[arg-type]
```
sfc-gh-azhan, line 2222: can you wrap it into a function and use function name to brief what this method does?

sfc-gh-azhan: what does this mean `(columns == slice(None) or len(columns) > 1)`?

sfc-gh-azhan: this `type: ignore[arg-type]` actually indicate something is wrong. You didn't consider all type cases.

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -2213,6 +2216,36 @@ def set_frame_2d_labels(
...
+        col_len = (
+            len(internal_frame.data_column_snowflake_quoted_identifiers)
+            if columns == slice(None)
+            else len(columns)  # type: ignore[arg-type]
+        )
+        item = get_item_series_as_single_row_frame(
+            item, col_len, move_index_to_cols=True
+        )
+
+        if is_scalar(index):
```
sfc-gh-azhan, line 2234: what happens if index is not scalar?

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
sfc-gh-azhan, line 2222: This should be done in `_set_2d_labels_helper_for_frame_item`

sfc-gh-azhan (changes requested):
Please describe what was the issue?

src/snowflake/snowpark/modin/pandas/indexing.py
sfc-gh-rdurrani: Added tests!

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
sfc-gh-rdurrani: I believe this is checking to see if more than one column is being set. As for the arg-type, I think that is because its ignoring if the columns is a SnowflakeQueryCompiler? I've added a test for that case, and will fix it!

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
sfc-gh-rdurrani: If index is not scalar, we don't have to append it to the item to match index - it should either be slice(None) or an internalframe, which we handle in the rest of the method.

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
sfc-gh-rdurrani: I actually think it needs to be done in this method, since we need to modify `item` before the map is created (which is passed into `_set_2d_labels_helper_for_frame_item`, and we need the modified item later on in this method.

sfc-gh-rdurrani: I can move this into the conditional for `if item_is_frame` though!

sfc-gh-rdurrani:
Will add additional tests once the match by position or labels issue is resolved: https://snowflake.slack.com/archives/C04HF38JFAQ/p1727828020400139?thread_ts=1727824503.275869&cid=C04HF38JFAQ

src/snowflake/snowpark/modin/plugin/docstrings/base.py
sfc-gh-helmeleegy: Thanks!

CHANGELOG.md
```diff
@@ -56,6 +56,7 @@
 - Fixed a bug where transposing a dataframe would map `Timedelta` index levels to integer column levels.
 - Fixed a bug where `Resampler` methods on timedelta columns would produce integer results.
 - Fixed a bug where `pd.to_numeric()` would leave `Timedelta` inputs as `Timedelta` instead of converting them to integers.
+- Fixed `loc` set when setting row of DataFrame with Series value.
```
sfc-gh-azhan, line 59: single row or rows?

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -2068,6 +2072,76 @@ def set_frame_2d_labels(
     # map from item's data column label to its position in the joined frame or item_column_values
     item_data_col_label_to_pos_map: dict[Hashable, int] = {}
     if item_is_frame:
+        from snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler import (
```
sfc-gh-azhan, line 2075: why not move them into `_set_2d_labels_helper_for_frame_item` since item is a frame.

CHANGELOG.md
sfc-gh-rdurrani: both - will update.

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
sfc-gh-rdurrani: This codepath also modifies index, which is used later on in the post-processing steps of the current function, so it makes sense to perform the computation here imo.

sfc-gh-azhan: can you wrap them into a helper function and we can clearly see what's the input and output are?

sfc-gh-rdurrani: sure!

sfc-gh-helmeleegy (approved):
LGTM. Thanks for addressing the comments.

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -1954,6 +1955,82 @@ def _set_2d_labels_helper_for_single_column_wise_item(
     ).result_frame
 
 
+def _convert_series_item_to_row_for_set_frame_2d_labels(
```
sfc-gh-azhan, line 1958: can you move the index operation to another helper function? Because the name of this function didn't say anything about changing the index.

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -1954,6 +1955,82 @@ def _set_2d_labels_helper_for_single_column_wise_item(
...
+
+    def slice_len(slice_obj: slice) -> int:
+        """Helper method to calculate length of slice object for columns."""
+        start = internal_frame.data_column_pandas_labels.index(slice_obj.start) or 0
+        end = internal_frame.data_column_pandas_labels.index(slice_obj.stop) or len(
+            internal_frame.data_column_snowflake_quoted_identifiers
+        )
+        return end - start
+
+    if columns == slice(None):
```
sfc-gh-azhan, line 1996: this can be a helper function like `get_column_length` in indexing_util.py.

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -1954,6 +1955,82 @@ def _set_2d_labels_helper_for_single_column_wise_item(
...
+        col_len = len(internal_frame.data_column_snowflake_quoted_identifiers)
+    elif isinstance(columns, slice):
+        col_len = slice_len(columns)
+    elif isinstance(columns, Sized):
+        col_len = len(columns)
+    else:
+        col_len = len(columns.index)
+
+    if isinstance(columns, SnowflakeQueryCompiler):
```
sfc-gh-azhan, line 2005: can you add comments about what you are trying to do here and also the next line?

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -1954,6 +1955,82 @@ def _set_2d_labels_helper_for_single_column_wise_item(
...
+    item = (
+        SnowflakeQueryCompiler(
+            get_item_series_as_single_row_frame(item, col_len, move_index_to_cols=True)
+        )
+        .reset_index(drop=True)
+        ._modin_frame
+    )
+
+    if is_scalar(index):
+        new_item = item.append_column("__index__", pandas_lit(index))
```
sfc-gh-azhan, line 2017: you should use some api in SQC to set index or reindex. Manually set the column can lead to potential bugs. Once you got new_item_sqc then you can set item = new_item_sqc._modin_frame.

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -2068,6 +2148,25 @@ def set_frame_2d_labels(
...
     if item_is_frame:
+        from snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler import (
+            SnowflakeQueryCompiler,
+        )
+
+        # If `item` is from a Series (rather than a Dataframe), flip the series item values to apply them
+        # across columns rather than rows.
+        is_multi_col_set = (
+            (isinstance(columns, Sized) and len(columns) > 1)
+            or isinstance(columns, slice)
```
sfc-gh-azhan, line 2159: slice and qc case can be single column right?

sfc-gh-rdurrani: Good catch, thank you!

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -2068,6 +2201,30 @@ def set_frame_2d_labels(
     # map from item's data column label to its position in the joined frame or item_column_values
     item_data_col_label_to_pos_map: dict[Hashable, int] = {}
     if item_is_frame:
+        from snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler import (
+            SnowflakeQueryCompiler,
+        )
+
+        # If `item` is from a Series (rather than a Dataframe), flip the series item values to apply them
+        # across columns rather than rows.
+        is_multi_col_set = (
```
sfc-gh-azhan, line 2210: can you use `col_info` those columns have been preprocessed there for the rest code to use. You don't need to worry about different types once you use `col_info`.

sfc-gh-rdurrani: Good catch, makes sense!

sfc-gh-azhan: Can you update it?

sfc-gh-rdurrani: Oh you meant to change the length check here - I can't do that since we want to follow this codepath even if the column key is a slice of length one or a Series of length one.

sfc-gh-azhan: Yes, I mean in all your new logic added in this PR. You should be able to just rely on `col_info` instead of `columns`.

sfc-gh-rdurrani: Not necessarily - some of the behavior depends on the specific type of the columns (e.g. if columns is a list or a snowpark pandas Series). Here, we want to specifically check that columns is either a list of more than one element, or a slice, or a SnowflakeQueryCompiler - not that it has multiiple columns, in which case, df.loc[:, pd.Series(["A"])] wouldn't be caught in this case, even though we would want it to.

sfc-gh-azhan (approved):

PR #2214, opened 2024-09-03
SNOW-1641644: Drop temp table directly at garbage collection instead of using multi-threading 

sfc-gh-jdu:

   

   Fixes SNOW-1641644

  see details in [discussion](https://snowflake.slack.com/archives/C04HF38JFAQ/p1725057972543059?thread_ts=1724782971.351959&cid=C04HF38JFAQ)

### DIFF
--- a/src/snowflake/snowpark/_internal/telemetry.py
+++ b/src/snowflake/snowpark/_internal/telemetry.py
@@ -79,6 +79,20 @@ class TelemetryField(Enum):
     QUERY_PLAN_HEIGHT = "query_plan_height"
     QUERY_PLAN_NUM_DUPLICATE_NODES = "query_plan_num_duplicate_nodes"
     QUERY_PLAN_COMPLEXITY = "query_plan_complexity"
+    # temp table cleanup
+    TYPE_TEMP_TABLE_CLEANUP = "snowpark_temp_table_cleanup"
+    NUM_TEMP_TABLES_CLEANED = "num_temp_tables_cleaned"
+    NUM_TEMP_TABLES_CREATED = "num_temp_tables_created"
+    TEMP_TABLE_CLEANER_ENABLED = "temp_table_cleaner_enabled"
+    TYPE_TEMP_TABLE_CLEANUP_ABNORMAL_EXCEPTION = (
+        "snowpark_temp_table_cleanup_abnormal_exception"
+    )
+    TEMP_TABLE_CLEANUP_ABNORMAL_EXCEPTION_TABLE_NAME = (
+        "temp_table_cleanup_abnormal_exception_table_name"
+    )
+    TEMP_TABLE_CLEANUP_ABNORMAL_EXCEPTION_MESSAGE = (
+        "temp_table_cleanup_abnormal_exception_message"
+    )
 
 
 # These DataFrame APIs call other DataFrame APIs
@@ -461,6 +475,44 @@ class TelemetryClient:
             TelemetryField.KEY_DATA.value: {
                 TelemetryField.SESSION_ID.value: session_id,
                 CompilationStageTelemetryField.KEY_REASON.value: reason,
+            },
+        }
+        self.send(message)
+
+    def send_temp_table_cleanup_telemetry(
+        self,
+        session_id: str,
+        temp_table_cleaner_enabled: bool,
+        num_temp_tables_cleaned: int,
+        num_temp_tables_created: int,
+    ) -> None:
+        message = {
+            **self._create_basic_telemetry_data(
+                TelemetryField.TYPE_TEMP_TABLE_CLEANUP.value
+            ),
+            TelemetryField.KEY_DATA.value: {
+                TelemetryField.SESSION_ID.value: session_id,
+                TelemetryField.TEMP_TABLE_CLEANER_ENABLED.value: temp_table_cleaner_enabled,
+                TelemetryField.NUM_TEMP_TABLES_CLEANED.value: num_temp_tables_cleaned,
+                TelemetryField.NUM_TEMP_TABLES_CREATED.value: num_temp_tables_created,
+            },
+        }
+        self.send(message)
+
+    def send_temp_table_cleanup_abnormal_exception_telemetry(
+        self,
+        session_id: str,
+        table_name: str,
+        exception_message: str,
+    ) -> None:
+        message = {
+            **self._create_basic_telemetry_data(
+                TelemetryField.TYPE_TEMP_TABLE_CLEANUP_ABNORMAL_EXCEPTION.value
+            ),
+            TelemetryField.KEY_DATA.value: {
+                TelemetryField.SESSION_ID.value: session_id,
+                TelemetryField.TEMP_TABLE_CLEANUP_ABNORMAL_EXCEPTION_TABLE_NAME.value: table_name,
+                TelemetryField.TEMP_TABLE_CLEANUP_ABNORMAL_EXCEPTION_MESSAGE.value: exception_message,
             },
         }
         self.send(message)
--- a/src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
+++ b/src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
@@ -4,9 +4,7 @@
 import logging
 import weakref
 from collections import defaultdict
-from queue import Empty, Queue
-from threading import Event, Thread
-from typing import TYPE_CHECKING, Dict, Optional
+from typing import TYPE_CHECKING, Dict
 
 from snowflake.snowpark._internal.analyzer.snowflake_plan_node import SnowflakeTable
 
@@ -33,12 +31,6 @@ class TempTableAutoCleaner:
         # to its reference count for later temp table management
         # this dict will still be maintained even if the cleaner is stopped (`stop()` is called)
         self.ref_count_map: Dict[str, int] = defaultdict(int)
-        # unused temp table will be put into the queue for cleanup
-        self.queue: Queue = Queue()
-        # thread for removing temp tables (running DROP TABLE sql)
-        self.cleanup_thread: Optional[Thread] = None
-        # An event managing a flag that indicates whether the cleaner is started
-        self.stop_event = Event()
 
     def add(self, table: SnowflakeTable) -> None:
         self.ref_count_map[table.name] += 1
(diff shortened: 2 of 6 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -643,11 +639,6 @@ def auto_clean_up_temp_table_enabled(self) -> bool:
         When setting this parameter to ``True``, Snowpark will automatically clean up temporary tables created by
         :meth:`DataFrame.cache_result` in the current session when the DataFrame is no longer referenced (i.e., gets garbage collected).
         The default value is ``False``.
-
-        Note:
```
sfc-gh-jdu, line 647: We can still provide this guarantee, by checking all entries of the count map whether a count reaches 0. But given we're not using watch thread, I don't think this guarantee makes much sense. We only need to do our best effort to clean up temp tables when this parameter is enabled. Alternatively, we can always easily add this guarantee later when the customer requests it.

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -54,53 +46,26 @@ def _delete_ref_count(self, name: str) -> None:
         self.ref_count_map[name] -= 1
         if self.ref_count_map[name] == 0:
             self.ref_count_map.pop(name)
```
sfc-gh-yzou, line 48: let's do not do pop here, if we keep the stop method, we can probably sent a teletemetry on close for the temp table that we didn't clean up (ones that whose reference that is 0, and ones that whose reference is not 0)

src/snowflake/snowpark/session.py
sfc-gh-yzou: This behavior is irrelevant with the threading work, it is a behavior of the cleaner. if we do not have this, what we are saying is we only clean up temp tables whose reference reach 0 after the cleaner is started. let's make sure this is documented clearly somewhere

tests/integ/test_temp_table_cleanup.py
```diff
@@ -139,7 +144,6 @@ def test_reference_count_map_multiple_sessions(db_parameters, session):
         assert session._temp_table_auto_cleaner.ref_count_map[table_name1] == 0
         assert new_session._temp_table_auto_cleaner.ref_count_map[table_name1] == 0
 
-        new_session._temp_table_auto_cleaner.start()
         del df2
         gc.collect()
         time.sleep(WAIT_TIME)
```
sfc-gh-yzou, line 149: this wait here seems going to be flaky, one thing you can do is keep track of all asycn_job in the auto_cleaner class, and here you can do a check to see if all job is finished. you can also have a flag to only record this for testing

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -54,53 +46,26 @@ def _delete_ref_count(self, name: str) -> None:
...
+            )._internal_collect_with_tag_no_telemetry(
+                block=False, statement_params={DROP_TABLE_STATEMENT_PARAM_NAME: name}
+            )
+            logging.debug(
+                f"Dropping {common_log_text} with query id {async_job.query_id}"
+            )
         except Exception as ex:
             logging.warning(
-                f"Cleanup Thread: Failed to drop {common_log_text}, exception: {ex}"
+                f"Failed to drop {common_log_text}, exception: {ex}"
```
sfc-gh-yzou, line 70: we can probably recored a telemetry later for failed table drops

tests/integ/test_temp_table_cleanup.py
sfc-gh-jdu: Yes, I will do it in a separate PR (tracked in https://snowflakecomputing.atlassian.net/browse/SNOW-1645523)

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: actually we might not need telemetry for this case, because we can group by temp table name from statement params, then we can find the corresponding query is successful or not.

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: I think we can have a telemetry here about 1) the number of temp tables cleaned up (value is 0) 2) the total number of temp table created by cache_result (length of the dict), when closing the session. So we can understand the percentage of temp table created from cache_result that are cleaned up

src/snowflake/snowpark/session.py
sfc-gh-jdu: We can say the temporary tables will only be dropped when this parameter is turned on during garbage collection, whereas the garbage collection in Python is triggered opportunistically and the timing is not guaranteed. I think it's also clear. Let me know wdyt.

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: I think actually it's a good point, and I added such telemetry and corresponding tests

src/snowflake/snowpark/session.py
```diff
@@ -645,9 +646,10 @@ def auto_clean_up_temp_table_enabled(self) -> bool:
         The default value is ``False``.
 
         Note:
-            Even if this parameter is ``False``, Snowpark still records temporary tables when
-            their corresponding DataFrame are garbage collected. Therefore, if you turn it on in the middle of your session or after turning it off,
-            the target temporary tables will still be cleaned up accordingly.
+            The temporary tables will only be dropped when this parameter is turned on during garbage collection. That is, if the
```
sfc-gh-helmeleegy, line 649: The wording of this note is very hard to read. Can we have another pass over it?

src/snowflake/snowpark/session.py
```diff
@@ -645,9 +646,10 @@ def auto_clean_up_temp_table_enabled(self) -> bool:
         The default value is ``False``.
```
sfc-gh-helmeleegy, line 646: The doctoring says:
> When setting this parameter to ``True``, Snowpark will automatically clean up temporary tables created by
        :meth:`DataFrame.cache_result` in the current session when the DataFrame is no longer referenced (i.e., gets garbage collected)
        
Do we make any distinction between calls to `cache_result` made by the user and those made by the system (e.g., during large query breakdown or repeated subquery elimination, or possibly as part of future optimizations)?

src/snowflake/snowpark/session.py
sfc-gh-jdu: yes fixed. Actually here I would like clarify that only if garbage collection when this parameter is on, table will be dropped. But garbage collection might not happen once the table object is not referenced anymore. It's controlled by python. So if a user turns on this parameter, a table object is not referenced whereas garbage collection doesn't occur, then user turns it off, finally the table will not be dropped.

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-yzou: you can only do this for queries that has reached to server, not for the ones that don't even have a query id.
i think what we can do here is check if the async job is not None, and if it doesn't have a query_id, if not, we can sent a separate telemetry over

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -33,74 +31,51 @@ def __init__(self, session: "Session") -> None:
...
             )  # pragma: no cover
 
-    def is_alive(self) -> bool:
-        return self.cleanup_thread is not None and self.cleanup_thread.is_alive()
+    @property
+    def num_temp_tables_created(self) -> int:
```
sfc-gh-yzou, line 73: if you move the telemetry into this class, i don't think you need those properties anymore.

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -33,74 +31,51 @@ def __init__(self, session: "Session") -> None:
...
+        return sum(v == 0 for v in self.ref_count_map.values())
 
-    def stop(self) -> None:
-        """
-        The cleaner will stop immediately and leave unfinished temp tables in the queue.
-        """
-        self.stop_event.set()
-        if self.is_alive():
-            self.cleanup_thread.join()
+    def reset_reference_count_map(self) -> None:
```
sfc-gh-yzou, line 80: is that only for testing purpose? can you just access the class to clear the map for just testing purpose?

tests/integ/test_temp_table_cleanup.py
sfc-gh-yzou: i see, the whole test is skipped at this moment

sfc-gh-yzou: let's make sure we turn this test on before we start turning the feature on

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: I still need these two properties for testing. I don't see the reason of removing it. If you really mind, I can change it to private method

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: removed

src/snowflake/snowpark/session.py
```diff
@@ -609,8 +606,12 @@ def close(self) -> None:
             raise SnowparkClientExceptionMessages.SERVER_FAILED_CLOSE_SESSION(str(ex))
         finally:
             try:
+                self._conn._telemetry_client.send_temp_table_cleanup_telemetry(
```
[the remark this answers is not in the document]
sfc-gh-jdu: good idea, added back

tests/integ/test_temp_table_cleanup.py
sfc-gh-jdu: Sure, I still think it's better to do it in next PR, which is much cleaner

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: async job will not be None and only query id might be None. I added such a telemetry.

src/snowflake/snowpark/_internal/compiler/telemetry_constants.py
```diff
@@ -15,3 +15,9 @@ class CompilationStageTelemetryField(Enum):
 class SkipLargeQueryBreakdownCategory(Enum):
     ACTIVE_TRANSACTION = "active transaction"
     VIEW_DYNAMIC_TABLE = "view or dynamic table command"
+
+
+class TempTableCleanupTelemetryField(Enum):
+    TYPE_TEMP_TABLE_CLEANUP = "snowpark_temp_table_cleanup"
```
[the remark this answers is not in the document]
sfc-gh-jdu: Sorry didn't realize it's in new compiliation stage, moved.

tests/integ/test_temp_table_cleanup.py
sfc-gh-yzou: sue, we can do it in a different pr

src/snowflake/snowpark/_internal/telemetry.py
```diff
@@ -441,3 +447,37 @@ def send_large_query_optimization_skipped_telemetry(
...
+            ),
+            TelemetryField.KEY_DATA.value: {
+                TelemetryField.SESSION_ID.value: session_id,
+                TelemetryField.NUM_TEMP_TABLES_CLEANED.value: num_temp_tables_cleaned,
+                TelemetryField.NUM_TEMP_TABLES_CREATED.value: num_temp_tables_created,
+            },
+        }
+        self.send(message)
+
+    def send_temp_table_cleanup_exception_telemetry(
```
sfc-gh-yzou, line 469: we can call this "abnormal_exception" since we are not sending this for all exeception,

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -33,74 +31,62 @@ def __init__(self, session: "Session") -> None:
...
+            )
+            query_id = async_job.query_id
+            logging.debug(f"Dropping {common_log_text} with query id {query_id}")
+        except Exception as ex:  # pragma: no cover
+            warning_message = f"Failed to drop {common_log_text}, exception: {ex}"
+            logging.warning(warning_message)
+            if query_id is None:
+                self.session._conn._telemetry_client.send_temp_table_cleanup_exception_telemetry(
+                    self.session.session_id,
+                    warning_message,
```
sfc-gh-yzou, line 73: can we separate the table name and exception message, so that we do not need to parse the fields for information

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -33,74 +31,62 @@ def __init__(self, session: "Session") -> None:
...
         """
         Decrements the reference count of a temporary table,
         and if the count reaches zero, puts this table in the queue for cleanup.
         """
         self.ref_count_map[name] -= 1
         if self.ref_count_map[name] == 0:
-            self.ref_count_map.pop(name)
-            # clean up
-            self.queue.put(name)
+            if self.session.auto_clean_up_temp_table_enabled:
```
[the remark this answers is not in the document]
sfc-gh-jdu: right, but it's just a dict in memory, which should be fine

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/telemetry.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -33,74 +31,62 @@ def __init__(self, session: "Session") -> None:
...
+            num_temp_tables_created=self.num_temp_tables_created,
+        )
+
+    @property
+    def num_temp_tables_created(self) -> int:
+        return len(self.ref_count_map)
+
+    @property
+    def num_temp_tables_cleaned(self) -> int:
+        return sum(v == 0 for v in self.ref_count_map.values())
```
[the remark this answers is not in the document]
sfc-gh-jdu: yes, created the jira and added todo there https://github.com/snowflakedb/snowpark-python/pull/2214#discussion_r1757333516

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -33,74 +31,62 @@ def __init__(self, session: "Session") -> None:
...
     def stop(self) -> None:
         """
-        The cleaner will stop immediately and leave unfinished temp tables in the queue.
+        Stops the cleaner (no-op) and sends the telemetry.
         """
-        self.stop_event.set()
-        if self.is_alive():
-            self.cleanup_thread.join()
+        self.session._conn._telemetry_client.send_temp_table_cleanup_telemetry(
+            self.session.session_id,
```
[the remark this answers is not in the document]
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -33,74 +31,62 @@ def __init__(self, session: "Session") -> None:
...
+            )._internal_collect_with_tag_no_telemetry(
+                block=False, statement_params={DROP_TABLE_STATEMENT_PARAM_NAME: name}
+            )
+            query_id = async_job.query_id
+            logging.debug(f"Dropping {common_log_text} with query id {query_id}")
+        except Exception as ex:  # pragma: no cover
+            warning_message = f"Failed to drop {common_log_text}, exception: {ex}"
+            logging.warning(warning_message)
+            if query_id is None:
+                self.session._conn._telemetry_client.send_temp_table_cleanup_exception_telemetry(
```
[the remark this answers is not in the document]
sfc-gh-jdu: done

sfc-gh-yzou (approved):

src/snowflake/snowpark/session.py
sfc-gh-helmeleegy: The most recent writeup is way clearer. Thanks for this update! I would also add that perhaps an example for a simple script where we can show where a temp table gets created, where it becomes not referenced anymore, and the scope (following this last point) within which it may get dropped assuming (1) python garbage collection is triggered, and (2) the parameter is on.

sfc-gh-aalam (approved):

src/snowflake/snowpark/session.py
sfc-gh-helmeleegy: If there are cases where this scenario can play out because of explicitly using `cache_result` or because of other (implicit) reasons for temp table creation, then we can also have an example for each such case.

src/snowflake/snowpark/session.py
sfc-gh-helmeleegy: Discussed offline. Resolving.

src/snowflake/snowpark/session.py
sfc-gh-jdu: yes, added!

sfc-gh-helmeleegy (approved):
LGTM.

PR #2223, opened 2024-09-04
SNOW-1063346: Remove modin/pandas/dataframe.py

sfc-gh-joshi:

   

   Fixes SNOW-1063346

This PR removes dataframe.py (the Snowpark pandas one, not the Snowpark Python one), following #2167 and #2205. Once more, preserved overrides are given a reason in code comments. The following implemented methods have been added to `dataframe_overrides.py`:
- __init__
- __dataframe__
- __and__, __rand__, __or__, __ror__
- apply, applymap
- columns
- corr
- dropna
- fillna
- groupby
- info
- insert
- isin
- join
- mask
- melt
- merge
- replace
- rename
- pivot_table
- pow
- rpow
- select_dtypes
- set_axis
- set_index
- shape
- squeeze
- sum
- stack
- transpose
- unstack
- value_counts
- where
- iterrows
- itertuples
- __repr__
- _repr_html_
- _to_datetime
- _to_pandas
- __setitem__


src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -3792,6 +3792,7 @@ def nunique():
 
         Examples
         --------
+        >>> import snowflake.snowpark.modin.pandas as pd
```
sfc-gh-azhan, line 3795: This should already been imported here snowflake/snowpark/modin/conftest.py

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -208,7 +208,7 @@ def to_pandas(
     **kwargs: Any,
 ) -> pandas.DataFrame:
     """
-    Convert Snowpark pandas DataFrame to `pandas.DataFrame <https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html>`_
+    Convert Snowpark pandas DataFrame to `pandas.DataFrame <https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html>`__
```
sfc-gh-azhan, line 211: should we remove the lines?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
```diff
@@ -7,23 +7,1474 @@
...
+        register_dataframe_accessor(base_method.__name__)(func)
+        return func
+
+    return decorator
+
+
+# === UNIMPLEMENTED METHODS ===
+# The following methods are not implemented in Snowpark pandas, and must be overridden on the
+# frontend. These methods fall into a few categories:
+# 1. Would work in Snowpark pandas, but we have not tested it.
```
sfc-gh-azhan, line 125: can we simply say all methods that we do not claim in supported API docs should be here?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
```diff
@@ -7,23 +7,1474 @@
...
+def from_dict(
+    cls, data, orient="columns", dtype=None, columns=None
+):  # pragma: no cover # noqa: PR01, RT01, D200
+    pass
+
+
+DataFrame.from_dict = from_dict
+
+
+@snowpark_pandas_telemetry_method_decorator
```
sfc-gh-azhan, line 424: is this decorator required for all override methods?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
sfc-gh-joshi: I was previously running into a warning from Sphinx ("Duplicate explicit target name pandas.dataframe") without the double underscore, which turns it into an anonymous link. It seems to no longer be an issue so I'll revert it.

src/snowflake/snowpark/modin/plugin/docstrings/series.py
sfc-gh-joshi: Reverted, I added this line by mistake.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
sfc-gh-joshi: No, but it is required for most. We don't record telemetry for many methods prefixed by an underscore (like `_to_datetime`).

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
sfc-gh-joshi: Eventually I think we should push the NotImplementedError down to the query compiler wherever possible rather than keeping the frontend method stub defined here. Keeping the stubs here just makes the migration easier for now.

sfc-gh-yzou: i think the answer to this is yes tat this moment. We are considering in the near future might potentially push this hanlization down to query compiler. 
However, @sfc-gh-joshi i am still questioning about this, because api not implemented is for the whole API, but modin have some checks performed at frontend layer, if we move it to query compiler layer, it just means the concrete implementation is missing, but user may still see some error if they didn't pass the correct parameter values inside etc. This behavior could be wired to uers.

src/snowflake/snowpark/modin/pandas/snow_partition_iterator.py
```diff
@@ -5,10 +5,9 @@
 from collections.abc import Iterator
 from typing import Any, Callable
 
+import modin.pandas.dataframe as DataFrame
 import pandas
 
-import snowflake.snowpark.modin.pandas.dataframe as DataFrame
```
sfc-gh-yzou, line 10: @sfc-gh-joshi i think you will need to move this file down to the extensions folder later

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -571,9 +567,7 @@ def __new__(
             attrs (Dict[str, Any]): The attributes of the class.
 
         Returns:
-            Union[snowflake.snowpark.modin.pandas.series.Series,
```
sfc-gh-yzou, line 574: where did dataframe and series go? doesn't it need to return the modin dataframe and series now?

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -183,6 +182,22 @@ def filter(
     pass  # pragma: no cover
 
 
+@register_base_not_implemented()
+def interpolate(
+    self,
+    method="linear",
+    *,
+    axis=0,
```
sfc-gh-yzou, line 190: where does this come from? i though we are only removing dataframe in this pr, why do we have a base override ?

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
```diff
@@ -55,7 +55,7 @@ def to_snowflake(
 
     See Also:
         - :func:`to_snowflake <snowflake.snowpark.modin.pandas.io.to_snowflake>`
-        - :func:`DataFrame.to_snowflake <snowflake.snowpark.modin.pandas.DataFrame.to_snowflake>`
```
sfc-gh-yzou, line 58: @sfc-gh-joshi i want to double check here those are just for doc link, it won't show up to user with the full path, right?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-joshi: The type signature here describes which classes use this in their `metaclass` field, e.g. `class DataFrameGroupBy(metaclass=TelemetryMeta)`. Since we no longer declare Series/DataFrame ourselves, we can't apply this metaclass to them.

Instead, we emulate the behavior of the metaclass in `src/.../modin/pandas/__init__.py` by looping over all of Series/DF's attributes and applying the telemetry annotation to each.

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-joshi: In modin 0.28.1 this method is defined on the base class, but our vendored copy still had it defined on DataFrame.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
sfc-gh-joshi: That's true, and I'm not sure if upstream Modin would be OK with pushing parameter validation to the query compiler in all cases. If that's a concern for a particular method we can leave it as an override here.

src/snowflake/snowpark/modin/plugin/extensions/series_extensions.py
sfc-gh-joshi: That's correct. This determines the path of the method to link to.

I'll see about fixing the link to `to_snowflake` after this PR, it might have been broken for a while now.

Screenshot 2024-09-06 at 12 06 37

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
sfc-gh-azhan: Can we have a unit test to check all public api in dataframe_overrides or series_override have telemetry decorator covered?

sfc-gh-joshi: Is there a good way to check this in unit tests? I don't think there's an easy way to see if a function has a telemetry annotation or not, without actually running the function and checking the logged telemetry.

sfc-gh-yzou (approved):

sfc-gh-azhan (approved):

PR #2225, opened 2024-09-04
[SNOW-1638242] Fix encoding of large decimal

sfc-gh-lspiegelberg:
When encoding decimals (e.g., `Decimal(3.141)`) the default context is precision 28. However, the calculation of how many bytes are needed to convert in `ast_utils.py` had a one-off error for these. Fixed by correctly rounding up the number of bytes required.

With this PR, `pytest src/snowflake/snowpark/functions.py::snowpark.functions.format_number` passes.


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -103,7 +104,9 @@ def build_expr_from_python_val(expr_builder: proto.Expr, obj: Any) -> None:
         unscaled_val = reduce(lambda val, digit: val * 10 + digit, dec_tuple.digits)
         if dec_tuple.sign != 0:
             unscaled_val *= -1
-        req_bytes = (unscaled_val.bit_length() + 7) // 8
+        # Need to round up.
+        req_bytes = int(math.ceil((unscaled_val.bit_length() + 7) / 8))
```
sfc-gh-oplaton, line 108: I'm not sure I understand. If using FP division and `math.ceil`, what's the point of `+ 7`?

sfc-gh-azwiegincew: It appears the argument to `ceil` is in the integer domain. `... / 8.0` should fix that and eliminate the need for `+7`.

sfc-gh-lspiegelberg: refactored and added comment.

tests/unit/test_ast_utils.py
```diff
@@ -0,0 +1,19 @@
...
+from snowflake.snowpark._internal.ast_utils import build_expr_from_python_val
+
+
+@pytest.mark.parametrize(
+    "value", [0, 1, -1, 3.141, "1234567", "12345678.9876543212345676543456"]
+)
+def test_decimal_encoding(value):
+    v = Decimal(value)
+    expr = proto.Expr()
+    build_expr_from_python_val(expr, v)
```
sfc-gh-oplaton, line 19: Optional: The test would be even more useful if it checked the value returned from `build_expr_from_python_val`.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -106,7 +106,11 @@ def build_expr_from_python_val(expr_builder: proto.Expr, obj: Any) -> None:
         unscaled_val = reduce(lambda val, digit: val * 10 + digit, dec_tuple.digits)
         if dec_tuple.sign != 0:
             unscaled_val *= -1
-        req_bytes = (unscaled_val.bit_length() + 7) // 8
+
+        # In two-complement -1 with one byte is 0xFF. We encode arbitrary length integers
+        # in full bytes. Therefore, round up to fullest byte. To restore the sign, add another byte.
+        req_bytes = unscaled_val.bit_length() // 8 + 1
```
sfc-gh-oplaton, line 112: This allocates one extra byte for positive powers of 256. If allocating the optimal number of bytes is a goal, the previous formula works well.

sfc-gh-oplaton (approved):

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: that's correct I believe, we need to have the extra byte to distinguish between positive and negative. Technically a bit would be sufficient, but we round up to full bytes.

tests/unit/test_ast_utils.py
sfc-gh-lspiegelberg: I started with this, but didn't want to add a dedicated decode routine. Thus this test just safeguards against the code not throwing exceptions.

PR #2229, opened 2024-09-04
[SNOW-1514712] UDxF work part II - support UDTFs

sfc-gh-lspiegelberg:
Support UDTFs in Snowpark IR.

Other:
- Fixes UDFRegistration to allow `register_from_file`
- Refactors copy&pasted code from `pandas_udf` and `pandas_udtf` to be cleaner.
- Implement table_function in local testing mode
- Adds support for `PandasSeriesType` and `PandasDataframeType` for SpDataType.
- Fixes base64 issue due to non-deterministic field ordering.


tests/ast/test_ast_driver.py
```diff
@@ -119,57 +119,63 @@ def run_test(session):
             [2, "two"],
             [3, "three"],
```
sfc-gh-lspiegelberg, line 120: Changing here emit mode to False, so the tested functions are the first time when any ast related functionality is invoked. Makes debugging easier.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -859,3 +879,102 @@ def build_udf(
...
+    if_not_exists: bool = False,
+    parallel: int = 4,
+    max_batch_size: Optional[int] = None,
+    strict: bool = False,
+    secure: bool = False,
+    external_access_integrations: Optional[List[str]] = None,
+    secrets: Optional[Dict[str, str]] = None,
+    immutable: bool = False,
+    comment: Optional[str] = None,
+    *,
```
sfc-gh-oplaton, line 905: Aside: To me, one of the mysteries of Python is how positional wildcard parameters fit together with the rest of the parameter list. For instance, why is this here, and not on line 892, where optional parameters start?

Another major mystery of Python is how to evolve function signatures in a backward compatible way.

sfc-gh-lspiegelberg: I modeled this after the API signature of `udtf`, but actually it would be much better to simply remove `*`. Don't really see a need for that here. Wydt?

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -859,3 +879,102 @@ def build_udf(
...
+        if isinstance(output_schema, DataType):
+            output_schema._fill_ast(ast.output_schema.udtf_schema__type.return_type)
+        elif isinstance(output_schema, Sequence) and all(
+            isinstance(el, str) for el in output_schema
+        ):
+            ast.output_schema.udtf_schema__names.schema.extend(output_schema)
+        else:
+            raise ValueError(f"Can not encode {output_schema} to AST.")
+
+    # if return_type is not None:
```
sfc-gh-oplaton, line 932: Commented-out code is hard to read. Is this functionality disabled because it's supposed to be reenabled later? Is it pseudo-code for what the code below actually does?

I suggest deleting the code and, if appropriate, leaving behind a comment about what's missing (if anything), what's feasible here (if it provides useful context for later iterations) etc.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: I'm fine with removing `*`, because other than ignoring arguments (possibly incorrectly), it doesn't add any functionality.

sfc-gh-oplaton: Is ignoring arguments desirable? Do callers of this function forward `*args` and `**kwargs` that they don't understand, and treat as opaque?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: let me delete this piece here! I think this a lapse on my end, thanks for spotting!

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -570,9 +583,11 @@ def build_expr_from_snowpark_column_or_table_fn(
     if isinstance(value, snowflake.snowpark.Column):
         build_expr_from_snowpark_column(expr_builder, value)
     elif isinstance(value, snowflake.snowpark.table_function.TableFunctionCall):
-        raise NotImplementedError(
-            "SNOW-1509198: No support for TableFunctionCall AST generation"
-        )
+        assert value._ast is not None, "TableFunctionCall must have ast assigned."
+        expr_builder.CopyFrom(value._ast)
+        # raise NotImplementedError(
```
sfc-gh-oplaton, line 588: Is this fully implemented now, or should this branch still raise at the end?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: This is actually a recently touched (Python3.9) feature, some context here https://peps.python.org/pep-0570/#specification. Basically `*` and `/` can be used to overcome some issues with mixing positional and named arguments.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: I think so, there are not very exhaustive tests for this though. Let me leave a TODO with tests.

sfc-gh-lspiegelberg: added comment.

sfc-gh-oplaton (approved):

tests/ast/test_ast_driver.py
```diff
@@ -232,7 +249,7 @@ def run_test(session):
 
 @pytest.mark.parametrize("test_case", load_test_cases(), ids=idfn)
 def test_ast(session, test_case):
-    actual, base64 = run_test(
+    actual, base64_str = run_test(
```
sfc-gh-azwiegincew, line 252: base64 is a str by definition, i don't think this name change is good

sfc-gh-azwiegincew (approved):

tests/ast/test_ast_driver.py
sfc-gh-lspiegelberg: it conflicts with the python module `base64`. I am accessing that particular module below. If you feel strongly, I can do `from base64 import ...` to call these functions and restore the name.

sfc-gh-azwiegincew: Ah, no, that's cool.

PR #2233, opened 2024-09-04
SNOW-1632701: Add uuid to SnowflakePlan

sfc-gh-aalam:

   

   Fixes SNOW-1632701

   Associate a `uuid` with each `SnowflakePlan` so that we can track all queries that originated from the same plan.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -254,12 +254,16 @@ class SnowflakePlan(LogicalPlan):
         # In the placeholder query, subquery (child) is held by the ID of query plan
         # It is used for optimization, by replacing a subquery with a CTE
         self.placeholder_query = placeholder_query
-        # encode an id for CTE optimization
+        # encode an id for CTE optimization. This is generated based on the main
+        # query and the associated query parameters. We use this id for equality comparison
+        # to determine if two plans are the same.
         self._id = encode_id(queries[-1].sql, queries[-1].params)
         self.referenced_ctes: Set[str] = (
             referenced_ctes.copy() if referenced_ctes else set()
         )
         self._cumulative_node_complexity: Optional[Dict[PlanNodeCategory, int]] = None
+        # UUID for the plan to uniquely identify the SnowflakePlan object. We also use this
+        # to UUID track queries that are generated from the same plan.
         self._uuid = str(uuid.uuid4())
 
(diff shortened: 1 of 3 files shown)


sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -260,6 +260,7 @@ def __init__(
             referenced_ctes.copy() if referenced_ctes else set()
         )
         self._cumulative_node_complexity: Optional[Dict[PlanNodeCategory, int]] = None
+        self._uuid = str(uuid.uuid4())
```
sfc-gh-yzou, line 263: so there is very embracing state here that the snowflake_plan its-self already have and self._id fields, and used as part of the __eq__ comparison, the id is generated based on the query and params, which may not be the exact indicator of whether two plan are the same plan. 
I think we can keep both, but let's make sure we we doc it clearly here about the difference. @sfc-gh-jdu i also start question is changing the __eq__ and __hash__ correct in this case? I got the point that we want to simplify the comparison, but that also kind of changes the meaning of the equivalence of two snowflake plan nodes here.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -272,6 +273,14 @@ def __eq__(self, other: "SnowflakePlan") -> bool:
     def __hash__(self) -> int:
         return hash(self._id) if self._id else super().__hash__()
 
+    @property
+    def uuid(self) -> str:
+        return self._uuid
+
+    @uuid.setter
+    def uuid(self, value: str) -> None:
```
sfc-gh-yzou, line 281: i don't think we should provide a setter function here, i don't think anyone is suppose to change this once produced.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -502,6 +511,7 @@ def __deepcopy__(self, memodict={}) -> "SnowflakePlan":  # noqa: B006
         copied_plan._is_valid_for_replacement = True
         if copied_source_plan:
             copied_source_plan._is_valid_for_replacement = True
+        copied_plan.uuid = self.uuid
```
sfc-gh-yzou, line 514: by definition, the copied_plan should have a different uuid here, right? because it is a different node

sfc-gh-yzou (approved):

PR #2241, opened 2024-09-05
SNOW-1548942 Implement support for dataframe apply axis=0

sfc-gh-evandenberg:

SNOW-1548942 Implement support for dataframe apply axis=0

This pr adds support for df.apply with axis=0


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -710,6 +710,9 @@ def python_type_to_snow_type(
             )
         return TimestampType(timezone), False
 
+    if tp == object:
```
sfc-gh-azhan, line 713: where do you need this? Maybe implement this in pandas_lit for now. This mapping may need more discussion I believe.

sfc-gh-evandenberg: This is needed to support 'object' type hint, which is where this is called, the test test_Apply_axis_0 with parameter [[[1.1, 2.2], [3, np.nan]], lambda x: "0" if x.sum() > 1 else 0, "object"]) which essientally creates a type hint -> object for the apply function.

It is called via

check_return_variant_and_get_return_Type (apply_utils)
--> get_types_from_type_hints (udf_utils)
-->  calls python_type_to_snow_type (type_utils)

The type hint will fail to be recognized with a TypeError("invalid type <class 'object'>") without this check.

@sfc-gh-jdu looks like your familiar with this code, can you check this is best way to support 'object' type hint?

sfc-gh-azhan: Why user want to create a type hint for object?

sfc-gh-evandenberg: It looks like the original frame/test_apply.py test (for axis=1) to check type hints (including 'object') was added by https://snowflakecomputing.atlassian.net/browse/SNOW-1042200 (@sfc-gh-lspiegelberg) as a performance improvement, he may have more insight why type hint for variant would be useful.  In this case we're building on top of existing udtf apply code (specifically groupby apply) which do check these type hints if they are provided as found here.

sfc-gh-azhan: I was wondering why it worked for axis=1 cases but not now and then I found this https://github.com/snowflakedb/snowpark-python/blob/f6438409cadd28bfc947dc6ce164f36718b36fb6/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py#L813
Can we instead reuse `deduce_return_type_from_function` for `check_return_variant_and_get_return_Type`?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -392,6 +393,8 @@
 MICROSECONDS_PER_SECOND = 10**6
 NANOSECONDS_PER_DAY = SECONDS_PER_DAY * NANOSECONDS_PER_SECOND
 
+TEMP_GROUPBY_COLUMN = "tmp_groupby"
```
sfc-gh-nkrishna, line 396: Nit: Let's define this in `groupby_utils.py` instead and import it here?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7858,56 +7875,258 @@ def apply(
...
+            ) in zip(
+                frame.data_column_pandas_labels,
+                frame.data_column_snowflake_quoted_identifiers,
+            ):
+                data_col_frame = InternalFrame.create(
+                    ordered_dataframe=frame.ordered_dataframe,
+                    data_column_snowflake_quoted_identifiers=[
+                        data_column_snowflake_quoted_identifier
+                    ],
+                    data_column_pandas_labels=[data_column_pandas_label],
```
sfc-gh-nkrishna, line 7915: Can you add a comment about what this is doing? My understanding is you are maintaining the index but grabbing only the last data column in the frame? If that is the case, why do this instead of using `drop`?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7858,56 +7875,258 @@ def apply(
...
+                data_col_result_frame = InternalFrame.create(
+                    ordered_dataframe=data_col_result_frame.ordered_dataframe,
+                    data_column_snowflake_quoted_identifiers=data_col_result_frame.data_column_snowflake_quoted_identifiers,
+                    data_column_pandas_labels=[data_column_pandas_label],
+                    data_column_pandas_index_names=data_col_frame.data_column_pandas_index_names,
+                    data_column_types=None,
+                    index_column_snowflake_quoted_identifiers=data_col_result_frame.index_column_snowflake_quoted_identifiers,
+                    index_column_pandas_labels=data_col_result_frame.index_column_pandas_labels,
+                    index_column_types=None,
+                )
```
sfc-gh-nkrishna, line 7947: Why not use `set_index_names` instead?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7858,56 +7875,258 @@ def apply(
...
+                                    :-1
+                                ],
+                                index_column_pandas_labels=qc_frame.index_column_pandas_labels[
+                                    :-1
+                                ],
+                                index_column_types=qc_frame.cached_index_column_snowpark_pandas_types[
+                                    :-1
+                                ],
+                            )
+                        )
```
sfc-gh-nkrishna, line 8007: I think this can be performed with either `drop` (using the `index` parameter, though that may not be implemented) or `iloc`. Can you check if that is possible?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7858,56 +7875,258 @@ def apply(
...
+                            ) == len(col_result_indexes)
+                            qc_result = col_results[0].concat(
+                                axis=1, other=col_results[1:], sort=not all_same_index
+                            )
+                        else:
+                            # If there's a mix of scalar and pd.Series output from the apply func, pandas stores the
+                            # pd.Series output as the value, which we do not currently support.
+                            raise NotImplementedError(
+                                "Nested pd.Series in result is not supported"
+                            )
```
sfc-gh-nkrishna, line 8027: Nit: add the method to the NotImplementedError
```suggestion
                            ErrorMessage.not_implemented((
                                "Nested pd.Series in result is not supported in DataFrame.apply(axis=0)"
                            )
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7858,56 +7875,258 @@ def apply(
...
+                else:
+                    if any(is_scalar for is_scalar in col_result_scalars):
+                        # If there's a mix of scalar and pd.Series output from the apply func, pandas stores the
+                        # pd.Series output as the value, which we do not currently support.
+                        raise NotImplementedError(
+                            "Nested pd.Series in result is not supported"
+                        )
```
sfc-gh-nkrishna, line 8034:
```suggestion
                            ErrorMessage.not_implemented((
                                "Nested pd.Series in result is not supported in DataFrame.apply(axis=0)"
                            )
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7858,56 +7875,258 @@ def apply(
...
+                        # If there are multiple output series with different indices, then line them up as a series output.
+                        all_same_index = all(
+                            all(i == col_result_indexes[0]) for i in col_result_indexes
+                        )
+                        # If the col results all have same index then we keep the existing index ordering.
+                        qc_result = col_results[0].concat(
+                            axis=1, other=col_results[1:], sort=not all_same_index
+                        )
+
+            # If result should be Series then change te data column label appropriately.
```
sfc-gh-nkrishna, line 8058:
```suggestion
            # If result should be Series then change the data column label appropriately.
```

tests/integ/modin/frame/test_apply_axis_0.py
```diff
@@ -0,0 +1,608 @@
...
+
+
+import scipy.stats  # noqa: E402
+
+
+@pytest.mark.parametrize(
+    "packages,expected_query_count",
+    [
+        (["scipy", "numpy"], 26),
+        (["scipy>1.1", "numpy<2.0"], 26),
```
sfc-gh-nkrishna, line 575: Why is the query count here so high? Can you add a comment?

tests/integ/modin/frame/test_apply_axis_0.py
```diff
@@ -0,0 +1,608 @@
...
+    ],
+)
+def test_apply_axis_0_bug_1650918(data, apply_func):
+    native_df = native_pd.DataFrame(data)
+    snow_df = pd.DataFrame(native_df)
+    eval_snowpark_pandas_result(
+        snow_df,
+        native_df,
+        lambda x: x.apply(apply_func, axis=0),
+    )
```
sfc-gh-nkrishna, line 565: Is this is raising SnowparkSQLException should be catch the exception with `pytest` instead of using `eval_snowpark_pandas_result`?

sfc-gh-nkrishna (commented):
Can you update the CHANGELOG and `dataframe_supported.rst` accordingly?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-evandenberg: Added comment. This is iterating through each of the data columns of the original df, so data_column_pandas_label, data_column_snowflake_quoted_identifier are the current column, we then create a frame reduced down to this single data column to call groupby_apply with apply function on.  So it's not always the last data column.  The index is preserved because the apply function might access via the index.

tests/integ/modin/frame/test_apply_axis_0.py
sfc-gh-evandenberg: This looks like an existing bug in the underlying udtf support not handling NULL, i created the jira.  Don't we want to use xfail since it is a known failing test, rather than having it "pass".  The idea here is we don't want to refactor/rewrite the test when the issue is fixed, we simply remove the xfail annotation, ie. test driven development.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-evandenberg: I tried drop with index, it fails with an error:

"snowflake.snowpark.exceptions.SnowparkSQLException: (1304): 01b6dabb-0001-a95e-0000-047d1166174a: 100038 (22018): Numeric value 'A' is not recognized"

Also doesn't seem like iloc works here as that selects based on the index range, however, here we want all the rows just to drop one part of the resulting multi-index remaining from the concat above.

sfc-gh-helmeleegy:
> Can you update the CHANGELOG and `dataframe_supported.rst` accordingly?

Can you also update the docstring in `src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py` accordingly?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-evandenberg: It also needs to set the data_column_pandas_label otherwise it won't have the expected result.  I updated the comment to mention this.

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-evandenberg: Updated

docs/source/modin/supported/dataframe_supported.rst
```diff
@@ -84,7 +84,7 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``any``                     | P                               |                                  | ``N`` for non-integer/boolean types                |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``apply``                   | P                               |                                  | ``N`` if ``axis == 0`` or ``func`` is not callable |
+| ``apply``                   | P                               |                                  | ``N`` if ``func`` is not callable |
```
sfc-gh-nkrishna, line 87: Nit: fix the end `|` bar to match the others

tests/integ/modin/frame/test_apply_axis_0.py
sfc-gh-nkrishna: Ah okay, I misunderstood. Yeah xfail with `eval_snowpark_pandas_result` is best here.

sfc-gh-nkrishna (approved):
Left one minor nit, but LGTM. Nice work!

docs/source/modin/supported/dataframe_supported.rst
sfc-gh-evandenberg: Done

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
```diff
@@ -738,8 +738,6 @@ def apply():
             * 0 or 'index': apply function to each column.
             * 1 or 'columns': apply function to each row.
 
-            Snowpark pandas does not yet support ``axis=0``.
```
sfc-gh-helmeleegy, line 741: Can we also re-enable the skipped doctest for axis=0?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7903,56 +7918,261 @@ def apply(
                 "Snowpark pandas apply API doesn't yet support DataFrame or Series in 'args' or 'kwargs' of 'func'"
             )
 
-        # get input types of all data columns from the dataframe directly
-        input_types = self._modin_frame.get_snowflake_type(
-            self._modin_frame.data_column_snowflake_quoted_identifiers
-        )
+        if axis == 0:
```
sfc-gh-azhan, line 7921: Screenshot 2024-09-06 at 4 09 07 PM
why not just reuse groupby.apply to implement this?

tests/integ/modin/frame/test_apply.py
```diff
@@ -554,33 +544,71 @@ def g(v):
...
-    [[[1.3, 2.5]], np.sqrt],
-    [[[1.3, 2.5]], "sqrt"],
-    [[[1.3, 2.5]], np.log],
-    [[[1.3, 2.5]], "log"],
-    [[[1.3, 2.5]], np.square],
-    [[[1.3, 2.5]], "square"],
+@pytest.mark.xfail(
+    strict=True,
+    raises=SnowparkSQLException,
+    reason="SNOW-1650918: Apply on dataframe data columns containing NULL fails with invalid arguments to udtf function",
```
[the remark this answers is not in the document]
sfc-gh-evandenberg: No this is an existing bug I found so I added a test case and jira.

tests/integ/modin/frame/test_apply.py
```diff
@@ -554,33 +544,71 @@ def g(v):
...
+    "data, apply_func, expected_query_count", TRANSFORM_DATA_FUNC_MAP
+)
+def test_basic_dataframe_transform(data, apply_func, expected_query_count):
+    if expected_query_count is None:
+        msg = "Snowpark pandas apply API only supports callables func"
+        with pytest.raises(NotImplementedError, match=msg):
+            snow_df = pd.DataFrame(data)
+            snow_df.transform(apply_func)
+    else:
+        msg = "SNOW-1650644 & SNOW-1345395: Avoid extra caching and repeatedly creating same temp function"
```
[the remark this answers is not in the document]
sfc-gh-evandenberg: I'm not that familiar with df.transform as that was not the main objective of this pr, it's just something that was enabled for free as a byproduct of supporting df.apply(axis=0).  There may be other ways to implement df.transform not sure on that, this the default implementation. 

@sfc-gh-helmeleegy Do you think we should continue failing df.transform here over performance?  The approach with df.apply(axis=0) is to reuse existing building blocks like groupby_apply so these are pre-existing perf issues that are otherwise orthogonal to this pr.  I was thinking we can make forward progress on perf issues in parallel since we also have a lot of focus there and can start getting feedback to prioritize our efforts.  Any thoughts here?

tests/integ/modin/frame/test_apply.py
```diff
@@ -554,33 +544,71 @@ def g(v):
...
-@sql_count_checker(query_count=0)
-def test_basic_dataframe_transform(data, apply_func):
-    msg = "Snowpark pandas apply API doesn't yet support axis == 0"
-    with pytest.raises(NotImplementedError, match=msg):
+@pytest.mark.parametrize(
+    "data, apply_func, expected_query_count", TRANSFORM_DATA_FUNC_MAP
+)
+def test_basic_dataframe_transform(data, apply_func, expected_query_count):
+    if expected_query_count is None:
+        msg = "Snowpark pandas apply API only supports callables func"
```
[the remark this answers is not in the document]
sfc-gh-evandenberg: The apply_func is executed within a udtf python code, it's expecting to be a python func rather than a sql function, so we would need to add an alternative implementation to support this small subset of callables.  Given this isn't related to the goals of this pr, I created a separate jira https://snowflakecomputing.atlassian.net/browse/SNOW-1657167 to look into this further.

tests/integ/modin/frame/test_apply_axis_0.py
```diff
@@ -0,0 +1,608 @@
...
+
+@pytest.mark.parametrize(
+    "data, func, return_type", BASIC_DATA_FUNC_PYTHON_RETURN_TYPE_MAP
+)
+@pytest.mark.modin_sp_precommit
+def test_axis_0_basic_types_without_type_hints(data, func, return_type):
+    # this test processes functions without type hints and invokes the UDTF solution.
+    native_df = native_pd.DataFrame(data, columns=["A", "b"])
+    snow_df = pd.DataFrame(data, columns=["A", "b"])
+    with SqlCounter(udtf_count=2):
```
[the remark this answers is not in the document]
sfc-gh-evandenberg: Added

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7903,56 +7918,261 @@ def apply(
...
+            for (
+                data_column_pandas_label,
+                data_column_snowflake_quoted_identifier,
+            ) in zip(
+                frame.data_column_pandas_labels,
+                frame.data_column_snowflake_quoted_identifiers,
+            ):
+                # Create a frame for the current data column which we will be passed to the apply function below.
+                # Note that we maintain the original index because the apply function may access via the index.
+                data_col_frame = InternalFrame.create(
```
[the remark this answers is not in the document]
sfc-gh-evandenberg: Good point, I tried it out and it does work.  The resulting query is a bit more complex because it requires the index to be passed in, so for example I can pass in index=slice(0,Integer.MAX_VALUE,1) and it filters on that.  While it's not a big difference, it seems like we can avoid any additional over changing the frame metadata directly as done here.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-evandenberg: The basic approach is to use groupby apply, however it is more nuanced, check out the last part of the doc (https://docs.google.com/document/d/1P2HsPYKXt_RQ-nevtGohwPercffHKNc8R_aDmHHzQnE/edit) which basically shows how to do this.  The fake_group index part can be avoided internally as there is a way to force grouping the entire index which is what we want in this case.  Note that I'm not able to reproduce your example above.

Screenshot 2024-09-09 at 11 00 32 AM

Also the behavior is slightly different if the apply_func maps to series vs scalar or combinations across columns.  The final result could be a dataframe, series or scalar.  The apply function is applied to each column separately and then results and joined or concatenated together as appropriate.  There's more book keeping necessary to handle all of these cases.

tests/integ/modin/frame/test_apply.py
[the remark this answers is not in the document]
sfc-gh-helmeleegy: @sfc-gh-evandenberg, would it be possible to run a small test to compare perf with and without caching for df.apply(axis=0) and df.transform()? If it becomes clear that caching is not helping in these cases (seems like it given the high number of queries), then I believe we should turn it off for those cases for now. Later, we should have a more comprehensive plan to decide on a general caching strategy for all the apply/transform use cases.

docs/source/modin/supported/dataframe_supported.rst
```diff
@@ -84,7 +84,7 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``any``                     | P                               |                                  | ``N`` for non-integer/boolean types                |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``apply``                   | P                               |                                  | ``N`` if ``axis == 0`` or ``func`` is not callable |
+| ``apply``                   | P                               |                                  | ``N`` if ``func`` is not callable                  |
```
sfc-gh-helmeleegy, line 87: Can we reference the list of limitations given for `GroupBy.apply` in `groupby_supported.rst`?

docs/source/modin/supported/dataframe_supported.rst
sfc-gh-helmeleegy, line 1: Should we also update the notes for `transform` now that axis=0 is supported?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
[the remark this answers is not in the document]
sfc-gh-azhan: You can pass index=slice(None,None) to skip index selection.

tests/integ/modin/frame/test_apply.py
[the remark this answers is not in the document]
sfc-gh-evandenberg: @sfc-gh-helmeleegy I ran the related unit tests with and without caching and found they are ~ 2.5x slower if there is no caching.  Yes, let's agree to follow up here (we have tracking jiras) on more comprehensive plan for general caching.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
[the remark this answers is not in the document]
sfc-gh-evandenberg: Updated

docs/source/modin/supported/dataframe_supported.rst
sfc-gh-evandenberg: Updated, it wasn't very clear but transform is only working if func is a callable, otherwise it is not supported.

docs/source/modin/supported/dataframe_supported.rst
sfc-gh-evandenberg: I looked at the limitations of groupby_supported.rst for GroupBy.apply and they are not hit in the case of Dataframe.apply(axis=0) so I don't think there's anything to document there since this is otherwise an implementation detail.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7905,56 +7920,258 @@ def apply(
...
 
-        # get input types of all data columns from the dataframe directly
-        input_types = self._modin_frame.get_snowflake_type(
-            self._modin_frame.data_column_snowflake_quoted_identifiers
-        )
+        if axis == 0:
+            frame = self._modin_frame
 
-        from snowflake.snowpark.modin.pandas.utils import try_convert_index_to_native
+            # To apply function to Dataframe with axis=0, we repurpose the groupby apply function by taking each
```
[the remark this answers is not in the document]
sfc-gh-azhan: BUT, if a df has 10 columns, then this approach is going to create 10 UDTFs and run them sequentially which is slow. We can define a single UDTF to handle the whole df in one go. Why not?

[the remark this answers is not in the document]
sfc-gh-evandenberg: Yeah, there are different approaches, here I'm reusing existing building block (groupby_apply) as much as possible so we can start enabling for customers, at which point we can see what kind of workloads and performance our customers are encountering and prioritize further improvements accordingly.

While an uber UDTF approach would reduce the number of round-trip queries, it's not clear to what extent the overall performance would be better.  For instance, partitioning by columns (as done here) should enable better scalability for larger data sets since they operate on column fraction of the total dataset at a time, and further these queries could be parallelized so as to substantially decreasing the overall latency.  While a single uber udtf approach would need the entire dataset in memory at one time but take similar overall execution time without the parallelization opportunity being in a single python method.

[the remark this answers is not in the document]
sfc-gh-azhan: No doubt this approach is ready to use at least.

The performance of a single UDTF approach should be better but I don't have the data to show how much it can be better IMO. Taking the 10 column case as an example, the current approach basically run 10 queries and each one UDTF basically runs the select query and apply the UDTF on the selected row sets. If we can combine them into one query and then the backend automatically will parallel the process for each rowsets across the warehouse instances.

But it's probably too early to optimize right now. It's better to add some perf tests for apply axis=0 first. cc @sfc-gh-nkumar @sfc-gh-helmeleegy @sfc-gh-lyambem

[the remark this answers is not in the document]
sfc-gh-evandenberg: I've created a jira https://snowflakecomputing.atlassian.net/browse/SNOW-1659582 to follow up on evaluating performance options when we have time.

Also had some side discussion with @sfc-gh-helmeleegy will share for visibility:

Screenshot 2024-09-10 at 2 22 32 PM

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7909,56 +7924,258 @@ def apply(
...
+            if len(col_results) == 1:
+                result_is_series = col_result_scalars[0]
+                qc_result = col_results[0]
+
+                # Squeeze to series if it is single column
+                qc_result = qc_result.columnarize()
+                if col_result_scalars[0]:
+                    qc_result = qc_result.reset_index(drop=True)
+            else:
+                single_row_output = all(len(index) == 1 for index in col_result_indexes)
```
sfc-gh-azhan, line 8018: would `len(index)` trigger eager evaluation again?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7909,56 +7924,258 @@ def apply(
...
+                    index_column_snowflake_quoted_identifiers=data_col_result_frame.index_column_snowflake_quoted_identifiers,
+                    index_column_pandas_labels=data_col_result_frame.index_column_pandas_labels,
+                    index_column_types=data_col_result_frame.cached_index_column_snowpark_pandas_types,
+                )
+
+                data_col_result_index = (
+                    data_col_result_frame.index_columns_pandas_index()
+                )
+                col_result_indexes.append(data_col_result_index)
+                col_result_scalars.append(
```
sfc-gh-azhan, line 8002: for some popular methods, `np.sum`, we should know whether the result is scalar or not so we can avoid eager evaluation here. Can be a TODO item.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-evandenberg: No, the index is retrieved earlier and this is the Index value used here so does not fetch again.

sfc-gh-azhan (approved):
LGTM as the first step to support axis=0. We should add some workflows to see how the performance looks like.

sfc-gh-helmeleegy (approved):
LGTM. I'm glad we could verify that the current implementation using some extra cache_result eager queries is (counter-intuitively) faster than the alternative implementation with no result caching and hence no eager queries.

Also glad that we have a number of JIRA tickets for possible performance improvement directions to pursue in the future.

Thanks for the nice work, Eric.

PR #2248, opened 2024-09-06
SNOW-1653121: Support some Timedelta aggregations on axis=0.

sfc-gh-mvashishtha:
Fixes SNOW-1653121

Test and support aggregation on axis=0. We still raise `NotImplementedError` if:

1) the aggregation requires concatenating a frame with timedelta types
2) the aggregation requires transposing a row containing a timedelta type and other types.

To facilitate this change, introduce the class `SnowparkPandasAggFunc`, which represents a Snowpark pandas aggregation, together with the bool `preserves_snowpark_pandas_type`. We can use `SnowparkPandasAggFunc` instead of the separate mappings `GROUPBY_AGG_PRESERVES_SNOWPARK_PANDAS_TYPE` and `GROUPBY_AGG_WITH_NONE_SNOWPARK_PANDAS_TYPES`.

This change also fixes the bug that timedelta aggregations like mean would produce the wrong type (and the wrong result) by truncating the float result if `preserves_snowpark_pandas_type`.


sfc-gh-mvashishtha:
marking as draft till I fix SNOW-1654730

sfc-gh-nkrishna (approved):
Nice work, LGTM!

tests/integ/modin/conftest.py
```diff
@@ -715,3 +715,29 @@ def numeric_test_data_4x4():
...
+                pd.Timedelta(minutes=-1),
+                pd.Timedelta(minutes=0),
+                pd.Timedelta(minutes=5),
+                pd.Timedelta(minutes=6),
+            ],
+            "C": [
+                None,
+                pd.Timedelta(nanoseconds=5),
+                pd.Timedelta(nanoseconds=0),
+                pd.Timedelta(nanoseconds=4),
```
sfc-gh-nkumar, line 740: not sure if this covered in other tests. It would good to include a column withe all None values.

sfc-gh-nkumar (approved):

tests/integ/modin/conftest.py
sfc-gh-mvashishtha: Done.

PR #2249, opened 2024-09-06
SNOW-1557539 count describe queries for test

sfc-gh-evandenberg:

   Fixes SNOW-1557539 count describe queries for test

Add describe query to query listener history (opt-in)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,6 @@ This is a re-release of 1.22.0. Please refer to the 1.22.0 release notes for det
   - `is_transient`
   - `data_retention_time`
   - `max_data_extension_time`
-- Added support for passing parameter `include_describe` to `Session.query_history`.
 
 #### Bug Fixes
 
@@ -80,6 +79,7 @@ This is a re-release of 1.22.0. Please refer to the 1.22.0 release notes for det
 
 - Added support for type coercion when passing columns as input to UDF calls.
 - Added support for `Index.identical`.
+- Added support for passing parameter `include_describe` to `Session.query_history`.
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Added support for `TimedeltaIndex.mean` method.
 - Added support for some cases of aggregating `Timedelta` columns on `axis=0` with `agg` or `aggregate`.
 - Added support for `by`, `left_by`, `right_by`, `left_index`, and `right_index` for `pd.merge_asof`.
+- Added support for passing parameter `include_describe` to `Session.query_history`.
 
 #### Bug Fixes
 
@@ -79,7 +80,6 @@ This is a re-release of 1.22.0. Please refer to the 1.22.0 release notes for det
 
 - Added support for type coercion when passing columns as input to UDF calls.
 - Added support for `Index.identical`.
-- Added support for passing parameter `include_describe` to `Session.query_history`.
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ This is a re-release of 1.22.0. Please refer to the 1.22.0 release notes for det
   - `is_transient`
   - `data_retention_time`
   - `max_data_extension_time`
+- Added support for passing parameter `include_describe` to `Session.query_history`.
 
 #### Bug Fixes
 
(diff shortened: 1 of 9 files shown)


src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -244,9 +244,18 @@ def _get_string_datum(self, query: str) -> Optional[str]:
...
-            run_new_describe(self._cursor, query), self.max_string_size
+            self._run_new_describe(self._cursor, query), self.max_string_size
         )
 
+    def _run_new_describe(self, cursor: SnowflakeCursor, query: str) -> Union[List[ResultMetadata], List["ResultMetadataV2"]]:
+        result_metadata = run_new_describe(cursor, query)
+
+        describe_listeners = [listener for listener in self._query_listener if hasattr(listener, 'describe_listener')]
+        for listener in describe_listeners:
+            listener._add_query(QueryRecord(cursor.sfqid, query, True))
```
sfc-gh-yixie, line 255: This has an impact to existing customers. How about add an optional parameter `session.query_history(include_describe: bool=False)`?

sfc-gh-evandenberg: Updated the QueryHistory with include_describe

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -244,9 +244,23 @@ def _get_string_datum(self, query: str) -> Optional[str]:
...
+            self._run_new_describe(self._cursor, query), self.max_string_size
         )
 
+    def _run_new_describe(
+        self, cursor: SnowflakeCursor, query: str
+    ) -> Union[List[ResultMetadata], List["ResultMetadataV2"]]:
+        result_metadata = run_new_describe(cursor, query)
+
+        for listener in filter(
+            lambda listener: hasattr(listener, "include_describe")
```
sfc-gh-aalam, line 256:
```suggestion
            lambda listener: listener.include_describe
```

sfc-gh-aalam, line 256: I don't understand why we need a `hasattr` check here. isn't `listener.include_describe` sufficient?

sfc-gh-evandenberg: For example, in the case of SqlCounter we don't inherit from QueryHistory, we just implement _add_query directly on the class and add as a listener.  However, adding this property breaks the client unless the new property is added.  May be we could do this differently but this isn't an uncommon pattern so added the check here to avoid potentially breaking clients that do this kind of duck typing.

src/snowflake/snowpark/session.py
```diff
@@ -3399,7 +3399,7 @@ def flatten(
         set_api_call_source(df, "Session.flatten")
         return df
 
-    def query_history(self) -> QueryHistory:
+    def query_history(self, include_describe: bool = False) -> QueryHistory:
```
sfc-gh-yixie, line 3402: Need to update the docstring for the new parameter `include_describe`.
Pls also list this as a new feature in the changelog.

sfc-gh-aalam (approved):

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
   - `is_transient`
   - `data_retention_time`
   - `max_data_extension_time`
+- Added support for passing parameter `include_describe` to `Session.query_history`.
```
sfc-gh-aalam, line 43: please move this to new section

sfc-gh-yixie (approved):

src/snowflake/snowpark/session.py
```diff
@@ -3460,16 +3460,19 @@ def flatten(
...
         """Create an instance of :class:`QueryHistory` as a context manager to record queries that are pushed down to the Snowflake database.
 
+        Args:
+            include_describe: Include query notifications for describe queries
+
```
sfc-gh-helmeleegy, line 3468: Can we also add a code example?

sfc-gh-helmeleegy:
Will we plan for the opt-in parameter to be opt-out at some point?

sfc-gh-evandenberg:
> Will we plan for the opt-in parameter to be opt-out at some point?

@sfc-gh-yixie @sfc-gh-aalam Any thoughts here?  I'm not sure describe queries are as interesting to track by default but may make sense at some point.

sfc-gh-helmeleegy (approved):
LGTM. thanks, Eric!

sfc-gh-aalam:
> > Will we plan for the opt-in parameter to be opt-out at some point?
> 
> @sfc-gh-yixie @sfc-gh-aalam Any thoughts here? I'm not sure describe queries are as interesting to track by default but may make sense at some point.

I think this is minor enough that we can keep it opt-in forever.

PR #2256, opened 2024-09-09
Improve warning message for read_snowflake to include materialization reason

sfc-gh-yzou:
1. Fill out the following pre-review checklist:

2. Please describe how your code solves the related issue.
read_snowflake will trigger table materialization under cases for hybrid table, iceberg table etc. include the reason inside the warning message to make the reason more clear to users

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/utils.py
@@ -265,8 +265,9 @@ def _create_read_only_table(
         temp_table_name = random_name_for_temp_object(TempObjectType.TABLE)
 
         _logger.warning(
-            f"Data from source table/view '{table_name}' is being copied into a new "
-            f"temporary table '{temp_table_name}'. DataFrame creation might take some time."
+            f"Snapshot source table/view '{table_name}' failed due to reason: `{materialization_reason}'. Data from "
+            f"source table/view '{table_name}' is being copied into a new "
+            f"temporary table '{temp_table_name}' for snapshotting. DataFrame creation might take some time."
         )
 
         statement_params = get_default_snowpark_pandas_statement_params()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,12 +109,14 @@
 
 #### Improvements
 
+- Improve concat, join performance when operations are performed on series coming from the same dataframe by avoiding unnecessary joins.
 - Refactored `quoted_identifier_to_snowflake_type` to avoid making metadata queries if the types have been cached locally.
 - Improved `pd.to_datetime` to handle all local input cases. 
 - Create a lazy index from another lazy index without pulling data to client.
 - Raised `NotImplementedError` for Index bitwise operators.
 - Display a clearer error message when `Index.names` is set to a non-like-like object.
 - Raise a warning whenever MultiIndex values are pulled in locally.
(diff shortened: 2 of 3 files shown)


sfc-gh-nkumar (approved):

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -265,8 +265,9 @@ def _create_read_only_table(
         temp_table_name = random_name_for_temp_object(TempObjectType.TABLE)
 
         _logger.warning(
-            f"Data from source table/view '{table_name}' is being copied into a new "
-            f"temporary table '{temp_table_name}'. DataFrame creation might take some time."
+            f"Snapshot source table/view '{table_name}' failed due to reason: `{materialization_reason}'. Data from "
```
sfc-gh-azhan, line 268: should we add a limit of rows here? If customer work on a huge table, this can be expensive!.

sfc-gh-yzou: sorry, what do you mean limit of rows? do you mean add in the message like `if your table is more than 10,000 rows, the materialization could be slow'?

sfc-gh-yzou: or do you mean change how the materialization happens, like when materialization, we put an limit there to only materialize a subset of the table. 
I don't think we should do that without customer notice, since that is just going to be partial of the data. we can potentially offer an option in the api to only get a set of the data, but i think we should check with customer if they would like such option.

sfc-gh-azhan: In this case, we could set a default value of limit of rows and raise a warning message and tell customer if this limit is not enough, they can set to None to ignore it or use a larger limit. 
What I'm worry about here is that there are lots of huge tables in Snowflake, for those cases, copy can be quite expensive and slow. They won't get any result in this case and also have to pay the bill for the long or cancelled query. cc @sfc-gh-dolee @sfc-gh-qding

sfc-gh-dolee: The row limit should be set at the to_pandas level so that the row limit logic + warning is shared across everything that leads to a materialization step. I don't think we should do anything custom only for Multi-Index.

sfc-gh-yzou: I think we can move this discussion offline. This pr is just trying to improve the current warning message, so that the user can know clearly why the materialization happens. 
Maybe it could be a good option to user, but i think we might need to understand more about customer's need, if a subset of data is not what users want, that seems a waste also. 
@sfc-gh-dolee i agree that it shouldn't be something we do specially for mutli-index

sfc-gh-azhan: First of all, we didn't talk about multi-index here.

to_pandas is too late, in this case, if the target table is 1PB, the copy is done after calling read_snowflake. The to_pandas limit is usually specified by the running environment, like notebook. We didn't add a default limit for to_pandas though.

sfc-gh-yzou: in fact, the fundamental solution is to support the snapshotting for most commonly used cases.

sfc-gh-azhan (approved):
approve for this warning.

PR #2260, opened 2024-09-10
SNOW-1657456 Improved  to persist the original timezone offset for TIMESTAMP_TZ type

sfc-gh-azhan:

   

   Fixes SNOW-1657456

The problem was that to_pandas use Snowflake Python connector to read query result into a pandas dataframe. However, the connector will convert TIMESTAMP_TZ result to timestamps using the local session timezone which does not match with the pandas behavior in our scenario. 
The goal is to preserve the original timezone offsets for timestamp_tz.
The idea here is to convert timestamp_tz to string before calling to_pandas and then cast them back to datetime64tz on the native pandas result.


sfc-gh-helmeleegy:
This looks great! One quick ask: In the PR description, can you start with explaining the original problem and then explain how this PR provides a solution for it?

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1384,10 +1387,36 @@ def snowpark_to_pandas_helper(
             [typeof(col(id)) for id in variant_type_identifiers],
         )
 
+    # Step 3.2: cast timestamp_tz to string to preserve their original timezone offsets
```
sfc-gh-helmeleegy, line 1390: Can you update the docstring with this new behavior as well? Right now, it only talks about handling variant types.

sfc-gh-helmeleegy: Unrelated to this PR, but we should also mention the timedelta handling in the docstring.

sfc-gh-azhan: I don't think we should talk about those in the docstring because they match exactly the pandas behavior. We only mention those that dismatch.

sfc-gh-helmeleegy: The way the docstring is written today seems to be addressing the internal developer and describing what the function does. I believe it matched the original implementation. But as more updates were made, it somewhat became outdated. I don't have a strong opinion here though. It just felt that since we are discussing the special handling of variants, mentioning other types of special handling would be appropriate.

sfc-gh-helmeleegy (approved):
LGTM. Thanks for working on this!

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1493,7 +1550,10 @@ def convert_str_to_timedelta(x: str) -> pd.Timedelta:
             # multiple timezones. So here we cast the index to the index_type when ret = pd.Index(...) above cannot
             # figure out a non-object dtype. Note that the index_type is a logical type may not be 100% accurate.
             if is_object_dtype(ret.dtype) and not is_object_dtype(index_type):
-                ret = ret.astype(index_type)
+                try:
+                    ret = ret.astype(index_type)
+                except ValueError:  # e.g., Tz-aware datetime.datetime cannot be converted to datetime64
+                    pass
```
sfc-gh-nkrishna, line 1556: In this case thoughts on using `tz_localize(None)` for naive local time or `tz_convert(None)` for UTC time, then calling `astype`. Might be better than catching and passing the ValueError

sfc-gh-nkrishna (commented):
Left a question and one nit

sfc-gh-nkrishna (approved):

sfc-gh-jjiao:
qq: this is not in Snowpark Pandas GA release, right? so this fix is not in the scope of the GA ?

sfc-gh-nkrishna:
> qq: this is not in Snowpark Pandas GA release, right? so this fix is not in the scope of the GA ?

Correct, this not a GA release blocker/requirement

PR #2266, opened 2024-09-10
REFACTOR: Reuse `pd.date_range` for resample computation

sfc-gh-nkrishna:
This PR refactors Snowpark pandas `resample` implementation to use `pd.date_range` for the resulting index (complete with missing bins) rather than compute the number of rows required in the output and using a generator function. `pd.date_range` uses a generator under-the-hood.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/resample_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/resample_utils.py
@@ -2,7 +2,6 @@
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
 
-import math
 from typing import Any, Literal, NoReturn, Optional, Union
 
 import pandas as native_pd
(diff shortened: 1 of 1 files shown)


src/snowflake/snowpark/modin/plugin/_internal/resample_utils.py
```diff
@@ -478,39 +468,15 @@ def get_expected_resample_bins_frame(
...
-        (native_pd.to_datetime(end_date) - native_pd.to_datetime(start_date))
-        / to_offset(rule)
-        + 1
-    )
-
-    expected_resample_bins_snowpark_frame = pd.session.generator(
-        all_resample_bins_col, rowcount=rowcount
-    )
-
+    expected_resample_bins_snowpark_frame = pd.date_range(
```
sfc-gh-helmeleegy, line 471: Can you explain in the PR description (not necessarily as a code comment) why the new implementation is equivalent to the old one?

sfc-gh-mvashishtha (approved):

sfc-gh-helmeleegy (approved):
LGTM, thanks Naren!

PR #2267, opened 2024-09-10
SNOW-1654416: Add support for TimedeltaIndex.mean

sfc-gh-nkumar:
SNOW-1654416: Add support for TimedeltaIndex.mean


tests/integ/modin/index/test_timedelta_index_methods.py
```diff
@@ -128,3 +128,29 @@ def test_timedelta_total_seconds():
...
+
+@sql_count_checker(query_count=0)
+def test_timedelta_index_mean_invalid_axis():
+    native_index = native_pd.TimedeltaIndex([1, 2, 3])
+    snow_index = pd.Index(native_index)
+    with pytest.raises(IndexError, match="tuple index out of range"):
+        native_index.mean(axis=1)
+    # Snowpark pandas raises ValueError instead of IndexError.
+    with pytest.raises(ValueError, match="axis should be 0 for TimedeltaIndex.mean"):
+        snow_index.mean(axis=1)
```
sfc-gh-nkrishna, line 156: Unsure if the `to_pandas()` is necessary here with SQL Simplifier turned off.
```suggestion
        snow_index.mean(axis=1).to_pandas()
```

sfc-gh-nkrishna (approved):

src/snowflake/snowpark/modin/plugin/extensions/timedelta_index.py
```diff
@@ -407,17 +412,39 @@ def mean(
...
-        -----
-        mean is only defined for Datetime and Timedelta dtypes, not for Period.
         """
+        if axis:
+            # Native pandas raises IndexError: tuple index out of range
+            # We raise a different more user-friendly error message.
+            raise ValueError(
+                f"axis should be 0 for TimedeltaIndex.mean, found '{axis}'"
+            )
+        frame = self._query_compiler._modin_frame
```
sfc-gh-azhan, line 436: why we don't have a QC method to do so? Or alternatively, you can do something similar to `to_series()` and `mean` for this

sfc-gh-nkrishna: I think `agg` is still in-progress by @sfc-gh-mvashishtha, but should be similar?

sfc-gh-nkumar: > why we don't have a QC method to do so?

In my understanding we don't have such QC method to operate on index columns, only on data columns. @sfc-gh-yzou would have more context why we don't have that?

> you can do something similar to to_series() and mean for this

We don't have that support yet. It throws NotImplementedError but It would definitely makes sense to reuse code once that is available.

sfc-gh-azhan: Please add a todo link to https://snowflakecomputing.atlassian.net/browse/SNOW-1620439 and we can refactor this code later.

sfc-gh-azhan (approved):

src/snowflake/snowpark/modin/plugin/extensions/timedelta_index.py
```diff
@@ -407,17 +412,39 @@ def mean(
...
-
-        Notes
-        -----
-        mean is only defined for Datetime and Timedelta dtypes, not for Period.
         """
+        if axis:
+            # Native pandas raises IndexError: tuple index out of range
+            # We raise a different more user-friendly error message.
+            raise ValueError(
+                f"axis should be 0 for TimedeltaIndex.mean, found '{axis}'"
```
sfc-gh-helmeleegy, line 434: This is a bit weird. If `axis` has to be 0, what's the point of having it as a parameter in the first place?

sfc-gh-helmeleegy (approved):
LGTM - except for my lack of understanding of the constraint on the axis parameter. I think it should be fine though given that this is how pandas behaves.

src/snowflake/snowpark/modin/plugin/extensions/timedelta_index.py
sfc-gh-nkumar: yeah, pandas is 30% weird, 70% useful.

src/snowflake/snowpark/modin/plugin/extensions/timedelta_index.py
sfc-gh-nkumar: done

PR #2273, opened 2024-09-11
SNOW-1636678 add server side param for complexity bounds

sfc-gh-aalam:

   

   Fixes SNOW-1636678

   This PR makes the following changes:
   1. Read default values for complexity score bounds from session parameters. These bounds are used in large query breakdown optimization.
   2. Allow users to easily change the bounds using `session` object.
   3. Send telemetry about updates made to bounds by the user.


src/snowflake/snowpark/session.py
```diff
@@ -223,6 +223,16 @@
...
 )
+_PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_UPPER_BOUND = (
+    "PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_UPPER_BOUND"
+)
+_PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_LOWER_BOUND = (
+    "PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_LOWER_BOUND"
+)
+# The complexity score lower bound is set to match COMPILATION_MEMORY_LIMIT
+# in Snowflake. This is the limit where we start seeing compilation errors.
+DEFAULT_COMPLEXITY_SCORE_LOWER_BOUND = 10_000_000
```
sfc-gh-helmeleegy, line 234: I'm assuming the compilation memory limit is measured in bytes. Can we also assume that our complexity score is measured in bytes? Wasn't it more based on the number of plan nodes?

If they don't use the same units, then is there a mapping between such units that we use?

src/snowflake/snowpark/session.py
```diff
@@ -223,6 +223,16 @@
...
     "PYTHON_SNOWPARK_USE_LARGE_QUERY_BREAKDOWN_OPTIMIZATION"
 )
+_PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_UPPER_BOUND = (
+    "PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_UPPER_BOUND"
+)
+_PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_LOWER_BOUND = (
+    "PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_LOWER_BOUND"
+)
+# The complexity score lower bound is set to match COMPILATION_MEMORY_LIMIT
+# in Snowflake. This is the limit where we start seeing compilation errors.
```
sfc-gh-helmeleegy, line 233: Do we know if this is a soft limit or a hard limit? IoW, does the compiler intentionally errors out as soon as the limit is exceeded, or does it keep going anyway with undefined behavior?

src/snowflake/snowpark/session.py
sfc-gh-helmeleegy, line 234: I'm also assuming that this limit is configurable on the compiler side. Is there a way to pull the currently configured value instead of hard coding it here - to make sure the two configurations are in sync?

src/snowflake/snowpark/session.py
```diff
@@ -784,6 +810,24 @@ def large_query_breakdown_enabled(self, value: bool) -> None:
                 "value for large_query_breakdown_enabled must be True or False!"
             )
 
+    @large_query_breakdown_complexity_bounds.setter
+    def large_query_breakdown_complexity_bounds(self, value: Tuple[int, int]) -> None:
```
sfc-gh-yzou, line 814: might be easier to take a lower and upper bound there, then you can reconstruct the tuple internally, and we can make lower or upper optional also

sfc-gh-yzou (approved):
left a comment

src/snowflake/snowpark/session.py
sfc-gh-aalam: @sfc-gh-yzou this is a `@setter` method. I'm afraid we have to abide by this format.

sfc-gh-jdu (approved):

PR #2274, opened 2024-09-11
SNOW-1661142 Fix index name behavior

sfc-gh-vbudati:
   Fixes SNOW-1661142

Fixed a bug where updating an index's name updates the parent's index name when it is not supposed to. This is done by verifying that the query_compiler recorded during the index's creation matches that of its parent object when the parent object must be updated.
```py
>>> df = pd.DataFrame(
...     {
...         "A": [0, 1, 2, 3, 4, 4],
...         "B": ['a', 'b', 'c', 'd', 'e', 'f'],
...     },
...     index = pd.Index([1, 2, 3, 4, 5, 6], name = "test"),
... )
>>> index = df.index
>>> df
      A  B
test      
1     0  a
2     1  b
3     2  c
4     3  d
5     4  e
6     4  f
>>> index.name = "test2"
>>> 
>>> df
       A  B
test2      
1      0  a
2      1  b
3      2  c
4      3  d
5      4  e
6      4  f
>>> df.dropna(inplace=True)
>>> index.name = "test3"
>>> df
       A  B
test2        # <--- name should not update      
1      0  a
2      1  b
3      2  c
4      3  d
5      4  e
6      4  f
```
For the full discussion, see thread: https://docs.google.com/document/d/1vdllzNgeUHMiffFNpm9SD1HOYUk8lkMVp14HQDoqr7s/edit?disco=AAABVbKjFJ0


src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
```diff
@@ -163,6 +163,7 @@ def __new__(
         index._query_compiler = query_compiler
         # `_parent` keeps track of any Series or DataFrame that this Index is a part of.
         index._parent = None
```
sfc-gh-yzou, line 165: how about get a class like IndexParent, and it records the necessary parent information, the constructor can take the df and record the information property, and it can also provide an interface like no_change_on_parent (you can use other name if you have other ones in mind) to help decide wether a replacement should happen, this could help wrap all parent logic at one place and simplify the code handling in the future

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -726,7 +730,9 @@ def name(self, value: Hashable) -> None:
         if not is_hashable(value):
             raise TypeError(f"{type(self).__name__}.name must be a hashable type")
         self._query_compiler = self._query_compiler.set_index_names([value])
-        if self._parent is not None:
+        # Update the name of the parent's index only if the parent's current query compiler
+        # matches the recorded query compiler (_parent_qc).
+        if self._parent is not None and self._parent_qc is self._parent._query_compiler:
```
sfc-gh-yzou, line 735: also now after updating, you should make sure the parent_qc is updated to the new one, you can also wrap that logic in the IndexParent class

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-yzou: also, i have a quick question, shouldn't datetime_index an extension of index class, and it can call super().__init__ to init the common part, right? how come we are initialization the same thing at different place?

sfc-gh-vbudati: `__init__` in Index.py is used only for docstring generation, the actual initialization is happening in `__new__`. I'm not sure why it was designed this way.
But I created a new class like you suggested to abstract away the common logic.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: done!

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -2262,7 +2262,7 @@ def _get_index(self):
         return self._query_compiler.index
 
     idx = Index(query_compiler=self._query_compiler)
-    idx._set_parent(self)
+    idx._parent.set_parent(self)
```
sfc-gh-azhan, line 2265: can we update `_set_parent` instead?

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -71,6 +71,41 @@
 }
 
 
+class IndexParent:
+    def __init__(self) -> None:
```
sfc-gh-azhan, line 75: Can we always init with values not None? if index._parent is None, then no parent, if it is not, it should mean parent exists.

sfc-gh-azhan (changes requested):

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-yzou: I think the _init_ can always take a df as input, index._parent can remain optional, which makes the relationship more clear

src/snowflake/snowpark/modin/plugin/extensions/datetime_index.py
sfc-gh-yzou: i think the __new__ is implemented because the returned object could be different based on the type. i don't think we have to fix that in this pr, just curious why we didn't call super class for the common part? cc @sfc-gh-azhan do you know?

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-yzou: actually do we need that one line wrapper? the current place seems the only place calls it

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: done! The init should now take `parent` and be the only way to set the parent details.

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
sfc-gh-vbudati: I reverted it to `_set_parent()` since that seems like a "safer" way to set the parent details without exposing too much

sfc-gh-azhan (approved):

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -726,10 +760,10 @@ def name(self, value: Hashable) -> None:
         if not is_hashable(value):
             raise TypeError(f"{type(self).__name__}.name must be a hashable type")
         self._query_compiler = self._query_compiler.set_index_names([value])
+        # Update the name of the parent's index only if the parent's current query compiler
```
sfc-gh-yzou, line 763: can you update the doc for the name function to make it clear that the inplace replacement only works when no inplace update has been applied to parent

sfc-gh-yzou (approved):
@sfc-gh-vbudati Left some comment about docs and comments, please take care of it before merging!

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-vbudati: done!

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -71,6 +71,35 @@
...
+            The parent object that the Index is a part of.
+        """
+        assert isinstance(parent, (DataFrame, Series))
+        self._parent = parent
+        self._parent_qc = parent._query_compiler
+
+    def check_and_update_parent_qc_index_names(self, names: list) -> None:
+        """
+        Update the Index and its parent's index names if the parent's current query compiler matches
+        the recorded query compiler (`_parent_qc`).
```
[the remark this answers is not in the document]
sfc-gh-vbudati: done!

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 - Added support for `TimedeltaIndex.mean` method.
 
+#### Bug Fixes
+
+- Fixed a bug where an `Index` object created from a `Series`/`DataFrame` incorrectly updates the `Series`/`DataFrame`'s index name when it is not supposed to.
```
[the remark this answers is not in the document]
sfc-gh-vbudati: done!

PR #2278, opened 2024-09-11
[SNOW-1638290] Add GetItemObject entity and fix Dataframe.random_split

sfc-gh-lspiegelberg:
Supports new GetItemObject Snowpark IR entity, fixes `Dataframe.random_split`.

Other:
- Changes AST symbol capture to use builtin python `ast` module instead of regex.
- Allow capture of multiple targets in assign statements for AST symbol capture.
- Fix `DataFrame.join` to reproduce in unparser keywords arguments that were missing before.


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -46,6 +45,53 @@
 SRC_POSITION_TEST_MODE = False
 
 
+class ExtractAssignmentVisitor(ast.NodeVisitor):
+    def __init__(self) -> None:
+        super().__init__()
+        self.symbols: Optional[Union[str, List[str]]] = None
+
+    def visit_Assign(self, node) -> None:
```
sfc-gh-oplaton, line 53: I think this can be `visit_assign`, or `visit_assign_node`. The current capitalization seems unidiomatic for Python and is, frankly, more distracting to read. (`VisitAssign` or `visit_assign`)

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -46,6 +45,53 @@
...
+    """
+    try:
+        tree = ast.parse(source_line.strip())
+        v = ExtractAssignmentVisitor()
+        v.visit(tree)
+        return v.symbols
+    except Exception:
+        # It may happen that an incomplete source line is submitted.
+        # For this reason, try to parse one more time if '=' is found.
+        if "=" in source_line:
```
sfc-gh-oplaton, line 82: Shouldn't this be a retry loop, or a pass that fixes up all potentially broken source lines? Or a pass that fixes up all source lines before attempting the tree traversal?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: Unfortunately that won't work because this here is overriding a base-class method from `ast.NodeVisitor`. The naming indeed is weird, and an oddity of python. For context, the pattern is
`visit_<classname>`. https://docs.python.org/3/library/ast.html#ast.NodeVisitor.visit

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: I think only a single source line is given. In python to the best of my knowledge the left-side of an assign can't be broken up along multiple lines. Therefore, from an input parameter perspective we only need a single line. However, we could simplify by basically always searching for the first `'='` occurence and then using the substition with `None` (or some other generic token).

Do you think this would be a good refactor?

Python doesn't allow for `'='` as part of the name of an identifier, so we should be safe with this approach.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: If `ast.NodeVisitor` follows a know Python idiom that violates the more common naming convention, a comment isn't strictly necessary, but having one in `ast.NodeVisitor` itself might help.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: You're right that would be a legal case. If we wanted to support this, we would have to do some eager line parsing till we succeed parsing something including an AST. Another problematic case may be

`df = df.select(...); df2 = df.select(...)`.

Here, the wrong symbol would be picked up for the second statement. I don't think we'll have a perfect solution without a lot of additional work. My suggestion is therefore to take up a more incremental approach. With this PR, the basic case of multiple targets is supported similar to the existing code. The existing code does not handle your example, not the multi-statement per-line case.

Do you think we could defer implementation for these more complex cases? The worst that will happen currently is that the symbol doesn't get captured and something like `res1, res2, ...` will get generated. Wdyt?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: `ast.NodeVisitor` is a builtin class. It comes as part of the python standard library :). Ok, in this I'll leave as is.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: We'll end up parsing Python for this at some point, and I agree it's not worth doing now. Let's go with a simple hack until someone breaks it.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: Oh, it's not our `ast` 😄  OK, that makes sense.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: Sgtm, I want to pick up your code regarding the nesting here. I think when applying `source_line = source_line[: source_line.find("=")] + " = None"` directly, the code can be simplified. Will refactor and add a comment to explain.

sfc-gh-lspiegelberg: refactored.

sfc-gh-oplaton (approved):

PR #2282, opened 2024-09-11
SNOW-1418523 make analyzer server connection thread safe

sfc-gh-aalam:

   

   Fixes SNOW-1418523

   **[NOTE] Easier to review by selecting `hide whitespace` option when viewing diffs** 
   This PR makes the following updates:
   1. `server_connection.py`: 
   - Adds `_thread_store` to create a `_cursor` per thread to enable concurrent querying
   - Protects `_query_listener` updates using `_lock` so all queries can be added in a thread-safe manner
   2. `session.py`: 
   - Create per thread `_analyzer` to enable concurrent dataframe operations.
   - Use lock to protect generation of actions ids
   - Use lock to protect running alter query tag and updating `_query_tag`.
   - Protect `_session_stage` creation with a lock so only one session stage is created and used.


sfc-gh-aalam:
https://ci-dev-142.int.snowflakecomputing.com/job/PythonStoredProcBuildPrecommitTest/1322/

sfc-gh-yuwang (approved):

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -183,6 +185,12 @@ def __init__(
             "_skip_upload_on_content_match" in signature.parameters
         )
 
+    @property
+    def _cursor(self) -> SnowflakeCursor:
+        if not hasattr(self._thread_stored, "cursor"):
+            self._thread_stored.cursor = self._conn.cursor()
+        return self._thread_stored.cursor
```
sfc-gh-aling, line 192: does this operation need to be protected by self._lock?

src/snowflake/snowpark/session.py
```diff
@@ -623,8 +620,19 @@ def __str__(self):
...
+
+    @property
+    def _analyzer(self) -> Analyzer:
+        if not hasattr(self._thread_store, "analyzer"):
+            self._thread_store.analyzer = (
+                Analyzer(self)
+                if isinstance(self._conn, ServerConnection)
+                else MockAnalyzer(self)
+            )
+        return self._thread_store.analyzer
```
sfc-gh-aling, line 635: same as above, do we need to protect this as well?

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: this is a `threading.local()` object so each thread has its own copy. We don't need a lock for this step.

src/snowflake/snowpark/session.py
sfc-gh-aalam: same. this is a thread local object. this the how we create a per-thread instance of cursor and analyzer.

sfc-gh-aling (approved):

PR #2284, opened 2024-09-12
SNOW-1662105, SNOW-1662657: Support `by`, `left_by`, `right_by` for `pd.merge_asof`

sfc-gh-nkrishna:
SNOW-1662105, SNOW-1662657

This PR refactors `join_utils.py` to make `left_on` and `right_on` optional arguments, as they are not required for "cross" or "asof" joins. It also support `by`, `left_by`, `right_by` for `pd.merge_asof`.


sfc-gh-jjiao:
@sfc-gh-nkrishna  I created a JIRA ticket for this refactor work from you, could you please or do you mind I linking it to this PR? https://snowflakecomputing.atlassian.net/browse/SNOW-1662105  Thanks

sfc-gh-vbudati (approved):
Thanks Naren!

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -161,40 +204,31 @@ def join(
...
+        assert (
+            right_match_col
+        ), "ASOF join was not provided a column identifier to match on for the right table"
+        assert (
+            match_comparator
+        ), "ASOF join was not provided a comparator for the match condition"
+        assert_snowpark_pandas_types_match(
+            left, right, [left_match_col], [right_match_col]
+        )
+    if left_on and right_on:
```
sfc-gh-azhan, line 223: we should keep the check for how is not `asof` or `cross`.

sfc-gh-azhan (changes requested):

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkrishna: Which check are you referring to? We keep the check for snowpark pandas types for all join types, and we assert that `how` is in `JoinTypeLit`.

sfc-gh-azhan: The length assertion below in line 224. We should make sure this is always run when how is not "asof"

sfc-gh-nkrishna: The assertion should always be run if `left_on` and `right_on` are not empty or None. Even in the case of `asof` that check gets run if `on` is specified. Maybe I'm misunderstanding?

sfc-gh-helmeleegy:
Perhaps @sfc-gh-nkumar can take a quick look?

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -103,12 +103,55 @@ class JoinOrAlignInternalFrameResult(NamedTuple):
     result_column_mapper: JoinOrAlignResultColumnMapper
 
 
+def assert_snowpark_pandas_types_match(
+    left: InternalFrame,
+    right: InternalFrame,
+    left_identifiers: list[str],
+    right_identifiers: list[str],
+) -> None:
+    """
```
sfc-gh-yzou, line 112: i assume that is a direct code movement without change, is that right?

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-yzou, line 223: the check should be if left_on or right_on, if no right on is provided, or no left_on is provided, that would invalid case also

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -161,40 +204,31 @@ def join(
...
-                rt = rt if rt is not None else right.get_snowflake_type(right_on[i])
-                raise ValueError(
-                    f"You are trying to merge on {type(lt).__name__} and {type(rt).__name__} columns for key '{key}'. "
-                    f"If you wish to proceed you should use pd.concat"
-                )
-
-    assert_snowpark_pandas_types_match()
+    if how == "asof":
+        assert (
+            left_match_col
```
sfc-gh-yzou, line 212: for asof join, we can provide both match col and join on column, what would happen if both on and match columns are provided? if we only have tested support for match column, let's error out if both are provided today

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -161,40 +204,31 @@ def join(
...
+            match_comparator
+        ), "ASOF join was not provided a comparator for the match condition"
+        assert_snowpark_pandas_types_match(
+            left, right, [left_match_col], [right_match_col]
+        )
+    if left_on and right_on:
+        assert len(left_on) == len(
+            right_on
+        ), "left_on and right_on must be of same length or both be None"
+        if join_key_coalesce_config is not None:
```
sfc-gh-yzou, line 227: the join key validation key is also valid for asof, right? but the left_on might be None now.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -161,40 +204,31 @@ def join(
...
+            assert len(join_key_coalesce_config) == len(
+                left_on
+            ), "join_key_coalesce_config must be of same length as left_on and right_on"
```
sfc-gh-yzou, line 230: we should also check if how is not asof, no match column should be provided

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -161,40 +204,31 @@ def join(
...
+        assert_snowpark_pandas_types_match(left, right, left_on, right_on)
```
sfc-gh-yzou, line 231: i think you can have a variable left_join_key, and right_join_key, and set the value correctly for both asof and non-asof, and use it when passing to _create_internal_frame_with_join_or_align_result also

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkrishna: Yes no code change beyond changing the variable names (using `left/right_identifiers` instead of `left/right_on`.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkrishna: Changed check to `or` condition.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkrishna: We can have both `on` and `match` columns for ASOF Join (in Snowflake SQL + Snowpark Python APIs). The Python API does have tests for both though even though we don't use it in the pandas layer. 

I don't think we need to error out in pandas if both are provided, but can change it to do so if you think it's better.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkrishna: Update: changed this to an "if/else" ASOF case to make it clearer.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkrishna: Done

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-yzou: What i am not worried here is not whether we error it out or not, but more about that we are not handling it correctly. For example, if i provide bout match col and on, should the join key be both match_col and on columns, or just on columns, or just match cols?
if we support both, we should make sure we have test for it

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -216,8 +265,8 @@ def assert_snowpark_pandas_types_match() -> None:
         left,
         right,
         how,
-        left_on,
-        right_on,
+        left_join_key,
```
sfc-gh-yzou, line 268: do we also need to update JoinOrAlignOrderedDataframeResultHelper construction also, seems we might missed that part before

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1213,6 +1213,7 @@ def join(
             assert match_comparator, "match_comparator was not provided to ASOF Join"
             snowpark_dataframe = left_snowpark_dataframe_ref.snowpark_dataframe.join(
                 right=right_snowpark_dataframe_ref.snowpark_dataframe,
+                on=on,
```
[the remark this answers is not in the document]
sfc-gh-nkrishna: Good point, done

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkrishna: @sfc-gh-azhan @sfc-gh-yzou I added support in `pd.merge_asof` for `by`, `left_by`, and `right_by` with tests to show how `left_on` and `right_on` are used with ASOF Join since it wasn't much additional work and it's a good test case for this.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1197,22 +1197,29 @@ def join(
...
             right_match_col = Column(right_identifiers_rename_map[right_match_col])
             # ASOF Join requires the use of match_condition
-            assert match_comparator, "match_comparator was not provided to ASOF Join"
+            assert (
+                match_comparator
+            ), "ASOF join was not provided a comparator for the match condition"
+
+            on = None
+            for left_col, right_col in zip(left_on_cols, right_on_cols):
+                eq = Column(left_col).__eq__(Column(right_col))
```
sfc-gh-nkrishna, line 1217: `equal_null` does not work here as I get:

`SQLCompilationError: ON clause for ASOF JOIN must contain conjunctions of equality conditions only. Disjunctions are not allowed.`

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -161,40 +206,48 @@ def join(
...
+        assert (
+            left_match_col is None
+            and right_match_col is None
+            and match_comparator is None
+        ), f"match condition should not be provided for {how} join"
+        if join_key_coalesce_config is not None:
+            assert len(join_key_coalesce_config) == len(
+                left_join_key
+            ), "join_key_coalesce_config must be of same length as left_on and right_on"
+        assert_snowpark_pandas_types_match(left, right, left_join_key, right_join_key)
```
sfc-gh-azhan, line 250: we don't need to write `assert_snowpark_pandas_types_match` twice.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-azhan: Does this match with pandas behavior?

sfc-gh-nkrishna: Yes there are tests in `test_merge_asof.py` that have NULL/NaN values and that introduce NULL/NaN values after performing the merge.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-nkrishna: Good catch, fixed

sfc-gh-azhan (approved):

PR #2287, opened 2024-09-12
Capture AST for DataFrame.pivot and DataFrame.unpivot

sfc-gh-oplaton:

   

   Fixes SNOW-1491297

Capture AST for `DataFrame.pivot` and `DataFrame.unpivot`.
Add expectation test for `pivot`.
I was unable to resolve the local testing issue for `unpivot`. Adding a disabled test for now, and will revisit the test issue separately.


src/snowflake/snowpark/dataframe.py
```diff
@@ -2237,19 +2244,32 @@ def unpivot(
             ---------------------------------------------
             <BLANKLINE>
         """
+        # AST.
```
sfc-gh-lspiegelberg, line 2247: can you add `_emit_ast` as guard? We'll use that for rollout parameter.

src/snowflake/snowpark/dataframe.py
```diff
@@ -2237,19 +2244,32 @@ def unpivot(
...
+        self.set_ast_ref(ast.df)
+        ast.value_column = value_column
+        ast.name_column = name_column
+        for c in column_list:
+            build_expr_from_snowpark_column_or_col_name(ast.column_list.add(), c)
+
         column_exprs = self._convert_cols_to_exprs("unpivot()", column_list)
         unpivot_plan = Unpivot(value_column, name_column, column_exprs, self._plan)
 
+        df: DataFrame
```
sfc-gh-lspiegelberg, line 2259: nit: can do `df = ... if ... else ...`. pre-commit will auto-format.

tests/ast/data/DataFrame.unpivot.test.DISABLED
```diff
@@ -0,0 +1,12 @@
+## TEST CASE
+
```
sfc-gh-lspiegelberg, line 2: maybe let's add ticket to fix this?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2192,7 +2193,12 @@ def pivot(
         """
 
         if _emit_ast:
```
sfc-gh-lspiegelberg, line 2195: before this if, there should be something like `stmt = None`. Else, for `_emit_ast=False` the return statement will raise an undefined name exception.

src/snowflake/snowpark/dataframe.py
```diff
@@ -2204,6 +2210,7 @@ def pivot(
             snowflake.snowpark.relational_grouped_dataframe._PivotType(
                 pc[0], pivot_values, default_on_null
```
sfc-gh-lspiegelberg, line 2211: nit: we might want to rename `ast_stmt` to `_ast_stmt` or so.

sfc-gh-lspiegelberg:
Few nits, do you have by chance the error that happens in local testing mode for unpivot?

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: That approach will cost more than it saves now.

A feature flag:

- Starts out as "off" by default
- Early adopters can set it to "on", everyone else has it off
- At some point, the feature transitions to on by default
- Some users might hit blocking issues. They can set the feature flag to "off"
- At some point, all blocking issues are fixed for good.
- The feature flag disappears altogether, together with any fallback code paths that might have been present

`emit_ast` has semantics orthogonal to the rollout status of the feature. Using `emit_ast` for two purposes at once implies that:

- We introduce `emit_ast` in the signature of all public APIs, not just the ones called internally from other public APIs
- Whoever has to remove the "feature flag" uses of `emit_ast` has to differentiate those uses from the "silence internal uses of public APIs" uses of `emit_ast` in function signatures and in implementations. It is this last part that will be very expensive for whoever does it.

sfc-gh-oplaton:
> Few nits, do you have by chance the error that happens in local testing mode for unpivot?

The verbatim error is:

```
session = <snowflake.snowpark.session.Session object at 0x16ddc9e80>
test_case = TestCase(filename='DataFrame.unpivot.test', source='\ndf = session.create_dataframe(\n    [(1, "electronics", 100, 200... df.unpivot("sales", "month", ["jan", "feb"]).sort("empid")\n\n\n', expected_ast_base64='', expected_ast_unparsed='\n')

    @pytest.mark.parametrize("test_case", load_test_cases(), ids=idfn)
    def test_ast(session, test_case):
>       actual, base64_str = run_test(
            session, test_case.filename.replace(".", "_"), test_case.source
        )

/Users/oplaton/src/snowparkpy/tests/ast/test_ast_driver.py:252: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Users/oplaton/src/snowparkpy/tests/ast/test_ast_driver.py:243: in run_test
    base64_batches = test_module.run_test(session)
/var/folders/5r/s68_pbc972lfk7s9qms41w1c0000gn/T/test_asthorw47_6.py:94: in run_test
    ???
/Users/oplaton/src/snowparkpy/src/snowflake/snowpark/_internal/telemetry.py:212: in wrap
    r = func(*args, **kwargs)
/Users/oplaton/src/snowparkpy/src/snowflake/snowpark/dataframe.py:2263: in unpivot
    from_=SelectSnowflakePlan(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <snowflake.snowpark._internal.analyzer.select_statement.SelectSnowflakePlan object at 0x16dd3a640>
snowflake_plan = <snowflake.snowpark._internal.analyzer.unary_plan_node.Unpivot object at 0x16e05dac0>

    def __init__(self, snowflake_plan: LogicalPlan, *, analyzer: "Analyzer") -> None:
        super().__init__(analyzer)
        self._snowflake_plan: SnowflakePlan = (
            snowflake_plan
            if isinstance(snowflake_plan, SnowflakePlan)
            else analyzer.resolve(snowflake_plan)
        )
>       self.expr_to_alias.update(self._snowflake_plan.expr_to_alias)
E       AttributeError: 'NoneType' object has no attribute 'expr_to_alias'

/Users/oplaton/src/snowparkpy/src/snowflake/snowpark/_internal/analyzer/select_statement.py:552: AttributeError
```

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/dataframe.py
sfc-gh-lspiegelberg: I might have not been precise here: We'll add a decorator to guard the rollout of the IR feature for [SNOW-1557209](https://snowflakecomputing.atlassian.net/browse/SNOW-1557209), i.e. something like

```
@guard
def public_api(..., _emit_ast=True):
   ...
```
This guard will simply pass `_emit_ast=False` to the API. The behavior of the decorator `guard` will be controlled by a variable then (feature flag). For existing APIs, the sort-of established rule right now is that will have a parameter `_emit_ast:bool=True`.

However, I sense that you have some very good points here. Should we schedule a call with @sfc-gh-azwiegincew to align on the strategy with `_emit_ast`?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2237,19 +2244,32 @@ def unpivot(
...
                         unpivot_plan, analyzer=self._session._analyzer
                     ),
                     analyzer=self._session._analyzer,
                 )
             )
-        return self._with_plan(unpivot_plan)
+            if self._select_statement
+            else self._with_plan(unpivot_plan)
+        )
+        df._ast_id = stmt.var_id.bitfield1
```
sfc-gh-lspiegelberg, line 2271:
```
if _emit_ast:
  df._ast_id = ...
```

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: The decorator is an implementation detail of the rollout. That's less interesting. The salient point is that we'd end up using `emit_ast` for two purposes, so (assuming we want to finish the rollout and retire the feature flag), untangling the two purposes at the end will result in non-trivial work, and in changes to the public API itself.

sfc-gh-lspiegelberg: Hm, how could we do better?

sfc-gh-oplaton: One question I'd ask is "how do other Snowpark features get rolled out?" Is the decorator + parameter pattern common?

Another question would be - with or without a decorator, if `emit_ast` didn't exist, how would feature control work? I presume some piece of code would fetch configuration from one or more places (`.toml` file, environment variable, command-line flag, server response etc), make a policy decision and pass it around - by storing it in a global variable, or by returning it from a (memoized?) helper function etc.

In any case, I propose treating feature control completely separately from `emit_ast`.

For functions that don't have an `emit_ast` parameter, the logic would be:

```python
    if enable_snowpark_ir:
        ...
```

Functions that have an `emit_ast` parameter can do one of:

```python
    if enable_snowpark_ir:
        if _emit_ast:
            ...
```

This variant would be preferable if the two `if` statements result in distinct `else` clauses. If both `else` clauses do nothing, simpler variants will work.

```python
    if enable_snowpark_ir and _emit_ast:
        ...
```

```python
    _emit_ast = _emit_ast and enable_snowpark_ir
    if _emit_ast:
        ...
```

sfc-gh-lspiegelberg: Thanks for providing some more in-depth context here! I'd actually say it'd be good to discuss this in today's standup. For the scope of this PR, it sounds to me that we're aligned that adding a `_emit_ast` argument similar to the existing APIs would no matter which route we go for the feature flag be a good idea. I'd therefore suggest to for now simply use `_emit_ast` without a double-purpose in mind (i.e., simply to emit/not emit AST) and circle back on the feature flag in the standup. That would allow to land this PR, and not introduce other, new concepts with it. Wdyt? Is this a reasonable approach for this PR?

sfc-gh-oplaton: Sure, I can do that.

tests/ast/data/DataFrame.unpivot.test.DISABLED
sfc-gh-oplaton: Filed https://snowflakecomputing.atlassian.net/browse/SNOW-1663699

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: Done.

sfc-gh-lspiegelberg (approved):
Great work!

PR #2290, opened 2024-09-12
SNOW-1657460 Improved  results for TIMESTAMP_TZ type to show correct timezone offset.

sfc-gh-azhan:

   

   Fixes SNOW-1657460

Previously we use "datetime64[ns]" as the dtype for datetime with timezone which is missing the tz info. This change will extract the timezone from timestamp_tz column and show the timezone offset when calling dtypes.

### DIFF
--- a/src/snowflake/snowpark/modin/pandas/indexing.py
+++ b/src/snowflake/snowpark/modin/pandas/indexing.py
@@ -51,11 +51,9 @@ from pandas._typing import AnyArrayLike, Scalar
 from pandas.api.types import is_bool, is_list_like
 from pandas.core.dtypes.common import (
     is_bool_dtype,
-    is_datetime64_any_dtype,
     is_integer,
     is_integer_dtype,
     is_numeric_dtype,
-    is_timedelta64_dtype,
     pandas_dtype,
 )
 from pandas.core.indexing import IndexingError
@@ -846,7 +844,7 @@ class _LocIndexer(_LocationIndexerBase):
             period = pd.Period(parsed, freq=reso.attr_abbrev)
 
             # partial string indexing only works for DatetimeIndex
-            if is_datetime64_any_dtype(self.df._query_compiler.index_dtypes[0]):
+            if self.df._query_compiler.is_datetime64_any_dtype(idx=0, is_index=True):
                 return slice(
                     pd.Timestamp(period.start_time, tzinfo=tzinfo),
                     pd.Timestamp(period.end_time, tzinfo=tzinfo),
@@ -927,7 +925,7 @@ class _LocIndexer(_LocationIndexerBase):
         row_loc = self._try_partial_string_indexing(row_loc)
 
         # Check if self or its index is a TimedeltaIndex. `index_dtypes` retrieves the dtypes of the index columns.
-        if is_timedelta64_dtype(self.df._query_compiler.index_dtypes[0]):
+        if self.df._query_compiler.is_timedelta64_dtype(idx=0, is_index=True):
             # Convert row_loc to timedelta format to perform exact matching for TimedeltaIndex.
             row_loc = self._convert_to_timedelta(row_loc)
 
--- a/src/snowflake/snowpark/modin/pandas/indexing.py
+++ b/src/snowflake/snowpark/modin/pandas/indexing.py
@@ -924,7 +924,7 @@ class _LocIndexer(_LocationIndexerBase):
         row_loc, col_loc = self._parse_get_row_and_column_locators(key)
         row_loc = self._try_partial_string_indexing(row_loc)
 
-        # Check if self or its index is a TimedeltaIndex. `index_dtypes` retrieves the dtypes of the index columns.
+        # Check if self or its index is a TimedeltaIndex.
         if self.df._query_compiler.is_timedelta64_dtype(idx=0, is_index=True):
             # Convert row_loc to timedelta format to perform exact matching for TimedeltaIndex.
             row_loc = self._convert_to_timedelta(row_loc)
--- a/src/snowflake/snowpark/modin/plugin/_internal/type_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/type_utils.py
@@ -1,7 +1,7 @@
 #
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
-from functools import reduce
+from functools import lru_cache, reduce
 from typing import Any, Callable, Union
 
 import numpy as np
@@ -31,6 +31,7 @@ from pandas.core.dtypes.common import is_bool_dtype, is_float_dtype, is_integer_
 
 from snowflake.snowpark import Column
 from snowflake.snowpark._internal.type_utils import infer_type, merge_type
+from snowflake.snowpark.dataframe import DataFrame as SnowparkDataFrame
 from snowflake.snowpark.functions import (
     builtin,
     cast,
@@ -39,6 +40,7 @@ from snowflake.snowpark.functions import (
     floor,
     iff,
     length,
+    to_char,
     to_varchar,
     to_variant,
 )
@@ -446,3 +448,23 @@ def is_compatible_snowpark_types(sp_type_1: DataType, sp_type_2: DataType) -> bo
     if isinstance(sp_type_1, StringType) and isinstance(sp_type_2, StringType):
         return True
     return False
+
+
+@lru_cache
+def _get_timezone_from_timestamp_tz(
+    snowpark_dataframe: SnowparkDataFrame, snowflake_quoted_identifier: str
+) -> Union[str, DatetimeTZDtype]:
+    tz_df = (
+        snowpark_dataframe.filter(col(snowflake_quoted_identifier).is_not_null())
+        .select(to_char(col(snowflake_quoted_identifier), format="TZHTZM").as_("tz"))
+        .group_by(["tz"])
+        .agg()
+        .limit(2)
+        .to_pandas()
+    )
+    assert (
+        len(tz_df) > 0
+    ), f"col {snowflake_quoted_identifier} does not contain valid timezone offset"
+    if len(tz_df) == 2:  # multi timezone cases
+        return "object"
+    return DatetimeTZDtype(tz="UTC" + tz_df.iloc[0, 0].replace("Z", ""))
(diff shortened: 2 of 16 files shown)


tests/integ/modin/series/test_astype.py
```diff
@@ -286,7 +287,7 @@ def test_astype_from_DatetimeTZDtype_to_datetime64(from_tz):
     native = native_pd.Series([0, 1, 2, 3], dtype=from_dtype)
     snow = pd.Series(native)
     expected_dtype = get_expected_dtype(to_dtype)
-    with SqlCounter(query_count=1):
+    with SqlCounter(query_count=3):
```
sfc-gh-helmeleegy, line 290: Why is the increase in the query count?

sfc-gh-azhan: the code below called `.astype` twice which trigger the query to get tz each time.

sfc-gh-azhan: I could keep improving this in a follow up PR.

sfc-gh-azhan: actually this can be an easy fix. let me do it now.

sfc-gh-helmeleegy: Much better. Thanks!

sfc-gh-helmeleegy (approved):
LGTM, thanks Andong!

src/snowflake/snowpark/modin/pandas/indexing.py
```diff
@@ -927,7 +925,7 @@ def __getitem__(
         row_loc = self._try_partial_string_indexing(row_loc)
 
         # Check if self or its index is a TimedeltaIndex. `index_dtypes` retrieves the dtypes of the index columns.
```
sfc-gh-rdurrani, line 927: Comment needs to be updated to reflect change.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -487,18 +488,19 @@ def dtypes(self) -> native_pd.Series:
...
+            else TypeMapper.to_pandas(t)
+            for i, t in type_map.items()
         ]
 
-        from snowflake.snowpark.modin.pandas.utils import try_convert_index_to_native
-
         return native_pd.Series(
             data=types,
-            index=try_convert_index_to_native(self._modin_frame.data_columns_index),
+            index=self._modin_frame.data_columns_index,
```
sfc-gh-rdurrani, line 503: Why do we no longer need the try_convert_index...? I'm assuming that its inclusion here previously was a bug, since the data_columns_index should always be a native index right?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -512,13 +514,65 @@ def index_dtypes(self) -> list[Union[np.dtype, ExtensionDtype]]:
...
+        )
         return [
-            TypeMapper.to_pandas(t)
-            for t in self._modin_frame.get_snowflake_type(
-                self._modin_frame.index_column_snowflake_quoted_identifiers
-            )
+            self._modin_frame.get_datetime64tz_from_timestamp_tz(i)
+            if t == TimestampType(TimestampTimeZone.TZ)
+            else TypeMapper.to_pandas(t)
+            for i, t in type_map.items()
```
sfc-gh-rdurrani, line 524: Perhaps make into a function, since the same code is used above as well?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -512,13 +514,65 @@ def index_dtypes(self) -> list[Union[np.dtype, ExtensionDtype]]:
...
         ]
 
+    def is_timestamp_type(self, idx: int, is_index: bool = True) -> bool:
+        """Return True if index is TIMESTAMP TYPE.
```
sfc-gh-rdurrani, line 528:
```suggestion
        """Return True if column at index is TIMESTAMP TYPE.
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -512,13 +514,65 @@ def index_dtypes(self) -> list[Union[np.dtype, ExtensionDtype]]:
...
+        )
+
+    def is_datetime64_any_dtype(self, idx: int, is_index: bool = True) -> bool:
+        """Helper method similar to is_datetime64_any_dtype, but it avoids extra query for DatetimeTZDtype.
+
+        Args:
+            idx: the index of the column
+            is_index: whether it is an index or data column
+        """
+        return self.is_timestamp_type(idx, is_index) or is_datetime64_any_dtype(
```
sfc-gh-rdurrani, line 550: Should we reverse the order of the operands in the `or`, so we can shortcut if the dtypes caches are filled, and not have the query from get_snowflake_type?

tests/integ/modin/frame/test_dtypes.py
```diff
@@ -373,21 +373,27 @@ def test_insert_multiindex_multi_label(label1, label2):
...
+        else 1
     )
-    assert roundtripped.dtype == expected.dtype
+    with SqlCounter(query_count=qc):
+        created = pd.Series(dataframe_input, dtype=input_dtype)
+        # For snowpark pandas type mapping
+        assert created.dtype == logical_dtype
+        roundtripped = created.to_pandas()
+        assert_series_equal(
+            roundtripped, expected, check_dtype=False, check_index_type=False
```
sfc-gh-rdurrani, line 394: Can't we just set `check_dtype=True`, and avoid the below line?

tests/integ/modin/series/test_astype.py
```diff
@@ -218,11 +219,11 @@ def test_astype_to_DatetimeTZDtype(from_dtype, to_tz):
                 # use dypte=str instead of StringDType
                 native_pd.Series(seed, dtype=str).astype(to_dtype)
     else:
-        with SqlCounter(query_count=1):
+        with SqlCounter(query_count=2):
             s = pd.Series(seed, dtype=from_dtype).astype(to_dtype)
             # Snowflake timestamp_tz column's metadata does not contain the tzinfo so it cannot provide dtype as
```
sfc-gh-rdurrani, line 224: Isn't this comment no longer true?

sfc-gh-rdurrani (commented):
left some questions/nits, but overall lgtm!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-azhan: This was not necessary because data_columns_index always return a native index.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-azhan: Probably we only need the first part. Let me try.

sfc-gh-azhan: It works.

src/snowflake/snowpark/modin/plugin/_internal/type_utils.py
```diff
@@ -446,3 +448,23 @@ def is_compatible_snowpark_types(sp_type_1: DataType, sp_type_2: DataType) -> bo
...
+@lru_cache
+def _get_timezone_from_timestamp_tz(
+    snowpark_dataframe: SnowparkDataFrame, snowflake_quoted_identifier: str
+) -> Union[str, DatetimeTZDtype]:
+    tz_df = (
+        snowpark_dataframe.filter(col(snowflake_quoted_identifier).is_not_null())
+        .select(to_char(col(snowflake_quoted_identifier), format="TZHTZM").as_("tz"))
+        .group_by(["tz"])
+        .agg()
+        .limit(2)
```
sfc-gh-nkrishna, line 462: Why limit 2 here? Can you add a comment?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -512,12 +528,59 @@ def index_dtypes(self) -> list[Union[np.dtype, ExtensionDtype]]:
...
                 self._modin_frame.index_column_snowflake_quoted_identifiers
-            )
-        ]
+                if is_index
+                else self._modin_frame.data_column_snowflake_quoted_identifiers
+            )[idx],
+            TimestampType,
+        )
+
+    def is_datetime64_any_dtype(self, idx: int, is_index: bool = True) -> bool:
```
sfc-gh-nkrishna, line 551: Maybe move this functions and the other similar ones into timestamp_utils?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -477,6 +478,28 @@ def wrap(*args, **kwargs):  # type: ignore
...
+        """
+        type_map = self._modin_frame.quoted_identifier_to_snowflake_type(
+            snowflake_quoted_identifiers
+        )
+        return [
+            self._modin_frame.get_datetime64tz_from_timestamp_tz(i)
+            if t == TimestampType(TimestampTimeZone.TZ)
+            else TypeMapper.to_pandas(t)
+            for i, t in type_map.items()
+        ]
```
sfc-gh-nkrishna, line 501: Do you think this function (and `get_datetime64tz_from_timestamp_tz`) can be incorporated into `cached_data_column_snowpark_pandas_types`? They seem somewhat related to me.

sfc-gh-azhan: This PR is not able to do that now. This cache here is on the snowpark dataframe or sql layer. It cache the result if the snowpark dataframe/sql doesn't change. 

To be able to do cache in `cached_data_column_snowpark_pandas_types`, we have to track all cases that may change timezone.

sfc-gh-rdurrani (approved):
LGTM!

sfc-gh-nkrishna (approved):
LGTM, nice work!

PR #2295, opened 2024-09-16
[SNOW-1625268] Add AST merge gate for ast phase0.

sfc-gh-lspiegelberg:
Adds merge gate to run ast expectation tests as part of Github actions workflow.

Other:
- Fixes unboundlocalerror for pivot.
- Changes ast expectation tests to run client language independent.
- Unifies AST timezone encoding for `datetime.datetime`/`datetime.time`.
- Provides fix for windows test by injecting timezone via environment. Though not working because of Windows issues, this allows to have a similar test environment as in linux/mac.


tests/ast/test_ast_driver.py
```diff
@@ -28,6 +31,7 @@ class TestCase:
     source: str
     expected_ast_base64: str
     expected_ast_unparsed: str
+    __test__: bool = False  # Add this to suppress pytest collection warning
```
sfc-gh-azwiegincew, line 34: Set this to `True` ...; also period at the end.

sfc-gh-lspiegelberg: It's purposefully set to False, this is because pytest will assume any class starts with `Test` is a testcase (https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#conventions-for-python-test-discovery). This behavior can be suppressed with `__test__:bool = False` which will get rid off a warning that this class does not represent a valid test case (in the pytest way).

tests/integ/conftest.py
```diff
@@ -329,7 +329,7 @@ def _notify(self, query_record: QueryRecord, *args, **kwargs) -> None:
 # Disable phase0 Snowpark IR check by setting autouse=False,
 # by default enabled for all tests (except modin/) to check that Snowpark python requests contain
 # the dataframeAst property holding the encoded representation.
```
sfc-gh-lspiegelberg, line 331: this will need to get fixed in another PR. deactivating for now, so the integ tests are run.

sfc-gh-azwiegincew (approved):
Thanks for doing this!

PR #2297, opened 2024-09-16
Remove phase 1 code

sfc-gh-azwiegincew:
Phase 1 client now lives in the branch `ast-phase-1`.


src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -367,14 +367,9 @@ def notify_query_listeners(self, query_record: QueryRecord, **kwargs) -> None:
     def execute_and_notify_query_listener(
         self, query: str, **kwargs: Any
     ) -> SnowflakeCursor:
-
         results_cursor = self._cursor.execute(query, **kwargs)
```
sfc-gh-lspiegelberg, line 370: I think we need to keep these changes here. We will use them to track that in phase0 dataframeAst is part of the query. Can we revert the QueryRecord/listener changes?

sfc-gh-oplaton (approved):

PR #2302, opened 2024-09-17
SNOW-1663726 make session config updates thread safe

sfc-gh-aalam:

   

   Fixes SNOW-1663726

   In this PR we make the following changes:
   - We protect all session config variable updates using a lock
   - We create a per thread `_plan_builder` to isolate usage.
   - Created `ConfigContext` to set a context and read a snapshot of config values from session. Use the `ConfigContext` object for `Analyzer`, `PlanCompiler` and `SnowflakePlanBuilder`.
     - created methods `do_analyzer` and `_compile` that runs using a single `ConfigContext`


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -452,7 +451,7 @@ def cumulative_node_complexity(self, value: Dict[PlanNodeCategory, int]):
         self._cumulative_node_complexity = value
 
     def __copy__(self) -> "SnowflakePlan":
-        if self.session._cte_optimization_enabled:
+        if self.cte_optimization_enabled:
```
sfc-gh-aalam, line 454: this seems wrong

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -166,18 +167,33 @@ def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
...
+        self.plan_builder.config_context = self.config_context
 
     def analyze(
         self,
         expr: Union[Expression, NamedExpression],
         df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
         parse_local_name=False,
+    ) -> str:
+        # Set the config context for analysis step
+        with self.config_context:
```
sfc-gh-yzou, line 181: I see analyze is also called during resolve stage, and we create one snapshot when WithStatement is used, does that mean the config snapshot can change during the resolve stage?

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -166,18 +167,33 @@ def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
...
+            return self.do_analyze(
+                expr, df_aliased_col_name_to_real_col_name, parse_local_name
+            )
+
+    def do_analyze(
```
sfc-gh-yzou, line 186: another thing here is that caller can still call do_analyze independently if they want, and this may not have a snapshot

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -166,18 +167,33 @@ def __init__(self, session: "snowflake.snowpark.session.Session") -> None:
...
+        self,
+        expr: Union[Expression, NamedExpression],
+        df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
```
sfc-gh-yzou, line 189: one way i can think about is to let every function that you want to protect with take a configContext parameter, and let session provide a getConfigSnapshot capability.
in terms code change, i think it is probably similar as what you have here. 

When i look at the problem, the issue seems very easy to address during the new compilation stage, because the code generator is create per compilation. However, for the general analyzer, it is one per session, which makes things little bit harder.

If we don't do any snapshot here, and do not consider about the old cte optimization. is the problem only happens during the plan building for the new compilation? if that is the case, maybe we can handle the new compilation stage, and put a rule in the future that all implementation has to be thread safe.

src/snowflake/snowpark/modin/plugin/_internal/generator_utils.py
```diff
@@ -225,10 +225,11 @@ def generate_irregular_range(
 
     # TODO: SNOW-1646883 fix invalid identifier error when sql_simplifier_enabled is True
     session = get_active_session()
-    sql_simplifier_enabled = session.sql_simplifier_enabled
-    session.sql_simplifier_enabled = False
-    num_offsets = session.range(start=0, end=periods, step=1)
-    session.sql_simplifier_enabled = sql_simplifier_enabled
+    with session._lock:
+        sql_simplifier_enabled = session.sql_simplifier_enabled
+        session.sql_simplifier_enabled = False
```
sfc-gh-yzou, line 230: hmm, I thought we are putting the lock on the sql_simplfier_enable function, if we do that, do we still need to do the change here?

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-aalam: I updated call from `analyze` -> `do_analyze` in `resolve`. I created a snapshot in `resolve` using with statement. Since a single analyzer shared the same `config_context`, `do_analyze` uses the same context during resolve. same goes for `plan_builder`

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-aalam: `do_analyze` is only called from within Analyzer similar to `do_resolve`. I can convert them into `_do_analyze` to make this a private method to ensure a snapshot is created by calling `analyze()`

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -518,6 +523,56 @@ def add_aliases(self, to_add: Dict) -> None:
...
+            "large_query_breakdown_complexity_bounds",
+            "large_query_breakdown_enabled",
+        }
+
+    def __getattr__(self, name: str) -> Any:
+        if name in self.configs:
+            return getattr(self.session, name)
+        raise AttributeError(f"ConfigContext has no attribute {name}")
+
+    def __enter__(self) -> "ConfigContext":
```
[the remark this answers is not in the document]
sfc-gh-aalam: can you share an example of called it without with statement. One can call `context_config.__enter__()` and `context_config.__exit__()` and they will behave as any regular method call. But the purpose of these methods it to be used inside a context to call `__enter__` and `__exit__` methods is a less verbose manner

src/snowflake/snowpark/modin/plugin/_internal/generator_utils.py
sfc-gh-aalam: this is done so that the value stays constant during `session.range` invocation. This was the intention of the old code implementation and lock just holds the same guarantee.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-aalam: > However, for the general analyzer, it is one per session, which makes things little bit harder.

We have one session per-thread. So the config context snapshot we are taking here is taken in each thread.

> if that is the case, maybe we can handle the new compilation stage, and put a rule in the future that all implementation has to be thread safe.

I do not understand this completely. Do you mean we handle plan builder same way we deal with new compilation stage?

src/snowflake/snowpark/_internal/analyzer/config_context.py
```diff
@@ -0,0 +1,37 @@
...
+
+    On instantiation, this object stores the configuration from the session object
+    and returns the stored configuration attributes when requested.
+    """
+
+    def __init__(self, session) -> None:
+        self.session = session
+        self.configs = {
+            "_query_compilation_stage_enabled",
+            "cte_optimization_enabled",
```
sfc-gh-yzou, line 19: instead of doing this, it might be more clear to provide a property interface in the session that teaks a snapshot. so it is the interface is at a closer place with the properties.

Given the fact that we are adding more and more configs,  i am thinking it might worth to do a refactoring to centralize all configs, and wrap all of them into a class to reduce the possiblility for developer make mistakes.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -531,6 +542,10 @@ def __init__(
         # on the optimized plan. During the final query generation, no schema query is needed,
         # this helps reduces un-necessary overhead for the describing call.
         self._skip_schema_query = skip_schema_query
+        self._config_context: Optional[ConfigContext] = None
```
sfc-gh-yzou, line 545: The config context here should probably been done in the same way as what we have done for analyze and resolve as parameter for each builder.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -564,9 +579,10 @@ def build(
             ), "No schema query is available in child SnowflakePlan"
             new_schema_query = schema_query or sql_generator(child.schema_query)
 
+        config_context = self._config_context or ConfigContext(self.session)
```
sfc-gh-yzou, line 582: each build function seems can take a new snapshot, but multiple build can be called during the whole resolve, and they are suppose to get the same snapshot, it is safer and cleaner to simply extend each interface to take the config context.

src/snowflake/snowpark/_internal/compiler/plan_compiler.py
```diff
@@ -46,6 +47,7 @@ class PlanCompiler:
 
     def __init__(self, plan: SnowflakePlan) -> None:
         self._plan = plan
+        self._config_context = ConfigContext(self._plan.session)
```
sfc-gh-yzou, line 50: That is good, we are only using one snapshot during the whole compilation stage

src/snowflake/snowpark/modin/plugin/_internal/generator_utils.py
sfc-gh-yzou: that code here is hacky and bad. do you know what is the ticket SNOW-1646883 referring for, i think we should fix that and remove this code, we are not suppose to change the config in our code like this.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: I intend to remove `config_context` from plan_builder once old CTE implementation is gone. That is the only reason why plan builder needs this info. I can own this task. Without that, the refactor will be too big and it would be prone to more errors. I want to avoid that.

src/snowflake/snowpark/modin/plugin/_internal/generator_utils.py
sfc-gh-aalam: I'll fix this bug and remove the hack

sfc-gh-aalam: seems like it is fixed: I removed the hack here: https://github.com/snowflakedb/snowpark-python/pull/2375

src/snowflake/snowpark/_internal/compiler/plan_compiler.py
```diff
@@ -49,6 +49,13 @@ class PlanCompiler:
 
     def __init__(self, plan: SnowflakePlan) -> None:
         self._plan = plan
+        session = plan.session
+        self.cte_optimization_enabled = session.cte_optimization_enabled
```
sfc-gh-yzou, line 53: if we are restricting the change to config runtime, this is not necessary anymore, right?

sfc-gh-aalam: we are only putting restriction on `cte_optimization_enabled`. Since for other, we still need to take a snapshot, I'm taking a snapshot for cte param for completeness.

src/snowflake/snowpark/session.py
```diff
@@ -827,12 +868,18 @@ def large_query_breakdown_enabled(self, value: bool) -> None:
         materialize the partitions, and then combine them to execute the query to improve
         overall performance.
         """
+        if threading.active_count() > 1:
```
sfc-gh-yzou, line 871: you can probably have a small utility function and call that for all configs with the config name as a parameter

src/snowflake/snowpark/_internal/compiler/plan_compiler.py
sfc-gh-yzou: since now we are putting protection on all config variables, i think we don't need to do that anymore

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -297,6 +298,15 @@ def normalize_path(path: str, is_local: bool) -> str:
     return f"'{path}'"
 
 
+def warn_session_config_update_in_multithreaded_mode(config) -> None:
+    if threading.active_count() > 1:
+        logger.warning(
+            "You might have more than one threads sharing the Session object trying to update "
+            f"{config}. This is currently not thread-safe and may cause unexpected behavior. "
```
sfc-gh-yzou, line 305: the sense here sounds little bit wired, and updating the config its-self is thread safe, it just make the other operations unsafe. maybe say updating f"{config}" while other tasks are running can potentially cause unexpected behavior.

src/snowflake/snowpark/session.py
```diff
@@ -637,6 +639,16 @@ def _analyzer(self) -> Analyzer:
             )
         return self._thread_store.analyzer
 
+    @property
+    def _plan_builder(self):
+        if not hasattr(self._thread_store, "plan_builder"):
```
sfc-gh-yzou, line 644: oh, i didn't notice we have plan builder here, i thought plan builder is just part of the analyzer, what is the plan builder used here?, and what is the reason that we moved it to a property

sfc-gh-yzou: and the plan builder seems thread local to me

src/snowflake/snowpark/_internal/compiler/plan_compiler.py
```diff
@@ -151,8 +154,7 @@ def compile(self) -> Dict[PlanQueryType, List[Query]]:
             return queries
         else:
             final_plan = self._plan
-            if self._plan.session.cte_optimization_enabled:
-                final_plan = final_plan.replace_repeated_subquery_with_cte()
+            final_plan = final_plan.replace_repeated_subquery_with_cte()
```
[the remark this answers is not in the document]
sfc-gh-aalam: yes. `replace_repeated_subquery_with_cte()` does this check automatically.

src/snowflake/snowpark/session.py
sfc-gh-aalam: This plan builder is different from analyzer's plan builder. Only used by
1. DataFramReader to create read_file plan
2. FileOperation to create file_operation_plan

sfc-gh-aalam: I added it as part of old changes. reverted back to original now since plan builder should be thread-safe with restricted config updates

src/snowflake/snowpark/_internal/utils.py
sfc-gh-aalam: updated

sfc-gh-yzou (approved):

PR #2313, opened 2024-09-17
SNOW-1665697: add open_new_result in document API

sfc-gh-zhan:
Description
add open_new_result in document API, https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/1.8.0/api/snowflake.snowpark.files.SnowflakeFile.open

Testing
add unit test

   Fixes SNOW-1665697


sfc-gh-zhan:

src/snowflake/snowpark/files.py
```diff
@@ -122,6 +124,20 @@ def isatty(self) -> None:
         """
         raise NotImplementedError(_DEFER_IMPLEMENTATION_ERR_MSG)
 
+    @classmethod
+    @private_preview(version="1.22.1")
+    def open_new_result(cls, mode: str = "w") -> SnowflakeFile:
```
sfc-gh-lschmidt, line 129: Do you think we need a * here? Would that give us more flexibility in the future? Maybe we can check with python folks?

sfc-gh-zhan: verified with Afroz, it will show if version >= 1.22.1. and the version is getting from [version.py](https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/version.py)

sfc-gh-lschmidt: No, I mean a * in the function definition. Like in the others.

sfc-gh-zhan: Named arguments must follow bare *, I think for now since mode is a required arguments we can leave it here. once we have some unrequired arguments, we can update it like `(cls, mode, *, [arguments])`, similar as current `open(cls, ..., *,...)`

sfc-gh-lschmidt: hrm okay, it is probably okay then. I think we monkey patch the other one after this one, so it should still work. Thanks for checking.

sfc-gh-lschmidt (approved):

src/snowflake/snowpark/files.py
```diff
@@ -122,6 +125,20 @@ def isatty(self) -> None:
...
+        """
+        Returns a :class:`~snowflake.snowpark.file.SnowflakeFile`.
+        In UDF and Stored Procedures, the object works like a Python IOBase object and as a wrapper for an IO stream of remote files. The IO Stream is to support the file operations defined in this class.
+
+        You can use this particular scoped URL to access files from within Snowflake (through another UDF or the COPY FILES command), but not from outside of Snowflake as a pre-signed URL. The scoped URL is valid for 24 hours.
+
+        Args:
+            mode: A string used to mark the type of an IO stream.
+        """
+        return cls("new results file", mode, require_scoped_url=0, from_result_api=True)
```
sfc-gh-aalam, line 140: if we always write the result in one location - `"new results file"`, we should probably mention that in the documentation.

sfc-gh-aalam (approved):

src/snowflake/snowpark/files.py
sfc-gh-zhan: yes, but it's more like a placeholder for the file_uri. also user should read/write file through scoped url, maybe it's better to hide this info to user?

sfc-gh-lschmidt: Hrm, reading through the comments. The user of this class doesn't need to do anything with scoped URLs. Probably would be best not to mention that and instead do something like:

"This stream will open a writable result file that will be materialized when it's returned by a UDF or Stored Procedure. When the file is materialized then file_uri will be set to a scoped URL that temporarily references this file."

sfc-gh-zhan:
Hi @sfc-gh-aalam looks like I'm not authorized to merge, could you help to merge it? thanks!

PR #2315, opened 2024-09-18
[SIT-2372] Add the `DataFrame.map` method

sfc-gh-lfallasavendano:
Adds the `DataFrame.map` method which applies a Python function to every element of a DataFrame.

   Fixes [SIT-2372](https://snowflakecomputing.atlassian.net/browse/SIT-2372)

   Adds the `DataFrame.map` method that applies a Python function to every row of a DataFrame.

[SIT-2372]: https://snowflakecomputing.atlassian.net/browse/SIT-2372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfc-gh-lfallasavendano:

sfc-gh-mvega (approved):
lgtm

sfc-gh-yixie:
@sfc-gh-aalam could you help review this PR?

src/snowflake/snowpark/dataframe.py
```diff
@@ -4188,6 +4189,117 @@ def _explain_string(self) -> str:
 
         return f"{msg}\n--------------------------------------------"
 
+    def map(
+        self,
+        func: Callable,
+        output_types: list[StructType],
+        output_column_names: Optional[list[str]] = None,
```
sfc-gh-yixie, line 4196: I suggest adding a `*` before this parameter to make these optional parameters key-word arguments.

src/snowflake/snowpark/dataframe.py
```diff
@@ -4188,6 +4189,117 @@ def _explain_string(self) -> str:
 
         return f"{msg}\n--------------------------------------------"
 
+    def map(
```
sfc-gh-yixie, line 4192: A UDTF call may have other parameters like partitioning.

src/snowflake/snowpark/dataframe.py
sfc-gh-lfallasavendano: Thanks for your suggestion.

I just removed the `extra_packages` parameter and added a `**kwargs` parameter to pass arguments to the UTDF registration (as with other similar methods).

src/snowflake/snowpark/dataframe.py
sfc-gh-lfallasavendano: As part of the previous comment, the`**kwargs` parameter can be used to pass arguments to the UDTF registration.

sfc-gh-aalam: @sfc-gh-lfallasavendano A more fundamental question here is why are we using UDTF to implement map. `.map` is supposed to give 1-1 input -> output map for each row. UDF is more aligned to do this.

tests/integ/scala/test_datatype_suite.py
```diff
@@ -381,7 +381,7 @@ def test_structured_dtypes_select(structured_type_session, examples):
     query, expected_dtypes, expected_schema = examples
     df = _create_test_dataframe(structured_type_session)
     flattened_df = df.select(
-        df.map["k1"].alias("value1"),
+        df.Map["k1"].alias("value1"),
```
sfc-gh-aalam, line 384: this change implies any customers who are using this way to access column names `map` will break. This is a breaking change then.
cc: @sfc-gh-yixie

src/snowflake/snowpark/dataframe.py
sfc-gh-lfallasavendano: @sfc-gh-aalam  the reason for using UDTFs instead of UDFs is to support scenarios where non-scalar values are returned from the function that is being applied. For example:

```
df = session.create_dataframe([ [10,'a'], [20,'b'], [30,'c'] ], 
                              schema=StructType([StructField('N_ID', IntegerType()), StructField('NAME', StringType())]))

# Returning rows
df.map(lambda r: Row("---" + r.NAME + "---", r.N_ID), output_types=[StringType(), IntegerType()]).show()
-------------------
|"C_1"    |"C_2"  |
-------------------
|---c---  |30     |
|---a---  |10     |
|---b---  |20     |
-------------------


# Returning tuples:
def do_something(row):
    return ("---" + row.NAME + "---", row.N_ID)
df.map(do_something, output_types=[StringType(), IntegerType()]).show()
-------------------
|"C_1"    |"C_2"  |
-------------------
|---c---  |30     |
|---a---  |10     |
|---b---  |20     |
-------------------
```

tests/integ/scala/test_datatype_suite.py
sfc-gh-lfallasavendano: @sfc-gh-aalam  perhaps we can change the name of this new method to avoid this situation. For example, we could use `apply_func`. Do you agree with this alternative?

PR #2316, opened 2024-09-18
SNOW-1642189: collect telemetry about concurrency usage

sfc-gh-aalam:

   

   Fixes SNOW-1642189

   In this PR, we associate thread id to with collect and file upload APIs. The number of unique thread ids in a session can then be used to gauge the number of users using this feature.


src/snowflake/snowpark/_internal/telemetry.py
```diff
@@ -341,6 +345,7 @@ def send_upload_file_perf_telemetry(
                 TelemetryField.KEY_CATEGORY.value: TelemetryField.PERF_CAT_UPLOAD_FILE.value,
                 TelemetryField.KEY_FUNC_NAME.value: func_name,
                 TelemetryField.KEY_DURATION.value: duration,
+                TelemetryField.THREAD_IDENTIFIER.value: threading.get_ident(),
```
sfc-gh-aling, line 348: I see we have other telemetry method like `send_function_usage_telemetry`, `send_session_created_telemetry`, do we need to set this field in other methods?

sfc-gh-aalam: actually you bring a great point. I don't think we should update other telemtery but we can add a new telemetery. I can create a new telemetry function `send_new_thread_created` whenever we create a new cursor. This will not be exactly accurate but will give us close to accurate telemetry.

sfc-gh-jdu (approved):

sfc-gh-aling (approved):

sfc-gh-helmeleegy (approved):
LGTM.

PR #2317, opened 2024-09-18
SNOW-1652608: Fix column labels when transposing frame with Timedelta index levels.

sfc-gh-mvashishtha:
Fixes SNOW-1652608

Before this PR, using `loc` to get a single row of a dataframe with a `Timedelta` label gave a series with an integer name instead of with a `Timedelta` name, e.g. `pd.DataFrame(['a'], index=[pd.Timedelta(1)]).loc[pd.Timedelta(1)]`. This PR fixes that bug as well.


tests/integ/modin/frame/test_loc.py
```diff
@@ -3933,17 +3933,11 @@ def test_raise_set_cell_with_list_like_value_error():
             "1 day",
             2,
             3,
-            marks=pytest.mark.xfail(
-                reason="SNOW-1652608 result series name incorrectly set"
```
sfc-gh-helmeleegy, line 3937: Are we doing a transpose operation as part of processing `loc`?
I wasn't clear how is this related to fixing a bug in transpose.

src/snowflake/snowpark/modin/plugin/_internal/transpose_utils.py
```diff
@@ -108,6 +111,73 @@ def prepare_and_unpivot_for_transpose(
...
+    is a SnowparkPandasType, the intermediate transpose result for each column
+    uses the Snowpark representation of the row label rather than the Snowpark
+    pandas representation. For example, if a row has pandas label
+    pd.Timedelta(7), then that row's label in Snowpark is the number 7, so the
+    intermediate transpose result would have a column named 7 instead of
+    pd.Timedelta(7). This method uses the index types of the original frame to
+    fix the pandas labels of column levels that come from SnowparkPandasType
+    index levels.
+
+    Examples
```
sfc-gh-helmeleegy, line 131: I think it would be better to move the examples to the end of the docstring after the argument and returned value are explained.

sfc-gh-helmeleegy (approved):
LGTM overall. Just a couple of minor nits/questions. Thanks for fixing this bug.

tests/integ/modin/frame/test_transpose.py
```diff
@@ -61,6 +62,42 @@ def test_dataframe_transpose_set_single_index(
         )
 
 
+@pytest.mark.parametrize(
+    "transpose_operation, expected_query_count",
```
sfc-gh-lmukhopadhyay, line 66: nit- why parameterize on expected_query_count if both examples are 1, instead of `@sql_count_checker(query_count=1)`?

sfc-gh-lmukhopadhyay (approved):
Lgtm, just 1 nit!

tests/integ/modin/frame/test_loc.py
sfc-gh-mvashishtha: I added an explanation in the PR description.

tests/integ/modin/frame/test_transpose.py
sfc-gh-mvashishtha: Fixed.

I copied this code from `transpose_and_double_transpose_parameterize`, which has the redundant query count, and modified the `marks` for one of the test cases.

PR #2326, opened 2024-09-19
[SNOW-1632896] Detect valid condition for flatten nested select projection complexity

sfc-gh-yzou:

   

NOW-1632896

snowflake optimization can merge the select statement when possible, to get a more accurate estimation of the query plan complexity, we want to detect such case and adjust the complexity calculation.
For example, SELECT COL1 + 2 AS COL1, COL2 FROM (SELECT COL1 + 1 AS COL1, COL2 FROM TEST_TABLE) can be merged as SELECT (COL1  + 1) + COL1 + 2 AS COL1, COL2 FROM TEST_TABLE.

A parameter _try_merge_projection_complexity is added to SelectStatement to indicate that we should try to merge the select projection complexity with the child (from).


src/snowflake/snowpark/_internal/select_projection_complexity_utils.py
```diff
@@ -0,0 +1,235 @@
...
+# Be careful when adding functions in this list, it must statisfy the following condition:
+# 1) not aggregation
+# 2) no data generation
+# 3) not nondeterministic with unique reference,
+VALID_PROJECTION_MERGE_FUNCTIONS = (
+    "add_months",
+    "bitnot",
+    "bitshiftleft",
+    "bitshiftright",
+    "bround",
```
sfc-gh-aalam, line 25: would suggest we sort this

src/snowflake/snowpark/_internal/select_projection_complexity_utils.py
```diff
@@ -0,0 +1,235 @@
...
+from snowflake.snowpark._internal.analyzer.expression import (
+    Expression,
+    FunctionExpression,
+)
+from snowflake.snowpark._internal.analyzer.table_function import TableFunctionExpression
+from snowflake.snowpark._internal.analyzer.window_expression import WindowExpression
+
+# Set of functions (defined in snowpark-python/src/snowflake/snowpark/functions.py) that do not
+# block merge projections of nested select.
+# Be careful when adding functions in this list, it must statisfy the following condition:
```
sfc-gh-jdu, line 16: How do we maintain this list, if in the future we add such a function meeting this requirements? Do we expect developers to add it here

src/snowflake/snowpark/_internal/valid_view_merge_functions.py
```diff
@@ -0,0 +1,191 @@
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
+# aggregation, data generation or nondeterministic with unique reference
+VALID_VIEW_MERGE_FUNCTIONS = (
```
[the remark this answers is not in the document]
sfc-gh-jdu: if it's complement, we can programmatically generate this list?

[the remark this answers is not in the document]
sfc-gh-yzou: Oh, @sfc-gh-aalam , @sfc-gh-jdu  sorry, that is not used anywhere, i forgot to remove that file, it is the same as VALID_PROJECTION_MERGE_FUNCTIONS.

@sfc-gh-jdu I have thought about this, they way i try to get the set of functions is to run a sql query to get all functions and filtered based on some properties. I also did some manual verification to make sure the current set is valid. This is a very restricted set at this moment.

Since this requires sync with server side, so i didn't really do it on session init to avoid unnecessary overhead. Let me put some instructions here about  how to pull the functions and property for check.

src/snowflake/snowpark/_internal/select_projection_complexity_utils.py
sfc-gh-yzou: sg! sorted

src/snowflake/snowpark/_internal/select_projection_complexity_utils.py
sfc-gh-yzou: That is a good question, yes. right now, I expect developers to add this when they start adding new functions support. the only problem is i could not add instructions to check the function property here because there is an internal parameter needs to be flipped on. Since it requires turn on an internal parameter, we can't do autosync at client side also. I checked the functions satisfies the properties on prod1, there are 673 now, here we covered about 200, which is not bad. Right now, i only added this 200 is more for safety. 

Do we have any place to doc those snowflake internal involved instructions?
The other way we can handle this is like how is_data_generator is added there, also add other properties as part of the FunctionExpression, however, i would like to enforce the configuration of those properties values if we do that. This will continue require developer to explicitly maintain those properties, but might be better since it enforces user think about this when adding new function support. However, the problematic part is about functions.builtin, which is called by snowpark pandas and also open for user usage.  i think we can say if the property is UNKNOWN, we just do not consider it is valid, which may not be valid. 
If we prefer to go through the other way, i think we can do it in a different pr, because it will require touch of a lot of file, and i will need to think about what is the best way to add this since it contains bunch of property.

src/snowflake/snowpark/_internal/valid_view_merge_functions.py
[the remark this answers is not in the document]
sfc-gh-jdu: I see, thanks

src/snowflake/snowpark/_internal/select_projection_complexity_utils.py
sfc-gh-jdu: Got it, I'm just not sure how to ensure the developer is aware of this list when adding a new function. Maybe we can add a merge gate like changelog, to check whether any function is added without modifying this list.

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/select_projection_complexity_utils.py
sfc-gh-yzou: created ticket to follow up here https://snowflakecomputing.atlassian.net/browse/SNOW-1678506

PR #2331, opened 2024-09-19
SNOW-1640167: Test Timedelta input and output in apply-like methods.

sfc-gh-mvashishtha:
Fixes SNOW-1640167

Test two scenarios for apply-like methods `apply`, `applymap`, `groupby.transform`, and `groupby.apply`:

1) input dataframe has timedelta data columns or index columns
2) output includes timedelta values.

Situation 1) currently raises `NotImplementedError` and will continue to do so after this commit. Recognizing that we have timedelta instead of the underlying integer as input inside the UD(T)F requires adding extra metadata. Postpone this task till a user asks for it. In this commit, just check for the `NotImplementedError`.

Situation 2) currently puts string outputs instead of `Timedelta` in the result. That's a bug, but it does follow what Snowflake does in this case. Fixing the bug is possible, but requires adding some extra metadata to the UD(T)F output. For now defer this task until a feature notices the discrepancy. In this commit, just check for an `AssertionError` that the Snowpark pandas result is not correct.

### DIFF
--- a/tests/integ/modin/frame/test_apply.py
+++ b/tests/integ/modin/frame/test_apply.py
@@ -35,6 +35,15 @@ from tests.integ.modin.utils import (
 # test data which has a python type as return type that is not a pandas Series/pandas DataFrame/tuple/list
 BASIC_DATA_FUNC_PYTHON_RETURN_TYPE_MAP = [
     [[[1.1, 2.2], [3, np.nan]], np.min, "float"],
+    param(
+        [[1.0, 2.2], [3, np.nan]],
+        lambda x: native_pd.Timedelta(1),
+        "native_pd.Timedelta",
+        id="return_timedelta_scalar",
+        marks=pytest.mark.xfail(
+            strict=True, raises=AssertionError, reason="SNOW-1619940"
+        ),
+    ),
     [[[1.1, 2.2], [3, np.nan]], lambda x: x.sum(), "float"],
     [[[1.1, 2.2], [3, np.nan]], lambda x: x.size, "int"],
     [[[1.1, 2.2], [3, np.nan]], lambda x: "0" if x.sum() > 1 else 0, "object"],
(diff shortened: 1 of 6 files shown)


tests/integ/modin/series/test_apply.py
```diff
@@ -171,9 +189,28 @@ def test_apply_and_map_basic_with_type_hints(data, func, return_type):
...
+        [None, 1],
+        [np.nan, None, 2.1],
+        param(
+            [native_pd.Timedelta(1), native_pd.Timedelta(2)],
+            id="input_timedleta_columns",
+            marks=pytest.mark.xfail(strict=True, raises=NotImplementedError),
+        ),
+    ],
+)
 @sql_count_checker(query_count=4, udf_count=1)
```
sfc-gh-mvashishtha, line 212: This test function used to be parametrized by `BASIC_DATA_FUNC_RETURN_TYPE_MAP`, but the test function confusingly used the `data` from that constant but ignored the `func` and `return_type` params. I wanted to attach `xfail`s to the `BASIC_DATA_FUNC_RETURN_TYPE_MAP` params that reflect what happens when you call the `func` with the given input, but this function applies a different `func`, so I'm copying just the test data into a new parameter list here, and using custom `xfail`s.

sfc-gh-jkew (approved):

tests/integ/modin/frame/test_apply.py
```diff
@@ -124,6 +143,17 @@ def foo(row) -> str:
         eval_snowpark_pandas_result(snow_df, df, lambda x: x.apply(foo, axis=1))
 
 
+@pytest.mark.xfail(strict=True, raises=NotImplementedError)
```
sfc-gh-helmeleegy, line 146: I thought for NotImplementedError, we would typically use `pytest.raises(NotImplementedError)` instead of xfails, as we generally wanted to lower the number of xfails in our test suite?

sfc-gh-helmeleegy: Yeah, the points you make do make sense. I think one benefit of `pytest.raises` is that you can also match on the error message. Also, looking at how commonly used each pattern is in our test suite (including Snowpark python tests), I could find about 300 instances of `xfail` vs about 1700 instances of `pytest.raises`. But still it's fair to say that both patterns are common enough.

sfc-gh-helmeleegy (approved):
LGTM.

PR #2336, opened 2024-09-20
SNOW-1674875: merge gate for multithreaded wrapper tests

sfc-gh-aalam:

   

   Fixes SNOW-1674875

   This PR adds:
   - `multithreading_mode` which runs tests marked with `multithreaded_run` concurrently in multiple threads using shared session.
   - merge gates to run snowpark and snowpandas tests in multithreaded mode
   - tags a subset of tests with `multithreaded_run`


tests/utils.py
```diff
@@ -135,6 +136,30 @@ def running_on_jenkins() -> bool:
     return RUNNING_ON_JENKINS
 
 
+def running_in_multi_threaded_mode() -> bool:
+    """Whether tests are currently running in multi-threaded mode."""
+    return os.getenv("SNOWPARK_MULTITHREADING_MODE") == "true"
+
+
+def multithreaded_run(num_threads: int) -> None:
```
sfc-gh-aling, line 144: nit: you can make default num_threads=5 so that you do not need to specify num_thread in each call which defaults to 5

tests/conftest.py
```diff
@@ -85,6 +86,14 @@ def cte_optimization_enabled(pytestconfig):
...
 
 
+@pytest.fixture(scope="session", autouse=True)
+def multithreading_mode(pytestconfig):
+    enabled = pytestconfig.getoption("multithreading_mode")
+    if enabled:
+        os.environ["SNOWPARK_MULTITHREADING_MODE"] = "true"
+    return enabled
+
+
```
sfc-gh-aling, line 96: remember to run in stored proc, I remember previously when I called similar function in local testing test setup and it broke sproc test

tests/integ/scala/test_dataframe_suite.py
```diff
@@ -181,6 +182,7 @@ def test_view_should_be_updated(session, local_testing_mode):
             Utils.drop_view(session, view_name)
 
 
+@multithreaded_run(num_threads=5)
```
sfc-gh-aling, line 185: what's the criteria of selecting tests to run in multithread mode?

I'm wondering if we can have all tests run in multithreaded_run? it might be an overkill to run the whole test set in each PR but we can also do the whole tests multithread mode in daily running?

tests/integ/modin/test_modin_stored_procedures.py
```diff
@@ -11,7 +11,8 @@
     supported_modin_version,
     supported_pandas_version,
 )
-from tests.integ.utils.sql_counter import sql_count_checker
+from tests.integ.sql_counter import sql_count_checker
+from tests.utils import multithreaded_run
```
sfc-gh-aalam, line 15:
```suggestion
from tests.integ.utils.sql_counter import sql_count_checker
from tests.utils import multithreaded_run
```

tests/integ/scala/test_dataframe_suite.py
sfc-gh-aalam: These tests are selected at random to get diverse coverage of different APIs. No strict rule here. Since each test is run x5, I want to not run this for every test. I think it will be useful for any new apis that are added to simple put this tag and run test their code to get some level of confidence.

sfc-gh-aling (approved):

sfc-gh-jrose (approved):

tests/conftest.py
```diff
@@ -86,6 +87,14 @@ def cte_optimization_enabled(pytestconfig):
     return pytestconfig.getoption("enable_cte_optimization")
 
 
+@pytest.fixture(scope="session", autouse=True)
+def multithreading_mode(pytestconfig):
+    enabled = pytestconfig.getoption("multithreading_mode")
+    if enabled:
+        os.environ["SNOWPARK_MULTITHREADING_MODE"] = "true"
```
sfc-gh-yzou, line 94: why do we need an env setup here?

tests/utils.py
```diff
@@ -135,6 +136,30 @@ def running_on_jenkins() -> bool:
     return RUNNING_ON_JENKINS
 
 
+def running_in_multi_threaded_mode() -> bool:
+    """Whether tests are currently running in multi-threaded mode."""
```
sfc-gh-yzou, line 140: can we do it like how cte, sql simplifier is triggered , instead of checking environment variable

tox.ini
```diff
@@ -90,11 +90,13 @@ commands =
...
     notdoctest: {env:SNOWFLAKE_PYTEST_CMD} -m "{env:SNOWFLAKE_TEST_TYPE} or udf" {posargs:} tests
+    multithreaded: {env:SNOWFLAKE_PYTEST_CMD} --multithreading_mode -m "{env:SNOWFLAKE_TEST_TYPE} and multithreaded_run" {posargs:} tests
     notudfdoctest: {env:SNOWFLAKE_PYTEST_CMD} -m "{env:SNOWFLAKE_TEST_TYPE} and not udf" {posargs:} tests
     local: {env:SNOWFLAKE_PYTEST_CMD} --local_testing_mode -m "integ or unit or mock" {posargs:} tests
     dailynotdoctest: {env:SNOWFLAKE_PYTEST_DAILY_CMD} -m "{env:SNOWFLAKE_TEST_TYPE} or udf" {posargs:} tests
     # Snowpark pandas commands:
     snowparkpandasnotdoctest: {env:MODIN_PYTEST_CMD} --durations=20 -m "{env:SNOWFLAKE_TEST_TYPE}" {posargs:} {env:SNOW_1314507_WORKAROUND_RERUN_FLAGS} tests/unit/modin tests/integ/modin tests/integ/test_df_to_snowpark_pandas.py
+    snowparkpandasmultithreaded: {env:MODIN_PYTEST_CMD} --multithreading_mode --skip_sql_count_check --durations=20 -m "{env:SNOWFLAKE_TEST_TYPE} and multithreaded_run" {posargs:} {env:SNOW_1314507_WORKAROUND_RERUN_FLAGS} tests/integ/modin tests/integ/test_df_to_snowpark_pandas.py
```
sfc-gh-yzou, line 99: here we are only running tests/integ/modin tests/integ/test_df_to_snowpark_pandas.py? is that intended? i see you updated a set of modin test there

tox.ini
```diff
@@ -90,11 +90,13 @@ commands =
     notudf: {env:SNOWFLAKE_PYTEST_CMD} -m "{env:SNOWFLAKE_TEST_TYPE} and not udf" {posargs:} src/snowflake/snowpark tests
     udf: {env:SNOWFLAKE_PYTEST_CMD} -m "{env:SNOWFLAKE_TEST_TYPE} or udf" {posargs:} src/snowflake/snowpark tests
     notdoctest: {env:SNOWFLAKE_PYTEST_CMD} -m "{env:SNOWFLAKE_TEST_TYPE} or udf" {posargs:} tests
+    multithreaded: {env:SNOWFLAKE_PYTEST_CMD} --multithreading_mode -m "{env:SNOWFLAKE_TEST_TYPE} and multithreaded_run" {posargs:} tests
```
sfc-gh-yzou, line 93: so we are running all tests, but some tests are written in multithreaded way?

tests/conftest.py
```diff
@@ -29,6 +29,7 @@ def pytest_addoption(parser):
     parser.addoption("--disable_sql_simplifier", action="store_true", default=False)
     parser.addoption("--local_testing_mode", action="store_true", default=False)
     parser.addoption("--enable_cte_optimization", action="store_true", default=False)
+    parser.addoption("--multithreading_mode", action="store_true", default=False)
```
sfc-gh-yzou, line 32: now since you have a parameter now, the multithreading_mode can probably work in a very similar way as other parameters

tests/integ/modin/conftest.py
```diff
@@ -79,8 +79,12 @@ def auto_annotate_sql_counter(request):
 
 @pytest.fixture(autouse=True)
 def check_sql_counter_invoked(request):
+    from tests.conftest import SKIP_SQL_COUNT_CHECK
+
```
sfc-gh-yzou, line 83: since we have moved sql_counter for snowpark oython usage, you probably need to bump this up to snowpark python level to allow skip sql counter check at snowpark python level

tox.ini
sfc-gh-aalam: we are running both `tests/integ/modin` and `tests/integ/test_df_to_snowpark_pandas.py`. The list is same as regular moding tests

tox.ini
sfc-gh-aalam: tests that are tagged with `@multithreaded_run()` with run in multi-threaded way. Others will run regularly

tests/integ/modin/conftest.py
sfc-gh-aalam: that's already done here, right. We are importing from `tests.conftest` instead of `tests.modin.conftest`

tests/utils.py
sfc-gh-aalam: Yep. I'll make that update.

tests/conftest.py
```diff
@@ -86,6 +87,17 @@ def cte_optimization_enabled(pytestconfig):
     return pytestconfig.getoption("enable_cte_optimization")
 
 
+MULTITHREADING_MODE_ENABLED = False
+
+
+@pytest.fixture(scope="session", autouse=True)
+def multithreading_mode(pytestconfig):
+    enabled = pytestconfig.getoption("multithreading_mode")
```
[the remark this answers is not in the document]
sfc-gh-yzou: i think i got what are trying to do here, the flag MULTITHREADING_MODE_ENABLED is only used to indicate whether to apply multi-threading change to the test or not, not really control the thread-safe session object implementation. Maybe call it MULTITHREADING_TEST_MODE_ENABLED to be more clear

tox.ini
sfc-gh-yzou: that means we will have a lot of duplicated tests with other daily job, note that all our jobs are run on the same cluster, running to many tests could make our merge gate slow. One thing we can do is to use markers, and only run the tests with the marker, you can take a look at the maker "@pytest.mark.modin_sp_precommit", which is used to control the tests that are run in the precommit job.

tests/conftest.py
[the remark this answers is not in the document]
sfc-gh-aalam: I added the param protection later but I feel these merge gates should enable the param protection and run them in multithreading mode.

tests/integ/modin/conftest.py
sfc-gh-aalam: hmm... I understand your point here but shouldn't the tests fail if that is the case?

tests/mock/conftest.py
```diff
@@ -10,8 +10,20 @@
 
 
 @pytest.fixture(scope="function")
-def session():
-    with Session(MockServerConnection()) as s:
+def mock_server_connection(multithreading_mode_enabled):
+    options = {
```
sfc-gh-aalam, line 14: local testing does not make real connection so testing enabled/disabled mode using parameter.

tests/utils.py
```diff
@@ -135,6 +136,26 @@ def running_on_jenkins() -> bool:
...
 
+def multithreaded_run(num_threads: int = 5) -> None:
+    """When multithreading_mode is enabled, run the decorated test function in multiple threads."""
+    from tests.conftest import MULTITHREADING_TEST_MODE_ENABLED
+
+    def decorator(func):
+        @pytest.mark.multithreaded_run
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            if MULTITHREADING_TEST_MODE_ENABLED:
```
sfc-gh-yzou, line 147: for those tests, probably doesn't work when the thread safe session object is not enabled, you might have to add a check here once you have the feature flag

sfc-gh-yzou (approved):

tests/integ/modin/conftest.py
sfc-gh-aalam: I see what you mean here. If we add sql counter check to snowpark python and also run those tests in do multithreaded run with them, then, this needs to move up.

tests/utils.py
sfc-gh-aalam: I seem to have removed config which creates session in thread-safe mode with `--multithreading_mode` option. I have readded it now.

tests/conftest.py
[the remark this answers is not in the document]
sfc-gh-aalam: I accidentlaly remove this change when merging: https://github.com/snowflakedb/snowpark-python/pull/2336/files#diff-3eb784db46c74f591684379e15fe5cb643571423d06dfa1f317f4d73e251542bR213-R216. Added it back now.

PR #2340, opened 2024-09-20
[SNOW-1632898] Adjust SelectStatement projection complexity calculation

sfc-gh-yzou:

   

SNOW-1632898

1) adjust the projection complexity calculation, and cumulative complexity for selectStatement when _merge_projection_complexity_with_subquery set to True.

2) update the complexity calculation for snowflake plan to directly get the complexity from source plan, and update the reset to reset the cumulative complexity.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -25,6 +25,7 @@ import snowflake.snowpark._internal.utils
 from snowflake.snowpark._internal.analyzer.cte_utils import encode_id
 from snowflake.snowpark._internal.analyzer.query_plan_analysis_utils import (
     PlanNodeCategory,
+    subtract_complexities,
     sum_node_complexities,
 )
 from snowflake.snowpark._internal.analyzer.table_function import (
@@ -656,6 +657,7 @@ class SelectStatement(Selectable):
         # _merge_projection_complexity_with_subquery is used to indicate that it is valid to merge
         # the projection complexity of current SelectStatement with subquery.
         self._merge_projection_complexity_with_subquery = False
+        self._projection_complexity: Optional[List[PlanNodeCategory, int]] = None
 
     def __copy__(self):
         new = SelectStatement(
@@ -844,17 +846,7 @@ class SelectStatement(Selectable):
         complexity = {}
         # projection component
         complexity = (
-            sum_node_complexities(
-                complexity,
-                *(
-                    getattr(
-                        expr,
-                        "cumulative_node_complexity",
-                        {PlanNodeCategory.COLUMN: 1},
-                    )  # type: ignore
-                    for expr in self.projection
-                ),
-            )
+            sum_node_complexities(*self.projection_complexities)
             if self.projection
             else complexity
         )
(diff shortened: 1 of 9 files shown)


tests/integ/test_large_query_breakdown.py
```diff
@@ -108,15 +108,16 @@ def test_no_valid_nodes_found(session, large_query_df, caplog):
 def test_large_query_breakdown_with_cte_optimization(session):
     """Test large query breakdown works with cte optimized plan"""
     session._cte_optimization_enabled = True
+    # session.large_query_breakdown_enabled = False
```
sfc-gh-aalam, line 111: remove

tests/integ/test_deepcopy.py
```diff
@@ -50,6 +52,57 @@
 ]
 
 
+def create_df_with_deep_nested_with_column_dependencies(
+    session, temp_table_name, nest_level: int
+) -> DataFrame:
+    """
```
sfc-gh-aalam, line 58: we could add it to test utils but this is also fine.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -894,6 +895,27 @@ def individual_node_complexity(self) -> Dict[PlanNodeCategory, int]:
...
+                assert isinstance(self.from_, SelectStatement)
+                self._cumulative_node_complexity = subtract_complexities(
+                    self._cumulative_node_complexity,
+                    sum_node_complexities(*self.from_.projection_complexities),
+                )
+
+        return self._cumulative_node_complexity
+
+    @cumulative_node_complexity.setter
+    def cumulative_node_complexity(self, value: Dict[PlanNodeCategory, int]):
```
sfc-gh-aalam, line 916: would this not be inherited from `Selectable`?

tests/integ/test_nested_select_plan_analysis.py
```diff
@@ -44,51 +55,161 @@ def setup(request, session):
...
     df_res = df_res.select(col("a") * 2, (col("c") + 2).as_("d"))
-    verify_dataframe_select_statement(df_res, can_be_merged_when_enabled=True)
+    verify_dataframe_select_statement(
+        df_res,
+        can_be_merged_when_enabled=True,
+        complexity_before_merge={
+            PlanNodeCategory.LOW_IMPACT: 4,
+            PlanNodeCategory.COLUMN: 10,
+            PlanNodeCategory.LITERAL: 13,
+        },
```
sfc-gh-aalam, line 136: this is an interesting pattern. I would have assumed that after merge, the complexity goes up but it is going down in almost all tests I have seen. is this expected?

sfc-gh-yzou: yes. Under some case it will, think about the effect is very similar as SQL simpllifier, when we have SELECT col1 + 1, col2 from (select col1 + 2 as col1, col2 from table), the effect is like select col1 + 1 + 2, col2 from table, you can see that the query after merging is actually simplier.

The case that will start exploding is the case when there are more than one dupliation, like select col1 + 1 + col2, col2 from (select col1 + 2 as col1, col2 + 2  as col2 from table), the final effect is like select (col1 + 2) + 1 + (col2 + 2), col2 + 2 form table, the inlining will make the total complexity increase.

tests/integ/test_large_query_breakdown.py
sfc-gh-yzou: removed

tests/integ/test_nested_select_plan_analysis.py
sfc-gh-yzou: if you take a look of the test case like test_large_query_breakdown_with_cte_optimization, i actually added more reference to the col, and the complexity increase much faster, and the number of loops to reach complexity boundary becomes samller

tests/integ/test_deepcopy.py
sfc-gh-yzou: i was actually trying to see if there is such file, but didn't find it. did i miss it somewhere?

sfc-gh-aalam (approved):

sfc-gh-jdu (approved):

PR #2344, opened 2024-09-23
SNOW-1617634: implement one-pass algorithm

sfc-gh-aalam:

   

   Fixes SNOW-1617634

   Implement large query breakdown using one-pass algorithm. All existing tests must pass.

### DIFF
--- a/src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
+++ b/src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
@@ -417,4 +417,6 @@ class LargeQueryBreakdown:
 
         if candidate_node is not None:
             # Update the priority queue with the new eligible node in the ancestors.
-            heapq.heappush(self.priority_queue, HeapNode(-candidate_score, candidate_node))
+            heapq.heappush(
+                self.priority_queue, HeapNode(-candidate_score, candidate_node)
+            )
--- a/src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
+++ b/src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
@@ -138,7 +138,7 @@ class LargeQueryBreakdown:
         # of all valid nodes eligible to breakdown. Implemented using a min-heap gives us quick
         # access to the node with the highest complexity score.
         # When None, initialize the priority queue.
-        self.priority_queue = None
+        self.priority_queue: Optional[List[HeapNode]] = None
 
     def apply(self) -> List[LogicalPlan]:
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -377,8 +403,23 @@ def _replace_child_and_update_ancestors(
                 # Skip if the node is already updated.
                 continue
 
+            valid_to_breakdown, score = self._is_node_valid_to_breakdown(node)
+            if valid_to_breakdown and score >= candidate_score:
```
sfc-gh-yzou, line 407: it seems that we are rechecking among all nodes has been updated? However, it seems to me that we should only need to check the candidates in the existing heap, because large query break down is suppose to reduce the complexity score, there shouldn't be a case that it is originally not a candidate becomes a new candidate, right?

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -377,8 +403,23 @@ def _replace_child_and_update_ancestors(
...
 
             parents = self._parent_map[node]
             nodes_to_reset.extend(parents)
+
+        if candidate_node is not None:
+            # Assertion for type check. we must have initialized the priority queue
+            # by the time we reach here.
+            assert self.priority_queue is not None
+            # Update the priority queue with the new eligible node in the ancestors.
+            heapq.heappush(
```
sfc-gh-yzou, line 423: it doesn't seem we adjust the queue its-self, what happens if the same candidate node occurs in the queue twice? the new score should be the correct score used, but if we did not pop out the old candidate, we might still use the old score for comparison
for example if I have (100, node1), and then node 1 scoe is updated to 50 after breakdown, and it is valid, we will have both (100, node1), (50, node1) in the heap, right?

sfc-gh-yzou: do we have a test for such case also?

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -2,6 +2,7 @@
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
 
+import heapq
```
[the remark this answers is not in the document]
sfc-gh-aalam: I was using sortedcontainers earlier. when reimplementing, I realized we can use heapq which does not add any additional dependencies.

PR #2346, opened 2024-09-24
SNOW-1566370: remove duplicate with query complexity

sfc-gh-aalam:

   

   Fixes SNOW-1566370

   This PR makes two updates:
   1. Update the `referenced_ctes` datatype from `Set[str]` to `Set[WithQueryBlock]`. Now, instead of storing the names of all referenced_ctes, we will the set of `WithQueryBlock` the subtree of a `TreeNode`.
   2. Using `cumulative_complexity_score` of `WithQueryBlock` and their count, adjust the complexity score of the current node to reflect the correct value.


src/snowflake/snowpark/_internal/analyzer/query_plan_analysis_utils.py
```diff
@@ -51,8 +52,17 @@ def sum_node_complexities(
     return dict(counter_sum)
 
 
-def get_complexity_score(
-    cumulative_node_complexity: Dict[PlanNodeCategory, int]
-) -> int:
+def get_complexity_score(node) -> int:
```
sfc-gh-yzou, line 55: put type hint here, the node can only be SnowflakePlan or Selectable

src/snowflake/snowpark/_internal/analyzer/query_plan_analysis_utils.py
```diff
@@ -51,8 +52,17 @@ def sum_node_complexities(
...
     """Calculates the complexity score based on the cumulative node complexity"""
-    return sum(cumulative_node_complexity.values())
+    adjusted_cumulative_complexity = node.cumulative_node_complexity.copy()
```
sfc-gh-yzou, line 57: I see we are trying re-adjust the complexity here because we might counting the same WithQueryBlock child multiple times, which is wired, because the complexity is not correctly calculated at each node.
Instead of doing this, how about adjust the cumulative_complexity_score to ignore the children complexity, because the WithQueryBlock is very similar as a select table, and we can still argue the logical node complexity is calculate correctly because the WithQueryBlock is defined at the  top. Then similar as how we correct the final query generated by attach the WithQuery on top of the current generated query, we simply add the WithQueryBlock complexity here.

In that way, the whole way for reason about the WithQueryBlock can be consistent, and we also get a reasonable correct calculated complexity at each logical node.

sfc-gh-yzou: if we do this, we don't need to change the referenced_ctes(self) -> Dict[WithQueryBlock, int], but just Set[WithQueryBlock]

sfc-gh-aalam: hmm...that's a better way

sfc-gh-yzou (approved):

sfc-gh-jdu (approved):

PR #2356, opened 2024-09-26
Factor out a subroutine to check for missing AST + abstract away temp table names in AST tests

sfc-gh-azwiegincew:


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -50,6 +50,15 @@
 # Test mode. In test mode, the source filename is ignored.
 SRC_POSITION_TEST_MODE = False
 
+# Debug check for missing AST. This is invoked with various arguments that are expected to be non-NULL if the AST
```
sfc-gh-lspiegelberg, line 53: Let's move the comment here into the docstring? I.e.
```
def debug_check_missing_ast(ast, container):
   """Debug check..."
   ...
   
```

The parameters are also missing type hints, please add (the precommit should complain about this)

src/snowflake/snowpark/dataframe.py
```diff
@@ -584,15 +584,9 @@ def _set_ast_ref(self, sp_dataframe_expr_builder: Any) -> None:
...
-        # TODO: remove the None guard below once we generate the correct AST.
-        if self._ast_id is None:
-            if FAIL_ON_MISSING_AST:
-                _logger.debug(self._explain_string())
-                raise NotImplementedError(
-                    f"DataFrame with API usage {self._plan.api_calls} is missing complete AST logging."
-                )
-        else:
-            sp_dataframe_expr_builder.sp_dataframe_ref.id.bitfield1 = self._ast_id
+        # TODO: remove once we generate the correct AST.
```
sfc-gh-lspiegelberg, line 587: nit: I think we can rename to `check_missing_ast` - no need for the debug prefix :)

tests/ast/test_ast_driver.py
```diff
@@ -248,7 +249,8 @@ def run_test(session):
         sys.modules[test_name] = test_module
         spec.loader.exec_module(test_module)
         base64_batches = test_module.run_test(session)
-        unparser_output = render(base64_batches) if pytest.unparser_jar else ""
+        raw_unparser_output = render(base64_batches) if pytest.unparser_jar else ""
```
sfc-gh-lspiegelberg, line 252: Should we enumerate them?
I.e., First encountered temp table -> SNOWPARK_TEMP_TABLE_1 and so on.
This would allow to keep subsequent API accesses testing for the right thing when two tables are involved e.g.

sfc-gh-lspiegelberg: (side note: When comparing protobufs - I removed the temp tables altogether)

sfc-gh-lspiegelberg: can we add a comment on why we sub the temp tables here?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: the linter GH action doesn't pass yet, can we fix?

src/snowflake/snowpark/dataframe.py
sfc-gh-azwiegincew: I prefer the debug in the name to indicate this isn't meant to be a production check. It's a fairly common practice (e.g. https://support.microsoft.com/en-us/topic/fix-assert-failure-of-the-debugchk-macro-when-you-delete-a-wi-fi-profile-in-a-debug-build-of-windows-embedded-compact-7-aa89c56d-ea8b-74f4-bd1b-751bcffe3dbb).

tests/ast/test_ast_driver.py
sfc-gh-azwiegincew: We're coming across these in the Spark project. This was meant to be a quick change – it's not worth spending the time on the fancier version you're suggesting at the moment.

sfc-gh-lspiegelberg (approved):

PR #2357, opened 2024-09-26
[DRAFT] [SNOW-1625268] Add github action to ensure AST correctness

sfc-gh-lspiegelberg:
tbd,

### DIFF
--- a/src/snowflake/snowpark/dataframe_na_functions.py
+++ b/src/snowflake/snowpark/dataframe_na_functions.py
@@ -392,7 +392,8 @@ class DataFrameNaFunctions:
         if not value_dict:
             new_df = copy.copy(self._dataframe)
             add_api_call(new_df, "DataFrameNaFunctions.fill")
-            new_df._ast_id = stmt.var_id.bitfield1
+            if _emit_ast:
+                new_df._ast_id = stmt.var_id.bitfield1
             return new_df
         if not all(
             [
@@ -594,7 +595,8 @@ class DataFrameNaFunctions:
         elif len(subset) == 0:
             new_df = copy.copy(self._dataframe)
             add_api_call(new_df, "DataFrameNaFunctions.replace")
-            new_df._ast_id = stmt.var_id.bitfield1
+            if _emit_ast:
+                new_df._ast_id = stmt.var_id.bitfield1
             return new_df
 
         if isinstance(to_replace, dict):
@@ -615,7 +617,8 @@ class DataFrameNaFunctions:
         if not replacement:
             new_df = copy.copy(self._dataframe)
             add_api_call(new_df, "DataFrameNaFunctions.replace")
-            new_df._ast_id = stmt.var_id.bitfield1
+            if _emit_ast:
+                new_df._ast_id = stmt.var_id.bitfield1
             return new_df
         if not all(
             [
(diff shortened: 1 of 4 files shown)


tests/integ/test_dataframe.py
```diff
@@ -175,7 +175,7 @@ def test_read_stage_file_show(session, resources_path, local_testing_mode):
             session.read.option("purge", False)
             .schema(user_schema)
             .csv(test_file_on_stage)
```
sfc-gh-lspiegelberg, line 177: `_show_string` is an internal API. Therefore pass explicitly `_emit_ast`. We do not mark internal APIs with `@publicapi`.

src/snowflake/snowpark/dataframe_na_functions.py
```diff
@@ -392,7 +392,8 @@ def fill(
         if not value_dict:
             new_df = copy.copy(self._dataframe)
             add_api_call(new_df, "DataFrameNaFunctions.fill")
-            new_df._ast_id = stmt.var_id.bitfield1
+            if _emit_ast:
+                new_df._ast_id = stmt.var_id.bitfield1
```
sfc-gh-oplaton, line 396: Intuitively, this pattern seems suspicious. Is it OK for two distinct object to share the same `VarId`?

tests/integ/scala/test_dataframe_join_suite.py
sfc-gh-oplaton, line 1: This, and other test files that follow the same pattern are surprising: Why don't `_show_string` and other methods involved take the default value of `_emit_ast` from the session, the same way as public APIs? What prevents that from working? (Or why is it a bad idea, if it is?)

sfc-gh-lspiegelberg: That's a great question! My reasoning was here as follows:

1.) `_show_string` is a private API. It is not exposed to the user, and therefore doesn't get a `@publicapi` decorator.
2.) Because it is private, the caller of this API is responsible for the state (i.e., passing `_emit_ast`).
3.) Unfortunately, in the tests the API design is not correctly accounted for. A private API is used to test things. The contract of `show()` is to output to stdout. Hence, a test should capture stdout and compare. However, for whichever reason this is not done. Instead a private (!) function is used to write end-to-end integration tests. I saw 3 options:
a.) Make `_show_string()` a public API, that doesn't great.
b.) Rewrite all tests to use `show()` and stdout capture -> a lot of work.
c.) Go a hybrid, basically honor the original test intention but have it do the private contract: Pass `_emit_ast` and ensure state because internals are getting called. -> that's what I did.

to sum up, the correct way to do it for an integration test would be:
1.) make `_show_string()` a public API OR
2.) do not test internal functions in integration tests. Internals are tested with unit tests at most, never integration tests. Else, you screwed up API design.

tests/integ/test_dataframe.py
sfc-gh-lspiegelberg: @sfc-gh-oplaton I tried to explain the reasoniing here.

sfc-gh-oplaton (approved):

src/snowflake/snowpark/dataframe_na_functions.py
sfc-gh-lspiegelberg: that's a good question, here I basically simply assigned (like in other APIs) the id of the stmt to the new dataframe returned. The if stmt is needed, as the code is mixed (in some APIs we have a clear separation of AST code vs. existing, in others it is mixed in because of some parameter processing).

PR #2358, opened 2024-09-26
SNOW-1692775: Remove convert_dtype warning in Series.apply

sfc-gh-helmeleegy:

   

   Fixes SNOW-1692775

   Remove convert_dtype warning in Series.apply.

### DIFF
--- a/tests/integ/modin/series/test_apply.py
+++ b/tests/integ/modin/series/test_apply.py
@@ -2,7 +2,6 @@
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
 import datetime
-import logging
 import math
 from typing import Callable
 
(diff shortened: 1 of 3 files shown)


sfc-gh-nkrishna (approved):
Do you know the original reason for having this warning?

tests/integ/modin/series/test_apply.py
```diff
@@ -482,17 +480,6 @@ def f(x) -> float:
...
 
 
-@sql_count_checker(query_count=3)
-def test_apply_convert_dtype(caplog):
-    snow_series = pd.Series([1])
-
-    caplog.clear()
-    WarningMessage.printed_warnings.clear()
-    with caplog.at_level(logging.WARNING):
-        snow_series.apply(lambda x: x, convert_dtype=True)
```
sfc-gh-joshi, line 492: Do we want to keep a test around to enforce that no warning is raised, and that the result dtype might differ from pandas?

sfc-gh-joshi (approved):
LGTM!

tests/integ/modin/series/test_apply.py
sfc-gh-helmeleegy: I don't think it makes a lot of sense to add a test to enforce that no warning is raised. If we do the same everywhere else, we'll end up with so many test :) Also, since the parameter is deprecated now, I think it's not worth investing much more into it.

PR #2361, opened 2024-09-26
Fix df alias test failure for new query generation when sql simplifier disabled

sfc-gh-yzou:
1. Please describe how your code solves the related issue.
a) Fix alias query generation after transformation by propogating the alias map correctly to snowflake pln
b) fix a test that is actually doing wrong check

### DIFF
--- a/src/snowflake/snowpark/_internal/compiler/query_generator.py
+++ b/src/snowflake/snowpark/_internal/compiler/query_generator.py
@@ -116,11 +116,13 @@ class QueryGenerator(Analyzer):
                 assert logical_plan.source_plan is not None
                 # when encounter a SnowflakePlan with no queries, try to re-resolve
                 # the source plan to construct the result
-                res = self.do_resolve(logical_plan.source_plan)
-                resolved_children[logical_plan] = res
-                resolved_plan = res
-            else:
-                resolved_plan = logical_plan
+                from snowflake.snowpark._internal.compiler.utils import (
+                    resolve_and_update_snowflake_plan,
+                )
+
+                resolve_and_update_snowflake_plan(logical_plan, self)
+
+            resolved_plan = logical_plan
 
         elif isinstance(logical_plan, SnowflakeCreateTable):
             from snowflake.snowpark._internal.compiler.utils import (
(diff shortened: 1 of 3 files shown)


sfc-gh-yzou:
new test run https://github.com/snowflakedb/snowpark-python/actions/runs/11062453132

tests/integ/compiler/test_query_generator.py
```diff
@@ -99,13 +99,13 @@ def check_generated_plan_queries(plan: SnowflakePlan) -> None:
...
+    plan_queries: List[str], post_action_queries: List[str], num_queries: int
 ) -> None:
-    assert len(plan_queries) == 3
+    assert len(plan_queries) == num_queries
     assert plan_queries[0].startswith("CREATE")
     assert plan_queries[1].startswith("INSERT")
-    assert plan_queries[2].startswith("SELECT")
-    assert len(post_action_queries) == 1
+    assert plan_queries[-1].startswith("SELECT")
+    assert len(post_action_queries) == int(num_queries / 2)
```
sfc-gh-aalam, line 108: `len(post_action_queries) == int(num_queries / 2)` this hold in most cases but might not be true in general. We can add a comment mentioning that.

sfc-gh-aalam (approved):

sfc-gh-jdu (approved):

tests/integ/compiler/test_query_generator.py
sfc-gh-yzou: good point, when create is not followed with INSERT, this will not be true. Added comment for this test utitlity function

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -116,11 +116,13 @@ def do_resolve_with_resolved_children(
...
                 # the source plan to construct the result
-                res = self.do_resolve(logical_plan.source_plan)
-                resolved_children[logical_plan] = res
-                resolved_plan = res
-            else:
-                resolved_plan = logical_plan
+                from snowflake.snowpark._internal.compiler.utils import (
+                    resolve_and_update_snowflake_plan,
+                )
+
```
sfc-gh-aalam, line 122: .

PR #2363, opened 2024-09-27
[SNOW-1678113] Improve pivot_table performance

sfc-gh-yzou:

   

SNOW-1678113]

In the reported case, the unstack calls pivot table underneath, and customer end with about 2000 columns after pivot. The pivot it's self took about 1~2 seconds to finish, but the after pivot it start looping over all columns and calling append_columns in each loop here [snowpark-python/src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py at 272e4e1ee5da84f8ac0abfefda95aab3b0bf4d7e · snowflakedb/snowpark-python](https://github.com/snowflakedb/snowpark-python/blob/272e4e1ee5da84f8ac0abfefda95aab3b0bf4d7e/src/snowflake/snowpark/modin/plugin/_internal/pivot_utils.py#L704). The append columns eventually calls select with all existing projected columns, and the a check is performed on each column here [snowpark-python/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py at main · snowflakedb/snowpark-python](https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py#L616)

our profiling shows that each check too about 0.015s, and 2000 check is about 30seconds, and the snowpark select took about 0.5 seconds since it need to perform sql simplification. Since there is an outer loop of 2000, overall it could take about (30.5*2000)s , which is close to 16 h.

In order to handle the issue, we did the following:
1) use "*" for append_columns to avoid checking for each columns
2) instead of calling append_column in each loop, try to get all columns to append and only call append_columns once.
with manual testing, the customer case now took about 8.002492904663086 s to finish the unstack reported

TODO:
add this to our performance benchmark https://snowflakecomputing.atlassian.net/browse/SNOW-1706311

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -334,7 +334,7 @@ class OrderedDataFrame:
 
         return row_number().over(Window.order_by(self._ordering_snowpark_columns())) - 1
 
-    @property
+    @cached_property
     def projected_column_snowflake_quoted_identifiers(self) -> list[str]:
         """
         Returns:
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -545,9 +545,14 @@ class OrderedDataFrame:
             extracted as the quoted identifier.
         2) when it is a str
             a) if it is a start (*), then all projected columns snowflake quoted identifiers are added
-            b) otherwise, it is treated as a name of existing column, and only active columns of the current
-                ordered dataframe can be used.
+            b) otherwise, it is treated as a name of existing column, and a validation check is applied
+                to ensure that only active columns of the current ordered dataframe can be used
         e) AssertionError is raised for all cases can not be handled.
+
+        Args:
+            col: ColumnOrName, Snowpark Column expression or column name
+            active_columns: the active columns of the current ordered dataframe to perform the check against,
+                includes all projected columns, row position columns and ordering columns.
         """
         from snowflake.snowpark.modin.plugin._internal.utils import (
             is_valid_snowflake_quoted_identifier,
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -483,10 +483,6 @@ class OrderedDataFrame:
         """Returns a Snowpark session object associated with this ordered dataframe."""
         return self._dataframe_ref.snowpark_dataframe.session
 
-    @cached_property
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -612,6 +611,7 @@ def select(
         new_snowpark_column_objects: list[Column] = []
         # a list of snowflake quoted identifiers as projected columns for new OrderedDataFrame
         new_projected_columns: list[str] = []
+        active_columns = self._get_active_column_snowflake_quoted_identifiers()
```
sfc-gh-evandenberg, line 614: Do you have any profile / perf numbers on how much this improved the runtime?  Were we able to get the earlier repro (that surfaced this) to complete in reasonable time?

sfc-gh-evandenberg (approved):
LGTM

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-yzou: Yes, it is in the description. In short, it reduces from around 16 h to 8.002492904663086 s

sfc-gh-yzou: Oh, sorry, you are just talking about this line,  so each check iteration is taken about 0.015 s, half of the time it is trying to getting the active columns, so each iteration is not slow, but if there are too many iterations, it becomes large.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -537,7 +537,7 @@ def _get_active_column_snowflake_quoted_identifiers(
         return column_quoted_identifiers
 
     def _extract_quoted_identifiers_from_column_or_name(
-        self, col: ColumnOrName
+        self, col: ColumnOrName, active_columns: list[str]
     ) -> list[str]:
         """
```
sfc-gh-helmeleegy, line 542: Can you update the doc string by providing a description for both arguments?

tests/unit/modin/test_utils.py
```diff
@@ -841,44 +851,54 @@ def test_try_convert_to_simple_slice(input, output):
 
 
 def test_append_columns():
```
sfc-gh-helmeleegy, line 853: Do we need to update an existing test, or could we just add a new test? Existing tests should still pass, correct?

sfc-gh-helmeleegy (approved):
Thanks, Yun, for catching and fixing this. LGTM. Just left one nit and one question (already answered).

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-yzou: Yes, updated

PR #2367, opened 2024-09-27
SNOW-1694806: Add support for rolling and expanding count with timedelta.

sfc-gh-mvashishtha:
Fixes SNOW-1694806

Add support for rolling().count() and expanding().count() for both Series and DataFrames.

For other rolling and expanding methods, match pandas behavior, which is to raise some kind of error.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -63,6 +63,7 @@ from pandas.api.types import (
 from pandas.core.dtypes.base import ExtensionDtype
 from pandas.core.dtypes.common import is_dict_like, is_list_like, pandas_dtype
 from pandas.core.indexes.base import ensure_index
+from pandas.errors import DataError
 from pandas.io.formats.format import format_percentiles
 from pandas.io.formats.printing import PrettyDict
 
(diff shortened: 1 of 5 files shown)


sfc-gh-vbudati (approved):
Looks great, thanks Mahesh!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -14134,6 +14136,12 @@ def _window_agg(
                 ErrorMessage.not_implemented(  # pragma: no cover
                     f"Window aggregation does not support the aggregation {repr_aggregate_function(agg_func, agg_kwargs)}"
                 )
+            if (
+                snowflake_agg_func.snowpark_aggregation is not count
```
sfc-gh-nkrishna, line 14140: Can you add a comment saying that `count` does not output Timedelta columns which is why it is supported?

sfc-gh-nkrishna (approved):
LGTM, left one comment

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: Done.

tests/integ/modin/window/test_expanding.py
```diff
@@ -174,3 +186,69 @@ def test_expanding_aggregation_series_unsupported(agg_func, agg_func_kwargs):
...
+            ),
+            lambda object: getattr(object.expanding(), agg_func)(numeric_only=False),
+            expect_exception=True,
+            expect_exception_type=DataError,
+            # pandas error message is different for dataframe and series, but
+            # Snowpark pandas always uses the same message. Accept either
+            # message.
+            expect_exception_match=(
+                f"(?:{re.escape('No numeric types to aggregate')})|"
+                + f"(?:{re.escape(_TIMEDELTA_ROLLING_AGGREGATION_NOT_SUPPORTED)})"
```
sfc-gh-jjiao, line 220: qq: do we need to re.escape those literals?

sfc-gh-jjiao (approved):

tests/integ/modin/window/test_expanding.py
sfc-gh-mvashishtha: We want to interpret the portions like `)|` and `(?:` as regex, not as literals within the regex, so we don't need to escape them.

PR #2375, opened 2024-09-30
SNOW-1646883: remove disable sql simplifier hack

sfc-gh-aalam:

   

   Fixes SNOW-1646883

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/generator_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/generator_utils.py
@@ -223,12 +223,8 @@ def generate_irregular_range(
         while start + periods * offset <= end:
             periods += 1
 
-    # TODO: SNOW-1646883 fix invalid identifier error when sql_simplifier_enabled is True
     session = get_active_session()
-    sql_simplifier_enabled = session.sql_simplifier_enabled
-    session.sql_simplifier_enabled = False
     num_offsets = session.range(start=0, end=periods, step=1)
-    session.sql_simplifier_enabled = sql_simplifier_enabled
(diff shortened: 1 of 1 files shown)


sfc-gh-mvashishtha (approved):

src/snowflake/snowpark/modin/plugin/_internal/generator_utils.py
```diff
@@ -223,12 +223,8 @@ def generate_irregular_range(
         while start + periods * offset <= end:
             periods += 1
 
-    # TODO: SNOW-1646883 fix invalid identifier error when sql_simplifier_enabled is True
     session = get_active_session()
-    sql_simplifier_enabled = session.sql_simplifier_enabled
```
sfc-gh-yzou, line 228: @sfc-gh-aalam is this fixed https://snowflakecomputing.atlassian.net/browse/SNOW-1646883, what is the problem ?

sfc-gh-aalam: I don't know. It used to raise error for test described in the ticket. Seems to have fixed now so this change is no longer required

sfc-gh-azhan (approved):

src/snowflake/snowpark/modin/plugin/_internal/generator_utils.py
sfc-gh-yzou: also @sfc-gh-nkrishna that code seems again violates the architecture in our Snowpark Pandas, that there should be no direct interact with snowpark dataframe, the right way is to provide a static function inside ordered dataframe for range here, and return an ordered dataframe. Then all rest operations should happen on the ordered dataframe. can we get a refactoring task to refact that part?

PR #2378, opened 2024-09-30
[SIT-2372] `map` function

sfc-gh-lfallasavendano:
Adds the `map` function to apply a Python function to each row of a DataFrame.

   Fixes [SIT-2372](https://snowflakecomputing.atlassian.net/browse/SIT-2372)

      - [X] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing

   Adds the `map` function which applies a Python function to every row of a `DataFrame`.

[SIT-2372]: https://snowflakecomputing.atlassian.net/browse/SIT-2372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfc-gh-mvega (approved):
looks good to me... again. 
the new approach of making mpa part of functions.py does not alter to much the original implementation.

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8701,123 @@ def make_interval(
...
+):
+    """Returns a new DataFrame with the result of applying `func` to each of the
+    rows of the specified DataFrame.
+
+    Args:
+        dataframe: The DataFrame instance.
+        func: A function to be applied to every row of the DataFrame.
+        output_types: A list of types for values generated by the ``func``
+        output_column_names: A list of names to be assigned to the resulting columns.
+        wrap_row: A flag to determine if the current row must be wrapped in an instance of the :class:`Row` class.
```
sfc-gh-aalam, line 8722: `wrap_row` which are the scenarios when wrap_row should be false? based on the example, it seems to be when num rows output = 1

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8701,123 @@ def make_interval(
...
+def map(
+    dataframe,
+    func: Callable,
+    output_types: List[StructType],
+    output_column_names: Optional[List[str]] = None,
+    wrap_row=True,
+    **kwargs: Dict[str, Any],
+):
+    """Returns a new DataFrame with the result of applying `func` to each of the
+    rows of the specified DataFrame.
```
sfc-gh-aalam, line 8715: can you also include that this function registers a UDTF internally and applies the given function to each column of the dataframe.

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8701,123 @@ def make_interval(
...
+        **kwargs,
+    )
+
+    if not output_column_names:
+        output_column_names = [field.name for field in output_schema]
+
+    _output_cols = [
+        column.alias(desired_name)
+        for column, desired_name in zip(output_cols, output_column_names)
+    ]
```
sfc-gh-aalam, line 8821: why not integrate this when defining the schema

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8701,123 @@ def make_interval(
...
+    _input_cols = [dataframe[column_name] for column_name in dataframe.columns]
+    return dataframe.join_table_function(map_udtf(*_input_cols)).select(*_output_cols)
```
sfc-gh-aalam, line 8823:
```suggestion
    return dataframe.join_table_function(map_udtf(*dataframe.columns))
```
is this simpler?

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8701,123 @@ def make_interval(
...
+        |900    |
+        ---------
+        <BLANKLINE>
+
+    Note:
+        The result of the `func` function must be either a scalar value or a tuple containing the same number of elements
+        as specified in the `output_types` argument.
+    """
+    if "packages" not in kwargs:
+        kwargs["packages"] = ["snowflake-snowpark-python"]
```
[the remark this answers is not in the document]
sfc-gh-lfallasavendano: We need to include the `snowflake-snowpark-python` package because the handler uses the `snowflake.snowpark.Row` class . See the `_MapFunc` handler .

If we don't include this package we get this error: `ModuleNotFoundError: No module named 'snowflake'`

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: `wrap_row` should be False when `func` is a function that has the same number of parameters as the number of columns of the DataFrame. For example:

```
map(mydf.select(mydf[0], lit(100)),
       lambda x, y: x * y * x, 
       [IntegerType()], 
       wrap_row=False).show()
```

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: We can set the names when creating the schema, however we still need this code and the last `.select` projection since `join_table_function` includes all the columns from the original DataFrame. 

For example:

```python
another_df = session.create_dataframe([('a',10), ('b', 20), ('d', 30)], 
                                      schema=StructType([StructField('name', StringType()),
                                                         StructField('numval', IntegerType())]))

r1 = map(another_df, lambda row: (row[0] + "_", row[1] * row[1]), [StringType(), IntegerType()], output_column_names=['A','B'])
r1.show()
```
This generates:
```
---------------------------------
|"NAME"  |"NUMVAL"  |"A"  |"B"  |
---------------------------------
|a       |10        |a_   |100  |
|b       |20        |b_   |400  |
|d       |30        |d_   |900  |
---------------------------------
```

Notice that in this case it is assumed that the resulting DataFrame should have two columns.  By including the original columns we can introduce bugs for code that uses column access by index . For example:
```
r1 = map(another_df, lambda row: (row[0] + "_", row[1] * row[1]), [StringType(), IntegerType()], 
r2 = map(r1, lambda r: r[1] * 2, [IntegerType()])
```

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8701,123 @@ def make_interval(
             nanoseconds,
         )
     )
+
+
+def map(
+    dataframe,
+    func: Callable,
```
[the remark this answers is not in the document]
sfc-gh-lfallasavendano: Applied this change with some minor modifications because importing `DataFrame` directly introduces a circular reference.

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: Agreed, I just added this to the function documentation.

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: Agreed, I just removed the assignment to `_input_cols` and changed the invocation to `map_udtf`. However we cannot remove the final `.select(*_output_calls)` call because `join_table_function` returns all the columns from `dataframe`. (See previous comment)

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8711,135 @@ def make_interval(
...
+        output_schema=StructType(output_schema),
+        input_types=input_types,
+        is_permanent=False,
+        **kwargs,
+    )
+
+    if not output_column_names:
+        output_column_names = [field.name for field in output_schema]
+
+    _output_cols = [
```
sfc-gh-aalam, line 8838: If we implement output_column_names as done above, we can just do `dataframe.join_table_function(map_udtf(*dataframe.columns)).select(output_column_names)` in the final return

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8711,135 @@ def make_interval(
...
+            return tuple(result)
+        elif isinstance(result, tuple):
+            return result
+        else:
+            return (result,)
+
+    class _MapFunc:
+        def process(self, *argv):
+            if wrap_row:
+                argv = [input_row_obj(*argv)]
```
sfc-gh-aalam, line 8823:
```suggestion
            if wrap_row:
                argv = [Row(*argv)]
```

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8711,135 @@ def make_interval(
...
+    input_types = [field.datatype for field in dataframe.schema.fields]
+    num_fields = len(dataframe.schema.fields)
+    output_schema = [
+        StructField(f"c_{i + 1}", output_types[i]) for i in range(len(output_types))
+    ]
+    output_cols = [
+        col(f"${i + 1}") for i in range(num_fields, num_fields + len(output_types))
+    ]
+    input_row_obj = snowflake.snowpark.Row(*dataframe.columns) if wrap_row else None
+
```
sfc-gh-aalam, line 8811:
```suggestion

```

src/snowflake/snowpark/functions.py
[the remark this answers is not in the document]
sfc-gh-lfallasavendano: Ok, this change is applied on the last commit.

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: The value of this parameter is difficult to infer at runtime since it requires knowing how the argument of a function is used in its body.

For example, it is difficult to know (from the implementation of `map`) that these two scenarios require different values for `wrap_row`:

```python
#1
def foo(x):
    return x[0]*x[1]
...
map(df1, foo, [IntegerType()])
```
```python
#2
def foo(x):
    return x*x
...
map(df2, foo, [IntegerType()])
```

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8711,135 @@ def make_interval(
             nanoseconds,
         )
     )
+
+
+if TYPE_CHECKING:
+    import snowflake.snowpark
```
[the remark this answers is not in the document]
sfc-gh-lfallasavendano: Ok, this change is applied on the last commit.

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8711,135 @@ def make_interval(
...
+    """
+    if "packages" not in kwargs:
+        kwargs["packages"] = ["snowflake-snowpark-python"]
+    if len(output_types) == 0:
+        raise ValueError("output_types cannot be empty.")
+    input_types = [field.datatype for field in dataframe.schema.fields]
+    num_fields = len(dataframe.schema.fields)
+    output_schema = [
+        StructField(f"c_{i + 1}", output_types[i]) for i in range(len(output_types))
+    ]
```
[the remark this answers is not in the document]
sfc-gh-lfallasavendano: Applied in a new commit.

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: @sfc-gh-aalam  the problem with selecting the columns by name (via the `output_column_names`created in the previous comment) is that we can get an ambiguos column reference. The reason for this is that the `join_table_function` includes the original columns of the DataFrame. The following scenarios illustrate this problem:

1. When chaining several `map` calls (because of the column names we generate):

```python
another_df = session.create_dataframe([('a',10), ('b', 20), ('d', 30)], 
                                      schema=StructType([StructField('name', StringType()),
                                                         StructField('numval', IntegerType())]))

r1 = map(another_df, lambda row: (row[0] + "_", row[1] * row[1]), [StringType(), IntegerType()])
r2 = map(r1, lambda row: row[0], [StringType()])
r2.show()
```

2. When processing a DataFrame that already has  `C1,C2,...` columns (or `C_1, C_2,...` before this change):

```python
another_df = session.create_dataframe([('a',10), ('b', 20), ('d', 30)], 
                                      schema=StructType([StructField('C1', StringType()),
                                                         StructField('C2', IntegerType())]))

r1 = map(another_df, lambda row: (row[0] + "_", row[1] * row[1]), [StringType(), IntegerType()])
r1.show()
```

In both these scenarios we get the following error if we apply this change:

```
SnowparkSQLException: (1304): 01b773a7-0a0c-34b2-0000-1c550d43832a: 002028 (42601): SQL compilation error:
ambiguous column name 'C1'
```

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: @sfc-gh-aalam the goal of this code is to support scenarios where the user uses column names in the function that is applied to each row. For example:

```python
another_df = session.create_dataframe([('a',10), ('b', 20), ('d', 30)], 
                                      schema=StructType([StructField('NAME', StringType()),
                                                         StructField('NUMVAL', IntegerType())]))

r1 = map(another_df, lambda row: (row.NAME + "_", row.NUMVAL * row.NUMVAL), [StringType(), IntegerType()])
r1.show()
```

This is the reason for creating a `Row` object from the column names of the current dataframe (next comment).

If we do not keep this information we get the following error:

```
...
    raise AttributeError(f"Row object has no attribute {item}")
AttributeError: Row object has no attribute NAME
```

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: This is related to the previous comment.  We create `input_row_obj` to allow the user to access columns by name.

src/snowflake/snowpark/functions.py
sfc-gh-aalam: I don't understand why we need to create `input_row_obj`. Maybe I don't understand `Row` correctly. There are two cases here.

Case 1: `wrap_row = False`
- `input_row_obj = None`
- We don't execute line 8822-8823:
```
if wrap_row:
    argv = [input_row_obj(*argv)]
```

Case 2: `wrap_row = True`
- `input_row_obj = Row`
- we execute line 8822-8823:
```
if wrap_row:
    argv = [Row(*argv)]
```

If that is the case, why do we even need to create `input_row_obj`. can't we just do
```
if wrap_row:
   argv = [Row(*argv)]
```

sfc-gh-lfallasavendano: We need `input_row_obj` to create Row object instances that have the same column names as the ones in the current DataFrame.  We want to support these scenarios:

```python
map(df, lambda row: (row.C1 + 1, row.C1 * 2), [IntegerType(), IntegerType()])
```

Notice that in `functions.py` in line 8810 we create `input_row_obj` like this:

```
input_row_obj = snowflake.snowpark.Row(*dataframe.columns) if wrap_row else None
```

As described in the second example for the [documentation](https://docs.snowflake.com/developer-guide/snowpark/reference/python/latest/snowpark/api/snowflake.snowpark.Row) ,   `Row` instances are ['callable'](https://github.com/snowflakedb/snowpark-python/blob/cef750a346d60ecb5b0fe4d3e7fe6e5ad2571a5e/src/snowflake/snowpark/row.py#L199) and can be used to create other `Row` instances.

sfc-gh-aalam: Okay. this makes sense to me. Thanks for clarifying.

src/snowflake/snowpark/functions.py
sfc-gh-aalam: make sense. We should  move this part in `_output_cols` initialization too.

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8704,139 @@ def make_interval(
...
+        raise ValueError(
+            "'output_column_names' and 'output_types' must be of the same size."
+        )
+
+    output_schema = [
+        StructField(name, type_)
+        for name, type_ in zip(output_column_names, output_types)
+    ]
+    output_cols = [
+        col(f"${i + 1}") for i in range(num_fields, num_fields + len(output_types))
```
sfc-gh-aalam, line 8808:
```suggestion
        col(f"${i + num_fields + 1}").alias(col_name) for i, col_name in enumerate(output_column_names)
```

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8704,139 @@ def make_interval(
...
+            if wrap_row:
+                argv = [input_row_obj(*argv)]
+            result = func(*argv)
+            yield wrap_result(result)
+
+    map_udtf = dataframe._session.udtf.register(
+        _MapFunc,
+        output_schema=StructType(output_schema),
+        input_types=input_types,
+        is_permanent=False,
```
sfc-gh-aalam, line 8831: let user decide if they want this registration to be permanent. If would allow users to not do repeated registration since it is a time consuming process.

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: Applied in a new commit.

src/snowflake/snowpark/functions.py
sfc-gh-lfallasavendano: Applied in a new commit.

src/snowflake/snowpark/functions.py
```diff
@@ -8700,3 +8704,135 @@ def make_interval(
...
+            "'output_column_names' and 'output_types' must be of the same size."
+        )
+
+    output_schema = [
+        StructField(name, type_)
+        for name, type_ in zip(output_column_names, output_types)
+    ]
+
+    output_cols = [
+        col(f"${i + num_fields + 1}").alias(col_name)
```
sfc-gh-aalam, line 8809:
```suggestion
        col(f"${i + num_fields + 1}").alias(col_name)  # this is done to avoid collision with original table columns
```

sfc-gh-aalam (approved):

sfc-gh-fgonzalezmendez (approved):

sfc-gh-aherreraaguilar (approved):

src/snowflake/snowpark/functions.py
sfc-gh-fgonzalezmendez: Fixed in the commit [fbeaf06](https://github.com/snowflakedb/snowpark-python/pull/2378/commits/fbeaf06cf9e062c0860e6737ad82faf9fcaf939a)

PR #2379, opened 2024-09-30
SNOW-1672576 Dataframe.to_pandas and Dataframe.to_pandas_batch

sfc-gh-evandenberg:
Fixes SNOW-1672576 Dataframe.to_pandas and Dataframe.to_pandas_batch


src/snowflake/snowpark/dataframe.py
```diff
@@ -916,9 +916,17 @@ def to_pandas(
             2. If you use :func:`Session.sql` with this method, the input query of
             :func:`Session.sql` can only be a SELECT statement.
         """
-
         if _emit_ast:
```
sfc-gh-lspiegelberg, line 919: the stmt code here looks good to me! I think what is missing is an eval. `to_pandas` should be similar to collect() that it should issue the REST request. Cf. the implementation of `collect()` for example. If you write a test like

```
df = ...
with session.ast_listener() as al:
   df.to_pandas()
assert al.base64_batches is not None # <-- should contain stmt + eval. Flushed via `dataframeAst` property as part of json.
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -916,9 +916,17 @@ def to_pandas(
...
+            if statement_params is not None:
+                for k, v in statement_params.items():
+                    t = ast.statement_params.add()
+                    t._1 = k
+                    t._2 = v
+            ast.block = block
+            self._set_ast_ref(ast.df)
 
         with open_telemetry_context_manager(self.to_pandas, self):
             result = self._session._conn.execute(
```
sfc-gh-lspiegelberg, line 932: the kwargs here prob. should contain. `dataframeAst`

sfc-gh-lspiegelberg (commented):
Can you add the eval? Else looks good to me. Feel to shorten PR description, no need for the checkboxes...

src/snowflake/snowpark/dataframe.py
sfc-gh-evandenberg: done

src/snowflake/snowpark/dataframe.py
sfc-gh-evandenberg: Added eval

tests/ast/data/DataFrame.to_pandas.test
```diff
@@ -0,0 +1,39 @@
...
+pandas_df = df.to_pandas(statement_params={"SF_PARTNER": "FAKE_PARTNER"}, block=False)
+
+## EXPECTED UNPARSER OUTPUT
+
+df = session.table("test_table")
+
+pandas_df = Add a handler for Empty$ in Unparser.spDataframeExpr().to_pandas()
+
+pandas_df
+
```
sfc-gh-lspiegelberg, line 20: missing handler? need to update expectations? Can you sync with your snowpark change/rebuild the unparser?

sfc-gh-lspiegelberg (commented):
Looks all good, except I think you need to run one more time the `--update-expectations`. In the tests it shows things like
```
Add a handler for Empty$ in Unparser.spDataframeExpr().to_pandas_batches()
```

sfc-gh-lspiegelberg (approved):
LGTM!

PR #2382, opened 2024-10-01
SNOW-1638298 add memo to deepcopy

sfc-gh-aalam:

   

   Fixes SNOW-1638298

   Add memoization to deep copy so that we don't create new copies of shared children. Original and duplicated plans after this change can been seen below:

### Original Plan
original_plan

---
### Deep copy without memoization
image

---
### Deep Copied Plan with memoization (This PR)
deep_copied_plan


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -474,7 +482,11 @@ def __init__(
             self._schema_query = sql
             self._query_param = params
 
-    def __deepcopy__(self, memodict={}) -> "SelectSQL":  # noqa: B006
+    def __deepcopy__(self, memodict=None) -> "SelectSQL":  # noqa: B006
```
sfc-gh-yzou, line 485: @sfc-gh-aalam a clarify question so the memodict is used to avid repeated deepcopy of the same node, right? for example if a node whose id exists in he meodict, it will not repeated copy the node, is the correct?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -474,7 +482,11 @@ def __init__(
...
+        if memodict is None:
+            memodict = {}
```
sfc-gh-yzou, line 487: what is the benefit to have None here? does it do the same thing as passing memodict={}?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -489,24 +489,38 @@ def __copy__(self) -> "SnowflakePlan":
...
 
-    def __deepcopy__(self, memodict={}) -> "SnowflakePlan":  # noqa: B006
-        copied_source_plan = (
-            copy.deepcopy(self.source_plan) if self.source_plan else None
-        )
+    def __deepcopy__(self, memodict=None) -> "SnowflakePlan":  # noqa: B006
+        if memodict is None:
+            memodict = {}
+
+        if (self_id := id(self)) in memodict:
```
sfc-gh-yzou, line 496: can we add a comment here that if the node has already been copied, return the copied node

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yzou: for other nodes that are not selectable and snowflakeplane, do they atuomatically added it to memodict? and check the memodict?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -489,24 +489,38 @@ def __copy__(self) -> "SnowflakePlan":
...
+        else:
+            if (source_id := id(self.source_plan)) in memodict:
+                copied_source_plan = memodict[source_id]
+            else:
+                copied_source_plan = copy.deepcopy(self.source_plan, memodict)
+                memodict[source_id] = copied_source_plan
+
         copied_plan = SnowflakePlan(
-            queries=copy.deepcopy(self.queries) if self.queries else [],
+            queries=copy.deepcopy(self.queries, memodict) if self.queries else [],
```
sfc-gh-yzou, line 509: for queries, i don't think we need memodict, it is not a logical plan node

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -489,24 +489,38 @@ def __copy__(self) -> "SnowflakePlan":
...
             schema_query=self.schema_query,
-            post_actions=copy.deepcopy(self.post_actions)
+            post_actions=copy.deepcopy(self.post_actions, memodict)
```
sfc-gh-yzou, line 511: similarly, those copy probably don't need any memodict

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -483,14 +495,16 @@ def __deepcopy__(self, memodict={}) -> "SelectSQL":  # noqa: B006
             # object construct, and correct the fields after.
             convert_to_select=False,
             analyzer=self.analyzer,
-            params=deepcopy(self.query_params),
+            params=deepcopy(self.query_params, memodict),
```
sfc-gh-yzou, line 498: for query params, we don't need memodict

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -691,27 +712,33 @@ def __copy__(self):
...
+            memodict = {}
+        if (self_id := id(self)) in memodict:
+            return memodict[self_id]
         copied = SelectStatement(
-            projection=deepcopy(self.projection),
-            from_=deepcopy(self.from_),
-            where=deepcopy(self.where),
-            order_by=deepcopy(self.order_by),
-            limit_=deepcopy(self.limit_),
+            projection=deepcopy(self.projection, memodict),
```
sfc-gh-yzou, line 721: another simple optimizaiton we can do is that expression doesn't really need a copy, because we are not really doing any update to expressions, but this could be little bit wired, i think we can keep the deep copy here for now

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: We avoid deepcopy, YES. But also, since we have already seen the old object, we will return the same old deep copied object.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: I guess you are right. they are the same

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: For other plans, we can call `deepcopy` with and without `memodict`. If it is called without a memodict, the internal implementation with create a new memodict. Otherwise, we use the memodict passed by the user.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: `ScalarSubQueryExpression` contains a plan. keeping deepcopy here is better.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -692,12 +704,15 @@ def __copy__(self):
...
+            from_=deepcopy(self.from_, memodict),
+            where=deepcopy(self.where, memodict),
+            order_by=deepcopy(self.order_by, memodict),
+            limit_=deepcopy(self.limit_, memodict),
```
sfc-gh-yzou, line 715: limit is just an int, we probably don't need the memodict here

sfc-gh-yzou (approved):

sfc-gh-jdu (approved):

PR #2386, opened 2024-10-01
SNOW-1445416, SNOW-1445419: Implement DataFrame/Series.attrs

sfc-gh-joshi:

   

   Fixes SNOW-1445416 and SNOW-1445419

Implements `DataFrame`/`Series.attrs` by adding a new query compiler variable `_attrs` that is read out by frontend objects. A new annotation on the query compiler, `_propagate_attrs_on_methods`, will either copy `_attrs` from `self` to the return value, or reset `_attrs` on the return value.

I initially intended to implement this solely at the frontend layer with the override system (similar to how telemetry is added to all methods), but this created difficulties when preserving `attrs` across in-place operations like `df.columns = [...]`, and could create ambiguity if the frame had a column named `"_attrs"`. Implementing propagation at the query compiler level is simpler.

This PR also adds a new `test_attrs=True` parameter to `eval_snowpark_pandas_result`. `eval_snowpark_pandas_result` will set a dummy value of `attrs` on its inputs, and ensure that if the result is a DF/Series, the `attrs` field on the result matches that of pandas. Since pandas isn't always consistent about whether it propagates attrs or resets it (for some methods, the behavior depends on the input, and for some methods, it is inconsistent between Series/DF), setting `test_attrs=False` skips this check. When I encountered such inconsistent methods, I elected to have Snowpark pandas always propagate `attrs`, since it seems unlikely that users would rely on the `attrs` of a result being empty if they did not explicitly set it.


sfc-gh-joshi:
Marking as WONTFIX: the hackier parts were already removed by removing the modin/pandas folder from our codebase. Remaining work is blocked on a task from the open source side: https://github.com/modin-project/modin/issues/7404

tests/integ/modin/frame/test_aggregate.py
```diff
@@ -158,7 +158,14 @@ def test_corr_negative(numeric_native_df, method):
 @sql_count_checker(query_count=1)
 def test_string_sum(data, numeric_only_kwargs):
     eval_snowpark_pandas_result(
-        *create_test_dfs(data), lambda df: df.sum(**numeric_only_kwargs)
+        *create_test_dfs(data),
+        lambda df: df.sum(**numeric_only_kwargs),
+        # apparently, pandas doesn't propagate attrs if the frame is empty
```
sfc-gh-helmeleegy, line 163: I'm assuming that we do propagate `attrs` in this case. If so, can we mention that in the comment?

tests/integ/modin/frame/test_apply_axis_0.py
```diff
@@ -323,6 +323,7 @@ def test_groupby_apply_constant_output():
         snow_df,
         native_df,
         lambda df: df.groupby(by=["fg"], axis=0).apply(lambda x: [1, 2]),
+        test_attrs=False,  # native pandas is inconsistent about whether it propagates attrs
```
sfc-gh-helmeleegy, line 326: What does inconsistent mean here?

tests/integ/modin/frame/test_aggregate.py
sfc-gh-helmeleegy: There are many similar occurrences in the other tests. Can we mention our behavior there as well?

tests/integ/modin/frame/test_apply_axis_0.py
sfc-gh-helmeleegy: There are many similar occurrences in the other tests. Can we clarify the native pandas inconsistency there as well?

sfc-gh-joshi: By "inconsistent," I mean that some invocations of the function will propagate while some will not, depending on the arguments provided. In this particular test case, I think `attrs` isn't propagated because there's no relationship between the return of the applied lambda `lambda x: [1, 2]` and the data in the original frame; we don't really have a good way to check for this.

There are some cases we _can_ check:
```
>>> native_df = native_pd.DataFrame({"a": [1,2,3], "b": [4,5,6]})
>>> native_df.attrs = {"k": "v"}
>>> native_df.groupby("a").apply(sum).attrs  # function argument causes reset
{}
>>> native_df.groupby("a").apply("sum").attrs  # string argument causes propagation
{'k': 'v'}
```

Here, I think it's possible to tell based on the type of the provided aggregation, but others are not possible to determine or would require us to look at the input frame, and I don't think it's worth the effort to identify every single one of these cases in native pandas.

Do you have any thoughts on how to make this more clear?

sfc-gh-helmeleegy: Thanks for the explanation. Yes, I don't think it's worth it. But in terms of clarifying the comment, maybe it's worth adding what you mentioned above: "Some invocations of the function will propagate while some will not, depending on the arguments provided."

sfc-gh-joshi: Sounds good, updated comments to be more clear.

sfc-gh-helmeleegy (approved):
LGTM. Great work adding all the tests!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -411,7 +412,95 @@
...
+                result._attrs = copy.deepcopy(self._attrs)
+            return result
+
+        return typing.cast(T, wrap)
+
+    def reset_attrs_decorator(method: T) -> T:
+        @functools.wraps(method)
+        def wrap(self, *args, **kwargs):  # type: ignore
+            result = method(self, *args, **kwargs)
+            if isinstance(result, SnowflakeQueryCompiler) and len(self._attrs):
```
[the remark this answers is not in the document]
sfc-gh-joshi: It checks that `self._attrs` was not empty, and avoids setting `result._attrs` if so, since intermediate calls to query compiler operations may have already set `result._attrs = self._attrs`. This lets us potentially avoid creating a new empty dict for `result._attrs` if it's already empty (not a very important optimization, but I think it's good practice anyway).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -411,7 +412,95 @@
...
+
+        return typing.cast(T, wrap)
+
+    for attr_name, attr_value in cls.__dict__.items():
+        # concat is handled explicitly because it checks all of its arguments
+        # agg is handled explicitly because it sometimes resets and sometimes propagates
+        if attr_name.startswith("_") or attr_name in ["concat", "agg"]:
+            continue
+        if attr_name in _RESET_ATTRS_METHODS or any(
+            attr_name.startswith(prefix) for prefix in ["expanding", "rolling"]
```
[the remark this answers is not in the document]
sfc-gh-joshi: Some groupby methods (like `groupby_agg` and `groupby_fillna`) actually _do_ copy `attrs` to their output in native pandas, it just has to be checked on a method-by-method basis.

sfc-gh-nkumar (approved):

PR #2389, opened 2024-10-02
SNOW-1625385, SNOW-1625384: Test and/or fix timedelta in integ/modin/{pivot,tools}

sfc-gh-mvashishtha:
- Test that pivot() with timedelta inputs is not supported.
  - pivot() is complicated so defer this feature until a user requests it.
- Test that timedelta is invalid input to to_datetime()
- Fix a bug where to_numeric() would leave timedelta values alone instead of converting them to integer.

Fixes SNOW-1625385
Fixes SNOW-1625384


sfc-gh-nkrishna (approved):
LGTM!

PR #2395, opened 2024-10-03
SNOW-1445732: Implement `Series.items()`

sfc-gh-nkrishna:
SNOW-1445732

This PR implements `Series.items()` by using `SnowparkPandasRowPartitionIterator` in much the same manner as `DataFrame.iterrows()`.


src/snowflake/snowpark/modin/plugin/utils/frontend_constants.py
```diff
@@ -43,6 +43,12 @@
     "Scalar key incompatible with {} value"
 )
 
+SERIES_ITEMS_WARNING_MESSAGE = (
+    "{} will result eager evaluation and potential data pulling, which is inefficient. For efficient Snowpark "
+    "pandas usage, consider rewriting the code with an operator (such as Series.apply) which "
+    "can work on the entire Series in one shot."
```
sfc-gh-mvashishtha, line 49:
```suggestion
    "{} may result in executing one more queries to fetch each row of this series. For better " +
    "performance, consider instead using a method like Series.apply that Snowpark pandas " +
    "can execute lazily and without fetching data from Snowflake."
```

src/snowflake/snowpark/modin/plugin/extensions/series_overrides.py
```diff
@@ -203,9 +207,22 @@ def item(self):  # noqa: RT01, D200
...
+    """
+    Iterate over ``Series`` rows as (index, value) tuples.
+    """
+
+    def items_builder(s):
+        """Return tuple of the given ``Series`` in the form (index, value)."""
+        return s.name, s.squeeze()
+
+    # Raise warning message since Series.items is very inefficient.
+    WarningMessage.single_warning(SERIES_ITEMS_WARNING_MESSAGE.format("Series.items"))
```
sfc-gh-mvashishtha, line 223: I don't see a reason to use string formatting here.

src/snowflake/snowpark/modin/plugin/extensions/series_overrides.py
```diff
@@ -203,9 +207,22 @@ def item(self):  # noqa: RT01, D200
...
+
+    return SnowparkPandasRowPartitionIterator(DataFrame(self), items_builder, True)
```
sfc-gh-mvashishtha, line 225: There was some controversy about the `enable_partition_with_native_pandas=True` option when it was introduced and I think it was originally set to `False`. I don't have time to investigate right now. looks like iterrows and itertuples do different things.

tests/integ/modin/series/test_items.py
```diff
@@ -0,0 +1,45 @@
...
+
+@pytest.mark.parametrize(
+    "series",
+    [
+        native_pd.Series(
+            ["bear", "bear", "marsupial"],
+            index=["panda", "polar", "koala"],
+        ),
+        native_pd.Series(data=["a"]),
+        native_pd.Series(native_pd.timedelta_range(10, periods=10)),
```
sfc-gh-mvashishtha, line 36: test an empty series?

tests/integ/modin/series/test_items.py
```diff
@@ -0,0 +1,45 @@
...
+    pandas_list = list(pandas_result)
+    assert len(snow_list) == len(pandas_list), "lengths of items are not equal."
+    if len(snow_list) == 0:
+        # Expect no queries if there are no columns.
+        with SqlCounter(query_count=0):
+            return
+    for ((snow_index, snow_value), (pandas_index, pandas_value)) in zip(
+        snow_list, pandas_list
+    ):
+        with SqlCounter(query_count=0):
```
sfc-gh-mvashishtha, line 23: Can we adjust the SQL counters (including the `sql_count_checker`) to make it clear when exactly we execute the query? could be useful for people reading the test.

sfc-gh-nkrishna: Added a comment saying that one row count query is issued when building the generator. It can't be captured in `assert_items_results_equal`.

src/snowflake/snowpark/modin/plugin/extensions/series_overrides.py
sfc-gh-nkrishna: I followed what is done in `itertuples()`. Experimenting around with it, significantly more queries are issued when the flag is set to False so I think this is reasonable to partition with pandas and throw the warning here.
```
    enable_partition_with_native_pandas: bool, default False
        When True, retrieve the table as partitions. Each partition is a pandas DataFrame which is iterated over until
        exhausted, and the next partition is pulled in.
        When False, iterate over the Snowpark pandas DataFrame directly row-by-row.
```

sfc-gh-mvashishtha: `enable_partition_with_native_pandas=True` is likely to be faster for common use cases, but I don't have confidence in its implementation because we [skip](https://github.com/snowflakedb/snowpark-python/blame/560d56318208d9285c381897a46f0e747755820c/tests/integ/modin/frame/test_itertuples.py#L156) all the test cases that do `itertuples()` on large dataframes.

For `items()`, can we test `PARTITION_SIZE`, `PARTITION_SIZE - 1`, `PARTITION_SIZE + 1`, `PARTITION_SIZE * 2`, and `PARTITION_SIZE * 2 + 1`, and check manually that the those test cases are not too slow?

tests/integ/modin/series/test_items.py
sfc-gh-mvashishtha: The counts will change to something like [this](https://github.com/snowflakedb/snowpark-python/blame/560d56318208d9285c381897a46f0e747755820c/tests/integ/modin/frame/test_itertuples.py#L161) if we use the test data lengths I suggest [here](https://github.com/snowflakedb/snowpark-python/pull/2395/files/e99b9b6b56bf89d9e6ac679137620ea5c3f29aa7..0373649aba5c31cf99a4c0399bf6a5008ea0313e#r1786732371).

tests/integ/modin/series/test_items.py
```diff
@@ -0,0 +1,52 @@
...
+            assert snow_index == pandas_index
+            # Workaround using NumPy since nan != nan
```
sfc-gh-mvashishtha, line 26: This seems  a little awkward. Can we use [`assert_values_equal`](https://github.com/snowflakedb/snowpark-python/blob/560d56318208d9285c381897a46f0e747755820c/tests/integ/modin/utils.py#L481)?

tests/integ/modin/series/test_items.py
sfc-gh-mvashishtha: I don't think any of these series are empty. `native_pd.Series(index=["a"])` has length 1.

tests/integ/modin/series/test_items.py
sfc-gh-nkrishna: Oh didn't realize that util function existed, neat!

tests/integ/modin/series/test_items.py
sfc-gh-nkrishna: Will add

src/snowflake/snowpark/modin/plugin/extensions/series_overrides.py
sfc-gh-nkrishna: Added a test, I experimented with the parameter set to True and it takes around 5-6 for `PARTITION_SIZE` and about 13 seconds for `PARTITION_SIZE * 2 + 1`. Set to False is takes significantly longer in both cases so I think having the parameter at True here is okay especially given the warning we provide.

tests/integ/modin/series/test_items.py
```diff
@@ -0,0 +1,91 @@
...
+    "size",
+    [
+        PARTITION_SIZE - 1,
+        PARTITION_SIZE,
+        PARTITION_SIZE + 1,
+        PARTITION_SIZE * 2,
+        PARTITION_SIZE * 2 + 1,
+    ],
+)
+@pytest.mark.skipif(running_on_public_ci(), reason="slow test")
```
sfc-gh-mvashishtha, line 74: Can we skip all the tests except `PARTITION_SIZE * 2`? Skipped tests are not useful, but OTOH we did [skip](https://github.com/snowflakedb/snowpandas/blame/dde2b5ddaaef5534da53a20422a2071e3ad9f2bf/tests/integ/modin/frame/test_itertuples.py#L153) the itertuples test cases that went through ~28 partitions. Let's try to just do 2 partitions in CI and skip the rest here?

sfc-gh-nkrishna: SG, I'll just add that test in CI and remove the rest

tests/integ/modin/series/test_items.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.mark.parametrize(
+    "series",
+    [
+        native_pd.Series(
+            ["bear", "bear", "marsupial"],
+            index=["panda", "polar", "koala"],
+        ),
+        native_pd.Series(data=["a"]),
```
sfc-gh-lmukhopadhyay, line 47: Maybe add cases with non-unique index and when data is a dict? For ex, `native_pd.Series(index=["a", "a", "a"])` and `native_pd.Series(data={'a': 1, 'b': 2, 'c': 3})`

src/snowflake/snowpark/modin/plugin/utils/frontend_constants.py
```diff
@@ -43,6 +43,12 @@
     "Scalar key incompatible with {} value"
 )
 
+SERIES_ITEMS_WARNING_MESSAGE = (
+    "Series.items may result in executing one more queries to fetch each row of this series. For better "
```
sfc-gh-lmukhopadhyay, line 47:
```suggestion
    "Series.items may result in executing one more query to fetch each row of this series. For better "
```

sfc-gh-lmukhopadhyay (commented):
1 nit and suggested tests to add, otherwise LGTM

sfc-gh-lmukhopadhyay (approved):

sfc-gh-mvashishtha (approved):

PR #2400, opened 2024-10-04
[SNOW-1625268] Fix pandas test with AST disabled.

sfc-gh-lspiegelberg:
Fixes Snowpark pandas API merge gates, after this PR they are green and subsequent PRs should maintain the CI passing.

Details:
- Refactored Column class `_emit_ast` pickup with new helper function to ensure identical behavior for binary operations.
- Fixed SQLCounter to use updated QueryListener protocol.
- Fixed `publicapi` decorator for the case where a function is passed keyword only args (i.e., no positional args).


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -7025,12 +7025,17 @@ def _apply_with_udtf_and_dynamic_pivot_along_axis_1(
                 pandas_labels=["partition_id"]
             )[0]
```
sfc-gh-lspiegelberg, line 7026: needed to fix this here because of an edge case test where multiple sessions where present.

sfc-gh-vbudati: just curious, why would there be a case where there's multiple sessions present?

sfc-gh-vbudati (approved):
Thanks Leonhard!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lspiegelberg: Not sure, I am following here the pandas tests, specifically `TestApplyLikeMethodsWithMultipleSessionsSnow1216902` where multiple sessions are tested.

PR #2401, opened 2024-10-04
SNOW-1720835 param protect thread-safe client side changes

sfc-gh-aalam:

   

   Fixes SNOW-1720835

   This PR adds param protection and uses `DummyLock` and `DummyThreadLocal` objects when thread-safe session is not enabled for session. This will be removed once thread-safe session is fully rolled out.


src/snowflake/snowpark/session.py
```diff
@@ -602,6 +599,26 @@ def __init__(
...
             ),
         )
+
+        self._thread_safe_session_enabled: bool = (
+            self._conn._get_client_side_session_parameter(
+                _PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION, False
+            )
+        )
+        self._thread_store = (
+            threading.local() if self._thread_safe_session_enabled else dict()
```
sfc-gh-yzou, line 609: instead of dict, maybe just use optional here, and only use the _thread_store value if the parameter is on

sfc-gh-aalam: I would then have to create something like `__analyzer` and `__conn` since the `_analyzer` is already created as a property now. I think since we have tests passing, it should be okay. I am using a `DummyThreadLocal` class now. Everything will be removed once I rollout parameter.

tests/unit/test_dataframe.py
```diff
@@ -146,10 +146,8 @@ def test_select_bad_input():
     )
 
 
-def test_join_bad_input():
-    mock_connection = mock.create_autospec(ServerConnection)
-    mock_connection._conn = mock.MagicMock()
-    session = snowflake.snowpark.session.Session(mock_connection)
+def test_join_bad_input(mock_server_connection):
```
sfc-gh-yzou, line 149: Do we have a test to make sure that when multi-threading is off, only one conn is created

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -323,6 +323,10 @@ def __init__(self, options: Optional[Dict[str, Any]] = None) -> None:
             self._oob_telemetry.disable()
         else:
             self._oob_telemetry.log_session_creation(self._connection_uuid)
+        # thread safe param protection
+        self._thread_safe_session_enabled = self._options.get(
```
sfc-gh-yzou, line 327: is that needed here? or should always be false for the mock?

sfc-gh-aalam: agreed

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -675,6 +680,19 @@ def warning(self, text: str) -> None:
         self.count += 1
 
 
+# TODO: SNOW-1720855: Remove DummyLock and DummyThreadLocal after the rollout
+class DummyLock:
+    def __enter__(self):
```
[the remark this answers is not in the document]
sfc-gh-aalam: added explanation.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -171,6 +171,21 @@ def __init__(
...
+        ) or self._lower_case_parameters.get(
+            "python_snowpark_enable_thread_safe_session", False
+        )
+        self._lock = (
+            threading.RLock() if self._thread_safe_session_enabled else DummyLock()
+        )
+        self._thread_store = (
+            threading.local()
+            if self._thread_safe_session_enabled
+            else DummyThreadLocal()
```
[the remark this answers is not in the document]
sfc-gh-aalam: DummyThreadLocal is just an empty class that lets us use the same interface as thread local without creating a thread local object. When we say multithreading is enabled, we are skipping creation of locks and thread-local objects.

src/snowflake/snowpark/mock/_connection.py
sfc-gh-aalam: We will read connection parameter given by user. If not, we will disable by default.

tests/integ/conftest.py
```diff
@@ -209,6 +209,8 @@ def session(
     session = (
         Session.builder.configs(db_parameters)
         .config("local_testing", local_testing_mode)
+        # Enable thread safe session for testing
+        .config("PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION", True)
```
sfc-gh-aalam, line 213: for tests, we are adding multithreading gates which will run all snowpark tests in enabled and disabled mode: https://github.com/snowflakedb/snowpark-python/pull/2336

tests/unit/test_dataframe.py
sfc-gh-aalam: let me add one.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -171,6 +173,22 @@ def __init__(
             except TypeError:
                 pass
 
+        # thread safe param protection
+        self._thread_safe_session_enabled = self._lower_case_parameters.get(
+            "python_snowpark_enable_thread_safe_session",
+            self._get_client_side_session_parameter(
+                "PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION", False
+            ),
+        )
```
sfc-gh-aalam, line 182: If connection parameter sets `PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION`, we read the value, otherwise we read the value from session.

tests/unit/test_dataframe.py
sfc-gh-aalam: added in `integ/test_multithreading.py::test_num_cursors_created`

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 ### Snowpark Python API Updates
 
-- Updated `Session` class to be thread-safe. This allows concurrent dataframe transformations, dataframe actions, UDF and store procedure registration, and concurrent file uploads.
+- Updated `Session` class to be thread-safe. This allows concurrent dataframe transformations, dataframe actions, UDF and store procedure registration, and concurrent file uploads. Setting parameter `"PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION"` when creating the `Session` object will enable thread-safe mode.
```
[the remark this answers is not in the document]
sfc-gh-yzou: actually, how can customer enable PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION, is the only way to enable this is to turn the parameter on at server side, or customer have other ways to turn this parameter on? i am trying to see shall we expose that to customer

src/snowflake/snowpark/mock/_connection.py
sfc-gh-yzou: i recall in the doc that supporting thread safe for local testing is out of scope at this moment, so i don't expect anyone can turn this on for mock. if that is the case, then i think this flag should always just be false, even if "PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION" is set to true. is there any code under mock that is currently using this parameter?

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-yzou: i think i see where the problem is now, we are also making the ServerConnection object thread safe, and the parameter here is a connection level parameter not session level parameter, that means customer will not be able to turn this on and off through the session object, they will have to turn it on an off through connection, and in general starting a new connection/session is required to turn this off.

sfc-gh-yzou: is setting through connection parameter a documented way for user to turn things on and off?

sfc-gh-yzou: and is _get_client_side_session_parameter also going to take care of the connection parameter, from the test, i believe it should be taken care also

tests/integ/conftest.py
sfc-gh-yzou: we should do it like how others are turning the feature on by flipping on the feature flag in our testing account, so that the change can be turned on and off without pr changes

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -675,6 +680,25 @@ def warning(self, text: str) -> None:
         self.count += 1
 
 
+# TODO: SNOW-1720855: Remove DummyLock and DummyThreadLocal after the rollout
+class DummyLock:
+    """This is a dummy lock that is used in place of threading.Rlock when multithreading is
+    disabled."""
+
+    def __enter__(self):
```
sfc-gh-yzou, line 688: This seems requiring the lock class to be used with WithStatement, otherwise it will throw an error. You might want to add a comment somewhere to notify the developer about this

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -323,6 +323,8 @@ def __init__(self, options: Optional[Dict[str, Any]] = None) -> None:
             self._oob_telemetry.disable()
         else:
             self._oob_telemetry.log_session_creation(self._connection_uuid)
+        # thread safe param protection
+        self._thread_safe_session_enabled = True
```
sfc-gh-yzou, line 327: shouldn't that be false?

src/snowflake/snowpark/session.py
```diff
@@ -611,6 +616,27 @@ def __init__(
                 DEFAULT_COMPLEXITY_SCORE_UPPER_BOUND,
             ),
         )
+
+        self._thread_store = (
```
sfc-gh-yzou, line 620: I see quite many place is using this, maybe add a simple funciton in the utils.py for create_thread_store and create_lock to use it at different places

src/snowflake/snowpark/_internal/server_connection.py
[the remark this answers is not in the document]
sfc-gh-yzou: I guess that should be fine, i was mainly want to keep the code the same as original, where you can init the cursor on the creation of SingleThreadLocal like what it does before, instead of wait until the call of cursor property.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: I documented in changelog.

I added this parameter mainly for testing since the locks cannot be changed during session.

src/snowflake/snowpark/_internal/utils.py
sfc-gh-aalam: I'll add methods for `acquire` and `release`. I expect this to be removed soon after we roll out thread-safe session for everyone.

src/snowflake/snowpark/session.py
sfc-gh-aalam: added

tests/integ/conftest.py
sfc-gh-aalam: adding that in the PR mentioned above.

src/snowflake/snowpark/_internal/utils.py
sfc-gh-yzou: so typically we will remove things after 1 or 2 quaters after rollout to be safe

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: remove changelog as per offline discussion

tests/integ/test_multithreading.py
```diff
@@ -32,6 +35,21 @@
 from tests.utils import IS_IN_STORED_PROC, IS_LINUX, IS_WINDOWS, TestFiles, Utils
 
 
+@pytest.fixture(scope="module")
```
sfc-gh-yzou, line 38: let's use a function scope here to avoid messing with other tests

tests/integ/test_multithreading.py
```diff
@@ -32,6 +35,21 @@
...
+def session(
```
sfc-gh-yzou, line 39: let's call this mutlithread_session instead of session

sfc-gh-yzou (approved):

tests/integ/test_multithreading.py
sfc-gh-aalam: renamed to threadsafe_session. this would not mess with others, right?

sfc-gh-jdu (approved):

PR #1542, opened 2024-05-09
SNOW-1296779: Remove sproc fallback code from Snowpark pandas

sfc-gh-nkrishna:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-1296779, SNOW-1254730

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   This PR removes unused sproc fallback code from Snowpark pandas, not that all APIs using fallback have been replaced with NotImplementedError.

PR #2404, opened 2024-10-04
NO-SNOW: fix sp-precommit test for large_query_breakdown and cte

sfc-gh-aalam:

   

   Fixes SNOW-NNNNNNN

   Test fails because it cannot detect fixture `sql_simplifier_enabled`. Reading the value from session to workaround the issue.

example: https://ci-dev-133.int.snowflakecomputing.com/job/RT-PC-TestWorker/203222/testReport/junit/(root)/t_python_sp_integ_sp/snowpark_test_large_query_breakdown_sql/

### DIFF
--- a/tests/integ/utils/sql_counter.py
+++ b/tests/integ/utils/sql_counter.py
@@ -2,6 +2,7 @@
 #
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
+import functools
 import inspect
 import os
 import re
@@ -11,7 +12,6 @@ import traceback
 from typing import Dict, List, Optional, Union
 
 import pytest
-from decorator import decorator
 
 from snowflake.snowpark.query_history import QueryRecord
 from snowflake.snowpark.session import Session
@@ -364,9 +364,7 @@ class SqlCounter:
         self.session = None
 
 
-@decorator
 def sql_count_checker(
-    func,
     no_check=None,
     high_count_expected=False,
     high_count_reason=None,
@@ -376,17 +374,8 @@ def sql_count_checker(
     udf_count=None,
     udtf_count=None,
     union_count=None,
-    *args,
-    **kwargs,
 ):
     """SqlCounter decorator that automatically validates the sql counts when test finishes."""
-    sql_counter = SqlCounter(
-        no_check=no_check,
-        log_stack_trace=False,
-        high_count_expected=high_count_expected,
-        high_count_reason=high_count_reason,
-    )
-
     all_args = inspect.getargvalues(inspect.currentframe())
     count_kwargs = {
         key: value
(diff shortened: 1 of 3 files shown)


sfc-gh-aalam:
https://ci-dev-142.int.snowflakecomputing.com/job/PythonStoredProcBuildPrecommitTest/1366/

sfc-gh-aling (approved):

tests/integ/utils/sql_counter.py
```diff
@@ -364,9 +364,7 @@ def _mark_as_dead(self):
         self.session = None
 
 
-@decorator
 def sql_count_checker(
-    func,
```
sfc-gh-aalam, line 369: updating the implementation because `decorator` package is not available in SP

sfc-gh-jrose (approved):

tests/integ/utils/sql_counter.py
```diff
@@ -376,17 +374,8 @@ def sql_count_checker(
     udf_count=None,
     udtf_count=None,
     union_count=None,
-    *args,
```
sfc-gh-evandenberg, line 379: The failure 

"sql_count_checker() got an unexpected keyword argument 'fallback_count'"

happens because the *args, **kwargs are removed here.

Not sure if that was intentional.  When I add them back, then the test_sql_counter.py tests pass.

tests/integ/utils/sql_counter.py
```diff
@@ -395,15 +384,29 @@ def sql_count_checker(
...
-        try:
-            sql_counter.close()
-        except Exception:
-            pass
-    return result
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            sql_counter = SqlCounter(
+                no_check=no_check,
```
sfc-gh-evandenberg, line 391: It looks like all tests that contain 'fallback_count=#' are disabled from being counted by pytest.mark.xfail:

@pytest.mark.xfail(
    reason="SNOW-1336091: Snowpark pandas cannot run in sprocs until modin 0.28.1 is available in conda",
    strict=True,
    raises=RuntimeError,
)

but when you actually run the tests, they fail with some other error such as NotImplementedError.

I believe a long time ago we would go through a 'fallback' to run the code using pandas within a sproc.  However, it looks like this was removed by https://github.com/snowflakedb/snowpark-python/pull/1542/files .  I'd recommend to check with @sfc-gh-nkrishna here.  Perhaps the disabled tests need to be re-evaluated to validate not implemented error rather than skipped in this way.

sfc-gh-evandenberg (commented):
Please see comments regarding failures:

sql_count_checker() got an unexpected keyword argument 'fallback_count'

tests/integ/utils/sql_counter.py
sfc-gh-aalam: Thanks for this. I assumed these were `args` and `kwargs` for the `func` we are decorating so I moved them down to `wrapper`. I moved it back up now.

sfc-gh-nkrishna (approved):
LGTM

PR #2405, opened 2024-10-07
SNOW-1418533 handle dropping temp objects in post actions

sfc-gh-aalam:

   

   Fixes SNOW-1418533

   This PR updates the way we generate name for temp objects. This is done to protect unexpected behavior when query generated by snowflake plan is as follows:
   i. Create temp object
   ii. Build sql query involving the temp object
   iii. Drop temp object in `post_actions`.

   When a dataframe with queries that create the same temp object and drop the same temp object, when run using multiple threads might cause a failure if one thread has not started working on temp object but a different thread has already dropped the temp object.

   In this PR, we update temp name creation for those objects that are dropped in a subsequent post action. We add `temp_name_placeholder` in `Query` class and generate a temp object name at query submission stage similar to `query_id_placeholder`.


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1673,6 +1678,7 @@ def __init__(
         *,
         query_id_place_holder: Optional[str] = None,
         is_ddl_on_temp_object: bool = False,
+        temp_name_place_holder: Optional[Tuple[str, TempObjectType]] = None,
```
sfc-gh-jdu, line 1681: is it possible that you can have two temp_name_place_holders, like joining two dataframes with temp tables?

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: a join statement when resolve can contain multiple temp_name_place_holders, but currently, when we generate temp object like temp tables or temp file format, we create them one at a time.

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@
 
 ### Snowpark Python API Updates
 
+- Updated `Session` class to be thread-safe. This allows concurrent dataframe transformations, dataframe actions, UDF and store procedure registration, and concurrent file uploads.
```
sfc-gh-yzou, line 58: this message is not clear about the change here. Please update the change log to reflect the actual change in this pr, and the reason to add this to help making the session class to be safe

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -731,7 +730,7 @@ def large_local_relation_plan(
         source_plan: Optional[LogicalPlan],
         schema_query: Optional[str],
     ) -> SnowflakePlan:
-        temp_table_name = random_name_for_temp_object(TempObjectType.TABLE)
+        temp_table_name = f"temp_name_placeholder_{generate_random_alphanumeric()}"
```
sfc-gh-yzou, line 733: isn't that going to impact the df.queries? not just people will see a different table, but is people going to see a non-sense temp table name called "temp_name_placeholder_xxx"?

sfc-gh-yzou: Is large_local_relation_plan it only case we will have internally created query scoped temp table?

sfc-gh-yzou: A more efficient way for those is probably making the temp table session scoped, and let @sfc-gh-jdu 's temp table clean up to handle the table drop off, i guess we can do this as future improvement

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -645,10 +648,20 @@ def get_result_set(
                 final_queries = []
                 last_place_holder = None
                 for q in main_queries:
+                    query = q.sql
+                    if q.temp_name_place_holder:
```
sfc-gh-yzou, line 652: instead of doing it here, might be better to do it at the execute_queries when calling plan_compiler to get the final query, so that this can be applied to all user facing queries

tests/integ/test_multithreading.py
```diff
@@ -633,6 +641,109 @@ def change_config_value(session_):
     )
 
 
+@pytest.mark.xfail(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="local testing does not execute sql queries",
+    run=False,
+)
+def test_temp_name_placeholder_for_sync(threadsafe_session):
+    from snowflake.snowpark._internal.analyzer import analyzer
```
sfc-gh-yzou, line 650: I just recall, do we have any multi-threading tests with the new query compilation stage?

src/snowflake/snowpark/async_job.py
```diff
@@ -405,8 +407,12 @@ def result(
             else:
                 raise ValueError(f"{async_result_type} is not supported")
         for action in self._post_actions:
+            query = action.sql
+            if self._placeholders:
```
sfc-gh-yzou, line 411: why do we need a special handling for async job ?

sfc-gh-aalam: because when dropping temp objects after async job is done, we need to drop the correct temp object. This information needs to be propagated to async job as well.

tests/integ/test_multithreading.py
sfc-gh-aalam: we do not at the moment

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1719,6 +1765,13 @@ def __init__(
             if query_id_place_holder
             else f"query_id_place_holder_{generate_random_alphanumeric()}"
         )
+        # This is a temporary workaround to handle the case when a snowflake plan is created
```
sfc-gh-yzou, line 1768: i am not sure if that is a temporary workaround. The sessions scoped temp table could be a long term solution, but it may have other draw back also, you can mention this as an alternative solution below

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1719,6 +1765,13 @@ def __init__(
...
+        # in the following way in a multi-threaded environment:
+        # 1. Create a temp object
+        # 2. Use the temp object in a query
+        # 3. Drop the temp object
```
sfc-gh-yzou, line 1772: this comment explained the scenario that needs that field, but not really explained how this field can be used to solve the problem, can you add some more detailed comment here

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -666,6 +669,7 @@ def get_result_set(
                     num_statements=len(main_queries),
                     params=params,
                     ignore_results=ignore_results,
+                    async_post_actions=post_actions,
```
sfc-gh-yzou, line 672: is that fixing a bug?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1711,6 +1756,7 @@ def __init__(
         *,
         query_id_place_holder: Optional[str] = None,
         is_ddl_on_temp_object: bool = False,
+        temp_name_place_holder: Optional[Tuple[str, TempObjectType]] = None,
```
sfc-gh-yzou, line 1759: let's call this temp_obj_name_placeholder to the more clear

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@
 
 #### Improvements
 
-- Disables sql simplification when sort is performed after limit. 
+- Disables sql simplification when sort is performed after limit.
```
sfc-gh-yzou, line 18: @sfc-gh-aalam do we have a ticket to track the documentation for multi-threading session object, i think you will want to mention this behavior in the doc on release

tests/integ/test_multithreading.py
sfc-gh-yzou: @sfc-gh-aalam can you get a ticket to add the tests? we should make sure all things are tested out

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: yes indeed.

sfc-gh-aalam: https://github.com/snowflakedb/snowpark-python/pull/2405/files#diff-a7f9cc0095b005bbd89b9100dc2d75b6b26c7ad70923bc502917ee46c1989bb4R449-R452

tests/integ/test_multithreading.py
sfc-gh-aalam: https://snowflakecomputing.atlassian.net/browse/SNOW-1758742

CHANGELOG.md
sfc-gh-aalam: created: https://snowflakecomputing.atlassian.net/browse/SNOW-1758759

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: earlier it would pull post_actions from snowflake plan which may not have drop table commands coming from large query breakdown

sfc-gh-aalam:
https://ci-dev-142.int.snowflakecomputing.com/job/PythonStoredProcBuildPrecommitTest/1412/

sfc-gh-yzou (approved):

PR #2413, opened 2024-10-08
SNOW-1727534: Implement `Resampler.indices`

sfc-gh-nkrishna:
SNOW-1727534

This PR adds support for `Resampler.indices`.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12260,8 +12266,9 @@ def resample(
 
         Returns
         -------
-        SnowflakeQueryCompiler
-            Holds an ordered frame with the result of the resample operation.
+        SnowflakeQueryCompiler or collections.defaultdict[Hashable, list]
+            Holds an ordered frame with the result of the resample operation, or a dictionary
```
sfc-gh-helmeleegy, line 12270:
```suggestion
            A SnowflakeQueryCompiler that holds an ordered frame with the result of the resample operation,
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12260,8 +12266,9 @@ def resample(
...
+            if resample_method is 'indices'.
```
sfc-gh-helmeleegy, line 12271:
```suggestion
            or a dictionary if resample_method is 'indices'.
```

sfc-gh-helmeleegy: This is just to make it clear that we are not returning a SnowflakeQueryCompiler that holds a dictionary.

sfc-gh-helmeleegy (approved):
Nice work. LGTM. Thanks!

tests/integ/modin/resample/test_resample.py
```diff
@@ -19,7 +18,24 @@
...
+        "sum",
+        "std",
+        "var",
+        "count",
+        "size",
+        "first",
+        "last",
+        "quantile",
+        "nunique",
+    ],
```
sfc-gh-jjiao, line 37: for my education, what is the reason we duplicate the content of IMPLEMENTED_AGG_METHODS  here?  but the `IMPLEMENTED_AGG_METHODS` still defined `src/snowflake/snowpark/modin/plugin/_internal/resample_utils.py` and has the same content, right?

sfc-gh-nkrishna: We add "indices" to `IMPLEMENTED_AGG_METHODS` b/c in `snowflake_query_compiler.py::resample` we have a check like follows:

```
elif resample_method in IMPLEMENTED_AGG_METHODS:
      <perform resample>
else:
            ErrorMessage.not_implemented(
                f"Resample Method {resample_method} has not been implemented."
            )
```

sfc-gh-nkrishna: But you bring up a good point, it's a bit clunky to relist all of these, so instead did this in the test:

```
# Parametrize on all IMPLEMENTED_AGG_METHODS except 'indices' which is tested in a separate file
agg_func = pytest.mark.parametrize("agg_func", IMPLEMENTED_AGG_METHODS[:13])
```

sfc-gh-jjiao: Thanks @sfc-gh-nkrishna for the update. I did later realize I missed the line you added "indices" to IMPLEMENTED_AGG_METHODS.  I think `IMPLEMENTED_AGG_METHODS[:13]` might be prone to make the test failed if others update the IMPLEMENTED_AGG_METHODS ,  we need to skip some agg methods for the rest, maybe consider what I suggested in my updated comment? 

```
agg_func = pytest.mark.parametrize(
    "agg_func",
     list(filter(lambda x: x not in  ["indices"], IMPLEMENTED_AGG_METHODS))
 )
```

tests/integ/modin/resample/test_resample.py
```diff
@@ -19,7 +19,8 @@
 )
 from tests.integ.utils.sql_counter import SqlCounter, sql_count_checker
 
-agg_func = pytest.mark.parametrize("agg_func", IMPLEMENTED_AGG_METHODS)
+# Parametrize on all IMPLEMENTED_AGG_METHODS except 'indices' which is tested in a separate file
+agg_func = pytest.mark.parametrize("agg_func", IMPLEMENTED_AGG_METHODS[:13])
```
sfc-gh-jjiao, line 23: are we only skipping some specific agg functions for the test?

would it be better to filter out specific functions than relies on the first 12 (in case other update the IMPLEMENTED_AGG_METHODS later ) 

```
agg_func = pytest.mark.parametrize(
    "agg_func",
     list(filter(lambda x: x not in  ["indices"], IMPLEMENTED_AGG_METHODS))
 )
```

sfc-gh-nkrishna: Good point, done!

sfc-gh-helmeleegy: Perhaps another option is to IMPLEMENTED_AGG_PROPERTIES separate from IMPLEMENTED_AGG_METHODS?

sfc-gh-nkrishna: For now since the only property currently implemented is `indices`, I'll go with Jialin's suggestion here. I don't think it matters too much here naming-wise since the code paths for other agg methods and 'indices' are quite similar.

sfc-gh-jjiao (approved):
LGTM, Thanks!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12394,6 +12410,7 @@ def resample(
                     numeric_only=resample_method_kwargs.get("numeric_only", False),
                     is_series_groupby=is_series,
                 )
+
```
sfc-gh-joshi, line 12413:
```suggestion
```

sfc-gh-joshi (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: This was intentional – there is a big block of case statements right before this so having the extra line helps with readability :)

PR #2420, opened 2024-10-08
Snowpark IR table functions

sfc-gh-oplaton:

   

   Fixes SNOW-1491303, SNOW-1621201, SNOW-1629946, SNOW-1621208

Pick up the [model changes](https://github.com/snowflakedb/snowflake/pull/217938) from the corresponding server-side work.

Support `DataFrame.join_table_function`.

Support `DataFrame.select` with `TableFunctionCall` parameters.

Switch `functions.explode`, `explode_outer`, `flatten` and `functions.table_function` to look like built-in function. We don't use  the fact that they're table functions anywhere yet (as opposed to scalar functions). We can reintroduce the distinction easily, if necessary.

Simplify `Session.table_function` by taking advantage of indirect function calls.

Support `TableFunctionCall` and its member functions.

Introduce `build_call_table_function_apply` and `build_indirect_table_fn_apply` for the specialized cases that require the corresponding entities.
Remove `build_session_table_fn_apply` and `build_table_fn_apply`.
Update type annotations.

There are instances that require accumulating a `proto.Expr` (similar to much of the column and built-in function functionality) until the root value can be assigned in the context of an `AstBatch`. Introduce a `build_intermediate_stmt` which takes any Python object that has an `_ast` attribute and assigns the value in a statement if necessary.

Random thought: The IR client captures all built-in function calls by unqualified name. We should fix this at some point.


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -386,53 +386,34 @@ def build_udtf_apply(
     build_fn_apply_args(ast, *args, **kwargs)
 
 
-def build_session_table_fn_apply(
+def build_call_table_function_apply(
     ast: proto.Expr,
```
sfc-gh-lspiegelberg, line 390: is there a reason why the alternate syntax `["database", "schema", "entity"]` is not supported anymore? It seems prevalent throughout snowpark, so maybe we should keep supporting it here?

sfc-gh-lspiegelberg: nit: I actually liked the `session` being part of the function name, it makes it clear what the call origin is. Unless this is a generalization...

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -386,53 +386,34 @@ def build_udtf_apply(
...
-        ast.apply_expr.fn.table_fn.call_type.WhichOneof("variant") is not None
-    ), f"Explicitly set the call type before calling this function {str(ast.apply_expr.fn.table_fn)}"
-    if not expr.fn.table_fn.call_type.table_fn_call_type__session_table_fn:
-        assert (
-            name is not None
-        ), f"Table function name must be provided {str(ast.apply_expr.fn.table_fn)}"
-        _set_fn_name(name, expr.fn.table_fn)
-    with_src_position(expr.fn.table_fn)
-    build_fn_apply_args(ast, *args, **kwargs)
+    if hasattr(func, "_ast_stmt"):
```
sfc-gh-lspiegelberg, line 408: I would actually prefer to have this more explicit, i.e. no indirect attribute lookup but rather:
```
if isinstance(func, TableFunctionCall):
   stmt = func._ast_stmt.
```
that is much easier to read. You can use local imports to prevent circular import issues. Much preferred to the `hasattr` solution.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -386,53 +386,34 @@ def build_udtf_apply(
...
+        stmt = func._ast_stmt
+        fn_expr = expr.fn.indirect_table_fn_call_ref
+        fn_expr.id.bitfield1 = stmt.var_id.bitfield1
+    else:
+        fn_expr = expr.fn.indirect_table_fn_name_call
+        _set_fn_name(func, fn_expr)
+    with_src_position(fn_expr)
```
sfc-gh-lspiegelberg, line 415: the `with_src_position` should be moved outside of this helper function, i.e.
```
expr = with_src_position(...)
build_indirect_table_fn_apply(expr, ...)
```
This helper func should only be concerned with building up the expr (without symbol/src extract)

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -1188,3 +1169,13 @@ def build_udtf(
         t = ast.kwargs.add()
         t._1 = k
         build_expr_from_python_val(t._2, v)
+
+
+def build_intermediate_stmt(ast_batch: AstBatch, o: Any) -> None:
+    if hasattr(o, "_ast_stmt"):
```
sfc-gh-lspiegelberg, line 1175: can we add docstring and make the name maybe more specific?

Similar to the comment above, can we avoid the `hasattr` pattern here and make things more explicit?

src/snowflake/snowpark/session.py
```diff
@@ -2106,48 +2106,15 @@ def table_function(
...
-                )
-            elif isinstance(func_name, str):
-                build_session_table_fn_apply(
-                    expr, func_name, *func_arguments, **func_named_arguments
-                )
-            elif isinstance(func_name, list):
-                build_session_table_fn_apply(
-                    expr, func_name, *func_arguments, **func_named_arguments
-                )
+            ast = with_src_position(stmt.expr.sp_session_table_function, stmt)
```
sfc-gh-lspiegelberg, line 2111: why not `func =  with_src_position(stmt.expr.sp_session_table_function.func, stmt)`, and then pass `func` instead of `ast.func` to build_indirect_table_fn_apply?

sfc-gh-lspiegelberg (commented):
looks good, some nits & open question regarding intermediate stmt. Great to get this done soon!

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: This is a misleading diff: `build_session_table_fn_apply` is gone altogether, and `build_call_table_function_apply` is a new thing. They just happen to share similar surrounding text.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: Multiple types are involved, though. `Callable` objects created from `functions.table_function` also have this property.

Which one do you prefer?

1. Explicit type enumeration (`if isinstance(func, (TableFunctionCall, Callable)):`)
2. Just look up the desired field

sfc-gh-lspiegelberg: I think 1. would be easier, thinking here of maintenance. If we ever wanted to support a new entity, very clear what to do/which branch logic takes. Looking up a potentially existing field seems more confusing imho.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1374,6 +1376,10 @@ def select(
                         f"Called '{table_func.user_visible_name}' and '{e.user_visible_name}'."
                     )
                 table_func = e
+                if _emit_ast and ast:
```
sfc-gh-evandenberg, line 1379:
```suggestion
                if ast:
```
_emit_ast must be True if ast is not None.  The other branches are inconsistent here as well.

tests/ast/data/DataFrame.join_table_function.test.DISABLED
```diff
@@ -0,0 +1,40 @@
+## TEST CASE
```
sfc-gh-evandenberg, line 1: Should this test be enabled with this pr?

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: Done

src/snowflake/snowpark/dataframe.py
sfc-gh-lspiegelberg: This may be confusing, we sometimes use in the code `ast=None` to auto-fill the ast (e.g., in the context of Column/Literals). Therefore we need both `_emit_ast` and `ast` in the check. I agree this is not great, we should clean it up and simplify.

tests/ast/data/DataFrame.join_table_function.test.DISABLED
sfc-gh-lspiegelberg: +1

sfc-gh-oplaton: Still having trouble with local tests. I'm working on finding a way to test this with integration tests.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: Oh, and with respect to `functions.call_table_function`, I think there's no technical reason why it couldn't take a `database.schema.function` name passed as a list, and the AST supports this internally, but the API itself is inconsistent with the rest of Snowpark. I won't try to fix that in this change.

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: I think this pattern matches what `build_builtin_fn_apply` and other sibling functions do. We should establish a uniform convention and follow it everywhere.

Also, if a convention requires all callers to do the exact same work everywhere, maybe the work does belong inside the callee.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1374,6 +1376,10 @@ def select(
...
+                    build_intermediate_stmt(self._session._ast_batch, table_func)
```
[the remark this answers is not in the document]
sfc-gh-oplaton: Added a few remarks to the docstring of `add_intermediate_stmt`.

src/snowflake/snowpark/session.py
sfc-gh-oplaton: For consistency with all other occurrences of `ast = with_src_position(stmt.expr.<<sp_specific_variant>>, stmt)` in the Snowpark codebase.

sfc-gh-lspiegelberg (approved):

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-oplaton: Done.

tests/ast/data/DataFrame.join_table_function.test.DISABLED
sfc-gh-oplaton: Will continue working on enabling these tests in some form (integration test) separately from this PR.

sfc-gh-evandenberg: Okay, work on separate from this pr.

sfc-gh-evandenberg (approved):
LGTM

src/snowflake/snowpark/dataframe.py
sfc-gh-oplaton: Will not take the (valid) suggestion yet. We have cleanup to do around the logic that enables/disables AST creation.

PR #2422, opened 2024-10-09
SNOW-1692063 - Preparatory updates to support more numpy ufunc mappings

sfc-gh-jkew:
Expands on the the earlier support for ufuncs by making the ufunc mapping clearer, and adding telemetry support. Another CL will add log support.

* Add support for subtract, multiply, divide, and true_divide ufuncs
* Fill out the list of possible ufuncs from https://numpy.org/doc/stable/reference/ufuncs.html
* Unimplemented ufuncs are marked with NotImplemented for visibility
* Add __array_ufunc__ to telemetry to get some usage statistics


tests/integ/modin/test_numpy.py
```diff
@@ -68,34 +68,46 @@ def test_logical_operators():
...
         assert_array_equal(np.array(snow_result), np.array(pandas_result))
 
     with SqlCounter(query_count=1):
         # Test binary numpy ufunc
-        snow_result = np.add(snow_df["A"], snow_df["A"])
-        pandas_result = np.add(pandas_df["A"], pandas_df["A"])
-        assert_array_equal(np.array(snow_result), np.array(pandas_result))
+        snow_result = np_ufunc(snow_df["A"], snow_df["B"])
+        pandas_result = np_ufunc(pandas_df["A"], pandas_df["B"])
+        assert_almost_equal(np.array(snow_result), np.array(pandas_result), 3)
```
sfc-gh-mvashishtha, line 92: nit

```suggestion
        assert_almost_equal(np.array(snow_result), np.array(pandas_result), decimal=3)
```

sfc-gh-mvashishtha (commented):
Great! I'm leaving two minor comments.

Great branch name.

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
```diff
@@ -119,9 +119,87 @@ def map_to_bools(inputs: Any) -> Any:
...
+# well to provide the reader with an understanding of the current
+# breadth. If a new ufunc is defined by numpy the absence of that
+# function from this list also implies that is it NotImplemented
+#
+# Numpy ufunc Reference
 # https://numpy.org/doc/stable/reference/ufuncs.html
 numpy_to_pandas_universal_func_map = {
+    # Math functions
     "add": lambda obj, inputs, kwargs: obj.__add__(*inputs, **kwargs),
+    "subtract": lambda obj, inputs, kwargs: obj.__sub__(*inputs, **kwargs),
```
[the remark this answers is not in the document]
sfc-gh-jkew: Good catch. I think I should fix it in this PR. I think the kwargs here should be mostly ignored possibly.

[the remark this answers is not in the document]
sfc-gh-jkew: Yeah. I believe the right behavior should be to ignore kwargs, but maybe throw an error if out is used ( since that type of operation would not be allowed w/ snowpandas.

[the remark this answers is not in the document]
sfc-gh-jkew: BigFrames will return NotImplemented if kwargs is > 0. PySpark Pandas will return NotImplemented if out is anything other than None and also ignores kwargs.

Overall this is a really good catch.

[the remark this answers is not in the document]
sfc-gh-jkew: @sfc-gh-mvashishtha take a look at the solution I pushed. let me know if it makes sense.

tests/integ/modin/test_numpy.py
```diff
@@ -108,6 +109,17 @@ def test_np_ufunc_notimplemented():
...
 
 
+@sql_count_checker(query_count=0)
+def test_np_ufunc_with_out_notimpl():
+    data = {
+        "A": [1],
+    }
+    snow_df = pd.DataFrame(data)
+    with pytest.raises(TypeError):
+        # heaviside is unlikely to be implemented any time soon
```
sfc-gh-mvashishtha, line 119:
```suggestion
```

tests/integ/modin/test_numpy.py
```diff
@@ -98,6 +98,7 @@ def test_np_ufunc_arithmetic_operators(np_ufunc):
         assert_array_equal(np.array(snow_result), np.array(pandas_result))
 
 
+@sql_count_checker(query_count=2)
```
sfc-gh-mvashishtha, line 101: FYI: I was curious about why we would have any queries at all. It turns out the numpy error message here will print representations of both the inputs, so we execute the same query twice to print `df["A"]`.

sfc-gh-mvashishtha (approved):
LGTM apart from a minor suggestion. Thank you!

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: SGTM

tests/integ/modin/test_numpy.py
sfc-gh-jkew: So this is actually coming from numpy error handling; and it is called from a C function when the __array_ufunc__ lookup ultimately failed. I don't think we necessarily need to fix this because:
1. It occurs on an error condition only
2. Its not in our code exactly; it's numpy formatting errors.

This was a good question though.

sfc-gh-dpetersohn (approved):
Very nice, LGTM. I imagine we will continue adding support for more of these in the future.

CHANGELOG.md
```diff
@@ -43,6 +43,8 @@
 - Added support for `DataFrame.tz_localize` and `Series.tz_localize`.
 - Added support for `DataFrame.tz_convert` and `Series.tz_convert`.
 - Added support for applying Snowpark Python functions (e.g., `sin`) in `Series.map`, `Series.apply`, `DataFrame.apply` and `DataFrame.applymap`.
+- Added support for `np.subtract`, `np.multiply`, `np.divide`, `np.true_divide`
```
sfc-gh-helmeleegy, line 46:
```suggestion
- Added support for `np.subtract`, `np.multiply`, `np.divide`, and `np.true_divide`.
```

CHANGELOG.md
```diff
@@ -43,6 +43,8 @@
...
+- Added support for tracking usages of `__array_ufunc__`
```
sfc-gh-helmeleegy, line 47:
```suggestion
- Added support for tracking usages of `__array_ufunc__`.
```

sfc-gh-helmeleegy (approved):
Looks great. Thanks for pushing our numpy coverage further!

sfc-gh-nkumar (approved):

PR #2424, opened 2024-10-09
SNOW-1675191: Fix reindex with non overlapping index types

sfc-gh-lmukhopadhyay:

   

   Fixes SNOW-1675191

   Fixed a bug in reindex when the new index is a Series of non overlapping type. The new index frame is converted to variant type. 

For example, the example below reindexing a df with int index with string index would previously error.
```
df = pd.DataFrame(
    {'one': [200, 300],
     'two': [200, 200]},
)
df.reindex(index=pd.Series(data=["A", "B"]))
```


src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -1154,6 +1154,29 @@ def join_on_index_columns(
...
+        frame: InternalFrame whose type needs to be converted.
+    Returns:
+        Frame with updated columns.
+    """
+    frame_ids = frame.data_column_snowflake_quoted_identifiers
+    frame_id_to_type_map = frame.quoted_identifier_to_snowflake_type(frame_ids)
+    frame_to_variant = {}
+    for frame_id in frame_ids:
+        frame_type = frame_id_to_type_map[frame_id]
+        if frame_type not in [VariantType, TimestampType]:
```
sfc-gh-nkrishna, line 1172: Why not do this for timestamp as well?

sfc-gh-nkrishna (commented):
Overall LGTM, some questions. Can you add a code example of the bug in the PR description?

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-lmukhopadhyay: We can actually do this for timestamp too, I added a test to confirm this. The case that doesn't work (which has negative tests already) is reindexing a datetime type frame with a non-matching index type.

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@
 - Fixed a bug where `Resampler` methods on timedelta columns would produce integer results.
 - Fixed a bug where `pd.to_numeric()` would leave `Timedelta` inputs as `Timedelta` instead of converting them to integers.
 - Fixed `loc` set when setting a single row, or multiple rows, of a DataFrame with a Series value.
+- Fixed a bug where `reindex` failed when the new index is a Series with non-overlapping types from the original index.
```
sfc-gh-rdurrani, line 70:
```suggestion
- Fixed a bug where `reindex` fails when the new index is a Series with non-overlapping types from the original index.
```
i think we use present tense for changelogs?

tests/integ/modin/frame/test_reindex.py
```diff
@@ -612,3 +613,65 @@ def test_reindex_with_lazy_index():
...
+@sql_count_checker(query_count=1, join_count=1)
+def test_reindex_str_index_with_tuple_index():
+    index = ["A", "B", "C", "D"]
+    native_df = native_pd.DataFrame(
+        {"one": [200, 200, 404, 404], "two": [200, 200, 404, 404]}, index=index
+    )
+    snow_df = pd.DataFrame(native_df)
+    nat_df = native_df.reindex(index=native_pd.Series(data=[("A", "B"), ("C", "D")]))
+    res_df = snow_df.reindex(index=pd.Series(data=[("A", "B"), ("C", "D")]))
+    # return res_df
```
sfc-gh-rdurrani, line 627: Please remove!

tests/integ/modin/frame/test_reindex.py
```diff
@@ -612,3 +613,65 @@ def test_reindex_with_lazy_index():
...
+@sql_count_checker(query_count=1, join_count=1)
+def test_reindex_int_index_with_tuple_index():
+    native_df = native_pd.DataFrame(
+        {"one": [200, 200, 404, 404], "two": [200, 200, 404, 404]}
+    )
+    snow_df = pd.DataFrame(native_df)
+    nat_df = native_df.reindex(index=native_pd.Series(data=[("A", "B"), ("C", "D")]))
+    idx = pd.Series(data=[("A", "B"), ("C", "D")])
+    res_df = snow_df.reindex(index=idx)
+    # return res_df
```
sfc-gh-rdurrani, line 640:
```suggestion
```

tests/integ/modin/frame/test_reindex.py
```diff
@@ -612,3 +613,65 @@ def test_reindex_with_lazy_index():
...
+@sql_count_checker(query_count=1, join_count=1)
+def test_reindex_int_index_with_str_index():
+    native_df = native_pd.DataFrame(
+        {"one": [200, 200, 404, 404], "two": [200, 200, 404, 404]}
+    )
+    snow_df = pd.DataFrame(native_df)
+    nat_df = native_df.reindex(index=native_pd.Series(data=["A", "C"]))
+    idx = pd.Series(data=["A", "C"])
+    res_df = snow_df.reindex(index=idx)
+    # return res_df
```
sfc-gh-rdurrani, line 653:
```suggestion
```

tests/integ/modin/index/test_df_series_creation_with_index.py
```diff
@@ -342,31 +344,36 @@ def test_create_with_index_as_data_and_series_as_index(
...
         ),  # all index values match
         (
             native_pd.Series(["A", "V", "D", "R"]),
             native_pd.Index([10, 20, 30, 40], name="none"),
+            True,
         ),  # no index values match
         (
             native_pd.Series([], name="empty series", dtype="int64"),
             native_pd.Index([], name="empty index", dtype="int64"),
+            False,
```
sfc-gh-rdurrani, line 369: Why are we not checking index type here?

tests/integ/modin/index/test_df_series_creation_with_index.py
```diff
@@ -501,6 +509,7 @@ def test_create_df_with_empty_df_as_data_and_index_as_index(native_df, native_in
         pd.DataFrame(snow_df, index=snow_index),
         native_pd.DataFrame(native_df, index=native_index),
         check_column_type=False,
+        check_index_type=False,
```
sfc-gh-rdurrani, line 512: Same here

tests/integ/modin/series/test_reindex.py
```diff
@@ -328,15 +325,26 @@ def perform_reindex(series):
...
+        {"prices": [100, 101, np.nan, 100, 89, 88]}, index=date_index
+    )
     date_index = pd.date_range("1/1/2010", periods=6, freq="D")
     snow_series = pd.Series(
         {"prices": [100, 101, np.nan, 100, 89, 88]}, index=date_index
     )
 
-    with pytest.raises(SnowparkSQLException, match=".*Timestamp 'A' is not recognized"):
-        snow_series.reindex(list("ABC")).to_pandas()
+    def perform_reindex(series):
```
sfc-gh-rdurrani, line 339: We could probably just do this as a lambda right? Both branches of the if statement lead to the same thing?

sfc-gh-rdurrani (changes requested):
Overall lgtm, left some nits.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -1169,7 +1169,7 @@ def convert_index_type_to_variant(
     frame_to_variant = {}
     for frame_id in frame_ids:
         frame_type = frame_id_to_type_map[frame_id]
-        if frame_type not in [VariantType, TimestampType]:
+        if frame_type != VariantType:
```
sfc-gh-nkrishna, line 1172: Could be wrong, but shouldn't this be since VariantType is a class:
```suggestion
        if not isinstance(frame_type, VariantType):
```

tests/integ/modin/index/test_df_series_creation_with_index.py
sfc-gh-lmukhopadhyay: After an empty index is converted to variant, the inferred_type is now `empty` instead of `integer`. We are currently not ensuring dtype, so would not returning the same index type in this case be a blocker?

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-lmukhopadhyay: That's right, that should be either VariantType() or use `isinstance`, thanks for catching that!

CHANGELOG.md
sfc-gh-lmukhopadhyay: I think I saw both but this change is more grammatically correct :)

sfc-gh-nkrishna (approved):
LGTM!

tests/integ/modin/index/test_reindex.py
```diff
@@ -168,7 +168,10 @@ def test_non_overlapping_datetime_index():
 def test_non_overlapping_different_types_index_negative_SNOW_1622502():
     date_index = pd.date_range("1/1/2010", periods=6, freq="D")
 
-    with pytest.raises(SnowparkSQLException, match=".*Timestamp 'A' is not recognized"):
+    with pytest.raises(
+        SnowparkSQLException,
+        match='.*Failed to cast variant value "A" to TIMESTAMP_NTZ',
```
sfc-gh-helmeleegy, line 173: I believe the `.*` is not needed here.

tests/integ/modin/frame/test_reindex.py
```diff
@@ -281,7 +281,8 @@ def test_reindex_index_non_overlapping_different_types_index_negative(self):
         )
 
         with pytest.raises(
-            SnowparkSQLException, match=".*Timestamp 'A' is not recognized"
+            SnowparkSQLException,
+            match='.*Failed to cast variant value "A" to TIMESTAMP_NTZ',
```
sfc-gh-helmeleegy, line 285: I believe the `.*` is not needed here.

sfc-gh-helmeleegy (approved):
LGTM, just a couple minot nits.

tests/integ/modin/index/test_reindex.py
sfc-gh-lmukhopadhyay: That's true, it can be removed

tests/integ/modin/index/test_df_series_creation_with_index.py
sfc-gh-rdurrani: Hmm - was the index type the same before? It would be nice to have the index type be correct, but I don't think it should be a blocker - maybe add it as a low/medium priority ticket as follow up work?

sfc-gh-rdurrani (approved):
LGTM!

tests/integ/modin/index/test_df_series_creation_with_index.py
sfc-gh-lmukhopadhyay: Good point! I created this ticket to keep track of this https://snowflakecomputing.atlassian.net/browse/SNOW-1740545

PR #2433, opened 2024-10-10
SNOW-1692063 - More functions for numpy compatibility

sfc-gh-jkew:
- logical comparison functions
- a few more math functions
- modified chain testing test to work with operations that result in mixed types


CHANGELOG.md
```diff
@@ -53,8 +53,9 @@
 - Added support for `DataFrame.tz_localize` and `Series.tz_localize`.
 - Added support for `DataFrame.tz_convert` and `Series.tz_convert`.
 - Added support for applying Snowpark Python functions (e.g., `sin`) in `Series.map`, `Series.apply`, `DataFrame.apply` and `DataFrame.applymap`.
-- Added support for `np.subtract`, `np.multiply`, `np.divide`, `np.true_divide`
-- Added support for tracking usages of `__array_ufunc__`
+- Added support for `np.subtract`, `np.multiply`, `np.divide`, and `np.true_divide`.
```
sfc-gh-nkumar, line 56: We have released 1.23.0. This will be moved to new release version. Please rebase to get updated changelog.

sfc-gh-nkumar (approved):
Please move changelog to new release version. Otherwise looks good.

CHANGELOG.md
```diff
@@ -53,8 +54,8 @@
...
```
sfc-gh-yzou, line 57: we probably want to keep the original word, since it has already be published

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Snowpark pandas API Updates
 
 #### New Features
+- Added support for `np.float_power`, `np.mod`, `np.remainder`, `np.greater`, `np.greater_equal`, `np.less`, `np.less_equal`, `np.not_equal`, and `np.equal`.
```
sfc-gh-yzou, line 12: maybe let's call this "added numpy compatibility support" to make a difference from the regular Snowpark pandas support

tests/integ/modin/test_numpy.py
```diff
@@ -69,9 +69,25 @@ def test_logical_operators():
 
 
 @pytest.mark.parametrize(
-    "np_ufunc", [np.add, np.subtract, np.divide, np.multiply, np.true_divide]
+    "np_ufunc",
+    [
+        np.add,
```
sfc-gh-yzou, line 74: For many of those, do we have any test with data that has NULL or NaN inside, i believe they might have some difference there, most about the mapping to Snowpark Pandas, for example, are they mapped to snowpark pandas missing value or actual NaN.
Maybe we can add some of those test in a different pr

sfc-gh-yzou (approved):

CHANGELOG.md
sfc-gh-jkew: done

PR #2434, opened 2024-10-10
SNOW-1678419: Fix sql simplifier enabled

sfc-gh-aalam:

   

   Fixes SNOW-1678419

- Fix bug in select statement where we ignored expressions that contain ScalarSubquery plan.
- Add test to check correctness of this case.


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -795,6 +795,7 @@ def sql_query(self) -> str:
             placeholder = f"{analyzer_utils.LEFT_PARENTHESIS}{self.from_._id}{analyzer_utils.RIGHT_PARENTHESIS}"
             self._sql_query = self.placeholder_query.replace(placeholder, from_clause)
         else:
+            self.analyzer.subquery_plans = []
```
sfc-gh-yzou, line 798: this is probably going to be problematic also, because this part can potentially be done during the middle of resolving, and if we clears the subquery_plans here, we might cause problem to the plan node that is actually for resolving

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -816,6 +817,19 @@ def sql_query(self) -> str:
                 else snowflake.snowpark._internal.utils.EMPTY_STRING
             )
             self._sql_query = f"{analyzer_utils.SELECT}{self.projection_in_str}{analyzer_utils.FROM}{from_clause}{where_clause}{order_by_clause}{limit_clause}{offset_clause}"
+
+            for plan in self.analyzer.subquery_plans:
+                for query in plan.queries[:-1]:
```
sfc-gh-yzou, line 822: also, if we look at the with_subqueries implementation in the analyzer.py it also handles the schema query, we will need to do the same thing, i think we might still need to do that at the analyzer level and just call the same function to propogate the field properly

tests/integ/test_simplifier_suite.py
```diff
@@ -205,6 +206,38 @@ def get_max_nesting_depth(query):
     )
 
 
+def test_in_with_subquery_multiple_query(session):
+    # multiple queries
+    original_threshold = analyzer.ARRAY_BIND_THRESHOLD
+    try:
+        analyzer.ARRAY_BIND_THRESHOLD = 2
+        df0 = session.create_dataframe([[1], [2], [5], [7]], schema=["a"])
+        df = session.create_dataframe(
```
sfc-gh-yzou, line 215: please search for SNOW-1678419 in the code base to enable the test that is currently disabled

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: The original reason why we handle `with_subqueries` is handle the condition when
- current plan's schema query = a subquery in subquery_plans

This does not apply to select statement since subquery plans for us can only come from filter expressions.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: you are right. I missed this possibility. I can read current value before resetting it and calling analyze.

tests/integ/test_simplifier_suite.py
sfc-gh-aalam: thanks. added

tests/integ/test_simplifier_suite.py
```diff
@@ -205,6 +206,38 @@ def get_max_nesting_depth(query):
...
+            schema=["a", "b", "c", "d"],
+        )
+        df_filter = df0.filter(col("a") < 3)
+        df_in = df.filter(~df["a"].in_(df_filter))
+        df_result = df_in.union_all(df_in).select("*")
+
+        """
+        'CREATE  OR  REPLACE  SCOPED TEMPORARY  TABLE SNOWPARK_TEMP_TABLE_B7KIT7YVL9("A" BIGINT NOT NULL , "B" STRING NOT NULL , "C" BIGINT NOT NULL , "D" BIGINT NOT NULL )', 'INSERT  INTO SNOWPARK_TEMP_TABLE_B7KIT7YVL9("A", "B", "C", "D") VALUES (?, ?, ?, ?)',
+        'WITH SNOWPARK_TEMP_CTE_5PCVK6WSYG AS ( SELECT  *  FROM ( SELECT "A", "B", "C", "D" FROM ( SELECT  *  FROM (SNOWPARK_TEMP_TABLE_B7KIT7YVL9))) WHERE NOT "A" IN (( SELECT "A" FROM ( SELECT  *  FROM (SNOWPARK_TEMP_TABLE_OYX4PWAYFQ)) WHERE ("A" < 3))))( SELECT  *  FROM (SNOWPARK_TEMP_CTE_5PCVK6WSYG)) UNION ALL ( SELECT  *  FROM (SNOWPARK_TEMP_CTE_5PCVK6WSYG))'
+        """
```
sfc-gh-aalam, line 226:
```suggestion
```

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yzou: so here is what i see how plan node with subquery works today, snowpark does resolve for every intermediate step, however, for node with subquery, it is not correct after resolve, https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/analyzer/analyzer.py#L780

this corrects the post-actions, queries as well as the schema queries. For select statement, it seems the same to me, where we also need to correct the schema query, otherwise the schema query will not be valid because it refers to a table that is not created. for example, following is the schema query for the select Statement 
```
 SELECT NOT "A" IN (( SELECT "A" FROM ( SELECT  *  FROM (SNOWPARK_TEMP_TABLE_3QWYPIQ4JV)) WHERE ("A" < 2))) AS "IN_RESULT" FROM ( SELECT "A", "B", "C", "D" FROM ( SELECT $1 AS "A", $2 AS "B", $3 AS "C", $4 AS "D" FROM  VALUES (0 :: BIGINT, 'a' :: STRING, 0 :: BIGINT, 0 :: BIGINT)))
```

as you can see it is referring to a SNOWPARK_TEMP_TABLE_3QWYPIQ4JV that is not defined anywhere.

So the selectstatement should be handled in the same way as how we handle the resolved snowflakeplane  (because selectstatement its-self treated as a resolved node), the problem today seems that we only handled the resolved SnowflakePlan, but not the selectStatement when the logical plan passed is a selectStatement here. Therefore, other than handling resolved SnowflakePlan here https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/analyzer/analyzer.py#L781, we should also handle the logical plan if it is a SelectStatement. The resolve step will record all subquery resolved

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -779,6 +779,31 @@ def projection_in_str(self) -> str:
             )
         return self._projection_in_str
 
+    def with_subqueries(self, subquery_plans: List[SnowflakePlan]) -> "SelectStatement":
+        """Update pre-actions, post-actions and schema to capture necessary subquery_plans
+        encountered during plan resolution."""
+        for plan in subquery_plans:
+            for query in plan.queries[:-1]:
+                if self.pre_actions is None:
+                    self.pre_actions = []  # pragma: no cover
```
sfc-gh-yzou, line 788: How can we have no cover for this line, most of the case , pre_actions should be None, right?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -779,6 +779,31 @@ def projection_in_str(self) -> str:
...
+                    self.post_actions = []  # pragma: no cover
+                if query not in self.post_actions:
+                    self.post_actions.append(query)
+
+            if (self._schema_query is not None) and (plan.schema_query is not None):
+                self._schema_query = self._schema_query.replace(
+                    plan.queries[-1].sql, plan.schema_query
+                )
+
+        if self._snowflake_plan is not None:
```
sfc-gh-yzou, line 802: i don't think you need to do this, all snowflake_plan should be handled during resolve right before you call 
```
if isinstance(logical_plan, SelectStatement):
               logical_plan.with_subqueries(self.subquery_plans)
```

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -779,6 +779,31 @@ def projection_in_str(self) -> str:
...
+                    self.pre_actions = []  # pragma: no cover
+                if query not in self.pre_actions:
+                    self.pre_actions.append(query)
+            for query in plan.post_actions:
+                if self.post_actions is None:
+                    self.post_actions = []  # pragma: no cover
+                if query not in self.post_actions:
+                    self.post_actions.append(query)
+
+            if (self._schema_query is not None) and (plan.schema_query is not None):
```
sfc-gh-yzou, line 797: should call self.schema_query for this, instead of calling the cached one, otherwise if the schema_query cache is None, when calling schema_query, we will still see the same problem.

tests/integ/test_cte.py
```diff
@@ -902,10 +902,6 @@ def test_window_function(session):
 
 
 def test_in_with_subquery_multiple_query(session):
-    if session._sql_simplifier_enabled:
-        pytest.skip(
```
sfc-gh-yzou, line 906: can you add a test in the regular test_dataframe to check the query generated without any new optimizations? and also check the schema query by calling df.schema

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -779,6 +779,8 @@ def resolve(self, logical_plan: LogicalPlan) -> SnowflakePlan:
 
         if self.subquery_plans:
             result = result.with_subqueries(self.subquery_plans)
+            if isinstance(logical_plan, SelectStatement):
```
sfc-gh-yzou, line 782: we probably should handle this for Selectable in general, not just SelectStatement

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -779,6 +779,31 @@ def projection_in_str(self) -> str:
...
 
+    def with_subqueries(self, subquery_plans: List[SnowflakePlan]) -> "SelectStatement":
+        """Update pre-actions, post-actions and schema to capture necessary subquery_plans
+        encountered during plan resolution."""
+        for plan in subquery_plans:
+            for query in plan.queries[:-1]:
+                if self.pre_actions is None:
+                    self.pre_actions = []  # pragma: no cover
+                if query not in self.pre_actions:
+                    self.pre_actions.append(query)
```
sfc-gh-yzou, line 790: this part of code seems repeated code block with the with_subqueries, and seems missing the API call handling also. can we have a utility function to handle the with_subqueries, maybe move the with_subqueries to a util file, and take a snowflakeplan or selectable as input and return a snowflakeplan or selectable

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 - Added support for 'Service' domain to `session.lineage.trace` API.
 - Added support for `copy_grants` parameter when registering UDxF and stored procedures.
 
+#### Bug Fixes
+
+- Fixed a bug where in expression values created using a dataframe would throw errors like `Object 'SNOWPARK_TEMP_TABLE_ABCDXYZ123' does not exist or not authorized`.
```
sfc-gh-yzou, line 14: This message is pretty vague, maybe let's more clear on the fix, like
```
fix the pre_action and post_action query propagation when in expression is used in selects.
```

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: It does not.

resolve() calls do_resolve()
in do_resolve(), we only recursively call resolve again for `child in logical_plan.children`. Since for selectables, `logical_plan.children = []`, it we don't call it for the underlying `SnowflakePlan`

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: This is how we create `schema_query` for different `Selectable` classes

1. `SelectableEntity`: `SELECT * FROM ENTITY`
2. `SelectSQL`: `sql_query`
3. `SetStatement`: build a schema query using `attributes`
4. `SelectSnowflakePlan`: `self.snowflake_plan.schema_query`
5. `SelectTableFunction`: `self.snowflake_plan.schema_query`
6. `SelectStatement`: 
```
        if self._schema_query:
            return self._schema_query
        if not self.projection:
            self._schema_query = self.from_.schema_query
            return self._schema_query
        self._schema_query = f"{analyzer_utils.SELECT}{self.projection_in_str}{analyzer_utils.FROM}({self.from_.schema_query})"
        return self._schema_query
```

for 1, 2, 3 schema_query is independent of subquery plans
for 4, 5 schema_query will be updated when we call `self._snowflake_plan = self._snowflake_plan.with_subqueries(subquery_plans)`

this only needs to be handled for `SelectStatement`

tests/integ/test_cte.py
sfc-gh-aalam: added a test but not sure what exactly to test with schema.

should I do something like
```
df_filter._plan.queries[-1] not in df_in._plan.schema_query
```

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: for `SnowflakePlan`, `with_subqueries` does not do an in-place replacement. It creates a new `SnowflakePlan` with updated values. As a result, first level of call after applying `with_subqueries` works fine but the next levels pull snowflake_plan from `Selectable._snowflake_plan`. Since is it already resolved, it does not populate `self.subquery_plans` inside analyzer. Therefore, we need this to update `_snowflake_plan` for the selectable.

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1239,6 +1265,16 @@ def limit(self, n: int, *, offset: int = 0) -> "SelectStatement":
 
         return new
 
+    def with_subqueries(self, subquery_plans: List[SnowflakePlan]) -> "SelectStatement":
+        super().with_subqueries(subquery_plans)
+        for plan in subquery_plans:
```
sfc-gh-yzou, line 1270: i see that other than the selectStatement, and setStatement. All rest just get the schema query from the attached snowflake plan. SetStatement also just get the attribute from the child snowflake plan directly. Therefore, probably only SelectStatement need a special handling here. Please comment it clearly, the code are quite confusing here for people do not understand how code generation works completely

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -779,6 +779,8 @@ def resolve(self, logical_plan: LogicalPlan) -> SnowflakePlan:
 
         if self.subquery_plans:
             result = result.with_subqueries(self.subquery_plans)
```
sfc-gh-yzou, line 781: add a comment here about what are we doing here

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -383,6 +383,32 @@ def referenced_ctes(self) -> Dict[WithQueryBlock, int]:
...
+    def with_subqueries(self, subquery_plans: List[SnowflakePlan]) -> "Selectable":
+        """Update pre-actions, post-actions and schema to capture necessary subquery_plans
+        encountered during plan resolution."""
+        for plan in subquery_plans:
+            for query in plan.queries[:-1]:
+                self.merge_into_pre_action(query)
+            for query in plan.post_actions:
+                self.merge_into_post_action(query)
+
+        if self._snowflake_plan is not None:
```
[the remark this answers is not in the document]
sfc-gh-aalam: I responded to your earlier comment about the same thing here: https://github.com/snowflakedb/snowpark-python/pull/2434#discussion_r1823514550

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: I actually think this is not needed for selectables at all. I wrote a summary here https://github.com/snowflakedb/snowpark-python/pull/2434#discussion_r1823386898 but if you notice, `schema_query` for `SelectStatement` is always made using `from_.schema_query`. In case of `SnowflakePlan`, default `schema_query` is made from given query so the adjustment is required.

tests/integ/test_dataframe.py
```diff
@@ -1278,6 +1278,33 @@ def test_join_left_outer(session):
...
+        df0 = session.create_dataframe([[1], [2], [3]], schema=["a"])
+        df1 = session.create_dataframe(
+            [[1, 11, 111], [2, 22, 222], [3, 33, 333]], schema=["a", "b", "c"]
+        )
+        df_filter = df0.filter(df0.a < 3)
+        df_in = df1.filter(~df1.a.in_(df_filter))
+
+        df_in = df_in.select("a", "b")
+        Utils.check_answer(df_in, [Row(3, 33)])
+        assert df_filter._plan.queries[-1].sql not in df_in._plan.schema_query
```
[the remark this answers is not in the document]
sfc-gh-aalam: I updated this based on what you suggested earlier.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -779,6 +779,9 @@ def resolve(self, logical_plan: LogicalPlan) -> SnowflakePlan:
...
+            # update the pre and post actions for selectable if it has subqueries
```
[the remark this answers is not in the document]
sfc-gh-aalam: done

tests/integ/test_dataframe.py
```diff
@@ -1278,6 +1278,34 @@ def test_join_left_outer(session):
...
+        df1 = session.create_dataframe(
+            [[1, 11, 111], [2, 22, 222], [3, 33, 333]], schema=["a", "b", "c"]
+        )
+        df_filter = df0.filter(df0.a < 3)
+        df_in = df1.filter(~df1.a.in_(df_filter))
+
+        df_in = df_in.select("a", "b")
+        Utils.check_answer(df_in, [Row(3, 33)])
+        # check that schema query does not depend on temp tables
+        assert "SNOWPARK_TEMP_TABLE_" not in df_in._plan.schema_query
```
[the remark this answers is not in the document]
sfc-gh-aalam: done

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -383,6 +383,44 @@ def referenced_ctes(self) -> Dict[WithQueryBlock, int]:
...
+
+    def with_subqueries(
+        self, subquery_plans: List[SnowflakePlan], snowflake_plan: SnowflakePlan
+    ) -> "Selectable":
+        """Update pre-actions, post-actions and schema to capture necessary subquery_plans
+        encountered during plan resolution. All updates are in-place.
+
+        Args:
+            subquery_plans: List of subquery plans encountered during plan resolution.
+            snowflake_plan: The snowflake plan which is created and updated using
```
sfc-gh-yzou, line 410: -> the snowflake corresponding to the fully resolved snowflake plan for the current selectable with the subquery plan handled

sfc-gh-yzou (approved):

PR #2435, opened 2024-10-10
SNOW-1727473: Test and fix timedelta handling in tests/integ/modin/frame: part 2/n.

sfc-gh-mvashishtha:
Fixes SNOW-1727473

Test and/or fix timedelta handling for the methods tested in tests/integ/modin/frame/, alphabetically from test_pct_change.py through test_set_index.py.

Make the following behavior changes:
- Fixed a bug where `DataFrame` and `Series` `pct_change()` would raise `TypeError` when input contained timedelta columns.
- Fixed a bug where `replace()` would sometimes propagate `Timedelta` types incorrectly through `replace()`. Instead raise `NotImplementedError` for `replace()` on `Timedelta`.
- Fixed a bug where `DataFrame` and `Series` `round()` would raise `AssertionError` for `Timedelta` columns. Instead raise `NotImplementedError` for `round()` on `Timedelta`.

Apart from those changes, just test that we can handle timedelta inputs correctly.


src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -1889,16 +1894,23 @@ def pct_change(
...
-        kwargs["axis"] = self._get_axis_number(kwargs["axis"])
+    kwargs["axis"] = self._get_axis_number(kwargs.get("axis", 0))
 
     # Attempting to match pandas error behavior here
     if not isinstance(periods, int):
         raise TypeError(f"periods must be an int. got {type(periods)} instead")
 
+    if (
+        kwargs["axis"] == 1
+        and any(is_timedelta64_dtype(t) for t in self._get_dtypes())
```
sfc-gh-azhan, line 1905: There is a `is_timedelta64_dtype` function in QC which is optimized for DatetimeTZDtype. Can you try to use that one?

sfc-gh-mvashishtha: Done.

sfc-gh-azhan (approved):

PR #2437, opened 2024-10-11
[SNOW-1731783] Refactor node query comparison for Repeated subquery elimination

sfc-gh-yzou:

   

SNOW-1731783

In the previous repeated subquery elimination, we updated the node comparison for SnowflakePlan for Selectable to 
```
    def __eq__(self, other: "SnowflakePlan") -> bool:
        if not isinstance(other, SnowflakePlan):
            return False
        if self._id is not None and other._id is not None:
            return isinstance(other, SnowflakePlan) and self._id == other._id
        else:
            return super().__eq__(other)

    def __hash__(self) -> int:
```
where the id is generated based on the query and query parameter, this means two node are treated as the same if they have same type and same query. This make sense when we do repeated subquery elimination, but not expected by other transformations.

Refactor the comparison to make sure that we only use the id comparison for repeated subquery eliminations, not for others.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/cte_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/cte_utils.py
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
 
 def find_duplicate_subtrees(root: "TreeNode") -> Set[str]:
     """
-    Returns a set containing all duplicate subtrees in query plan tree.
+    Returns a set of TreeNode encoded_id that indicates all duplicate subtrees in query plan tree.
     The root of a duplicate subtree is defined as a duplicate node, if
         - it appears more than once in the tree, AND
         - one of its parent is unique (only appear once) in the tree, OR
@@ -50,7 +50,6 @@ def find_duplicate_subtrees(root: "TreeNode") -> Set[str]:
     """
     id_count_map = defaultdict(int)
     id_parents_map = defaultdict(set)
-    # node_parents_map = defaultdict(set)
 
     def traverse(root: "TreeNode") -> None:
         """
@@ -157,6 +156,14 @@ def create_cte_query(root: "TreeNode", duplicated_node_ids: Set[str]) -> str:
 def encoded_query_id(
     query: str, query_params: Optional[Sequence[Any]] = None
 ) -> Optional[str]:
+    """
+    Encode the query and its query parameter into an id using sha256.
+
+
+    Returns:
+        If encode succeed, return the first 10 encoded value.
+        Otherwise, return None
+    """
     string = f"{query}#{query_params}" if query_params else query
     try:
         return hashlib.sha256(string.encode()).hexdigest()[:10]
@@ -168,9 +175,15 @@ def encoded_query_id(
 def encode_id(
     node_type_name: str, query: str, query_params: Optional[Sequence[Any]] = None
 ) -> str:
-    string = f"{query}#{query_params}" if query_params else query
-    try:
-        return hashlib.sha256(string.encode()).hexdigest()[:10] + node_type_name
-    except Exception as ex:
-        logging.warning(f"Encode SnowflakePlan ID failed: {ex}")
+    """
+    Encode given query, query parameters and the node type into an id.
+
+    If query and query parameters can be encoded successfully using sha256,
+    return the encoded query id + node_type_name.
+    Otherwise, generate a uuid.
+    """
+    query_id = encoded_query_id(query, query_params)
+    if query_id is not None:
+        return query_id + node_type_name
+    else:
         return str(uuid.uuid4())
--- a/src/snowflake/snowpark/_internal/analyzer/cte_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/cte_utils.py
@@ -4,6 +4,7 @@
 
 import hashlib
 import logging
+import uuid
 from collections import defaultdict
 from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence, Set, Tuple, Union
 
@@ -24,9 +25,7 @@ if TYPE_CHECKING:
     TreeNode = Union[SnowflakePlan, Selectable]
 
 
-def find_duplicate_subtrees(
-    root: "TreeNode",
-) -> Tuple[Set["TreeNode"], Dict["TreeNode", Set["TreeNode"]]]:
+def find_duplicate_subtrees(root: "TreeNode") -> Set[str]:
     """
     Returns a set containing all duplicate subtrees in query plan tree.
     The root of a duplicate subtree is defined as a duplicate node, if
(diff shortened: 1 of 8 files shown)


sfc-gh-aalam:
> This make sense when we do repeated subquery elimination, but not expected by other transformations.

Can you give an example of where this does not make sense?

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
```diff
@@ -144,16 +149,41 @@ def build_plan_to_query_map_in_post_order(root: "TreeNode") -> None:
...
     try:
         return hashlib.sha256(string.encode()).hexdigest()[:10]
     except Exception as ex:
         logging.warning(f"Encode SnowflakePlan ID failed: {ex}")
         return None
+
+
+def encode_id(
+    node_type_name: str, query: str, query_params: Optional[Sequence[Any]] = None
+) -> str:
```
sfc-gh-aalam, line 177: can we call this `encode_node_id`?

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
```diff
@@ -144,16 +149,41 @@ def build_plan_to_query_map_in_post_order(root: "TreeNode") -> None:
...
+
+    If query and query parameters can be encoded successfully using sha256,
+    return the encoded query id + node_type_name.
+    Otherwise, generate a uuid.
+    """
+    query_id = encoded_query_id(query, query_params)
+    if query_id is not None:
+        return query_id + node_type_name
+    else:
+        return str(uuid.uuid4())
```
sfc-gh-aalam, line 189: do we want the same node to have the same `encode_id`. If so, we can rely on `id(node)` which gives you deterministic result, instead of `uuid4()`

sfc-gh-yzou:
@sfc-gh-aalam 
"""
Can you give an example of where this does not make sense?
"""
i am not sure for which transformation this could make sense, unless the transformation want to treat two nodes with the same query as the same node, for example, for your large query breakdown, i don't think you want to treat two nodes with the same query as the same node during transformation. Or even cte transformation, during the actual node transformation, we don't want to treat two nodes as the same node during the plan transformation, because the are different nodes.  This could cause us to missing apply transformation on some nodes and cause potential problem, because we could incorrectly thought some nodes have been handled.
In general, we should never change node comparison unless it is by design, however, for node in plan tree, that should not be the case.

This has been causing problem to our ctc workloads, where we missed some handling of some nodes, I have tried it with our benchmark workloads, the problem has been gone with this refactoring.

sfc-gh-aalam (approved):
LGTM. a few comments.

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
```diff
@@ -144,16 +149,41 @@ def build_plan_to_query_map_in_post_order(root: "TreeNode") -> None:
...
+    """
+    Encode given query, query parameters and the node type into an id.
+
+    If query and query parameters can be encoded successfully using sha256,
+    return the encoded query id + node_type_name.
+    Otherwise, generate a uuid.
+    """
+    query_id = encoded_query_id(query, query_params)
+    if query_id is not None:
+        return query_id + node_type_name
```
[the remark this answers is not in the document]
sfc-gh-jdu: +1

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
```diff
@@ -144,16 +149,41 @@ def build_plan_to_query_map_in_post_order(root: "TreeNode") -> None:
...
+
+
+def encode_id(
+    node_type_name: str, query: str, query_params: Optional[Sequence[Any]] = None
+) -> str:
+    """
+    Encode given query, query parameters and the node type into an id.
+
+    If query and query parameters can be encoded successfully using sha256,
+    return the encoded query id + node_type_name.
```
sfc-gh-jdu, line 182: why do we need node_type_name?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -506,13 +500,21 @@ def sql_query(self) -> str:
...
+    def encoded_id(self) -> str:
         """
         Returns the id of this SelectSQL logical plan. The original SQL is used to encode its ID,
         which might be a non-select SQL.
         """
-        return encode_id(self.original_sql, self.query_params)
+        return encode_id(type(self).__name__, self.original_sql, self.query_params)
+
+    @cached_property
+    def encoded_query_id(self) -> Optional[str]:
```
sfc-gh-jdu, line 512: these two properties seem having the same docstring?

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
sfc-gh-helmeleegy: +1

sfc-gh-helmeleegy:
> > This make sense when we do repeated subquery elimination, but not expected by other transformations.
> 
> Can you give an example of where this does not make sense?

Also, it sounds like this a behavioral change. Are there production workloads that can be impacted? Which ones?

sfc-gh-yzou:
@sfc-gh-helmeleegy there should be no user facing behavior change, this equivalence check is introduced by the cte transformation before, and only used internally for node comparison, no others should rely on this fact. This is counted as a refactor work, so no existing workload should be impacted

tests/integ/test_deepcopy.py
```diff
@@ -401,31 +401,15 @@ def test_deepcopy_no_duplicate(session, generator):
...
-    # note that two nodes with same plan._id can have different id(plan) since SnowflakePlan inherits plan._id
-    # from its source selectable.
-    #   plan._id: calculated by hashing the query sql and query params of plan
-    #   type(plan): the type of the plan
-    #   id(plan): the memory address of the plan object
-    # If the same plan._id has multiple id(plan), it means the deepcopy is duplicating source nodes
-
     def traverse_plan(plan, plan_id_map):
-        plan_id = plan._id
-        plan_type = type(plan)
```
sfc-gh-yzou, line 414: @sfc-gh-aalam i am not sure what are we trying to test here, if two nodes have the same query and same type, there should still be two nodes after deepcopy, because they are two different nodes.

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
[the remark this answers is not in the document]
sfc-gh-yzou: updated

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
sfc-gh-yzou: i actually call this encode_node_id_with_query to be more clear of the encoded id

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
sfc-gh-yzou: this is used for cached property, so with uuid4 generation it will also be deterministitic, but we can use id also.

src/snowflake/snowpark/_internal/analyzer/cte_utils.py
sfc-gh-yzou: that is following the previous equivalence check, we require the node type to be the same, for example, a snowflake plan node and selectstatement node with the same query is not counted as the same

sfc-gh-jdu (approved):

sfc-gh-aalam (approved):

PR #2441, opened 2024-10-11
SNOW-1692063 - numpy log support ( using snowpark functions )

sfc-gh-jkew:
Support numpy.log functions, which is mapped to the snowpark log function. Snowpark functions originally did not support mixed positional parameters so this PR implements a way to pass named parameters to the snowpark function which is reconstituted into a set of positional parameters.

For example, the function:
```
snowflake.snowpark.functions.log(base: Union[Column, str, int, float], 
                                   x: Union[Column, str, int, float]) → [Column]
```
Assumes that base is first and the column reference is specified second. To support numpy.log2 we would have to call `log(base=2, x=col)` but the ordering here is not clear, either `base` could be a column or `x` could be a column, or technically both.

To fix this we inspect the function when kwargs is used with a snowpark function. The argument names to the function are then compared to the kwargs and a new set of positional arguments is created for the call.

This does *not* handle cases where a snowpark function might require multiple col references, and all non-col arguments need to be specified explicitly as named arguments.

An alternative approach would be to create some sort of metadata registry for functions so we can more explicitly align positional parameters without reflection.

**Checklist:**


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8572,11 +8572,37 @@ def wrapped_func(*args, **kwargs):  # type: ignore[no-untyped-def] # pragma: no
...
+            for arg in positional_args:
+                if arg in kwargs:
+                    resolved_positional.append(kwargs[arg])
+                else:
+                    if not col_specified:
+                        resolved_positional.append(col)
+                        col_specified = True
+                    else:
+                        ErrorMessage.not_implemented(
+                            "kwargs specified with apply cannot have multiple column references"
```
sfc-gh-azhan, line 8602: can you provide more info like which key has multiple column references?

sfc-gh-jkew: Ok. I re-worded things and added a test to make sure I was catching this case. The error message is a little complicated, but I summed it up to something more like:

"Unspecified Argument: {arg} - when using apply with kwargs, all function arguments should be specified except the single column reference (if applicable)."

It's a bit of a weird scenario, but if we are limiting this to snowpark functions for now I'm not sure if the user would run into this accidentally.

sfc-gh-azhan: I just want to deliver more info from this message back to our telemetry so we can reason and reproduce on our side.

sfc-gh-azhan (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8572,11 +8572,37 @@ def wrapped_func(*args, **kwargs):  # type: ignore[no-untyped-def] # pragma: no
...
+            for arg in positional_args:
+                if arg in kwargs:
+                    resolved_positional.append(kwargs[arg])
+                else:
+                    if not col_specified:
+                        resolved_positional.append(col)
+                        col_specified = True
+                    else:
+                        ErrorMessage.not_implemented(
+                            f"Unspecified Argument: {arg} - when using apply with kwargs, all function arguments should be specified except the single column reference (if applicable)."
```
sfc-gh-joshi, line 8602: will this message will be raised if the function takes multiple columns? should we make clearer that this isn't yet supported?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8572,11 +8572,37 @@ def wrapped_func(*args, **kwargs):  # type: ignore[no-untyped-def] # pragma: no
     def _apply_snowpark_python_function_to_columns(
         self,
         snowpark_function: Callable,
+        kwargs: Any,  # possible named arguments which need to be added
```
sfc-gh-joshi, line 8575: should the typing here be `dict[Any, Any]`? also, should we make it `**kwargs` to match the usual python convention?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: So this should *only* be hit on the snowpark functions we explicitly allow - and... yeah; say you have a function like:

`func(A=<scalar>, B=<col>, C=<col>)` 

and you pass a kwargs is { A: "blabla" } then the code would assume that B references "self" and then it would throw an error when it saw another parameter "C"

To be honest; if we didn't already limit these functions to snowpark ones I think I would try to articulate a different path - but supporting apply with functions which have multiple column references is a little outside the bounds of pandas too.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: I typed this as dict[str, Any] ( which is sort of similar to how we type other internal functions ). I can't do **kwargs here because it's an explicit argument passed through from other functions as opposed to an optional set of arguments.

sfc-gh-joshi (approved):

PR #2442, opened 2024-10-14
SNOW-1043520: Add option and partition_by for dataframe writer

sfc-gh-aalam:

   

   Fixes SNOW-1043520, SNOW-826851

   This PR adds methods `partition_by`, `option` and `options` to `DataFrameWriter` and make it more daisy-chain friendly. This makes `DataFrameWriter` similar to `DataFrameReader`.


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -803,6 +811,22 @@ def get_copy_into_table_options(
...
+    """Method that extracts options for COPY INTO LOCATION command into file
+    format type options and copy into location options.
+    """
+    file_format_type_options = options.get("FORMAT_TYPE_OPTIONS", {})
+    copy_into_location_options = options.get("COPY_INTO_LOCATION_OPTIONS", {})
+    for k, v in options.items():
+        if k in COPY_INTO_LOCATION_OPTIONS:
+            copy_into_location_options[k] = v
+        elif k not in NON_FORMAT_TYPE_OPTIONS:
+            file_format_type_options[k] = v
```
sfc-gh-aling, line 826: just for my own learning purposes as I see three categories here:
1. copy into option
2. file format which is not copy into option
3. non file format which is not copy into option

so the options in `copy_into_location_options` doesn't overlap `file_format_type_options` at all?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -803,6 +811,22 @@ def get_copy_into_table_options(
     return file_format_type_options, copy_options
 
 
+def get_copy_into_location_options(
```
sfc-gh-aling, line 814: nit: the signature name implies it's only returning copy_into_location_options but this function also returns file_format_type_options, or file_format_type_options is part of copy_into_location_options?

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -92,6 +95,26 @@ def mode(self, save_mode: str) -> "DataFrameWriter":
...
+        self._partition_by = expr
+        return self
+
+    def option(self, key: str, value: Any) -> "DataFrameWriter":
+        """Depending on the ``file_format_type`` specified, you can include more format specific options.
+        Use the options documented in the `Format Type Options <https://docs.snowflake.com/en/sql-reference/sql/copy-into-location.html#format-type-options-formattypeoptions>`__.
+        """
+        aliased_key = get_aliased_option_name(key, WRITER_OPTIONS_ALIAS_MAP)
+        self._cur_options[aliased_key] = value
+        return self
```
sfc-gh-aling, line 111: do we need to clean up old ones after execution or before setting new values? otherwise the options keeps accumulated? I'm thinking of the following case:

df.write.options(a=1) # this has option a = 1
df.write.options(b=2) # this would have option a= 1 and b = 2

sfc-gh-aalam: This is intentional to allow daisy chaining. We want to allow df.writer to be something like
```
df.writer.option("opt1", val1).option("opt2", val2)
```

note that df.reader already supports the same pattern. We are extending the same functionality to writer.

src/snowflake/snowpark/_internal/utils.py
sfc-gh-aalam: this is true. I just followed the same pattern for copy into table. I can refactor both functions so they match

src/snowflake/snowpark/_internal/utils.py
sfc-gh-aalam: copy into location: https://docs.snowflake.com/en/sql-reference/sql/copy-into-location is used for dataframe writer and,
copy into table: https://docs.snowflake.com/en/sql-reference/sql/copy-into-location is used for dataframe reader.

For the two, COPY OPTIONS supported are slightly different. I renamed the dictionaries so that this distinction is more obvious. let me know if this is better.

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-aling: I see, if this is an existing pattern than I'm good with it

src/snowflake/snowpark/_internal/utils.py
sfc-gh-aling: I see, thanks! it's clear to me now

sfc-gh-aling (approved):

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -839,10 +847,29 @@ def check_is_pandas_dataframe_in_to_pandas(result: Any) -> None:
...
+    format type options and copy options.
+    """
+    file_format_type_options = options.get("FORMAT_TYPE_OPTIONS", {})
+    copy_options = options.get("COPY_OPTIONS", {})
+    for k, v in options.items():
+        if k in COPY_INTO_TABLE_COPY_OPTIONS:
+            copy_options[k] = v
+        elif k not in NON_FORMAT_TYPE_OPTIONS:
+            file_format_type_options[k] = v
+    return file_format_type_options, copy_options
```
sfc-gh-jrose, line 860: These two function are almost entirely the same. It might be cleaner to make something like this in order to share logic:
```python
def _get_options(
    options: Dict[str, Any],
    allowed_options: Dict[str, Any],
) -> Tuple[Dict[str, Any], Dict[str, Any]]:
    """Method that extracts options for COPY INTO LOCATION command into file
    format type options and copy options.
    """
    file_format_type_options = options.get("FORMAT_TYPE_OPTIONS", {})
    copy_options = options.get("COPY_OPTIONS", {})
    for k, v in options.items():
        if k in COPY_OPTIONS:
        if k in allowed_options:
            copy_options[k] = v
        elif k not in NON_FORMAT_TYPE_OPTIONS:
            file_format_type_options[k] = v
    return file_format_type_options, copy_options

get_copy_into_table_options = partial(_get_options, allowed_options=COPY_INTO_TABLE_COPY_OPTIONS)
get_copy_into_location_options = partial(_get_options, allowed_options= COPY_INTO_LOCATION_COPY_OPTIONS)
```

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/utils.py
sfc-gh-aalam: that's a great suggestion. Updated

PR #2443, opened 2024-10-14
SNOW-1677859 Add table functions passthough dataframeAst

sfc-gh-evandenberg:
Fixes SNOW-1677859 Add table functions passthough dataframeAst


src/snowflake/snowpark/table.py
```diff
@@ -369,6 +374,21 @@ def sample(
 
         from snowflake.snowpark.mock._connection import MockServerConnection
 
+        stmt = None
+        if _emit_ast:
+            # AST.
+            stmt = self._session._ast_batch.assign()
+            ast = with_src_position(stmt.expr.sp_table_sample, stmt)
+            if frac:
+                ast.probability_fraction.value = frac
```
sfc-gh-oplaton, line 383: This looks surprising to me. In general, I think we used `build_expr_from_python_val` for this type of translation. We need to apply the same pattern to all values.

src/snowflake/snowpark/table.py
```diff
@@ -485,6 +506,34 @@ def update(
...
+                    build_expr_from_snowpark_column_or_python_val(t._2, v)
+            if condition is not None:
+                build_expr_from_snowpark_column(ast.condition, condition)
+            if source is not None:
+                source._set_ast_ref(ast.source)
+            if statement_params is not None:
+                for k, v in statement_params.items():
+                    t = ast.statement_params.add()
+                    t._1 = k
+                    t._2 = v
```
sfc-gh-oplaton, line 529: These few lines repeat several times in this module, and I bet the pattern shows up in a few other modules, too. Time for a `build_expr_from_dict`?

src/snowflake/snowpark/table.py
sfc-gh-evandenberg: This is the same code as the dataframe.sample, I can update that location as well.

Looking at the helper code, it seems similar but also sets some src_position information with the type.

src/snowflake/snowpark/mock/_connection.py
```diff
@@ -700,15 +703,18 @@ def execute(
             rows = [rows] if to_iter else rows
 
         # Notify query listeners.
+        self.notify_mock_query_record_listener(kwargs)
+
+        return iter(rows) if to_iter else rows
+
```
sfc-gh-lspiegelberg, line 709: Why do we need a new function `notify_mock_query_record_listener(kwargs)` here? Shouldn't it be sufficient to directly call ` self.notify_query_listeners` or so?

src/snowflake/snowpark/table.py
sfc-gh-evandenberg: I tried this but it fails:

        elif isinstance(obj, float):
>           ast = with_src_position(expr_builder.float64_val)
E           AttributeError: float64_val

It looks like this is only used if it's a variant type.

src/snowflake/snowpark/mock/_connection.py
sfc-gh-evandenberg: To avoid duplicating the code relate to the mock plan query record, request-id and passing through the dataframeAst.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -1310,3 +1310,18 @@ def build_sproc(
         t = ast.kwargs.add()
         t._1 = k
         build_expr_from_python_val(t._2, v)
+
+
+def build_expr_from_dict_str_str(
+    ast_dict: proto.Tuple_String_String, dict_str_str: Dict[str, str]
+) -> None:
+    """Populate the AST strutcure with dictionary for str -> str.
```
sfc-gh-oplaton, line 1318: `s/strutcure/structure/`

sfc-gh-evandenberg: Fixed

sfc-gh-oplaton (approved):

PR #2446, opened 2024-10-14
[SNOW-1731549] Disable sql simplification when sort is performed after limit

sfc-gh-yzou:

SNOW-1731549

Today, sql simplification flattens df.limit(2).orderby(col1) into one SelectStatement, which always put order by clause in front of the limit clause during code generation. 
However, limit order by and order by limit can cause big performance difference, because limit can stop table scanning whenever the number of record is satisfied.

Therefore, this pr disallow sql simplification when the current SelectStatement has a limit clause to avoid moving order by in front of limit to respect what the user have coded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@
 
 #### Improvements
 - Disables sql simplification when sort is performed after limit. 
-  - Improve performance of generated query since limit stops table scanning as soon as number of record is satisfied.
-  - Note before sort after limit and limit after sort will always return top k result, but now sort after limit will no longer return deterministic result.
+  - Respect the original clause order and improve performance of generated query since limit stops table scanning as soon as number of record is satisfied.
+  - Note that before `df.sort().limit()` and `df.limit().sort()` will generate same query with effect of `df.sort().limit()`. This will be no longer the case after the change, `df.limit().sort()` will keep its original clause order.
 
 ### Snowpark pandas API Updates
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
 
 #### New Features
 
+#### Improvements
+- Disables sql simplification when sort is performed after limit. 
+  - Improve performance of generated query since limit stops table scanning as soon as number of record is satisfied.
+  - Note before sort after limit and limit after sort will always return top k result, but now sort after limit will no longer return deterministic result.
+
 ### Snowpark pandas API Updates
 
 #### New Features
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 #### New Features
 
+#### Improvements
+- Disables sql simplification when sort is performed after limit. 
+  - Improve performance of generated query since limit stops table scanning as soon as number of record is satisfied.
+  - Note before sort after limit and limit after sort will always return top k result, but now sort after limit will no longer return deterministic result.
```
sfc-gh-jdu, line 12: This sentence seems not easy to read, maybe use an example?

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 #### New Features
 
+#### Improvements
+- Disables sql simplification when sort is performed after limit.
```
sfc-gh-jdu, line 10: Are following details needed to be known by users publicly? Or Snowflake compiler details? If not, maybe we don't need to mention them?

CHANGELOG.md
sfc-gh-yzou: added an example with the snowpark code

CHANGELOG.md
sfc-gh-yzou: For this one, yes. This is not snowflake compiler details, more of snowpark generated query behavior. Event hough we haven't provide any guarantee, we still want to notify the user about this change, in case this is causing any problem to there workloads

sfc-gh-jdu (approved):

sfc-gh-yuwang (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
 
 #### New Features
 
+#### Improvements
+- Disables sql simplification when sort is performed after limit. 
+  - Respect the original clause order, before `df.sort().limit()` and `df.limit().sort()` are the same as `df.sort().limit()`.
```
sfc-gh-aalam, line 13: I understand what you mean here is that before this change, we should treat `df.sort().limit()` the same as `df.limit().sort()` but with this change, we fix it. @sfc-gh-pspaeth can you suggest how we can update this changelog.

sfc-gh-aalam (approved):

CHANGELOG.md
sfc-gh-yzou: updated based on the feedback

sfc-gh-azhan (approved):

PR #2448, opened 2024-10-14
SNOW-1709861: table clean up uses its own cursor

sfc-gh-aling:

   

   Fixes SNOW-NNNNNNN

auto table clean up could be kicked off by python gc at random timing, this caused flakiness in our async query execution as the random execution of async table clean up might overwrite the status of the shared session object leading to incorrect async query id being stored.

to improve, here we create a low level new connector cursor to execute drop temp table each time to avoid side effect


src/snowflake/snowpark/session.py
```diff
@@ -585,11 +585,12 @@ def __init__(
...
-            self._conn._get_client_side_session_parameter(
-                _PYTHON_SNOWPARK_AUTO_CLEAN_UP_TEMP_TABLE_ENABLED, False
-            )
-        )
+        self._auto_clean_up_temp_table_enabled = True
+        # self._auto_clean_up_temp_table_enabled: bool = (
+        #     self._conn._get_client_side_session_parameter(
+        #         _PYTHON_SNOWPARK_AUTO_CLEAN_UP_TEMP_TABLE_ENABLED, False
+        #     )
+        # )
```
sfc-gh-aling, line 593: this is for testing purpose only, will revert before merging the PR

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
```diff
@@ -68,13 +69,28 @@ def drop_table(self, name: str) -> None:  # pragma: no cover
...
         query_id = None
         try:
-            async_job = self.session.sql(
-                f"drop table if exists {name} /* internal query to drop unused temp table */",
-            )._internal_collect_with_tag_no_telemetry(
-                block=False, statement_params={DROP_TABLE_STATEMENT_PARAM_NAME: name}
-            )
-            query_id = async_job.query_id
-            logging.debug(f"Dropping {common_log_text} with query id {query_id}")
+            if (
```
sfc-gh-jdu, line 72: why do we need to skip it in stored proc?

tests/integ/test_deepcopy.py
```diff
@@ -387,9 +387,6 @@ def test_deep_nested_select(session):
     [
         lambda session_: session_.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"]),
         lambda session_: session_.sql("select 1 as a, 2 as b"),
-        lambda session_: session_.table(
```
sfc-gh-jdu, line 390: we don't need this change?

sfc-gh-aling:
successful daily precommit running without any flaky test: https://github.com/snowflakedb/snowpark-python/actions/runs/11372006647

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-aling: this is equivalent to our server_connection.execute method:

https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/server_connection.py#L562-L581

tests/integ/test_deepcopy.py
sfc-gh-aling: this test cases use cache_result, the table will be auto cleaned in the middle of execution as it's not referenced.
@sfc-gh-aalam is onboard with the change

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-jdu: can you add this check here instead? https://github.com/snowflakedb/snowpark-python/blob/952599b1395dad9047dfc75f51324eb6014ada2a/src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py#L54, which seems cleaner

tests/integ/test_deepcopy.py
sfc-gh-jdu: hmm I don't get it, from code, it seems based_df shouldn't be gced, then if it's cleaned in the middle of execution, isn't unexpected?

sfc-gh-aling: the clean up code is working expected.

if you check the table set up process, we don't keep reference of the originally created cache_result() table, instead we just use the table_name which is a string var.
```python
session_.table(
     session_.sql("select 1 as a, 2 as b").cache_result().table_name
)
```

in this way, no one is referencing the cache_result() table, so it gets cleaned up.

if we want to table not cleaned up, the test needs to be rewritten into the following format
```
cache_df = session_.sql("select 1 as a, 2 as b").cache_result()
session_.table(cache_df.table_name)
```

src/snowflake/snowpark/_internal/temp_table_auto_cleaner.py
sfc-gh-aling: sure, I can make the change

sfc-gh-jdu (approved):

tests/integ/test_deepcopy.py
sfc-gh-aalam: yeah, we can remove this test.

src/snowflake/snowpark/session.py
sfc-gh-aalam: reminder to add it back.

sfc-gh-aalam (approved):

PR #2450, opened 2024-10-14
SNOW-1730546: Bug fix for to_csv on stage with empty strings

sfc-gh-nkumar:
Fixes: SNOW-1730546

RCA: 
We have two different code paths depending on destination location.
When calling to_csv on the local path we use native pandas to_csv. This works because native pandas doesn’t differentiate between null and empty strings when writing to csv. Empty strings are treated as null values. As a side effect, calling to_csv and read_csv again will result in replacing all empty strings with NULL/NA.

When calling to_csv on the snowflake stage path we use snowflake’s copy command. Snowflake copy command is not able to handle empty strings out of box and requires users to pass FIELD_OPTIONALLY_ENCLOSED_BY (which is NONE by default).

Solution:
Always set EMPTY_FIELD_AS_NULL = False when to_csv on stage is called.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 
 - Added numpy compatibility support for `np.float_power`, `np.mod`, `np.remainder`, `np.greater`, `np.greater_equal`, `np.less`, `np.less_equal`, `np.not_equal`, and `np.equal`.
 
+#### Bug Fixes
+- Fixed a bug where `to_csv` on Snowflake stage fails when data contains empty strings.
+
 ## 1.23.0 (2024-10-09)
 
 ### Snowpark Python API Updates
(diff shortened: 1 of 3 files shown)


sfc-gh-jjiao:
Thanks @sfc-gh-nkumar for the proposed fix.

Have you seen this doc  https://docs.snowflake.com/en/user-guide/data-unload-considerations ? It talks about other options EMPTY_FIELD_AS_NULL and NULL_IF, could we see whether those are useful for this issue too?  

I feel if we set EMPTY_FIELD_AS_NULL to TRUE, it will matches pandas behavior as well?  

But I am not quite understanding a statement in that doc " If the EMPTY_FIELD_AS_NULL option is TRUE (which is prohibited)",  it said setting EMPTY_FIELD_AS_NULL to TRUE is prohibited?

sfc-gh-nkumar:
> Thanks @sfc-gh-nkumar for the proposed fix.
> 
> Have you seen this doc https://docs.snowflake.com/en/user-guide/data-unload-considerations ? It talks about other options EMPTY_FIELD_AS_NULL and NULL_IF, could we see whether those are useful for this issue too?
> 
> I feel if we set EMPTY_FIELD_AS_NULL to TRUE, it will matches pandas behavior as well?
> 
> But I am not quite understanding a statement in that doc " If the EMPTY_FIELD_AS_NULL option is TRUE (which is prohibited)", it said setting EMPTY_FIELD_AS_NULL to TRUE is prohibited?

Good callout. I have included this option in doc for further discussion.

sfc-gh-nkrishna (approved):
LGTM!

sfc-gh-joshi (approved):

PR #2454, opened 2024-10-15
SNOW-1690923 Encode Intervals

sfc-gh-vbudati:

   Fixes SNOW-1690923

Encoding the Interval object in AST. Relevant test is `interval.test`.


src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -379,7 +380,10 @@ def __init__(
         millisecond: Optional[int] = None,
         microsecond: Optional[int] = None,
         nanosecond: Optional[int] = None,
+        _emit_ast: bool = True,
     ) -> None:
+        from snowflake.snowpark._internal.ast_utils import with_src_position
```
sfc-gh-vbudati, line 385: Had to import here to prevent circular imports issue.

sfc-gh-vbudati:
I ran the linter which is why there's 100+ file changes - most of these changes are in the test data directory and I think are superficial.  I'm not sure how to fix the pyright/precommit type checking test.

src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -405,6 +409,14 @@ def __init__(
         if nanosecond is not None:
             self.values_dict["NANOSECOND"] = nanosecond
 
+        if self._ast is None and _emit_ast:
+            expr = proto.Expr()
```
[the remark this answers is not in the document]
sfc-gh-evandenberg: Can you repro locally?

python -m tox -e pyright

It is saying you may need a new version: 
"Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`"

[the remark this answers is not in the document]
sfc-gh-vbudati: I get the same errors locally, I'm not sure what to do about the first error though. There's other parts of the code that use `proto.Expr()` as well so I'm not sure why it's erroring here

src/snowflake/snowpark/_internal/analyzer/expression.py
sfc-gh-lspiegelberg: So far in the code, we tried to avoid building expressions directly. The reasoning was that Expressions should disappear as they're not a first-class entity in snowpark. Instead we chose an approach where always the caller (i.e. the function that creates Interval) creates the AST and assigns it to the expression if need be. I.e.,

```
def build_expr_from_snowpark_column_or_python_val(
    expr_builder: proto.Expr, value: ColumnOrLiteral
) -> None:
    """Copy from a Column object's AST, or copy a literal value into an AST expression.

    Args:
        ast (proto.Expr): A previously created Expr() IR entity intance to be filled
        value (ColumnOrLiteral): The value from which to populate the provided ast parameter.

    Raises:
        TypeError: The Expr provided should only be populated from a Snowpark Column with a valid _ast field or a literal value
    """
    if isinstance(value, snowflake.snowpark.Column):
        build_expr_from_snowpark_column(expr_builder, value)
    elif isinstance(value, VALID_PYTHON_TYPES_FOR_LITERAL_VALUE):
        build_expr_from_python_val(expr_builder, value)
    elif isinstance(value, Expression):
        # Expressions must be handled by caller.
        pass
    else:
        raise TypeError(f"{type(value)} is not a valid type for Column or literal AST.")
```

src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -405,6 +409,14 @@ def __init__(
...
+            ast = with_src_position(expr.sp_interval)
+            # Set the AST values based on the values_dict.
+            for k, v in self.values_dict.items():
+                getattr(ast, k.lower()).value = v
```
sfc-gh-lspiegelberg, line 417: let's try to avoid indirection via getattr and use the values explicitly. It's much easier to read and reason about :)

src/snowflake/snowpark/_internal/analyzer/expression.py
sfc-gh-lspiegelberg: we prob. can move the AST logic to build_expr_from_snowpark_column_or_python_val and have an `Interval` case there. I don't think the original intention of snowpark was to use `Interval` directly, but rather the `make_interval` function.

tests/ast/data/interval.test
```diff
@@ -0,0 +1,81 @@
...
+            [datetime.datetime(2010, 1, 1), datetime.datetime(2011, 1, 1)],
+            [datetime.datetime(2012, 1, 1), datetime.datetime(2013, 1, 1)],
+        ],
+        schema=["a", "b"],
+    )
+
+df2 = df1.select(
+    df1["a"]
+    + Column(
+        Interval(
```
sfc-gh-lspiegelberg, line 16: let's use make_interval here to test. Not sure we should support the `Column(Interval(...))` syntax at all.

tests/ast/data/interval.test
```diff
@@ -0,0 +1,81 @@
...
+        )
+    )
+)
+
+
+## EXPECTED UNPARSER OUTPUT
+
+df1 = session.create_dataframe([[datetime.datetime(2010, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=-18000), name="EST")), datetime.datetime(2011, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=-18000), name="EST"))], [datetime.datetime(2012, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=-18000), name="EST")), datetime.datetime(2013, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=-18000), name="EST"))]], schema=["a", "b"])
+
+df2 = df1.select(df1["a"] + Interval(quarter=1, month=1, week=2, day=2, hour=2, minute=3, second=3, millisecond=3, microsecond=4, nanosecond=4))
```
sfc-gh-lspiegelberg, line 71: does this work in snowpark?

sfc-gh-vbudati: I believe it does - I pulled this from `test_interval`:
```py
def test_interval(session):
    df1 = session.create_dataframe(
        [
            [datetime.datetime(2010, 1, 1), datetime.datetime(2011, 1, 1)],
            [datetime.datetime(2012, 1, 1), datetime.datetime(2013, 1, 1)],
        ],
        schema=["a", "b"],
    )
    df2 = df1.select(
        df1["a"]
        + make_interval(
            quarters=1,
            months=1,
            weeks=2,
            days=2,
            hours=2,
            minutes=3,
            seconds=3,
            milliseconds=3,
            microseconds=4,
            nanoseconds=4,
        )
    )
    verify_column_result(
        session,
        df2,
        [
            '"(""A"" + INTERVAL \'1 QUARTER,1 MONTH,2 WEEK,2 DAY,2 HOUR,3 MINUTE,3 SECOND,3 MILLISECOND,4 MICROSECOND,4 NANOSECOND\')"',
        ],
        [TimestampType(timezone=TimestampTimeZone.NTZ)],
        None,
    )
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -1358,7 +1358,7 @@ def select(
         for e in exprs:
             if isinstance(e, Column):
                 names.append(e._named())
-                if _emit_ast and ast:
+                if _emit_ast and ast and e._ast:
```
sfc-gh-lspiegelberg, line 1361: Curious, why do you also need to check `e._ast` here?

tests/ast/data/interval.test
```diff
@@ -0,0 +1,772 @@
...
+        days=4,
+        hours=5,
+        minutes=6,
+        seconds=7,
+        milliseconds=8,
+        microseconds=9,
+        nanoseconds=10
+    )
+)
+
```
sfc-gh-lspiegelberg, line 59: can you add some example where it's with
```
 make_interval(
        years=1)
``` or so only? I.e., only a few arguments set? Would be great to see that the unparser doesn't produce a lot of `None` then.

File: tests/ast/data/interval.test [hunk 8]
```diff
@@ -0,0 +1,772 @@
...
+
+## EXPECTED UNPARSER OUTPUT
+
+df1 = session.create_dataframe([[datetime.datetime(2010, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=-18000), name="EST")), datetime.datetime(2011, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=-18000), name="EST"))], [datetime.datetime(2012, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=-18000), name="EST")), datetime.datetime(2013, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=-18000), name="EST"))]], schema=["a", "b"])
+
+df2 = df1.select(df1["a"] + make_interval(None, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, None, None))
```
sfc-gh-lspiegelberg, line 65: maybe let's get rid off the None here in the unparser and use keyword arguments? Should make a much more succinct syntax. Also easier to read.

src/snowflake/snowpark/dataframe.py
sfc-gh-vbudati: I added the check here since `make_interval` creates a Column object but it has no associated `_ast` field. Since `_emit_ast` is True by default, select tries to perform `ast.cols.append(e._ast)` and fails since `e._ast` is None (here `e` is the Column created by `make_interval`).

I don't think it is correct to do `_emit_ast=False` on the "user" side of things since it is supposed to be an interval parameter.

tests/ast/data/interval.test
sfc-gh-vbudati: sounds good! I turned the parameters into keyword arguments and only display them if they are non-None values.

tests/ast/data/interval.test
sfc-gh-vbudati: done!

src/snowflake/snowpark/dataframe.py
sfc-gh-lspiegelberg: oh, I think in that case you can remove the check for `e._ast` then. Simply assign in `make_interval` the `_ast` field. Trying to avoid here one-off solutions for an API. I think it should work without this (added) check. Can we try?

src/snowflake/snowpark/functions.py
```diff
@@ -9830,16 +9830,46 @@ def make_interval(
...
     """
-
+    ast = None
     if _emit_ast:
-        raise NotImplementedError("TODO SNOW-1690923: Add interval support to IR.")
+        ast = proto.Expr()
+        # Encode the parameters as kwargs to make them more readable.
+        # If any of the parameters are None, ignore them.
+        kwargs = {}
+        if years:
```
sfc-gh-lspiegelberg, line 9839: use `is not None`. Technically, `years=0` will also not be encoded.

src/snowflake/snowpark/functions.py
```diff
@@ -9852,5 +9882,9 @@ def make_interval(
             milliseconds,
             microseconds,
             nanoseconds,
-        )
+        ),
+        _emit_ast=False,
     )
+
+    res._ast = ast
```
sfc-gh-lspiegelberg, line 9889: this here should allow you to remove the check in `select`.

sfc-gh-lspiegelberg (approved):
please remove the added check in `select`, consider using `is not None`, then good to go :)

src/snowflake/snowpark/functions.py
sfc-gh-vbudati: ooh yeah, I found an issue with the test. Thanks for pointing this out!!

src/snowflake/snowpark/functions.py
sfc-gh-vbudati: fixed!

src/snowflake/snowpark/dataframe.py
sfc-gh-vbudati: removed the check, there was a bug in my test that made me run into this. Thanks for pointing it out!

PR #2462, opened 2024-10-15
SNOW-1728988: Cache attributes on SelectStatement to reduce describe query 

sfc-gh-jdu:

   

   Fixes SNOW-1728988

   When sql simplifier is on, it can resolve the scenarios where 1) select() is called because df._plan is not carried over, 2) sql simplifier can't flatten the query

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -1183,7 +1183,8 @@ class SelectStatement(Selectable):
             new = SelectStatement(
                 from_=self.to_subqueryable(), where=col, analyzer=self.analyzer
             )
-        new._attributes = self._attributes
+        if self.analyzer.session.reduce_describe_query_enabled:
+            new._attributes = self._attributes
 
         return new
 
@@ -1209,7 +1210,8 @@ class SelectStatement(Selectable):
                 order_by=cols,
                 analyzer=self.analyzer,
             )
-        new._attributes = self._attributes
+        if self.analyzer.session.reduce_describe_query_enabled:
+            new._attributes = self._attributes
 
         return new
 
@@ -1292,7 +1294,8 @@ class SelectStatement(Selectable):
             new.pre_actions = new.from_.pre_actions
             new.post_actions = new.from_.post_actions
             new._merge_projection_complexity_with_subquery = False
-        new._attributes = self._attributes
+        if self.analyzer.session.reduce_describe_query_enabled:
+            new._attributes = self._attributes
 
         return new
 
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -683,6 +683,8 @@ class SelectStatement(Selectable):
         self._projection_complexities: Optional[
             List[Dict[PlanNodeCategory, int]]
         ] = None
+        # Metadata/Attributes for the plan
+        self._attributes: Optional[List[Attribute]] = None
 
     def __copy__(self):
         new = SelectStatement(
@@ -1181,6 +1183,7 @@ class SelectStatement(Selectable):
             new = SelectStatement(
                 from_=self.to_subqueryable(), where=col, analyzer=self.analyzer
             )
+        new._attributes = self._attributes
 
         return new
 
@@ -1206,6 +1209,8 @@ class SelectStatement(Selectable):
                 order_by=cols,
                 analyzer=self.analyzer,
             )
+        new._attributes = self._attributes
+
         return new
 
     def set_operator(
@@ -1287,6 +1292,8 @@ class SelectStatement(Selectable):
             new.pre_actions = new.from_.pre_actions
             new.post_actions = new.from_.post_actions
             new._merge_projection_complexity_with_subquery = False
+        new._attributes = self._attributes
+
         return new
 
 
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1181,6 +1183,7 @@ def filter(self, col: Expression) -> "SelectStatement":
             new = SelectStatement(
                 from_=self.to_subqueryable(), where=col, analyzer=self.analyzer
             )
+        new._attributes = self._attributes
```
sfc-gh-helmeleegy, line 1186: Wouldn't this also apply to set operations like union and intersection?

sfc-gh-helmeleegy: Can we first check if the input data types are the same, then it can be propagated to the output. I believe this would be the more common case, and we should be able to recognize it.

sfc-gh-jdu: Actually it's [data coerction](https://docs.snowflake.com/en/sql-reference/data-type-conversion#implicit-casting-coercion) in Snowflake, and there are many [complicated rules](https://docs.snowflake.com/en/sql-reference/data-type-conversion#data-types-that-can-be-cast) on the server side. Even if they are the same type, there may still be data coercion happening (such as two varchar values with different length). So I don't think it's a good idea to have such rules on the client side, and we should only rely on the server side for data coercion. This type of operation is not in the scope of reducing describe queries.

sfc-gh-helmeleegy: Hmm, I think this will reduce describe queries. We're just worried that the rules we use on the client side may end up being different than the server side, right? Would we still be worried if the data types for both operands are exactly the same (including lengths for archer, etc)? Or is it that we don't have this information on the client side so we cannot be 100% sure? I'm just trying to understand the concern.

sfc-gh-jdu: My original concern is just about data coercion, but as you said, if two types are exactly the same, we can try to infer too and we can investigate it in the future, if this is a common pattern that we should optimize.

sfc-gh-helmeleegy: Yes, I agree it doesn't have to be in this PR. For Snowpark pandas, we do use union operations pretty commonly. And I think that in most cases, the data types are identical.

sfc-gh-jdu: got it, yeah if it's pretty common, we can definitely do it.

sfc-gh-helmeleegy (approved):
LGTM. Good progress with reducing describe queries. Thanks, Jianzhun.

sfc-gh-aalam (approved):
would you have to update some tests as a consequence because this change will reduce num describe queries?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yzou, line 1186: flag control

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -400,12 +400,20 @@ def with_subqueries(self, subquery_plans: List["SnowflakePlan"]) -> "SnowflakePl
...
+
         if self._attributes is not None:
             return self._attributes
         assert (
             self.schema_query is not None
         ), "No schema query is available for the SnowflakePlan"
         self._attributes = analyze_attributes(self.schema_query, self.session)
+        # We need to cache attributes on SelectStatement too because df._plan is not
+        # carried over to next SelectStatement (e.g., check the implementation of df.filter()).
+        if isinstance(self.source_plan, SelectStatement):
```
sfc-gh-yzou, line 415: flag_control here

tests/integ/test_reduce_describe_query.py
```diff
@@ -31,20 +31,37 @@ def setup(session):
...
+    lambda session: session.sql("SELECT 1 AS a, 2 AS b").select("b"),
+    lambda session: session.sql("SELECT 1 AS a, 2 AS b").select(
+        "a", lit("2").alias("c")
+    ),
 ]
 
 # Create from Values
-create_from_values_funcs = []
+create_from_values_funcs = [
+    lambda session: session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"]),
```
sfc-gh-yzou, line 45: can we setup this test suite with the control on and off to make sure things all works as expected when the flag is on or off

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-jdu: maybe we don't need this one because if the parameter is off, _attributes is always None. But we can also add

tests/integ/test_reduce_describe_query.py
sfc-gh-jdu: sure

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-yzou: you can have parameter turned off during middle. Let's protect the code for extra safety, and when we do copy, don't we also need to copy the attribute over?

sfc-gh-jdu: sure, let me add the parameter protection. I'd prefer not to copy the attributes for now, because `select()` use copy to create a new select_statement, where attributes may not be the same https://github.com/snowflakedb/snowpark-python/blob/016b063f4f95c0604d9572be5e2f7a68b437a4c7/src/snowflake/snowpark/_internal/analyzer/select_statement.py#L1114. Rather copying attributes then resetting attributes outside of copy, not copying it for now is safer

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: added

sfc-gh-yzou (approved):

PR #2466, opened 2024-10-17
SNOW-1747461: Support `on` parameter with `Resampler`

sfc-gh-nkrishna:
SNOW-1747461

This PR adds support for the `on` parameter with `Resampler` and does some small cleanup removing repeated calls to utility functions for Resample methods.


tests/integ/modin/resample/test_resample_on.py
```diff
@@ -0,0 +1,75 @@
...
+    native_df = native_pd.DataFrame(
+        data={
+            "A": np.random.randn(15),
+            "B": native_pd.date_range("2020-01-01", periods=15, freq="1s"),
+        },
+        index=native_pd.date_range("2020-10-01", periods=15, freq="1s"),
+    )
+    snow_df = pd.DataFrame(native_df)
+    with pytest.raises(TypeError):
+        native_df.resample(rule="2s", on="A").min()
```
sfc-gh-nkumar, line 55: native pandas has different message here?

tests/integ/modin/resample/test_resample_on.py
```diff
@@ -0,0 +1,75 @@
...
+        data={
+            "A": np.random.randn(15),
+            "B": native_pd.date_range("2020-01-01", periods=15, freq="1s"),
+        },
+        index=native_pd.date_range("2020-10-01", periods=15, freq="1s"),
+    )
+    snow_df = pd.DataFrame(native_df)
+    with pytest.raises(KeyError):
+        native_df.resample(rule="2s", on="invalid").min()
+    with pytest.raises(IndexError):
```
sfc-gh-nkumar, line 74: Just curios, why we are raising different error type here?

src/snowflake/snowpark/modin/plugin/_internal/resample_utils.py
```diff
@@ -359,21 +359,25 @@ def perform_resample_binning_on_frame(
         The internal frame with a single DatetimeIndex column
         to perform resample binning on.
 
+    datetime_index_col : str
+        The datetime-like column to use for resampling.
```
sfc-gh-nkumar, line 363: nit: can you please also mention if this is identifier or label for improved readability.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12322,7 +12322,12 @@ def resample(
 
         validate_resample_supported_by_snowflake(resample_kwargs)
 
-        frame = self._modin_frame
+        axis = resample_kwargs.get("axis", 0)
+        rule = resample_kwargs.get("rule")
+        on = resample_kwargs.get("on")
+
+        # Supplying 'on' to Resampler replaces the existing index of the DataFrame with the 'on' column
+        frame = self.set_index(on)._modin_frame if on else self._modin_frame
```
sfc-gh-nkumar, line 12330: This weird edge case might happen here if on label has "Falsy" value, for example `0`.
Can you please replace this with 
`frame = self.set_index(on)._modin_frame if on is not None else self._modin_frame`

Maybe also add this test case for completeness.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -12420,41 +12429,42 @@ def resample(
                 # Call groupby_size directly on the dataframe or series with the index reset
                 # to ensure we perform count aggregation on row positions which cannot be null
                 qc = (
-                    SnowflakeQueryCompiler(frame)
+                    SnowflakeQueryCompiler(resampled_frame)
                     .reset_index()
                     .groupby_size(
-                        by="index",
-                        axis=resample_kwargs.get("axis", 0),
+                        by=on if on else "index",
```
sfc-gh-nkumar, line 12435: same as above, please check for check if `on is not None`

tests/integ/modin/resample/test_resample_on.py
sfc-gh-nkrishna: Yes different error messages since native pandas supports `PeriodIndex` where we don't, so instead Snowpark pandas error is just "Only valid with DatetimeIndex or TimedeltaIndex"

tests/integ/modin/resample/test_resample_on.py
sfc-gh-nkrishna: IndexError is what we raise when we do `qc.set_index(on)` if `on` is not a valid label. Do you think this is okay or should I add an additional check and return a KeyError like pandas?

src/snowflake/snowpark/modin/plugin/_internal/resample_utils.py
sfc-gh-nkrishna: Good point, done

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkrishna: `on` is guaranteed to be a string so I don't think this case gets hit but added it to be safe.

sfc-gh-helmeleegy (approved):
LGTM. Thanks for the quick response!

sfc-gh-nkumar (approved):

PR #2468, opened 2024-10-17
Snowpark IR: Create the test tables lazily

sfc-gh-oplaton:

   

   Fixes SNOW-1740795

The existing functionality in `test_ast_driver.py` creates all test tables, unconditionally, for each test. This is not a problem in local testing mode, but the table creation time dominates the run time when running the same tests against `sfctest0`. Most tables are unused by a majority of the tests, so they can be created lazily, only by the tests that need them.

Move table creation out of `run_test`.
Introduce the `Tables` helper class, which holds all the cached properties.
Initialize each table lazily and return its name.

It would have been ideal to create both the `session` fixture and the `table` fixture at a Pytest session or module scope. Unfortunately, local testing mocks are incompatible with this approach, so the fixtures are still function-scoped.

The change does not meaningfully change the performance of the local testing mode. Performance doesn't regress, but doesn't improve noticeably, either. The improvement is approximately 1 second out of a baseline of 1 minute 25 seconds.

Running the same tests against `sfctest0` improves by about 61%, from approximately 13 minutes 24 seconds to 5 minutes 11 seconds.


tests/ast/conftest.py
```diff
@@ -42,6 +43,108 @@ def pytest_configure(config):
     pytest.update_expectations = config.getoption("--update-expectations")
 
 
+class Tables:
```
sfc-gh-evandenberg, line 46:
```suggestion
class TestTables:
```

tests/ast/conftest.py
```diff
@@ -42,6 +43,108 @@ def pytest_configure(config):
...
+            [
+                [1, "one"],
+                [2, "two"],
+                [3, "three"],
+            ],
+            schema=["num", 'Owner\'s""opinion.s'],
+        )
+
+    def _save_table(self, name: str, *args, **kwargs):
+        kwargs.pop("_emit_ast", None)
```
sfc-gh-evandenberg, line 136: Don't see the property methods above adding _emit_ast, _ast_stmt, _ast, so what does this popping achieve?

tests/ast/conftest.py
sfc-gh-oplaton: I agree that `TestTables`, or a similar name that describes their nature would be better. I'm not deeply familiar with Pytest - do you happen to know whether it will interpret the class name in any way? Normally, it picks up all `test*` and `Test*` entities and tries to use them as test classes/functions. This is why I avoided both prefixes in all names here.

tests/ast/conftest.py
sfc-gh-oplaton: Please reopen the thread if necessary. Marking as resolved.

sfc-gh-evandenberg (approved):
LGTM

PR #2469, opened 2024-10-17
SNOW-1231581: Enable local testing mode when running doctests.

sfc-gh-jrose:

   

   Fixes SNOW-1231581

   This PR adds the `--local_testing_mode` flag to doctests. This flag is already available for the regular tests, but did not previously work when running doctests in snowflake.snowpark.functions for example.

  It also automatically marks any failures as expected if they are caused by local testing not having implemented a function yet.


tests/conftest.py
```diff
@@ -27,9 +27,12 @@ def is_excluded_frontend_file(path):
 
 def pytest_addoption(parser):
     parser.addoption("--disable_sql_simplifier", action="store_true", default=False)
-    parser.addoption("--local_testing_mode", action="store_true", default=False)
     parser.addoption("--enable_cte_optimization", action="store_true", default=False)
     parser.addoption("--skip_sql_count_check", action="store_true", default=False)
+    if not any(
+        "--local_testing_mode" in opt.names() for opt in parser._anonymous.options
+    ):
```
sfc-gh-aling, line 34: I'm curious what is this change for/what are `parser._anonymous.options`?

sfc-gh-jrose: `parser._anonymous.options` is where all of the options that have been added by the `addoptions` call are stored. This branch is in both `src/conftest.py` and `tests/conftest.py`. In some cases tox uses one conftest or the the other and in some it uses both so the option needs to be added at least once, but cannot be added more times without an error.

If both test locations shared a base path that we could add another the option in a shared conftest, but I don't think we want to add that to the basepath of the project.

sfc-gh-aalam (approved):

sfc-gh-aling (approved):

tests/conftest.py
sfc-gh-aling: gotcha, thanks for the explanantion.

PR #2472, opened 2024-10-17
SNOW-1615885: Don't use `snowflake_plan.attributes` in SQL simplifier and infer quoted identifiers 

sfc-gh-jdu:

   

   Fixes SNOW-1615885

   Quoted identifiers (column names) are actually already inferred in derive_column_states, we can just use it instead of calling snowflake_plan.attributes to retrieve metadata from server side

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
@@ -2,18 +2,70 @@
 # Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
 #
 
-from typing import List, Optional
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, DefaultDict, Dict, List, Optional
 
-from snowflake.snowpark._internal.analyzer.expression import Attribute
+from snowflake.snowpark._internal.analyzer.expression import Attribute, Expression, Star
 from snowflake.snowpark._internal.analyzer.snowflake_plan_node import Limit, LogicalPlan
+from snowflake.snowpark._internal.analyzer.unary_expression import UnresolvedAlias
+
+if TYPE_CHECKING:
+    from snowflake.snowpark._internal.analyzer.analyzer import Analyzer
+
+
+@dataclass(frozen=True)
+class PlanMetadata:
+    """
+    Metadata of a plan including attributes (schema) and quoted identifiers (column names).
+    """
+
+    attributes: Optional[List[Attribute]]
+    quoted_identifiers: Optional[List[str]]
+
+    def __post_init__(self):
+        # If attributes is not None, then quoted_identifiers will be explicitly set to None.
+        # If quoted_identifiers is not None, then attributes will be None because we can't infer data types.
+        assert not (self.attributes is not None and self.quoted_identifiers is not None)
+
+
+def infer_quoted_identifiers_from_expressions(
+    expressions: List[Expression],
+    analyzer: "Analyzer",
+    df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
+) -> Optional[List[str]]:
+    """
+    Infer quoted identifiers from (named) expressions.
+    The list of quoted identifier will be only returned
+    if and only if the identifier can be derived from all expressions.
+    """
+    from snowflake.snowpark._internal.analyzer.select_statement import parse_column_name
+    from snowflake.snowpark._internal.utils import quote_name
+
+    result = []
+    for e in expressions:
+        # If we do select *, we may not be able to get all current quoted identifiers
+        # (e.g., when SQL simplifier is disabled), so we just be conservative and do
+        # not perform any inference in this case.
+        if isinstance(e, UnresolvedAlias) and isinstance(e.child, Star):
+            return None
+        column_name = parse_column_name(
+            e, analyzer, df_aliased_col_name_to_real_col_name
+        )
+        if column_name is not None:
+            result.append(quote_name(column_name))
+        else:
+            return None
+    return result
 
 
 def infer_metadata(
-    source_plan: LogicalPlan,
-) -> Optional[List[Attribute]]:
+    source_plan: Optional[LogicalPlan],
+    analyzer: "Analyzer",
+    df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
+) -> PlanMetadata:
     """
     Infer metadata from the source plan.
-    Returns the metadata including attributes (schema).
+    Returns the metadata including attributes (schema) and quoted identifiers (column names).
     """
     from snowflake.snowpark._internal.analyzer.select_statement import (
         Selectable,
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -369,8 +369,17 @@ def column_states(self) -> ColumnStateDict:
         Refer to class ColumnStateDict.
         """
         if self._column_states is None:
+            if self.analyzer.session.reduce_describe_query_enabled:
+                # data types are not needed in SQL simplifier, so we
+                # just create dummy data types here.
+                column_attrs = [
+                    Attribute(q, DataType())
```
sfc-gh-jdu, line 376: I'd prefer to use this "dummy" approach instead of removing data types by converting `list[Attribute]` to `list[str]`, which will result in massive code changes (because other code relies on the fact that `column_states.project` is  `list[Attribute]`, e.g., https://github.com/snowflakedb/snowpark-python/blob/47cadd26f25781eeb7898f9e291fe83132b1cf4e/src/snowflake/snowpark/dataframe.py#L4208-L4213). They are not easily parameter protected.

src/snowflake/snowpark/session.py
```diff
@@ -589,7 +589,7 @@ def __init__(
         )
         self._reduce_describe_query_enabled: bool = (
             self._conn._get_client_side_session_parameter(
-                _PYTHON_SNOWPARK_REDUCE_DESCRIBE_QUERY_ENABLED, False
+                _PYTHON_SNOWPARK_REDUCE_DESCRIBE_QUERY_ENABLED, True
```
sfc-gh-jdu, line 592: will be flipped back once finishing testing

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -274,8 +274,18 @@ def __init__(
...
+        # If _attributes is not None, then _quoted_identifiers will be None.
+        # If _quoted_identifiers is not None, then _attributes will be None.
         if session.reduce_describe_query_enabled and self.source_plan is not None:
-            self._attributes = infer_metadata(self.source_plan)
+            self._attributes, self._quoted_identifiers = infer_metadata(
+                self.source_plan,
+                self.session._analyzer,
+                self.df_aliased_col_name_to_real_col_name,
+            )
+        assert not (
```
sfc-gh-yzou, line 286: add a comment here about why we have such check here

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -22,31 +56,61 @@ def infer_metadata(
...
-            and source_plan.from_._snowflake_plan._attributes is not None
         ):
-            attributes = source_plan.from_.snowflake_plan._attributes
+            # only set attributes and quoted_identifiers if they are not set in previous step
+            if (
+                attributes is None
+                and source_plan.from_._snowflake_plan._attributes is not None
+            ):
+                attributes = source_plan.from_._snowflake_plan._attributes
+            if (
```
sfc-gh-yzou, line 103: i think we want to do elif here, because if the attributes is already set, we do not need to set up the quoted_identifier anymore

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -2,18 +2,52 @@
...
+            return None
+    return result
 
 
 def infer_metadata(
     source_plan: LogicalPlan,
-) -> Optional[List[Attribute]]:
+    analyzer: "Analyzer",
+    df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
+) -> Tuple[Optional[List[Attribute]], Optional[List[str]]]:
```
sfc-gh-yzou, line 47: Instead of a tuple, let's get a NamedTuple Maybe call it PlanMetadata with attributes, and quoted_identifier

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -22,31 +56,61 @@ def infer_metadata(
...
+            analyzer,
+            df_aliased_col_name_to_real_col_name,
+        )
     # If source_plan is a SelectStatement, SQL simplifier is enabled
     elif isinstance(source_plan, SelectStatement):
         # When attributes is cached on source_plan, just use it
         if source_plan._attributes is not None:
             attributes = source_plan._attributes
+        # When _column_states.projection is available, just use it, which is
+        # 1) cached on self._snowflake_plan._quoted_identifiers
```
sfc-gh-helmeleegy, line 86: I don't understand the relationship between (1) and (2). Is it OR, or AND? If it's OR, then can we clarify in the comment? If it's AND, then how come it's both cached and inferred at the same time?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-jdu: > so for column expression, especially the Star(), does it require the correct DataType, or any data type is fine?
I don't think a correct data type is needed, but we need a data type object here because of alias map https://github.com/snowflakedb/snowpark-python/blob/8439f86474b1117bd3d8a73d0151aabd5466c054/src/snowflake/snowpark/_internal/analyzer/analyzer.py#L368-L371. I agree that having such `_output` method is confusing and error prone. 

> by introducing a new _output methods to selectStatement, when the flag is true, we just read the quoted_identifier and construct list of attributes with DataType, otherwise return the original projection.
I think the above implementation is exactly doing what you describe here? So how about having this implementation for now, and let's create another ticket to see whether we can refactor and remove `_output` method later? cc @sfc-gh-yixie

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-jdu: It's OR, clarified in comments

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-jdu: yes good catch!

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -2,18 +2,57 @@
...
+    from snowflake.snowpark._internal.analyzer.analyzer import Analyzer
+
+
+def infer_quoted_identifiers_from_expressions(
+    expressions: List[Union[Expression, NamedExpression]],
+    analyzer: "Analyzer",
+    df_aliased_col_name_to_real_col_name: DefaultDict[str, Dict[str, str]],
+) -> Optional[List[str]]:
+    """
+    Infer quoted identifiers from (named) expressions.
```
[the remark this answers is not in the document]
sfc-gh-jdu: done

tests/integ/test_reduce_describe_query.py
```diff
@@ -23,12 +30,16 @@
 ]
 
 param_list = [False, True]
+temp_table_name = random_name_for_temp_object(TempObjectType.TABLE)
 
 
-@pytest.fixture(params=param_list, autouse=True)
+@pytest.fixture(params=param_list, autouse=True, scope="module")
```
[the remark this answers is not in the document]
sfc-gh-jdu: yes https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -2,18 +2,59 @@
...
+from snowflake.snowpark._internal.analyzer.expression import Attribute, Expression, Star
 from snowflake.snowpark._internal.analyzer.snowflake_plan_node import Limit, LogicalPlan
+from snowflake.snowpark._internal.analyzer.unary_expression import UnresolvedAlias
+
+if TYPE_CHECKING:
+    from snowflake.snowpark._internal.analyzer.analyzer import Analyzer
+
+
+class PlanMetadata(NamedTuple):
+    attributes: Optional[List[Attribute]]
```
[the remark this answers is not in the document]
sfc-gh-jdu: good idea, done

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -2,51 +2,156 @@
...
 
-def infer_metadata(
-    source_plan: LogicalPlan,
-) -> Optional[List[Attribute]]:
+
+class PlanMetadata:
     """
-    Infer metadata from the source plan.
-    Returns the metadata including attributes (schema).
+    Metadata of a plan including attributes (schema) and quoted identifiers (column names).
```
sfc-gh-yzou, line 17: Oh sorry, i don't really mean to wrap the init with just source plan. I kind of want to make this class similar as the internalFrame in our pandas that is immutable, every time there is change just create a new one instead of do inplace update
```
@dataclass(frozen=True)
class InternalFrame:
```

The infer_quoted_identifiers_from_expressions and infer_medata data can still work like before be provided as utility function.

sfc-gh-jdu: no worries, done

sfc-gh-yzou (approved):

sfc-gh-aalam (approved):

PR #2476, opened 2024-10-18
SNOW-1738518: Test and fix timedelta in tests/integ/modin/frame: part 3/3.

sfc-gh-mvashishtha:
Fixes SNOW-1738518

Test and/or fix timedelta handling for the methods tested in tests/integ/modin/frame/, alphabetically from test_setitem.py through test_where.py.

Make the following behavior changes:
- Support timedelta in `value_counts()
- Fixed a bug where inserting timedelta values into an existing column would silently convert the values to integers instead of raising `NotImplementedError`.
- Fixed a bug where `DataFrame.shift()` on axis=0 and axis=1 would fail to propagate timedelta types.
- `DataFrame.abs()`, `DataFrame.__neg__()`, `DataFrame.stack()`, and `DataFrame.unstack()` now raise `NotImplementedError` for timedelta inputs instead of failing to propagate timedelta types.

Apart from those changes, just test that we can handle timedelta inputs correctly.


src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -96,6 +96,7 @@
     "Must have equal len keys and value when setting with an iterable"
 )
 LOC_SET_ITEM_EMPTY_ERROR = "The length of the value/item to set is empty"
+_LOC_SET_NON_TIMEDELTA_TO_TIMEDELTA_ERROR = "Snowpark pandas does not yet support assigning timedelta values to an existing column."
```
sfc-gh-lmukhopadhyay, line 99: nit - Why is only this constant an internal variable?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -18224,6 +18272,11 @@ def stack(
         sort : bool, default True
             Whether to sort the levels of the resulting MultiIndex.
         """
+        # stack() may create a column that includes values from multiple input
+        # columns. Tracking types in that case is not simple, so we don't
+        # handle the client-side timedelta type as an input.
+        self._raise_not_implemented_error_for_timedelta()
```
sfc-gh-lmukhopadhyay, line 18278: Is there a ticket to track stack and unstack for timedelta as well?

sfc-gh-lmukhopadhyay (approved):
Overall LGTM just a couple comments

tests/integ/modin/frame/test_squeeze.py
```diff
@@ -39,33 +44,54 @@ def test_1d(axis):
...
+                snow_df,
+                native_df,
                 lambda df: df.squeeze(axis=axis),
             )
+
+
+@pytest.mark.xfail(
+    strict=True,
+    raises=AssertionError,
+    reason="Transpose produces a column with both an int value and a timedelta value, so it can't preserve the timedelta type for the timdelta row.",
```
sfc-gh-nkrishna, line 90:
```suggestion
    reason="Transpose produces a column with both an int value and a timedelta value, so it can't preserve the timedelta type for the timedelta row.",
```

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -2404,13 +2413,36 @@ def generate_updated_expr_for_existing_col(
...
 
-        col_obj_type = (
-            origin_col_type
-            if col_obj_type == origin_col_type or (is_scalar(item) and pd.isna(item))
-            else None
-        )
-
-        return SnowparkPandasColumn(col_obj, col_obj_type)
+        if (
+            # In these cases, we can infer that the resulting column has a
```
sfc-gh-nkrishna, line 2417: These comments are great!

sfc-gh-nkrishna (approved):
One minor nit + please address Labanya's comments!

src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
sfc-gh-mvashishtha: Outside of this file, it's only used in tests, so it's not part of the interface of this module.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: No, there are a few methods that don't work with timedelta, and we don't plan to implement them. We'll file tickets for them as needed in response to user feedback.

tests/integ/modin/frame/test_take.py
```diff
@@ -12,7 +12,10 @@
 
 
 @pytest.mark.parametrize("test_multiindex", [False, True])
-def test_df_take(float_native_df, test_multiindex):
+@pytest.mark.parametrize("dtype", ["float", "timedelta64[ns]"])
```
sfc-gh-rdurrani, line 15: maybe just have two different dataframes that we can use? Minor nit, feel free to ignore.

PR #2488, opened 2024-10-22
SNOW-1518382, SNOW-1344931: Implement np.may_share_memory

sfc-gh-jkew:
Fixes SNOW-1344931 and SNOW-1518382 - 

Scikit-learn will often use `np.may_share_memory` on objects passed to functions. When used with snowpandas this call will immediately result in a TypeError:

`TypeError: no implementation found for 'numpy.may_share_memory' on types that implement __array_function__: [<class 'numpy.ndarray'>, <class 'modin.pandas.dataframe.DataFrame'>]`

This change uses the numpy mapping function (similar to the np.where mapping) to provide a function which always returns False. This allows for greater compatibility with scikit-learn functions.

checklist:


tests/integ/modin/test_scikit.py
```diff
@@ -0,0 +1,41 @@
...
+import pytest
+from numpy.testing import assert_array_equal
+
+import snowflake.snowpark.modin.plugin  # noqa: F401
+from tests.integ.utils.sql_counter import SqlCounter, sql_count_checker
+
+
+# SNOW-1344931 Support MaxAbsScalar by supporting numpy.may_share_memory
+def test_scikit_maxabs():    
+    from sklearn.preprocessing import MaxAbsScaler
```
sfc-gh-mvashishtha, line 13: nit: move this import to the top of the file

tests/integ/modin/test_scikit.py
```diff
@@ -0,0 +1,41 @@
...
+    # functions
+    with SqlCounter(query_count=3):
+        result = MaxAbsScaler().fit_transform(X)
+        native_result = MaxAbsScaler().fit_transform(X_native)
+        assert_array_equal(np.array(result), np.array(native_result))
+        
+# SNOW-1518382 Support PCA by supporting numpy.may_share_memory
+# similar to the test for MaxAbsScalar
+def test_scikit_pca():    
+    from sklearn.decomposition import PCA
```
sfc-gh-mvashishtha, line 31: nit: move this import to the top as well

sfc-gh-mvashishtha (approved):
thanks @sfc-gh-jkew ! `may_share_memory` was on my mind for interoperability with scikit-learn.

tests/integ/modin/test_scikit.py
sfc-gh-jkew: done

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@
 - Improved generated SQL query for `iloc` and `iat` when the row key is a scalar.
 - Removed all joins in `iterrows`.
 - Improved documentation for `Series.map` to reflect the unsupported features.
+- Added support for `np.may_share_memory` which is used internally by many scikit-learn functions. This method will always return false
```
sfc-gh-joshi, line 48:
```suggestion
- Added support for `np.may_share_memory` which is used internally by many scikit-learn functions. This method will always return false when called with a Snowpark pandas object.
```

tests/integ/modin/test_numpy.py
```diff
@@ -40,6 +40,17 @@ def test_np_where(cond, query_count):
...
 
+def test_np_may_share_memory():
+    data = {
+        "A": [0, 1, 2, 0, 1, 2, 0, 1, 2],
+    }
+    snow_df_A = pd.DataFrame(data)
+    snow_df_B = pd.DataFrame(data)
+    # np.may_share_memory always returns False
+    with SqlCounter(query_count=0):
+        assert not np.may_share_memory(snow_df_A, snow_df_B)
```
sfc-gh-joshi, line 51: Can you add a couple cases that mix native pandas/numpy arrays with Snowpark pandas objects? Also, if we call something like `np.may_share_memory(native_pd.DataFrame([]), pd.DataFrame([]))` (which I believe would use the native pandas frame's `__array_function__` rather than our implementation), would it still return False as expected?

sfc-gh-joshi (approved):
LGTM, just some minor comments.

tests/integ/modin/test_numpy.py
sfc-gh-jkew: Good point. Added tests. I think mixed arguments would still fail when you have dask and snowpandas together, in which case I think the first one seen in the argument list would be selected, but it's been a while since I have looked at the C code that iterates through those objects.

What is a bit interesting about this function though is that if I were to change it to always return "True" the scikit-learn functions would still work. Maybe something else would break, but those functions still work. I think the calling of this function within scikit-learn is a little vestigial, but don't know enough to understand exactly why.

sfc-gh-joshi: From numpy docs:

> A return of True does not necessarily mean that the two arrays share any element. It just means that they might.

I wonder if a return of `True` just makes scikit-learn more conservative, and forces it to choose algorithms that will copy data instead of operating in-place.

sfc-gh-lmukhopadhyay (approved):
Super useful, LGTM!

PR #2489, opened 2024-10-22
SNOW-1632899: make select statement breakable

sfc-gh-aalam:

   

   Fixes SNOW-1632899

   In this change, we allow `SelectStatement` nodes to be a valid candidate when no pipeline breakers are found.


src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -382,6 +382,19 @@ def get_name(node: Optional[LogicalPlan]) -> str:
             name = f"{name} :: ({get_name(node.snowflake_plan.source_plan)})"
         elif isinstance(node, SetStatement):
             name = f"{name} :: ({node.set_operands[1].operator})"
+        elif isinstance(node, SelectStatement):
+            properties = []
```
sfc-gh-aalam, line 386: adding some more details to `SelectStatement` by highlighting attributes that are set for this node.

tests/integ/test_large_query_breakdown.py
```diff
@@ -410,10 +429,11 @@ def test_sort(session):
     for i in range(160):
         df1 = df1.with_column("A", col("A") + lit(i))
         df2 = df2.with_column("B", col("B") + lit(i))
-    df1_with_sort = df1.order_by("A")
 
-    union_df = df1_with_sort.union_all(df2)
-    final_df = union_df.with_column("A", col("A") + lit(1))
+    # when sort is applied on the final dataframe, the final result should be the same
+    # it will now be broken down at select statement
```
sfc-gh-aalam, line 434:
```suggestion
    # when sort is applied on the final dataframe, the final result should be the same as
    # when no optimization is applied.
    # A cut will be made at SelectStatement
```

tests/integ/test_large_query_breakdown.py
```diff
@@ -426,19 +447,6 @@ def test_sort(session):
     assert len(plan_queries["post_actions"]) == 1
     assert plan_queries["post_actions"][0].startswith("DROP  TABLE  If  EXISTS")
 
-    # when sort is applied on the final dataframe, it should not be broken down
-    union_df = df1.union_all(df2)
-    final_df = union_df.with_column("A", col("A") + lit(1)).order_by("A")
-
```
sfc-gh-aalam, line 432: The older version of this test was ensuring that when there is a sort node within the pipeline, we break it down, and if the sort is at root node, we do not break it down.

With cutting select statement nodes being allowed, we will cut at a select statement even when only pipeline breaker(sort) is at root.

Updated test now only ensures that the final result with and without large query breakdown are same as when optimization is disabled.

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -316,14 +317,31 @@ def _find_node_to_breakdown(
                         # don't traverse subtrees if parent is a valid candidate
                         next_level.append(child)
 
+                    relaxed_validity_status, _ = self._is_node_valid_to_breakdown(
```
sfc-gh-yzou, line 320: if the node is already a valid candidate, seems we do not need to check whether it is relaxed or not anymore, right?

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -452,7 +476,9 @@ def _contains_external_cte_ref(self, node: TreeNode, root: TreeNode) -> bool:
 
         return False
 
-    def _is_node_pipeline_breaker(self, node: LogicalPlan) -> bool:
+    def _is_node_pipeline_breaker(
```
sfc-gh-yzou, line 479: this function is checking pipeline breaker, for those special, case, maybe get a separate function for it, so that we do not mess the function responsibility. Or since this is just a speical check of select statement, you can just direclty do it at function _is_node_valid_to_breakdown at the end

src/snowflake/snowpark/_internal/compiler/utils.py
sfc-gh-yzou: Why do we want so many details here? What would be the benefit?

sfc-gh-aalam: I wanted to see if select statement was breaking with only projection or if it was because of sort. so added all details here.

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
sfc-gh-aalam: It is not only checking if node is a pipeline breaker but also checking if score is valid and if there is an external CTE ref. There is enough overlap that I can repurpose this function for that.

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
sfc-gh-aalam: yeah. that is true.

sfc-gh-aalam: updated now

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
sfc-gh-yzou: In fact, i looked at the code for _is_node_valid_to_breakdown, you might want to introduce a status called VALID_RELAXED_NODE, and introduce a function called _is_relaxed_pipeline_breaker function, and when _is_relaxed_pipeline_breaker and _is_node_pipeline_breaker are both false, set the valid status to InvalidNodesInBreakdownCategory.NON_PIPELINE_BREAKER, if _is_node_pipeline_breaker is true, keep the original valid status, if _is_node_pipeline_breaker is false, and _is_relaxed_pipeline_breaker is true, set the valid status to VALID_RELAXED_NODE, then at the caller you just need to check the valid status

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -316,14 +317,33 @@ def _find_node_to_breakdown(
                         # don't traverse subtrees if parent is a valid candidate
                         next_level.append(child)
 
+                        # If node is not a pipeline breaker, we allow select statements to be
+                        # considered as valid candidates for partitioning.
+                        relaxed_validity_status, _ = self._is_node_valid_to_breakdown(
```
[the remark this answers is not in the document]
sfc-gh-aalam: yes we do want to explore children if parent is not a pipeline breaker. This is how pipeline breaker even with a lower score will have precedence over relaxed candidate with higher score.

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -356,7 +380,7 @@ def _get_partitioned_plan(self, root: TreeNode, child: TreeNode) -> SnowflakePla
         return temp_table_plan
 
     def _is_node_valid_to_breakdown(
-        self, node: TreeNode, root: TreeNode
+        self, node: TreeNode, root: TreeNode, relaxed: bool
```
sfc-gh-yzou, line 383: why do we need a relaxed flag here? _is_node_valid_to_breakdown function should be able to handle detect node is strict or relaxed in one pass, we shouldn't need to call this function twice to detect valid node or relaxed valid node. What is the benefit of having the relaxed flag here?

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -377,10 +407,17 @@ def _is_node_valid_to_breakdown(
             is_valid = False
             validity_status = InvalidNodesInBreakdownCategory.SCORE_ABOVE_UPPER_BOUND
 
-        if is_valid and not self._is_node_pipeline_breaker(node):
+        # check pipeline breaker condition
+        pipeline_breaker = (
```
sfc-gh-yzou, line 411: you should just need to do checks at this place in one pass, if it is neither  _is_node_pipeline_breaker or _is_relaxed_pipeline_breaker, then validty_status will be InvalidNodesInBreakdownCategory.NON_PIPELINE_BREAKER

if _is_node_pipeline_breaker is True, node status doesn't need to change
if not _is_node_pipeline_breaker and _is_relaxed_pipeline_breaker, node status changed to InvalidNodesInBreakdownCategory.VALID_NODE_RELAXED

sfc-gh-aalam: true. can be done in a single pass.

sfc-gh-yzou (approved):

PR #2494, opened 2024-10-23
[SNOW-1758768] Add repeated node complexity distribution telemetry

sfc-gh-yzou:

   

SNOW-1758768

Add telemetry to record the distribution of repeated nodes complexity. Note, we are recording the number of repeated nodes, not just the cte nodes. For example, if n2 occured twice during the visit of the whole tree, it will be counted twice, instead of once.

The complexity bucket used today are :
```
 1) low complexity
        bin 0: <= 10,000; bin 1: > 10,000, <= 100,000; bin 2: > 100,000, <= 500,000
    2) medium complexity
        bin 3: > 500,000, <= 1,000,000;  bin 4: > 1,000,000, <= 5,000,000
    4) large complexity
        bin 5: > 5,000,000, <= 10,000,000;  bin 6: > 10,000,000
```

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -363,16 +363,12 @@ class SnowflakePlan(LogicalPlan):
         return self._output_dict
 
     @cached_property
-    def num_duplicate_nodes(self) -> int:
-        duplicated_nodes = find_duplicate_subtrees(self)
-        return len(duplicated_nodes)
-
-    @cached_property
     def plan_state(self) -> Dict[PlanState, Any]:
         from snowflake.snowpark._internal.analyzer.select_statement import (
             SelectStatement,
         )
 
+        # calculate plan height and num_selects_with_complexity_merged
         height = 0
         num_selects_with_complexity_merged = 0
         current_level = [self]
@@ -387,9 +383,16 @@ class SnowflakePlan(LogicalPlan):
                     num_selects_with_complexity_merged += 1
             height += 1
             current_level = next_level
+        # calculate the repeated node status
+        cte_nodes, duplicated_node_complexity_distribution = find_duplicate_subtrees(
+            self, propagate_complexity_hist=True
+        )
+
         return {
             PlanState.PLAN_HEIGHT: height,
             PlanState.NUM_SELECTS_WITH_COMPLEXITY_MERGED: num_selects_with_complexity_merged,
+            PlanState.NUM_CTE_NODES: len(cte_nodes),
+            PlanState.DUPLICATED_NODE_COMPLEXITY_DISTRIBUTION: duplicated_node_complexity_distribution,
         }
 
     @property
--- a/src/snowflake/snowpark/_internal/compiler/cte_utils.py
+++ b/src/snowflake/snowpark/_internal/compiler/cte_utils.py
@@ -5,16 +5,22 @@
 import hashlib
 import logging
 from collections import defaultdict
-from typing import TYPE_CHECKING, Dict, Optional, Set
+from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple
 
 from snowflake.snowpark._internal.analyzer.snowflake_plan_node import WithQueryBlock
+
+from snowflake.snowpark._internal.analyzer.query_plan_analysis_utils import (
+    get_complexity_score,
+)
 from snowflake.snowpark._internal.utils import is_sql_select_statement
 
 if TYPE_CHECKING:
     from snowflake.snowpark._internal.compiler.utils import TreeNode  # pragma: no cover
 
 
-def find_duplicate_subtrees(root: "TreeNode") -> Set[str]:
+def find_duplicate_subtrees(
+    root: "TreeNode", propagate_complexity_hist: bool = False
+) -> Tuple[Set[str], Optional[List[int]]]:
     """
     Returns a set of TreeNode encoded_id that indicates all duplicate subtrees in query plan tree.
     The root of a duplicate subtree is defined as a duplicate node, if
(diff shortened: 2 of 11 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -387,9 +383,16 @@ def plan_state(self) -> Dict[PlanState, Any]:
...
             current_level = next_level
+        # calculate the repeated node status
+        cte_nodes, duplicated_node_complexity_distribution = find_duplicate_subtrees(
+            self, propagate_complexity_hist=True
+        )
+
         return {
             PlanState.PLAN_HEIGHT: height,
             PlanState.NUM_SELECTS_WITH_COMPLEXITY_MERGED: num_selects_with_complexity_merged,
+            PlanState.NUM_CTE_NODES: len(cte_nodes),
```
sfc-gh-jdu, line 394: it's the same as "num_duplicate_nodes" before, right? I'm not sure whether we should change the name since it will break the existing query in dashboard

src/snowflake/snowpark/_internal/compiler/cte_utils.py
```diff
@@ -71,12 +84,57 @@ def is_duplicate_subtree(encoded_node_id_with_query: str) -> bool:
...
+    else:
+        return (duplicated_node_ids, None)
+
+
+def get_duplicated_node_complexity_distribution(
+    duplicated_node_id_set: Set[str], id_complexity_map: Dict[str, List[int]]
+) -> List[int]:
+    """
+    Calculate the complexity distribution for the detected repeated node. The complexity are categorized as following:
+    1) low complexity
```
sfc-gh-jdu, line 109: just curious, where are these ranges from?

tests/integ/test_telemetry.py
```diff
@@ -607,6 +607,7 @@ def test_execute_queries_api_calls(session, sql_simplifier_enabled):
             "query_plan_height": query_plan_height,
             "query_plan_num_duplicate_nodes": 0,
```
sfc-gh-jdu, line 608: you might want to remove this

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yzou: There is no change to the telemetry node, it is just a change for our internal code to be more clear at least from our internal developer's point of view. If you look at the telemetry sent here https://github.com/snowflakedb/snowpark-python/pull/2494/files#diff-c46aa591b19bb02d4b80a69dda3c5b4ff9e2ff4ad18cf1170779aa678ccbdafaR194 it is still using the same name

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-yzou: the upper bound 10,000,000 is from the default exhaust bound, the rest is really just by sense, if we have other suggestions about the bins, please feel free to put the suggestion, i just don't want to send a big list for the complexity if we have a lot of duplications

tests/integ/test_telemetry.py
sfc-gh-yzou: oh, why is that? we are still sending this. So the query_plan_num_duplicate_nodes is the number of cte nodes. the query_plan_duplicated_node_complexity_distribution is the actual distribution of repeated nodes. so it is different, if we sum all of them together, we will get the actual number of repeated nodes, not the cte nodes.

sfc-gh-jdu: yeah I missed that

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/compiler/cte_utils.py
```diff
@@ -71,12 +84,57 @@ def is_duplicate_subtree(encoded_node_id_with_query: str) -> bool:
...
+    4) large complexity
+        bin 5: > 5,000,000, <= 10,000,000;  bin 6: > 10,000,000
+
+    Returns:
+        A list with size 7, each element corresponds number of repeated nodes with complexity falls into the bin.
+    """
+    node_complexity_dist = [0] * 7
+    for node_id in duplicated_node_id_set:
+        for complexity_score in id_complexity_map[node_id]:
+            if complexity_score <= 10000:
```
sfc-gh-aalam, line 122: can't we do

```
complexity_score = id_complexity_map[node_id]
repetition_count = id_count_map[node_id]
node_complexity_dist[..] += repetition_count
```

src/snowflake/snowpark/_internal/compiler/cte_utils.py
```diff
@@ -48,6 +57,10 @@ def traverse(root: "TreeNode") -> None:
             next_level = []
             for node in current_level:
                 id_count_map[node.encoded_node_id_with_query] += 1
+                if propagate_complexity_hist:
+                    id_complexity_map[node.encoded_node_id_with_query].append(
+                        get_complexity_score(node)
+                    )
```
sfc-gh-aalam, line 63: shouldn't two nodes with same `encoded_node_id_with_query` have the same complexity? we are simply going to make a list of `[complexity] * id_count_map[node_id]`

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-yzou: so i was recording the complexity for each node because theoritically they can be different node end with same query. However, if they are the same query, theoretically i would in general expect the complexity to be same or close, since that information is mainly used for estimation, i think we can just use the repetition_count to make things simpiler

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-yzou: answered below

PR #2497, opened 2024-10-23
SNOW-1738538 Fix multi ast eval validation in decoding expectation test ast

sfc-gh-evandenberg:
Fix multi ast eval validation in decoding expectation test ast

### DIFF
--- a/src/snowflake/snowpark/_internal/ast_utils.py
+++ b/src/snowflake/snowpark/_internal/ast_utils.py
@@ -1346,18 +1346,19 @@ def base64_str_to_request(base64_str: str) -> proto.Request:
     return message
 
 
-def merge_requests(messages: List[proto.Request]) -> proto.Request:
-    message = proto.Request()
+def merge_requests(requests: List[proto.Request]) -> proto.Request:
+    """Merge list of requests into a single request through accumulating the request body segments in same order."""
+    request = proto.Request()
 
     # Copy the client_version, etc as part of first message.
-    message.CopyFrom(messages[0])
+    request.CopyFrom(requests[0])
 
-    for next_message in messages[1:]:
-        for next_stmt in next_message.body:
-            stmt = message.body.add()
+    for next_request in requests[1:]:
+        for next_stmt in next_request.body:
+            stmt = request.body.add()
             stmt.CopyFrom(next_stmt)
 
-    return message
+    return request
 
 
 def base64_lines_to_request(base64_lines: str) -> proto.Request:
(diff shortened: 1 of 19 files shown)


src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -1342,7 +1342,19 @@ def ClearTempTables(message: proto.Request) -> None:
 
 def base64_str_to_request(base64_str: str) -> proto.Request:
```
sfc-gh-lspiegelberg, line 1343: I think we should be here a bit more careful with API naming and semantics.

Technically, `"\n"` is not a valid character in base64. The function itself right now does exactly one thing, which is taking a valid base64_str (for which the user is responsible, we can do a check) and decodes it. I would suggest to keep the basic function as is, but for convenience add a new one which does the repeated decode and then merges the result together.

I.e.,
```
def base64_str_to_request(base64_str: str) -> proto.Request:
  ....
 < add check that characters are only in `[0-9+/=a-zA-Z]` which is base64 encoding>
 
 def base64_lines_to_requst(base64_lines: str) -> proto.Request:
    # allow here character "\n"
    requests = [base64_str_to_request]
    return merge_requests(requests)
```

This is I think anyways the logic you have in mind here, but easier to understand

sfc-gh-evandenberg: Updated

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -1342,12 +1342,31 @@ def ClearTempTables(message: proto.Request) -> None:
...
     message = proto.Request()
-    message.ParseFromString(base64.b64decode(base64_str.strip()))
+    message.ParseFromString(base64.b64decode(base64_str))
     return message
 
 
-def base64_str_to_textproto(base64_str: str) -> str:
-    request = base64_str_to_request(base64_str)
+def merge_requests(messages: List[proto.Request]) -> proto.Request:
+    message = proto.Request()
```
sfc-gh-lspiegelberg, line 1350: nit: maybe add docstring comment on what the function does (i.e., merge all requests into single request. Order is the same as messages).

nit: rename potentially messages -> requests?

sfc-gh-lspiegelberg (approved):

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-lspiegelberg: Empty case is also not handled in this function. I.e.

```
if not messages:
   return message
```

sfc-gh-evandenberg: Updated

PR #2498, opened 2024-10-23
[SNOW-1760495] Reduce test errors on AST enabled merge gate / pass test_dataframe.py

sfc-gh-lspiegelberg:
Fixes several tests to reduce test failures for the merge-gate with `session.ast_enabled=True`. With this PR all tests except for the interval related ones pass for `test_dataframe.py`.

Details:
- Adds `ast_id` to shallow copy protocol for Dataframe.
- Modify APIs so checks happen before AST encoding to pass negative tests.
- Add missing `ast_stmt` to `select(table_function(...))`
- Fix filter test.
- Fix overflow issue when using `datetime.time` due to timezone encoding.
- Support array type (array.py from stdlib) by mapping to list.
- Fix na functions to carry out checks for `subset` parameter.
- Explicitly type check for `col` and `column` as expected in existing tests.
- Add `NotImplementedError` for interval type in `make_interval`.
- Fix wrong usage of Expression in `function.col(...)`, use `Column` instead in `apply_in_pandas`.


src/snowflake/snowpark/dataframe.py
```diff
@@ -878,7 +879,9 @@ def __copy__(self) -> "DataFrame":
             new_plan._query_params = self._select_statement.query_params
         else:
             new_plan = copy.copy(self._plan)
-        return DataFrame(self._session, new_plan)
+        df = DataFrame(self._session, new_plan)
+        df._ast_id = self._ast_id
```
sfc-gh-oplaton, line 883: If `DataFrame` never, ever changes in place, this approach is safe. If any meaningful `DataFrame` API mutates the object in place, using the same variable ID for two objects will lead to very surprising results at some point in the future.

We might have this pattern in other places, too.

src/snowflake/snowpark/dataframe_na_functions.py
```diff
@@ -185,10 +197,8 @@ def drop(
         # if subset is not provided, drop will be applied to all columns
         if subset is None:
             subset = self._dataframe.columns
-        elif isinstance(subset, str):
+        if isinstance(subset, str):
```
sfc-gh-oplaton, line 200: This change seems unnecessary, `elif` was doing the correct thing.

src/snowflake/snowpark/dataframe_na_functions.py
```diff
@@ -376,10 +388,8 @@ def fill(
 
         if subset is None:
             subset = self._dataframe.columns
-        elif isinstance(subset, str):
+        if isinstance(subset, str):
```
sfc-gh-oplaton, line 391: Unnecessary?

sfc-gh-oplaton (approved):
Generally looks good. Minor inconsistencies in handling `subset` when it's `None` vs `str`. 2 functions now use `if` after `if`, the third one uses the previous pattern.

src/snowflake/snowpark/dataframe.py
```diff
@@ -5287,7 +5306,10 @@ def random_split(
             if seed:
                 ast.seed.value = seed
             if statement_params:
-                ast.statement_params = statement_params
+                for k, v in statement_params.items():
```
sfc-gh-evandenberg, line 5309: Reuse ast_utils.build_expr_from_dict_str_str
```suggestion
                build_expr_from_dict_str_str(ast.statement_params, statement_params)
```

src/snowflake/snowpark/dataframe_na_functions.py
```diff
@@ -63,6 +63,16 @@ def _is_value_type_matching_for_na_function(
...
 
 
+def _check_subset_parameter(subset: Optional[Union[str, Iterable[str]]]) -> None:
+    """Produces exception when invalid subset parameter was passed."""
+    if (
+        subset is not None
+        and not isinstance(subset, str)
+        and not isinstance(subset, (list, tuple))
+    ):
+        raise TypeError("subset should be a list or tuple of column names")
```
sfc-gh-evandenberg, line 73:
```suggestion
        raise TypeError("subset should be a single column name, list or tuple of column names")
```

src/snowflake/snowpark/dataframe.py
sfc-gh-evandenberg: What is reason for this, performance?

sfc-gh-evandenberg (approved):
LGTM with some suggestions.

src/snowflake/snowpark/dataframe.py
sfc-gh-lspiegelberg: I think this a copy-by-reference vs. deep-copy concern here. Indeed we should be careful regarding this. Atm, this is done using copy-by-reference. However, we may want to add a separate IR API to emulate the copy behavior (maybe reuse cache?)

cc @sfc-gh-azwiegincew 

I'll add a JIRA with todo so we can pickup the copy behavior there.

src/snowflake/snowpark/dataframe_na_functions.py
sfc-gh-lspiegelberg: good spot, unified to `elif`

src/snowflake/snowpark/dataframe_na_functions.py
sfc-gh-lspiegelberg: unified to `elif`

src/snowflake/snowpark/dataframe_na_functions.py
sfc-gh-lspiegelberg: I like the suggestion, sounds much better. Will have to update existing tests.

PR #2499, opened 2024-10-23
SNOW-1518378 - Provide a numpy compatibility mapping to np.full_like

sfc-gh-jkew:
SNOW-1518378 - Provide a numpy compatibility mapping to np.full_like

This provides a numpy compat mapping to np.full_like; which fills a dataframe/series with a fill value the same shape as the input.

checklist:


tests/integ/modin/test_numpy.py
```diff
@@ -57,6 +57,45 @@ def test_np_may_share_memory():
...
+        assert_array_equal(np.array(snow_result), np.array(pandas_result))
+
+    with pytest.raises(TypeError):
+        np.full_like(snow_df, 1234, subok=False)
+
+    with pytest.raises(TypeError):
+        np.full_like(snow_df, 1234, order="D")
+
+    with pytest.raises(TypeError):
+        np.full_like(snow_df, 1234, dtype=int)
```
sfc-gh-mvashishtha, line 96: Can we split these into another test case?

sfc-gh-jkew: I'm trying to keep it to one test per api; because we may end up moving things around later and we only want to test the mapping itself - not exactly numpy.

tests/integ/modin/test_numpy.py
```diff
@@ -86,6 +86,17 @@ def test_full_like():
...
         assert_array_equal(np.array(snow_result), np.array(pandas_result))
 
+    with SqlCounter(query_count=1):
+        pandas_result = np.full_like(pandas_df, fill_value=4, shape=())
+        # breakpoint()
+        snow_result = np.full_like(snow_df, fill_value=4, shape=())
+        assert_array_equal(np.array(snow_result), np.array(pandas_result))
+
+    with SqlCounter(query_count=1):
+        snow_result = np.full_like(snow_df, fill_value=4, shape=(4))
```
sfc-gh-mvashishtha, line 96: this is equivalent to `shape=4`. did you mean to test `shape=(4,)`?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
```diff
@@ -112,6 +112,36 @@ def may_share_memory_mapper(a: Any, b: Any, max_work: Optional[int] = None) -> b
...
+) -> Union[pd.DataFrame, pd.Series]:
+    if not subok:
+        return NotImplemented
+    if not order == "K":
+        return NotImplemented
+    if dtype is not None:
+        return NotImplemented
+
+    result_shape = shape
+    if isinstance(result_shape, tuple) and len(result_shape) == 0:
```
sfc-gh-mvashishtha, line 131: nit: this won't do the right thing for `np.full_like(pd.Series(1), 4, shape=[])`. I think you want `is_list_like` or something like that.

tests/integ/modin/test_numpy.py
```diff
@@ -57,6 +57,56 @@ def test_np_may_share_memory():
...
+        assert_array_equal(np.array(snow_result), np.array(pandas_result))
+
+    with SqlCounter(query_count=1):
+        snow_result = np.full_like(snow_df, "numpy is the best")
+        pandas_result = np.full_like(pandas_df, "numpy is the best")
+        assert_array_equal(np.array(snow_result), np.array(pandas_result))
+
+    with SqlCounter(query_count=1):
+        pandas_result = np.full_like(pandas_df, fill_value=4, shape=())
+        # breakpoint()
```
sfc-gh-mvashishtha, line 91:
```suggestion
```

tests/integ/modin/test_numpy.py
sfc-gh-jkew: Added (4,); but I actually did want (4) in this case. So now I have both cases.

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-jkew: good catch

tests/integ/modin/test_numpy.py
sfc-gh-jkew: removed

tests/integ/modin/test_numpy.py
sfc-gh-mvashishtha: I would delete the `(4)` test case since it means the same thing as `4` and it looks like it means `tuple(4)`.

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-mvashishtha: Are you still missing a fix for this place a `shape=[]` test case?

sfc-gh-mvashishtha (commented):
I still have a couple of small unresolved comments

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-jkew: I will return a TypeError in this case.

sfc-gh-mvashishtha (approved):

CHANGELOG.md
```diff
@@ -80,6 +80,9 @@
 - Added support for applying Snowpark Python function `snowflake_cortex_summarize`.
 - Added support for `DataFrame.attrs` and `Series.attrs`.
 - Added support for `DataFrame.style`.
+- Added support for `Index.to_numpy`.
```
sfc-gh-rdurrani, line 83: These are probably merge artifacts?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
```diff
@@ -112,6 +112,36 @@ def may_share_memory_mapper(a: Any, b: Any, max_work: Optional[int] = None) -> b
...
+    dtype: Optional[Any] = None,
+    order: Optional[str] = "K",
+    subok: Optional[bool] = True,
+    shape: Optional[tuple[Any]] = None,
+) -> Union[pd.DataFrame, pd.Series]:
+    if not subok:
+        return NotImplemented
+    if not order == "K":
+        return NotImplemented
+    if dtype is not None:
```
sfc-gh-rdurrani, line 127: Why can't we support dtype here? It just overrides the datatype provided by a right?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
```diff
@@ -112,6 +112,36 @@ def may_share_memory_mapper(a: Any, b: Any, max_work: Optional[int] = None) -> b
...
+        return NotImplemented
+
+    result_shape = shape
+    if isinstance(result_shape, tuple) and len(result_shape) == 0:
+        result_shape = (1,)
+    if isinstance(result_shape, int):
+        result_shape = (result_shape,)
+    if result_shape is None:
+        result_shape = a.shape
+    if len(result_shape) == 2:
```
sfc-gh-rdurrani, line 137: Maybe check if `result_shape` is list-like? otherwise something like result_shape = 'ab' would enter this conditional?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
```diff
@@ -112,6 +112,36 @@ def may_share_memory_mapper(a: Any, b: Any, max_work: Optional[int] = None) -> b
...
+        height, width = result_shape  # type: ignore
+        return pd.DataFrame(fill_value, index=range(height), columns=range(width))
```
sfc-gh-rdurrani, line 139: Don't we need to set dtype here? Also if a DataFrame contains multiple dtypes, are the returned object's columns supposed to be the same type as each of the columns mapped positionally?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
```diff
@@ -112,6 +112,36 @@ def may_share_memory_mapper(a: Any, b: Any, max_work: Optional[int] = None) -> b
...
+    if len(result_shape) == 1:
+        return pd.Series(fill_value, index=range(result_shape[0]))
```
sfc-gh-rdurrani, line 141: dtype?

tests/integ/modin/test_numpy.py
```diff
@@ -57,6 +57,63 @@ def test_np_may_share_memory():
...
+        "B": [True, False, True, True, False, True, False, False, False],
+        "C": ["a", "b", "c", "d", "a", "b", "c", "d", "e"],
+    }
+    snow_df = pd.DataFrame(data)
+    pandas_df = native_pd.DataFrame(data)
+
+    with SqlCounter(query_count=2):
+        snow_result = np.full_like(snow_df, 1234)
+        pandas_result = np.full_like(pandas_df, 1234)
+        assert_array_equal(np.array(snow_result), np.array(pandas_result))
```
sfc-gh-rdurrani, line 72: Should we be checking if the dataframes are equal?

sfc-gh-rdurrani (changes requested):
Left some nits, but great work overall!

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-jkew: These are numpy types; and within the context of mapping them to snowpandas they don't make sense and it would be overly complicated to support the options.

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-jkew: No. Everything ends up being the fill_value, and where there are mixed types the resulting dataframe is an object. I think that's a bug in pandas/numpy we don't necessarily want to keep. There's other bugs in the pandas implementation. You can use full_like with a mixed type dataframe with any full_value; but if you use it on a dataframe with only a single type the fill_value has to be the same.

The pandas implementation doesn't make a lot of sense. Here we basically just pass in the value and do the default thing.

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-jkew: I don't think that can happen. Numpy parses the shape argument as well into integers.

CHANGELOG.md
sfc-gh-jkew: yup, thanks!

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-rdurrani: Just tried this on your branch:
```python
In [1]: im();

In [2]: df = pd.DataFrame([[1, 2, 3], [4, 5, 6]])

In [4]: np.full_like(df, 4, shape='ab')
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 np.full_like(df, 4, shape='ab')

File ~/Documents/snowpark-python/src/snowflake/snowpark/modin/plugin/extensions/base_extensions.py:38, in __array_function__(self, func, types, args, kwargs)
     33 from snowflake.snowpark.modin.plugin.utils.numpy_to_pandas import (
     34     numpy_to_pandas_func_map,
     35 )
     37 if func.__name__ in numpy_to_pandas_func_map:
---> 38     return numpy_to_pandas_func_map[func.__name__](*args, **kwargs)
     39 else:
     40     # per NEP18 we raise NotImplementedError so that numpy can intercept
     41     return NotImplemented

File ~/Documents/snowpark-python/src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py:139, in full_like_mapper(a, fill_value, dtype, order, subok, shape)
    137 if len(result_shape) == 2:
    138     height, width = result_shape  # type: ignore
--> 139     return pd.DataFrame(fill_value, index=range(height), columns=range(width))
    140 if len(result_shape) == 1:
    141     return pd.Series(fill_value, index=range(result_shape[0]))

TypeError: 'str' object cannot be interpreted as an integer
```

This is what happens if a string is passed in - I think this message is probably ok right?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-rdurrani: That makes sense to me, but can we document these divergences in behavior?

sfc-gh-rdurrani (approved):
LGTM - minor nit - maybe we should document how our behavior diverges from pandas for this API?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-jkew: Yeah, it's same as w/ numpy essentially.

PR #2503, opened 2024-10-24
SNOW-1722639: Add support for DataFrame.style

sfc-gh-helmeleegy:

   

   Fixes SNOW-1722639

   Add support for DataFrame.style.


docs/source/modin/supported/dataframe_supported.rst
```diff
@@ -46,7 +46,7 @@ Attributes
 +-----------------------------+---------------------------------+----------------------------------------------------+
 | ``size``                    | Y                               |                                                    |
 +-----------------------------+---------------------------------+----------------------------------------------------+
-| ``style``                   | N                               |                                                    |
+| ``style``                   | Y                               |                                                    |
```
sfc-gh-nkrishna, line 49: Should we add a note here saying this defaults to pandas?

src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
```diff
@@ -5125,7 +5125,23 @@ def attrs():
...
+        --------
+        io.formats.style.Styler
+        Helps style a DataFrame or Series according to the data with HTML and CSS.
+
+        Examples
+        --------
+        >>> df = pd.DataFrame({'A': [1, 2, 3]})
+        >>> df.style
+
+        Please see `Table Visualization <https://pandas.pydata.org/docs/user_guide/style.html>`_ for more examples.
```
sfc-gh-nkrishna, line 5143: Should we add a Notes section here in the docstring mentioning that this defaults to pandas behavior?

sfc-gh-nkrishna (approved):
LGTM, just some documentation questions

tests/integ/modin/frame/test_style.py
sfc-gh-joshi, line 1: Can you test a couple simple style operations like in https://github.com/pandas-dev/pandas/blob/68d9dcab5b543adb3bfe5b83563c61a9b8afae77/doc/source/user_guide/style.ipynb#L4 ? It seems like the `Styler` object is meant to be immutable so it shouldn't make a difference if we chain operations to `df.style`, but I think it would be good to be safe.

docs/source/modin/supported/dataframe_supported.rst
sfc-gh-jjiao: > Should we add a note here saying this defaults to pandas?

+1, good idea

sfc-gh-jjiao (approved):

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
```diff
@@ -382,10 +382,9 @@ def __delitem__(self, key):
 
 
 @register_dataframe_accessor("style")
-@dataframe_not_implemented()
 @property
 def style(self):  # noqa: RT01, D200
-    pass  # pragma: no cover
+    return self._to_pandas().style
```
sfc-gh-joshi, line 387: I think materializing is fine for now, but if performance becomes a concern we should consider returning a custom object in the future. A styler can be configured to hide rows/columns and filter data, which we would be able to do before any materializations.

sfc-gh-joshi (approved):

tests/integ/modin/frame/test_style.py
sfc-gh-helmeleegy: We are already using `df.style.to_string()` in the test. Why would the other operations behave differently than `to_string()`?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
sfc-gh-helmeleegy: Agreed. There could be opportunities to push some of the styler post processing down to the snowflake engine. But I'd defer this for now - until there is a clear signal from customers that this is high priority. I think the current implementation handles the most common use cases. Note that the [docs](https://github.com/pandas-dev/pandas/blob/68d9dcab5b543adb3bfe5b83563c61a9b8afae77/doc/source/user_guide/style.ipynb#L4) say:
> The Styler, which can be used for large data but is primarily designed for small data, currently has the ability to output to these formats

PR #2505, opened 2024-10-24
SNOW-1738540 Add Nop / No-execution mode for running ast expectation tests

sfc-gh-evandenberg:
This adds a NopConnection, NopAnalyzer, NopExecutionPlan which adds a new backend that does no execution, ie. nothing is executed or sent to the server.  This is useful for AST expectation testing to avoid blocking on local testing gaps and performance issues, in many cases the gaps blocked writing tests.  The biggest challenge is supporting a no execution mode is there is schema validation code scattered throughout the dataframe API that fails without attributes.  To mitigate this the NopAnalyzer makes a best effort to figure out the dataframe attributes, while it isn't guaranteed to always be correct in all cases, it effectively unblocks tests because majority of the time complete accuracy is not required on the client side.  Note there is no use case for AST serialization in local testing since nothing is send to the snowflake server.

Fixed Session.table_function.test issue where the table function wasn't being passed in a valid form.

### DIFF
--- a/src/snowflake/snowpark/mock/_nop_analyzer.py
+++ b/src/snowflake/snowpark/mock/_nop_analyzer.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
+from typing import Dict, List, Union
+
+from snowflake.snowpark._internal.analyzer.expression import Expression, Star
+from snowflake.snowpark._internal.analyzer.select_statement import (
+    Selectable,
+    SelectSnowflakePlan,
+    SelectStatement,
+)
+from snowflake.snowpark._internal.analyzer.snowflake_plan_node import LogicalPlan
+from snowflake.snowpark.mock._analyzer import MockAnalyzer
+from snowflake.snowpark.mock._nop_plan import NopExecutionPlan, resolve_attributes
+from snowflake.snowpark.mock._plan import MockExecutionPlan
+from snowflake.snowpark.mock._select_statement import (
+    MockSelectableEntity,
+    MockSelectExecutionPlan,
+    MockSelectStatement,
+    MockSetOperand,
+    MockSetStatement,
+)
+
+
+class NopSetStatement(MockSetStatement):
+    @property
+    def attributes(self):
+        return [val.expression for val in self.column_states.data.values()]
+
+
+class NopSelectStatement(MockSelectStatement):
+    @property
+    def attributes(self):
+        return (
+            resolve_attributes(self.from_)
+            if isinstance(self.projection[0], Star)
+            else self.projection
+        )
+
+    def _make_nop_select_statement_copy(self, statement: "MockSelectStatement"):
+        if isinstance(statement, NopSelectStatement):
+            return statement
+
+        nop_statement = NopSelectStatement(from_=self.from_, analyzer=self.analyzer)
+        nop_statement.__dict__.update(statement.__dict__)
+        return nop_statement
+
+    def select(self, cols: List[Expression]) -> "SelectStatement":
+        return self._make_nop_select_statement_copy(super().select(cols))
+
+    def filter(self, col: Expression) -> "MockSelectStatement":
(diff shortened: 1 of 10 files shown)


sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/mock/_nop_analyzer.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
```
sfc-gh-lspiegelberg, line 1: Should we remove as part of this PR (or maybe followup) code like the following which has been used so far to make some tests pass? This here looks like a MUCH BETTER solution, so I'd prefer to get rid off the old code.

```
 if (
            isinstance(self._session._conn, MockServerConnection)
            and self._session._conn._suppress_not_implemented_error
        ):
            # Allow AST tests to pass.
            return []
```

links for examples:
- https://github.com/snowflakedb/snowpark-python/blob/ls-SNOW-1491199-merge-phase0-server-side/src/snowflake/snowpark/dataframe.py#L1834
- https://github.com/snowflakedb/snowpark-python/blob/ls-SNOW-1491199-merge-phase0-server-side/src/snowflake/snowpark/dataframe.py#L4009

Ideally we would remove from session the `_suppress_not_implemented_error` attribute.

src/snowflake/snowpark/mock/_nop_connection.py
```diff
@@ -0,0 +1,208 @@
...
+        ) -> List[Row]:
+            current_schema = self.conn._get_current_parameter("schema")
+            current_database = self.conn._get_current_parameter("database")
+            name = get_fully_qualified_name(name, current_schema, current_database)
+            self.table_registry[name] = table
+            return [Row(status=f"Table {name} successfully created.")]
+
+    def __init__(self, options: Optional[Dict[str, Any]] = None) -> None:
+        super().__init__(options)
+        self._conn._session_parameters = {
```
sfc-gh-lspiegelberg, line 83: nit: _session_parameters should be Dict[str,str], at least many APIs use that as type hint.

src/snowflake/snowpark/mock/_nop_analyzer.py
sfc-gh-evandenberg: Created https://snowflakecomputing.atlassian.net/browse/SNOW-1782942 to follow up on this and other clean up of dead code related to emit ast processing in local test mode.

src/snowflake/snowpark/mock/_nop_connection.py
sfc-gh-evandenberg: MockServerConnection defines as Dict[str, Any] and I checked a few places the retrieval code assumes the value is a boolean.

sfc-gh-lspiegelberg: that sounds like something out of scope for this PR, but we prob. should consider fixing in the existing API. So basically the `statement_params` is not uniquely typed. Ideally in the AST we would represent this with one object/type.

tests/ast/data/RelationalGroupedDataFrame.test
```diff
@@ -45,9 +45,9 @@ res8 = df.group_by("a").count()
 
 df = session.create_dataframe([("SF", 21.0), ("SF", 17.5), ("SF", 24.0), ("NY", 30.9), ("NY", 33.6)], schema=["location", "temp_c"])
```
sfc-gh-lspiegelberg, line 47: is the DoubleType() -> FloatType() change correct?

sfc-gh-lspiegelberg (approved):
lgtm, only thing I am wondering about is the type change.

tests/ast/data/RelationalGroupedDataFrame.test
sfc-gh-evandenberg: This looks potentially a bug or difference in how local testing mode works.

For local testing, uses _extract_schema_and_data_from_pandas_df which maps to DoubleType while for non-local testing, uses the mapping PYTHON_TO_SNOW_TYPE_MAPPINGS (type_utils.py) to map 'float' -> 'FloatType'.

PR #2517, opened 2024-10-26
Skip child replace for new query compilation stage if the child has been updated already

sfc-gh-yzou:
In replace_child, if the child of a parent has already been updated, the old child won't exist anymore, and an error could raise with the current code. However, if the child has already been updated with the new child, we should simply skip the update.

For this case, the original plan have 4 resolved snowflake plan node that is pointing to the same source plan, therefore updating the child of one parent automatically updates the other parents.

daily precommit test verification https://github.com/snowflakedb/snowpark-python/actions/runs/11545868356


sfc-gh-aalam (approved):

sfc-gh-jdu (approved):

sfc-gh-helmeleegy (approved):
lgtm

PR #2540, opened 2024-10-30
SNOW-1748104,SNOW-1748107, SNOW-1748108: Add support for pd.read_pickle/read_html/read_xml

sfc-gh-helmeleegy:

   

   Fixes SNOW-1748104,SNOW-1748107, SNOW-1748108

   Add support for pd.read_pickle/read_html/read_xml.


tests/integ/modin/io/test_read_xml.py
```diff
@@ -0,0 +1,39 @@
...
+from io import StringIO
+
+import modin.pandas as pd
+import pandas as native_pd
+
+from tests.integ.modin.utils import assert_frame_equal
+from tests.integ.utils.sql_counter import SqlCounter
+
+
+def test_read_xml():
```
sfc-gh-jkew, line 13: test_read_xml is the only one I would really want tests on - specifically for the xpath, iterparse.

src/snowflake/snowpark/modin/plugin/io/snow_io.py
```diff
@@ -593,28 +593,18 @@ def read_gbq(
...
-    ):
-        pass  # pragma: no cover
+    def read_html(cls, **kwargs) -> list[SnowflakeQueryCompiler]:
+        """
+        Read HTML tables into a list of query compilers.
+        """
+        return [cls.from_pandas(df) for df in pandas.read_html(**kwargs)]
+
+    @classmethod
+    def read_xml(cls, **kwargs) -> SnowflakeQueryCompiler:
```
sfc-gh-jkew, line 603: To prevent xml entity resolution security problems I would throw an error if:
* storage_options is used as a parameter
* if parser is changed from the default, lxml
* if stylesheet is used
* and /possibly/ if namespaces is used.

I know pandas supports all these but we can add support in later if need be and we don't have to deal with that class of issue

src/snowflake/snowpark/modin/plugin/io/snow_io.py
```diff
@@ -665,13 +655,11 @@ def read_sas(cls, **kwargs):  # noqa: PR01
...
 
     @classmethod
-    @pandas_module_level_function_not_implemented()
-    def read_pickle(
-        cls,
-        filepath_or_buffer,
-        **kwargs,
-    ):
-        pass  # pragma: no cover
+    def read_pickle(cls, **kwargs) -> SnowflakeQueryCompiler:
```
sfc-gh-jkew, line 658: This one... I think I would either not implement this one, or force the user to acknowledge that they trust the underlying data they are de-pickling by forcing the use of an argument:

`pd.read_pickle(unpicking_trusted_data=False)`

Only if the user flips that to "True" would it make sense. I would run that by @sfc-gh-dolee though.

sfc-gh-jkew (changes requested):

tests/integ/modin/io/test_read_xml.py
sfc-gh-helmeleegy: Added more tests.

src/snowflake/snowpark/modin/plugin/io/snow_io.py
sfc-gh-helmeleegy: John and I discussed offline. Below is the summary of our convo.

1- Even if we blocked some of the native pandas functionality when directly calling Snowpark pandas, users can still access that same blocked functionality by simply calling native pandas themselves and then passing the resulting df into Snowpark pandas:

```
native_df = pandas.read_*(<some_param_combination>)
df = pd.DataFrame(native_df)
```

2- We are exposing the exact same signatures as native pandas, and furthermore we are explicitly mentioning in our docs that we are using native pandas internally to support those methods. So the user should not be surprised by any behavior because of using specific native pandas functionalities that they explicitly invoked.

3- Based on the above, we believe the liability should not be on us (Snowflake) but rather on the end users in those cases (similar to the case when users are simply using native pandas without using Snowpark pandas at all).

@sfc-gh-dolee would love to get your input as well.

sfc-gh-helmeleegy: @sfc-gh-jkew I checked with Doris and she agrees with the above.

In her words:
> I agree with your points in the comments. I don’t see how this would pose any additional security risk over what we already support with native pandas today.

sfc-gh-jkew (approved):

sfc-gh-nkumar (approved):

PR #2542, opened 2024-10-31
SNOW-1777142: Implement groupby.agg(len)

sfc-gh-mvashishtha:
Fixes SNOW-1777142.

Map `len` to the same Snowpark aggregation functions that `"size"` maps to.

Also fix a bug where aggregating a single-column dataframe with a single callable function (e.g. `pd.DataFrame([0]).agg(len)`) would fail to transpose the result. One of the test cases that we already have for `agg("size")` revealed this bug for `agg(len)`.


CHANGELOG.md
```diff
@@ -26,7 +26,11 @@
 
 - Added support for `Index.to_numpy`.
 - Added support for `DataFrame.align` and `Series.align` for `axis=0`.
-- Added support for `size` in `GroupBy.aggregate`, `DataFrame.aggregate`, and `Series.aggregate`.
+- Added support for aggregation functions `"size"` and `len` in `GroupBy.aggregate`, `DataFrame.aggregate`, and `Series.aggregate`.
```
sfc-gh-vbudati, line 29: nit: Is there a reason `size` is in quotes but len isn't?

sfc-gh-mvashishtha: You can agg with the function `len` but not the string `"len"`, and you can agg with the string `"size"` but usually not with `size` :)

sfc-gh-vbudati (approved):
Thanks Mahesh!

sfc-gh-joshi (approved):
LGTM

sfc-gh-joshi:
Actually, one q: should we update the support docs for this? We explicitly list the aggregation functions we support, but would we just consider this a subset of `agg("size")`?

sfc-gh-mvashishtha:
> should we update the support docs for this? 

Good point! Done.

CHANGELOG.md
sfc-gh-helmeleegy: Are these limitations coming from native pandas?!

sfc-gh-mvashishtha: `size` is a variable that's not bound to anything in python. the note here was wrong before my change since we could aggregate by `"size"` but not by `size`.

CHANGELOG.md
```diff
@@ -30,6 +30,11 @@
 - Added support for `pd.read_pickle` (Uses native pandas for processing).
 - Added support for `pd.read_html` (Uses native pandas for processing).
 - Added support for `pd.read_xml` (Uses native pandas for processing).
+- Added support for aggregation functions `"size"` and `len` in `GroupBy.aggregate`, `DataFrame.aggregate`, and `Series.aggregate`.
+
+#### Bug Fixes
+
+- Fixed a bug where aggregating a single-column dataframe with a single callable function (e.g. `pd.DataFrame([0]).agg(np.mean)`) would fail to transpose the result.
```
sfc-gh-helmeleegy, line 37: Is there a test added for this fix?

sfc-gh-helmeleegy (approved):
Overall, lgtm. Just left a couple of questions.

CHANGELOG.md
sfc-gh-mvashishtha: Yes, the test case is tests/integ/modin/frame/test_aggregate.py::test_agg_with_single_col

PR #2545, opened 2024-10-31
SNOW-859943: Add basic support for functions.window

sfc-gh-jrose:

   

   Fixes SNOW-859943

   This PR adds partial support for functions.window. This involved a few notable changes:

- Moved utils around time part aliases from local testing mock utils to the internal anlyzer utils. The pyspark equivalent uses interval strings to support this feature so we need some amount of parsing logic for compatibility.
- This function does not have a sql equivalent so it is instead composed of dataframe operations that are equivalent to the logic needed. 
    - The resulting column would have an overly verbose name so I've instead aliased it to the default value of `window`. These columns are often used as aggregate keys though which cannot be aliased. In order to support this use case I've modified the aggregation logic to remove the alias if present. This has the side-effect of also allowing users to have a statement like this: `df.groupby(upper(col("cat")).alias("cat").agg(...)`. The resulting aggregate key column would have the name `cat` instead of `UPPER(""CAT"")"`
    - Users may want to immediately re-alias the column so I've modified `Column` so that if you alias an already aliased column it replaces the alias instead of trying to alias twice. This allows a statement like `col("cat").alias("a1").alias("a2")` which results in a column of name `a2`.
- This addition does not have full parity with the pyspark window function yet.  The slideDuration parameter causes this function to go from a simple mapping to something that would require a udtf. That would have a significantly different performance profile so I've left it unimplemented until we can get better sql support.
- Modified local testing groupby to allow an arbitrary function expression. As far as I can tell, in live mode function expressions are allowed as groupby keys as long as they aren't aggregate or generator expressions.


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 - Added support for `Index.to_numpy`.
 - Added support for `DataFrame.align` and `Series.align` for `axis=0`.
 - Added support for `size` in `GroupBy.aggregate`, `DataFrame.aggregate`, and `Series.aggregate`.
+- Added partial support for `snowflake.snowpark.functions.window`
```
sfc-gh-aalam, line 30: move it up to 1.25.0

sfc-gh-aalam: should we call it partial support? lets documents its capabilities to function description and not word it as partial support.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -224,6 +224,66 @@ class TempObjectType(Enum):
...
+    "weekiso": {"weekiso", "week_iso", "weekofyeariso", "weekofyear_iso"},
+    "yearofweek": {"yearofweek"},
+    "yearofweekiso": {"yearofweekiso"},
+    "epoch_second": {"epoch_second", "epoch", "epoch_seconds"},
+    "epoch_millisecond": {"epoch_millisecond", "epoch_milliseconds"},
+    "epoch_microsecond": {"epoch_microsecond", "epoch_microseconds"},
+    "epoch_nanosecond": {"epoch_nanosecond", "epoch_nanoseconds"},
+    "timezone_hour": {"timezone_hour", "tzh"},
+    "timezone_minute": {"timezone_minute", "tzm"},
+}
```
sfc-gh-aalam, line 263: did you do a lot of trial and error for this? Are these aliases documented somewhere?

src/snowflake/snowpark/functions.py
```diff
@@ -4594,6 +4595,90 @@ def dayofyear(e: ColumnOrName) -> Column:
...
 
+def window(
+    timeColumn: ColumnOrName,
+    windowDuration: str,
+    slideDuration: Optional[str] = None,
+    startTime: Optional[str] = None,
+) -> Column:
+    """
+    Converts a time column into a window object with start and end times. Window start times areinclusive
+    inclusive while end times are exclusive. For example 9:30 is in the window [9:30, 10:00), but not [9:00, 9:30).
```
sfc-gh-aalam, line 4606:
```suggestion
    Converts a time column into a window object with start and end times. Window start times are
    inclusive while end times are exclusive. For example 9:30 is in the window [9:30, 10:00), but not [9:00, 9:30).
```

src/snowflake/snowpark/functions.py
```diff
@@ -4594,6 +4595,90 @@ def dayofyear(e: ColumnOrName) -> Column:
     return builtin("dayofyear")(c)
 
 
+def window(
+    timeColumn: ColumnOrName,
+    windowDuration: str,
+    slideDuration: Optional[str] = None,
+    startTime: Optional[str] = None,
```
sfc-gh-aalam, line 4602: camelCase is generally not pythonic. Can we convert this into snake_case

src/snowflake/snowpark/functions.py
```diff
@@ -4594,6 +4595,90 @@ def dayofyear(e: ColumnOrName) -> Column:
...
+        <BLANKLINE>
+    """
+    if slideDuration:
+        # SNOW-1063685: slideDuration changes this function from a 1:1 mapping to a 1:N mapping. That
+        # currently would require a udtf which may have significantly different performance.
+        raise NotImplementedError(
+            "snowflake.snowpark.functions.window does not support slideDuration parameter yet."
+        )
+
+    epoch = lit("1970-01-01 00:00:00").cast(TimestampType())
```
sfc-gh-aalam, line 4660: do we need to care about timezone here?

src/snowflake/snowpark/functions.py
```diff
@@ -4594,6 +4595,90 @@ def dayofyear(e: ColumnOrName) -> Column:
...
+        |  "start": "2024-10-31 09:02:00.000"  |
+        |}                                     |
+        ----------------------------------------
+        <BLANKLINE>
+    """
+    if slideDuration:
+        # SNOW-1063685: slideDuration changes this function from a 1:1 mapping to a 1:N mapping. That
+        # currently would require a udtf which may have significantly different performance.
+        raise NotImplementedError(
+            "snowflake.snowpark.functions.window does not support slideDuration parameter yet."
```
sfc-gh-aalam, line 4657: since @sfc-gh-qding volunteered as a sql expert for our team, we can discuss if doing this is possible without udtf.

src/snowflake/snowpark/_internal/utils.py
sfc-gh-jrose: There's a handy chart here: https://docs.snowflake.com/en/sql-reference/functions-date-time#label-supported-date-time-parts

src/snowflake/snowpark/functions.py
sfc-gh-jrose: I agree, but to maintain compatibility with the pyspark implementation I think camelCase might be required. I could modify so that either version is accepted.

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -871,9 +871,16 @@ def execute_mock_plan(
...
-                    )
+                    column_name = plan.session._analyzer.analyze(exp)
+                    if isinstance(exp, FunctionExpression):
+                        materialized_column = calculate_expression(
+                            exp, child_rf, plan.session._analyzer, expr_to_alias
+                        )
+                        # Only function expressions that are a mapping of existing columns can be aggregated on.
+                        # Any increase or reduction in number of rows is an invalid function expression.
+                        if len(materialized_column) == len(child_rf):
+                            child_rf[column_name] = materialized_column
```
sfc-gh-aling, line 882: is this related to the window function support in local testing as groupby expression?
also I don't understand why we want to materialized_column here -- what's the case it's addressing. is there an example?

tests/integ/test_df_aggregate.py
sfc-gh-aling, line 1: do we have test for the new window function?

I saw pyspark that it has example of using groupby + windows function: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.window.html

src/snowflake/snowpark/functions.py
sfc-gh-aling: IHMO api style consistency within the snowpark lib outweighs consistency with pyspark.

src/snowflake/snowpark/mock/_plan.py
sfc-gh-jrose: This is related to having any function expression in the groupby. For example:
```df2 = df.group_by(upper(df.a)).agg(max_(df.b))```
This is perfectly valid in live mode, but does not work in local testing at the moment.

tests/integ/test_df_aggregate.py
sfc-gh-jrose: I'll add a doctest example that shows it being used as a groupby expression.

sfc-gh-aalam (approved):

src/snowflake/snowpark/functions.py
sfc-gh-aalam: yup. the kwarg conversion can be handled by conversion tools

sfc-gh-aling (approved):

PR #2548, opened 2024-10-31
SNOW-1760495 Reduce number of failing tests for merge gate with enabled AST

sfc-gh-lspiegelberg:
Reduces number of failing tests from 39 to 26 when enabling AST for snowpark.

Details:
- Separated logic to check types/values of parameters for `.agg()` in both dataframe/grouped dataframe into separate helper function.
- Fixed AST emit logic for COUNT which now for COUNT(*) emits * instead of COUNT(1) in the unparser.
- Fix tests where statement_params were used with incorrect type, added additional check for correct type in testing.
- Fix multi-thread error in MockConnection when notifying listeners by adding global mutex.

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -1171,6 +1171,8 @@ def check_agg_exprs(
     """Helper function to raise exceptions when invalid exprs have been passed."""
     from snowflake.snowpark import Column
 
+    exprs, _ = parse_positional_args_to_list_variadic(*exprs)
+
     # special case for single list or tuple
     if is_valid_tuple_for_agg(exprs):
         exprs = [exprs]
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -1155,6 +1155,45 @@ def check_flatten_mode(mode: str) -> None:
         raise ValueError("mode must be one of ('OBJECT', 'ARRAY', 'BOTH')")
 
 
+def is_valid_tuple_for_agg(e: Union[list, tuple]) -> bool:
+    from snowflake.snowpark import Column
+
+    return len(e) == 2 and isinstance(e[0], (Column, str)) and isinstance(e[1], str)
+
+
+def check_agg_exprs(
+    exprs: Union[
+        "snowflake.snowpark.Column",
+        Tuple["snowflake.snowpark.ColumnOrName", str],
+        Dict[str, str],
+    ]
+):
+    """Helper function to raise exceptions when invalid exprs have been passed."""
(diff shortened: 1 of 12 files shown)


tests/ast/data/functions.test
```diff
@@ -664,7 +664,7 @@ df32 = df.select(avg(col("A")), avg(col("B")))
 
 df33 = df.select(corr(col("A"), col("B")))
```
sfc-gh-lspiegelberg, line 666: the encoding here used to be wrong. I.e., due to some logic `count(*)` got demoted to `count(1)`. Yet, `count(*)` is closer to original. Changed here to be more correct.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1155,6 +1155,47 @@ def check_flatten_mode(mode: str) -> None:
         raise ValueError("mode must be one of ('OBJECT', 'ARRAY', 'BOTH')")
```
sfc-gh-lspiegelberg, line 1157: Refactored from agg (dataframe.py and relational_grouped_dataframe.py) to move checks before AST generation.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1918,6 +1919,9 @@ def agg(
             - :meth:`DataFrame.group_by`
         """
 
+        # Check exprs arguments.
```
sfc-gh-azwiegincew, line 1922: This comment doesn't add much value thanks to all the well chosen names in this context.

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -257,18 +263,15 @@ def agg(
...
-                and isinstance(e[1], str)
-            )
-
         exprs, is_variadic = parse_positional_args_to_list_variadic(*exprs)
+
         # special case for single list or tuple
         if is_valid_tuple_for_agg(exprs):
             exprs = [exprs]
 
+        # Check exprs parameter.
```
sfc-gh-azwiegincew, line 272: Ditto.

tests/utils.py
```diff
@@ -353,9 +353,19 @@ def check_answer(
         actual: Union[Row, List[Row], DataFrame],
         expected: Union[Row, List[Row], DataFrame],
         sort=True,
-        statement_params=None,
+        statement_params: Optional[Dict[str, str]] = None,
         float_equality_threshold=0.0,
     ) -> None:
+
+        # Check that statement_params are passed as Dict[str,str]
```
sfc-gh-azwiegincew, line 360: Space after `,`

sfc-gh-azwiegincew (approved):

PR #2136, opened 2024-08-21
[SNOW-1491199] Snowpark server-side work phase0

sfc-gh-lspiegelberg:
Introduces AST encoding as part of server-side snowpark phase0 work.

PR #2552, opened 2024-11-01
SNOW-1491199 Server-Side Snowpark Phase 0: Core AST functionality and utilities

sfc-gh-oplaton:

   

   Fixes SNOW-1491199

Introduce functionality that supports the AST collection logic.
Add `batch.py` and `utils.py` to `src/snowflake/snowpark/_internal/ast`.

`AstBatch` supports the collection of sequences of statements within the `Session` object. Statements can be either `Assign` or `Eval`. `AstBatch` accumulates statements and can then flush them into a `Request` object.

`ast/utils.py` is the workhorse of the AST collection functionality. The job of the functions in this module is to traverse Python object graphs that might be used as inputs to Snowpark APIs. The `build_*` and `fill_*` functions are the core of the AST extraction logic.

Follow-up work:

* Enable `pyright` checks for the `ast` directory.
* Improve test coverage when bringing in the rest of the AST collection changes, and remove the `no cover` pragmas.


src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,94 @@
...
+from dataclasses import dataclass
+from typing import Callable, Tuple
+
+import snowflake.snowpark._internal.proto.generated.ast_pb2 as proto
+from snowflake.snowpark.version import VERSION
+
+# The current AST version number (generated by the DSL).
+CLIENT_AST_VERSION = proto.__Version__.MAX_VERSION
+
+
```
sfc-gh-lspiegelberg, line 19: shall we remove `expr_to_daframe_expr`? it's unused.

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,94 @@
...
+    var_id: int
+    func: Callable
+
+
+class AstBatch:
+    # Function used to generate request IDs. This is overridden in some tests.
+    generate_request_id = uuid.uuid4
+
+    def __init__(self, session) -> None:
+        self._session = session
```
sfc-gh-lspiegelberg, line 39: Don't want to bloat the scope of this PR unnecessarily, but I feel this PR may be a good place to add proper docstrings to all of the methods of AstBatch here. Some documentation could be useful for the team, especially for folks to understand assign/eval concept.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: Done.

tests/unit/test_ast.py
```diff
@@ -0,0 +1,56 @@
...
+        assert next(ast_batch._id_gen) == 1
+
+    def test_assign(self, fake_session):
+        ast_batch = AstBatch(fake_session)
+        stmt = ast_batch.assign()
+        assert stmt is not None
+
+    def test_eval(self, fake_session):
+        ast_batch = AstBatch(fake_session)
+        ast_batch.eval(ast_batch.assign("foo"))
```
sfc-gh-azhan, line 37: what to assert?

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,114 @@
...
+
+import snowflake.snowpark._internal.proto.generated.ast_pb2 as proto
+from snowflake.snowpark.version import VERSION
+
+# The current AST version number (generated by the DSL).
+CLIENT_AST_VERSION = proto.__Version__.MAX_VERSION
+
+
+@dataclass
+class TrackedCallable:
```
sfc-gh-azhan, line 21: can you add comment to explain what this is used for?

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,114 @@
...
+        stmt.eval.var_id.CopyFrom(target.var_id)
+
+    def flush(self) -> Tuple[str, str]:
+        """Ties off a batch and starts a new one. Returns the tied-off batch."""
+        batch = str(base64.b64encode(self._request.SerializeToString()), "utf-8")
+        self._init_batch()
+        return (str(self._request_id), batch)
+
+    def _init_batch(self):
+        self._request_id = AstBatch.generate_request_id()  # Generate a new unique ID.
```
sfc-gh-azhan, line 88: add a single line comment to explain what init do here.

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,114 @@
...
+from typing import Callable, Tuple
+
+import snowflake.snowpark._internal.proto.generated.ast_pb2 as proto
+from snowflake.snowpark.version import VERSION
+
+# The current AST version number (generated by the DSL).
+CLIENT_AST_VERSION = proto.__Version__.MAX_VERSION
+
+
+@dataclass
```
sfc-gh-azhan, line 20: why this is a dataclass?

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python3
```
sfc-gh-azhan, line 1: Can we move those AST files under a new folder `ast/` and add its ownership to IR team only.
Also, can we add this new ast folder in pyright check in `tox.ini` and make sure type hints have been added for ast code.

tests/unit/test_ast.py
```diff
@@ -0,0 +1,56 @@
+#
```
sfc-gh-azhan, line 1: let's move those tests under tests/unit/ast and add ownership to IR team.

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,114 @@
...
+    def assign(self, symbol=None):
+        """
+        Creates a new assignment statement.
+
+        Args:
+            symbol: An optional symbol to name the new variable.
+        """
+        stmt = self._request.body.add()
+        # TODO: extended BindingId spec from the branch snowpark-ir.
+        stmt.assign.uid = next(self._id_gen)
```
sfc-gh-azhan, line 65: Can we have a method called `get_next_id()`?

sfc-gh-azhan, line 65: Do we need to make sure AST is thread safe? cc @sfc-gh-aalam @sfc-gh-aling 

If current analyzer is thread safe, then AST code should be thread safe too right?

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,114 @@
...
+        Creates a new evaluation statement.
+
+        Args:
+            target: The variable to evaluate.
+        """
+        stmt = self._request.body.add()
+        stmt.eval.uid = next(self._id_gen)
+        stmt.eval.var_id.CopyFrom(target.var_id)
+
+    def flush(self) -> Tuple[str, str]:
```
sfc-gh-azhan, line 81: let's use namedtuple for better readability.

src/snowflake/snowpark/_internal/ast.py
```diff
@@ -0,0 +1,114 @@
...
+
+        (major, minor, micro, releaselevel, serial) = sys.version_info
+        self._request.client_language.python_language.version.major = major
+        self._request.client_language.python_language.version.minor = minor
+        self._request.client_language.python_language.version.patch = micro
+        self._request.client_language.python_language.version.label = releaselevel
+
+        self._request.client_ast_version = CLIENT_AST_VERSION
+
+    # TODO(SNOW-1491199) - This method is not covered by tests until the end of phase 0. Drop the pragma when it is covered.
```
sfc-gh-azhan, line 104: nit: maybe we can add callable related code later.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -0,0 +1,1370 @@
...
+# The path to the snowpark package.
+SNOWPARK_LIB_PATH = Path(__file__).parent.parent.resolve()
+
+# Test mode. In test mode, the source filename is ignored.
+SRC_POSITION_TEST_MODE = False
+
+_logger = getLogger(__name__)
+
+
+# TODO(SNOW-1491199) - This method is not covered by tests until the end of phase 0. Drop the pragma when it is covered.
```
sfc-gh-azhan, line 59: Since all code here are not used and tested, why we include it here? We could include those utility file when it is used in the future PRs.

The reason I suggest removing it here is that
1. It is hard to review all of them without context and tests;
2. If we just let it in, later we may forget to review them.

sfc-gh-oplaton: We discussed this when splitting the changes. To see how all code is used together, we can look at the staging branch, https://github.com/snowflakedb/snowpark-python/tree/ls-SNOW-1491199-merge-phase0-server-side

Instead of pursuing a single, huge PR (https://github.com/snowflakedb/snowpark-python/pull/2136), our finer-grained choices are to merge at roughly module level, in a bottom-up fashion (this approach), or at an even finer-grained level, "vertically" - add a set of features and all internal functionality that they depend on.

Splitting the top level change at the finest-grained level is a lot more work than organizing modules in a bottom-up fashion. The context is available in https://github.com/snowflakedb/snowpark-python/pull/2136

tests/unit/test_ast.py
sfc-gh-oplaton: There aren't any particularly interesting properties to test after this step. I can delete the test altogether, since it's already the same functionality that we test in `test_flush`, or I can leave it in place as a smoke test (if it passes, it means nothing is obviously broken).

sfc-gh-azhan: ok, maybe just add comment say this is a smoke test.

sfc-gh-oplaton: Done.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: Added docstring.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: I assume the question implies that a named tuple would work equally well. I have a very minor preference for data classes over named tuples, but not for internal types, it's a lot less important here.

I left things as they are for now. (But added a named tuple for the return value of `AstBatch.flush()`, so now we do things slightly inconsistently).

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: Done.

sfc-gh-oplaton: When the thread safety work came up, the discussion I remember was that `_ast_batch` is a field of `Session`, and thus needs to be thread safe similarly to `Session`. If this happens by virtue of methods getting called through APIs already synchronized by the `Session` mutex, nothing to do for us. Otherwise, we can add internal synchronization to `AstBatch`. I just need to know what the contract needs to be.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: As I mentioned before, merging back into `main` in a very fine-grained fashion probably doesn't pass a cost/benefit analysis - it's a lot more work to sort the changes this way, and for very limited benefit, since we aim to have all the production code merged within another 5-10 pull requests.

src/snowflake/snowpark/_internal/ast/ast.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    Several Snowpark APIs that deal with stored procedures and user-defined functions accept callables as arguments.
+    This class is a pair of a callable and an ID that is used to reference it in the AST. Distinct objects get distinct IDs.
+    It is undesirable for the same callable to have multiple IDs due to constraints in other parts of the system.
+    """
+
+    var_id: int
+    func: Callable
+
+
```
sfc-gh-lspiegelberg, line 32: nit: should this be a dataclass as well? it makes not much sense to use a namedtuple.

Why?
-> why should this be true at all? `isinstance(SerializedBatch, tuple)`?

sfc-gh-lspiegelberg (approved):

src/snowflake/snowpark/_internal/ast.py
sfc-gh-lspiegelberg: namedtuple is a bad idea, unless you need the `is-a` tuple relationship - no need to use it.

Inheritance should be used carefully, as it comes with certain implications.

Can we use dataclass here to stay consistent?

sfc-gh-azhan: Oh my original question is why this needs to be immutable? I guess that's the reason to use `dataclass` right?

If it just a simple group of immutable items, I would prefer using namedtuple. If you are going to define methods then go with `dataclass`. Just a suggestion.

sfc-gh-lspiegelberg: It think it boils down to modeling here, each object has different contracts:

What is a tuple?
A tuple is fixed-length sequence of items (so, order matters!).
-> A namedtuple expands a tuple. So a namedtuple "is-a" tuple.

What is a dataclass?
A dataclass is a trait to mix into a class.

For context, cf.

https://peps.python.org/pep-0557/#rationale

> No base classes or metaclasses are used by Data Classes. Users of these classes are free to use inheritance and metaclasses without any interference from Data Classes. The decorated classes are truly “normal” Python classes. The Data Class decorator should not interfere with any usage of the class.

https://peps.python.org/pep-0557/#why-not-just-use-namedtuple

A namedtuple has a good chunk of disadvantages, my biggest pet peeve here is that it messes with the inheritance. Imho, inheritance should be used when it makes sense to specify relationships. Traits/mixins/interfaces are preferred when possible.

sfc-gh-azhan: Yes, if we don't need inheritance or just used it as a immutable message, namedtuple is fine.

src/snowflake/snowpark/_internal/ast/ast.py
sfc-gh-oplaton: The original signature of `AstBatch.flush()` advertised the return type as a `Tuple[str, str]`.

If I make this a `dataclass`, all consumers of that serialized batch (not many, but a few places in the code) will need coordinated changes. I think we should do that in a follow-up PR.

sfc-gh-azhan: I think namedtuple should be fine since we are not going to inherit on it and it only add a name to the existing tuple.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: Done. I also renamed `ast.py` to `batch.py` to avoid having to deal with `ast/ast.py`.

sfc-gh-azhan: great! let's update ownership file and enable this folder with pyright.

sfc-gh-oplaton: * Updated ownership
* Added pyright checks

Pyright checks yield around 130 warnings and errors. A small number of them are actual minor issues. A much larger number of them has to do with protocol buffers and setting up tooling (e.g. `"Expr" is not a known member of module "snowflake.snowpark._internal.proto.generated.ast_pb2"`).

I'll file a separate task and do the pyright work in a separate pull request, if that's OK with you.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-lspiegelberg: @sfc-gh-azhan the other way round :) Because we do not need inheritance and a relationship to tuple, imho we should use dataclass. Tbh, I do not think that this a very important detail. Yet, it would make sense to keep things succinct with @dataclass.

sfc-gh-azhan: I try to understand why the relationship with tuple is a problem?

src/snowflake/snowpark/_internal/ast.py
sfc-gh-azhan: Yes, thanks for exploring this. I see people use type-protobuf instead but it seems only works with protobuf~=5.28.3. Let's revisit it later.

sfc-gh-oplaton: I fully support getting the type annotations right, around our protocol buffer types and in Python, by the way. I find it critical for understanding the code base at all.

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -0,0 +1,133 @@
...
+
+    The core statement types are:
+    - Assign: Creates a new variable and assigns a value to it.
+    - Eval: Evaluates a variable.
+    """
+
+    # Function used to generate request IDs. This is overridden in some tests.
+    generate_request_id = uuid.uuid4
+
+    def __init__(self, session) -> None:
```
sfc-gh-azhan, line 51: Can we have all arguments and return value with type hints? Unfortunately we don't have a lint for it due to protobuf. So let's try to treat this as a best practice for now. 

Sorry I was insisted on this but the type hints actually help a lot when we deploying pandas APIs. It is especially helpful to detect bugs earlier.

Let me try to see how to enforce pyright on this later.

tests/unit/test_ast.py
sfc-gh-azhan: Please help to move this too.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-azhan: make sense. Let's move forward since no concern here.
Also, let's add one line comment to describe `AstBatch` is not thread safe for now.

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-vbudati: +1

sfc-gh-oplaton: Done.

src/snowflake/snowpark/_internal/ast.py
sfc-gh-oplaton: Done.

tests/unit/test_ast.py
sfc-gh-oplaton: Sorry, just noticed this comment. For whatever reason, it doesn't appear in the main "conversation" view on the pull request. Doing this now.

sfc-gh-oplaton: Done.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -799,6 +807,155 @@ def warning(name: str, text: str, warning_times: int = 1) -> None:
...
+            return session.ast_enabled  # noqa: B012
+
+
+def publicapi(func) -> Callable:
+    """decorator to safeguard public APIs with global feature flags."""
+
+    # TODO(SNOW-1491199) - This method is not covered by tests until the end of phase 0. Drop the pragma when it is covered.
+    @functools.wraps(func)
+    def func_call_wrapper(*args, **kwargs):  # pragma: no cover
+        # warning(func.__qualname__, warning_text)
```
sfc-gh-azhan, line 855: remove this line?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -799,6 +807,155 @@ def warning(name: str, text: str, warning_times: int = 1) -> None:
...
+                kwargs["_emit_ast"] = args[0].ast_enabled
+            elif isinstance(
+                args[0],
+                snowflake.snowpark.relational_grouped_dataframe.RelationalGroupedDataFrame,
+            ):
+                kwargs["_emit_ast"] = args[0]._df._session.ast_enabled
+            else:
+                kwargs["_emit_ast"] = infer_ast_enabled_from_global_sessions(func)
+
+        # TODO: Could modify internal docstring to display that users should not modify the _emit_ast parameter.
```
sfc-gh-azhan, line 952: Do you mean customers should not modify? Other users like us or other developers can?

sfc-gh-azhan (approved):

PR #2560, opened 2024-11-04
SNOW-1760495 Reduce more merge gate test failures

sfc-gh-lspiegelberg:
Reduces merge gate (ast-enabled) test failures from 26 down to 7.

Details:
- Fixes usage of `lit` in test suite, adds `datatype` parameter.
- Fixes table init usaage
- Fixes sampleby, drop_duplicates by setting correct ast ref
- Fixes negative UDTF, UDAF test
- Refactors checks for imports/output_schema for UDxFs/sprocs.


src/snowflake/snowpark/modin/plugin/_internal/timestamp_utils.py
```diff
@@ -443,43 +443,35 @@ def raise_if_to_datetime_not_supported(
...
 
 def convert_dateoffset_to_interval(
-    value: native_pd.DateOffset,
+    value: native_pd.DateOffset, _emit_ast: bool = True
 ) -> Interval:
     """
     Converts a pandas DateOffset where value is treated as a timedelta to a Snowpark
     Interval keyword. DateOffset with parameters that replace the offset value is not
     yet supported, so a NotImplemented error is raised.
     """
```
sfc-gh-lspiegelberg, line 452: Was talking with @sfc-gh-nkrishna about this here, after some discussion concluded to add comment to clarify what's missing in pandas.

src/snowflake/snowpark/functions.py
```diff
@@ -358,15 +360,19 @@ def lit(literal: LiteralType, _emit_ast: bool = True) -> Column:
         ---------------------------------------------------------------------------------------
         <BLANKLINE>
     """
+
+    assert not isinstance(
+        literal, Column
+    ), "Do not use lit(Column(...)), type hint does not allow this syntax."
```
sfc-gh-vbudati, line 366: Is this just reinforcing the warning issued by the type hint? It makes sense to keep the assert if the warning is silent/can be easily ignored.
I see that in the rest of the file there's a lot of code of the form `n if isinstance(n, Column) else lit(n)` and this assertion is supposed to prevent folks from simply doing `lit(n)`?

src/snowflake/snowpark/window.py
```diff
@@ -255,6 +255,15 @@ def _spec() -> "WindowSpec":
     rowsBetween = rows_between
 
 
+def _check_window_position_parameter(
+    value: Union[int, WindowRelativePosition], name=""
```
sfc-gh-vbudati, line 259: nit: add type hint for `name`

sfc-gh-vbudati (commented):
Left a few nits but looks great! Thanks Leonhard!

src/snowflake/snowpark/functions.py
sfc-gh-lspiegelberg: I decided to add an explicit assert here to prevent wrong usage. Unfortunately, in the code there's often something like `lit(Column(...))`. This is not correct though, because a `Column` can be an arbitrary expression. To fix this, I added the `isinstance(...)` code. I was debating to introduce a helper function, but felt there's not really a point for a single if stmt. Better to be explicit there.

sfc-gh-vbudati: Makes sense, thank you for making this change!

PR #2569, opened 2024-11-05
SNOW-1787507: Fix DataFrame.dropna() with subset empty, scalar, or native_pd.Index.

sfc-gh-mvashishtha:
Fixes SNOW-1787507

- Fix bugs where `DataFrame.dropna()` would:
  - Treat an empty `subset` (e.g. `[]`) as if it specified all columns instead of no columns.
  - Raise a `TypeError` for a scalar `subset` instead of filtering on just that column.
  - Raise a `ValueError` for a `subset` of type `pandas.Index` instead of filtering on the columns in the index.


sfc-gh-nkrishna (approved):
Thanks for the quick fix!

sfc-gh-vbudati (approved):

PR #2573, opened 2024-11-06
SNOW-1787415 Fix binary op expr with alias and send dataframe ast only once for multi-queries

sfc-gh-evandenberg:
SNOW-1787415 Fix binary op expr with alias and send dataframe ast only once for multi-queries

### DIFF
--- a/tests/ast/ast_test_utils.py
+++ b/tests/ast/ast_test_utils.py
@@ -45,13 +45,13 @@ def render(ast_base64: Union[str, List[str]], unparser_jar: Optional[str]) -> st
     return res.stdout
 
 
-def generate_error_trace_info(python_text, ex=None):
+def generate_error_trace_info(python_text, exception=None):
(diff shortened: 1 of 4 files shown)


tests/ast/ast_test_utils.py
```diff
@@ -44,11 +45,13 @@ def render(ast_base64: Union[str, List[str]], unparser_jar: Optional[str]) -> st
     return res.stdout
 
 
-def generate_error_trace_info(python_text):
+def generate_error_trace_info(python_text, ex=None):
```
sfc-gh-vbudati, line 48: nit: I think it would be more readable to have exception/a full word as the parameter name instead of ex.

sfc-gh-vbudati (approved):
Thanks for fixing this!

tests/ast/ast_test_utils.py
sfc-gh-evandenberg: Updated

PR #2578, opened 2024-11-07
SNOW-1791084: Cache metadata on SelectStatement.from_._snowflake_plan

sfc-gh-jdu:

   

   Fixes SNOW-1791084

   Another optimization we can do further by caching metadata on the child of SelectStatement

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
@@ -161,9 +161,10 @@ def infer_metadata(
                 quoted_identifiers = [
                     c.name for c in source_plan._column_states.projection
                 ]
-            # When source_plan has the same metadata as its child plan, we can use it directly
+            # When source_plan doesn't have a projection, it's a simple `SELECT * from ...`,
+            # which means source_plan has the same metadata as its child plan, we can use it directly
             if (
-                is_select_statement_child_metadata_same(source_plan)
+                source_plan.projection is None
                 and source_plan.from_._snowflake_plan is not None
             ):
                 # only set attributes and quoted_identifiers if they are not set in previous step
@@ -190,7 +191,7 @@ def infer_metadata(
     return PlanMetadata(attributes=attributes, quoted_identifiers=quoted_identifiers)
 
 
-def cache_metadata_on_select_statement(
+def cache_metadata_if_select_statement(
     source_plan: Optional[LogicalPlan], metadata: PlanMetadata
 ) -> None:
     """
@@ -203,12 +204,13 @@ def cache_metadata_on_select_statement(
         and source_plan.analyzer.session.reduce_describe_query_enabled
     ):
         source_plan._attributes = metadata.attributes
-        # When source_plan has the same metadata as its child plan,
+        # When source_plan doesn't have a projection, it's a simple `SELECT * from ...`,
+        # which means source_plan has the same metadata as its child plan,
         # we should cache it on the child plan too.
         # This is necessary SelectStatement.select() will need the column states of the child plan
         # (check the implementation of derive_column_states_from_subquery().
         if (
-            is_select_statement_child_metadata_same(source_plan)
+            source_plan.projection is None
             and source_plan.from_._snowflake_plan is not None
         ):
             source_plan.from_._snowflake_plan._metadata = metadata
--- a/src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
@@ -190,7 +190,7 @@ def infer_metadata(
 
 
 def cache_metadata_on_select_statement(
-    source_plan: LogicalPlan, metadata: PlanMetadata
+    source_plan: Optional[LogicalPlan], metadata: PlanMetadata
 ) -> None:
     """
     Cache metadata on a SelectStatement source plan.
--- a/src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
@@ -74,9 +74,7 @@ def is_select_statement_child_metadata_same(
     When source_plan doesn't have a projection, it's a simple `SELECT * from ...`,
     which has the same metadata as it's child plan (source_plan.from_).
     """
-    return (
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -362,6 +362,11 @@ def attributes(self) -> List[Attribute]:
             self.source_plan, SelectStatement
         ):
             self.source_plan._attributes = attributes
+            if (
```
sfc-gh-helmeleegy, line 365: Looks good, but we'll need to add a comment with explanation here too.

sfc-gh-aalam: +1
At first glance the assignment seems backwards but the schema query for `SnowflakePlan` comes from the `SelectStatement` so once a describe query gets executed, the results are being propagated back down to the `SelectStatement`. Is there a possibility we need to propagate the metadata down to even more `SelectStatement`s?

sfc-gh-jdu: Yes, added.

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -182,3 +188,27 @@ def infer_metadata(
...
+        source_plan._attributes = metadata.attributes
+        # When source_plan has the same metadata as its child plan,
+        # we should cache it on the child plan too.
+        # This is necessary SelectStatement.select() will need the column states of the child plan
+        # (check the implementation of derive_column_states_from_subquery().
+        if (
+            is_select_statement_child_metadata_same(source_plan)
+            and source_plan.from_._snowflake_plan is not None
+        ):
+            source_plan.from_._snowflake_plan._metadata = metadata
```
sfc-gh-aalam, line 214: does this need to be propagated down recursively? `source_plan.from_.from_.snowflake_plan._metadata`?

sfc-gh-jdu: Theoretically yes, but we shouldn't need it because the duplicate query is triggered here https://github.com/snowflakedb/snowpark-python/blob/cf454e9813bcb494c23093f04b4668acc8bf538d/src/snowflake/snowpark/_internal/analyzer/select_statement.py#L719 because source_plan.from_ doesn't have a column states. But if the metadata is cached, then it won't go to next level

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-helmeleegy: How about Afroz's question. Can we think of cases where the propagation can go even further down?

sfc-gh-helmeleegy: Just noticed you addressed the question in the other thread. SGTM.

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-aalam: I guess not since we only come here when projection is None.

sfc-gh-aalam (approved):

sfc-gh-helmeleegy (approved):
LGTM, thanks Jianzhun!

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -182,3 +188,27 @@ def infer_metadata(
...
+def cache_metadata_on_select_statement(
+    source_plan: Optional[LogicalPlan], metadata: PlanMetadata
+) -> None:
+    """
+    Cache metadata on a SelectStatement source plan.
+    """
+    from snowflake.snowpark._internal.analyzer.select_statement import SelectStatement
+
+    if (
+        isinstance(source_plan, SelectStatement)
```
sfc-gh-yzou, line 202: if the source plan is None, will this throw an error? maybe just do a check of source_plan is None, then return

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -66,6 +67,16 @@ def infer_quoted_identifiers_from_expressions(
...
 
 
+def is_select_statement_child_metadata_same(
+    source_plan: "SelectStatement",
+) -> bool:
+    """
+    When source_plan doesn't have a projection, it's a simple `SELECT * from ...`,
+    which has the same metadata as it's child plan (source_plan.from_).
+    """
+    return source_plan.projection is None
```
sfc-gh-yzou, line 77: that seems a single line of check and only called at one place, do we really need a separate function for this?

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -182,3 +188,27 @@ def infer_metadata(
             quoted_identifiers = None
 
     return PlanMetadata(attributes=attributes, quoted_identifiers=quoted_identifiers)
+
+
+def cache_metadata_on_select_statement(
+    source_plan: Optional[LogicalPlan], metadata: PlanMetadata
```
sfc-gh-yzou, line 194: since this function is called  cache_metadata_on_select_statement, let's make the caller do a check and make sure it passes a SelectStatement node inside. 
Or call this function cache_metadata_if_select_statement

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-jdu: this is not necessary in python because isinstance(source_plan, SelectStatement) will return False if source_plan is None (because it is NoneType)

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-jdu: yeah let me remove this function

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-jdu: I'd prefer to call this function cache_metadata_if_select_statement and keep all logic here (more clear)

sfc-gh-yzou (approved):

PR #2585, opened 2024-11-07
[SNOW-1791241] Add a pd modular level flag to turn scoped read only temp table off

sfc-gh-yzou:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/utils.py
@@ -109,6 +109,7 @@ READ_ONLY_TABLE_SUFFIX = "READONLY"
 METADATA_ROW_POSITION_COLUMN = "METADATA$ROW_POSITION"
 # read only table is only supported for base table or base temporary table
 READ_ONLY_TABLE_SUPPORTED_TABLE_KINDS = ["LOCAL TEMPORARY", "BASE TABLE"]
+ENABLE_SCOPED_TEMP_READ_ONLY_TABLE = "enable_scoped_data_snapshot"
 UNDEFINED = "undefined"
 # number of digits used to generate a random suffix
 _NUM_SUFFIX_DIGITS = 4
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/modin/plugin/extensions/pd_extensions.py
```diff
@@ -23,11 +23,14 @@
 from snowflake.snowpark.modin.plugin.utils.warning_message import (
     materialization_warning,
 )
+from snowflake.snowpark.modin.plugin.extensions.configs import Configs
 
 register_pd_accessor("Index")(Index)
 register_pd_accessor("DatetimeIndex")(DatetimeIndex)
 register_pd_accessor("TimedeltaIndex")(TimedeltaIndex)
 
+register_pd_accessor("configs")(Configs)
```
sfc-gh-joshi, line 32: I'm unsure about adding a distinct `pd.configs` variable for settings like this. modin and native pandas already provide their own configuration mechanisms:
- [`modin.config`](https://modin.readthedocs.io/en/stable/flow/modin/config.html) variables control partitioning and other behavior in OSS modin; we currently still use a vendored copy of `modin/config.py`
- [`pd.set_option`/`pd.get_option`](https://pandas.pydata.org/docs/user_guide/options.html) controls display options and some other configuration for native pandas, which we sometimes use in Snowpark pandas

It's possible for us to add a Modin environment variable since we still have a vendored copy, but this might not be a solution that works in the long term. I don't know if it's possible for us to add new items to `pd.set_options`.

I would prefer renaming this new variable to something like `pd.snowflake_options` to make clear that it is specific to Snowpark pandas, and avoid confusing users into thinking it's a native pandas API.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -273,7 +273,9 @@ def _create_read_only_table(
     readonly_table_name = (
         f"{random_name_for_temp_object(TempObjectType.TABLE)}{READ_ONLY_TABLE_SUFFIX}"
     )
-    use_scoped_temp_table = session._use_scoped_temp_objects
+    use_scoped_temp_table = (
+        session._use_scoped_temp_objects and pd.configs.enable_scoped_snapshot
```
sfc-gh-joshi, line 277: Does this behavior make sense if `session._use_scoped_temp_objects` is False and `pd.configs.enable_scoped_snapshot` is True?

Would it be reasonable to instead have `pd.configs.enable_scoped_snapshot` just be an alias for `session._use_scoped_temp_objects`, so that if a user does `pd.configs.enable_scoped_snapshot = False`, it will delegate the set to `session._use_scoped_temp_objects` instead? I'm worried about having two different configuration variables that seem to address the same behavior.

src/snowflake/snowpark/modin/plugin/extensions/pd_extensions.py
sfc-gh-yzou: In fact, i can not do it like this. This will not work for native app. i need to do it in the snowflake way. I need to close this pr

PR #2587, opened 2024-11-08
[SNOW-1791241] Add a session parameter to disable scoped read only temp table

sfc-gh-yzou:

   

SNOW-1791241

scoped read only table doesn't work with dynamic pivot in notebook today, investigation needed. 
We added a flag to turns the scoped temp read only table off for snowpark padnas.

Follow up: https://snowflakecomputing.atlassian.net/browse/SNOW-1793002 enable the flag at server side for native app

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
   - Treat an empty `subset` (e.g. `[]`) as if it specified all columns instead of no columns.
   - Raise a `TypeError` for a scalar `subset` instead of filtering on just that column.
   - Raise a `ValueError` for a `subset` of type `pandas.Index` instead of filtering on the columns in the index.
+- Disable creation of scoped read only table to mitigate dynamic pivot unfound table issue in notebook env.
 
 #### Improvements
 - Improve np.where with scalar x value by eliminating unnecessary join and temp table creation.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@
   - Treat an empty `subset` (e.g. `[]`) as if it specified all columns instead of no columns.
   - Raise a `TypeError` for a scalar `subset` instead of filtering on just that column.
   - Raise a `ValueError` for a `subset` of type `pandas.Index` instead of filtering on the columns in the index.
-- Disable creation of scoped read only table to mitigate dynamic pivot unfound table issue in notebook env.
+- Disable creation of scoped read only table to mitigate Disable creation of scoped read only table to mitigate `TableNotFoundError` when using dynamic pivot in notebook environment.
 
 #### Improvements
 - Improve np.where with scalar x value by eliminating unnecessary join and temp table creation.
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -238,6 +238,10 @@ _PYTHON_SNOWPARK_LARGE_QUERY_BREAKDOWN_COMPLEXITY_LOWER_BOUND = (
 _PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION = (
     "PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION"
 )
+# Flag for controlling the usage of scoped temp read only table.
+_PYTHON_SNOWPARK_ENABLE_SCOPED_TEMP_READ_ONLY_TABLE = (
+    "PYTHON_SNOWPARK_ENABLE_SCOPED_TEMP_READ_ONLY_TABLE"
+)
 # The complexity score lower bound is set to match COMPILATION_MEMORY_LIMIT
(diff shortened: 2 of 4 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -541,6 +545,11 @@ def __init__(
                 _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, True
             )
         )
+        self._use_scoped_temp_read_only_table: bool = (
+            self._conn._get_client_side_session_parameter(
+                _PYTHON_SNOWPARK_ENABLE_SCOPED_TEMP_READ_ONLY_TABLE, False
```
sfc-gh-deliu, line 550: If it's for native app, I think it's easier to look up from `self._conn._conn._session_parameters['<param_name>']`, similar to what we did for `ENABLE_FIX_1375538`, WDYT @sfc-gh-sfan?

sfc-gh-helmeleegy (changes requested):
I think we should have a changelog entry for this.

src/snowflake/snowpark/session.py
sfc-gh-yzou: oh, i believe that function does look into the _conn._session_parameters, here is the implementation 
```
    def _get_client_side_session_parameter(self, name: str, default_value: Any) -> Any:
        """It doesn't go to Snowflake to retrieve the session parameter.
        Use this only when you know the Snowflake session parameter is sent to the client when a session/connection is created.
        """
        return (
            self._conn._session_parameters.get(name, default_value)
            if self._conn._session_parameters
            else default_value
        )
```

sfc-gh-yzou:
@sfc-gh-helmeleegy added a changelog as part of fix.

sfc-gh-jjiao (approved):

sfc-gh-helmeleegy:
Let's wait for the response of Rich Murnane to our question [here](https://snowflakecomputing.atlassian.net/browse/SNOW-1645085?focusedCommentId=5821610) - to be able to assess the impact - before we go ahead and merge this.

sfc-gh-yzou:
@sfc-gh-helmeleegy I don't think we need to wait for the native app response, our impact on notebook is much bigger than the native app at this moment, the native app user can stay with 1.24.0 if they would like, or turn on the flag if the user would like to.

sfc-gh-helmeleegy:
> @sfc-gh-helmeleegy I don't think we need to wait for the native app response, our impact on notebook is much bigger than the native app at this moment, the native app user can stay with 1.24.0 if they would like, or turn on the flag if the user would like to.

SGTM. I will approve. The native app response when it comes (hopefully soon) can help us decide if there are users we need to communicate with to turn off the flag.

sfc-gh-helmeleegy (approved):
LGTM.

src/snowflake/snowpark/session.py
sfc-gh-deliu: Cool! I think you need to define a server side parameter too, similar to `PYTHON_SNOWPARK_ENABLE_THREAD_SAFE_SESSION` above: https://sourcegraph.c2.us-central1.gcp-dev.app.snowflake.com/github.com/snowflakedb/snowflake/-/blob/GlobalServices/src/main/java/com/snowflake/core/parameters/SnowparkEcosystemParameters.java?L774:42

And default it to false.

sfc-gh-yzou: @sfc-gh-deliu i will follow up with you on this, i though you mentioned that i don't need a server side parameter, and you can set this up in the session automatically

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
   - Treat an empty `subset` (e.g. `[]`) as if it specified all columns instead of no columns.
   - Raise a `TypeError` for a scalar `subset` instead of filtering on just that column.
   - Raise a `ValueError` for a `subset` of type `pandas.Index` instead of filtering on the columns in the index.
+- Disable creation of scoped read only table to mitigate dynamic pivot unfound table issue in notebook env.
```
sfc-gh-aalam, line 56: I would suggest we describe the kind of error this would bubble earlier. For example 
```suggestion
- Disable creation of scoped read only table to mitigate `TableNotFoundError` when using dynamic pivot in notebook environment.
```

sfc-gh-aalam (approved):

CHANGELOG.md
sfc-gh-yzou: udpated

src/snowflake/snowpark/session.py
sfc-gh-yzou: @sfc-gh-deliu i double checked with @sfc-gh-sfan , we do not need a definition for this parameter at server side. but he suggest we have a parameter control for turning this on for native app. I think we can do that when doing the server side change

sfc-gh-deliu (approved):

PR #2588, opened 2024-11-08
SNOW-1760495 Fix remaining merge gate failures.

sfc-gh-lspiegelberg:
Fixes remaining tests for AST enabled merge gate.

Details:
- changes window IR to use Expr instead of int to support make_interval.
- moves checks for orderby in front.
- fixes package moduletype management for sproc.
- fixes check for create_map
- fixes bug in AST serialization for dataframewriter


sfc-gh-vbudati (approved):
Thanks Leonhard!

PR #2591, opened 2024-11-08
SNOW-1787415 Fix dataframe join with sort deterministic

sfc-gh-evandenberg:
SNOW-1787415 Fix dataframe join with sort deterministic

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -43,6 +43,7 @@ from snowflake.snowpark._internal.analyzer.binary_plan_node import (
     UsingJoin,
     create_join_type,
 )
+from snowflake.snowpark._internal.analyzer.analyzer_utils import unquote_if_quoted
 from snowflake.snowpark._internal.analyzer.expression import (
     Attribute,
     Expression,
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -206,6 +207,16 @@ def _generate_prefix(prefix: str) -> str:
     return f"{prefix}_{generate_random_alphanumeric(_NUM_PREFIX_DIGITS)}_"
 
 
+def _generate_deterministic_prefix(prefix, exclude_prefixes):
+    # Generate deterministic prefix while ensuring it doesn't exist in the exclude list.
```
sfc-gh-lspiegelberg, line 211: can we turn the comment here into a docstring, I think that prob. documents this function sufficiently.

src/snowflake/snowpark/dataframe.py
```diff
@@ -206,6 +207,16 @@ def _generate_prefix(prefix: str) -> str:
     return f"{prefix}_{generate_random_alphanumeric(_NUM_PREFIX_DIGITS)}_"
 
 
+def _generate_deterministic_prefix(prefix, exclude_prefixes):
```
sfc-gh-lspiegelberg, line 210: can we add type hints?

tests/integ/conftest.py
```diff
@@ -313,3 +313,11 @@ def temp_schema(connection, session, local_testing_mode) -> None:
             )
             yield temp_schema_name
             cursor.execute(f"DROP SCHEMA IF EXISTS {temp_schema_name}")
+
+
+@pytest.fixture(scope='function', autouse=True)
+def clear_session_ast_batch_on_validate_ast(session, validate_ast):
+    yield
+    if validate_ast:
+        # After each test, flush the AST batch so it does not pollute the next test validation.
```
sfc-gh-lspiegelberg, line 322: nit: Can move the comment into docstring at top

src/snowflake/snowpark/dataframe.py
```diff
@@ -267,6 +278,7 @@ def _disambiguate(
         for n in lhs_names
         if n in set(rhs_names) and n not in normalized_using_columns
     ]
+    all_names = [unquote_if_quoted(n) for n in lhs_names + rhs_names]
```
sfc-gh-lspiegelberg, line 282: this is basically a snowpark python fix, right? AST does not seem to be affected?

Love the additional test.

sfc-gh-oplaton:
If this patch doesn't have a hard dependency on AST logic, please consider merging into `main` instead.

sfc-gh-lspiegelberg (approved):
LGTM, similar to Ovidiu's comment on slack - we prob. could push the fix directly to snowpark python's main branch.

PR #2594, opened 2024-11-08
SNOW-1438001: Add support for list values in Series.str.len

sfc-gh-helmeleegy:

   

   Fixes SNOW-1438001

   Add support for list values in Series.str.len.


sfc-gh-joshi (approved):
One question, otherwise LGTM: would this work for other arbitrary collections (dict/set/tuples)?

sfc-gh-helmeleegy:
> One question, otherwise LGTM: would this work for other arbitrary collections (dict/set/tuples)?

No, those will need additional logic.

docs/source/modin/supported/series_str_supported.rst
```diff
@@ -76,7 +76,7 @@ the method in the left column.
 +-----------------------------+---------------------------------+----------------------------------------------------+
 | ``join``                    | N                               |                                                    |
 +-----------------------------+---------------------------------+----------------------------------------------------+
-| ``len``                     | Y                               |                                                    |
+| ``len``                     | P                               | Only string and list data values are supported.    |
```
sfc-gh-yzou, line 79: i think we want to be clear that only string and list columns are supported, because i can have a column with mixed string and list data, but i don't think it work correctly today

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -16550,11 +16551,22 @@ def str_len(self, **kwargs: Any) -> "SnowflakeQueryCompiler":
...
+        if isinstance(
+            self._modin_frame.quoted_identifier_to_snowflake_type([col]).get(col),
+            ArrayType,
+        ):
+            return SnowflakeQueryCompiler(
+                self._modin_frame.apply_snowpark_function_to_columns(
+                    lambda col: array_size(col)
+                )
+            )
+        else:
```
sfc-gh-yzou, line 16564: since you are doing the type check anyway, how about just do a check of string type and throw an unsupported error for other cases ? length is only working for string, right?

tests/integ/modin/series/test_str_accessor.py
```diff
@@ -449,6 +449,13 @@ def test_str_len():
     eval_snowpark_pandas_result(snow_ser, native_ser, lambda ser: ser.str.len())
 
 
+@sql_count_checker(query_count=1)
+def test_str_len_list():
```
sfc-gh-yzou, line 453: we can add the example df I sent to you here to  check that after casting to array, this function can work

docs/source/modin/supported/series_str_supported.rst
sfc-gh-helmeleegy: Added more clarification.

sfc-gh-yzou (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -16550,11 +16551,22 @@ def str_len(self, **kwargs: Any) -> "SnowflakeQueryCompiler":
         SnowflakeQueryCompiler representing result of the string operation.
         """
         # TODO SNOW-1438001: Handle dict, list, and tuple values for Series.str.len().
```
sfc-gh-nkumar, line 16553: nit: remove `list` from todo comment.

sfc-gh-nkumar (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: I added the NotImplementedError but then had to remove it again because I realized that we do support a special case of mixed types (strings + numerical values, where the numerical values are automatically coerced into strings by snowflake). These test cases would break if we only allow string or array column types.

PR #2603, opened 2024-11-11
SNOW-1787415 Fix dataframe join suite failures related to formatting of remapping left and right columns of join.

sfc-gh-evandenberg:
SNOW-1787415 Fix dataframe join suite failures related to formatting of remapping left and right columns of join.


sfc-gh-lspiegelberg (approved):
lgtm, the createmap is handled in another PR.

PR #2605, opened 2024-11-11
SNOW-1794356: [API Coverage] DataType

sfc-gh-yuwang:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -133,7 +133,7 @@ commands = pytest {env:SNOWFLAKE_PYTEST_VERBOSITY:} {env:SNOWFLAKE_PYTEST_COV_CM
 [testenv:coverage]
 description = [run locally after tests]: combine coverage data and create report
 deps = {[testenv]deps}
-       coverage
+       coverage!=7.6.5
 skip_install = True
 passenv =
     DIFF_AGAINST


tests/unit/test_types.py
```diff
@@ -1048,3 +1048,51 @@ def test_snow_type_to_dtype_str():
 
     with pytest.raises(TypeError, match="invalid DataType"):
         snow_type_to_dtype_str(None)
+
+
+def test_datatype():
+    tpe = DataType()
+    assert tpe.simple_string() == "data"
+    assert tpe.json() == '"data"'
+
```
sfc-gh-aalam, line 1057: missing `json`, `json_value` tests.

tests/unit/test_types.py
```diff
@@ -1048,3 +1048,51 @@ def test_snow_type_to_dtype_str():
...
+    assert tpe.simple_string() == "boolean"
+
+    tpe = ByteType()
+    assert tpe.simple_string() == "byte"
+
+    tpe = DateType()
+    assert tpe.simple_string() == "date"
+
+    tpe = DecimalType()
+    assert tpe.simple_string() == "decimal"
```
sfc-gh-aalam, line 1071: Missing arrayType and Maptype

```
>>> mp.simpleString()
'map<string,int>'
>>> mp.json()
'{"keyType":"string","type":"map","valueContainsNull":true,"valueType":"integer"}'
>>> mp.jsonValue()
{'type': 'map', 'keyType': 'string', 'valueType': 'integer', 'valueContainsNull': True}

>>> at
ArrayType(StringType(), True)
>>> at.simpleString()
'array<string>'
>>> at.json()
'{"containsNull":true,"elementType":"string","type":"array"}'
>>> at.jsonValue()
{'type': 'array', 'elementType': 'string', 'containsNull': True}
```

tests/unit/test_types.py
```diff
@@ -1048,3 +1048,51 @@ def test_snow_type_to_dtype_str():
...
+    assert tpe.simple_string() == "short"
+
+    tpe = StringType()
+    assert tpe.simple_string() == "string"
+
+    tpe = StructType()
+    assert tpe.simple_string() == "struct"
+
+    tpe = TimestampType()
+    assert tpe.simple_string() == "timestamp"
```
sfc-gh-aalam, line 1095: let's create them with NTZ, LTZ, TZ and expect

```
timestamp<ntz>
timestamp<ltz>
timestamp<tz>
```
respectively

sfc-gh-aalam (changes requested):

src/snowflake/snowpark/types.py
```diff
@@ -237,17 +269,30 @@ def __init__(
         key_type: Optional[DataType] = None,
         value_type: Optional[DataType] = None,
         structured: bool = False,
+        value_contains_null: bool = True,
     ) -> None:
         self.structured = structured
         self.key_type = key_type if key_type else StringType()
         self.value_type = value_type if value_type else StringType()
+        self.value_contains_null = value_contains_null
```
sfc-gh-aalam, line 277: what funcitonal value does it serve for pyspark. do we do the same thing?

tests/unit/test_types.py
```diff
@@ -1048,3 +1048,137 @@ def test_snow_type_to_dtype_str():
...
+    assert tpe.simple_string() == "timestamptz"
+    assert tpe.json() == '"timestamp"'
+    assert tpe.type_name() == "timestamp"
+    assert tpe.json_value() == "timestamp"
+
+    tpe = TimestampType(TimestampTimeZone.LTZ)
+    assert tpe.simple_string() == "timestampltz"
+    assert tpe.json() == '"timestamp"'
+    assert tpe.type_name() == "timestamp"
+    assert tpe.json_value() == "timestamp"
```
sfc-gh-aalam, line 1148:
```
>>> TimestampNTZType().json()
'"timestamp_ntz"'
>>> TimestampNTZType().jsonValue()
'timestamp_ntz'
```
this is pyspark behavior

tests/unit/test_types.py
```diff
@@ -1048,3 +1048,137 @@ def test_snow_type_to_dtype_str():
...
+    assert tpe.json_value() == "struct"
+
+    tpe = TimestampType()
+    assert tpe.simple_string() == "timestamp"
+    assert tpe.json() == '"timestamp"'
+    assert tpe.type_name() == "timestamp"
+    assert tpe.json_value() == "timestamp"
+
+    tpe = TimestampType(TimestampTimeZone.TZ)
+    assert tpe.simple_string() == "timestamptz"
```
sfc-gh-aalam, line 1139: can we either do
```
assert tpe.simple_string() == "timestamp<tz>"
```

or 
```
assert tpe.simple_string() == "timestamp_tz"
```

Option 2 will be better for compatibility

tests/unit/test_types.py
```diff
@@ -1048,3 +1048,137 @@ def test_snow_type_to_dtype_str():
...
+    assert tpe.type_name() == "short"
+    assert tpe.json_value() == "short"
+
+    tpe = StringType()
+    assert tpe.simple_string() == "string"
+    assert tpe.json() == '"string"'
+    assert tpe.type_name() == "string"
+    assert tpe.json_value() == "string"
+
+    tpe = StructType()
```
sfc-gh-aalam, line 1126:
```
>>> st = StructType([StructField("f1", StringType())])
>>> st.simpleString()
'struct<f1:string>'
>>> StructType().simpleString()
'struct<>'
```

sfc-gh-aalam (changes requested):

src/snowflake/snowpark/types.py
```diff
@@ -338,14 +395,15 @@ def _strip_unnecessary_quotes(string: str) -> str:
         return string[1:-1] if result else string
 
 
-class StructField:
+class StructField(DataType):
```
sfc-gh-aalam, line 398: lets remove it from this change and add it later with BCR changes

src/snowflake/snowpark/types.py
```diff
@@ -338,14 +395,15 @@ def _strip_unnecessary_quotes(string: str) -> str:
...
     """Represents the content of :class:`StructField`."""
 
     def __init__(
         self,
         column_identifier: Union[ColumnIdentifier, str],
         datatype: DataType,
         nullable: bool = True,
+        metadata: Optional[Dict[str, Any]] = None,
```
sfc-gh-aalam, line 406: same question for this. How is it used in pyspark?

tests/unit/test_types.py
```diff
@@ -1048,3 +1048,156 @@ def test_snow_type_to_dtype_str():
...
+
+    tpe = TimeType()
+    assert tpe.simple_string() == "time"
+    assert tpe.json() == '"time"'
+    assert tpe.type_name() == "time"
+    assert tpe.json_value() == "time"
+
+    tpe = ArrayType(IntegerType())
+    assert tpe.simple_string() == "array<int>"
+    assert tpe.json() == '{"elementType":"integer","type":"array"}'
```
sfc-gh-aalam: can we also add an example of `Array(Array(Integer()))`

tests/unit/test_types.py
```diff
@@ -1048,3 +1048,156 @@ def test_snow_type_to_dtype_str():
 
     with pytest.raises(TypeError, match="invalid DataType"):
         snow_type_to_dtype_str(None)
+
+
+def test_datatype():
+    tpe = DataType()
+    assert tpe.simple_string() == "data"
+    assert tpe.json() == '"data"'
+    assert tpe.type_name() == "data"
```
sfc-gh-aalam, line 1057: how about using pytest.mark.parametrize

src/snowflake/snowpark/types.py
sfc-gh-yuwang: There is no internal pyspark function use this variable, it is mainly used in user-defined function to record information in structfield

src/snowflake/snowpark/types.py
```diff
@@ -41,6 +42,23 @@ def __repr__(self):
     def is_primitive(self):
         return True
 
+    @classmethod
+    def type_name(cls) -> str:
+        return cls.__name__[:-4].lower()
+
+    def simple_string(self) -> str:
```
sfc-gh-jdu, line 49: why are these two methods not classmethod, and having different return types?

sfc-gh-jdu: Or you can do `simple_string = type_name` directly

src/snowflake/snowpark/types.py
```diff
@@ -41,6 +42,23 @@ def __repr__(self):
...
+        return self.type_name()
+
+    def json_value(self) -> Union[str, Dict[str, Any]]:
+        return self.type_name()
+
+    def json(self) -> str:
+        return json.dumps(self.json_value(), separators=(",", ":"), sort_keys=True)
+
+    typeName = type_name
```
sfc-gh-jdu, line 58: Does pyspark have typeName only, or also type_name? If typeName only, I don't think we need to add type_name

sfc-gh-yuwang: they only have typeName, but type_name is a python style naming that we should apply to our code right? We just want to provide compatibility with pyspark

src/snowflake/snowpark/types.py
sfc-gh-yuwang: I am following pyspark's practice, i think they both have str as a return type?

src/snowflake/snowpark/types.py
sfc-gh-jdu: Then I don't think we need to add type_name and other methods that pyspark doesn't have - we only need to add APIs that pyspark has. cc @sfc-gh-aalam

src/snowflake/snowpark/types.py
sfc-gh-jdu: yeah, but why `def json_value(self) -> Union[str, Dict[str, Any]]:` is different?

src/snowflake/snowpark/types.py
```diff
@@ -41,6 +42,23 @@ def __repr__(self):
...
+    simpleString = simple_string
+    jsonValue = json_value
```
sfc-gh-aalam, line 60: can we implement these using classmethod as well so we don't have to re-add aliases each time.

tests/unit/test_types.py
```diff
@@ -1048,3 +1048,193 @@ def test_snow_type_to_dtype_str():
...
+            "decimal(20,10)",
+            '"decimal(20,10)"',
+            "decimal",
+            "decimal(20,10)",
+        ),
+        (DoubleType(), "double", '"double"', "double", "double"),
+        (FloatType(), "float", '"float"', "float", "float"),
+        (IntegerType(), "int", '"integer"', "integer", "integer"),
+        (LongType(), "bigint", '"long"', "long", "long"),
+        (ShortType(), "smallint", '"short"', "short", "short"),
```
sfc-gh-aalam, line 1072: can you look into some other snowflake types as well and see what is appropriate implementation for them. For example, I think Vector type may need some thought. You can create a separate JIRA for them as well if that is more convenient.

src/snowflake/snowpark/types.py
sfc-gh-aalam: > a dict from string to simple type that can be toInternald to JSON automatically

looks like it is used for `toInternal`? can you check if we can perform anything similar. My guess would be no. 

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.types.StructField.html

src/snowflake/snowpark/types.py
sfc-gh-yuwang: I think there two reasons we cannot do that, correct me if I was wrong.
1. in pyspark, only type_name is classmethod, do we want this different behavior?
2. even if we make them classmethod, we still need to re-add aliases each time because the aliases will always point to functions in DataType(parent class) if we don't re-add them.

tests/unit/test_types.py
sfc-gh-yuwang: I will create another Jira and another PR to do it

src/snowflake/snowpark/types.py
sfc-gh-aalam: I think you can use something like this to call the class version of `simple_string`

https://github.com/snowflakedb/snowpark-python/blob/8b6090ab2e2fde1408afa38693b3d223970890d7/src/snowflake/snowpark/mock/_functions.py#L140-L145

tests/unit/test_types.py
sfc-gh-aalam: sg!

src/snowflake/snowpark/types.py
sfc-gh-aalam: snowpark way is to follow snake_case and we create alias for pyspark counterpart for better migration. This is what I think we should keep following along

https://github.com/snowflakedb/snowpark-python/blob/8b6090ab2e2fde1408afa38693b3d223970890d7/src/snowflake/snowpark/dataframe.py#L4317-L4345

src/snowflake/snowpark/types.py
sfc-gh-yuwang: I took a look at metadata in StructField, though it says can be convert to json in toInternal automaticaly, I didn't see pyspark's code use it anywhere, seems like it is more like a record of structfield that can be preserved during transformation , serialization, etc

src/snowflake/snowpark/types.py
sfc-gh-yuwang: discussed offline, we decide to keep what we have now since these function need variable from instance and cannot be made classmethod

sfc-gh-jdu (approved):

sfc-gh-aalam (approved):

PR #2620, opened 2024-11-13
SNOW-1787415 Fix dataframe ast to register udtf for replay execution using client created udtf in same client session

sfc-gh-evandenberg:
Add Udtf registration that reuses the client registered UDTF.  Doesn't need all the previous parameters as they were only required for original creation of udtf.

Fix Full AST Validation to use same session for replay execution of unparser AST, thereby allowing reuse of client defined UDTFs.

### DIFF
--- a/src/snowflake/snowpark/_internal/ast_utils.py
+++ b/src/snowflake/snowpark/_internal/ast_utils.py
@@ -1171,7 +1171,7 @@ def build_udtf(
     statement_params: Optional[Dict[str, str]] = None,
     is_permanent: bool = False,
     session: "snowflake.snowpark.session.Session" = None,
-    udtf_name: Optional[Union[str, Iterable[str]]] = None,
+    _registered_object_name: Optional[Union[str, Iterable[str]]] = None,
     **kwargs,
 ):
     """Helper function to encode UDTF parameters (used in both regular and mock UDFRegistration)."""
@@ -1184,7 +1184,7 @@ def build_udtf(
         ast.handler,
         handler,
         session._ast_batch if session is not None else None,
-        udtf_name,
+        _registered_object_name,
     )
 
     if output_schema is not None:
--- a/src/snowflake/snowpark/_internal/ast_utils.py
+++ b/src/snowflake/snowpark/_internal/ast_utils.py
@@ -953,7 +953,7 @@ def build_proto_from_callable(
     expr_builder: proto.SpCallable,
     func: Union[Callable, Tuple[str, str]],
     ast_batch: Optional[AstBatch] = None,
-    object_name: Optional[str] = None,
+    object_name: Optional[Union[str, Iterable[str]]] = None,
 ):
     """Registers a python callable (i.e., a function or lambda) to the AstBatch and encodes it as SpCallable protobuf."""
 
@@ -979,7 +979,7 @@ def build_proto_from_callable(
         expr_builder.name = func.__name__
 
     if object_name is not None:
-        expr_builder.object_name = object_name
+        build_sp_table_name(expr_builder.object_name, object_name)
 
 
 def build_udf(
@@ -1171,7 +1171,7 @@ def build_udtf(
     statement_params: Optional[Dict[str, str]] = None,
     is_permanent: bool = False,
     session: "snowflake.snowpark.session.Session" = None,
-    udtf_name: str = None,
+    udtf_name: Optional[Union[str, Iterable[str]]] = None,
     **kwargs,
 ):
     """Helper function to encode UDTF parameters (used in both regular and mock UDFRegistration)."""
--- a/src/snowflake/snowpark/_internal/ast_utils.py
+++ b/src/snowflake/snowpark/_internal/ast_utils.py
@@ -16,7 +16,7 @@ from functools import reduce
 from logging import getLogger
 from pathlib import Path
 from types import ModuleType
-from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 import dateutil
 from dateutil.tz import tzlocal
@@ -44,7 +44,7 @@ from snowflake.snowpark._internal.type_utils import (
     ColumnOrSqlExpr,
 )
 from snowflake.snowpark._internal.utils import str_to_enum
-from snowflake.snowpark.types import DataType, StructType
+from snowflake.snowpark.types import DataType
 
 # This flag causes an explicit error to be raised if any Snowpark object instance is missing an AST or field, when this
 # AST or field is required to populate the AST field of a different Snowpark object instance.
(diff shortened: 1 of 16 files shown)


tests/ast/ast_test_utils.py
```diff
@@ -320,17 +319,13 @@ def notify_full_ast_validation(query_record: QueryRecord, *args, **kwargs):
     # Ensure sql counter uses the original session and not the validation session.
     tests.integ.utils.sql_counter._active_session = session
```
sfc-gh-lspiegelberg, line 321: can we avoid using globals?

sfc-gh-evandenberg: I refactored it to be more encapsulated.  The thing is we have a lot of usages of SqlCounter() throughout the test code, and what happens is the full ast validation path will trigger additional queries which then get added to the count, so we need a way to suppress this from happening when executing the validation queries.  Rather than directly interact with globals use TLS state instead which is more friendly for parallel test execution as well.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -953,6 +953,7 @@ def build_proto_from_callable(
     expr_builder: proto.SpCallable,
     func: Union[Callable, Tuple[str, str]],
     ast_batch: Optional[AstBatch] = None,
+    object_name: Optional[str] = None,
```
sfc-gh-lspiegelberg, line 956: we may want to type it with the weird `Optional[Union[str, List[str]]]` that is used across snowpark to identify objects.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -977,6 +978,9 @@ def build_proto_from_callable(
         # Use the actual function name. Note: We do not support different scopes yet, need to be careful with this then.
         expr_builder.name = func.__name__
 
+    if object_name is not None:
+        expr_builder.object_name = object_name
```
sfc-gh-lspiegelberg, line 982: can use spTableName logic here (which distinguishes between flat and structured form), there should be a helper existing in ast_utils.py

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -1166,7 +1170,8 @@ def build_udtf(
     comment: Optional[str] = None,
     statement_params: Optional[Dict[str, str]] = None,
     is_permanent: bool = False,
-    session=None,
+    session: "snowflake.snowpark.session.Session" = None,
+    udtf_name: str = None,
```
sfc-gh-lspiegelberg, line 1174: type hint? Optional[str]? Prob. with spTableName change the type name for that should be propagated here.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1359,31 +1359,24 @@ def select(
         if not exprs:
             raise ValueError("The input of select() cannot be empty")
 
-        # AST.
-        stmt = _ast_stmt
-        ast = None
-
-        if _emit_ast and _ast_stmt is None:
```
sfc-gh-lspiegelberg, line 1366: why is this code moved below? Any reason? E.g., test failure?

src/snowflake/snowpark/functions.py
```diff
@@ -8758,7 +8758,7 @@ def udtf(
     else:
         udtf_registration_method = session.udtf.register
 
-    if handler is None:
+    if handler is None and "object_name" not in kwargs:
```
sfc-gh-lspiegelberg, line 8761: should this be `... and kwargs.get("object_name") is None` to handle the None case inclusively?

src/snowflake/snowpark/mock/_udtf.py
```diff
@@ -71,36 +71,18 @@ def _do_register_udtf(
...
+        if "object_name" in kwargs:
             stmt = self._session._ast_batch.assign()
             ast = with_src_position(stmt.expr.udtf, stmt)
+            ast_id = stmt.var_id.bitfield1
 
-            build_udtf(
-                ast,
+            return MockUserDefinedTableFunction(
                 handler,
-                output_schema=output_schema,
```
sfc-gh-lspiegelberg, line 84: I assume we do not need those args in the mock plan?

tests/ast/ast_test_utils.py
```diff
@@ -12,12 +12,19 @@
...
+from tests.integ.utils.sql_counter import (
+    enable_sql_counting,
+    suppress_sql_counting,
+)
 from snowflake.snowpark import Session
 from snowflake.snowpark._internal.ast_utils import base64_lines_to_request
 from snowflake.snowpark.query_history import AstListener, QueryRecord
 
 
+suppress_ast_listener_reentry = False
```
sfc-gh-lspiegelberg, line 24: nit: Uppercase for these variables?

src/snowflake/snowpark/dataframe.py
sfc-gh-evandenberg: Yes, IDs were generated out of order because the stmt is created first and then ID's associated with any select expressions (in particular indirect table fn apply case) are created second.  However, the unparser resolves variables such that it expects any referenced variables (for example the table fn apply case) to have previously been defined and fails otherwise.

src/snowflake/snowpark/mock/_udtf.py
sfc-gh-evandenberg: Yes, the use_ast path isn't relevant for local testing mock.  For the nop plan we only use these fields of the mock UDTF.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -977,6 +978,9 @@ def build_proto_from_callable(
         # Use the actual function name. Note: We do not support different scopes yet, need to be careful with this then.
         expr_builder.name = func.__name__
 
+    if object_name is not None:
+        build_sp_table_name(expr_builder.object_name, object_name)
+
```
sfc-gh-lspiegelberg, line 984: should we update udaf/sproc as well?

src/snowflake/snowpark/dataframe.py
sfc-gh-lspiegelberg: sgtm, maybe let's add a comment to not forget this train of thought? Purposefully most of the code has been written so far in the form
```
if _emit_ast:
    ....

<snowpark code>

return <snowpark object>
```
This was with an eye towards removing the `<snowpark code>` / bringing it server-side. For that process, it may be easier to have a comment here to explain how this needs to be done here :)

src/snowflake/snowpark/mock/_udtf.py
```diff
@@ -71,36 +71,18 @@ def _do_register_udtf(
         _emit_ast: bool = True,
         **kwargs,
     ) -> UserDefinedTableFunction:
-
-        # Capture original parameters.
-        ast = None
-        if _emit_ast:
+        if kwargs.get("object_name") is not None:
```
sfc-gh-lspiegelberg, line 74: what if `_emit_ast = False` and `kwargs.get(...) is not None`?

src/snowflake/snowpark/udtf.py
```diff
@@ -908,41 +908,23 @@ def _do_register_udtf(
...
         **kwargs,
     ) -> UserDefinedTableFunction:
-
-        check_output_schema_type(output_schema)
-        check_imports_type(imports, "udtf-level")
-
-        # Capture original parameters.
-        ast, ast_id = None, None
-        if _emit_ast:
+        if "object_name" in kwargs:
```
sfc-gh-lspiegelberg, line 911: nit: maybe let's use one syntax here? (the is not None or in syntax?) No preference

tests/ast/test_ast_driver.py
```diff
@@ -107,6 +107,12 @@ def indent_lines(source: str, n_indents: int = 0):
     return "\n".join(map(lambda line: indent * n_indents + line, source.split("\n")))
 
 
+def normalize_temp_names(s: str):
+    return re.sub(
```
sfc-gh-lspiegelberg, line 111: can we add a docstring to describe what this function does? `normalize` itself is not very descriptive...

src/snowflake/snowpark/_internal/ast_utils.py
sfc-gh-evandenberg: Yes, I was planning to do in another pr after this one merges.

src/snowflake/snowpark/dataframe.py
sfc-gh-evandenberg: Added

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -953,6 +953,7 @@ def build_proto_from_callable(
     expr_builder: proto.SpCallable,
     func: Union[Callable, Tuple[str, str]],
     ast_batch: Optional[AstBatch] = None,
```
sfc-gh-lspiegelberg, line 955: nit: Could do `_registered_object_name` here as well.

src/snowflake/snowpark/_internal/ast_utils.py
```diff
@@ -1166,7 +1170,8 @@ def build_udtf(
     comment: Optional[str] = None,
     statement_params: Optional[Dict[str, str]] = None,
     is_permanent: bool = False,
-    session=None,
+    session: "snowflake.snowpark.session.Session" = None,
+    udtf_name: Optional[Union[str, Iterable[str]]] = None,
```
sfc-gh-lspiegelberg, line 1174: can we rename to not clash with `name`? Simply doing `_registered_object_name` should do.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -100,6 +100,7 @@
     "_from_pandas_udf_function",
     "input_names",  # for pandas_udtf
     "max_batch_size",  # for pandas_udtf
+    "_registered_object_name",  # db object name if already registered
```
sfc-gh-lspiegelberg, line 103: nit: remove `if already registered`, replace with something like `object name within Snowflake (post-registration)` or so.

sfc-gh-lspiegelberg: If it's None I assume the object has not been registered, if not none then it should be a valid object name?

src/snowflake/snowpark/dataframe.py
```diff
@@ -1446,6 +1441,23 @@ def select(
                     "The input of select() must be Column, column name, TableFunctionCall, or a list of them"
                 )
 
+        # AST.
+        stmt = _ast_stmt
+        ast = None
+
+        # Note it's intentional the column expressions are AST serializerd earlier (ast_cols) to ensure any
```
sfc-gh-lspiegelberg, line 1448: typo: serializerd -> serialized

src/snowflake/snowpark/mock/_udtf.py
```diff
@@ -140,17 +124,50 @@ def _do_register_udtf(
...
+                strict=strict,
+                secure=secure,
+                external_access_integrations=external_access_integrations,
+                secrets=secrets,
+                immutable=immutable,
+                comment=comment,
+                statement_params=statement_params,
+                is_permanent=is_permanent,
+                session=self._session,
+                udtf_name=udtf_name,
```
sfc-gh-lspiegelberg, line 156: can we rename udtf_name?

sfc-gh-lspiegelberg (approved):
lgtm, some nits. Would advise to not use udtf_name.

PR #2628, opened 2024-11-14
[SNOW-1798141] Improve align operation for align are performed on row position column by avoid unnecessary filter

sfc-gh-yzou:

   

SNOW-1798141

The align operations is a very complicated operation when the alignment is not applied on the objects coming from the same dataframe, where extra steps are required to check whether columns are aligned etc. The extra checks requires generations of bunch of window function, which is inefficient.

However, when alignment is applied on the row positions columns, even if they are not coming from the same dataframe, we know they are unique, and some extra operations could be avoided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,7 +136,7 @@
 
 - Improve np.where with scalar x value by eliminating unnecessary join and temp table creation.
 - Improve get_dummies performance by flattening the pivot with join.
-- Improve align performance when align on row position column by removing unnecessary window functions.
+- Improve align performance when aligning on row position column by removing unnecessary window functions.
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,7 @@
 
 - Improve np.where with scalar x value by eliminating unnecessary join and temp table creation.
 - Improve get_dummies performance by flattening the pivot with join.
+- Improve align performance when align on row position column by removing unnecessary window functions.
 
 
 
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -1723,7 +1723,7 @@ class OrderedDataFrame:
         elif join_filter is not None:
             joined_ordered_frame = joined_ordered_frame.filter(join_filter)
 
-        joined_ordered_frame.sort(ordering_columns)
+        joined_ordered_frame = joined_ordered_frame.sort(ordering_columns)
 
         # call select to make sure only the result_projected_column_snowflake_quoted_identifiers are projected
         # in the join result
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -1531,6 +1531,7 @@ class OrderedDataFrame:
         right_count = coalesce(max_(right_row_pos).over() + 1, lit(0))
         eq_row_pos_count = sum_(iff(left_row_pos == right_row_pos, 1, 0)).over()
 
+        # align_on_row_position_column = False
         ordering_columns = joined_ordered_frame.ordering_columns
         if align_on_row_position_column:
             align_filter = None
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -1560,10 +1560,10 @@ class OrderedDataFrame:
             extra_columns_to_append.append(col_matching_expr)
 
             # Define the final ordering column.
-            # As we mentioned in docstring, when left_on_cols and right_on_cols matches, the left
-            # and right frame is merged row by row with the original order, and the row order of
+            # As we mentioned in docstring, when left_on_cols and right_on_cols match, the left
+            # and right frames are merged row by row with the original order, and the row order of
             # original frame is retained.
-            # However, when left_on_cols and right_on_cols doesn't match, we need to sort lexicographically
+            # However, when left_on_cols and right_on_cols don't match, we need to sort lexicographically
             # on the join keys for `outer` align, and preserve left order followed by right order for `left` align.
             # This means the ordering column changes based on the result of column matching situation. Due
             # to lazy evaluation, we do not know the column matching situation util the query is evaluated.
@@ -1571,7 +1571,7 @@ class OrderedDataFrame:
             # input frames have matching left_on_cols and right_on_cols, otherwise this will be set to constant
             # 1 (a dummy ordering column has no effect).
             # Note that this is only needed by `outer` methods because it needs to sort on join keys. For `left`,
-            # preserve the left order followed by right order can give the correct order for both matching case
+            # preserving the left order followed by right order can give the correct order for both matching case
             # and non-matching case.
             if how == "outer":
                 global_order_col_identifier = (
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -1443,9 +1443,15 @@ class OrderedDataFrame:
         left = self.ensure_row_position_column()
         right = right.ensure_row_position_column()
 
+        # whether the alignment is performed on the row position column of each dataframe.
+        # In other words, this indicates whether the alignment is applied on a unique column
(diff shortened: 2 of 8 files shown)


tests/integ/utils/sql_counter.py
```diff
@@ -388,6 +394,10 @@ def sql_count_checker(
             filter(lambda k: k[0].endswith("_count"), all_args.locals.items())
         )
     }
+    # also look into kwargs for count information
+    for (key, value) in kwargs.items():
+        if key.endswith("_count"):
+            count_kwargs.update({key: value})
```
sfc-gh-evandenberg, line 400: Why is this necessary?  I think can avoid this if add window_count=None to argument list of sql_count_checker.

sfc-gh-yzou: The problem is the way we treat those xxx_count = None, seems check if xxxx_count is 0, so if i add window_count=None in the argument, the tests with sql_counter_check annotation without window_count configured but have window function inside will fail.

It seems to me the one marked with explicit parameters like query_count are counter check required, but for the ones configured in kwargs are optional. At this moment, i do not want to enforce the check window function in the annotation yet.

sfc-gh-nkumar (approved):

tests/integ/modin/frame/test_align.py
```diff
@@ -17,33 +17,42 @@
 from tests.integ.utils.sql_counter import SqlCounter, sql_count_checker
 
 
+def eval_dataframe_align_result(
```
sfc-gh-lmukhopadhyay, line 20: Thanks for refactoring this @sfc-gh-yzou! Can this test modification and tests for `on_row_position_columns` be added to the align series tests as well (series/test_align.py), or refactor in another PR?

tests/integ/utils/sql_counter.py
sfc-gh-evandenberg: I think this can be fixed by adding window_count=None to sql_count_checker API and treating None count_value as an opt-out, which is the default if it is not specified.

So in sql_count_checker can do:
```    
    count_kwargs = {
        key: value
        for key, value in list(
            filter(lambda k: k[0].endswith("_count") and k[1] is not None, all_args.locals.items())
        )
    }
```

and then won't need to make this change.

I tried out and seems to work as expected if window_count not specified (or None) or window_count is not None in sql_count_checker decorator.

sfc-gh-yzou: So I am not sure if that is the change we want to make here, because the annotation today took bunch of _count parameter here https://github.com/snowflakedb/snowpark-python/blob/main/tests/integ/utils/sql_counter.py#L375. And when expected_count = None, we check 0.  In other words, for all annotation check, if there is an explicit declaration of _count, the count check will be enforced and 0 will be used if None. 
If i filter out all _count parameter now, that basically means for the existing test, we will no longer check the parameter anymore, i would prefer not do this to keep out check level.

tests/integ/modin/frame/test_align.py
sfc-gh-yzou: added a refactor for series also.

sfc-gh-lmukhopadhyay (approved):
LGTM!

CHANGELOG.md
```diff
@@ -121,6 +121,7 @@
 
 - Improve np.where with scalar x value by eliminating unnecessary join and temp table creation.
 - Improve get_dummies performance by flattening the pivot with join.
+- Improve align performance when align on row position column by removing unnecessary window functions.
```
sfc-gh-helmeleegy, line 124:
```suggestion
- Improve align performance when aligning on row position column by removing unnecessary window functions.
```

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1521,55 +1538,62 @@ def align(
...
-                OrderingColumn(global_order_col_identifier)
-            ] + ordering_columns
+            col_matching_expr = (
+                (left_count == right_count) & (left_count == eq_row_pos_count)
+            ).as_(col_matching_identifier)
+            col_matching_column = Column(col_matching_identifier)
+            extra_columns_to_append.append(col_matching_expr)
+
+            # Define the final ordering column.
+            # As we mentioned in docstring, when left_on_cols and right_on_cols matches, the left
```
sfc-gh-helmeleegy, line 1563:
```suggestion
            # As we mentioned in docstring, when left_on_cols and right_on_cols matche, the left
```

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1521,55 +1538,62 @@ def align(
...
+            # and right frame is merged row by row with the original order, and the row order of
```
sfc-gh-helmeleegy, line 1564:
```suggestion
            # and right frames are merged row by row with the original order, and the row order of
```

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1521,55 +1538,62 @@ def align(
...
+            # original frame is retained.
+            # However, when left_on_cols and right_on_cols doesn't match, we need to sort lexicographically
```
sfc-gh-helmeleegy, line 1566:
```suggestion
            # However, when left_on_cols and right_on_cols don't match, we need to sort lexicographically
```

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1521,55 +1538,62 @@ def align(
...
+            # on the join keys for `outer` align, and preserve left order followed by right order for `left` align.
+            # This means the ordering column changes based on the result of column matching situation. Due
+            # to lazy evaluation, we do not know the column matching situation util the query is evaluated.
+            # In order to achieve this, we add a column 'ordering_col' which is set to left row position if
+            # input frames have matching left_on_cols and right_on_cols, otherwise this will be set to constant
+            # 1 (a dummy ordering column has no effect).
+            # Note that this is only needed by `outer` methods because it needs to sort on join keys. For `left`,
+            # preserve the left order followed by right order can give the correct order for both matching case
```
sfc-gh-helmeleegy, line 1574:
```suggestion
            # preserving the left order followed by right order can give the correct order for both matching case
```

tests/integ/modin/series/test_align.py
```diff
@@ -94,19 +71,30 @@ def test_align_basic_series_reorder_index(join, axis):
...
-    )
-    ser = pd.Series(native_ser)
-    other_ser = pd.Series(native_other_ser)
-    left, right = ser.align(other_ser, join=join, axis=axis)
-    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(left, native_left)
-    assert_snowpark_pandas_equals_to_pandas_without_dtypecheck(right, native_right)
+    eval_series_align_result(native_ser, native_other_ser, join=join, axis=axis)
+
+
+@pytest.mark.parametrize("join", ["outer", "inner", "left", "right"])
```
sfc-gh-helmeleegy, line 77: I'm not clear if we support `right` or not. The docstring in `ordered_dataframe.py` does not mention it.

tests/integ/utils/sql_counter.py
sfc-gh-yzou: added more comments

tests/integ/modin/series/test_align.py
sfc-gh-yzou: that is for align api, not the ordered dataframe, it seems supported right now. i am also not very sure how it is supported @sfc-gh-lmukhopadhyay

CHANGELOG.md
sfc-gh-yzou: updated

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-yzou: updated

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-yzou: updated

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-yzou: updated

sfc-gh-jdu (approved):

sfc-gh-helmeleegy (approved):

PR #2632, opened 2024-11-14
SNOW-1757892: Improve Dataframe.print_schema for structured types

sfc-gh-jrose:

   

   Fixes SNOW-1757892

   This PR adds support to DataFrame.print_schema for nested data types. Before this change a nested schema would be printed like so:
```
root
 |-- "MAP": MapType(StringType(), ArrayType(StructType([StructField('FIELD1', StringType(), nullable=True), StructField('FIELD2', LongType(), nullable=True)]))) (nullable = True)
```

After the change the nested schema will instead look like this:
```
root
 |-- "MAP": MapType (nullable = True)
 |   |-- key: StringType()
 |   |-- value: ArrayType
 |   |   |-- element: StructType
 |   |   |   |-- "FIELD1": StringType() (nullable = True)
 |   |   |   |-- "FIELD2": LongType() (nullable = True)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
       - `base_location`
       - `catalog_sync`
       - `storage_serialization_policy`
+- Added support for nested data types to `DataFrame.print_schema`
 
 #### Bug Fixes
 
--- a/docs/source/snowpark/dataframe.rst
+++ b/docs/source/snowpark/dataframe.rst
@@ -63,6 +63,8 @@ DataFrame
     DataFrame.orderBy
     DataFrame.order_by
     DataFrame.pivot
+    DataFrame.print_schema
+    DataDrame.printSchema
     DataFrame.randomSplit
     DataFrame.random_split
     DataFrame.rename
--- a/docs/source/snowpark/dataframe.rst
+++ b/docs/source/snowpark/dataframe.rst
@@ -64,7 +64,7 @@ DataFrame
     DataFrame.order_by
     DataFrame.pivot
     DataFrame.print_schema
-    DataDrame.printSchema
+    DataFrame.printSchema
     DataFrame.randomSplit
     DataFrame.random_split
     DataFrame.rename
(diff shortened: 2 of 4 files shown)


sfc-gh-jdu (approved):

sfc-gh-aalam (approved):

sfc-gh-wshangguan:
Hi @sfc-gh-jrose , this definitely improves the behavior. But still some pointes to note
1. can we return the string instead of only printing it?
2. In my local test, seems like every time I try to pass in nested structure like `Array<Array<Integer>>`, the returned dataframe only gives me `Array<String>` type, which doesn't align with the example in this PR description. Am I missing anything?
3. Can we support accepting a parameter, which indicates the printed column name? Basically takes in new column names and print_schema with them.

sfc-gh-jrose:
@sfc-gh-wshangguan 
1. We could add a utility method for that. Nobody has submitted a feature request for that yet.
2. Structured type support has not been rolled out yet. The Snowflake api always sends Snowpark semi-structured data. What you're seeing with `Array<String>` is the api returning structured data as json encoded strings.
3. I believe this use-case would be supported by a statement like `df.select(<columns>).print_schema()`.

sfc-gh-wshangguan:
@sfc-gh-jrose 
Thanks for the explanation!
1. Yeah, this would be helpful for accelerated spark api project
2. Gotcha.
3. Right, that would be the common use case. What I was talking about is that for spark api project, we have a sort of column naming mapping between spark df and snowpark df. This means the column names in snowpark df print schema are inconsistent with spark df column name. If we can accept an optional parameter in print_schema, which replaces the default column names in print_schema, that can solve the problem.

sfc-gh-wshangguan:
One more request is that we would need the level parameter similar to what spark df has in https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.printSchema.html

PR #2640, opened 2024-11-15
SNOW-1803811: Allow mixed-case field names for struct type columns

sfc-gh-jrose:

   

   Fixes SNOW-1803811

   This PR addresses an issue with StructType columns where names for StructFields are always uppercased. Snowflake always uppercases column names, but fields that are internal to StructType columns do not have casing enforced. Ideally we would not treat StructFields as column objects, but separating out that responsibility would require a BCR.

This change gets around the BCR requirement by using a feature flag. When enabled it shouldn't be much of a change as it assumes that only the top level of StructFields in a struct object are columns. Any structfields that are nested deeper in the schema are instead treated a OBJECT field names. This hierarchy should only exist in structured columns.


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
       - `catalog_sync`
       - `storage_serialization_policy`
 - Added support for nested data types to `DataFrame.print_schema`
+- Added support for mixed case field names in struct type columns.
```
sfc-gh-aalam, line 32: does this only affected structured types or all `StructType`s? If it is the latter, then it would be a bcr, no?

sfc-gh-jrose: This only affects structured types because currently semi-structured objects don't become StructType columns, but instead get converted to MapType for some reason.

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -291,7 +291,7 @@ def convert_sp_to_sf_type(datatype: DataType) -> str:
     if isinstance(datatype, StructType):
         if datatype.structured:
             fields = ", ".join(
-                f"{field.name} {convert_sp_to_sf_type(field.datatype)}"
+                f"{field.raw_name} {convert_sp_to_sf_type(field.datatype)}"
```
sfc-gh-aling, line 294: I kinda have the same problem as Afroz's on BCR.. previous we have name upper cased but now we do not, this won't break users?

tests/integ/scala/test_datatype_suite.py
```diff
@@ -106,7 +106,7 @@ def _create_test_dataframe(s):
                     StructType(
                         [
                             StructField("A", StringType(16777216), nullable=True),
-                            StructField("B", DoubleType(), nullable=True),
+                            StructField('"b"', DoubleType(), nullable=True),
```
sfc-gh-aling, line 109: we we need extra double quote here?

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jrose: As far as I know nobody is using StructType columns yet due to them requiring structured types be enabled for their account.

src/snowflake/snowpark/types.py
```diff
@@ -407,6 +407,7 @@ class ColumnIdentifier:
     """Represents a column identifier."""
 
     def __init__(self, normalized_name: str) -> None:
+        self.raw_name = normalized_name
```
sfc-gh-aling, line 410: If this is internal usage only then I prefer marking them as private `_raw_name`

src/snowflake/snowpark/types.py
```diff
@@ -487,6 +488,10 @@ def name(self) -> str:
         """Returns the column name."""
         return self.column_identifier.name
 
+    @property
+    def raw_name(self) -> str:
```
sfc-gh-aling, line 492: same as the the public vs private api comment

sfc-gh-jrose:
@sfc-gh-aling @sfc-gh-aalam 
I've completely refactored the approach to better avoid bcr concerns. I've changed around the logic to more closely match the server-side. Now top level StructFields will be assumed to describe columns as they previously have. Anything nested inside of the schema will instead be considered an actual StructField and have the correct naming logic applied. This approach also gets around the annoying quotation marks that would be added to inner fields.

@sfc-gh-lspiegelberg 
I've added you as a reviewer because this change breaks a few AST test and I'm hoping you can explain the process for updating those with the new expectations.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
   - Added `save` method to `DataFrameWriter` to work in conjunction with `format`.
   - Added support to read keyword arguments to `options` method for `DataFrameReader` and `DataFrameWriter`.
 - Relaxed the cloudpickle dependency for Python 3.11 to simplify build requirements. However, for Python 3.11, `cloudpickle==2.2.1` remains the only supported version.
+- Added support for mixed case field names in struct type columns.
```
sfc-gh-aalam, line 47: reminder to move it up to 1.27.0

src/snowflake/snowpark/types.py
```diff
@@ -552,29 +564,41 @@ def __init__(
...
+        self.name = column_identifier
+        self.is_column = is_column
         self.datatype = datatype
         self.nullable = nullable
 
     @property
     def name(self) -> str:
-        """Returns the column name."""
-        return self.column_identifier.name
+        return self.column_identifier.name if self.is_column else self._name
```
sfc-gh-aalam, line 577: does `is_column=True` enforce type of `self.column_identifier` is `ColumnIdentifier`? can we add an assert for that in `__init__`

sfc-gh-aalam (approved):
Looks good. left 2 minor comments

src/snowflake/snowpark/types.py
sfc-gh-jrose: `__init__` uses the setter for `self.name` which ensures that `self.column_identifier` is always `ColumnIdentifier`.

src/snowflake/snowpark/context.py
```diff
@@ -21,6 +21,10 @@
 _should_continue_registration: Optional[Callable[..., bool]] = None
 
 
+# Global flag that determines if structured type semantics should be used
+_should_use_structured_type_semantics = False
```
sfc-gh-aling, line 25: who's responsible for setting this parameter, also do we need server side parameter for this?

sfc-gh-jrose: This should not be enabled until structured type support is rolled out. There will eventually be a server side parameter that we can track instead of this flag.

src/snowflake/snowpark/types.py
```diff
@@ -552,29 +569,46 @@ def __init__(
         column_identifier: Union[ColumnIdentifier, str],
         datatype: DataType,
         nullable: bool = True,
+        is_column: bool = True,
```
sfc-gh-aling, line 572: this parameter will be part of the public api, do we expect customer to set this?

sfc-gh-jrose: I do not expect anyone to set this. Can make it a private parameter.

src/snowflake/snowpark/context.py
sfc-gh-aling: cool, if this is not rolled out yet then I'm good with this module var

sfc-gh-aling (approved):

PR #2657, opened 2024-11-20
SNOW-1818205: Add support for pd.json_normalize

sfc-gh-helmeleegy:

   

   Fixes SNOW-1818205

   Add support for pd.json_normalize.


src/snowflake/snowpark/modin/plugin/extensions/io_overrides.py
```diff
@@ -146,7 +145,14 @@ def json_normalize(
 ) -> pd.DataFrame:  # noqa: PR01, RT01, D200
     # TODO(https://github.com/modin-project/modin/issues/7104):
     # modin needs to remove defaults to pandas at API layer
-    pass  # pragma: no cover
+    _pd_json_normalize_signature = {
+        val.name
+        for val in inspect.signature(native_pd.json_normalize).parameters.values()
+    }
+    _, _, _, f_locals = inspect.getargvalues(inspect.currentframe())
```
sfc-gh-nkrishna, line 152: Why use inspect instead of passing the arguments directly?

sfc-gh-nkrishna (approved):
Left one question, but approving

src/snowflake/snowpark/modin/plugin/extensions/io_overrides.py
sfc-gh-helmeleegy: I believe this should help us stay consistent with upstream Modin.

src/snowflake/snowpark/modin/plugin/docstrings/io.py
```diff
@@ -343,6 +343,125 @@ def read_xml():
...
+    Parameters
+    ----------
+    data : dict or list of dicts
+        Unserialized JSON objects.
+    record_path : str or list of str, default None
+        Path in each object to list of records. If not passed, data will be assumed to be an array of records.
+    meta : list of paths (str or list of str), default None
+        Fields to use as metadata for each record in resulting table.
+    meta_prefix : str, default None
+        If True, prefix records with dotted (?) path, e.g. foo.bar.field if meta is [‘foo’, ‘bar’].
```
sfc-gh-rdurrani, line 356: Do we know why the question marks are here? seems like an artifact or typo in the original docs that we don't need to replicate.

src/snowflake/snowpark/modin/plugin/docstrings/io.py
```diff
@@ -343,6 +343,125 @@ def read_xml():
...
+        - ‘ignore’ : will ignore KeyError if keys listed in meta are not always present.
+        - ‘raise’ : will raise KeyError if keys listed in meta are not always present.
+    sep : str, default ‘.’
+        Nested records will generate names separated by sep. e.g., for sep=’.’, {‘foo’: {‘bar’: 0}} -> foo.bar.
+    max_level : int, default None
+        Max number of levels(depth of dict) to normalize. if None, normalizes all levels.
+
+    Returns
+    -------
+    frame : DataFrame
```
sfc-gh-rdurrani, line 370:
```suggestion
    A pandas on Snowflake DataFrame.
```

src/snowflake/snowpark/modin/plugin/extensions/io_overrides.py
```diff
@@ -146,7 +145,14 @@ def json_normalize(
 ) -> pd.DataFrame:  # noqa: PR01, RT01, D200
```
sfc-gh-rdurrani, line 145: do we need to remove the noqa, since we now support this and want it type checked?

sfc-gh-rdurrani (commented):
overall lgtm, but I left a couple of comments, and have a few questions:
1. Since we're defaulting to native pandas, should we provide some warning to the user or something, that the data will be processed serially in memory, and may be slow for large datasets?
2. Do we want to consider adding a distributed/threaded approach later on? (e.g. process batches of data records in parallel by loading them into tables (using https://docs.snowflake.com/en/user-guide/tutorials/script-data-load-transform-json), and then joining those tables?)

src/snowflake/snowpark/modin/plugin/docstrings/io.py
sfc-gh-joshi: we mostly use either "DataFrame" or "Snowpark pandas DataFrame" in other docstrings, so we should pick one of those for consistency

src/snowflake/snowpark/modin/plugin/extensions/io_overrides.py
sfc-gh-joshi: these #noqa marks are actually for lack of docstrings (PR01 is parameters, RT01 is return value, I forget what D200 is), which I think we're OK with because we define it elsewhere

src/snowflake/snowpark/modin/plugin/io/snow_io.py
```diff
@@ -177,6 +177,13 @@ def from_pandas(cls, df: pandas.DataFrame):
         """
         return cls.query_compiler_cls.from_pandas(df, pandas.DataFrame)
 
+    @classmethod
+    def json_normalize(cls, **kwargs):  # noqa: PR01
+        """
+        Normalize semi-structured JSON data into a query compiler representing a flat table.
+        """
+        return cls.from_pandas(pandas.json_normalize(**kwargs))
```
sfc-gh-joshi, line 185: should we mention in docs that we default to pandas for this? it's a bit of an edge case because the argument data is an in-memory object to begin with

sfc-gh-joshi (approved):

src/snowflake/snowpark/modin/plugin/docstrings/io.py
sfc-gh-helmeleegy: Updated.

src/snowflake/snowpark/modin/plugin/docstrings/io.py
sfc-gh-helmeleegy: Agree with Jonathan. We generally don't use "pandas on Snowflake" in the codebase.

src/snowflake/snowpark/modin/plugin/io/snow_io.py
sfc-gh-helmeleegy: I don't think we should. As you said, the input data is already in-memory. So the fact that we're using native pandas internally is an implementation detail that the end user need not worry about.

sfc-gh-helmeleegy:
> overall lgtm, but I left a couple of comments, and have a few questions:
> 
> 1. Since we're defaulting to native pandas, should we provide some warning to the user or something, that the data will be processed serially in memory, and may be slow for large datasets?
> 2. Do we want to consider adding a distributed/threaded approach later on? (e.g. process batches of data records in parallel by loading them into tables (using https://docs.snowflake.com/en/user-guide/tutorials/script-data-load-transform-json), and then joining those tables?)

As mentioned in another thread, the input data in this case is already in-memory.

PR #2659, opened 2024-11-20
SNOW-1748140: Modify schema_expression to be structured type aware.

sfc-gh-jrose:

   

   Fixes SNOW-1748140

This pull request includes several changes to improve the handling of structured data types in the `schema_expression` function to support structured types.

Schema expression has different case for when a field is nullable and when it is not. This change modifies the nullable case to skip the semi-structured branches of the if statement and instead use the default case so that `convert_sp_to_sf_type` does the work to specify the schema instead.

For non-nullable fields it relies on recursive calls to schema_expression in order to get good default values for the various nestsed data types and then casts a relevant data structured to the correct schema. Note that non-nullable columns do not have nullability respected in their child fields which is a limitation of the way we infer schemas using the dummy query approach. 

While writing some test cases for this I also found that the nullability of child fields was not set correctly when parsing metadata so I added that in type_utils as well.

I've opened two new bugs as a result of this change:

SNOW-1819531 - Large query breakdown does not appear to work correctly with the structured types.
SNOW-1819428 - When calling create_dataframe with a schema that contains a StructType column, the child fields do not have their nullability respected. I suspect this due to a similar limitation in how we generate schema strings.


src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -213,9 +213,33 @@ def schema_expression(data_type: DataType, is_nullable: bool) -> str:
         else:
             return "to_timestamp('2020-09-16 06:30:00')"
     if isinstance(data_type, ArrayType):
+        if data_type.structured:
+            element = schema_expression(data_type.element_type, is_nullable)
+            return f"to_array({element}) ::" + convert_sp_to_sf_type(data_type)
```
sfc-gh-aalam, line 218:
```suggestion
            return f"to_array({element}) ::{convert_sp_to_sf_type(data_type)}"
```
why not use this? also same below

tests/integ/scala/test_datatype_suite.py
```diff
@@ -970,3 +970,100 @@ def test_structured_type_print_schema(
...
+            " object_construct_keep_null('FIELD1', 'a' ::  STRING (16777216), 'FIELD2', 0 :: "
+            'DOUBLE) :: OBJECT(FIELD1 STRING(16777216), FIELD2 DOUBLE) AS "OBJ") UNION ( SELECT '
+            "object_construct_keep_null('a' ::  STRING (16777216), 0 :: DOUBLE) :: "
+            'MAP(STRING(16777216), DOUBLE) AS "MAP", to_array(0 :: DOUBLE) ::ARRAY(DOUBLE) AS "ARR", '
+            "object_construct_keep_null('FIELD1', 'a' ::  STRING (16777216), 'FIELD2', 0 :: "
+            'DOUBLE) :: OBJECT(FIELD1 STRING(16777216), FIELD2 DOUBLE) AS "OBJ")'
+        )
+    finally:
+        Utils.drop_table(structured_type_session, table_name)
+        Utils.drop_table(structured_type_session, non_null_table_name)
```
sfc-gh-aalam, line 1069: new line

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -213,9 +213,27 @@ def schema_expression(data_type: DataType, is_nullable: bool) -> str:
...
     if isinstance(data_type, ArrayType):
+        if data_type.structured:
+            element = schema_expression(data_type.element_type, is_nullable)
+            return f"to_array({element}) :: {convert_sp_to_sf_type(data_type)}"
         return "to_array(0)"
     if isinstance(data_type, MapType):
+        if data_type.structured:
+            key = schema_expression(data_type.key_type, is_nullable)
+            value = schema_expression(data_type.value_type, is_nullable)
+            return f"object_construct_keep_null({key}, {value}) :: {convert_sp_to_sf_type(data_type)}"
```
sfc-gh-jdu, line 224: should we determine whether keeping null values based on `is_nullable`?

tests/integ/scala/test_datatype_suite.py
```diff
@@ -970,3 +970,101 @@ def test_structured_type_print_schema(
...
+        non_null_table = structured_type_session.table(non_null_table_name)
+
+        assert table.schema == expected_schema
+        assert non_null_table.schema == expected_non_null_schema
+
+        # Dataframe.union forces a schema_expression call
+        assert table.union(table).schema == expected_schema
+        # Functions used in schema generation don't respect nested nullability so compare query string instead
+        non_null_union = non_null_table.union(non_null_table)
+        assert non_null_union._plan.schema_query == (
```
sfc-gh-jdu, line 1058: can you also create a test case for nested array and object? like `to_array(... to_array(...))`

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
sfc-gh-jrose: If we don't keep nulls and either the key or value gets evaluated to a `NULL :: ` type statement then that field would be dropped from the schema altogether. For this reason I think we always want nulls.

tests/integ/scala/test_datatype_suite.py
sfc-gh-jrose: Added

sfc-gh-jdu (approved):

PR #2665, opened 2024-11-22
SNOW-1787415 Fix UDF, UDAF, sproc functions with registered object_name for execution replay

sfc-gh-evandenberg:


src/snowflake/snowpark/_internal/proto/ast.proto
```diff
@@ -1501,6 +1502,11 @@ message SessionResetRequiredError {
   VarId var_id = 2;
 }
 
+// result.ir:23
```
sfc-gh-lspiegelberg, line 1505: This looks new - do not see an IR change, is this a new entity?

src/snowflake/snowpark/mock/_stored_procedure.py
```diff
@@ -262,37 +262,29 @@ def _do_register_sp(
...
-                session=self._session,
-                **kwargs,
+                strict=strict,
+                _ast=ast,
+                _ast_id=ast_id,
             )
+            self._registry[object_name] = sproc
+            return sproc
+
+        check_imports_type(imports, "stored-proc-level")
```
sfc-gh-lspiegelberg, line 287: The check should come first, there were a few tests that break otherwise.

src/snowflake/snowpark/mock/_udf.py
```diff
@@ -114,37 +114,27 @@ def _do_register_udf(
         _emit_ast: bool = True,
         **kwargs,
     ) -> UserDefinedFunction:
+        if kwargs.get("_registered_object_name") is not None:
```
sfc-gh-lspiegelberg, line 117: Can we define ast, ast_id first as variables? This may else lead to an unbound local error issue.

src/snowflake/snowpark/mock/_udf.py
```diff
@@ -153,7 +143,7 @@ def _do_register_udf(
                 raise_error=NotImplementedError,
             )
 
-        with self._lock:
+        with (self._lock):
```
sfc-gh-lspiegelberg, line 146: is the lock required for constructing the AST?

src/snowflake/snowpark/udaf.py
```diff
@@ -664,12 +668,45 @@ def _do_register_udaf(
...
+                kwargs["_registered_object_name"],
+                return_type,
+                input_types,
+                _ast=ast,
+                _ast_id=ast_id,
+            )
 
-        # AST. Capture original parameters, before any pre-processing.
-        ast = None
+        # get the udaf name, return and input types
```
sfc-gh-lspiegelberg, line 687: nit: udaf or UDAF?

src/snowflake/snowpark/udf.py
```diff
@@ -854,14 +855,42 @@ def _do_register_udf(
...
+                kwargs["_registered_object_name"],
+                _ast=ast,
+                _ast_id=ast_id,
+            )
 
         check_imports_type(imports, "udf-level")
 
-        # AST. Capture original parameters, before any pre-processing.
-        ast = None
+        # get the udf name, return and input types
```
sfc-gh-lspiegelberg, line 876: nit: `Get the UDF name, return and input types.`

Arthur is a fan of sentences that start with Uppercase letter and end with a dot.

tests/ast/ast_test_utils.py
```diff
@@ -68,6 +72,76 @@ def generate_error_trace_info(python_text, exception=None):
...
 
 
+def is_valid_ast_obj(ast):
+    return isinstance(ast, Iterable) or (
+        isinstance(ast, Message) and ast.ByteSize() > 0
+    )
+
+
+def get_dependent_var_ids(ast):
+    dependent_ids = set()
```
sfc-gh-lspiegelberg, line 82: can we add a doc string here? This function looks complex and could benefit from documentation.

tests/ast/ast_test_utils.py
```diff
@@ -68,6 +72,76 @@ def generate_error_trace_info(python_text, exception=None):
     return error_msg
 
 
+def is_valid_ast_obj(ast):
+    return isinstance(ast, Iterable) or (
```
sfc-gh-lspiegelberg, line 76: what does "valid" mean here? Can we add a docstring to state what we interpret as "valid"?

tests/ast/ast_test_utils.py
```diff
@@ -68,6 +72,76 @@ def generate_error_trace_info(python_text, exception=None):
...
+                if (
+                    isinstance(c, Iterable) and not isinstance(c, str) and len(c) > 0
+                ) or (isinstance(c, Message) and c.ByteSize() > 0):
+                    dependent_ids.update(get_dependent_var_ids(c))
+
+    return dependent_ids
+
+
+def create_minimal_full_ast_request(cur_request, prev_stmts, dependency_cache):
+    eval_stmts = []
```
sfc-gh-lspiegelberg, line 104: this may also benefit from a docstring. What is a minimal full ast request?

tests/ast/ast_test_utils.py
```diff
@@ -174,50 +248,20 @@ def notify_full_ast_validation_with_listener(
...
-    request = proto.Request()
-    request.client_language.python_language.version.major = 3
-    request.client_language.python_language.version.minor = 9
-    request.client_language.python_language.version.patch = 1
-    request.client_language.python_language.version.label = "final"
-
-    for next_stmt in prev_stmts + list(cur_request.body):
-        stmt = request.body.add()
-        stmt.CopyFrom(next_stmt)
+    request = create_minimal_full_ast_request(
```
sfc-gh-lspiegelberg, line 257: request -> minimal_request?

tests/ast/ast_test_utils.py
```diff
@@ -252,7 +296,7 @@ def notify_full_ast_validation_with_listener(
     )
 
     globals_dict = full_ast_validation_listener._globals
-    globals_dict["session"] = full_ast_validation_listener.session
+    skip_validation = False
```
sfc-gh-lspiegelberg, line 299: if this variable is False, can we remove it and simplify logic?

tests/ast/data/Session.call.test
```diff
@@ -13,7 +13,7 @@ df2 = session.call("my_sproc", 2, "one", dict(), False)
 
 ## EXPECTED UNPARSER OUTPUT
 
-my_sproc_sp = sproc(my_sproc, name="my_sproc", replace=True, comment="The parameters are useless.")
+my_sproc_sp = sproc("my_sproc", return_type=StringType(), input_types=[LongType(), StringType(), MapType(StringType(), StringType(), structured=False), BooleanType()], name="my_sproc", replace=True, comment="The parameters are useless.", _registered_object_name="\"MOCK_DATABASE\".\"MOCK_SCHEMA\".\"MY_SPROC\"")
```
sfc-gh-lspiegelberg, line 16: if we have two sprocs, would they both use the same registered object name? That doesn't sound right...

Prob. we'd need to:
1.) not print _registered_object_name here
2) Fix local testing to use unique names (prob. preferred)

tests/ast/data/col_literal.test
```diff
@@ -123,33 +123,33 @@ df = df.select(col("A") + float('-inf'))
 
 df = df.select(col("A") + float('nan'))
```
sfc-gh-lspiegelberg, line 125: just to verify: Is there `from decimal import Decimal` part of the ast runner setup?

sfc-gh-lspiegelberg:
Looks good! Went over and left some comments. There's also a good number of gates failing atm, we should definitely ensure they're all green before merging this back.

src/snowflake/snowpark/mock/_stored_procedure.py
sfc-gh-evandenberg: I don't think so.  The check_imports_type should only occur on the client side since that's where the sproc is actually created.  On the TCM / execution playback side we expect the sproc object already exists (the branch above) and therefore any related validations have occurred.  Note the branch above is used by the no-execution backend for testing.  Let me know if there's specific tests you think would fail as not seeing them.

src/snowflake/snowpark/udf.py
sfc-gh-evandenberg: Sure, this is the existing comment but I can clean it up.

tests/ast/ast_test_utils.py
sfc-gh-evandenberg: No longer used so removed.

tests/ast/ast_test_utils.py
sfc-gh-evandenberg: Updated the name and comments.

tests/ast/data/Session.call.test
sfc-gh-evandenberg: No, there is logic to generate the (registered) object name, it could be random or based on the "name" of the sproc.  So 2 different sprocs wouldn't normally end up with the same registered object name unless by random chance or they were given the same name.  Either way, the behavior would be same as if was entirely executed on the client.  Let me know if i misunderstood anything here.

PR #2670, opened 2024-11-22
SNOW-1819531: propagate referenced ctes to all root nodes

sfc-gh-aalam:

   

   Fixes SNOW-1819531

   i) https://github.com/snowflakedb/snowpark-python/pull/2659 re-enable test which is fixed by #2720 
  ii) fix a bug where `relaxed_pipeline_breaker` property for `SnowflakePlan` and `SelectSnowflakePlan` were not read correctly
  iii) updated plan plotter to color nodes that are not valid for replacement with red color.
  iv) updated plan plotter to print number of reference a node contains.


src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -195,6 +198,19 @@ def to_selectable(plan: LogicalPlan, query_generator: QueryGenerator) -> Selecta
         raise ValueError(f"parent type {type(parent)} not supported")
 
 
+def valid_for_replacement(node: LogicalPlan) -> bool:
+    if node._is_valid_for_replacement:
+        return True
+
+    if isinstance(node, SnowflakePlan) and node.source_plan is not None:
+        return node._is_valid_for_replacement
```
sfc-gh-yzou, line 206: i am kind of confused about why we need this function here?

sfc-gh-yzou: Can you also add comment for this function?

sfc-gh-aalam: while testing, I noticed that large query breakdown failed with error saying node is not valid for replacement for a `SnowflakePlan` whose child was a `WithQueryBlock`. At the time, I though this function would help us catch-up all such cases but on a second thought, it was a bug in repeated subquery elimination. I fixed the bug and removed this function now.

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -169,6 +169,7 @@ def do_resolve_with_resolved_children(
                 iceberg_config=logical_plan.iceberg_config,
                 table_exists=logical_plan.table_exists,
             )
+            resolved_plan.referenced_ctes = resolved_child.referenced_ctes
```
sfc-gh-yzou, line 172: it seems we missed the cte reference propagation in the plan builder for some case, we should double check the PlanBuilder code in snowflake_plan, instead of doing the fix here

src/snowflake/snowpark/_internal/compiler/repeated_subquery_elimination.py
```diff
@@ -171,6 +171,7 @@ def _update_parents(
                     with_block._is_valid_for_replacement = True
 
                     resolved_with_block = self._query_generator.resolve(with_block)
+                    resolved_with_block._is_valid_for_replacement = True
                     resolved_with_block_map[
```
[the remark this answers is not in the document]
sfc-gh-yzou: we set this flag here https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/compiler/query_generator.py#L225, do we have a case that didn't go into the do_resolve_with_resolved_children step, maybe we should overwrite at the do_resolve step also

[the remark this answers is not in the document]
sfc-gh-helmeleegy: What was the consequence of missing it? a missed optimization opportunity or a correctness issue?
Do we need to add more test coverage for the cte optimization?

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -477,6 +477,14 @@ def _is_relaxed_pipeline_breaker(self, node: LogicalPlan) -> bool:
         if isinstance(node, SelectStatement):
             return True
 
+        if isinstance(node, SnowflakePlan):
+            return node.source_plan is not None and self._is_relaxed_pipeline_breaker(
```
sfc-gh-helmeleegy, line 481: What's the difference between "relaxed pipeline breaker" and "pipeline breaker"?

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -477,6 +477,14 @@ def _is_relaxed_pipeline_breaker(self, node: LogicalPlan) -> bool:
...
+                node.source_plan
+            )
+
+        if isinstance(node, SelectSnowflakePlan):
+            return self._is_relaxed_pipeline_breaker(node.snowflake_plan)
```
sfc-gh-helmeleegy, line 486: It seems we need a test to cover this case.

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-yzou: After some thoughts, let's update the referenced_ctes definition to the cte referenced in the plan tree, and let's propagate  the ctes  reference for all nodes, in other words, let's deprecate this parameter https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L538, and make sure the referenced ctes is propagated correctly for all plan builder, please double check all places that directly create SnowflakePlan to make sure the referenced_ctes is propagated correctly.

sfc-gh-yzou: For the correction of code generation, we can use the way you have now, but please make sure we comment it out clearly about why things is done in such way

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -306,7 +309,17 @@ def get_snowflake_plan_queries(
 
     plan_queries = plan.queries
     post_action_queries = plan.post_actions
-    if len(plan.referenced_ctes) > 0:
+    if len(plan.referenced_ctes) > 0 and not isinstance(
+        plan.source_plan,
```
sfc-gh-yzou, line 313: adda a comment here about why are we doing this here

src/snowflake/snowpark/_internal/compiler/repeated_subquery_elimination.py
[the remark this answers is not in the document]
sfc-gh-yzou: the optimization may fail. @sfc-gh-aalam maybe it would be good to add a try catch for the optimization stage, and fallback to the original query if the optimization fails.

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
sfc-gh-aalam: pipeline breaker are all nodes like sort, pivot etc which are listed in the is_pipeline_breaker function.
relaxed pipeline breaker are those nodes which are not pipeline breakers but can be used to cut if no valid pipeline breaker is found. For now `SelectStatement` is the only relaxed pipeline breaker.

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
sfc-gh-aalam: agreed.

src/snowflake/snowpark/_internal/compiler/repeated_subquery_elimination.py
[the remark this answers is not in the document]
sfc-gh-aalam: https://snowflakecomputing.atlassian.net/browse/SNOW-1844094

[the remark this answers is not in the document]
sfc-gh-aalam: Screenshot 2024-12-03 at 2 43 49 PM

I'm investigating the source of this issue. But the problem is not with the `SnowflakePlan` but the `SelectSnowflakePlan` that is generated afterwards.

src/snowflake/snowpark/_internal/compiler/utils.py
```diff
@@ -121,7 +124,9 @@ def to_selectable(plan: LogicalPlan, query_generator: QueryGenerator) -> Selecta
             return plan
 
         snowflake_plan = query_generator.resolve(plan)
-        return SelectSnowflakePlan(snowflake_plan, analyzer=query_generator)
+        selectable = SelectSnowflakePlan(snowflake_plan, analyzer=query_generator)
+        selectable._is_valid_for_replacement = snowflake_plan._is_valid_for_replacement
+        return selectable
```
sfc-gh-aalam, line 129: before update:
Screenshot 2024-12-03 at 2 43 49 PM

after update:
Screenshot 2024-12-03 at 3 26 25 PM

sfc-gh-helmeleegy (commented):
Local testing tests are still failing.

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-yzou: If the cte reference is propagated correctly during plan builder, you shouldn't need to re-propagate here.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -565,9 +560,7 @@ def build(
             api_calls=select_child.api_calls,
             df_aliased_col_name_to_real_col_name=child.df_aliased_col_name_to_real_col_name,
             session=self.session,
-            referenced_ctes=child.referenced_ctes
```
sfc-gh-yzou, line 568: we probably also need to propagate the cte for the create_and_insert_plan https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L972 now need to have the cte propogated. 

For the subquery, we also need to set set the original cte in the newly resolved snowflake plan https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L326 which seems missing

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -941,7 +934,6 @@ def get_create_table_as_select_plan(child: SnowflakePlan, replace, error):
                 child,
                 source_plan,
                 is_ddl_on_temp_object=is_temp_table_type,
-                propagate_referenced_ctes=False,
```
sfc-gh-yzou, line 944: let's actually move that into a separate pr, and add more tests in test query generator with manually constructed plan tree to make sure the cte is propogated and used corretly

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -169,6 +182,7 @@ def do_resolve_with_resolved_children(
...
```
sfc-gh-yzou, line 185: i do not expect we need to do the propogation here if the plan is all resovled correctly

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: good catch.

sfc-gh-yzou (approved):

sfc-gh-jrose (approved):

sfc-gh-helmeleegy (commented):
LGTM.

PR #2678, opened 2024-11-25
SNOW-1787415 Fix udf, udaf, sproc functions with registered object name

sfc-gh-evandenberg:
SNOW-1787415 Fix udf, udaf, sproc functions with registered object name


sfc-gh-lspiegelberg (approved):
lgtm

PR #2682, opened 2024-11-26
SNOW-1818230: add concat_ws_ignore_nulls

sfc-gh-aalam:

   

   Fixes SNOW-1818230

   - Add `concat_ws_ignore_nulls` that ignores nulls and runs `concat_ws` operation.
   - Update `substring` function to make len argument optional to align with snowflake functionality.


src/snowflake/snowpark/functions.py
```diff
@@ -2628,18 +2628,28 @@ def substring(
...
+        ... )
+        >>> df.select(substring(col("S"), 2)).collect()
+        [Row(SUBSTRING("S", 2)='bc'), Row(SUBSTRING("S", 2)='ef')]
     """
     s = _to_col_if_str(str, "substring")
     p = pos if isinstance(pos, Column) else lit(pos)
-    length = len if isinstance(len, Column) else lit(len)
-    return builtin("substring")(s, p, length)
+    if len is not None:
+        length = len if isinstance(len, Column) else lit(len)
```
sfc-gh-jrose, line 2650: I've seen other places in function do this with `Column._to_expr(len)`.

src/snowflake/snowpark/functions.py
```diff
@@ -2935,6 +2945,39 @@ def concat_ws(*cols: ColumnOrName) -> Column:
...
+        |Hello,World                                       |
+        |                                                  |
+        |Hello                                             |
+        ----------------------------------------------------
+        <BLANKLINE>
+    """
+    columns = [_to_col_if_str(c, "concat_ws_ignore_nulls") for c in cols]
+    names = ",".join([c.get_name() for c in columns])
+    input_column_array = array_construct_compact(*columns)
+    reduced_result = builtin("reduce")(
```
sfc-gh-yixie, line 2973: reduce is to return a single value. Should it be filter?

sfc-gh-yixie: Another way is to use `concat_was(filter(...))` to remove the nulls first then concat.

sfc-gh-aalam: Return value of `filter` is an array. `concat_ws` does not take `ARRAY` as an input therefore we had to use reduce.

sfc-gh-jrose (approved):

src/snowflake/snowpark/functions.py
```diff
@@ -3030,18 +3030,27 @@ def substring(
...
+        >>> df.select(substring(col("S"), 2)).collect()
+        [Row(SUBSTRING("S", 2)='bc'), Row(SUBSTRING("S", 2)='ef')]
     """
     s = _to_col_if_str(str, "substring")
     p = pos if isinstance(pos, Column) else lit(pos, _emit_ast=_emit_ast)
-    length = len if isinstance(len, Column) else lit(len, _emit_ast=_emit_ast)
-    return builtin("substring", _emit_ast=_emit_ast)(s, p, length)
+    if len is None:
+        return builtin("substring", _emit_ast=_emit_ast)(s, p)
+    return builtin("substring", _emit_ast=_emit_ast)(s, p, Column._to_expr(len))
```
sfc-gh-lspiegelberg, line 3053: The `Column._to_expr` may cause issues when we add the AST here. We use instead the following and make it explicit:

```
len if isinstance(len, Column) else lit(len)
```

This should also have the correct type then, i.e. not using Expression objects to be passed to the callable created by builtin. Can we change to that?

Also, we're happy to help with adding the AST here. Can we file a ticket onto https://snowflakecomputing.atlassian.net/browse/SNOW-1489960 and add TODO in code with the ticket number to track the work?

If you'd like to give the AST directly a shot, happy to help out and walk you through the process - else also fine with JIRA ticket + TODO and our team will take care of it. Whichever works for you!

src/snowflake/snowpark/functions.py
```diff
@@ -3392,6 +3401,42 @@ def concat_ws(*cols: ColumnOrName, _emit_ast: bool = True) -> Column:
...
+        ----------------------------------------------------
+        |Hello,World                                       |
+        |                                                  |
+        |Hello                                             |
+        ----------------------------------------------------
+        <BLANKLINE>
+    """
+    columns = [_to_col_if_str(c, "concat_ws_ignore_nulls") for c in cols]
+    names = ",".join([c.get_name() for c in columns])
+    input_column_array = array_construct_compact(*columns, _emit_ast=_emit_ast)
```
sfc-gh-lspiegelberg, line 3431: you can do `_emit_ast=False` here. We'll capture the AST top-level in the body of this function. Basically the thumb rule should be whenever calling some other public function as part of the logic, use `_emit_ast=False`.

src/snowflake/snowpark/functions.py
```diff
@@ -3392,6 +3401,42 @@ def concat_ws(*cols: ColumnOrName, _emit_ast: bool = True) -> Column:
...
+    reduced_result = builtin("reduce", _emit_ast=_emit_ast)(
+        input_column_array, lit(""), sql_expr(f"(l, r) -> l || '{sep}' || r")
```
sfc-gh-lspiegelberg, line 3433: same here, prob. _emit_ast=False is correct.

src/snowflake/snowpark/functions.py
```diff
@@ -3392,6 +3401,42 @@ def concat_ws(*cols: ColumnOrName, _emit_ast: bool = True) -> Column:
...
+    )
+    return substring(reduced_result, 2, _emit_ast=_emit_ast).alias(
```
sfc-gh-lspiegelberg, line 3435: you can use _emit_ast=False for both, we'll need to explicitly create the AST for this and then assign to the columnar expression returned.

sfc-gh-lspiegelberg (commented):
left AST related comments

src/snowflake/snowpark/functions.py
sfc-gh-aalam: This is the same usage pattern as `to_date`. Shouldn't that also create problems?

sfc-gh-lspiegelberg: We fixed a lot of the `Column._to_expr(...)` cases. There may be some that fell through, merging long-standing feature branches is hard. But yes, we should not use `Column._to_expr(...)` anywhere. Better to be explicit. Happy to discuss/share more in detail.

src/snowflake/snowpark/functions.py
sfc-gh-lspiegelberg: we should not use `Column._to_expr(len)`. Happy to discuss why, it also clashes with the current AST design.

src/snowflake/snowpark/functions.py
```diff
@@ -3030,16 +3030,26 @@ def substring(
...
+
+    Example 2::
+        >>> df = session.create_dataframe(
+        ...     ["abc", "def"],
+        ...     schema=["S"],
+        ... )
+        >>> df.select(substring(col("S"), 2)).collect()
+        [Row(SUBSTRING("S", 2)='bc'), Row(SUBSTRING("S", 2)='ef')]
     """
     s = _to_col_if_str(str, "substring")
```
sfc-gh-lspiegelberg, line 3049: can we add `# TODO: SNOW-1831917 create ast`?

sfc-gh-lspiegelberg (approved):
LGTM (maybe add todo also for substring API, it likely will already work but not sure we have a test for the expanded version here)

src/snowflake/snowpark/functions.py
sfc-gh-aalam: this is not a new API. I am only updating the signature here

PR #2700, opened 2024-12-02
SNOW-1842459 Fix line no issues in ast expectation validation tests.

sfc-gh-evandenberg:
Fixes[SNOW-1842459](https://snowflakecomputing.atlassian.net/browse/SNOW-1842459)

The frame_info line information differs across python versions for chained operations, so to mitigate we only compare line no information for python 3.9 and not other versions.

### DIFF
--- a/src/snowflake/snowpark/_internal/ast/utils.py
+++ b/src/snowflake/snowpark/_internal/ast/utils.py
@@ -21,6 +21,7 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Tupl
 import dateutil
 from dateutil.tz import tzlocal
 from google.protobuf.text_format import MessageToString, Parse
+from google.protobuf.message import Message
 
 import snowflake.snowpark
 import snowflake.snowpark._internal.proto.generated.ast_pb2 as proto
@@ -1494,3 +1495,28 @@ def base64_lines_to_textproto(base64_str: str) -> str:
 def textproto_to_request(textproto_str: str) -> proto.Request:
     request = Parse(textproto_str, proto.Request())
     return request
+
+
+def clear_line_no_in_ast(ast: Any) -> None:
+    """Clear any 'src' information in the statement body."""
+    if isinstance(ast, Iterable) and not isinstance(ast, str):
+        for c in ast:
+            clear_line_no_in_ast(c)
+    elif hasattr(ast, "DESCRIPTOR"):
+        if hasattr(ast, "src"):
+            ast.ClearField("src")  # type: ignore[union-attr]
+
+        for f in ast.DESCRIPTOR.fields:
+            c = getattr(ast, f.name)
+            if (isinstance(c, Iterable) and not isinstance(c, str) and len(c) > 0) or (  # type: ignore[arg-type]
+                isinstance(c, Message) and c.ByteSize() > 0
+            ):
+                clear_line_no_in_ast(c)
+
+
+def clear_line_no_in_request(request: proto.Request) -> None:
+    """There are inconsistencies in the frame_info.line_no depending on the python version, this seems to be due to
+    fixes in determining better line_no info for chained python code, etc.  We will only record line_no info for
+    python 3.9 during test validation and exclude otherwise."""
+    for stmt in request.body:
+        clear_line_no_in_ast(stmt)
--- a/src/snowflake/snowpark/_internal/ast/utils.py
+++ b/src/snowflake/snowpark/_internal/ast/utils.py
@@ -1516,7 +1516,6 @@ def clear_line_no_in_ast(ast: Any) -> None:
 
 def clear_line_no_in_request(request: proto.Request) -> None:
     """There are inconsistencies in the frame_info.line_no depending on the python version, this seems to be due to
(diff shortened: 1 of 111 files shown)


src/snowflake/snowpark/_internal/ast/utils.py
```diff
@@ -1494,3 +1495,28 @@ def base64_lines_to_textproto(base64_str: str) -> str:
 def textproto_to_request(textproto_str: str) -> proto.Request:
     request = Parse(textproto_str, proto.Request())
     return request
+
+
+def clear_line_no_in_ast(ast: Any) -> None:
+    """Clear any 'src' information in the statement body."""
```
sfc-gh-lspiegelberg, line 1501: I am ok with a workaround for fixing this atm to unblock the release, however I actually think we may have a product ask to ensure we've have a proper src position story. Hence, mid-term we need to properly test that correct src positions are extracted. Are you planning on fixing the src code issue in another PR?

sfc-gh-evandenberg: The root cause is from the inspect library which behaves differently for chained code calls.

For example:
```
42:         df = df.select(
43:             when(col("a") == 1, "one")    ############### py3.10 -> start_line: 43
44:             .otherwise("other_one")
45:             .when(col("a") == 2, "two")
46:             .otherwise("other_two")      ################ py3.10 -> start_line: 46 (rather than 43)
47:             .when(col("a") == 3, "three")
48:             .otherwise("other_three")
49:             .when(col("a") == 4, "four")
50:             .otherwise("other_four")
51:             .when(col("a") == 5, "five")
52:             .otherwise("other_five")
53:         )
```

In python 3.9, inspect frame_info returns line 43 for all of the chained calls (lines 43-52).  Where as for python 3.10, inspect frame_info returns the actual correct line for each chained call (call on line 48 would return line 48.)

As you can see, whether this comes up depends on how the df source code was written.  We could also, for instance, change the test code to avoid hitting this issue.  But ultimately, we care about working with any customer source code, so I'd opt to not change the tests to subtle avoid this in some way.  

It seems we need to change to a different implementation (from inspect, have not looked into alternatives but may be there is something out there) or standardize on python 3.10+ (with updated inspect fixing the above line no bugs.).  Let me know if you have other ideas here.

sfc-gh-lspiegelberg: Hm, so to me it sounds like as-if Python 3.10 fixed some wrong behavior or improved the behavior. Considering all subsequent versions should have the Python3.10 behavior, I'd say given Python3.9 will at some point expire (EOL: 1st or 2nd quarter 2025, src: https://devguide.python.org/versions/), let's skip that part of the testing for Python3.9 and test only with 3.10+? Wdyt? This seems like the most pragmatic way to me. I do not really see a good way of fixing up 3.9, and it may not be worth it given 3.9 is already nearing its EOL.

I.e., let's call the clear logic here only for 3.9, update the tests/expectations using 3.10?

sfc-gh-evandenberg: Updated the expectation test file output based on python 3.11 based on offline conversation.

src/snowflake/snowpark/_internal/ast/utils.py
```diff
@@ -1494,3 +1495,28 @@ def base64_lines_to_textproto(base64_str: str) -> str:
...
+        for f in ast.DESCRIPTOR.fields:
+            c = getattr(ast, f.name)
+            if (isinstance(c, Iterable) and not isinstance(c, str) and len(c) > 0) or (  # type: ignore[arg-type]
+                isinstance(c, Message) and c.ByteSize() > 0
+            ):
+                clear_line_no_in_ast(c)
+
+
+def clear_line_no_in_request(request: proto.Request) -> None:
+    """There are inconsistencies in the frame_info.line_no depending on the python version, this seems to be due to
```
sfc-gh-lspiegelberg, line 1518: can we update the comment based on offline convo? I think it's also fine to remove any comment here mentioning the python version and simply describe what the function does. The whole intention on for which version we call this function should be commented where the function is called.

tests/ast/test_ast_driver.py
```diff
@@ -224,6 +226,12 @@ def test_ast(session, tables, test_case):
             ClearTempTables(actual_message)
             ClearTempTables(expected_message)
 
+            # If this is not python 3.9, then for the purposes of the expectation tests we will ignore the line_no
+            # information since it can be different based on various python bug fixes.
+            if sys.version_info.major != 3 or sys.version_info.minor != 11:
```
sfc-gh-lspiegelberg, line 231: we can assume sys.version_info.major to be 3 (you can add an assert if you feel strongly about safeguarding against 4).

Can we do
`if sys.version_info.minor <= 10:`?

Let's also update the comment.

sfc-gh-lspiegelberg (approved):
LGTM, few minor things left: please update comments & maybe also add to README/CI that we should use python 3.11 to update expectations. This is a change we need to communicate with the team.

sfc-gh-bmackas (approved):

PR #2711, opened 2024-12-04
SNOW-1672579 Encode `DataFrame.to_snowpark_pandas`

sfc-gh-vbudati:

   Fixes SNOW-1672579

- Added AST encoding for `DataFrame.to_snowpark_pandas`. 
- I had to add some Local testing functionality to help my expectation test to pass. Also needed to add some janky logic to create a temp read-only table in lieu of the table that is created by `to_snowpark_pandas`.
- I updated the script generating the relevant proto files to create them in the correct directory.


src/snowflake/snowpark/_internal/proto/update-from-local.sh
```diff
@@ -23,9 +23,9 @@ pushd $SRC_ROOT/Snowpark
 popd
```
sfc-gh-lspiegelberg, line 24: Thanks for fixing those

sfc-gh-lspiegelberg: easiest may be actually to set $SCRIPT_DIR=$SCRIPT_DIR/generated/

src/snowflake/snowpark/dataframe.py
```diff
@@ -1241,22 +1241,44 @@ def to_snowpark_pandas(
...
+            _emit_ast=False,
         )  # pragma: no cover
+        self._ast_id = ast_id  # reset the AST ID.
 
         snowpandas_df = pd.read_snowflake(
             name_or_query=temporary_table_name, index_col=index_col, columns=columns
         )  # pragma: no cover
 
+        if _emit_ast:
+            snowpandas_df._ast_id = stmt.var_id.bitfield1
```
sfc-gh-lspiegelberg, line 1280: shouldn't this be setting the snowpark dataframe within the ordered dataframe? I think we have no _ast_id for snowpark pandas yet.

tests/ast/data/Dataframe.to_snowpark_pandas.test
```diff
@@ -0,0 +1,186 @@
+## TEST CASE
+
+df = session.table(tables.table1)
+snowpark_pandas_df = df.to_snowpark_pandas()
+
+snowpark_pandas_df = df.to_snowpark_pandas(index_col="A")
+
+snowpark_pandas_df = df.to_snowpark_pandas(columns=["B", "A"])
```
sfc-gh-lspiegelberg, line 8: Can we add a test case where some pandas op is called after df.to_snowpark_pandas()? We want to make sure nothing breaks then. You can either extend the expectation test here, or add a new integ test under ast for that.

src/snowflake/snowpark/dataframe.py
sfc-gh-vbudati: done!

src/snowflake/snowpark/_internal/proto/update-from-local.sh
sfc-gh-vbudati: updated!

tests/ast/data/Dataframe.to_snowpark_pandas.test
sfc-gh-vbudati: done!

tests/ast/data/Dataframe.to_snowpark_pandas.test
```diff
@@ -0,0 +1,189 @@
...
+
+snowpark_pandas_df = df.to_snowpark_pandas(index_col=["A"], columns=["C", "B"])
+
+# Perform a pandas operation on the result to ensure nothing breaks.
```
sfc-gh-lspiegelberg, line 12: So I think this operation should then produce with AST enabled the internal Snowpark calls, or should it not? Just trying to understand the logic here. Perhaps you'll need to add `snowpark_pandas_df.groupby("A").to_numpy()` or so to trigger an eval... Yet our test logic should flush the last batch no matter what.

sfc-gh-vbudati: It won't produce an AST because I don't think we added any pandas AST logic

sfc-gh-vbudati: To avoid scope creep, I think it's a good idea to test the pandas AST logic in a different PR. This PR should only focus on whether the `to_snowpark_pandas` protobuf is generated correctly. Here's the ticket for it: https://snowflakecomputing.atlassian.net/browse/SNOW-1849281

Testing the pandas logic with local testing/nop testing is very janky so it will have to be in a separate test file.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1241,22 +1241,47 @@ def to_snowpark_pandas(
...
+                ast.columns.list.extend(
+                    columns if isinstance(columns, list) else [columns]
+                )
+
         # create a temporary table out of the current snowpark dataframe
         temporary_table_name = random_name_for_temp_object(
             TempObjectType.TABLE
         )  # pragma: no cover
+        ast_id = self._ast_id
+        self._ast_id = None  # set the AST ID to None to prevent AST emission.
```
sfc-gh-azwiegincew, line 1266: So setting `_emit_ast=False` wasn't enough? This seems like a bug.

src/snowflake/snowpark/dataframe.py
```diff
@@ -3925,7 +3950,7 @@ def write(self, _emit_ast: bool = True) -> DataFrameWriter:
         """
 
         # AST.
-        if _emit_ast and self._ast_id is not None:
+        if self._ast_id is not None:
```
sfc-gh-azwiegincew, line 3953: Oh, I see. This is a bit of a hack, but probably reasonable given the alternative.

tests/ast/test_ast_driver.py
```diff
@@ -94,6 +94,11 @@ def load_test_cases():
     Returns: a list of test cases.
     """
     test_files = DATA_DIR.glob("*.test")
+    if sys.version_info[1] < 9:
```
sfc-gh-azwiegincew, line 97: Should probably check that `[0]` == 3. Your grandchildren who will have to maintain this code will thank you ;).

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/dataframe.py
sfc-gh-vbudati: Yeah, since `write` is a property, I can't use the `_emit_ast` parameter.

PR #2713, opened 2024-12-04
[SNOW-1844465] Avoid creating a CTE out of simple select start on top of a select entity

sfc-gh-yzou:

   

   Fixes SNOW-1844465

The CTE optimization today will create a CTE out of a simple select * from TABLE, improve teh CTE by exclude such usage.

### DIFF
--- a/src/snowflake/snowpark/_internal/compiler/cte_utils.py
+++ b/src/snowflake/snowpark/_internal/compiler/cte_utils.py
@@ -58,7 +58,8 @@ def find_duplicate_subtrees(
 
     def is_simple_select_entity(node: "TreeNode") -> bool:
         """
-        Check if the current node is a simple select from Select
+        Check if the current node is a simple select on top of a SelectEntity, for example:
+        select * from TABLE. This check only works with selectable when sql simplifier is enabled.
         """
         if isinstance(node, SelectableEntity):
             return True
@@ -139,7 +140,6 @@ def find_duplicate_subtrees(
 
 def get_duplicated_node_complexity_distribution(
     duplicated_node_id_set: Set[str],
-    # id_complexity_map: Dict[str, int],
     id_node_map: Dict[str, List["TreeNode"]],
 ) -> List[int]:
     """
@@ -157,7 +157,6 @@ def get_duplicated_node_complexity_distribution(
     node_complexity_dist = [0] * 7
     for node_id in duplicated_node_id_set:
         complexity_score = get_complexity_score(id_node_map[node_id][0])
-        # complexity_score = id_complexity_map[node_id]
         repeated_count = len(id_node_map[node_id])
         if complexity_score <= 10000:
             node_complexity_dist[0] += repeated_count
--- a/src/snowflake/snowpark/_internal/compiler/cte_utils.py
+++ b/src/snowflake/snowpark/_internal/compiler/cte_utils.py
@@ -45,9 +45,40 @@ def find_duplicate_subtrees(
 
     This function is used to only include nodes that should be converted to CTEs.
     """
-    id_count_map = defaultdict(int)
+    id_node_map = defaultdict(list)
     id_parents_map = defaultdict(set)
-    id_complexity_map = defaultdict(int)
+
+    from snowflake.snowpark._internal.analyzer.select_statement import (
+        Selectable,
+        SelectStatement,
+        SelectableEntity,
+        SelectSnowflakePlan,
+    )
+    from snowflake.snowpark._internal.analyzer.snowflake_plan import SnowflakePlan
+
+    def is_simple_select_entity(node: "TreeNode") -> bool:
+        """
+        Check if the current node is a simple select from Select
+        """
+        if isinstance(node, SelectableEntity):
+            return True
+        if (
+            isinstance(node, SelectStatement)
+            and (node.projection is None)
+            and isinstance(node.from_, SelectableEntity)
+        ):
+            return True
+        if (
+            isinstance(node, SnowflakePlan)
+            and (node.source_plan is not None)
+            and isinstance(node.source_plan, (SnowflakePlan, Selectable))
+        ):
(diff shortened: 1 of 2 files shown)


sfc-gh-yzou:
pandas daily job started https://ci-dev-142.int.snowflakecomputing.com/job/SnowparkPythonSnowPandasDailyRegressRunner/1087/

snowpark daily job started https://github.com/snowflakedb/snowpark-python/actions/runs/12170949440

src/snowflake/snowpark/_internal/compiler/cte_utils.py
```diff
@@ -131,8 +156,9 @@ def get_duplicated_node_complexity_distribution(
     """
     node_complexity_dist = [0] * 7
     for node_id in duplicated_node_id_set:
-        complexity_score = id_complexity_map[node_id]
-        repeated_count = id_count_map[node_id]
+        complexity_score = get_complexity_score(id_node_map[node_id][0])
+        # complexity_score = id_complexity_map[node_id]
+        repeated_count = len(id_node_map[node_id])
```
sfc-gh-aalam, line 161:
```suggestion
        repeated_count = len(id_node_map[node_id])
```

src/snowflake/snowpark/_internal/compiler/cte_utils.py
```diff
@@ -114,8 +139,8 @@ def is_duplicate_subtree(encoded_node_id_with_query: str) -> bool:
 
 def get_duplicated_node_complexity_distribution(
     duplicated_node_id_set: Set[str],
-    id_complexity_map: Dict[str, int],
-    id_count_map: Dict[str, int],
+    # id_complexity_map: Dict[str, int],
+    id_node_map: Dict[str, List["TreeNode"]],
```
sfc-gh-aalam, line 143:
```suggestion
    id_node_map: Dict[str, List["TreeNode"]],
```

src/snowflake/snowpark/_internal/compiler/cte_utils.py
```diff
@@ -45,9 +45,40 @@ def find_duplicate_subtrees(
 
     This function is used to only include nodes that should be converted to CTEs.
     """
-    id_count_map = defaultdict(int)
+    id_node_map = defaultdict(list)
     id_parents_map = defaultdict(set)
-    id_complexity_map = defaultdict(int)
```
sfc-gh-aalam, line 50: I don't see the benefit of this approach over the previous approach. In this change, we are appending nodes to id_node_map list where we only access list[0] to calculate the complexity score.

sfc-gh-yzou: This is not needed anymore with the current change, the benefit is to avoid unnecessary data  structure, and it also cleaned up some unnecessary code  during the tree travel.

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-yzou: forgot to push the new change, removed

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-yzou: removed

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-aalam: I think appending nodes to a list only to get len of list later is not very efficient but I don't strongly disagree with your changes if it allows cleanup some code.

sfc-gh-aalam (approved):

tests/integ/test_cte.py
```diff
@@ -1005,6 +1006,25 @@ def test_time_series_aggregation_grouping(session, enable_sql_simplifier):
...
+    df = session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"])
+    df.write.save_as_table(table_name, table_type="temp")
+    df = session.table(table_name)
+    df_result = df.with_column("add_one", col("a") + 1).union(
+        df.with_column("add_two", col("a") + 2)
+    )
+    check_result(
+        session,
+        df_result,
+        expect_cte_optimized=False if session.sql_simplifier_enabled else True,
```
sfc-gh-helmeleegy, line 1020: Can we have test cases for when sql_simplifier_enabled is both True and False? The previous test uses a parameter for that.

tests/integ/test_cte.py
```diff
@@ -700,13 +700,14 @@ def test_table(session):
     check_result(
         session,
         df_result,
-        expect_cte_optimized=True,
+        expect_cte_optimized=False if session.sql_simplifier_enabled else True,
```
sfc-gh-helmeleegy, line 703: Same comment about using a parameter to test for both settings of sql_simplifier_enabled.

tests/integ/test_cte.py
sfc-gh-yzou: in our daily test, we actually runs all tests with sql simplifier disabled already here https://github.com/snowflakedb/snowpark-python/actions/runs/12169626312/job/33945760032, i don't think it is necessary to repeat the test here

tests/integ/test_cte.py
sfc-gh-yzou: Same as above, all those tests are run with both sql simplify enabled and disabled in our daily test today.

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-yzou: Oh, the main reason I need node here is because i need the node to help me getting the node property later, not really for clean up purpose, the clean up is a side effect after updating to the new data structure.

sfc-gh-helmeleegy (approved):
LGTM

PR #2724, opened 2024-12-06
SNOW-1830499: Incorporate arthurs suggested improvement

sfc-gh-batur:

   

   Fixes SNOW-1830499

### DIFF
--- a/tests/ast/decoder.py
+++ b/tests/ast/decoder.py
@@ -591,15 +591,11 @@ class Decoder:
 
             case "and":
                 # "and" is reserved keyword in python - so have to use getattr here.
-                left = self.decode_expr(getattr(expr, "and").lhs)
-                right = self.decode_expr(getattr(expr, "and").rhs)
-                return left & right
+                return self.binop(getattr(expr, "and"), lambda lhs, rhs: lhs & rhs)
 
             case "or":
                 # "or" is reserved keyword in python - so have to use getattr here.
-                left = self.decode_expr(getattr(expr, "or").lhs)
-                right = self.decode_expr(getattr(expr, "or").rhs)
-                return left | right
+                return self.binop(getattr(expr, "or"), lambda lhs, rhs: lhs | rhs)
 
             # DATAFRAME FUNCTIONS
             case "sp_create_dataframe":
--- a/tests/ast/decoder.py
+++ b/tests/ast/decoder.py
@@ -445,6 +445,9 @@ class Decoder:
         assert timezone.utcoffset(datetime.now()).total_seconds() == offset_seconds
         return timezone
 
+    def binop(self, lhs, rhs, fn):
+        return fn(self.decode_expr(lhs), self.decode_expr(rhs))
+
     def decode_expr(self, expr: proto.Expr) -> Any:
         match expr.WhichOneof("variant"):
             # COLUMN BINARY OPERATIONS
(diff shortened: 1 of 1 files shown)


tests/ast/decoder.py
```diff
@@ -445,6 +445,9 @@ def decode_timezone_expr(self, tz_expr: proto.PythonTimeZone) -> Any:
         assert timezone.utcoffset(datetime.now()).total_seconds() == offset_seconds
         return timezone
 
+    def binop(self, lhs, rhs, fn):
```
sfc-gh-azwiegincew, line 448: I think my original suggestion was cleaner – no need for the lhs and rhs being passed here, since we always dereference `.lhs` and `.rhs` (if not, then let's just not use this helper for those cases).

sfc-gh-batur: ok

tests/ast/decoder.py
```diff
@@ -554,71 +557,49 @@ def decode_expr(self, expr: proto.Expr) -> Any:
...
+                return self.binop(expr.pow, lambda lhs, rhs: lhs**rhs)
 
             case "and":
                 # "and" is reserved keyword in python - so have to use getattr here.
-                col_left = self.decode_expr(getattr(expr, "and").lhs)
-                col_right = self.decode_expr(getattr(expr, "and").rhs)
-                return col_left & col_right
+                left = self.decode_expr(getattr(expr, "and").lhs)
+                right = self.decode_expr(getattr(expr, "and").rhs)
+                return left & right
```
sfc-gh-azwiegincew, line 596: why not:
```
return self.binop(getattr(expr, "and"), lambda ...)
```

sfc-gh-azwiegincew: Also, I believe this is logical `and`, not `&`, which is represented by `bitand`

tests/ast/decoder.py
```diff
@@ -554,71 +557,49 @@ def decode_expr(self, expr: proto.Expr) -> Any:
...
-                col_left = self.decode_expr(expr.mod.lhs)
-                col_right = self.decode_expr(expr.mod.rhs)
-                return col_left % col_right
+                return self.binop(expr.mod, lambda lhs, rhs: lhs % rhs)
 
             case "pow":
-                col_left = self.decode_expr(expr.pow.lhs)
-                col_right = self.decode_expr(expr.pow.rhs)
-                return col_left**col_right
+                return self.binop(expr.pow, lambda lhs, rhs: lhs**rhs)
```
sfc-gh-azwiegincew, line 590: Spaces around the operator.

sfc-gh-batur: The black formatter actually removed it.

tests/ast/decoder.py
sfc-gh-batur: raise TypeError(
            "Cannot convert a Column object into bool: please use '&' for 'and', '|' for 'or', "
            "'~' for 'not' if you're building DataFrame filter expressions. For example, use df.filter((col1 > 1) & (col2 > 2)) instead of df.filter(col1 > 1 and col2 > 2)."
        )
E       TypeError: Cannot convert a Column object into bool: please use '&' for 'and', '|' for 'or', '~' for 'not' if you're building DataFrame filter expre

tests/ast/decoder.py
```diff
@@ -554,71 +557,45 @@ def decode_expr(self, expr: proto.Expr) -> Any:
...
 
             case "and":
                 # "and" is reserved keyword in python - so have to use getattr here.
-                col_left = self.decode_expr(getattr(expr, "and").lhs)
-                col_right = self.decode_expr(getattr(expr, "and").rhs)
-                return col_left & col_right
+                return self.binop(getattr(expr, "and"), lambda lhs, rhs: lhs & rhs)
```
sfc-gh-azwiegincew, line 594: & => and

sfc-gh-batur: raise TypeError(
            "Cannot convert a Column object into bool: please use '&' for 'and', '|' for 'or', "
            "'~' for 'not' if you're building DataFrame filter expressions. For example, use df.filter((col1 > 1) & (col2 > 2)) instead of df.filter(col1 > 1 and col2 > 2)."
        )
E       TypeError: Cannot convert a Column object into bool: please use '&' for 'and', '|' for 'or', '~' for 'not' if you're building DataFrame filter expressions. For example, use df.filter((col1 > 1) & (col2 > 2)) instead of df.filter(col1 > 1 and col2 > 2).

sfc-gh-batur: I think the formatting in the error message is messed up, but, that's the error that shows up if I use "and" instead of "&".

sfc-gh-batur: Same issue with | => or.

sfc-gh-azwiegincew (approved):

PR #2735, opened 2024-12-10
Feedback on public api

sfc-gh-aalam:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -846,8 +846,9 @@ def infer_ast_enabled_from_global_sessions(func: Callable) -> bool:  # pragma: n
                 f"Could not retrieve default session "
                 f"for function {func.__qualname__}, capturing AST by default."
             )
-            # session has not been created yet. To not lose information, always encode AST.
-            return True  # noqa: B012
+            # session has not been created yet, do not encode AST be default.
+            # TODO: flip this to True when we go GA.
+            return False  # noqa: B012
         else:
             return session.ast_enabled  # noqa: B012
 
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -868,6 +869,10 @@ def func_call_wrapper(*args, **kwargs):  # pragma: no cover
         if "_emit_ast" in func.__code__.co_varnames and "_emit_ast" not in kwargs:
             # No arguments, or single argument with function.
             if len(args) == 0 or (len(args) == 1 and isinstance(args[0], Callable)):
+                # this can be refactored to
+                #  - extract session
+                #  - check isinstance(session, snowflake.snowpark.Session)
+                #  - update kwargs["_emit_ast"] = value
```
sfc-gh-aalam, line 875: I noticed we are not checking `isinstance(session, Session)` so this would make it more consistent

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -846,8 +846,9 @@ def infer_ast_enabled_from_global_sessions(func: Callable) -> bool:  # pragma: n
                 f"Could not retrieve default session "
                 f"for function {func.__qualname__}, capturing AST by default."
             )
-            # session has not been created yet. To not lose information, always encode AST.
-            return True  # noqa: B012
+            # session has not been created yet, do not encode AST be default.
+            # TODO: flip this to True when we go GA.
+            return False  # noqa: B012
```
sfc-gh-oplaton, line 851: If we're moving away from hard-coded `True`, the value should come from an environmental variable, or some other suitable global configuration source. We don't want to be in a place where the feature is untestable while under development/opt-in mode.

sfc-gh-aalam: Why would it not be testable if user has a session and they have oped-in such that `session.ast_enabled` returns true.

This is the case when a valid session is not found, and we default to assuming that ast is enabled. I say, we default to disabled.

PR #2747, opened 2024-12-11
SNOW-1782778: Fix Function available memory exhausted

sfc-gh-yuwang:

   

   Fixes SNOW-1782778

   We hit this issue because when it is not a select statement, we do not have a limit on the query results, while regex is always not reliable enough to decide whether it is a select statement. Thus leading to accidental consume of huge queries.
This PR is meant to add a limit even if it is not select statement, we did this by consume it with iterator and return limited rows, which avoid this issue

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -491,7 +491,6 @@ class ServerConnection:
         async_post_actions: Optional[List[Query]] = None,
         **kwargs,
     ) -> Union[Dict[str, Any], AsyncJob]:
-        show_limit = kwargs.pop("show_limit", None)
         try:
             # Set SNOWPARK_SKIP_TXN_COMMIT_IN_DDL to True to avoid DDL commands to commit the open transaction
             if is_ddl_on_temp_object:
@@ -528,7 +527,6 @@ class ServerConnection:
                 results_cursor=results_cursor,
                 to_pandas=to_pandas,
                 to_iter=to_iter,
-                show_limit=show_limit,
             )
         else:
             return AsyncJob(
@@ -548,10 +546,7 @@ class ServerConnection:
         results_cursor: SnowflakeCursor,
         to_pandas: bool = False,
         to_iter: bool = False,
-        show_limit: int = None,
     ) -> Dict[str, Any]:
-        if show_limit is not None:
-            to_iter = True
         qid = results_cursor.sfqid
         if to_iter:
             new_cursor = results_cursor.connection.cursor()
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -580,6 +587,15 @@ def _to_data_or_iter(
...
                 iter(results_cursor) if to_iter else results_cursor.fetchall()
             )
+        # return limit number of rows
+        if show_limit is not None:
+            data_or_iter_result = []
+            for i, d in enumerate(data_or_iter):
+                if i >= show_limit:
+                    break
+                else:
+                    data_or_iter_result.append(d)
```
sfc-gh-aalam, line 597: I think we should update `result_set_to_rows` instead of making changes here. Can you take a look into that function and see if you also think that is a more appropriate place to make these changes.

I think we should get results using `to_iter=True` and parse results in `result_set_to_rows`

src/snowflake/snowpark/dataframe.py
```diff
@@ -4367,6 +4367,7 @@ def _show_string(
                 self.limit(n, _emit_ast=False)._plan, **kwargs
             )
         else:
+            kwargs["show_limit"] = n
             res, meta = self._session._conn.get_result_and_metadata(
                 self._plan, **kwargs
```
sfc-gh-aalam, line 4372:
```suggestion
            res, meta = self._session._conn.get_result_and_metadata(
                self._plan, show_limit=n, **kwargs
```
why not just do this?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -768,8 +770,9 @@ def get_result_set(
     def get_result_and_metadata(
         self, plan: SnowflakePlan, **kwargs
     ) -> Tuple[List[Row], List[Attribute]]:
-        result_set, result_meta = self.get_result_set(plan, **kwargs)
-        result = result_set_to_rows(result_set["data"])
+        show_limit = kwargs.pop("show_limit", None)
```
sfc-gh-aalam, line 773: I'm slightly inclined towards making the parameter part of the function signature and let kwargs be same as what users passed in

```
def get_result_and_metadata(
        self, plan: SnowflakePlan, limit: Optional[int] = None, **kwargs
    ) -> Tuple[List[Row], List[Attribute]]:
```

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -768,8 +770,9 @@ def get_result_set(
...
+        result_set, result_meta = self.get_result_set(plan, to_iter=True, **kwargs)
+        result = result_set_to_rows(result_set["data"], show_limit=show_limit)
         attributes = convert_result_meta_to_attribute(result_meta, self.max_string_size)
```
sfc-gh-aalam, line 776: how about naming the param `limit` to make it more general rather than `show_limit`.

sfc-gh-aalam (commented):
we can also make the update in `MockServerConnection` to maintain consistency.

sfc-gh-aalam:
don't forget to update PR description

PR #2748, opened 2024-12-11
New scripting environment for integration with the monorepo

sfc-gh-azwiegincew:
THIS IS DEFINITELY WIP!

I've screwed up my Cloud Workspace trying to make this work, so I haven't really tested scripts/copy-ast.sh.

I did test scripts/copy-remote-ast.sh whereby I run the script from my local machine. AST tests pass after this.

## BOILERPLATE


scripts/copy-ast.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Compute the source directory. We honor the MONOREPO_DIR environment variable if set,
+# and default to the standard Clound Workspace location otherwise.
```
sfc-gh-vbudati, line 4:
```suggestion
# and default to the standard Cloud Workspace location otherwise.
```
nit

sfc-gh-vbudati (approved):
Thanks Arthur!

src/snowflake/snowpark/_internal/proto/ast.proto
```diff
@@ -2853,4 +2862,4 @@ message UnaryOp {
     Neg neg = 1;
     Not not = 2;
   }
-}
+}
```
sfc-gh-azhan, line 2865: newline

src/snowflake/snowpark/_internal/proto/ast.proto
```diff
@@ -1,5 +1,5 @@
 // N.B. This file is generated by `ir-dsl-c`. DO NOT EDIT!
-// Generated from `{git@github.com:snowflakedb/snowflake.git}/Snowpark/ast`.
+// Generated from `scratch/bazel/36bac8ee52e5d1559627454b12aa1b28/execroot/Snowflake/Snowpark/ast`.
```
sfc-gh-azhan, line 2: nit: is it possible to replace this to the commit from mono-repo

sfc-gh-azhan (approved):

src/snowflake/snowpark/_internal/proto/ast.proto
sfc-gh-azwiegincew: This is codegened. I'll take care of this in a subsequent PR.

src/snowflake/snowpark/_internal/proto/ast.proto
sfc-gh-azwiegincew: Oooh, yeah. Let me do that separately.

sfc-gh-azwiegincew: Let me clarify: I'll figure out a way to get back to the GIT location instead of the weird scratch directory. I think that's the right solution – I generally believe generated code should point to where the source of truth actually lives in case someone wants to make changes.

sfc-gh-aalam (approved):

PR #2673, opened 2024-11-23
Adding a compatibility mode to ease migration usages

sfc-gh-mrojas:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-1824419, #2672

2. Fill out the following pre-review checklist:

   - [ X] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.
   - [x] I acknowledge that I have ensured my changes to be thread-safe. Follow the link for more information: [Thread-safe Developer Guidelines](https://docs.google.com/document/d/162d_i4zZ2AfcGRXojj0jByt8EUq-DrSHPPnTa4QvwbA/edit#bookmark=id.e82u4nekq80k)

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

This code, introduces a flag which once set, allows storing/getting custom values

PR #2751, opened 2024-12-11
SNOWVATION: Replace parameter boilerplate with utility classes.

sfc-gh-jrose:

   

   Fixes SNOW-NNNNNNN

The goal of this refactor is to provide a central location for defining parameters and other snowpark-python configuration. This should make it easier to add additional parameters in the future as well as gate more deprecated or exerimental logic behind parameters.

Session parmeters are now all added in `Session._initialize_config`. All definition needed for them should happen there.
Package level confuguration settings should similarly all be defined in src/snowflake/snowpark/_internal/config.py

Parameters and package level configs can all be accessed via session.conf.

As part of this change I've moved all of the session level parameters over to the new settings style, but there are likely package level configuration options that I have missed. For now the parameters that are exposed as properties of the session object remain available in this manner, but in the future we should standardize the access method to be through the config object.


src/snowflake/snowpark/session.py
```diff
@@ -760,16 +817,9 @@ def close(self) -> None:
                 _remove_session(self)
 
     @property
-    def conf(self) -> RuntimeConfig:
+    def conf(self) -> SettingStore:
```
sfc-gh-aalam, line 820: we should probably keep the name same for migration sake: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.conf.RuntimeConfig.html

tests/unit/test_session.py
```diff
@@ -630,18 +616,14 @@ def test_session_builder_app_name_existing_invalid_json_query_tag():
...
 )
 @pytest.mark.parametrize(
-    "parameter_name",
-    [
-        "_auto_clean_up_temp_table_enabled",
-        "_cte_optimization_enabled",
-        "_large_query_breakdown_enabled",
-    ],
+    "parameter_name", ["auto_clean_up_temp_table_enabled", "cte_optimization_enabled"]
 )
```
sfc-gh-aalam, line 620: merge artifact

tests/integ/test_temp_table_cleanup.py
```diff
@@ -253,9 +253,10 @@ def test_session_close(db_parameters, caplog):
 def test_auto_clean_up_temp_table_enabled_parameter(db_parameters, session, caplog):
     warning_dict.clear()
     with caplog.at_level(logging.WARNING):
-        session.auto_clean_up_temp_table_enabled = False
-    assert session.auto_clean_up_temp_table_enabled is False
+        session.auto_clean_up_temp_table_enabled = True
+    assert session.auto_clean_up_temp_table_enabled is True
     assert "auto_clean_up_temp_table_enabled is experimental" in caplog.text
+    session.auto_clean_up_temp_table_enabled = False
```
sfc-gh-aalam, line 259: do we not raise warning only when we enable the param?

src/snowflake/snowpark/_internal/config.py
```diff
@@ -0,0 +1,280 @@
...
+        name (str): The name of the setting that is used to reference it.
+        description (str): A docstring that describes the function of the setting.
+        default: (Any): The default value that a setting takes when not otherwise configured.
+        read_only (bool): Disallows modification of the setting when set to True.
+        experimental_since (str): When set this will warn users that changing the value of this
+            setting is experimental and may not be ready for production environments.
+    """
+
+    name: str
+    description: str | None = field(default=None)
```
sfc-gh-aalam, line 54:
```suggestion
    description: Optional[str] = field(default=None)
```
nit: we follow this style in the repo.

tests/integ/test_temp_table_cleanup.py
sfc-gh-jrose: The warning is only raised when setting to a non-default value.

sfc-gh-aalam:
would it be easy to cover changes made in this PR: https://github.com/snowflakedb/snowpark-python/pull/2673

sfc-gh-jrose:
> would it be easy to cover changes made in this PR: #2673

That wouldn't be too hard. I think I'd recommend downgrading from an error to a warning though so that users can be aware that they should clean things up.

sfc-gh-aalam (approved):
approving based on 
1. re-naming `SettingStore` to `RuntimeConfig` so that we don't change names of classes
2. adding a simple how-to maybe in contributing.md or config.py

PR #2753, opened 2024-12-12
SIT-2898: Add divnull and nullifzero

sfc-gh-aalam:

   

   Fixes SIT-2898

   Add functions - `divnull` and `nullifzero`


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 - Added support for `Series.str.ljust` and `Series.str.rjust`.
 - Added support for `Series.str.center`.
 - Added support for `Series.str.pad`.
+- Added support for the following functions
```
sfc-gh-yuwang, line 12: what about Added support for following functions in `functions.py`:

sfc-gh-yuwang:
I think we also need test for it? Otherwise it won't pass combine coverage

src/snowflake/snowpark/functions.py
```diff
@@ -2014,6 +2014,47 @@ def div0(
...
+        lit(dividend, _emit_ast=False)
+        if isinstance(dividend, (int, float))
+        else _to_col_if_str(dividend, "divnull")
+    )
+    divisor_col = (
+        lit(divisor, _emit_ast=False)
+        if isinstance(divisor, (int, float))
+        else _to_col_if_str(divisor, "divnull")
+    )
+    return dividend_col / nullifzero(divisor_col, _emit_ast=False)
```
sfc-gh-mrojas, line 2042: The problem with this is that in there are valid cases of divisoin that return 0, for example 0/100, 0/20

and with this approach we wont be able to distinguish those from failed ones. Also this we can parity with spark behavior

sfc-gh-mrojas: I just understood the nullifzero on the divisor that is great as divisions by zero will then be null

CHANGELOG.md
sfc-gh-aalam, line 12:
```suggestion
- Added support for the following functions in `functions.py`
```

sfc-gh-aalam:
> I think we also need test for it? Otherwise it won't pass combine coverage

I'm using doc tests to fulfill the coverage. Let's see if it fails. I'll add more test cases then.

sfc-gh-yuwang (approved):

sfc-gh-aling (approved):

sfc-gh-aherreraaguilar:
Hello @sfc-gh-aalam , I was checking this with @sfc-gh-omendezarias and realized that already exists the builtIn [div0null](https://docs.snowflake.com/en/sql-reference/functions/div0null) in snowflake? Is it better to use that instead?
Also, the name should match with the matching snowflake function "DIV0NULL"

PR #2755, opened 2024-12-12
SNOW-1801330: Map numpy functions in apply to builtin snowflake functions

sfc-gh-nkumar:

   

   Fixes SNOW-1801330

   Map numpy functions to builtin snowpark function if exists. In POC this provided 5x speedup.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
@@ -26,32 +26,7 @@ from snowflake.snowpark.modin.plugin._internal.type_utils import (
     pandas_lit,
     is_compatible_snowpark_types,
 )
-from snowflake.snowpark.functions import (
-    builtin,
-    col,
-    dense_rank,
-    ln,
-    log,
-    _log2,
-    _log10,
-    sin,
-    snowflake_cortex_sentiment,
-    snowflake_cortex_summarize,
-    udf,
-    to_variant,
-    when,
-    udtf,
-    exp,
-    cos,
-    tan,
-    sinh,
-    cosh,
-    tanh,
-    ceil,
-    floor,
-    trunc,
-    sqrt,
-)
+from snowflake.snowpark import functions as sp_func
 from snowflake.snowpark.modin.plugin._internal.frame import InternalFrame
 from snowflake.snowpark.modin.plugin._internal.ordered_dataframe import (
     OrderedDataFrame,
@@ -96,23 +71,23 @@ DEFAULT_UDTF_PARTITION_SIZE = 1000
 cloudpickle.register_pickle_by_value(sys.modules[__name__])
 
 SUPPORTED_SNOWPARK_PYTHON_FUNCTIONS_IN_APPLY = {
-    exp,
-    ln,
-    log,
-    _log2,
-    _log10,
-    sin,
-    cos,
-    tan,
-    sinh,
-    cosh,
-    tanh,
-    ceil,
-    floor,
-    trunc,
-    sqrt,
-    snowflake_cortex_sentiment,
-    snowflake_cortex_summarize,
+    sp_func.exp,
+    sp_func.ln,
+    sp_func.log,
+    sp_func._log2,
+    sp_func._log10,
+    sp_func.sin,
+    sp_func.cos,
+    sp_func.tan,
+    sp_func.sinh,
+    sp_func.cosh,
+    sp_func.tanh,
+    sp_func.ceil,
+    sp_func.floor,
+    sp_func.trunc,
+    sp_func.sqrt,
+    sp_func.snowflake_cortex_summarize,
+    sp_func.snowflake_cortex_sentiment,
 }
 
 
@@ -287,7 +262,7 @@ def create_udtf_for_apply_axis_1(
     ApplyFunc.end_partition._sf_vectorized_input = native_pd.DataFrame  # type: ignore[attr-defined]
 
     packages = list(session.get_packages().values()) + udf_packages
-    func_udtf = udtf(
+    func_udtf = sp_func.udtf(
         ApplyFunc,
         output_schema=PandasDataFrameType(
             [LongType(), StringType(), VariantType()],
@@ -709,7 +684,7 @@ def create_udtf_for_groupby_apply(
         excluded=existing_identifiers,
         wrap_double_underscore=False,
     )
-    return udtf(
+    return sp_func.udtf(
         ApplyFunc,
         output_schema=PandasDataFrameType(
             [StringType(), IntegerType(), VariantType(), IntegerType(), IntegerType()],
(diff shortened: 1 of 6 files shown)


tests/integ/modin/frame/test_applymap.py
```diff
@@ -107,7 +107,7 @@ def test_applymap_numpy(func):
     native_df = native_pd.DataFrame(data)
     snow_df = pd.DataFrame(data)
 
-    with SqlCounter(query_count=7, udf_count=1):
+    with SqlCounter(query_count=1):
```
sfc-gh-jkew, line 110: This is great

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -95,22 +71,66 @@
...
+    sp_func.cosh,
+    sp_func.tanh,
+    sp_func.ceil,
+    sp_func.floor,
+    sp_func.trunc,
+    sp_func.sqrt,
+    sp_func.snowflake_cortex_summarize,
+}
+
+NUMPY_FUNCTION_TO_SNOWFLAKE_FUNCTION = {
```
sfc-gh-jkew, line 92: The `numpy_to_pandas` file should changed as well, right? The numpy ufunc function mapping should be in one place.

sfc-gh-jkew (changes requested):

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-nkumar: They are coming from different sources. Let me see if it's possible to consolidate this into one.

sfc-gh-nkumar: one thing we can do is change entries in numpy_to_pandas.numpy_to_pandas_universal_func_map something like this
`"exp": lambda obj, inputs: obj.apply(sp_func.exp)` -> `"exp": lambda obj, inputs: obj.apply(np.exp)` 
And then numpy to snowpark function mapping will only be in `snowflake_query_compiler.applymap`
Does it make sense or were you thinking something else? @sfc-gh-jkew

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8755,6 +8756,12 @@ def applymap(
                     f"Snowpark pandas applymap API doesn't yet support Snowpark Python function `{func.__name__}` with args = '{args}'."
                 )
             return self._apply_snowpark_python_function_to_columns(func, kwargs)
+
+        # Check if the function is a known numpy function that can be translated to Snowflake function.
+        sf_func = NUMPY_FUNCTION_TO_SNOWFLAKE_FUNCTION.get(func)
+        if sf_func is not None:
+            return self._apply_snowpark_python_function_to_columns(sf_func, kwargs)
```
sfc-gh-helmeleegy, line 8763: Any idea why the code coverage is complaining about this line even though it's clear that there is a test covering it? The query_count is even reduced, which means that we must have entered into this branch!

sfc-gh-nkumar: that is strange. need to debug this.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -95,22 +71,66 @@
...
+    np.reciprocal: lambda col: 1 / col,
+    np.exp: sp_func.exp,
+    np.exp2: lambda col: sp_func.pow(2, col),
+    np.expm1: lambda col: sp_func.exp(col) - 1,
+    np.log: sp_func.ln,
+    np.log2: sp_func._log2,
+    np.log10: sp_func._log10,
+    np.log1p: lambda col: sp_func.ln(col + 1),
+    # Aggregate functions translate to identity functions when applied element wise
+    np.sum: lambda col: col,
```
sfc-gh-jkew, line 110: Is this right? `df.apply(np.sum)` would translate to `df.apply(lambda col: col)`

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-jkew: That might add an unnecessary level of indirection in some cases.  
   `df.apply(np.cos)` is the same as `np.cos(df)`and ultimately it should all map to `sp_func.cos`

The np.<function> mapping can be occasionally different than the df.apply(<function>) mapping, but for simple unary functions it's probably the same. We might be able to refactor this to treat unary functions differently, but that might be to confusing to the reader.

Maybe the best idea is to move `NUMPY_FUNCTION_TO_SNOWFLAKE_FUNCTION` into `numpy_to_pandas.py` - at least then we have all numpy compat stuff in the same file.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-nkumar: Yeah when applied element-wise aggregate functions are pretty much no-op. 
Screenshot 2024-12-13 at 11 37 04 AM

sfc-gh-nkumar: Just to clarify this is applicable to series.apply, series.map, df.applymap and df.map (which all work elementwise)

sfc-gh-nkumar: df.apply is not impacted by this PR.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-nkumar: Moved `NUMPY_FUNCTION_TO_SNOWFLAKE_FUNCTION` into `numpy_to_pandas.py`

sfc-gh-jkew (approved):

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -26,32 +26,7 @@
...
-    tan,
-    sinh,
-    cosh,
-    tanh,
-    ceil,
-    floor,
-    trunc,
-    sqrt,
-)
+from snowflake.snowpark import functions as sp_func
```
sfc-gh-jjiao, line 29: Good refactor, thanks

tests/integ/modin/frame/test_applymap.py
sfc-gh-jjiao: qq: what change has reduce the query_count from 7 to 1 and eliminate the udf call?

sfc-gh-jjiao (approved):

PR #2774, opened 2024-12-17
SNOW-1794377: Add support for Dataframe.toJSON

sfc-gh-jrose:

   

   Fixes SNOW-1794377

   This PR adds an approximation of pysparks Dataframe.toJSON method. Notably Snowpark does not support RDDs so instead we return a Dataframe. Additionally Snowflake JSON methods don't support the `use_unicode` flag that pyspark provides so that is left out as well. We could add it as a dummy value for compatibility, but I think we should stick to what the Snowflake API can provide.

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -5623,7 +5623,7 @@ Query List:
             >>> df = session.create_dataframe([(1, "a"), (2, "b")], schema=["a", "b"])
             >>> df.to_json().show()
             -------------------
-            |"TO_JSON"         |
+            |"TO_JSON"        |
             -------------------
             |{"A":1,"B":"a"}  |
             |{"A":2,"B":"b"}  |
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -5615,22 +5615,22 @@ Query List:
         ]
         return dtypes
 
-    def toJSON(self) -> "DataFrame":
+    def to_json(self) -> "DataFrame":
         """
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -4304,6 +4306,23 @@ def dtypes(self) -> List[Tuple[str, str]]:
         ]
         return dtypes
 
+    def toJSON(self) -> "DataFrame":
+        """
+        Converts each row of the DataFrame into a JSON string.
```
sfc-gh-aalam, line 4311: don't we also have to handle `_emit_ast` here?

sfc-gh-aalam (commented):
we also should update the dataframe.rst file

src/snowflake/snowpark/dataframe.py
sfc-gh-jrose: As far as I can tell none of the other dataframe functions handle _emit_ast yet so I assume this does not need to either.

src/snowflake/snowpark/dataframe.py
```diff
@@ -5613,6 +5615,23 @@ def dtypes(self) -> List[Tuple[str, str]]:
         ]
         return dtypes
 
+    def toJSON(self) -> "DataFrame":
+        """
```
sfc-gh-aalam, line 5619: a nitpick but I also feel like we should add to_json and add alias for `toJSON`. This has been pattern in snowpark repo. I don't like this practice right now but for migration purposes it is useful

sfc-gh-aalam (approved):

PR #2775, opened 2024-12-17
SNOW-1722641: Add support for Series.between

sfc-gh-joshi:

   

   Fixes SNOW-1722641

 Implements `Series.between`. The frontend implementation uses [modin's implementation](https://github.com/modin-project/modin/blob/1c4d173d3b2c44a1c1b5d5516552c7717b26de32/modin/pandas/series.py#L795), which passes the `modin.pandas.Series` object to the [native pandas method](https://github.com/pandas-dev/pandas/blob/9fe33bcbca79e098f9ba8ffd9fcf95440b95032b/pandas/core/series.py#L5362-L5380), which directly uses comparison operators to implement the method.


sfc-gh-jjiao (approved):
This is just fall back to native pandas, right?

tests/integ/modin/series/test_between.py
```diff
@@ -0,0 +1,71 @@
...
+
+@sql_count_checker(query_count=1)
+def test_series_between_default_inclusive():
+    eval_snowpark_pandas_result(
+        *create_test_series(list(range(0, 10))), lambda ser: ser.between(2, 8)
+    )
+
+
+# tuples of (data, low, high)
+BETWEEN_TEST_ARGUMENTS = [
```
sfc-gh-helmeleegy, line 25: Can we include nulls in the test arguments?

src/snowflake/snowpark/modin/plugin/extensions/series_overrides.py
```diff
@@ -149,11 +149,6 @@ def autocorr(self, lag=1):  # noqa: PR01, RT01, D200
     pass  # pragma: no cover
 
 
-@register_series_not_implemented()
```
sfc-gh-helmeleegy, line 152: So are you saying that all we needed to do was to inherit the modin implementation? Nice!

sfc-gh-helmeleegy: (I guess that's because it's re-written to other supported APIs.)

sfc-gh-joshi: Yep, the modin implementation basically works out of the box.

tests/integ/modin/series/test_between.py
sfc-gh-joshi: I just added, some and it looks like we have some inconsistencies with pandas because we have different behavior because comparing with a NULL value in snowflake always produces a NULL result:

```
>>> pd.Series([0.8, np.nan]) < 1  # snowpark pandas
0    True
1    None
dtype: object
>>> native_pd.Series([0.8, np.nan]) < 1
>>> native_pd.Series([0.8, np.nan]) < np.nan
0    False
1    False
dtype: bool
```
This behavior propagates to `between`:
```
>>> pd.Series([0.8, np.nan]).between(np.nan, 1)
0    None
1    None
dtype: object
>>> native_pd.Series([0.8, np.nan]).between(np.nan, 1)
0    False
1    False
dtype: bool
```

We could get around this by overriding the upstream method and just sticking an `astype(bool)` to coerce the `None` values, but this would break our principle of adhering to SQL behavior. Alternatively, we could just modify the tests to expect None results as needed. What do you think?

sfc-gh-joshi:
> This is just fall back to native pandas, right?

Yes, but not in a way that causes extra materialization (the native pandas method just uses comparison operators and &, so we can preserve lazy evaluation).

tests/integ/modin/series/test_between.py
sfc-gh-helmeleegy: Let's override the upstream method. Sticking to pandas behavior is the North Star. Snowflake sql behavior is the second best.

sfc-gh-helmeleegy:
> > This is just fall back to native pandas, right?
> 
> Yes, but not in a way that causes extra materialization (the native pandas method just uses comparison operators and &, so we can preserve lazy evaluation).

No, native pandas is not being used here. This was just falling back to modin implementation which itself routes to other supported Snowpark pandas methods (comparison operators).

tests/integ/modin/series/test_between.py
sfc-gh-joshi: Should we just override our implementations for the comparison (`<`/`<=` etc.) operators then? Those diverge from pandas (as the code example shows), and fixing those would fix `between` for free.

sfc-gh-joshi:
> No, native pandas is not being used here.

The modin frontend implementation ([here](https://github.com/modin-project/modin/blob/1c4d173d3b2c44a1c1b5d5516552c7717b26de32/modin/pandas/series.py#L795)) actually calls a native pandas method ([here](https://github.com/pandas-dev/pandas/blob/9fe33bcbca79e098f9ba8ffd9fcf95440b95032b/pandas/core/series.py#L5362-L5380)), but the native pandas method routes to Snowpark pandas comparison operators.

tests/integ/modin/series/test_between.py
sfc-gh-helmeleegy: On second thoughts, I think we should stay consistent with the other comparison operators. But let's still have tests covering nulls. You can check the tests used for the binary comparison ops.

tests/integ/modin/series/test_between.py
```diff
@@ -0,0 +1,71 @@
...
+        dt.datetime(2019, 1, 1, 0, 0),
+        dt.datetime(2021, 1, 1, 0, 0),
+    ),
+]
+
+
+@sql_count_checker(query_count=1)
+@pytest.mark.parametrize("data, low, high", BETWEEN_TEST_ARGUMENTS)
+@pytest.mark.parametrize("inclusive", ["both", "neither", "left", "right"])
+def test_series_between(data, low, high, inclusive):
```
sfc-gh-helmeleegy, line 47: We should also test when `low` and `high` are series. I think it will just work.

tests/integ/modin/series/test_between.py
sfc-gh-joshi: From what I could tell, the tests in `tests/integ/modin/binary/test_binary_op.py` actually avoided testing comparison operators on data with NULLs. I added some tests for `between` with NULL data, and used `astype(bool)` to cast NULL values at the end.

tests/integ/modin/series/test_between.py
sfc-gh-joshi: added

tests/integ/modin/series/test_between.py
sfc-gh-helmeleegy: Hmm, I thought they used `_gen_random_int_list_with_nones` for testing with NULLs. Also, I think this is where NULL comparison is handled in tests: https://github.com/snowflakedb/snowpark-python/blob/3c38e82a288e1c47f9034d381cbc1c175b520d31/tests/integ/modin/utils.py#L275

sfc-gh-joshi: All the tests I saw that used `_gen_random_int_list_with_nones` only used arithmetic operators, and skipped comparisons, like this one: https://github.com/snowflakedb/snowpark-python/blob/8e614ee5a4114c25857ac7d5cc56eb0ec439431a/tests/integ/modin/binary/test_binary_op.py#L662

sfc-gh-helmeleegy: Does this one use it?
https://github.com/snowflakedb/snowpark-python/blob/8e614ee5a4114c25857ac7d5cc56eb0ec439431a/tests/integ/modin/binary/test_binary_op.py#L1924

sfc-gh-joshi: Ah, I didn't see this test. It seems like these cast the native pandas data to "Float64" (capital F), which has the same behavior of `None < None` evaluating to `pd.NA`. The tests I added have mixed object + datetime data so we can't always do the same cast here.

sfc-gh-helmeleegy: That's fair. I just wanted to make sure that you're aware of those tests. I think we already have good coverage here.

sfc-gh-helmeleegy (approved):
Nice work. Thanks, Jonathan!

PR #2779, opened 2024-12-18
SNOW-1801330: Map numpy functions in df.apply to builtin functions

sfc-gh-nkumar:

   

   Fixes SNOW-1801330

   Map numpy universal functions (element-wise functions) and aggregate functions in df.apply to sql builtin functions.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -8783,17 +8783,16 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
                 )
             return self._apply_snowpark_python_function_to_columns(func, kwargs)
 
-        # Check if the function is a known numpy function that can be translated to Snowflake function.
+        # TODO SNOW-1739034: remove pragma no cover when apply tests are enabled in CI
+        # Check if the function is a known numpy function that can be translated to
+        # Snowflake function.
         sf_func = NUMPY_UNIVERSAL_FUNCTION_TO_SNOWFLAKE_FUNCTION.get(func)
-        if sf_func is not None:
-            # TODO SNOW-1739034: remove pragma no cover when apply tests are enabled in CI
-            return self._apply_snowpark_python_function_to_columns(
-                sf_func, kwargs
-            )  # pragma: no cover
+        if sf_func is not None:  # pragma: no cover
+            return self._apply_snowpark_python_function_to_columns(sf_func, kwargs)
 
-        if func in (np.sum, np.min, np.max):
+        if func in (np.sum, np.min, np.max):  # pragma: no cover
             # Aggregate functions applied element-wise to columns are no-op.
-            return self  # pragma: no cover
+            return self
 
         # Currently, NULL values are always passed into the udtf even if strict=True,
         # which is a bug on the server side SNOW-880105.
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -373,7 +373,7 @@ from snowflake.snowpark.modin.plugin.utils.error_message import ErrorMessage
 from snowflake.snowpark.modin.plugin.utils.warning_message import WarningMessage
 from snowflake.snowpark.modin.utils import MODIN_UNNAMED_SERIES_LABEL
 from snowflake.snowpark.modin.plugin.utils.numpy_to_pandas import (
-    NUMPY_FUNCTION_TO_SNOWFLAKE_FUNCTION,
+    NUMPY_UNIVERSAL_FUNCTION_TO_SNOWFLAKE_FUNCTION,
 )
 from snowflake.snowpark.session import Session
 from snowflake.snowpark.types import (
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8760,13 +8784,17 @@ def applymap(
...
         # Check if the function is a known numpy function that can be translated to Snowflake function.
-        sf_func = NUMPY_FUNCTION_TO_SNOWFLAKE_FUNCTION.get(func)
+        sf_func = NUMPY_UNIVERSAL_FUNCTION_TO_SNOWFLAKE_FUNCTION.get(func)
         if sf_func is not None:
             # TODO SNOW-1739034: remove pragma no cover when apply tests are enabled in CI
             return self._apply_snowpark_python_function_to_columns(
                 sf_func, kwargs
             )  # pragma: no cover
 
+        if func in (np.sum, np.min, np.max):
```
sfc-gh-helmeleegy, line 8794: What about other aggregate functions like mean, std, var, etc?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
```diff
@@ -308,10 +309,6 @@ def map_to_bools(inputs: Any) -> Any:
     np.log2: sp_func._log2,
     np.log10: sp_func._log10,
     np.log1p: lambda col: sp_func.ln(col + 1),
-    # Aggregate functions translate to identity functions when applied element wise
```
sfc-gh-helmeleegy, line 311: Is this removed because we added the block above for handling aggregate functions?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8760,13 +8784,17 @@ def applymap(
...
+            # Aggregate functions applied element-wise to columns are no-op.
+            return self  # pragma: no cover
```
sfc-gh-helmeleegy, line 8796: why `# pragma: no cover`?

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
```diff
@@ -291,7 +291,8 @@ def map_to_bools(inputs: Any) -> Any:
 }
 
 
-NUMPY_FUNCTION_TO_SNOWFLAKE_FUNCTION = {
+# Map numpy universal (element-wise) function to Snowflake function
```
sfc-gh-jkew, line 294: I would add a comment here that this is specifically for `df.apply(np.<func>)` as opposed to `np.<func>(df)` - we have two maps for ufuncs which are used in different ways - one which allows numpy to map functions to snowpandas, and another which allows snowpandas to map numpy functions to snowflake.

tests/integ/modin/frame/test_apply_axis_0.py
```diff
@@ -88,11 +88,12 @@ def test_axis_0_basic_types_without_type_hints(data, func, return_type):
     # this test processes functions without type hints and invokes the UDTF solution.
     native_df = native_pd.DataFrame(data, columns=["A", "b"])
     snow_df = pd.DataFrame(data, columns=["A", "b"])
+    # np.min is mapped to builtin function so no UDTF is required.
     with SqlCounter(
-        query_count=11,
-        join_count=2,
-        udtf_count=2,
-        high_count_expected=True,
+        query_count=1 if func == np.min else 11,
```
sfc-gh-jkew, line 93: This is a huge improvement.

sfc-gh-jkew (approved):
Looks good; but I would clear up the difference between the numpy ufunc maps with a comment.

tests/integ/modin/frame/test_apply_axis_0.py
sfc-gh-helmeleegy: Yes, whenever we can avoid the standard apply path, it's a huge win.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: Right now apply tests don't run due to CI. so code coverage check fails. This is conversation thread https://snowflake.slack.com/archives/C04HF38JFAQ/p1734113965907879

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-nkumar: Right.

src/snowflake/snowpark/modin/plugin/utils/numpy_to_pandas.py
sfc-gh-nkumar: Ack, let me add a comment.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: mean, std, var are mapped to builtin functions when used in df.apply. But when applied element-wise I have skipped it for now. Ideally users should not be using aggregate functions in element wise apply methods as it doesn't do much (most likely a user error).

sfc-gh-helmeleegy: Agreed. I just thought it's the same case for sum/min/max (when applied element-wise). So expanding that list would just be for completeness/consistency. But it's not necessary.

sfc-gh-helmeleegy (approved):
LGTM. Thanks, Naresh!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-nkumar: I see, I skipped them because output type will be float so we need to add casting np.mean(<int>) -> float.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8435,6 +8435,30 @@ def apply(
...
+            return self._apply_snowpark_python_function_to_columns(sf_func, kwargs)
+
+        if get_snowflake_agg_func(func, {}, axis) is not None:  # pragma: no cover
+            # np.std and np.var 'ddof' parameter defaults to 0 but
+            # df.std and df.var 'ddof' parameter defaults to 1.
+            # Set it here explicitly to 0 if not provided.
+            if func in (np.std, np.var) and "ddof" not in kwargs:
+                kwargs["ddof"] = 0
+            # np.median return NaN if any value is NaN while df.median skips NaN values.
+            # Set 'skipna' to false to match behavior.
```
sfc-gh-jjiao, line 8452: so for median, we will use np.median behavior?

sfc-gh-nkumar: Yes, default behavior for np.median is skipna == False. We need to map this to df.median/ df.agg('median') that by default is skipna == True. So we need to change this here so we can reuse existing code which implements df.agg('median').

sfc-gh-jjiao: ok, so func is the targeted function we need to mapped to

PR #2781, opened 2024-12-18
SNOW-1819521: Add support for Series.dt.strftime (10 directives)

sfc-gh-helmeleegy:

   

   Fixes SNOW-1819521

   Add support for Series.dt.strftime (10 directives).


docs/source/modin/supported/series_dt_supported.rst
```diff
@@ -98,7 +98,9 @@ the method in the left column.
 +-----------------------------+---------------------------------+----------------------------------------------------+
 | ``normalize``               | Y                               |                                                    |
 +-----------------------------+---------------------------------+----------------------------------------------------+
-| ``strftime``                | N                               |                                                    |
+| ``strftime``                | P                               | ``N`` if `date_format` contains directives other   |
```
sfc-gh-nkumar, line 101: just curios, how many directives are supported in pandas?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -18145,8 +18145,102 @@ def dt_strftime(self, date_format: str) -> None:
         Returns:
             New QueryCompiler containing formatted date-time values.
         """
-        ErrorMessage.not_implemented(
-            "Snowpark pandas doesn't yet support the method 'Series.dt.strftime'"
+
+        def strftime_func(column: SnowparkColumn) -> SnowparkColumn:
```
sfc-gh-nkumar, line 18149: I am wondering if it's possible to use `to_char` method with format. And translate pandas directives to snowflake directives https://docs.snowflake.com/en/sql-reference/functions-conversion#date-and-time-formats-in-conversion-functions
I believe that will be more efficient solution?

docs/source/modin/supported/series_dt_supported.rst
sfc-gh-helmeleegy: This should be the full list: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Yes, great idea. Will try that.

sfc-gh-helmeleegy: It may not actually work in the general case. The issue is that if the text aside from the directives inside `date_format` conflicts with the Snowflake directives, `to_char` will return the wrong result (e.g., if the text contained "MM"). Also, not all pandas directives are mappable to Snowflake directives.

So, I think we can later add an optimization to first check if the given `date_format` is eligible to be implemented using `to_char`. If so, we use `to_char`. Otherwise, we use the default approach (the one in this PR).

sfc-gh-nkumar: I see, thanks!

sfc-gh-nkumar (approved):

tests/integ/modin/series/test_dt_accessor.py
```diff
@@ -433,6 +433,71 @@ def test_days_in_month(property):
...
 
 
+@sql_count_checker(query_count=1)
+@pytest.mark.parametrize(
+    "date_format",
+    [
+        "a%d-%m-%Y-%H-%M-%S-%f-%j-%X-%%b",
+        "%d-%m-%Y-%H-%M-%S-%f-%j-%X-%%b",
+        "a%d-%m-%Y-%H-%M-%S-%f-%j-%X-%%",
+        "%d-%m-%Y-%H-%M-%S-%f-%j-%X-%%",
```
sfc-gh-joshi, line 443: can you add a couple cases for more pathological-looking strings, like `"%%%M"` or `"%%M"`? Also, what happens if we have an unfinished expression like `"abc%"`?

sfc-gh-joshi (approved):
LGTM, just one question about an edge case.

tests/integ/modin/series/test_dt_accessor.py
sfc-gh-helmeleegy: Added.

sfc-gh-helmeleegy: For the `"abc%"` case, it will be kept as is. So it's treated the same way as `"abc%%"`. This is how pandas behaves and we now have a test for it too.

PR #2789, opened 2024-12-19
Unify Snowflake object name handling in the Snowpark AST

sfc-gh-oplaton:

   

   Fixes SNOW-1621205

This is the client-side change corresponding to https://github.com/snowflakedb/snowflake/pull/240557

Unify Snowflake object name handling in the Snowpark AST.
Remove `FnName` and `SpTableName`. They both had `Flat` and `Structured` variants, but ultimately designate Snowflake object names.
Introduce `data SpName` and `entity SpNameRef` for referring to Snowflake objects by relative or fully qualified name.
Update `FnNameRefExpr`.
Use `SpNameRef` in a few places that used to use `List[String]`.


tests/ast/data/DataFrame.agg.test
```diff
@@ -239,49 +206,39 @@ body {
                   fn {
                     builtin_fn {
                       name {
-                        fn_name_flat {
-                          name: "col"
+                        name {
```
sfc-gh-vbudati, line 209: This looks like it's nested twice -- is this intentional?

sfc-gh-oplaton: Yes, the outer `name` is of type `SpNameRef` and comes from the trait `FnNameRefExpr`, and the inner `name` is the field of the `SpNameRef`, of type `SpName`.

src/snowflake/snowpark/_internal/ast/utils.py
```diff
@@ -358,26 +369,27 @@ def _set_fn_name(
...
-def build_sp_table_name(  # type: ignore[no-untyped-def] # TODO(SNOW-1491199) # Function is missing a return type annotation
-    expr_builder: proto.SpTableName, name: Union[str, Iterable[str]]
-):  # pragma: no cover
-    if isinstance(name, str):
-        expr_builder.sp_table_name_flat.name = name
-    elif isinstance(name, Iterable):
-        expr_builder.sp_table_name_structured.name.extend(name)
-    else:
-        raise ValueError(f"Invalid name type {type(name)} for SpTableName entity.")
+def build_sp_table_name(
```
sfc-gh-azwiegincew, line 379: Are these type-specific functions necessary? One thing I notice is that the re-emitted exceptions lose the diagnostic contents of the original exception.

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/_internal/ast/utils.py
sfc-gh-oplaton: > One thing I notice is that the re-emitted exceptions lose the diagnostic contents of the original exception.

That's weird, `raise ... from ...` should preserve the contents. The functions aren't strictly necessary, but I find the call sites more readable (the name clarifies the intent), and the errors should have been more readable, too.

sfc-gh-evandenberg (approved):
LGTM

PR #2805, opened 2024-12-20
SNOW-1865926: Infer schema for StructType columns from nested Rows

sfc-gh-jrose:

   

   Fixes SNOW-1865926

   With the new semantic mode enabled structured MapType and ArrayType are already used by default. This PR makes it so that it's possible to express StructType objects in a create_dataframe call by passing them as Row objects similar to how pyspark would handle such data.


sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -441,6 +442,8 @@ def infer_type(obj: Any) -> DataType:
             if key is not None and value is not None:
                 return MapType(infer_type(key), infer_type(value))
         return MapType(NullType(), NullType())
+    elif isinstance(obj, Row) and context._should_use_structured_type_semantics():
+        return infer_schema(obj)
```
sfc-gh-aling, line 446: is there a chance that the given datatype is not StructType while users still input a Row as data? if so what would happen, do we error out?

sfc-gh-jrose: I'm not sure what that scenario would look like. Can you give an example?

sfc-gh-aling: sorry I didn't make it clear, actually I had two questions:

1. when the infer schema logic would be triggered for `Row` values -- is it only when schema is not explicitly set?

2. for my own learning purposes, will following Row input + MapType be a valid input?
```python
    struct = Row(f1="v1", f2=2)
    df = structured_type_session.create_dataframe(
        [
            (struct),
        ],
        schema=[StructureType(MapType(xxx)],
    )
```

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jrose: 1. Yes, the schema is only inferred if no explicitly set.
2. Today this example would give an error like this:
```
>>> struct = Row(f1="v1", f2=2)
>>> df = session.create_dataframe([(struct,),], schema=StructType([StructField("test", MapType())]))
  File "src/snowflake/snowpark/_internal/utils.py", line 960, in func_call_wrapper
    return func(*args, **kwargs)
  File "src/snowflake/snowpark/session.py", line 3318, in create_dataframe
    raise TypeError(
TypeError: Cannot cast <class 'snowflake.snowpark.row.Row'>(Row(f1='v1', f2=2)) to MapType(StringType(), StringType()).
```
Currently createDataFrame does not know how to handle casting Rows to Maps. You could get it to work by calling Row.as_dict if you wanted it to be a MapType.

After this change it's improved slightly:
```
# Inferred
>>> df = session.create_dataframe([(struct),])
>>> df.schema
StructType([StructField('F1', StringType(), nullable=False), StructField('F2', LongType(), nullable=False)])
 
# or explicit
>>> df = session.create_dataframe([(struct),], schema=StructType([StructField('F1', StringType(), nullable=False), StructField('F2', LongType(), nullable=False)]))
```
Without this change the explicit schema still works.

PR #2815, opened 2024-12-29
SNOW-1869388 add memoization to to_selectable

sfc-gh-aalam:

   

   Fixes SNOW-1869388

   The process of adding WithQueryBlocks during repeated subquery elimination optimization takes place as follows:

- Identify repeated sub-query blocks
- Add a WithQueryBlock - a snowflake LogicalPlan node
- Create a SnowflakePlan out of the WithQueryBlock.

When parents for the WithQueryBlocks are updated, if the direct parent is a Selectable, we convert the SnowflakePlan into a Selectable using to_selectable. In case the WithQueryBlock has multiple parents, which by construction will always be true, a new Selectable object is created for each parent. See example below where new selectable nodes are highlighted in gray:

image

When large query block is applied on this plan, and the WithQueryBlock is chosen at eligible, since all these blocks have same statistics, they will all be eligible together and possible be materialized we create redundant materializations. The correct construction for such a case should look like the plan shown below:

image

### DIFF
--- a/src/snowflake/snowpark/_internal/compiler/query_generator.py
+++ b/src/snowflake/snowpark/_internal/compiler/query_generator.py
@@ -68,16 +68,29 @@ class QueryGenerator(Analyzer):
         # order of when the with query block is visited. The order is important to make sure the dependency
         # between the CTE definition is satisfied.
         self.resolved_with_query_block: Dict[str, Query] = {}
+        # This is a memoization dict for storing the selectable for a SnowflakePlan when to_selectable
+        # method is called with the same SnowflakePlan. This is used to de-duplicate nodes created during
+        # compilation process
+        self._to_selectable_memo_dict = {}
 
     def to_selectable(self, plan: LogicalPlan) -> Selectable:
         """Given a LogicalPlan, convert it to a Selectable."""
         if isinstance(plan, Selectable):
             return plan
 
-        snowflake_plan = self.resolve(plan)
-        selectable = SelectSnowflakePlan(snowflake_plan, analyzer=self)
-        selectable._is_valid_for_replacement = snowflake_plan._is_valid_for_replacement
-        return selectable
+        with self.session._lock:
+            if (
+                isinstance(plan, SnowflakePlan)
+                and plan._uuid in self._to_selectable_memo_dict
+            ):
+                return self._to_selectable_memo_dict[plan._uuid]
+            snowflake_plan = self.resolve(plan)
+            selectable = SelectSnowflakePlan(snowflake_plan, analyzer=self)
+            selectable._is_valid_for_replacement = (
+                snowflake_plan._is_valid_for_replacement
+            )
+            self._to_selectable_memo_dict[snowflake_plan._uuid] = selectable
+            return selectable
 
     def generate_queries(
         self, logical_plans: List[LogicalPlan]
--- a/src/snowflake/snowpark/_internal/compiler/query_generator.py
+++ b/src/snowflake/snowpark/_internal/compiler/query_generator.py
@@ -79,8 +79,9 @@ class QueryGenerator(Analyzer):
             return plan
 
         plan_id = hex(id(plan))
-        if isinstance(plan, SnowflakePlan) and plan_id in self._to_selectable_memo_dict:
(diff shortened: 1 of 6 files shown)


tests/unit/conftest.py
```diff
@@ -79,6 +79,7 @@ def mock_session(mock_analyzer) -> Session:
     fake_session = mock.create_autospec(Session)
     fake_session._cte_optimization_enabled = False
     fake_session._analyzer = mock_analyzer
+    fake_session._lock = mock.MagicMock()
```
sfc-gh-helmeleegy, line 82: Can we have a more descriptive name to this lock -- similar to `_plan_lock` below?

sfc-gh-helmeleegy (approved):
Looks good, thanks!

tests/unit/conftest.py
sfc-gh-aalam: this lock is actually part of `Session` object and used as a general purpose lock for serialization: https://github.com/snowflakedb/snowpark-python/blob/ba313012d75f613ecefd70a1058043374df4daf9/src/snowflake/snowpark/session.py#L644

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -68,16 +68,29 @@ def __init__(
...
+                isinstance(plan, SnowflakePlan)
+                and plan._uuid in self._to_selectable_memo_dict
+            ):
+                return self._to_selectable_memo_dict[plan._uuid]
+            snowflake_plan = self.resolve(plan)
+            selectable = SelectSnowflakePlan(snowflake_plan, analyzer=self)
+            selectable._is_valid_for_replacement = (
+                snowflake_plan._is_valid_for_replacement
+            )
+            self._to_selectable_memo_dict[snowflake_plan._uuid] = selectable
```
sfc-gh-yzou, line 92: we probably should remember the original logical plan id instead of the resolved plan id, because the different snowflake plan might be created for the same orignal plan node

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -68,16 +68,29 @@ def __init__(
...
-        return selectable
+        with self.session._lock:
+            if (
+                isinstance(plan, SnowflakePlan)
+                and plan._uuid in self._to_selectable_memo_dict
+            ):
+                return self._to_selectable_memo_dict[plan._uuid]
+            snowflake_plan = self.resolve(plan)
+            selectable = SelectSnowflakePlan(snowflake_plan, analyzer=self)
+            selectable._is_valid_for_replacement = (
```
sfc-gh-yzou, line 89: in fact, it should always be true in query generator instead of propogating from the snowlfake plan

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -68,16 +68,29 @@ def __init__(
...
     def to_selectable(self, plan: LogicalPlan) -> Selectable:
         """Given a LogicalPlan, convert it to a Selectable."""
         if isinstance(plan, Selectable):
             return plan
 
-        snowflake_plan = self.resolve(plan)
-        selectable = SelectSnowflakePlan(snowflake_plan, analyzer=self)
-        selectable._is_valid_for_replacement = snowflake_plan._is_valid_for_replacement
-        return selectable
+        with self.session._lock:
```
sfc-gh-yzou, line 81: why do we need a lock here, but not at other place in query_generator?

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-jdu: Yea it's true

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/compiler/query_generator.py
sfc-gh-aalam: We don't need it. query generator is not shared between threads so we should be good. That's why it is not added in other places.

sfc-gh-yzou (approved):

PR #2816, opened 2024-12-30
SNOW-1865904 fix query gen when nested cte node is partitioned

sfc-gh-aalam:

   

   Fixes SNOW-1865904

   Update query generation for with query block where cte query is updated if the child plan is re-resolved during optimization process.

### DIFF
--- a/src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
+++ b/src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
@@ -565,19 +565,14 @@ class LargeQueryBreakdown:
         temp_table_selectable.post_actions = [drop_table_query]
 
         parents = self._parent_map[child]
-        updated_nodes = set()
         for parent in parents:
             replace_child(parent, child, temp_table_selectable, self._query_generator)
 
         nodes_to_reset = list(parents)
         while nodes_to_reset:
             node = nodes_to_reset.pop()
-            if node in updated_nodes:
-                # Skip if the node is already updated.
-                continue
 
             update_resolvable_node(node, self._query_generator)
-            updated_nodes.add(node)
 
             parents = self._parent_map[node]
             nodes_to_reset.extend(parents)
--- a/src/snowflake/snowpark/_internal/compiler/plan_compiler.py
+++ b/src/snowflake/snowpark/_internal/compiler/plan_compiler.py
@@ -185,7 +185,6 @@ class PlanCompiler:
                     error_type=type(e).__name__,
                     error_message=str(e),
                 )
-                pass
 
         return self.replace_temp_obj_placeholders(queries)
 
--- a/src/snowflake/snowpark/_internal/compiler/query_generator.py
+++ b/src/snowflake/snowpark/_internal/compiler/query_generator.py
@@ -218,11 +218,10 @@ class QueryGenerator(Analyzer):
 
         elif isinstance(logical_plan, WithQueryBlock):
             resolved_child = resolved_children[logical_plan.children[0]]
-            # record the CTE definition of the current block
-            if logical_plan.name not in self.resolved_with_query_block:
-                self.resolved_with_query_block[
-                    logical_plan.name
-                ] = resolved_child.queries[-1]
+            # record/update the CTE definition of the current block
+            self.resolved_with_query_block[logical_plan.name] = resolved_child.queries[
+                -1
+            ]
 
             resolved_plan = self.plan_builder.with_query_block(
                 logical_plan,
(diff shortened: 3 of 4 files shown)


tests/integ/test_large_query_breakdown.py
```diff
@@ -734,6 +735,48 @@ def test_optimization_skipped_with_exceptions(
...
+    df2 = df2.group_by("B").agg(sum_distinct(col("A")).alias("A"))
+    mid_final_df = df1.union_all(df2)
+
+    mid1 = mid_final_df.filter(col("A") > 10)
+    mid2 = mid_final_df.filter(col("B") > 3)
+    final_df = mid1.union_all(mid2)
+
+    with SqlCounter(query_count=1, describe_count=0):
+        queries = final_df.queries
+        # TODO: update when to_selectable memoization is merged
```
sfc-gh-helmeleegy, line 767: Add a Jira ticket?

src/snowflake/snowpark/_internal/compiler/query_generator.py
```diff
@@ -218,11 +218,11 @@ def do_resolve_with_resolved_children(
 
         elif isinstance(logical_plan, WithQueryBlock):
             resolved_child = resolved_children[logical_plan.children[0]]
-            # record the CTE definition of the current block
-            if logical_plan.name not in self.resolved_with_query_block:
-                self.resolved_with_query_block[
-                    logical_plan.name
-                ] = resolved_child.queries[-1]
+            # record the CTE definition of the current block and update the query when
```
sfc-gh-helmeleegy, line 221: For clarity, I think it should it be `or` instead of `and`, right?
```suggestion
            # record the CTE definition of the current block or update the query when
```

sfc-gh-helmeleegy (approved):
LGTM, just a couple of nits. Thanks!

tests/integ/test_large_query_breakdown.py
sfc-gh-aalam: I'll just wait for https://github.com/snowflakedb/snowpark-python/pull/2815 to merge and I'll update this comment

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
```diff
@@ -565,19 +565,14 @@ def _replace_child_and_update_ancestors(
...
 
         parents = self._parent_map[child]
-        updated_nodes = set()
         for parent in parents:
             replace_child(parent, child, temp_table_selectable, self._query_generator)
 
         nodes_to_reset = list(parents)
         while nodes_to_reset:
             node = nodes_to_reset.pop()
-            if node in updated_nodes:
```
sfc-gh-yzou, line 575: can you give some more detailed description about what is the problem if we skip the node update here ? the cte optimizaiton is doing a level to level update, so it can skip the node update if it is already updated, but for large query breakdown, i recall you were doing a dfs, right?

src/snowflake/snowpark/_internal/compiler/plan_compiler.py
```diff
@@ -185,7 +185,6 @@ def compile(self) -> Dict[PlanQueryType, List[Query]]:
                     error_type=type(e).__name__,
                     error_message=str(e),
                 )
-                pass
```
sfc-gh-yzou, line 188: do we have a bug before?

sfc-gh-aalam: no. I just noticed an unnecessary `pass` so I cleaned it up.

src/snowflake/snowpark/_internal/compiler/large_query_breakdown.py
sfc-gh-aalam: consider a tree below
```
     node1
        |
     node2
   /       \
node3      node4
````

suppose we start by re-resolving `node3 -> node2 -> node1`. In this process, `node2` and `node1` are marked in `updated_nodes`. Now, when we go updating the ancestors of `node4`, re-resolution of `node2` and `node1` would be skipped. This is not ideal if `node4` update can also trigger a re-update of `node2` and `node1`.

For example, this could be problematic is when `node3` and `node4` before the update had `referenced_cte` map. After the first update `node3 -> node2 -> node1`, `node2` will be resolved with an older version of `node4`. If after a re-resolved `node4` there are no referenced_ctes, then `node2` will not be updated with this information.

sfc-gh-yzou: so it is because you are doing a dfs, if you are doing updated in the order of node3, node4 then node2, last node1 you shouldn't run into problems.

sfc-gh-yzou (approved):

sfc-gh-jdu (approved):

PR #2826, opened 2025-01-06
SNOW-1794509 Merging from main into decoder branch.

sfc-gh-batur:

   

   Fixes SNOW-1794509


tests/ast/decoder.py
```diff
@@ -195,10 +179,10 @@ def decode_table_name_expr(self, table_name: proto.SpTableName) -> str:
         str
             The decoded table name.
         """
-        if hasattr(table_name, "sp_table_name_flat"):
-            return table_name.sp_table_name_flat.name
-        elif hasattr(table_name, "sp_table_name_structured"):
-            return table_name.sp_table_name_structured.name
+        if table_name.name.HasField("sp_name_flat"):
```
sfc-gh-vbudati, line 182: I don't know if this issue pops up here but I believe that using `HasField` raises a ValueError if the field is not defined in the protobuf object.

tests/ast/decoder.py
```diff
@@ -160,28 +160,12 @@ def decode_dsl_map_expr(self, map_expr: Iterable) -> dict:
...
-        str
-            The decoded function name.
-        """
-        if hasattr(fn_name, "fn_name_flat"):
-            return fn_name.fn_name_flat.name
-        elif hasattr(fn_name, "fn_name_structured"):
-            return fn_name.fn_name_structured.name
-        else:
-            raise ValueError("Function name not found in proto.FnName")
+    def convert_name_to_list(self, name: any) -> List:
```
sfc-gh-vbudati, line 163: Is there a reason to not unify this method with `decode_name_expr`? I see that it is called only after `decode_name_expr` is.

sfc-gh-vbudati (commented):
Looks good for most part, I have a couple questions though regarding the decoder. Thanks Bala!

sfc-gh-evandenberg:
Can you fix up the description?  It doesn't contain the jira number.

sfc-gh-batur:
> Can you fix up the description? It doesn't contain the jira number.

Merges require Jira numbers ?

tests/ast/decoder.py
sfc-gh-batur: But not always. There are places where it is not called and I thought it might be better to have separate functions rather than a single function with a parameter...

tests/ast/decoder.py
sfc-gh-batur: The documentation says that calling it on a "non-message" type will throw an error. In this case name is a message type - so I think we should be ok. hasattr doesn't work because both the fields (flat and structured) are present and it returns true for both.

tests/ast/decoder.py
sfc-gh-vbudati: Hmm fair point since, we can keep it as it is then

tests/ast/decoder.py
sfc-gh-vbudati: sounds good, thanks Bala!

PR #2827, opened 2025-01-07
SNOW-1847174: Support include_groups=False in groupby.apply.

sfc-gh-mvashishtha:
Support include_groups=False in groupby.apply.

When include_groups=False, we exclude grouping columns from the inputs to the user-supplied `func`.

This commit changes `func` in some groupby.apply tests where `func` assumes that the grouping columns are in the inputs.

Fixes SNOW-1847174

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -4003,7 +4003,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
             series_groupby:
                 Whether we are performing a SeriesGroupBy.apply() instead of a DataFrameGroupBy.apply()
             include_groups:
-                When True, will apply func to the groupings in the case that
+                When True, will include grouping keys when calling func in the case that
                 they are columns of the DataFrame.
             force_single_group:
                 Force single group (empty set of group by labels) useful for DataFrame.apply() with axis=0
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -4066,7 +4066,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
                 if series_groupby
                 else (
                     # For DataFrameGroupBy, if include_groups, we apply the
-                    # function to all data columnns. Otherwise, we exclude
+                    # function to all data columns. Otherwise, we exclude
                     # data columns that we are grouping by.
                     include_groups
                     or identifier not in by_snowflake_quoted_identifiers_list
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -3979,6 +3979,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         agg_args: Any,
         agg_kwargs: dict[str, Any],
         series_groupby: bool,
+        include_groups: bool,
         force_single_group: bool = False,
         force_list_like_to_series: bool = False,
     ) -> "SnowflakeQueryCompiler":
@@ -4001,6 +4002,9 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
                 Keyword arguments to pass to agg_func when applying it to each group.
             series_groupby:
                 Whether we are performing a SeriesGroupBy.apply() instead of a DataFrameGroupBy.apply()
+            include_groups:
+                When True, will apply func to the groupings in the case that
+                they are columns of the DataFrame.
             force_single_group:
                 Force single group (empty set of group by labels) useful for DataFrame.apply() with axis=0
             force_list_like_to_series:
@@ -4018,14 +4022,6 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
                 f"No support for groupby.apply with parameters by={by}, "
                 + f"level={level}, and axis={axis}"
             )
-
-        if "include_groups" in agg_kwargs:
-            # exclude "include_groups" from the apply function kwargs
-            include_groups = agg_kwargs.pop("include_groups")
-            if not include_groups:
-                ErrorMessage.not_implemented(
-                    f"No support for groupby.apply with include_groups = {include_groups}"
-                )
 
         sort = groupby_kwargs.get("sort", True)
         as_index = groupby_kwargs.get("as_index", True)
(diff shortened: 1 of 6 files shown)


sfc-gh-joshi (approved):

sfc-gh-mvashishtha:
@sfc-gh-rdurrani @sfc-gh-joshi let me know if you can figure out why tests show that I don't have coverage of my query compiler changes: https://github.com/snowflakedb/snowpark-python/actions/runs/12642483047/job/35228662346

sfc-gh-joshi:
> @sfc-gh-rdurrani @sfc-gh-joshi let me know if you can figure out why tests show that I don't have coverage of my query compiler changes: https://github.com/snowflakedb/snowpark-python/actions/runs/12642483047/job/35228662346
> 
> I merged the latest main into my branch, ran some tests, and checked that i was hitting lines 4050,4076 in `src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py`.

The test file (`tests/integ/modin/groupby/test_groupby_apply.py`) has a module-level skipif if the installed pandas is 2.2.3, since any test that tries to construct a UDF won't be able to do so because it can't find that pandas version in the Snowflake Anaconda channel. The code path for groupby_apply thus isn't getting run in the merge gate.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4001,6 +4002,9 @@ def groupby_apply(
                 Keyword arguments to pass to agg_func when applying it to each group.
             series_groupby:
                 Whether we are performing a SeriesGroupBy.apply() instead of a DataFrameGroupBy.apply()
+            include_groups:
+                When True, will apply func to the groupings in the case that
```
sfc-gh-rdurrani, line 4006:
```suggestion
                When True, will include grouping keys when calling func in the case that
```

Minor nit: I feel like the current documentation provided by pandas is a little unclear, and that this may be clearer? Feel free to ignore if you disagree!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4051,17 +4047,36 @@ def groupby_apply(
...
-        # For SeriesGroupBy, this frame may also include some grouping columns
-        # that `func` should not take as input. In that case, the only column
-        # that `func` takes as input is the last data column, so grab just that
-        # column with a slice starting at index -1 and ending at None.
-        input_data_column_identifiers = (
-            self._modin_frame.data_column_snowflake_quoted_identifiers[
-                slice(-1, None) if series_groupby else slice(None)
-            ]
-        )
+        input_data_column_positions = [
```
sfc-gh-rdurrani, line 4050: Maybe we can have an external if-else rather than nesting it in the list comprehension? May make it clearer to read.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8514,6 +8531,7 @@ def wrapped_func(*args, **kwargs):  # type: ignore[no-untyped-def] # pragma: no
                     series_groupby=True,
                     force_single_group=True,
                     force_list_like_to_series=True,
+                    include_groups=True,
```
sfc-gh-rdurrani, line 8534: Maybe we should add a comment here explaining that we use the same codepath for apply and groupby_apply, and we're using a hack for apply that forces a single group and we want to include_group since we now support that parameter?

tests/integ/modin/groupby/test_groupby_apply.py
```diff
@@ -509,7 +534,7 @@ def operation(df: native_pd.DataFrame) -> native_pd.DataFrame:
                         native_pd.DataFrame(
                             {
                                 "custom_sum": [26, 30, 30, 46],
-                                "custom_string": ["k0e", "k1d-b", "k0c", "k1k0d-b"],
+                                "custom_string": ["ee", "dd-b", "cc", "dbd-b"],
```
sfc-gh-rdurrani, line 537: why are these strings changed?

sfc-gh-rdurrani (commented):
Left some minor nits and q's, but lgtm overall!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: I have a slight preference for list comprehensions in cases like this. I'll leave this one as is.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: Other comments in this function explain that we're using `groupby_apply` to implement `apply`. My view is that the code `include_groups=True` itself tells the reader enough to figure out that we want to include the groups. I prefer not to add extra comments here justifying the decision to include groups.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: Agreed; done.

tests/integ/modin/groupby/test_groupby_apply.py
sfc-gh-mvashishtha: Because we're summing a different column: https://github.com/snowflakedb/snowpark-python/blob/1f08f7636f1872864b5c1f6f6d100bca03abc771/tests/integ/modin/groupby/test_groupby_apply.py#L83

I couldn't use the original column because it was (at least sometimes) one of the group keys.

sfc-gh-jjiao (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4051,17 +4047,36 @@ def groupby_apply(
...
+                    # the last data column, so take just that column.
+                    # include_groups has no effect.
+                    i
+                    == len(self._modin_frame.data_column_snowflake_quoted_identifiers)
+                    - 1
+                )
+                if series_groupby
+                else (
+                    # For DataFrameGroupBy, if include_groups, we apply the
+                    # function to all data columnns. Otherwise, we exclude
```
sfc-gh-helmeleegy, line 4069:
```suggestion
                    # function to all data columns. Otherwise, we exclude
```

sfc-gh-helmeleegy (approved):
Looks good, thanks Mahesh!

sfc-gh-rdurrani (approved):
lgtm!

PR #2828, opened 2025-01-07
SNOW-1871175: Add support for specifying a schema string for `DataFrame.create_dataframe`

sfc-gh-jdu:

   

   Fixes SNOW-1871175

   Support schema string

### DIFF
--- a/src/snowflake/snowpark/_internal/type_utils.py
+++ b/src/snowflake/snowpark/_internal/type_utils.py
@@ -993,8 +993,8 @@ def get_string_length(type_str: str) -> Optional[int]:
 
 def extract_bracket_content(type_str: str, keyword: str) -> str:
     """
-    Given a string that starts with e.g. 'array<', returns the content inside the top-level <...>.
-    e.g., "array<int>" => "int"
+    Given a string that starts with e.g. "array<", returns the content inside the top-level <...>.
+    e.g., "array<int>" => "int". It also parses the nested array like "array<array<...>>".
     Raises ValueError on mismatched or missing bracket.
     """
     prefix_pattern = rf"(?i)^\s*{keyword}\s*<"
--- a/src/snowflake/snowpark/_internal/type_utils.py
+++ b/src/snowflake/snowpark/_internal/type_utils.py
@@ -969,6 +969,15 @@ DECIMAL_RE = re.compile(
 STRING_RE = re.compile(r"^\s*(varchar|string|text)\s*\(\s*(\d*)\s*\)\s*$")
 # support type string format like "  string  (  23  )  "
 
+ARRAY_RE = re.compile(r"(?i)^\s*array\s*<")
+# support type string format like starting with "array<..."
+
+MAP_RE = re.compile(r"(?i)^\s*map\s*<")
+# support type string format like starting with "map<..."
+
+STRUCT_RE = re.compile(r"(?i)^\s*struct\s*<")
+# support type string format like starting with "array<..." starting with "struct<..."
+
 
 def get_number_precision_scale(type_str: str) -> Optional[Tuple[int, int]]:
     decimal_matches = DECIMAL_RE.match(type_str)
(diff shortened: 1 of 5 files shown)


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -979,7 +988,147 @@ def get_string_length(type_str: str) -> Optional[int]:
...
+    inside_chars: List[str] = []
+    i = start_index
+    while i < len(type_str):
+        c = type_str[i]
+        if c == "<":
+            bracket_depth += 1
+            # we don't store the opening bracket in 'inside_chars'
+            # if bracket_depth was 0 -> 1, to skip the outer bracket
+            if bracket_depth > 1:
+                inside_chars.append(c)
```
sfc-gh-aling, line 1015: is this form allowed `array<<<...>>>`?

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -979,7 +988,147 @@ def get_string_length(type_str: str) -> Optional[int]:
...
+    """
+    Parse something like "a: int, b: string, c: array<int>"
+    into StructType([StructField('a', IntegerType()), ...]).
+    """
+    fields = []
+    field_defs = split_top_level_comma_fields(fields_str)
+    for field_def in field_defs:
+        # Try splitting on colon first, else whitespace
+        if ":" in field_def:
+            left, right = field_def.split(":", 1)
```
sfc-gh-aling, line 1045: should we consider multiple colon cases like "a:b:c"? or this is handled by upstream/downstream logic already

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -979,7 +988,147 @@ def get_string_length(type_str: str) -> Optional[int]:
...
+    parts = []
+    bracket_depth = 0
+    start_idx = 0
+    for i, c in enumerate(s):
+        if c in ["<", "("]:
+            bracket_depth += 1
+        elif c in [">", ")"]:
+            bracket_depth -= 1
+            if bracket_depth < 0:
+                raise ValueError(f"Mismatched bracket in '{s}'.")
```
sfc-gh-aling, line 1077: I feel this bracket check logic has repeated multiple times
do you think it's possible to check the bracket match as the initial step for only one time for the whole input string, and then in the downstream logic we can only focus on extracting the types

src/snowflake/snowpark/session.py
```diff
@@ -3173,7 +3184,14 @@ def create_dataframe(
...
         schema_query = None
-        if isinstance(schema, StructType):
+        if isinstance(schema, str):
+            schema = type_string_to_type_object(schema)
+            if not isinstance(schema, StructType):
+                raise ValueError(
+                    f"Invalid schema string: {schema}. "
+                    f"You should provide a valid schema string representing a struct type."
+                )
+        if isinstance(schema, (StructType, str)):
```
sfc-gh-aling, line 3194: would the schema still be of type `str` after being processed by the `type_string_to_type_object`?
I feel we don't need to the `str` in the instance check here
```suggestion
        if isinstance(schema, StructType):
```

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: for something like `struct<i: array<...>>`

src/snowflake/snowpark/session.py
sfc-gh-jdu: yea it's by mistake

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: nah, PySpark's simpleString format only considers the first colon or whitespace.

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: We need to parse bracket to split fields, and extract names and types anyway. There is indeed a duplicate of validating whether the bracket expression is valid or not, maybe we can remove it. But to make the function self-contained, maybe let's still keep it? They are also covered in the test.

tests/unit/test_types.py
```diff
@@ -1440,3 +1445,390 @@ def test_maptype_alias():
...
+    ), f"Expected {expected_field_kv}, got {dt.fields[1]}"
+
+
+def test_type_string_to_type_object_implicit_struct_with_spaces():
+    """
+    Test spacing variations. E.g. "  col1  :   int  ,  col2  :  map< string , decimal(5,2) >  ".
+    """
+    dt = type_string_to_type_object(
+        "  col1 :  int  ,  col2 :   map< string , decimal(5,2) > "
+    )
```
sfc-gh-aalam, line 1707:
```suggestion
    dt = type_string_to_type_object(
        "  col1 :  int  ,  col2 :   map< string , decimal( 5 , 2 ) > "
    )
```
can we add spacing here too

tests/integ/test_dataframe.py
```diff
@@ -4447,3 +4447,155 @@ def test_map_negative(session):
...
+def test_create_dataframe_with_implicit_struct_datetime(session):
+    """
+    Another example mixing basic data with boolean and dates, ensuring
+    the implicit struct string handles them properly.
+    """
+    data = [
+        [True, datetime.date(2020, 1, 1)],
+        [False, datetime.date(2021, 12, 31)],
+    ]
+    schema_str = "flag: boolean, d: date"
```
sfc-gh-aalam, line 4567: is something like `string(100), timestamp(3)` supported?

src/snowflake/snowpark/session.py
```diff
@@ -3061,9 +3062,15 @@ def create_dataframe(
...
             schema: A :class:`~snowflake.snowpark.types.StructType` containing names and
                 data types of columns, or a list of column names, or ``None``.
-                When ``schema`` is a list of column names or ``None``, the schema of the
-                DataFrame will be inferred from the data across all rows. To improve
-                performance, provide a schema. This avoids the need to infer data types
+
+                - When passing a **string**, it can be either an *explicit* struct
+                  (e.g. ``"struct<a: int, b: string>"``) or an *implicit* struct
+                  (e.g. ``"a: int, b: string"``). Internally, the string is parsed and
+                  converted into a :class:`StructType` using Snowpark's type parsing.
```
sfc-gh-aalam, line 3069: does this mean a valid struct must contain `col_name: data_type` pair?

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -979,7 +988,147 @@ def get_string_length(type_str: str) -> Optional[int]:
...
+            bracket_depth -= 1
+            if bracket_depth < 0:
+                raise ValueError(f"Mismatched '>' in '{type_str}'")
+            if bracket_depth == 0:
+                if i != len(type_str) - 1:
+                    raise ValueError(
+                        f"Unexpected characters after closing '>' in '{type_str}'"
+                    )
+                # done
+                return "".join(inside_chars).strip()
```
sfc-gh-jrose, line 1026: Is this faster than just appending to a string sequentially?

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -979,7 +988,147 @@ def get_string_length(type_str: str) -> Optional[int]:
...
+    if not match:
+        raise ValueError(
+            f"'{type_str}' does not match expected '{keyword}<...>' syntax."
+        )
+
+    start_index = match.end() - 1  # position at '<'
+    bracket_depth = 0
+    inside_chars: List[str] = []
+    i = start_index
+    while i < len(type_str):
```
sfc-gh-jrose, line 1008: should type_str be stripped in case of trailing whitespace?

sfc-gh-jrose, line 1008:
```suggestion
    for c in type_str[start_index:]:
```

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -979,7 +988,147 @@ def get_string_length(type_str: str) -> Optional[int]:
...
+
+def split_top_level_comma_fields(s: str) -> List[str]:
+    """
+    Splits 's' by commas not enclosed in matching brackets.
+    Example: "int, array<long>, decimal(10,2)" => ["int", "array<long>", "decimal(10,2)"].
+    """
+    parts = []
+    bracket_depth = 0
+    start_idx = 0
+    for i, c in enumerate(s):
```
sfc-gh-jrose, line 1071: Something like `array<long), decimal(10, 2>` would appear to be valid under this scheme. I also suspect something like `int, array<long` might slip through.

tests/integ/test_dataframe.py
sfc-gh-jdu: nah, because pyspark doesn't support it, we don't need to support it either for now.

src/snowflake/snowpark/session.py
sfc-gh-jdu: yes

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: Yea I originally wrote like this but turns out I also need to maintain `i`, which is easier to understand

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: hmm it's just a more convenient way to write code in one line :)

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: good point, fixed

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-jdu: It may be valid for this string but invalid for `type_string_to_type_object` with other checks. `array<long), decimal(10, 2>` is not valid, because `array` doesn't accept two elements; `int, array<long` is also not valid as `array<long` is not a valid type. I added such test cases in unit tests too.

sfc-gh-jrose (approved):

sfc-gh-wshangguan (commented):
Is this change able to parse "not null", e.g. `"struct struct<i: integer not null>"`

sfc-gh-aalam (approved):

PR #2831, opened 2025-01-07
Disable python 3.8 incompatible doctest

sfc-gh-jrose:

   

   Fixes SNOW-NNNNNNN

   Two of the doctests for snowflake.snowpark.Session require packages that are not available anymore for python 3.8 which causes a failure in our daily precommit tests. To fix this there are a few options:
   
1 Skip the doctests entirely.
2 Add some code in the doctest that skips it conditionally.
3 Intercept the doctest finder so that it does not include those tests when running in python 3.8.

I've opted for option 3 so that the tests can still be run on other python versions and there isn't code that might be confusing to readers in the examples.

### DIFF
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -538,13 +538,9 @@ jobs:
       - name: Install tox
         run: python -m pip install tox
         # only run doctest for macos on aws
-      - name: Check Resources
-        run: |
-          df -h
-          free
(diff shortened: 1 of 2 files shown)


sfc-gh-aalam (approved):

sfc-gh-aling (approved):
or we should remove python 3.8 from our matrix?

depending on the announcement: https://docs.snowflake.com/en/release-notes/clients-drivers/snowpark-python-2024

we no long claim to support 3.8 after v1.24.

but before removing the 3.8 pipelines we should get the team aware

PR #2835, opened 2025-01-08
SNOW-1794373: Support DataFrameWriter.insertInto/insert_into

sfc-gh-aling:

   

   Fixes SNOW-NNNNNNN

Implement DataFrameWriter.insert_into/insertInto which adds dataframe content to existing table requiring same schema;
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriter.insertInto.html

support in both live and local testing mode


src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -913,4 +914,45 @@ def parquet(
...
+            >>> df = session.create_dataframe([["John", "Berry"]], schema = ["FIRST_NAME", "LAST_NAME"])
+            >>> df.write.save_as_table("my_table", table_type="temporary")
+            >>> df2 = session.create_dataframe([["Rick", "Berry"]], schema = ["FIRST_NAME", "LAST_NAME"])
+            >>> df2.write.insert_into("my_table")
+            >>> session.table("my_table").collect()
+            [Row(FIRST_NAME='John', LAST_NAME='Berry'), Row(FIRST_NAME='Rick', LAST_NAME='Berry')]
+        """
+        full_table_name = (
+            table_name if isinstance(table_name, str) else ".".join(table_name)
+        )
```
sfc-gh-aalam, line 944: reminder to add a TODO to built ast for this API https://snowflakecomputing.atlassian.net/browse/SNOW-1489960

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -913,4 +914,45 @@ def parquet(
...
+        validate_object_name(full_table_name)
+        qualified_table_name = (
+            parse_table_name(table_name) if isinstance(table_name, str) else table_name
+        )
+
+        target_table = self._dataframe._session.table(qualified_table_name)
+        if target_table.schema != self._dataframe.schema:
+            raise SnowparkClientException(
+                f"Schema of the DataFrame: {self._dataframe.schema} does not match the schema of the table {full_table_name}: {target_table.schema}."
+            )
```
sfc-gh-aalam, line 954: I'm not sure if we should do this check on the client side. I think it is possible to append the data by type coercion even though the schemas are not an exact match.

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -913,4 +914,45 @@ def parquet(
...
+        """
+        full_table_name = (
+            table_name if isinstance(table_name, str) else ".".join(table_name)
+        )
+        validate_object_name(full_table_name)
+        qualified_table_name = (
+            parse_table_name(table_name) if isinstance(table_name, str) else table_name
+        )
+
+        target_table = self._dataframe._session.table(qualified_table_name)
```
sfc-gh-aalam, line 950: does it work in pyspark even when table does not exist? I think this would fail if table doesn't exist.

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-aling: added: https://snowflakecomputing.atlassian.net/browse/SNOW-1882075

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-aling: that's a good question about semantic alignment -- [pyspark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriter.insertInto.html) requires the **same schema**:
> It requires that the schema of the [DataFrame](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.html#pyspark.sql.DataFrame) is the same as the schema of the table.

I think we can be restrictive at the beginning (requiring same schema), and based upon user feedback relaxing the same schema requirement which is a non-breaking change.

wdyt?

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-aling: pyspark would fail.
The non-existing table error would be triggered by `if target_table.schema != self._dataframe.schema:` (this check serves both table check and schema check), I have a negative test for non-existing table

sfc-gh-jrose (approved):
LGTM. Only nit is that the pytest.raises could have some matches so verify the error message is expected.

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-aalam: I don't mind keeping it this way but our current `save_as_table` append mode does not do this check and rather relies on server side to make this decision.

Also in case of spark, their engine is closer to client so they can do this check on the client side. We can't do the same

sfc-gh-aalam (approved):

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-aling: I think you are right, it's better to be consistent with our own behavior.
updated the code to allow coercion

PR #2841, opened 2025-01-09
SNOW-1800374: adding support for options and partition_by to dataframewriter

sfc-gh-batur:

   

   Fixes SNOW-1800374

### DIFF
--- a/src/snowflake/snowpark/dataframe_writer.py
+++ b/src/snowflake/snowpark/dataframe_writer.py
@@ -16,6 +16,7 @@ from snowflake.snowpark._internal.analyzer.snowflake_plan_node import (
 )
 from snowflake.snowpark._internal.ast.utils import (
     build_expr_from_snowpark_column_or_col_name,
+    build_expr_from_snowpark_column_or_sql_str,
     debug_check_missing_ast,
     fill_sp_save_mode,
     fill_sp_write_file,
@@ -130,19 +131,33 @@ class DataFrameWriter:
 
         return self
 
-    def partition_by(self, expr: ColumnOrSqlExpr) -> "DataFrameWriter":
+    def partition_by(
+        self, expr: ColumnOrSqlExpr, _emit_ast: bool = True
+    ) -> "DataFrameWriter":
         """Specifies an expression used to partition the unloaded table rows into separate files. It can be a
         :class:`Column`, a column name, or a SQL expression.
         """
         self._partition_by = expr
+        # Update AST if it exists.
+        if _emit_ast:
+            if self._ast_stmt is not None:
+                build_expr_from_snowpark_column_or_sql_str(
+                    self._ast_stmt.expr.sp_dataframe_write.partition_by, expr
+                )
+
         return self
 
-    def option(self, key: str, value: Any) -> "DataFrameWriter":
+    def option(self, key: str, value: Any, _emit_ast: bool = True) -> "DataFrameWriter":
         """Depending on the ``file_format_type`` specified, you can include more format specific options.
         Use the options documented in the `Format Type Options <https://docs.snowflake.com/en/sql-reference/sql/copy-into-location.html#format-type-options-formattypeoptions>`__.
         """
         aliased_key = get_aliased_option_name(key, WRITER_OPTIONS_ALIAS_MAP)
         self._cur_options[aliased_key] = value
+        if _emit_ast:
+            if self._ast_stmt is not None:
+                t = self._ast_stmt.expr.options.add()
+                t._1 = aliased_key
+                t._2 = value
         return self
 
     def options(self, configs: Optional[Dict] = None, **kwargs) -> "DataFrameWriter":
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -130,19 +131,33 @@ def mode(self, save_mode: str, _emit_ast: bool = True) -> "DataFrameWriter":
...
 
-    def partition_by(self, expr: ColumnOrSqlExpr) -> "DataFrameWriter":
+    def partition_by(
+        self, expr: ColumnOrSqlExpr, _emit_ast: bool = True
+    ) -> "DataFrameWriter":
         """Specifies an expression used to partition the unloaded table rows into separate files. It can be a
         :class:`Column`, a column name, or a SQL expression.
         """
         self._partition_by = expr
+        # Update AST if it exists.
```
sfc-gh-lspiegelberg, line 141: the 'Update AST if it exists' sounds a bit weird to me, i.e. I think `_emit_ast` is simply doing what it means :) emitting the AST if true. Would suggest to remove the comment maybe altogether? We use `#AST.` sometimes in the code base (not perfect either, but like works).

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -130,19 +131,33 @@ def mode(self, save_mode: str, _emit_ast: bool = True) -> "DataFrameWriter":
...
+        if _emit_ast:
+            if self._ast_stmt is not None:
```
sfc-gh-lspiegelberg, line 143: maybe remove this if stmt? is it needed? should it be an assert instead?

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -130,19 +131,33 @@ def mode(self, save_mode: str, _emit_ast: bool = True) -> "DataFrameWriter":
...
-    def option(self, key: str, value: Any) -> "DataFrameWriter":
+    def option(self, key: str, value: Any, _emit_ast: bool = True) -> "DataFrameWriter":
         """Depending on the ``file_format_type`` specified, you can include more format specific options.
         Use the options documented in the `Format Type Options <https://docs.snowflake.com/en/sql-reference/sql/copy-into-location.html#format-type-options-formattypeoptions>`__.
         """
         aliased_key = get_aliased_option_name(key, WRITER_OPTIONS_ALIAS_MAP)
         self._cur_options[aliased_key] = value
+        if _emit_ast:
+            if self._ast_stmt is not None:
+                t = self._ast_stmt.expr.options.add()
```
sfc-gh-lspiegelberg, line 158: what happens in the case of `.option('k', 'v').option('k', 'v')` (repeated call with same key, same/different value?)

sfc-gh-lspiegelberg (commented):
Looks good to me, left some comments. Ping me once it is ready for merge for the final approval!

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-batur: Copy pasted from the mode function above. Happy to remove...

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-batur: again copy-pasted from the existing mode function above this one...

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-batur: dataframe_writer stores the values in a map -so I assume it'll overwrite the first value with the second. I'll add this to my test cases...

sfc-gh-lspiegelberg: yeah, I think the one challenge I see here is what we want to do in terms of unparsing/repro.

I.e, in the current modeling we basically have one dict to store the options, yet the APIs can be modified using

`.option(...).options(...).option(...)`

If we want to unparse code like this, I don't think we can do that atm. Yet, there's also an argument to be made that all of this `.option(...)...` chain can be folded into a single`.options(...)` call. Same for `.mode(...)`. I'll leave it to your call, but maybe we need to document this better. Wdyt?

sfc-gh-batur: A related issue. Say the user does:
.option(...).partition_by(...)
Depending upon how we write the unparser code - this can turn into:
.partition_by(...).option(...)
Semantically the two might be the same, but, it did surprise me to see this happen.

tests/ast/data/DataFrame.write.test
```diff
@@ -44,6 +50,12 @@ df.write.save_as_table("saved_table")
 
 df.write.mode("overwrite").save_as_table("saved_table", table_type="temporary")
```
sfc-gh-lspiegelberg, line 52: can we add some repeated calls of `.mode(...)`, `.option(...)`, `.options(...)`? Just to clarify behavior and preserve for the future.

sfc-gh-lspiegelberg (approved):
lgtm, just one ask/nit to clarify behavior for repeated calls.

sfc-gh-vbudati (approved):
Looks great. Thanks for adding more test cases as well!

sfc-gh-vbudati (approved):

PR #2849, opened 2025-01-10
SNOW-1877449:Exception should be thrown when create df with null value and nullable set to False

sfc-gh-yuwang:
…lable column

   Fixes SNOW-1877449

   This bug caused by ARRAY_BIND_THRESHOLD parameter in snowpark, we use it to decide whether we insert the data into temporary table when creating a dataframe. Currently the threshold is 512. So if row number * colum number <512, no insert query is executed, which means no error will be thrown when NULL value is in non-nullable column. This mean we had a false behavior before and some of the test needs to be fixed

when create a dataframe where number of columns * number of rows < 512, instead of inserting data into snowflake table, such query is executed:

SELECT "KEY" FROM ( SELECT $1 AS "KEY" FROM  VALUES (NULL :: INT)) LIMIT 10
this query would ignore all nullable settings on all columns, making snowpark have different behaviors when creating small and big dataframes.

This query cannot be modified so that it have the same behvior of inserting null value into a non-nullable table. So I add some assertions in our code to throw the same error, giving user consistent experiences.


src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -158,6 +158,19 @@ def is_large_local_data(self) -> bool:
...
+    def is_contain_illegal_null_value(self) -> bool:
+        from snowflake.snowpark._internal.analyzer.analyzer import ARRAY_BIND_THRESHOLD
+
+        rows_to_compare = min(
+            ARRAY_BIND_THRESHOLD // len(self.output) + 1, len(self.data)
+        )
+        for i in range(rows_to_compare):
+            for j in range(len(self.output)):
+                if self.data[i][j] is None and not self.output[j].nullable:
+                    return True
```
sfc-gh-jrose, line 171: Potentially a lot fewer comparisons if you restructured like this.
```suggestion
        for j in range(len(self.output)):
            if not self.output[j].nullable:
                for i in range(rows_to_compare):
                    if self.data[i][j] is None:
                        return True
```

sfc-gh-jrose (approved):

CHANGELOG.md
```diff
@@ -24,8 +24,10 @@
 #### Bug Fixes
 
 - Fixed a bug in local testing mode that caused a column to contain None when it should contain 0
+- Fixed a bug in StructField.from_json that prevented TimestampTypes with tzinfo from being parsed correctly.
 - Fixed a bug in `StructField.from_json` that prevented TimestampTypes with tzinfo from being parsed correctly.
```
sfc-gh-aalam, line 28: this needs a fix?

tests/integ/test_dataframe.py
```diff
@@ -1828,8 +1828,9 @@ def test_create_dataframe_with_schema_col_names(session):
         assert Utils.equals_ignore_case(field.name, expected_name)
 
     # the column names provided via schema keyword will overwrite other column names
+    struct_col_name = StructType([StructField(col, StringType()) for col in col_names])
     df = session.create_dataframe(
```
sfc-gh-aalam, line 1832: lets add a comment here for why this is needed.

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-yuwang: you are right, will change

CHANGELOG.md
sfc-gh-yuwang: ohh, this is issue from rebase, will change

PR #2858, opened 2025-01-13
[SNOW-1879403] Fix replace when passing `Column` object to avoid incorrect usage of `lit(...)` from within.

sfc-gh-lspiegelberg:

   Fixes SNOW-1879403

  `replace` supports both `Column` and literal objects. The code within the function directly called `lit(...)`, however `lit(...)` can't be called with a `Column` object. Adding if to fix this.


sfc-gh-aalam (commented):
I do not think this is the right approach. For example doing a cursory search, I found `uniform` could have the same problem. 

We removed the support for lit() to be safe from being given a column type with 1.26.0 which is a BCR. We agreed AST changes will not change current path unless ast changes are enabled. For now, we should go back to old implementation and re think on a better approach.

sfc-gh-lspiegelberg:
> I do not think this is the right approach. For example doing a cursory search, I found `uniform` could have the same problem.
> 
> We removed the support for lit() to be safe from being given a column type with 1.26.0 which is a BCR. We agreed AST changes will not change current path unless ast changes are enabled. For now, we should go back to old implementation and re think on a better approach.

Let's keep the scope of this PR narrow and concerned about the ticket filed. I'd like to get this fix in asap. We can figure a more general solution in a follow-up PR out.

sfc-gh-aalam (approved):

sfc-gh-jdu (approved):

PR #2862, opened 2025-01-14
SNOW-1871198: Add support for snowflake.snowpark.functions.from_json

sfc-gh-jrose:

   

   Fixes SNOW-1871198

This PR adds support for the `from_json` function. It works similarly to the pyspark equivalent with two notable exceptions:

1. pyspark allows pruning the schema of a json blob while it is parsed. Any fields defined in json that are not defined in the schema are dropped. Snowflake sql does not currently have equivalent functionality when parsing json so all fields need to be defined in the schema. The following example does not work at the moment:
```
>>> df.select(f.from_json(f.lit("{'a': 1, 'b': 2}"), schema=StructType([StructField("a", IntegerType())])).alias("out")).collect()
[Row(out=Row(a=1))]
```

2. snowpark does not have support for the Data Source Options that spark has so that parameter has been omitted from `from_json`.


sfc-gh-aalam (approved):
I'm happy with this change but I don't think this would be enough for SAS. It is worth checking with them that we only provide limited support

PR #2867, opened 2025-01-15
SNOW-1882151: Revert to old lit behavior

sfc-gh-aalam:

   

   Fixes SNOW-1882151

   This PR reverts back behavior of lit to support `Column` objects which was raising assertion errors and causing regressions to some customers.


sfc-gh-jrose (approved):
🎉

sfc-gh-jdu (approved):

src/snowflake/snowpark/functions.py
```diff
@@ -340,35 +340,36 @@ def column(name1: str, name2: Optional[str] = None, _emit_ast: bool = True) -> C
 
 @publicapi
 def lit(
-    literal: LiteralType, datatype: Optional[DataType] = None, _emit_ast: bool = True
+    literal: ColumnOrLiteral,
```
sfc-gh-lspiegelberg, line 343: this should be type hint `Any` to be compatible with PySpark.

src/snowflake/snowpark/functions.py
```diff
@@ -340,35 +340,36 @@ def column(name1: str, name2: Optional[str] = None, _emit_ast: bool = True) -> C
...
+        |     |     |     |      |                    |            |  2   |}                  |     |
+        |     |     |     |      |                    |            |]     |                   |     |
+        ---------------------------------------------------------------------------------------------
         <BLANKLINE>
     """
 
-    assert not isinstance(
-        literal, Column
-    ), "Do not use lit(Column(...)), type hint does not allow this syntax."
+    if isinstance(literal, Column):
```
sfc-gh-lspiegelberg, line 371: we need to add this case to IR/AST encoding. Right now, the repeated invocation of `lit` won't get reflected in AST. shouldn't you construct a new `Column` object out of the `lit` invocation? I.e., `Column(Literal(literal, ...))`?

sfc-gh-aalam: > shouldn't you construct a new Column object out of the lit invocation? I.e., Column(Literal(literal, ...))?

Old behavior as well as behavior in other libraries is to return the column as is in input object is a column. This is reflected in the updated documentation. see old behavior: https://github.com/snowflakedb/snowpark-python/blob/65581c71018c461165d95bd274e46805a2ee8e02/src/snowflake/snowpark/functions.py#L318

> we need to add this case to IR/AST encoding. Right now, the repeated invocation of lit won't get reflected in AST

Right now repeated invocation of `lit()` is redundant. If we want to capture it, we would have to create a new ticket.

src/snowflake/snowpark/functions.py
sfc-gh-aalam: Even though pyspark says it supports any, it raises TypeErrors for unsupported columns. In term of behavior, we are compatible

tests/ast/data/functions1.test
```diff
@@ -16,7 +16,7 @@ df05 = df.select(column("X", "A"))
```
sfc-gh-lspiegelberg, line 16: just to confirm, did you run update-expectations?

sfc-gh-lspiegelberg: AST merge gate seems to fail

src/snowflake/snowpark/functions.py
```diff
@@ -340,43 +340,48 @@ def column(name1: str, name2: Optional[str] = None, _emit_ast: bool = True) -> C
...
         <BLANKLINE>
     """
 
-    assert not isinstance(
-        literal, Column
-    ), "Do not use lit(Column(...)), type hint does not allow this syntax."
-
     ast = None
     if _emit_ast:
-        ast = proto.Expr()
```
sfc-gh-lspiegelberg, line 375: I just checked, the function `build_builtin_fn_apply` can actually take as argument a `Column` object (logic is in `build_fn_apply_args`). So I would revert the code change here, because basically when `lit` is invoked and `_emit_ast=True`, then always a new `proto.Expr()` object needs to be created.

src/snowflake/snowpark/functions.py
```diff
@@ -340,43 +340,48 @@ def column(name1: str, name2: Optional[str] = None, _emit_ast: bool = True) -> C
...
+        if isinstance(literal, Column):
+            ast = literal._ast
+        else:
+            ast = proto.Expr()
         if datatype is None:
             build_builtin_fn_apply(ast, "lit", literal)
         else:
             build_builtin_fn_apply(ast, "lit", literal, datatype)
+
+    if isinstance(literal, Column):
```
sfc-gh-lspiegelberg, line 382: I think it should not return the same column, but instead a new Column with the newly created AST?

sfc-gh-lspiegelberg: i.e., is it `Column(Literal(_literal_from_column(...)), _ast=ast)`?

tests/ast/data/functions1.test
```diff
@@ -16,7 +16,7 @@ df05 = df.select(column("X", "A"))
 
 df06 = df.select(column("A", None))
```
sfc-gh-lspiegelberg, line 18: what happens in the case `lit(col("A"))`? Could we add a test?

src/snowflake/snowpark/functions.py
sfc-gh-aalam: I don't understand this comment. Can you explain why

tests/ast/data/functions1.test
sfc-gh-aalam: I missed it.

CHANGELOG.md
```diff
@@ -53,8 +53,7 @@
 - Fixed a bug in local testing mode that caused a column to contain None when it should contain 0
 - Fixed a bug in `StructField.from_json` that prevented TimestampTypes with `tzinfo` from being parsed correctly.
 - Fixed a bug in function `date_format` that caused an error when the input column was date type or timestamp type.
-- Fixed a bug in dataframe that null value can be inserted in a non-nullable column.
```
sfc-gh-yuwang, line 56: I think you don't mean to remove this changelog right?

sfc-gh-aalam: oops. will fix that.

sfc-gh-yuwang (approved):

src/snowflake/snowpark/functions.py
sfc-gh-aalam:
```
     if isinstance(literal, Column):
         if _emit_ast:
             return Column(literal._expression, _ast=ast, _emit_ast=_emit_ast)
         return literal
```
I tried this and tests ast generation. This also doesn't encode `lit(lit())`. Any suggestions?

sfc-gh-lspiegelberg:
@sfc-gh-aalam I made a PR to support the lit(lit(...))/lit(col(...)) case. Please merge before merging this PR back to main. https://github.com/snowflakedb/snowpark-python/pull/2877

src/snowflake/snowpark/functions.py
```diff
@@ -341,43 +341,52 @@ def column(name1: str, name2: Optional[str] = None, _emit_ast: bool = True) -> C
...
     ast = None
     if _emit_ast:
         ast = proto.Expr()
         if datatype is None:
             build_builtin_fn_apply(ast, "lit", literal)
         else:
             build_builtin_fn_apply(ast, "lit", literal, datatype)
+
+    if isinstance(literal, Column):
+        if _emit_ast:
```
sfc-gh-lspiegelberg, line 381: you do not need this if here. If leads to unnecessary tech depth. Keep code simple!

src/snowflake/snowpark/functions.py
sfc-gh-lspiegelberg: @sfc-gh-aalam can we address this? Why not use `Any` as type hint?

sfc-gh-lspiegelberg (approved):
let's remove the unnecessary if, it's not needed. Good to go.

src/snowflake/snowpark/functions.py
sfc-gh-aalam: Because `ColumnOrLiteral` is the accurate type hint.

Pyspark's ref can be seen here: https://spark.apache.org/docs/3.5.0/api/python/_modules/pyspark/sql/functions.html#lit

src/snowflake/snowpark/functions.py
```diff
@@ -369,25 +369,28 @@ def lit(
...
-        if _emit_ast:
+        if isinstance(literal, Column):
             # Create new Column, and assign expression of current Column object.
             # This will encode AST correctly.
             c = Column("", _emit_ast=False)
             c._expression = literal._expression
             c._ast = ast
             return c
+        return Column(
+            Literal(literal, datatype=datatype), _ast=ast, _emit_ast=_emit_ast
```
sfc-gh-lspiegelberg, line 387: nit: can even do here _emit_ast=True

PR #2873, opened 2025-01-15
Simplify the implementation of the ast_enabled feature flag

sfc-gh-oplaton:

   

   Fixes SNOW-1874622

Previously, the logic that determined the value of `ast_enabled` tried to examine sessions present in process. The logic worked correctly for 0 or 1 sessions, but had bugs when more than one Snowpark session was active in process at a time. The transition from `enabled` to `disabled` and back to `enabled` was especially problematic. See the JIRA issue for more details.

The new functionality doesn't depend on sessions at all. Instead, a single global `ast_enabled` value exists, and the wrapper only allows safe transitions. Once the first session determines the value recommended by the server, the feature state freezes and cannot change back and forth anymore.

We can add more client-side policies if they ever become necessary.


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -801,153 +801,183 @@ def warning(name: str, text: str, warning_times: int = 1) -> None:
...
+    The enum value determines the precedence of the value.
+    """
+
+    LOCAL = auto()
+    """Some local criteria determined the value. This has the lowest precedence. Any other source can override it."""
+    SERVER = auto()
+    """The server set the value. This has the highest precedence. Nothing else can override it."""
+    TEST = auto()
+    """
+    Do not use this in production Snowpark code. Test code set the value. This has the highest precedence.
```
sfc-gh-azwiegincew, line 817: Typo: "Test code SETS the value."

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -801,153 +801,183 @@ def warning(name: str, text: str, warning_times: int = 1) -> None:
...
+                # Treat TEST as SERVER:
+                #   Only one place in the production code uses SERVER. Test code shouldn't use that enum value.
+                #   Treat TEST requests as final. No other test or production code can change the state that test setup code requested.
+                #   If cases where treating TEST different from SERVER come up, we can augment the logic below.
+                canonical_source = AstFlagSource.SERVER
+                _logger.info("Treating source = %s as %s", source, canonical_source)
+                source = canonical_source
+            if self._state == _AstFlagState.FINALIZED and self._ast_enabled != enable:
+                _logger.warning(
+                    "Cannot change AST state after it has been finalized. Frozen ast_enabled = %s. Ignoring value %s",
```
sfc-gh-azwiegincew, line 905: All these warnings are missing the trailing period.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -801,153 +801,183 @@ def warning(name: str, text: str, warning_times: int = 1) -> None:
...
+            else:
+                raise NotImplementedError(
+                    f"Unhandled transition. Current state: ast_enabled = {self._ast_enabled}, state = {self._state}. Request: source = {source}, enable = {enable}"
+                )
+
+
+_ast_state: _AstState = _AstState()
+
+
+def get_ast_enabled() -> bool:
```
sfc-gh-azwiegincew, line 940: I get it's a matter of style, but I'd suggest `is_ast_enabled` per Google C++ style.

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/_internal/utils.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/_internal/utils.py
sfc-gh-oplaton: Done.

src/snowflake/snowpark/session.py
```diff
@@ -797,16 +801,16 @@ def ast_enabled(self, value: bool) -> None:
         #     )
         # except Exception:
         #     pass
-        self._ast_enabled = value
```
sfc-gh-lspiegelberg, line 804: There's quite some logic buried in utils with the new helper class. Ultimately, I think the idea is that if the process starts with the flag as True, a user can flip it to False. However, then going back to True doesn't work. In a sense, the setter here could have a logic like:
```
if self._ast_enabled is False and value is True:
   raise ValueError("Can not enable AST mid-session...")
```
I am not a fan of replicating the code and introducing redundancy here, but I feel this behavior should be somewhere documented. E.g., either by adding
- a docstring here
- updating docs (this may be anyways a good idea with the new param introduced).

tests/unit/test_udf.py
```diff
@@ -18,14 +22,16 @@
 
 @mock.patch("snowflake.snowpark.udf.cleanup_failed_permanent_registration")
 def test_do_register_sp_negative(cleanup_registration_patch):
+    AST_ENABLED = False
```
sfc-gh-lspiegelberg, line 25: we could refactor lines 25-28 into a helper that gets used across the tests. There seems to be a lot of duplication.

sfc-gh-lspiegelberg (approved):
Looks great! Really like the narrowing down by disallowing the flip/flopping use case.

src/snowflake/snowpark/session.py
sfc-gh-oplaton: Note that `Session._ast_enabled` is gone - we want a single behavior per client, and individual sessions can't pick and choose.

`safe_transition` in [utils.py](https://github.com/snowflakedb/snowpark-python/blob/93d5748724de80330a39f2b6551622b9011dc0d9/src/snowflake/snowpark/_internal/utils.py#L910-L914) is the central point for this exact logic.

Adding a docstring, and will look into updating documentation, as well.

tests/unit/test_udf.py
sfc-gh-oplaton: `set_ast_state` should have sufficed. It's only the side effect of `Session.ast_enabled` that requires the duplication. Could probably do something about that, but it seemed relatively minor.

src/snowflake/snowpark/session.py
sfc-gh-lspiegelberg: sounds great, it may be also good to check/verify that `session.ast_listener()` throws some meaningful warning/action to take with the logic changes.

sfc-gh-oplaton: I added a docstring. I think we shouldn't document `Session.ast_enabled` beyond that, since it's a transitional property, and users can't do much with it.

sfc-gh-oplaton:
I need approval from one of the code owners, please!

@sfc-gh-smirzaei , @sfc-gh-aling , @sfc-gh-jrose

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -801,153 +801,183 @@ def warning(name: str, text: str, warning_times: int = 1) -> None:
...
+                # TEST behaviors override everything.
+                # If you see this code path running in production, the calling code is broken.
+                self._state = _AstFlagState.FINALIZED
+                self._ast_enabled = enable
+                return
+            if self._state == _AstFlagState.FINALIZED and self._ast_enabled != enable:
+                _logger.warning(
+                    "Cannot change AST state after it has been finalized. Frozen ast_enabled = %s. Ignoring value %s.",
+                    self._ast_enabled,
+                    enable,
```
sfc-gh-aalam, line 907: minor nits: can we use f-strings in all these logging statements

sfc-gh-aalam (approved):
Looks good. One minor nit. Thanks for the fix Ovidiu

src/snowflake/snowpark/_internal/utils.py
sfc-gh-oplaton: I don't mind making the change, since it's trivial, but I'm curious about one thing: I had the impression that, in Python, as in several other languages, logging statements generally use format strings because it allows them to delay turning objects into strings until they're sure the string is required.

Basically, `_logger.debug("foo = %s, bar = %s", foo, bar)` might avoid calling `str(foo)` and `str(bar)` altogether, whereas `_logger.debug(f"foo = {foo}, bar = {bar}")` formats one string eagerly, and calls `debug()`, which might discard the request altogether.

Performance won't matter in this particular case, but do we have a stance on this?

sfc-gh-jrose: I think I agree with @sfc-gh-oplaton. It's generally better style to let the logger figure out if the objects need to be rendered into strings.

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/utils.py
sfc-gh-oplaton: Alright, I'll merge with the format string in place. It's relatively mechanical to change it one way or another. Thanks!

PR #2877, opened 2025-01-16
[AST FIX] functions.py modification for lit(lit(lit(....)))

sfc-gh-lspiegelberg:
Changes to allow for proper lit(lit(...)) AST behavior.

### DIFF
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -377,9 +377,16 @@ def lit(
             build_builtin_fn_apply(ast, "lit", literal, datatype)
 
     if isinstance(literal, Column):
-        return literal
-
-    return Column(Literal(literal, datatype=datatype), _ast=ast, _emit_ast=_emit_ast)
+        # Create new Column, and assign expression of current Column object.
+        # This will encode AST correctly.
+        c = Column("", _emit_ast=False)
+        c._expression = literal._expression
+        c._ast = ast
+        return c
+    else:
+        return Column(
+            Literal(literal, datatype=datatype), _ast=ast, _emit_ast=_emit_ast
+        )
 
 
 @publicapi
--- a/tests/ast/data/col_lit.test
+++ b/tests/ast/data/col_lit.test
@@ -1,0 +1,443 @@
+## TEST CASE
+
+import datetime
+
+df = session.table(tables.table1)
+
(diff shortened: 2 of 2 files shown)


src/snowflake/snowpark/functions.py
```diff
@@ -377,9 +377,16 @@ def lit(
             build_builtin_fn_apply(ast, "lit", literal, datatype)
 
     if isinstance(literal, Column):
-        return literal
-
-    return Column(Literal(literal, datatype=datatype), _ast=ast, _emit_ast=_emit_ast)
+        # Create new Column, and assign expression of current Column object.
+        # This will encode AST correctly.
+        c = Column("", _emit_ast=False)
```
sfc-gh-jrose, line 382: I think the new column should have the same name as the old column.

sfc-gh-lspiegelberg: I think it does, because the `._expression` member is copied over. The constructor here is not called for correctness, but rather how the current API works. `Column` has as far as I see it two members:
`._expression` and `._ast`. Both will be set, so should be fine.

sfc-gh-jrose: Ah, you are correct. Thanks for explaining.

sfc-gh-jrose (approved):

src/snowflake/snowpark/functions.py
sfc-gh-aalam, line 382: just to be safe, can we enable this code path only when `_emit_ast=True`

src/snowflake/snowpark/functions.py
```diff
@@ -377,9 +377,16 @@ def lit(
...
+        c._expression = literal._expression
+        c._ast = ast
+        return c
+    else:
+        return Column(
```
sfc-gh-aalam, line 387: I prefer if we don't have the else here since the if statement always returns

src/snowflake/snowpark/functions.py
sfc-gh-lspiegelberg: which concern do you have? `Column("", _emit_ast=False)` should be side-effect free code, right?

sfc-gh-aalam: the only concern is to stay on a old code path. All ast related code changes should follow original codepath when `_emit_ast=False`

sfc-gh-aalam, line 382: also why is `_emit_ast=false` in `Column("", _emit_ast=False)`?

Issue #2888, opened 2025-01-20
SNOW-1793953: Allow relative path starting with / for GET


sfc-gh-tvanderkooy:
In native apps, GET supports a relative path to fetch files from the application version. For example:
```sql
GET /manifest.yml file:///to/some/local/path
```

Currently `file.get("/some/file")` will preprend `@` to the path and try to `GET '@/some/file'` which is never a valid stage path. We should allow `/` as a path prefix to support this case.

PR #1346, opened 2024-03-29
Add support for `snow://` stage paths

sfc-gh-dhung:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #1345

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

Adds support for `snow://` prefixed stage paths. Also fixes a bug where local file path is incorrectly extracted from single quoted paths: `'@path/file'` -> `file'` (note trailing quote)

PR #2470, opened 2024-10-17
SNOW-1300193: Add error message when trying to read from local directory

sfc-gh-jrose:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-1300193

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.
   - [x] I acknowledge that I have ensured my changes to be thread-safe. Follow the link for more information: [Thread-safe Developer Guidelines](https://docs.google.com/document/d/162d_i4zZ2AfcGRXojj0jByt8EUq-DrSHPPnTa4QvwbA/edit#bookmark=id.e82u4nekq80k)

3. Please describe how your code solves the related issue.

   This change adds an error message for when DataframeReader is passed a non-stage path. The SQL error for this case can be confusing to users so having a more direct error is desirable.

PR #2889, opened 2025-01-20
accept stage paths starting with / for GET

sfc-gh-tvanderkooy:

   

   Fixes SNOW-1793953 (https://github.com/snowflakedb/snowpark-python/issues/2888)

   This builds on the changes from #1346 to allow `/` as a remote path prefix so `file.get('/some/path')`will not try to prepend `@` to the path. This form of relative path is valid in some native application uses cases.

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -65,9 +65,11 @@ _logger = logging.getLogger("snowflake.snowpark")
 
 STAGE_PREFIX = "@"
 SNOWURL_PREFIX = "snow://"
+RELATIVE_PATH_PREFIX = "/"
 SNOWFLAKE_PATH_PREFIXES = [
     STAGE_PREFIX,
     SNOWURL_PREFIX,
+    RELATIVE_PATH_PREFIX,
 ]
 
 # Scala uses 3 but this can be larger. Consider allowing users to configure it.
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -69,6 +69,8 @@ RELATIVE_PATH_PREFIX = "/"
 SNOWFLAKE_PATH_PREFIXES = [
     STAGE_PREFIX,
     SNOWURL_PREFIX,
+]
+SNOWFLAKE_PATH_PREFIXES_FOR_GET = SNOWFLAKE_PATH_PREFIXES + [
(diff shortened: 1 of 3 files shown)


sfc-gh-skun (approved):

sfc-gh-mxu (approved):

sfc-gh-yb (approved):

sfc-gh-aalam:
@sfc-gh-tvanderkooy can you fix the failing tests please. change looks good

sfc-gh-tvanderkooy:
Turns out the test failures were because my change breaks #2470, which adds an explicit error for using a non-stage path with `session.read.<file_format>(<path)`. These functions reads files from a stage that are some table format (avro, csv, etc.) into a data frame, and the error is to help users that are trying to use these functions with local files (which doesn't work).

I think it's unlikely for native app providers to want to use this in their app -- any file that is valid for these functions they could simply share as a table instead of putting as a file in the version stage -- so I opted to NOT support the relative stage path syntax for this case and preserve the existing error message.

cc @sfc-gh-jrose @sfc-gh-aalam

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -372,7 +376,7 @@ def normalize_path(path: str, is_local: bool) -> str:
     a directory named "load data". Therefore, if `path` is already wrapped by single quotes,
     we do nothing.
     """
-    prefixes = ["file://"] if is_local else SNOWFLAKE_PATH_PREFIXES
+    prefixes = ["file://"] if is_local else SNOWFLAKE_PATH_PREFIXES_FOR_GET
```
sfc-gh-aalam, line 379: this change adds support when run inside a native apps environment. How does it affect when run outside NA? If it is not supported, then we should only allow the new prefix in NA env.

sfc-gh-tvanderkooy: Outside native apps this will be a SQL error. `SQL compilation error: invalid URL prefix found in: '/some/path'`

 But that's the current behaviour too (for GET) - today if you do `session.file.get('/some/path')`, snowpark will prepend `@` and run the GET, and then throw `SQL compilation error: missing stage name in URL: @/some/path` 

Unfortunately I'm not aware of a good way to check if this is NA env, it might be difficult to check that here.

sfc-gh-aalam (approved):

PR #2906, opened 2025-01-22
NO-SNOW: Fix stored proc precommit tests

sfc-gh-jrose:

   

   Fixes NO-SNOW

   The majority of these tests were failing due to results being in the wrong order. The operations in changed tests do not have a guaranteed order so I either moved them to use Utils.check_answer which sorts the data, or updated their sort statements to include enough columns to have a composite primary key. There was also a test that was trying to test structured type logic in an environment that does not support them.

### DIFF
--- a/tests/integ/scala/test_dataframe_aggregate_suite.py
+++ b/tests/integ/scala/test_dataframe_aggregate_suite.py
@@ -538,14 +538,14 @@ def test_group_by_grouping_sets(session):
         TestData.nurse(session)
         .group_by("medical_license", "radio_license")
         .agg(count(col("*")).as_("count"))
-        .sort(col("count"))
+        .sort(col("count"), col("medical_license"), col("radio_license"))
         .select("count", "medical_license", "radio_license"),
         [
             Row(1, "LVN", "General"),
-            Row(1, "RN", "Amateur Extra"),
             Row(1, "RN", None),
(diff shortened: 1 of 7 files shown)


tests/integ/scala/test_dataframe_aggregate_suite.py
```diff
@@ -538,14 +538,14 @@ def test_group_by_grouping_sets(session):
         TestData.nurse(session)
         .group_by("medical_license", "radio_license")
         .agg(count(col("*")).as_("count"))
-        .sort(col("count"))
+        .sort(col("count"), col("medical_license"), col("radio_license"))
```
sfc-gh-aling, line 541: unrelated to this PR, I'm just curious does spark give a deterministic column order when column order is not explicitly set by users

sfc-gh-aling (approved):

tests/integ/scala/test_dataframe_aggregate_suite.py
sfc-gh-jrose: No, dataframe operations are split into tasks that are executed by one or more worker nodes. Without an explicit ordering I don't think the results from each worker are sorted so it depends on the order you receive the data.

sfc-gh-aalam (approved):
Thanks for fixing these tests Jamison 🙌

PR #2912, opened 2025-01-22
SNOW-1878372: Fix analyzer access across threads

sfc-gh-aalam:

   

   Fixes SNOW-1878372

   In this change, we make the following changes to `Selectable`:
   - the analyzer object is accessed via a thread so that each selectable uses the analyzer created for the dedicated thread.
   - the analyzer can be set to a different analyzer only during optimization stage when `_is_valid_for_replacement=True`.

  We also update `Analyzer` to initialize `alias_maps_to_use` with an empty dict `{}` instead of `None`. In single threaded-case, `resolve` would initialize alias_maps_to_use to `{}` and then use it for `analyze` but in multithreaded-case, it is possible for one thread to call `resolve` and a different thread to call `analyze`.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -229,8 +229,10 @@ class Selectable(LogicalPlan, ABC):
         ] = None,  # Use Any because it's recursive.
     ) -> None:
         super().__init__()
-        # Due to multi-threading changes, we need to use the session object to
-        # access the analyzer created for current thread.
+        # With multi-threading support, each thread has its own analyzer which can be
+        # accessed through session object. Therefore, we need to store the session in
+        # the Selectable object and use the session to access the appropriate analyzer
+        # for current thread.
         self._session = analyzer.session
         self.pre_actions: Optional[List["Query"]] = None
         self.post_actions: Optional[List["Query"]] = None
--- a/src/snowflake/snowpark/_internal/analyzer/select_statement.py
+++ b/src/snowflake/snowpark/_internal/analyzer/select_statement.py
@@ -234,6 +234,9 @@ class Selectable(LogicalPlan, ABC):
         # the Selectable object and use the session to access the appropriate analyzer
         # for current thread.
         self._session = analyzer.session
+        # We create this internal object to be used for setting query generator during
+        # the optimization stage
+        self._analyzer = None
         self.pre_actions: Optional[List["Query"]] = None
         self.post_actions: Optional[List["Query"]] = None
         self.flatten_disabled: bool = False
@@ -250,7 +253,21 @@ class Selectable(LogicalPlan, ABC):
     @property
     def analyzer(self) -> "Analyzer":
         """Get the analyzer for used for the current thread"""
-        return self._session._analyzer
+        if self._analyzer is None:
+            return self._session._analyzer
+        return self._analyzer
+
+    @analyzer.setter
+    def analyzer(self, value: "Analyzer") -> None:
+        """For query optimization stage, we need to replace the analyzer with a query generator which
+        is aware of schema for the final plan and can compile WithQueryBlocks. Therefore we update the
+        setter to allow the analyzer to be set externally."""
+        if not self._is_valid_for_replacement:
+            raise ValueError(
+                "Cannot set analyzer for a Selectable that is not valid for replacement"
+            )
+
+        self._analyzer = value
 
     @property
     @abstractmethod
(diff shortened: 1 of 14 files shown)


tests/unit/test_query_plan_analysis.py
```diff
@@ -121,7 +121,7 @@ def test_selectable_entity_individual_node_complexity(mock_session, mock_analyze
     assert plan_node.individual_node_complexity == {PlanNodeCategory.COLUMN: 1}
 
 
-def test_select_sql_individual_node_complexity(mock_analyzer):
+def test_select_sql_individual_node_complexity(mock_session, mock_analyzer):
```
sfc-gh-jrose, line 124: What does adding mock_session do for these tests?

sfc-gh-aalam: that's a good question, calling mock session here sets `mock_analyzer.session=mock_session`. 

https://github.com/snowflakedb/snowpark-python/blob/7d4b43ce51a81afe8f31875e80bb1253d81bd9f5/tests/unit/conftest.py#L75-L82

this is probably not the best way to do it. open to suggestions.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -243,6 +250,25 @@ def __init__(
...
+    def analyzer(self, value: "Analyzer") -> None:
+        """For query optimization stage, we need to replace the analyzer with a query generator which
+        is aware of schema for the final plan and can compile WithQueryBlocks. Therefore we update the
+        setter to allow the analyzer to be set externally."""
+        if not self._is_valid_for_replacement:
+            raise ValueError(
+                "Cannot set analyzer for a Selectable that is not valid for replacement"
+            )
+
+        self._analyzer = value
```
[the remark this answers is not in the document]
sfc-gh-aalam: I did a codebase search and only time we re-assign analyzer is during compilation stage. That's why I added this check here so that this pattern is not easily done for future changes.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -243,6 +250,25 @@ def __init__(
         self._cumulative_node_complexity: Optional[Dict[PlanNodeCategory, int]] = None
         self._encoded_node_id_with_query: Optional[str] = None
 
+    @property
+    def analyzer(self) -> "Analyzer":
+        """Get the analyzer for used for the current thread"""
+        if self._analyzer is None:
```
sfc-gh-jdu, line 256:
```suggestion
        return self._analyzer or self._session._analyzer
```

sfc-gh-jdu (approved):

tests/unit/test_cte.py
```diff
@@ -84,7 +84,7 @@ def test_find_duplicate_subtrees(test_case):
     assert repeated_node_complexity == expected_repeated_node_complexity
 
 
-def test_encode_node_id_with_query_select_sql(mock_analyzer):
+def test_encode_node_id_with_query_select_sql(mock_session, mock_analyzer):
```
sfc-gh-aling, line 87: is mock_session used in this test?

sfc-gh-aalam: same as jamison's question. Linking old response: https://github.com/snowflakedb/snowpark-python/pull/2912#discussion_r1927517894

tests/unit/compiler/test_replace_child_and_update_node.py
```diff
@@ -455,7 +455,8 @@ def test_select_statement(
     new_replaced_plan = plan.children_plan_nodes[0]
     assert isinstance(new_replaced_plan, SelectSnowflakePlan)
     assert new_replaced_plan._snowflake_plan.source_plan == new_plan
-    assert new_replaced_plan.analyzer == mock_query_generator
+    # new_replaced_plan is created with QueryGenerator.to_selectable
+    assert new_replaced_plan.analyzer == mock_analyzer
```
sfc-gh-aling, line 459: is this change to pass the new "setter" check?

sfc-gh-aling (approved):

tests/unit/compiler/test_replace_child_and_update_node.py
sfc-gh-aalam: `new_replaced_plan` is created by calling within `QueryGenerator.to_selectable` which uses `self` as an analyzer. 

Before this change `Selectable.analyzer` would have returned the analyzer used to create the selectable i.e. `mock_query_generator`.

After this change, `Selectable.analyzer` will return the appropriate `session.analyzer`. Therefore, we need to update this test.

PR #2914, opened 2025-01-23
SNOW-1890740: Support more aggregation functions in pivot methods.

sfc-gh-mvashishtha:
Fixes SNOW-1890740

Added support for aggregations ``"count"``, ``"median"``, ``np.median``,
  ``"skew"``, ``"std"``, ``np.std`` ``"var"``, and ``np.var`` in
  ``pd.pivot_table()``, ``DataFrame.pivot_table()``, and ``pd.crosstab()``.


docs/source/modin/supported/agg_supp.rst
```diff
@@ -3,64 +3,56 @@
 Supported Aggregation Functions
 ====================================
 
-This page lists which aggregation functions are supported by ``DataFrame.agg``,
```
sfc-gh-mvashishtha, line 6: These descriptions are redundant, so I'm deleting them.

PR #2918, opened 2025-01-23
SNOW-1893703: Require SQL counter for local Snowpark pandas integration tests.

sfc-gh-mvashishtha:
Prior to this commit, we required SQL count checks in Github Actions
but not when running tests locally or on Jenkins. I have been surprised
many times to see my tests pass locally and fail in GitHub Actions
because they're missing sql count checks.

In this commit, require the count checks for local test runs, but leave
them as not required for Jenkins tests.

Fixes SNOW-1893703


sfc-gh-nkrishna (approved):
This is great, thanks!

sfc-gh-joshi (approved):

PR #2919, opened 2025-01-23
NO-SNOW: Add modin as dependency for AST tests.

sfc-gh-jrose:

   

   Fixes NO-SNOW

   One of the AST tests has been failing in daily-precommits due to requiring modin. Modin was not a tox dependency so I've added it.


sfc-gh-lspiegelberg:
Can you point out which test is failing due to Modin? Modin doesn't work with python 3.8, so adding it as a dependency may break those gates immediately.

tox.ini
```diff
@@ -60,7 +60,9 @@ setenv =
     unit: SNOWFLAKE_TEST_TYPE = unit
     integ: SNOWFLAKE_TEST_TYPE = integ
     doctest: SNOWFLAKE_TEST_TYPE = doctest
-    ast: SNOWFLAKE_TEST_TYPE = ast
```
sfc-gh-lspiegelberg, line 63: ast should not have a dependency on modin, we do not use anything from it for the AST tests to the best of my knowledge. Do you have a link/CI run to point why this fix is needed?

sfc-gh-jrose: Here's an example failure: https://github.com/snowflakedb/snowpark-python/actions/runs/12926763404/job/36050719429

sfc-gh-lspiegelberg: Thank you for the pointers. What baffles me that this fails, I don't recall the tests failing when the code was checked in. Did something change? To me the snowpark_pandas test is an integration test with the snowpark pandas api, it may make sense to separate these types of tests out. Quickly adding modin may fix this, but I am concerned about 3.8 (despite its EOL). Let's involve @sfc-gh-vbudati here, given she authored the originally code.

@sfc-gh-vbudati: Do you have an idea/suggestion how we could avoid adding modin as dependency and still test the to_snowpark_pandas API? Maybe we can convert this to an integration test and run only with python3.9+?

sfc-gh-vbudati: To my knowledge, the test was working when it was checked in. The error here (https://github.com/snowflakedb/snowpark-python/actions/runs/12926763404/job/36050719429) looks like it's because modin is not being imported. 

@sfc-gh-lspiegelberg the `to_snowpark_pandas.test` won't run unless the Python version is 3.9+ (look at `load_test_cases()` in `test_ast_driver.py`. So it should be fine to import modin after a check to see that the Python version is 3.9+.

PR #2923, opened 2025-01-23
SNOW-1893939: Add support for Index.drop_duplicates

sfc-gh-amaheshwary:

   Fixes [SNOW-1893939](https://snowflakecomputing.atlassian.net/browse/SNOW-1893939)

   Added support for Index.drop_duplicates.

[SNOW-1893939]: https://snowflakecomputing.atlassian.net/browse/SNOW-1893939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfc-gh-amaheshwary:

tests/integ/modin/index/test_drop_duplicates.py
```diff
@@ -0,0 +1,67 @@
...
+def test_drop_duplicates_nan_none(keep, expected):
+    # Note that Snowpark pandas treats np.nan and None the same
+    idx = pd.Index([np.nan, 3, 3, None, np.nan], dtype=object)
+
+    result = idx.drop_duplicates(keep=keep)
+    assert_index_equal(
+        result,
+        expected,
+    )
+
```
sfc-gh-nkrishna, line 57: Can you also add a test where keep is not specified explicitly (testing the default value of `first`)? I see it in the code but having a test is good too.

tests/integ/modin/index/test_drop_duplicates.py
```diff
@@ -0,0 +1,67 @@
...
+@sql_count_checker(query_count=1, join_count=2)
+def test_drop_duplicates_nan_none(keep, expected):
+    # Note that Snowpark pandas treats np.nan and None the same
+    idx = pd.Index([np.nan, 3, 3, None, np.nan], dtype=object)
+
+    result = idx.drop_duplicates(keep=keep)
+    assert_index_equal(
+        result,
+        expected,
+    )
```
sfc-gh-nkrishna, line 56: Can you also test a value of `keep` that is invalid? Ex. keep = "invalid" just to ensure we error correctly in the same manner as pandas?

sfc-gh-nkrishna (commented):
Some minor suggestions around adding more tests, but LGTM overall

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -533,10 +533,8 @@ def drop(
         # TODO: SNOW-1458146 implement drop
         pass  # pragma: no cover
 
-    @index_not_implemented()
-    def drop_duplicates(self) -> None:
-        # TODO: SNOW-1458147 implement drop_duplicates
-        pass  # pragma: no cover
+    def drop_duplicates(self, keep="first") -> None:
+        return self.to_series().drop_duplicates(keep=keep).index
```
sfc-gh-jjiao, line 537: naive question: is this operation being executed locally in client or it will also be lazily evaluation and been pushed down eventually?

sfc-gh-nkrishna: The `.index` part of this eagerly pulls the data. @sfc-gh-amaheshwary you should try call the query compiler's `drop_duplicates` instead of using `to_series`:
```
return self.__constructor__(
            query_compiler=self._query_compiler.copy(), name=name
        )
```
This above is for copy, you'll have to modify accordingly and pass the arguments for `drop_duplicates` correctly

sfc-gh-joshi: I believe the index constructor lets you take a Series directly, so you could keep it as `self.__constructor__(self.to_series().drop_duplicates(keep=keep))` without any materialization.

sfc-gh-jkew: It's not a totally naive question; for a long time .index would trigger a materialization of the index. Eventually this was replaced by a lazy index as described here: 

https://github.com/snowflakedb/snowpark-python/blob/249273fe7a59125dc09834738822ece2bbd43e8c/docs/source/modin/supported/index_supported.rst

It's worth double checking that no queries are run until the asserts occur in the tests though

sfc-gh-nkrishna: That should work, but it would still be an extra API call (i.e. the `to_series` is extraneous?)

tests/integ/modin/index/test_drop_duplicates.py
```diff
@@ -0,0 +1,67 @@
...
+
+@pytest.mark.parametrize(
+    "keep, expected",
+    [
+        ("first", native_pd.Index([np.nan, 3])),
+        ("last", native_pd.Index([3, np.nan])),
+        (False, native_pd.Index([], dtype="float64")),
+    ],
+)
+@sql_count_checker(query_count=1, join_count=2)
```
sfc-gh-jkew, line 47: It's a little surprising to me that this takes two joins. I know you are using series.drop_duplicates to do this but that also seems expensive for series. I assume that the reason for the joins is because we want to maintain ordering of the series/index and a simple UNIQUE/DISTINCT will not work, correct? That's crazy expensive for something users would expect to be cheap.

As a sort of meta-question I wonder if we should extend the api to drop_duplicates to allow order-free execution. That's not a question to ask today, but maybe @sfc-gh-helmeleegy has thoughts.

src/snowflake/snowpark/modin/plugin/extensions/index.py
sfc-gh-joshi: Yeah, calling the query compiler directly should be marginally better (it looks like `to_series` has some slight overhead for dealing with the `name` of the resulting series, which might result in some extra query text.)

src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -533,10 +533,10 @@ def drop(
         # TODO: SNOW-1458146 implement drop
         pass  # pragma: no cover
 
-    @index_not_implemented()
-    def drop_duplicates(self) -> None:
-        # TODO: SNOW-1458147 implement drop_duplicates
-        pass  # pragma: no cover
+    def drop_duplicates(self, keep="first") -> None:
+        if keep not in ("first", "last", False):
+            raise ValueError('keep must be either "first", "last" or False')
```
sfc-gh-nkrishna, line 538: Would you mind making this change in `base_overrides.py::drop_duplicates` as well for the Series/DataFrame implementation since you're already at that point in the codebase?

sfc-gh-nkrishna, line 538: Nit: quotes
```suggestion
            raise ValueError("keep must be either 'first', 'last' or 'False'")
```

tests/integ/modin/index/test_drop_duplicates.py
```diff
@@ -0,0 +1,75 @@
...
+        pandas_idx.drop_duplicates(),
+    )
+
+
+@sql_count_checker(query_count=0, join_count=0)
+def test_drop_duplicates_invalid_keep():
+    snow_idx = pd.Index(["a", "b", "b", "c", "a"], name="name")
+    with pytest.raises(ValueError) as ex_info:
+        snow_idx.drop_duplicates(keep="invalid")
+    assert 'keep must be either "first", "last" or False' in str(ex_info)
```
sfc-gh-nkrishna, line 75: I believe this should work, though I might have messed up some the escape quotes. It's a little bit cleaner to pass the matching string as a parameter to pytest than capturing the string and asserting afterwards:
```suggestion
    with pytest.raises(ValueError, match="keep must be either \'first\', \'last\' or \'False\'"):
        snow_idx.drop_duplicates(keep="invalid")
```

sfc-gh-jjiao (approved):
Great job for your first PR!

tests/integ/modin/index/test_drop_duplicates.py
sfc-gh-helmeleegy: The thing is that pandas drop_duplicates is more sophisticated than sql's select distinct. It allows you to choose whether to drop all duplicates, or keep just the first occurrence or last occurrence in each set of duplicate rows. But note that duplication is determined based on the data columns not the index columns. So the index columns can still be different, and hence specifying `keep='first'` will return a different result than `keep='last'`. Ensuring this behavior does require joins unfortunately. 

But there might be room for reducing the join count by 1 as today `drop_duplicates` is implemented on top of `duplicated`, such that a new join is added to the one join already used for `duplicated`. If `drop_duplicates` was implemented from scratch, it might be possible to avoid the second join.

PR #2932, opened 2025-01-24
Disable couple more slow test suites on github

sfc-gh-yzou:
Further reduce some slow tests suite


sfc-gh-helmeleegy (approved):

sfc-gh-jjiao:
Are the disablement temporary?  if we do not fix the root cause of the timeout,  I suspect we will have timeout again soon when people adding more new tests

sfc-gh-jjiao (approved):

PR #2943, opened 2025-01-27
SNOW-1856438: Directly support Snowflake Cortex functions Summarize and Sentiment with apply

sfc-gh-lmukhopadhyay:

   

   Fixes SNOW-1856438

   Supporting Snowflake Cortex LLM APIS Summarize and Sentiment  with df/ series apply

  Example:
```
 from snowflake.cortex import Sentiment

 content = "This is a really great sentence."
 s = pd.Series([content])
 s.apply(Sentiment).iloc[0]
 >>> 0.872
```


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8786,6 +8802,27 @@ def sf_function(col: SnowparkColumn) -> SnowparkColumn:
...
+    ) -> "SnowflakeQueryCompiler":
+        """Apply Snowflake Cortex function to columns."""
+
+        def sf_function(col: SnowparkColumn) -> SnowparkColumn:
+            if not kwargs:
+                return snowflake_function(col)
+            resolved_positional = []
+            resolved_positional.append(col)
+            for arg in kwargs:
+                resolved_positional.append(kwargs[arg])
```
sfc-gh-mvashishtha, line 8818: This code seems to assume that the kwargs come in the order of the arguments in the siginature `CLASSIFY_TEXT`, but that might not be true. e.g. the user may call `df.apply(ClassifyText, task_description="classify tweets", categories="["travel", "cooking"])` or `df.apply(ClassifyText, categories="["travel", "cooking"], task_description="classify tweets",)` but we'd produce different SQL function calls for each.

It seems like we should support `args` but not `kwargs`, instead of vice versa, since SQL doesn't really have a notion of keywords.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -0,0 +1,126 @@
...
+        summary = s.apply(Summarize).iloc[0]
+        # this length check is to get around the fact that this function may not be deterministic
+        assert 0 < len(summary) < len(content)
+
+
+@pytest.mark.skipif(
+    running_on_jenkins(),
+    reason="TODO: SNOW-1859087 snowflake.cortex.sentiment SSL error",
+)
+def test_apply_snowflake_cortex_sentiment(session):
```
sfc-gh-mvashishtha, line 45:
```suggestion
def test_apply_snowflake_cortex_sentiment_series(session):
```

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -0,0 +1,126 @@
...
+def test_apply_snowflake_cortex_negative(session):
+
+    # TODO: SNOW-1758914 snowflake.cortex.sentiment error on GCP
+    if session.connection.host == "sfctest0.us-central1.gcp.snowflakecomputing.com":
+        return
+
+    content = "One day I will see the world."
+    s = pd.Series([content])
+    df = pd.DataFrame([content])
+    with pytest.raises(NotImplementedError):
```
sfc-gh-mvashishtha, line 115: can we parametrize this test by `is_series` and the callable `operation` so that each test case checks for a single `NotImplementedError`?

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -1533,6 +1538,18 @@ def is_supported_snowpark_python_function(func: AggFuncType) -> bool:
...
 
 
+def is_supported_snowflake_cortex_function(func: AggFuncType) -> bool:
+    """Return True if the `func` is a supported Snowflake Cortex function."""
+    func_module = inspect.getmodule(func)
+    if func_module is None or "snowflake.cortex" not in func_module.__name__:
+        return False
+    if func not in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY:
+        ErrorMessage.not_implemented(
+            f"Snowflake Cortex function `{func.__name__}` is not supported yet."
```
sfc-gh-helmeleegy, line 1548: Can we use the same error message pattern used in `is_supported_snowflake_cortex_function` below?

CHANGELOG.md
```diff
@@ -89,14 +88,21 @@
 - Fixed a bug in `replace` and `lit` which raised type hint assertion error when passing `Column` expression objects.
 - Fixed a bug in `pandas_udf` and `pandas_udtf` where `session` parameter was erroneously ignored.
 
+#### Deprecations
+
+- Deprecated Snowpark Python function `snowflake_cortex_summarize`.
```
sfc-gh-helmeleegy, line 93: Didn't we also remove sentiment?

sfc-gh-helmeleegy: Ah ok, got it. It wasn't released yet.

sfc-gh-helmeleegy (approved):
Looks good in general, just left a small nit.

CHANGELOG.md
```diff
@@ -89,14 +88,21 @@
 - Fixed a bug in `replace` and `lit` which raised type hint assertion error when passing `Column` expression objects.
 - Fixed a bug in `pandas_udf` and `pandas_udtf` where `session` parameter was erroneously ignored.
 
+#### Deprecations
```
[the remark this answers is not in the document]
sfc-gh-helmeleegy: +1. Maybe "Removed Features" to better match "New Features".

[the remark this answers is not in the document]
sfc-gh-lmukhopadhyay: Yes good idea!

CHANGELOG.md
sfc-gh-lmukhopadhyay: Yes I think this changelog will have to be reformatted again once 1.27 release is out

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -1533,6 +1538,18 @@ def is_supported_snowpark_python_function(func: AggFuncType) -> bool:
     return True
 
 
+def is_supported_snowflake_cortex_function(func: AggFuncType) -> bool:
+    """Return True if the `func` is a supported Snowflake Cortex function."""
+    func_module = inspect.getmodule(func)
+    if func_module is None or "snowflake.cortex" not in func_module.__name__:
+        return False
```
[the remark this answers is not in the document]
sfc-gh-lmukhopadhyay: Yes we can do `if func in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY` instead, but we wouldn't be able to give the error `f"Snowflake Cortex function `{func.__name__}` is not supported yet."` as easily since support for all Cortex functions have not been tested yet right?

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-lmukhopadhyay: Do you mean `is_supported_snowpark_python_function` , or the apply method in snowflake_query_compiler?

sfc-gh-helmeleegy: Yes, the apply method, my bad.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: It is nice to reference Snowflake Cortex in the error message, so how about we only raise that error message if the function is in `snowflake.cortex`, but still use `SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY` to check for the supported Snowflake functions? I also suggest that we use a method like [this](https://github.com/snowflakedb/snowpark-python/blob/b2206b2281f9676073f58157e9d8c52b5a8eeeee/src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py#L88) to check whether the function is in `snowflake.cortex`. `getmodule` says "Try to guess which module an object was defined in," which sounds dubious.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: even with arg support, the syntax would be `s.apply( ClassifyText, args=(["travel", "cooking"], "classify tweets",))` where the user would have to add the arguments in the order of how they appear in the Cortex docs right? Since apply w/ Snowpark python functions don't support args yet, I was thinking we might want to behave similarly, and include in the docs that func args must be passed via the documented order?

tests/integ/modin/test_apply_snowflake_cortex_functions.py
sfc-gh-lmukhopadhyay: apply on series uses `qc.apply`, and apply on df uses `qc.applymap` which both have different unsupported arguments (na_action or raw for example) so I don't know if parameterizing on the df/ series input would make sense in this case?
Similarly, Cortex functions don't have the same arguments that are passed in apply (for ex Translate has `source_language` and `target_language` additional args where Sentiment has none). Is there a better way to parameterizing on the operation that I'm maybe missing?

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -1540,14 +1540,15 @@ def is_supported_snowpark_python_function(func: AggFuncType) -> bool:
 
 def is_supported_snowflake_cortex_function(func: AggFuncType) -> bool:
     """Return True if the `func` is a supported Snowflake Cortex function."""
-    func_module = inspect.getmodule(func)
-    if func_module is None or "snowflake.cortex" not in func_module.__name__:
+    cortex_func_names = [i[0] for i in inspect.getmembers(snowflake.cortex)]
```
sfc-gh-mvashishtha, line 1543: `inspect` can be slow sometimes. can we save `CORTEX_FUNCTION_NAMES` as a global tuple just once?

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -1540,14 +1540,15 @@ def is_supported_snowpark_python_function(func: AggFuncType) -> bool:
...
+    if func.__name__ not in cortex_func_names:
         return False
-    if func not in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY:
-        ErrorMessage.not_implemented(
-            f"Snowflake Cortex function `{func.__name__}` is not supported yet."
+    elif func in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY:
+        return True
+    else:
+        raise ErrorMessage.not_implemented(
+            f"Snowpark pandas apply API doesn't yet support Snowflake Cortex function `{func.__name__}`"
```
sfc-gh-mvashishtha, line 1550: It's not ideal to have this function raise an error for some unsupported functions and return `False` for others. I suggest that we remove this function and instead raise an error within the query compiler.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8810,8 +8810,8 @@ def _apply_snowflake_cortex_function_to_columns(
         """Apply Snowflake Cortex function to columns."""
 
         def sf_function(col: SnowparkColumn) -> SnowparkColumn:
-            if not kwargs:
-                return snowflake_function(col)
+            # if not kwargs:
+            #     return snowflake_function(col)
```
sfc-gh-mvashishtha, line 8814: did you mean to delete these?
```suggestion
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: What we have now allows `kwargs` but ignores the keywords. We generally want to avoid adding dimensions that have no effect to an API.

Why not pass the args and kwargs args directly to the snowflake ML function? I don't think we need to be consistent with our level of support for args and kwargs in Snowpark python functions.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -1533,6 +1538,19 @@ def is_supported_snowpark_python_function(func: AggFuncType) -> bool:
...
+    """Return True if the `func` is a supported Snowflake Cortex function."""
+    cortex_func_names = [i[0] for i in inspect.getmembers(snowflake.cortex)]
+    if func.__name__ not in cortex_func_names:
+        return False
+    elif func in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY:
+        return True
+    else:
+        raise ErrorMessage.not_implemented(
+            f"Snowpark pandas apply API doesn't yet support Snowflake Cortex function `{func.__name__}`"
+        )
```
sfc-gh-jjiao, line 1551: nit: is it better if we could check  `func in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY:` first and only when it is false, we could construct the list of `cortex_func_names` and check  `if func.__name__ not in cortex_func_names:` ?    it does not change the theoretical complexity , but it will do less work in some case

tests/integ/modin/test_apply_snowflake_cortex_functions.py
sfc-gh-lmukhopadhyay: Oh I see, I've updated this test now to parameterize the operation and is_series

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-lmukhopadhyay: Yes that makes sense to me, I've added the function check in the query compiler in this manner and removed `is_supported_snowflake_cortex_function`

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: @sfc-gh-mvashishtha I've now updated this so either args or kwargs are passed to the Cortex function as positional arguments. So both `s.apply(ClassifyText, args=(["travel", "cooking"]) `and `s.apply(ClassifyText, categories=["travel", "cooking"])` would work

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -456,6 +456,32 @@
 
 T = TypeVar("T", bound=Callable[..., Any])
 
+_CORTEX_FUNC_NAMES = [
```
sfc-gh-mvashishtha, line 459: nit: I like that this is global, but can we

1) store functions themselves instead of names, e.g. `snowflake.cortex.ClassifyText` instead of `"ClassifyText"`. We can use `func in _CORTEX_FUNCTIONS` instead of `func.__name__ in _CORTEX_FUNCTION_NAMES`
2) Keep the list global, but generate it dynamically, like [this](https://github.com/snowflakedb/snowpark-python/blob/e204d1e25cf3585f9385b9562c490fca98e282ba/src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py#L89)
3) Use a tuple instead of a list because tuples are immutable.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8481,11 +8507,12 @@ def apply(
                 ErrorMessage.not_implemented(
                     f"Snowpark pandas apply API doesn't yet support Snowflake Cortex function `{func.__name__}`with raw = {raw}."
                 )
-            if len(args) > 0:
+            return self._apply_snowflake_cortex_function_to_columns(func, args, kwargs)
+        else:
```
sfc-gh-mvashishtha, line 8511: nit: use `elif` instead of `else` with `if` in the body

Unfortunately GitHub won't let me make a code suggestion here.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8855,16 +8884,18 @@ def applymap(
...
-        if is_supported_snowflake_cortex_function(func):
+
+        if func in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY:
             if na_action:
                 ErrorMessage.not_implemented(
                     f"Snowpark pandas applymap API doesn't yet support Snowflake Cortex function `{func.__name__}` with na_action == '{na_action}'"
                 )
-            if len(args) > 0:
+            return self._apply_snowflake_cortex_function_to_columns(func, args, kwargs)
+        else:
```
sfc-gh-mvashishtha, line 8894: nit: use `elif` instead of `else` with `if` in the body

Unfortunately GitHub won't let me make a code suggestion here.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: @sfc-gh-lmukhopadhyay I'm really sorry for wasting your time with this args and kwargs question, but I remembered that we originally disallowed args and kwargs for Snowpark Python functions because we didn't want to deal with handling arguments that could refer to other columns. There are more details on this decision [here](https://docs.google.com/document/d/17Kk5YfXDF6tSfUSCqjadjniYsGX8Kc4lgG-mICMP_Pk/edit?tab=t.0#heading=h.cyi24m17upy5).

I think we have the same risk with the cortex functions, e.g. [`translate`](https://docs.snowflake.com/en/developer-guide/snowpark-ml/reference/latest/api/cortex/snowflake.cortex.translate#snowflake.cortex.translate) can take `Column` arguments for `from_language` and `to_language`.

I recommend that we stay consistent with our support for Snowpark Python functions and not support args or kwargs at all. I would also want to add code comments about both snowpark Python and Snowflake ML function support explaining why we dont' support args and kwargs. Those comments should have been there already.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8805,17 +8832,19 @@ def sf_function(col: SnowparkColumn) -> SnowparkColumn:
...
-            #     return snowflake_function(col)
             resolved_positional = []
             resolved_positional.append(col)
-            for arg in kwargs:
-                resolved_positional.append(kwargs[arg])
+            if args:
+                resolved_positional.append(args)
+            elif kwargs:
+                for arg in kwargs:
+                    resolved_positional.append(kwargs[arg])
```
sfc-gh-mvashishtha, line 8847: if we remove args and kwargs support as I recommend [here](https://github.com/snowflakedb/snowpark-python/pull/2943#discussion_r1942160949), we can delete this function.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -103,26 +138,47 @@ def test_apply_snowflake_cortex_classify_text(session):
...
-    with pytest.raises(NotImplementedError):
-        df.apply(ClassifyText, args=(["travel", "cooking"]))
-    with pytest.raises(NotImplementedError):
-        s.apply(Sentiment, na_action="ignore")
-    with pytest.raises(NotImplementedError):
-        df.apply(Sentiment, raw=True)
     with pytest.raises(NotImplementedError):
-        df.apply(Sentiment, axis=1)
+        modin_input = (pd.Series if is_series else pd.DataFrame)([content])
+        operation(modin_input)
```
sfc-gh-mvashishtha, line 184: nit: only included the `operation` call in the `raises` block.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -81,20 +82,54 @@ def test_apply_snowflake_cortex_sentiment_df(session):
...
-            ClassifyText, list_of_categories=["travel", "cooking"]
-        ).iloc[0]
-        text_class_label = text_class["label"]
+
+        modin_input = (pd.Series if is_series else pd.DataFrame)([content])
+        text_class = operation(modin_input)
+        if is_series:
+            text_class_label = text_class.iloc[0]["label"]
+        else:
+            # text_class_label = text_class["label"][0]
```
sfc-gh-mvashishtha, line 131:
```suggestion
```

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -81,20 +82,54 @@ def test_apply_snowflake_cortex_sentiment_df(session):
...
+        ),
+        param(
+            False,
+            (lambda df: df.apply(ClassifyText, categories=["travel", "cooking"])),
+            2,
+            id="df_cortex_classify_text_kwargs",
+        ),
+        param(
+            True,
+            (lambda s: s.apply(ClassifyText, args=(["travel", "cooking"]))),
```
sfc-gh-mvashishtha, line 102:
```suggestion
            (lambda s: s.apply(ClassifyText, args=(["travel", "cooking"],))),
```

`args` without the comma is the list `["travel", "cooking"]` itself rather than a tuple containing that list.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -81,20 +82,54 @@ def test_apply_snowflake_cortex_sentiment_df(session):
...
+            1,
+            id="series_cortex_classify_text_args",
+        ),
+        param(
+            False,
+            (lambda df: df.apply(ClassifyText, args=(["travel", "cooking"]))),
```
sfc-gh-mvashishtha, line 108:
```suggestion
            (lambda df: df.apply(ClassifyText, args=(["travel", "cooking"],))),
```

ditto

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: Yes good point, I was just changing to 1. myself due to udfs not having `__name__`. For 2.,  I believe you'd mentioned `inspect.getmembers` can be slow in the discussion [here](https://github.com/snowflakedb/snowpark-python/pull/2943#discussion_r1941905256), is that still a concern for generating the cortex func list dynamically?

sfc-gh-mvashishtha: If we're just doing it once, that's okay; I think the penalty is worth the benefit of being up-to-date with what's in `snowflake.cortex`. I just wanted to avoid generating the list on every `apply` call.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: I believe we already support kwargs for Snowpark Python and numpy functions right? Also, if we don't support args or kwargs, what would the apply user behavior look like? `s.apply(ClassifyText, categories=["travel", "cooking"])` would mean that ["travel", "cooking"] would be passed in as a kwarg by default if I'm understanding correctly.

sfc-gh-lmukhopadhyay: ^ discussed offline. Treating args / kwargs in a separate PR

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -82,78 +82,43 @@ def test_apply_snowflake_cortex_sentiment_df(session):
...
+                )
+            ),
+            id="df_cortex_unsupported_function_translate",
         ),
         param(
             True,
-            (lambda s: s.apply(ClassifyText, args=(["travel", "cooking"]))),
-            1,
-            id="series_cortex_classify_text_args",
+            (lambda s: s.apply(Sentiment, args=("hello"))),
```
sfc-gh-mvashishtha, line 105:
```suggestion
            (lambda s: s.apply(Sentiment, args=("hello",))),
```

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -82,78 +82,43 @@ def test_apply_snowflake_cortex_sentiment_df(session):
...
+            id="series_cortex_unsupported_args",
         ),
         param(
             False,
-            (lambda df: df.apply(ClassifyText, args=(["travel", "cooking"]))),
-            2,
-            id="df_cortex_classify_text_args",
+            (lambda df: df.apply(Sentiment, args=("hello"))),
```
sfc-gh-mvashishtha, line 110:
```suggestion
            (lambda df: df.apply(Sentiment, args=("hello",))),
```

sfc-gh-mvashishtha (commented):
I just have two minor comments

tests/integ/modin/test_apply_snowflake_cortex_functions.py
sfc-gh-lmukhopadhyay: Fixed!

sfc-gh-mvashishtha (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Release History
 
+## 1.28.0 (TBD)
+
+### Snowpark Python API Updates
+
+#### Removed Features
+
+- Removed support for Snowpark Python function `snowflake_cortex_summarize`. Users can use the Summarize function from Snowflake.Cortex instead.
+- Removed support for Snowpark Python function `snowflake_cortex_sentiment`. Users can use the Sentiment function from Snowflake.Cortex instead.
```
sfc-gh-aling, line 10: Removing public API is BCR change, we should discuss, please do not merge the PR

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+
+- Deprecated Snowpark Python function `snowflake_cortex_summarize`. Users can use the Summarize function from Snowflake.Cortex instead.
+- Deprecated Snowpark Python function `snowflake_cortex_sentiment`. Users can use the Sentiment function from Snowflake.Cortex instead.
+
+### Snowpark pandas API Updates
+
+#### Dependency Updates
+
+- Added a test dependency for `snowflake-ml-python`.
+
```
sfc-gh-aalam, line 17: I don't think we need to call out test dependency

src/snowflake/snowpark/functions.py
```diff
@@ -10744,13 +10745,15 @@ def make_interval(
...
     Summarizes the given English-language input text.
-
     Args:
         text: A string containing the English text from which a summary should be generated.
-
     Returns:
         A string containing a summary of the original text.
     """
+    warning(
+        "snowflake_cortex_summarize",
```
sfc-gh-aalam, line 10754: can you use `@deprecated` decorator

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -14,6 +14,7 @@
 import cloudpickle
 import numpy as np
 import pandas as native_pd
+import snowflake.cortex
 from pandas._typing import AggFuncType
```
sfc-gh-aalam, line 18: this looks like a required dependency.

sfc-gh-aling: yeah, looks like the ml lib should be part of the modin requirements

src/snowflake/snowpark/functions.py
```diff
@@ -10744,13 +10745,15 @@ def make_interval(
...
+        "snowflake_cortex_summarize is deprecated. Please consider using snowflake.cortex.Summarize instead.",
```
sfc-gh-aling, line 10755: better telling users where to get the method
```suggestion
        "snowflake_cortex_summarize is deprecated. Please consider installing snowflake-ml-python and using snowflake.cortex.Summarize instead.",
```

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Release History
 
+## 1.28.0 (TBD)
+
+### Snowpark Python API Updates
+
+#### Deprecations:
+
+- Deprecated Snowpark Python function `snowflake_cortex_summarize`. Users can use the Summarize function from Snowflake.Cortex instead.
+- Deprecated Snowpark Python function `snowflake_cortex_sentiment`. Users can use the Sentiment function from Snowflake.Cortex instead.
```
sfc-gh-aling, line 10: better call it out that the function is available in the ml library

src/snowflake/snowpark/functions.py
sfc-gh-jjiao: That sounds like a good idea, although I am not sure why the @deprecated decorator is rarely used (looks like only 2 usages in snowpark-python repo) .

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-lmukhopadhyay: I had added `snowflake-ml-python` to "modin-development" extra requirements, do you mean MODIN_REQUIREMENTS instead?

src/snowflake/snowpark/functions.py
sfc-gh-lmukhopadhyay: done!

CHANGELOG.md
sfc-gh-lmukhopadhyay: done!

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-lmukhopadhyay: I'm thinking I'll keep `snowflake-ml-python` in "modin-development" but call it out as a dependency in the changelog again?

sfc-gh-lmukhopadhyay: Added it to `MODIN_REQUIREMENTS` instead and updated changelog!

sfc-gh-aalam: I think this would create problems in the future: https://github.com/snowflakedb/snowflake-ml-python/blob/ae87f588c790961552cac41240d6458b8b342204/requirements.yml#L243-L245

We shouldn't be adding it as a dependency for snowpark-python package. Instead the functions should call out that this dependency is required to use cortex function.

When we do this, we should protect the cortex import with try. Something along the lines of

```
try:
    import snowflake.cortex
    SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY = {
        snowflake.cortex.Summarize,
        snowflake.cortex.Sentiment,
    }

    ALL_SNOWFLAKE_CORTEX_FUNCTIONS = tuple(
        i[1] for i in inspect.getmembers(snowflake.cortex)
    )
except ImportError:
    SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY = {}
    ALL_SNOWFLAKE_CORTEX_FUNCTIONS = tuple()
```

sfc-gh-aalam (approved):

PR #2947, opened 2025-01-28
Supporting stored procs in decoder

sfc-gh-batur:

   

   Fixes SNOW-1830568


tests/ast/test_ast_driver.py
```diff
@@ -195,6 +195,7 @@ def compare_base64_results(
     actual_message: proto.Request,
     expected_message: proto.Request,
     exclude_symbols_udfs_and_src: bool = False,
+    test_case: TestCase = None,
```
sfc-gh-vbudati, line 198: why not just take the test case name as the parameter instead of the whole test case?

sfc-gh-batur: Mostly in case we wanted more information from the test case, but, that's something we can do when there is a need for it..

sfc-gh-vbudati (approved):

PR #2950, opened 2025-01-28
SNOW-1897441: Fix missing row position sort in DataFrame.describe

sfc-gh-joshi:

   

   Fixes SNOW-1897441

When calling `describe` on a DataFrame with object columns, pandas will report a `top` column identifying the value that appears most frequently. If the top two values share the same frequency, [pandas documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.describe.html) indicates that it actually does not provide any stability guarantees:
> If multiple object values have the highest count, then the count and top results will be arbitrarily chosen from among those with the highest count.

Tests involving this behavior are currently failing on QA6, where it appears that the order of results returned by a GROUP BY/COUNT query has changed. This PR adds an additional sort on the row position column to ensure that the object value that appears first is always chosen first; this may not always agree with pandas (though pandas does this in all of our current tests), but at least keeps results the same between prod and qa6.

I ran `tests/integ/modin/frame/test_describe.py` with a qa6 account to verify everything passes.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -13558,8 +13558,14 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
                 axis=0,
             )
             # Compute top (the mode of each column) + freq (the number of times this mode appears).
-            top_freq_identifiers = padded_qc._modin_frame.ordered_dataframe.generate_snowflake_quoted_identifiers(
-                pandas_labels=["top", "freq"]
+            # Also create a new identifier to track min(__row_position__) of each group to ensure stability
+            new_identifiers = padded_qc._modin_frame.ordered_dataframe.generate_snowflake_quoted_identifiers(
+                pandas_labels=["top", "freq", "min_row_position"]
+            )
+            top_freq_identifiers = new_identifiers[0:2]
+            min_row_position_ident = new_identifiers[2]
+            row_position_ident = (
+                padded_qc._modin_frame.ordered_dataframe.row_position_snowflake_quoted_identifier
             )
             # To accommodate multi-level columns in the source frame, we generate a new index column
             # in the top/freq frame for each level. We transpose this frame later, so the columns
@@ -13608,7 +13614,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
                 #        IFF(a IS NULL, NULL, COUNT(a)) AS freq
                 # FROM df
                 # GROUP BY a
-                # ORDER BY freq DESC NULLS LAST
+                # ORDER BY freq DESC, MIN(__row_position__) ASC NULLS LAST
                 # LIMIT 1
                 #
                 # The resulting 1-row frame for column "a": [1, 1, 2] will have the form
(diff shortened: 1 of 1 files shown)


sfc-gh-lmukhopadhyay (approved):
LGTM, thanks Jonathan!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13558,8 +13558,14 @@ def describe(
                 axis=0,
             )
             # Compute top (the mode of each column) + freq (the number of times this mode appears).
-            top_freq_identifiers = padded_qc._modin_frame.ordered_dataframe.generate_snowflake_quoted_identifiers(
-                pandas_labels=["top", "freq"]
+            # Also create a new identifier to track min(__row_position__) of each group to ensure stability
+            new_identifiers = padded_qc._modin_frame.ordered_dataframe.generate_snowflake_quoted_identifiers(
```
sfc-gh-jjiao, line 13562: nit: new_identifiers is not specific enough,  there is nothing called `identifiers`, why don't you call it just `identifiers` ?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13558,8 +13558,14 @@ def describe(
...
+                pandas_labels=["top", "freq", "min_row_position"]
+            )
+            top_freq_identifiers = new_identifiers[0:2]
+            min_row_position_ident = new_identifiers[2]
```
sfc-gh-jjiao, line 13566: `min_row_position_ident` => `min_row_position_id`  or just `min_row_position_identifier`, I also saw it as `indent`

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13656,9 +13662,13 @@ def count_freqs(
                                 pandas_lit(None),
                                 count(col(col_ident)),
                             ).as_(freq_ident),
+                            min_(row_position_ident).as_(min_row_position_ident),
```
sfc-gh-jjiao, line 13665: qq: what does `min_(row_position_ident).as_(min_row_position_ident)` mean?   min_ something as another thing?

sfc-gh-jjiao (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: `min_` is the imported name of the snowpark `min` aggregation; we imported it like this to avoid conflicting with the python builtin `min`. This is equivalent to SQL `SELECT MIN(__row_position__) AS min_row_position`. The column is only used in the call to `OrderedDataFrame.sort`, but unlike SQL the API forces us to explicitly create a new named column instead of passing the aggregation expression directly.

sfc-gh-helmeleegy: Yes, it’s like:
SELECT MIN(A) AS A …

This is a very common pattern in the sql we generate.

sfc-gh-helmeleegy (approved):
Looks good, thanks Johnathan.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: It's `new_identifiers` because the ordered dataframe is generating a st of column identifiers that are guaranteed to be distinct from those existing in the frame.

I've now removed the variable and just assigned `top_identifier, freq_identifier, min_row_position_identifier` directly.

PR #2961, opened 2025-01-30
SNOW-1901507: Fix non-nullable struct type schema expressions

sfc-gh-jrose:

   

   Fixes SNOW-1901507

   This PR fixes an issue with `Session.create_dataframe` calls that would require a temp table creation. The issue caused incorrect schema_query generation when creating dataframes with non-nullable structured type columns. This was due to the temp table being created with a structured schema, but the schema query was expecting a string field instead.

  This PR fixes the problem by deferring temp table creation until after the schema attributes have been modified to account for data serialization. I've added a test that provides a scenario that would have thrown a sql exception previously.

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -3196,30 +3196,10 @@ class Session:
                     f"Invalid schema string: {schema}. "
                     f"You should provide a valid schema string representing a struct type."
                 )
+        create_temp_table = False
         if isinstance(schema, StructType):
             new_schema = schema
-            # SELECT query has an undefined behavior for nullability, so if the schema requires non-nullable column and
-            # all columns are primitive type columns, we use a temp table to lock in the nullabilities.
-            # TODO(SNOW-1015527): Support non-primitive type
-            if (
-                not isinstance(self._conn, MockServerConnection)
-                and any([field.nullable is False for field in schema.fields])
-                and all([field.datatype.is_primitive() for field in schema.fields])
-            ):
-                temp_table_name = random_name_for_temp_object(TempObjectType.TABLE)
-                schema_string = analyzer_utils.attribute_to_schema_string(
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -3296,6 +3276,28 @@ def convert_row_to_list(
             attrs.append(Attribute(quoted_name, sf_type, field.nullable))
             data_types.append(field.datatype)
 
+        if create_temp_table:
+            # SELECT query has an undefined behavior for nullability, so if the schema requires non-nullable column and
```
sfc-gh-aalam, line 3280: does create_temp_table become True in any case other then StructType schema?

sfc-gh-jrose: Not at the moment.

sfc-gh-aalam: I don't understand how is this change different from what was happening before the change. We are running the same code in both cases right? Does it not respect nullability before this change?

sfc-gh-jrose: Nullability was respected previously, but the attributes were different in that part of the function from later in the function. This change moves the bulk of the logic to later in the function after the attributes have been modified by this code:
https://github.com/snowflakedb/snowpark-python/blob/62208e380008ecbfc5c522bf0e4ec5a0e4705ab3/src/snowflake/snowpark/session.py#L3254-L3277

It then uses that modified attrs instead of generating it's own. Previously it did this:
```
schema_string = analyzer_utils.attribute_to_schema_string(
                    schema._to_attributes()
                )
```
Now it does this:
```
schema_string = analyzer_utils.attribute_to_schema_string(attrs)
```

sfc-gh-aalam (approved):

sfc-gh-yuwang (approved):

PR #2967, opened 2025-01-31
SNOW-1906607: Fix telemetry collection for all SnowflakePlan

sfc-gh-aalam:

   

   Fixes SNOW-1906607

   Currently snowflake plan metrics related telemetry is only collected for apis that are tagged with `df_collect_api_telemetry`. This PR fixes this issue and collects telemetry for all snowflake plans under type `snowpark_compilation_stage_statistics`.


src/snowflake/snowpark/_internal/telemetry.py
```diff
@@ -184,33 +183,7 @@ def wrap(*args, **kwargs):
...
-            uuid = args[0]._plan.uuid
-            api_calls[0][CompilationStageTelemetryField.PLAN_UUID.value] = uuid
-            api_calls[0][CompilationStageTelemetryField.QUERY_PLAN_COMPLEXITY.value] = {
-                key.value: value
-                for key, value in plan.cumulative_node_complexity.items()
-            }
-            api_calls[0][TelemetryField.THREAD_IDENTIFIER.value] = threading.get_ident()
-        except Exception:
-            pass
+        api_calls[0][TelemetryField.THREAD_IDENTIFIER.value] = threading.get_ident()
```
sfc-gh-yzou, line 186: would threading.get_ident() throw an exception under any case? if yes, let's keep the try catch part

src/snowflake/snowpark/_internal/telemetry.py
```diff
@@ -184,33 +183,7 @@ def wrap(*args, **kwargs):
         api_calls[0][TelemetryField.SQL_SIMPLIFIER_ENABLED.value] = args[
             0
         ]._session.sql_simplifier_enabled
-        try:
-            plan_state = plan.plan_state
-            api_calls[0][
-                CompilationStageTelemetryField.QUERY_PLAN_HEIGHT.value
-            ] = plan_state[PlanState.PLAN_HEIGHT]
-            api_calls[0][
-                CompilationStageTelemetryField.QUERY_PLAN_NUM_SELECTS_WITH_COMPLEXITY_MERGED.value
```
sfc-gh-yzou, line 193: @sfc-gh-aalam what is the kind of dicrepancy you are referring to in the slack thread https://snowflake.slack.com/archives/C03MJ5AA8CS/p1738347392255399?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -765,6 +773,45 @@ def get_result_set(
 
         return result, result_meta
 
+    def send_plan_metrics_telemetry(self, plan: SnowflakePlan) -> None:
+        """Extract the SnowflakePlan's metrics and including plan_state, uuid identifiers, complexity
+        classification breakdown, and complexity score.
```
sfc-gh-yzou, line 778: shall we get a parameter her to guard the sending of the new telemetry? it now occurs on  the critical pass of the query execution, and we do now know if there will be any side effect introduced

src/snowflake/snowpark/_internal/telemetry.py
sfc-gh-aalam: This is method of in built python module `threading` which is unlikely to fail under. Only issue I could find relating to this throwing an exception is this: https://github.com/python/cpython/issues/128189 which looks like it was more of a user error than a library error.

src/snowflake/snowpark/_internal/telemetry.py
sfc-gh-aalam: In our telemetry collected for job etl logs, we have sql statements submitted by snowpark client which have failed with error codes we are interested in, but the corresponding telemetry for the same plan uuid does not exist in client telemtery being sent from this part of code. This is a decorator -`df_collect_api_telemetry` is only put on select few apis and is not applied on all functions. As a result, will miss some important cases like `dataframe.write.save_as_table`.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aalam: there should ideally be no side effects but having a param protection sounds reasonable given how unstable we have been recently.

sfc-gh-yzou (approved):

src/snowflake/snowpark/_internal/telemetry.py
sfc-gh-helmeleegy: Unrelated, but do we think that any of the existing logic in the decorator may also be relevant to not-currently-decorated functions like `save_as_table`?

sfc-gh-helmeleegy (approved):
Looks good, thanks Afroz.

sfc-gh-jrose (approved):

PR #2993, opened 2025-02-07
SNOW-1887897: [DataSource] add custom schema support

sfc-gh-aling:

   

   Fixes SNOW-1887897

- add custom type support
- add dbms detection
- wrap exception
- column name related code optimization

### DIFF
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -46,6 +46,7 @@ from snowflake.snowpark._internal.type_utils import (
     convert_sp_to_sf_type,
     Connection,
     sql_server_type_to_snow_type,
+    type_string_to_type_object,
 )
 from snowflake.snowpark._internal.utils import (
     INFER_SCHEMA_FORMAT_TYPES,
@@ -58,12 +59,15 @@ from snowflake.snowpark._internal.utils import (
     get_temp_type_for_object,
     normalize_local_file,
     DATA_SOURCE_DBAPI_SIGNATURE,
+    detect_dbms,
+    DBMS_TYPE,
 )
 from snowflake.snowpark.column import METADATA_COLUMN_TYPES, Column, _to_col_if_str
 from snowflake.snowpark.dataframe import DataFrame
 from snowflake.snowpark.exceptions import (
     SnowparkSessionException,
     SnowparkClientException,
+    SnowparkDataframeReaderException,
 )
 from snowflake.snowpark.functions import sql_expr
 from snowflake.snowpark.mock._connection import MockServerConnection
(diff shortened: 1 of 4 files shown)


sfc-gh-jrose (approved):
Maybe unrelated to this change, but what happens if a user specifies a schema with unsupported types? Do the dbapis support things like GeometryType or VariantType?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1076,14 +1080,26 @@ def dbapi(
         max_workers: Optional[int] = None,
         query_timeout: Optional[int] = 0,
         fetch_size: Optional[int] = 0,
+        custom_schema: Optional[Union[str, StructType]] = None,
```
sfc-gh-yuwang, line 1083: in this design, this parameter have to be exact same column names as that in sql server/oracle db,
Do you think we should allow user to input a map like: {data_source_column_name_string -> StructType("new name", data_type)}, so that user can rename during read? They could also possibly only pick a part of columns(we can provide this functionality later)

tests/integ/test_data_source_api.py
```diff
@@ -348,3 +358,41 @@ def assert_datasource_statement_params_run_query(*args, **kwargs):
             single=True,
         )
         assert called == 2
+
+
+@pytest.mark.parametrize(
+    "custom_schema",
+    [
+        "id INTEGER, int_col INTEGER, real_col FLOAT, text_col STRING, blob_col BINARY, null_col STRING",
```
sfc-gh-yuwang, line 366: I think we also need test of custom_schema that include all types

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aling: this is alignment with pyspark. what you describe could be a new feature

tests/integ/test_data_source_api.py
sfc-gh-aling: updated

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1198,17 +1214,20 @@ def dbapi(
...
-        raw_schema = conn.cursor().execute(query).fetchall()
-        return self._to_snowpark_type(raw_schema), raw_schema
+        try:
+            query = f"""
+                SELECT COLUMN_NAME, DATA_TYPE, CHARACTER_MAXIMUM_LENGTH, NUMERIC_PRECISION, NUMERIC_SCALE, IS_NULLABLE
+                FROM INFORMATION_SCHEMA.COLUMNS
+                WHERE TABLE_NAME = '{table}'
+                """
+            raw_schema = conn.cursor().execute(query).fetchall()
+            return self._to_snowpark_type(raw_schema)
```
sfc-gh-yuwang, line 1226: we might have to use raw schema in some future external datasource we want to support, but we are good for sql server and oracle db

sfc-gh-yuwang: we can add it back any time

tests/integ/test_data_source_api.py
```diff
@@ -348,3 +367,55 @@ def assert_datasource_statement_params_run_query(*args, **kwargs):
...
+                StructField("double_col", DoubleType()),
+                StructField("decimal_col", DecimalType()),
+                StructField("map_col", MapType()),
+                StructField("array_col", ArrayType()),
+                StructField("var_col", VariantType()),
+            ]
+        ),
+    ],
+)
+def test_custom_schema(session, custom_schema):
```
sfc-gh-yuwang, line 402: is there a reason that we only test custom schema in sqllite3 while not test sql server?

sfc-gh-aling: this is to test the functionality of setting custom schema for the target snowflake db, the source database doesn't matter as long as we have the data prepared.

on the other hand, using sqlite3 because it's easy to set up, it's a python built-in lib using file as the storage. on the other hand, sql server requires docker image / or secret management which introduces extra cycles to dev

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aling: sounds good to me

PR #2996, opened 2025-02-10
SNOW-1893781: support ingest oracle db

sfc-gh-yuwang:

   

   Fixes SNOW-1893781


src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    "varbinary": BinaryType,
+    "image": BinaryType,
+    "sql_variant": VariantType,
+    "geography": GeographyType,
+    "geometry": GeometryType,
+    "uniqueidentifier": StringType,
+    "xml": StringType,
+    "sysname": StringType,
+}
+ORACLEDB_TYPE_TO_SNOW_TYPE = {
```
sfc-gh-aling, line 70: could you check if different oracledb version may have different types, for example, does oracledb 23ai introduce new types?

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,275 @@
+#
+# Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
+#
+
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
```
sfc-gh-aling, line 8: nit: redundant license info

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    "timestamp": TimestampType,
+    "longraw": BinaryType,
+    "raw": BinaryType,
+    "clob": StringType,
+    "nclob": StringType,
+    "blob": BinaryType,
+    "char": StringType,
+    "nchar": StringType,
+    "rowid": StringType,
+    "sys.anydata": VariantType,
```
sfc-gh-aling, line 93: this data looks very special, I see there are types like ANYDATA/ANYDATASET/ANYTYPE in https://docs.oracle.com/en/database/oracle/oracle-database/19/arpls/ANYDATA-TYPE.html

is this mapped to those types?

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,279 @@
...
+        remove_space_column_name = column[1].lower().replace(" ", "")
+        processed_column_name = (
+            "timestamp"
+            if remove_space_column_name.startswith("timestamp")
+            else remove_space_column_name
+        )
+        snow_type = ORACLEDB_TYPE_TO_SNOW_TYPE.get(processed_column_name, None)
+        if snow_type is None:
+            # TODO: SNOW-1912068 support types that we don't have now
+            raise NotImplementedError(f"sql server type not supported: {column[1]}")
```
sfc-gh-aling, line 167:
```suggestion
            raise NotImplementedError(f"oracle type not supported: {column[1]}")
```

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,279 @@
...
+#
+
+#
+# Copyright (c) 2012-2024 Snowflake Computing Inc. All rights reserved.
+#
+
+import datetime
+from typing import List, Any, Tuple, Protocol, Union
+from snowflake.connector.options import pandas as pd
+from dateutil import parser
```
sfc-gh-aling, line 12: I remember dateutil is not a required dependency of snowpark, it's from the optional pandas depdency.
can you double check it?

sfc-gh-yuwang: it is a required package for snowpark IR, so I think it is safe it use it here

src/snowflake/snowpark/_internal/data_source_utils.py
sfc-gh-yuwang: this is only map to SYS.ANYDATA, according to https://docs.google.com/document/d/1PLI9OnB2v2v6zJfs2rj9fLYdSymImwhX/edit,
the other two type is not defined yet

src/snowflake/snowpark/_internal/data_source_utils.py
sfc-gh-yuwang: as far as I see, seems like there is no new types introduced since 19c

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,347 @@
...
+    LongType,
+    TimeType,
+    ByteType,
+    DateType,
+    TimestampTimeZone,
+    StructType,
+    StructField,
+)
+
+_logger = logging.getLogger("snowflake.snowpark")
```
sfc-gh-aling, line 34: usually we do __name__ to use the module name

```suggestion
_logger = getLogger(__name__)
```

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,347 @@
...
+        for col in schema.fields:
+            if isinstance(col.datatype, TimestampType) and col.datatype.tz in [
+                TimestampTimeZone.TZ,
+                TimestampTimeZone.LTZ,
+            ]:
+                tz_data.append(col.name)
+        # apply read to LOB object, we currently have FakeOracleLOB because CLOB and BLOB is represented by an
+        # oracledb object and we cannot add it as our dependency in test, so we fake it in this way
+        # TODO: SNOW-1923698 remove FakeOracleLOB after we have test environment
+        df = df.map(lambda x: x.read() if isinstance(x, (LOB, FakeOracleLOB)) else x)
```
sfc-gh-aling, line 301: I think we should avoid having test related code from the production code.
what you can do is putting the dependencies into a develop requirements, and update tox.ini to have a new env datasource setup, you can check how modin does this (search `modin-development` in the code base).

we can do it in a separate PR.

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,347 @@
...
+                )
+            else:
+                cols.append(field.name)
+        return f"""select {" , ".join(cols)} from {table}"""
+    elif current_db == DBMS_TYPE.SQL_SERVER_DB or current_db == DBMS_TYPE.SQLITE_DB:
+        return f"select * from {table}"
+    else:
+        raise NotImplementedError(
+            f"currently supported drivers are pyodbc and oracledb, got: {current_db}"
+        )
```
sfc-gh-aling, line 339: same current_db comment

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+class FakeOracleLOB:
+    def __init__(self, value: str) -> None:
+        self.value = value
+
+    def read(self):
+        return self.value
```
sfc-gh-aling, line 347: let's remove it in the PR

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,347 @@
...
+            ):
+                cols.append(
+                    f"""TO_CHAR({field.name}, 'YYYY-MM-DD"T"HH24:MI:SS.FF6 TZH:TZM')"""
+                )
+            elif (
+                isinstance(field.datatype, TimestampType)
+                and field.datatype.tz == TimestampTimeZone.LTZ
+            ):
+                cols.append(
+                    f"""TO_CHAR({field.name} AT TIME ZONE SESSIONTIMEZONE, 'YYYY-MM-DD"T"HH24:MI:SS.FF6 TZH:TZM')"""
```
sfc-gh-aling, line 329: `YYYY-MM-DD"T"HH24:MI:SS.FF6 TZH:TZM` to what second precision (.FFx) does oracle db support? I know snowflake supports up to FF9

src/snowflake/snowpark/_internal/data_source_utils.py
sfc-gh-yuwang: can we remove this in the PR that I create new dependencies? The test would break without this or the new dependencies change

src/snowflake/snowpark/_internal/data_source_utils.py
sfc-gh-yuwang: this is also true for oracledb that is support up to FF9, updated

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,344 @@
...
+    ByteType,
+    DateType,
+    TimestampTimeZone,
+    StructType,
+    StructField,
+)
+
+_logger = logging.getLogger(__name__)
+
+SQL_SERVER_TYPE_TO_SNOW_TYPE = {
```
sfc-gh-jkew, line 35: Once you get enough of these you should be able to create a hierarchy of sorts to reduce duplication and make supporting multiple databases easier. Unfortunately it's not a clean hierarchy, but MySQL shares qualities with Oracle. Redshift and PostgreSQL are similar. DuckDB is largely PostgreSQL, etc.

At the root of this is ANSI SQL; though - so I would start with that as a base set of types.

You can do this work later; but it can eventually become unmanageable and you will have to refactor it.

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+def detect_dbms_pyodbc(dbapi2_conn):
+    """Detects the DBMS type for a pyodbc connection."""
+    import pyodbc
+
+    dbms_name = dbapi2_conn.getinfo(pyodbc.SQL_DBMS_NAME).lower()
+
+    # Set-based lookup for SQL Server
+    sqlserver_keywords = {"sql server", "mssql", "sqlserver"}
```
sfc-gh-jkew, line 142: freetds drivers being used for sql server may also be common on linux systems

src/snowflake/snowpark/_internal/data_source_utils.py
sfc-gh-yuwang: that is in our plan to refactor this part of code so that it would be easier to manage

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,344 @@
...
+        pass
+
+    def fetchone(self):
+        pass
+
+    def close(self):
+        pass
+
+
+def sql_server_to_snowpark_type(schema: List[tuple]) -> StructType:
```
sfc-gh-jkew, line 182: You may want to create a class for SQLServer and Oracle right away; and move them each into individual files. That will make the inevitable shift into a class hierarchy and mixins easier.

src/snowflake/snowpark/_internal/data_source_utils.py
sfc-gh-yuwang: yeah, we may support that in the future as well

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,344 @@
...
+        fields.append(StructField(column[0], data_type, bool(column[4].lower() == "y")))
+
+    return StructType(fields)
+
+
+def infer_data_source_schema(conn: Connection, table: str) -> StructType:
+    try:
+        current_db, driver_info = detect_dbms(conn)
+        cursor = conn.cursor()
+        if current_db == DBMS_TYPE.SQL_SERVER_DB:
```
sfc-gh-jkew, line 256: Make get_schema a function on the SQLServerClass and Oracle class.

The more I read this the more I think you should reorganize this for long term growth. Think 20-30 possible connectors eventually.

sfc-gh-yuwang: it's on our next step

sfc-gh-jkew (commented):
I would refactor this a bit into classes so it's easier to create many new data sources which are related to one-another through a mixin pattern.

src/snowflake/snowpark/_internal/data_source_utils.py
```diff
@@ -0,0 +1,344 @@
...
+        return df
+    elif current_db == DBMS_TYPE.ORACLE_DB:
+        # apply read to LOB object, we currently have FakeOracleLOB because CLOB and BLOB is represented by an
+        # oracledb object and we cannot add it as our dependency in test, so we fake it in this way
+        # TODO: SNOW-1923698 remove FakeOracleLOB after we have test environment
+        df = df.map(
+            lambda x: x.read()
+            if (hasattr(x, "__name__") and x.__name__.lower() == "lob")
+            else x
+        )
```
sfc-gh-aling, line 308: this is leaving notes for perf improvement, we should be able to track which columns is of what type so that we can just apply the function to the specific instead of map which is elementwise

src/snowflake/snowpark/_internal/data_source_utils.py
sfc-gh-aling: this is a good piece of suggestion, let's refactor in our next sprint

sfc-gh-aling (approved):

PR #2998, opened 2025-02-11
SNOW-1914292: Fix keyword argument handling for Snowpark Python functions in apply and applymap.

sfc-gh-lmukhopadhyay:

   

   Fixes SNOW-1914292

   Applying functions with keyword arguments was previously handled via aligning them to the positional Snowpark Python functions using snowpark_function.__code__.co_varnames. However, this does not work for special functions such as trunc(e, scale) or log(base, x). We now propose a different way to pass kwargs as positional arguments to the Snowpark python function using inspect.signature.parameters.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8785,21 +8785,20 @@ def sf_function(col: SnowparkColumn) -> SnowparkColumn:
...
                 if arg in kwargs:
                     resolved_positional.append(kwargs[arg])
                 else:
-                    if not col_specified:
+                    if not found_snowpark_column:
                         resolved_positional.append(col)
-                        col_specified = True
+                        found_snowpark_column = True
+                    elif params[arg].default is not inspect.Parameter.empty:
+                        # if the unspecified arg has a default value, don't need to add to resolved_positional
```
sfc-gh-mvashishtha, line 8800: We need to specify parameters with default values even if the user hasn't specified them. As an example, if we add support for [`array_sort`](https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/1.27.0/snowpark/api/snowflake.snowpark.functions.array_sort), which has `sort_ascending: Optional[bool] = True` as the second argument, and then `nulls_first` as the last argument, then for `df.apply(array_sort, nulls_first=True)` the current code would produce the Snowpark Python function call `array_sort(col, True)` , which is equivalent to `array_sort(array=col, sort_ascending=True, nulls_first=False)`. I don't think we can reproduce this bug with an [existing](https://github.com/snowflakedb/snowpark-python/blob/91871fa1fa3718b0c8bfaadea68ee1fea36dfd2c/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py#L84) supported Snowpark Python function, because none of them have an argument with a default as the non-last argument.

I think the solution is just to explicitly add the default argument to the positional arguments with `resolved_positional.append(params[arg].default)`.

sfc-gh-jjiao:
The description only has the link to the JIRA issue describing the problem and the root cause,  could we add a paragraph to describe the fix / solution?  This will help the reviewers to review too. Thanks  @sfc-gh-lmukhopadhyay

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: One thing is that for the Snowflake specific arguments such as '_emit_ast', '_ast', or 'session', I notice that this approach results in `func got multiple values for argument '_emit_ast'`. We might need to check if an argument is among these special cases and not append them as a positional argument.

CHANGELOG.md
```diff
@@ -25,6 +25,10 @@
 
 - Added support for applying Snowflake Cortex functions `Summarize` and `Sentiment`.
 
+#### Bug Fixes
+
+- Fixed a bug in `apply` where kwargs were not being passed into the applied function.
```
sfc-gh-jjiao, line 30: this only fix the kwards issue for apply or it also fix it for applymap?

tests/integ/modin/test_apply_snowpark_python_functions.py
```diff
@@ -51,6 +51,38 @@ def test_apply_log10():
     )
 
 
+@sql_count_checker(query_count=4)
+def test_apply_trunc_default_scale(session):
```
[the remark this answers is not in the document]
sfc-gh-lmukhopadhyay: I don't believe np.trunc has an arg for truncating to a specified number of places after the decimal point. Maybe np.round() would work for our case to test this

tests/integ/modin/test_apply_snowpark_python_functions.py
```diff
@@ -51,6 +51,38 @@ def test_apply_log10():
...
+
+
+@sql_count_checker(query_count=4)
+def test_apply_log_base(session):
+    from snowflake.snowpark.functions import log
+
+    native_s = native_pd.Series([7, 20.033, 4.09, 7.0, 8.33])
+    s = pd.Series(native_s)
+
+    assert_series_equal(s.apply(log, base=10), native_s.apply(np.log10))
```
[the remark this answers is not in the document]
sfc-gh-lmukhopadhyay: Do you maybe mean non-default value of x?

tests/integ/modin/test_apply_snowpark_python_functions.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: > Maybe np.round() would work for our case to test this

I think so.

> passing scale as a kwarg so that the column provides the values for e.
 
We can add this, though, right?

tests/integ/modin/test_apply_snowpark_python_functions.py
```diff
@@ -83,9 +80,30 @@ def test_apply_log_base(session):
     )
 
 
+@sql_count_checker(query_count=4)
+def test_apply_log_x(session):
```
sfc-gh-mvashishtha, line 84:
```suggestion
def test_apply_pass_base_as_kwarg(session):
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: > One thing is that for the Snowflake specific arguments such as '_emit_ast', '_ast', or 'session', I notice that this approach results in func got multiple values for argument '_emit_ast'. 

I don't understand what's happening here. Shouldn't we specify `_ast` or `_emit_ast` as well? Could you show how to reproduce that error if you don't include special cases for those parameters?

tests/integ/modin/test_apply_snowpark_python_functions.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: Never mind; I see now that snowpark python [`snowflake.snowpark.functions.log`](https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/1.8.0/api/snowflake.snowpark.functions.log) has no default value for `base`, so this test case is fine.

CHANGELOG.md
sfc-gh-lmukhopadhyay: This should be fixing the unspecified kwargs issue for all apply functions such as apply, applymap, and map

tests/integ/modin/test_apply_snowpark_python_functions.py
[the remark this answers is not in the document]
sfc-gh-lmukhopadhyay: Actually, np.round() or math.trunc() don't work with this case since both round the numbers instead of truncating. So far, this seems to but isn't very elegant:
`assert_series_equal(s.apply(trunc, scale=1), native_s.apply(lambda x: (math.floor(x * 10 ** 1) / 10 ** 1)), check_dtype=False)`

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: Replacing with the following in `_apply_snowpark_python_function_to_columns` (removing the `elif not is_external_kwarg(arg)` case) would produce the error above when running `test_apply_log_base`. 
```            
for arg in params:
                if arg in kwargs:
                    resolved_positional.append(kwargs[arg])
                else:
                    if not found_snowpark_column:
                        resolved_positional.append(col)
                        found_snowpark_column = True
                    elif params[arg].default is not inspect.Parameter.empty:
                        # if the unspecified arg has a default value, don't need to add to resolved_positional
                        resolved_positional.append(params[arg].default)
                    else:
                        ErrorMessage.not_implemented(
                            f"Unspecified Argument: {arg} - when using apply with kwargs, all function arguments should be specified except the single column reference (if applicable)."
                        )
```

sfc-gh-lmukhopadhyay: I see this [TODO](https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/utils.py#L1042) in _internal/utils.py:  `# TODO: Could modify internal docstring to display that users should not modify the _emit_ast parameter.` so it's probably preferable to not specify anything for the internal args such as _emit_ast and not append any values for those params?

sfc-gh-mvashishtha: Great find! Because of that code, `_emit_ast` must be specified as a kwarg, whereas we are trying to specify all arguments as positional. For the same reason, if we drop `_emit_ast` but try to specify `_ast` explicitly in a function that also takes `_emit_ast`, then we accidentally pass `_emit_ast` positionally and trigger the error.

First, let's file a bug pointing out that passing `_emit_ast` positionally raises an error. We can link to the code you showed me.

Then, how about we add lines at the beginning of this loop saying:

```python
                if arg in  ("_emit_ast", "_ast"):
                    continue
```

and add a comment 1) pointing to the bug 2) explaining that since those we expect those two parameters appear at the end of the function signature anyway, it's okay to drop them from the positional arguments.

and we can delete the utility function you added since its implementation is simple.

tests/integ/modin/test_apply_snowpark_python_functions.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: That's a good workaround, except I would 

1) define a utility function for trunc with `scale` 
2) use `math.trunc` instead of `math.floor` so we can check we round up for negative numbers.
3) add negative numbers to the test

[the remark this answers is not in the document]
sfc-gh-lmukhopadhyay: I added trunc_util in apply_utils.py. I used python round() function actually instead since math.trunc didn't work with a pandas df

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: Added! I filed the bug [here](https://snowflakecomputing.atlassian.net/browse/SNOW-1927811)

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8785,21 +8785,24 @@ def sf_function(col: SnowparkColumn) -> SnowparkColumn:
...
+            for arg in params:
                 if arg in kwargs:
                     resolved_positional.append(kwargs[arg])
                 else:
-                    if not col_specified:
+                    if not found_snowpark_column:
                         resolved_positional.append(col)
-                        col_specified = True
+                        found_snowpark_column = True
+                    # TODO: SNOW-1927811 Kwargs "_emit_ast" and "_ast" will appear in the function signature
```
sfc-gh-mvashishtha, line 8799: they're in the signature anyway, but we want to say that lower layers of snowpark python will try to add them to kwargs even if we add them to the positional arguments, so we should not add them as positional arguments.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -1638,3 +1638,24 @@ def make_result(value: Any) -> SnowparkColumn:
         return case_expression
 
     return do_map
+
+
+def trunc_util(
```
sfc-gh-mvashishtha, line 1643: This is a good utility function, but we should move it to a test file, because users of this module don't need access to this function.

tests/integ/modin/test_apply_snowpark_python_functions.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: `round` won't work because it rounds to the nearest multiple of `10 ** -scale` rather than to the nearest multiple closer to zero. In `test_apply_trunc_scale`, setting `scale=0` makes the test fail. You can `apply` `math.trunc`  to each element in the series.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-mvashishtha, line 1643: let's rename the function to `trunc_with_scale`

tests/integ/modin/test_apply_snowpark_python_functions.py
```diff
@@ -51,9 +51,79 @@ def test_apply_log10():
...
+    assert_series_equal(s.apply(trunc), native_s.apply(np.trunc))
+    assert_series_equal(s.map(trunc), native_s.map(np.trunc))
+    assert_frame_equal(
+        s.to_frame().applymap(trunc), native_s.to_frame().applymap(np.trunc)
+    )
+    assert_frame_equal(s.to_frame().apply(trunc), native_s.to_frame().apply(np.trunc))
+
+
+@sql_count_checker(query_count=3)
+def test_apply_trunc_scale(session):
```
sfc-gh-mvashishtha, line 69: Let's parametrize `scale` and include negative scales. Let's also add 2- and 3-digit positive and negative values in `native_s.`

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -1638,3 +1638,24 @@ def make_result(value: Any) -> SnowparkColumn:
...
+    e: native_pd.Series, scale: Union[Any, int, float] = 0
```
sfc-gh-mvashishtha, line 1644: for `df.apply(trunc_util)`, the input is a series, but for `series.apply(trunc_util)` or `df.applymap(trunc_util)`, the input is a scalar. Let's be consistent and say that `e` is a scalar in the signature here, and assert that [`is_number`](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_number.html). Then in the test we can call `df.apply(lambda series: series.apply(trunc_util)`.

sfc-gh-mvashishtha (commented):
I have some more minor comments about documentation and testing.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: Modifying this to: `Kwargs "_emit_ast" and "_ast" appear in the function signature and will be passed to the function by Snowpark Python so they should not be added as positional args here` if that's more clear

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: @sfc-gh-mvashishtha Also as of now, I don't think we can include the `elif params[arg].default is not inspect.Parameter.empty:` case since we don't have a Snowpark python function with a default argument that is not last positionally (and it triggers the codecov error). Maybe we can file a TODO bug to add this once such a function is being supported?

sfc-gh-mvashishtha: I would suggest

1) raise an error with `ErrorMessage.not_implemented`
2) exclude the line from codecov with `# pragma: no cover`, like here: https://github.com/snowflakedb/snowpark-python/blob/dde1206aa151c788755caf944a406e2e441e911a/src/snowflake/snowpark/modin/conftest.py#L19
3) comment that we think that we need to add the default value to the positional arguments, but we can't validate that solution yet because we don't reach this case in any of the currently supported Snowpark Python functions.

That way, we will fail fast if we hit this case, and we keep a record of a likely solution.

tests/integ/modin/test_apply_snowpark_python_functions.py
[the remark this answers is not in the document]
sfc-gh-lmukhopadhyay: I've updated trunc_with_scale to use math.trunc and parameterized scale!

tests/integ/modin/test_apply_snowpark_python_functions.py
```diff
@@ -65,22 +87,27 @@ def test_apply_trunc_default_scale(session):
...
 
 
+@pytest.mark.parametrize("scale", [-1, 0, 1, 2])
 @sql_count_checker(query_count=3)
-def test_apply_trunc_scale(session):
+def test_apply_trunc_scale(session, scale):
 
     native_s = native_pd.Series(
-        [-1.0, -0.9, -0.5, -0.2, 0.0, 0.2, 0.5, 0.9, 1.1, 3.14159]
+        [-1.0, -0.9, -0.5234, -0.268, 0.0, 0.23, 0.582, 0.9, 1.1, 3.14159]
```
sfc-gh-mvashishtha, line 95:
```suggestion
        [-1.0, -0.9, -0.5234, -0.268, 0.0, 0.23, 0.582, 0.9, 1.1, 3.14159, 31.4159, 39, -31.4159, -39]
```

We should add some  2- and 3-digit numbers

tests/integ/modin/test_apply_snowpark_python_functions.py
```diff
@@ -1,20 +1,42 @@
...
+    places after the decimal point.
+
+    Args
+    ----
+    e: The input expression to truncate
+    scale: The number of places after the decimal point which to truncate
+
+    Returns
+    -------
+        truncated input as an int
```
sfc-gh-mvashishtha, line 34:
```suggestion
        truncated input
```

tests/integ/modin/test_apply_snowpark_python_functions.py
```diff
@@ -1,20 +1,42 @@
...
+from pandas._typing import Scalar
+from pandas.core.dtypes.common import is_numeric_dtype
 
-from snowflake.snowpark.modin.plugin._internal.apply_utils import trunc_util
 from tests.integ.modin.utils import assert_frame_equal, assert_series_equal
 from tests.integ.utils.sql_counter import sql_count_checker
 from snowflake.snowpark.functions import trunc, sin, _log10, log, desc, asc
 
 
+def trunc_with_scale(e: Scalar, scale: Union[Any, int, float] = 0) -> Union[Any, float]:
```
sfc-gh-mvashishtha, line 20:
```suggestion
def trunc_with_scale(e: Scalar, scale: Union[Any, int, float] = 0) -> float
```

I think it's always a `float`, but I'm not sure.

sfc-gh-mvashishtha (commented):
a few more nits

tests/integ/modin/test_apply_snowpark_python_functions.py
sfc-gh-lmukhopadhyay: Added

tests/integ/modin/test_apply_snowpark_python_functions.py
sfc-gh-lmukhopadhyay: Yes it should be a float after the division

CHANGELOG.md
```diff
@@ -41,6 +41,10 @@
 
 - Fixed a bug in Dataframe.join that caused columns to have incorrect typing.
 
+#### Bug Fixes
```
sfc-gh-helmeleegy, line 44: This is misplaced. Can you please move it to the `Snowpark pandas API Updates` section?

sfc-gh-helmeleegy (approved):
Thanks for the fix, Labanya.

CHANGELOG.md
sfc-gh-jjiao: I think this is under pandas API update (line 30) ?

sfc-gh-jjiao (approved):

CHANGELOG.md
sfc-gh-lmukhopadhyay: It was under Snowpark Local Testing Updates by mistake. Should be fixed now!

PR #3000, opened 2025-02-11
SNOW-1007803: Add support for allow_missing_columns

sfc-gh-jrose:

   

   Fixes SNOW-1007803

   This change adds support for the `allow_missing_columns` to union_by_name and union_all_by_name.

In order to union the dataframe a null filled column is added for each missing column on each side of the union. In order to has correct typing information the schema has to be materialized. This may cause an extra describe query if typing information wasn't already available on the either dataframe.

I've also refactored error handling for this function. This could be considered a BCR, but the scenarios in which an error would be raised have not changed, only the error message. Previously if the right side has all of the left sides columns as well as some additional columns it would result in a sql exception rather than a client exception. I've merged this case with the case that columns are missing from the right side so that regardless of where the columns are missing from the error message provides the relevant information on the client side.


src/snowflake/snowpark/dataframe.py
```diff
@@ -2742,8 +2748,22 @@ def union_by_name(self, other: "DataFrame", _emit_ast: bool = True) -> "DataFram
...
+            |"A"  |"B"  |"C"   |
+            --------------------
+            |1    |2    |NULL  |
+            |1    |2    |3     |
+            --------------------
+            <BLANKLINE>
+
         Args:
             other: the other :class:`DataFrame` that contains the rows to include.
+            allow_missing_columns: When true includes missing columns in the final result. Missing values are Null filled. Default False.
```
sfc-gh-aalam, line 2766: can you file a ticket under this: https://snowflakecomputing.atlassian.net/browse/SNOW-1916219 to get this option added to ast

sfc-gh-aalam (approved):

src/snowflake/snowpark/dataframe.py
```diff
@@ -2791,35 +2834,71 @@ def union_all_by_name(
...
+                    materialized_names = {
+                        StructField(col, DataType()).name for col in missing_left
+                    }
+                    left = left.select(
+                        "*",
+                        *[
+                            lit(None).cast(dt_map[col]).alias(col)
+                            for col in materialized_names
+                        ],
+                    )
```
sfc-gh-aling, line 2878: nit: can we make this a reusable function? handling left and right seems identical

src/snowflake/snowpark/dataframe.py
```diff
@@ -2791,35 +2834,71 @@ def union_all_by_name(
...
+                return left._union_by_name_internal(
+                    right, is_all=is_all, _ast_stmt=_ast_stmt
+                )
+            else:
+                raise SnowparkClientExceptionMessages.DF_CANNOT_RESOLVE_COLUMN_NAME_AMONG(
+                    missing_left, missing_right
+                )
 
-        names = right_project_list + not_found_attrs
+        names = [right_attr_map[col] for col in left_attr_map.keys()]
```
sfc-gh-aling, line 2901: is the change equivalent? I feel the previous code was selecting all columns from right while the current change is selecting columns that show up in the left df.

I'm wondering for the following case:

left columns = ['a', 'b', 'c']
right columns = ['d', 'e']

names would equal to empty, and right_child will be selecting nothing?

sfc-gh-jrose: The functionality is not equivalent. Previously the presence of any items in `not_found_attrs` would cause a sql exception. The case is now raised client-side instead.

sfc-gh-aling (approved):

PR #3006, opened 2025-02-12
SNOW-1923616: [Local Testing] Fix error in CaseWhen statements

sfc-gh-jrose:

   

   Fixes SNOW-1923616

   This PR fixes a bug in CaseWhen statements that was caused when the else statement contained an expression. When evaluating an else expression the index on the input data would be a subset of the total data that matched rows in that should be evaluated in the else clause. This could have an inconsistent ordering which caused nans or Nones to show up in the results when they should not. This PR resets the index before applying the else expression so that it can evaluate correctly, then restores the index after computation is complete.

   I've also implemented functions.hour and functions.minute in order to verify that the fix works for a specific bug report.


src/snowflake/snowpark/mock/_functions.py
```diff
@@ -639,6 +639,22 @@ def mock_current_time(column_index):
...
+@patch("hour")
+def mock_hour(expr):
+    return ColumnEmulator(
+        data=[None if value is None else value.hour for value in expr],
+        sf_type=ColumnType(LongType(), False),
+    )
+
+
+@patch("minute")
+def mock_minute(expr):
```
sfc-gh-aling, line 651: I think these are two features to local testing as well?

sfc-gh-aling (approved):

sfc-gh-jdu (approved):

PR #3011, opened 2025-02-13
SNOW-1923645: [Local Testing] Add support for NULL_IF when loading csvs

sfc-gh-jrose:

   

   Fixes SNOW-1923645

   This PR adds support to the local testing csv reader for the NULL_IF/NULLVALUE option. This allows users to specify one or more values that should be considered as nulls when loading the data.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -1041,6 +1041,9 @@ def convert_value_to_sql_option(value: Optional[Union[str, bool, int, float]]) -
             )  # escape single quotes before adding a pair of quotes
             return f"'{value}'"
     else:
+        if isinstance(value, list):
+            # Snowflake sql uses round brackets for options that are lists
+            return str(tuple(value))
         return str(value)
 
 
--- a/src/snowflake/snowpark/mock/_snowflake_to_pandas_converter.py
+++ b/src/snowflake/snowpark/mock/_snowflake_to_pandas_converter.py
@@ -13,7 +13,7 @@ For full data type spec, please refer to https://docs.snowflake.com/en/sql-refer
 
 import datetime
 from decimal import Decimal
-from typing import Optional, Union
+from typing import List, Optional, Union
 
 from snowflake.snowpark.mock.exceptions import SnowparkLocalTestingException
 from snowflake.snowpark.types import (
@@ -41,7 +41,7 @@ def _integer_converter(
     value: str,
     datatype: DataType,
     field_optionally_enclosed_by: str = None,
-    null_if: Optional[bool] = None,
+    null_if: Optional[List[str]] = None,
 ) -> Optional[int]:
     if value is None or value == "" or null_if is not None and value in null_if:
         return None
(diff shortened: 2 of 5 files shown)


src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1041,6 +1041,9 @@ def convert_value_to_sql_option(value: Optional[Union[str, bool, int, float]]) -
             )  # escape single quotes before adding a pair of quotes
             return f"'{value}'"
     else:
+        if isinstance(value, list):
+            # Snowflake sql uses round brackets for options that are lists
+            return str(tuple(value))
```
sfc-gh-aling, line 1046: is this only used in local testing?
this change doesn't look safe to me `[1,2,3]` now becomes `(1,2,3)`, but surprisingly does not break any tests.

src/snowflake/snowpark/mock/_snowflake_to_pandas_converter.py
```diff
@@ -38,9 +38,12 @@
 
 
 def _integer_converter(
-    value: str, datatype: DataType, field_optionally_enclosed_by: str = None
+    value: str,
+    datatype: DataType,
+    field_optionally_enclosed_by: str = None,
+    null_if: Optional[bool] = None,
```
sfc-gh-aling, line 44: should the type hint be list[str]?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-jrose: That is intentional. Options do no use arrays in snowflake sql so the tuple representation is correct. Surprisingly I did not see any existing tests that passed multiple values to an option. I did double check in the docs that all options that accept multiple values expect the tuple syntax.

src/snowflake/snowpark/mock/_snowflake_to_pandas_converter.py
sfc-gh-jrose: fixed.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-jdu: it should be previously used in copy into

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-aling: interesting..so this is actually a bugfix for options of list.

> all options that accept multiple values

do you know what other feature would use this get_options_statement of list values (I see we have create table/put/get/ etc), I'm wondering if we could test those cases as well to ensure nothing is broken or the change makes us support a unsupported case before.

sfc-gh-jrose: I did not see any tests that used NULL_IF before and the other options act more like dicts.

sfc-gh-aling: also shall we have a bugfix changelog item?

sfc-gh-aling: thanks for checking!

sfc-gh-aling (approved):
LGTM. I see tests are failing, please fix them before merging

PR #3025, opened 2025-02-17
SNOW-1936459: Add modin as optional dependency to meta.yaml

sfc-gh-joshi:

   

   Fixes SNOW-1936459

Previously, when upgrading the version of the `modin` package required by Snowpark pandas, users in Snowflake environments (and our server-side test cases) ran into issues where the conda installation would pick a newer version of modin before Snowpark pandas released with support for it. This creates an annoying situation where we would need to attempt to synchronize the addition of a newer version of `modin` into the Snowflake Anaconda channel with the corresponding release of Snowpark Python that supports it.

As it turns out, conda recipes can define a [`run_constrained`](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#run-constrained) field to restrict optional dependencies like this. This will allow us to pull newer versions of `modin` into the Anaconda channel before updating Snowpark pandas to support it. (see [slack discussion](https://snowflake.slack.com/archives/C07BQJV0RLJ/p1738868920967189) for details)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -47,6 +47,9 @@ requirements:
     - protobuf >=3.20,<6
     - python-dateutil
     - tzlocal
+  run_constrained:
+    # Snowpark pandas
+    - modin==0.30.1
 
 test:
   imports:


sfc-gh-jrose (approved):

sfc-gh-yuwang:
I have some questions about this change:
1. is there a way that we can test this change works and will not affect other pipeline/tests ?
2. do we need to manually update meta.yaml every time pandas support a newer modin?
3. will this cause delay to users access to new modin? For example, if we support new modin (say 0.31.0) before snowpark 1.28.0 release, does user need to wait for 1.28.0 to release to use newer modin package?

sfc-gh-joshi:
> is there a way that we can test this change works and will not affect other pipeline/tests ?

I'm not sure, I don't know where in the build process this file is used.

> do we need to manually update meta.yaml every time pandas support a newer modin?

Yes, but this happens fairly infrequently.

> will this cause delay to users access to new modin? For example, if we support new modin (say 0.31.0) before snowpark 1.28.0 release, does user need to wait for 1.28.0 to release to use newer modin package?

Are you asking about users of the `modin` package without Snowpark pandas? `modin` has newer versions (up to 0.32.0) available in pypi and conda-forge, but not in Snowflake anaconda because adding newer versions would break existing Snowpark pandas users because the package resolver would always pick the newest version.

Currently, to guarantee that workflows don't break, users (and our Snowfort tests) must pin `modin==0.30.1` themselves. The latest available version of Snowpark Python is typically installed implicitly in notebooks/other Snowflake environments. Snowpark pandas will only ever support one version of modin at a time for the foreseeable future, and if we make a newer version of modin available in Snowflake, then the only way it can be used is if a newer Snowpark release supports it.

Within Snowflake, we basically expect users to exclusively use `modin` with Snowpark pandas, since the distributed computation backends it uses otherwise don't really work well in Snowflake (and also have some issues with the way we mock psutil).

sfc-gh-joshi:
> should we use modin<=0.30.1 instead? Or do we only support this version now?

We only support one specific version of modin at a time, since there's bugfixes/features that Snowpark pandas depends on that aren't available in lower versions. This might change when modin reaches 1.x and needs to respect semver, but there's no timeline for that at the moment.

sfc-gh-yuwang (approved):

PR #3058, opened 2025-02-21
SNOW-1906574: Add SelectFromFileNode for read file operations

sfc-gh-aalam:

   

   Fixes SNOW-1906574

   This PR add a new logical plan `SelectFromFileNode` whenever we do a read file operation like `df.reader.csv()`. This is done so that we can easily detect if the subtree contains a child from a `SelectFromFileNode` and skip creating CTE nodes in such cases. The skip logic is handled in PR: https://github.com/snowflakedb/snowpark-python/pull/3063

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,18 @@
 # Release History
+## 1.29.0 (TBD)
 
-## 1.28.0 (TBD)
+### Snowpark Python API Updates
+
+### Snowpark pandas API Updates
+
+### Snowpark Local Testing Updates
+
+#### Bug Fixes
+
+- Fixed a bug in `Series.rename_axis` where an `AttributeError` was being raised.
+
+
+## 1.28.0 (2025-02-20)
 
 ### Snowpark Python API Updates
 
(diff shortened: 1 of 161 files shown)


sfc-gh-jdu (approved):

sfc-gh-helmeleegy:
> This is done so that we can easily detect if the subtree contains a child from a `SelectFromFileNode` and skip creating CTE nodes in such cases.

Don't we have a lower complexity threshold? How does this rule relate to it?

sfc-gh-helmeleegy (approved):
Overall looks good, just had one question.

sfc-gh-aalam:
> > This is done so that we can easily detect if the subtree contains a child from a `SelectFromFileNode` and skip creating CTE nodes in such cases.
> 
> Don't we have a lower complexity threshold? How does this rule relate to it?

I don't think this change affects large query breakdown and complexity thresholds. We basically had a bug reported where CTEs created on top of `SELECT * FORM FILE` would raise XP errors in certain cases. That's why in this PR we add support to detect leaf nodes which are `SELECT * FROM FILE` and in https://github.com/snowflakedb/snowpark-python/pull/3063, we invalidate all ancestors to be considered for CTEs.

PR #3102, opened 2025-03-01
SNOW-856100: pd.Grouper support

sfc-gh-joshi:

   

   Fixes SNOW-856100

Adds support for `pd.Grouper` objects in groupby operations. These objects allow specification of individual labels or index levels, and allow resampling via the `freq` argument if the specified column is a datetime/timestamp.

This changes the code paths for almost every query compiler groupby operation. If a column would be resampled by a Grouper object, its values are now replaced by the corresponding bin it belongs to before the grouping operation is performed. Most operations (including all that invoke `groupby_agg`) have been updated to follow this behavior; some operations (notably `groupby_apply`, `groupby_size`, `groupby_groups`) have not yet been tested with this new resampling behavior.

This implementation also does not currently support resampling the same column twice (e.g. `df.groupby([pd.Grouper("col0", freq="1s"), pd.Grouper("col0", freq="2s")])` is invalid) because it replaces the original column with the resampled column. This restriction is not strictly necessary as we could theoretically append resampled versions as new columns, but is simpler to implement because existing groupby code assumes all columns involved in a grouping operation existed in the original frame.

### DIFF
--- a/tests/unit/modin/test_groupby_utils.py
+++ b/tests/unit/modin/test_groupby_utils.py
@@ -132,7 +132,7 @@ def test_is_groupby_value_label_like(val, expected_result):
         ),
     ],
 )
-def test_invalid_grouper_parameters(val, invalid_params):
+def test_grouper_parameters(val, invalid_params):
     if len(invalid_params) == 0:
         validate_grouper(val)
     else:
--- a/tests/unit/modin/test_groupby_utils.py
+++ b/tests/unit/modin/test_groupby_utils.py
@@ -14,6 +14,7 @@ from snowflake.snowpark.modin.plugin._internal.frame import InternalFrame
 from snowflake.snowpark.modin.plugin._internal.groupby_utils import (
     check_is_groupby_supported_by_snowflake,
     is_groupby_value_label_like,
+    validate_grouper,
 )
 from snowflake.snowpark.modin.plugin.compiler.snowflake_query_compiler import (
     SnowflakeQueryCompiler,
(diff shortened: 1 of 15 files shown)


tests/unit/modin/test_groupby_utils.py
```diff
@@ -89,8 +90,51 @@ def test_check_groupby_snowflake_execution_by_args_axis_1():
...
     ],
 )
 def test_is_groupby_value_label_like(val, expected_result):
     assert is_groupby_value_label_like(val) == expected_result
+
+
+@pytest.mark.parametrize(
+    "val, invalid_params",
+    [
+        (pd.Grouper(freq="1s"), []),  # parameters are valid, don't error
```
sfc-gh-helmeleegy, line 109: Why is this test case included with the test for invalid parameters?

tests/unit/modin/test_groupby_utils.py
```diff
@@ -89,8 +90,51 @@ def test_check_groupby_snowflake_execution_by_args_axis_1():
...
+        (
+            pd.Grouper(level=1, sort=True),
+            ["sort"],
+        ),  # sort=True invalid for non-datetime groupers
+        pytest.param(
+            pd.Grouper(freq="1s", sort=False),
+            ["sort"],
+            marks=pytest.mark.skip(
+                "sort=False is unsupported for datetime groupers, but pandas does not pass the sort argument for some reason"
```
sfc-gh-helmeleegy, line 118: I'm not sure I'm following this. Is `pandas` here referring to native pandas? If so, why do we need to care about pandas behavior here. We are only verifying that we are raising a `NotImplementedError` for unsupported params -- without comparing to native pandas, correct?

tests/unit/modin/test_groupby_utils.py
sfc-gh-joshi: It's testing to make sure the parameter combination is valid. I'll rename the test.

tests/unit/modin/test_groupby_utils.py
sfc-gh-joshi: The `pd.Grouper` object is inherited directly from native pandas, and our internal groupby code can only see the object returned after the Grouper constructor is called. When `pd.Grouper(freq="1s", sort=False)` is called, native pandas will return `TimeGrouper(freq=<Second>, axis=0, sort=True, dropna=True, closed='left', label='left', how='mean', convention='e', origin='start_day')`, so even though we don't support `sort=False`, because the returned object from native pandas doesn't respect the original value of `sort` in this case, it doesn't trigger the error.

sfc-gh-helmeleegy: That makes sense.

tests/unit/modin/test_groupby_utils.py
sfc-gh-helmeleegy: Thanks.

sfc-gh-helmeleegy (approved):
LGTM, modulo one minor nit.

sfc-gh-nkrishna (approved):
Great work, LGTM!

sfc-gh-jjiao:
When you said:
`This implementation also does not currently support resampling the same column twice (e.g. df.groupby([pd.Grouper("col0", freq="1s"), pd.Grouper("col0", freq="2s")]) is invalid) because it replaces the original column with the resampled column.`
did you mean does not currently support resampling the same column **more than once**?  meaning, resampling 3 or 4, 5 .. times also not supported, right?

sfc-gh-joshi:
> did you mean does not currently support resampling the same column more than once? meaning, resampling 3 or 4, 5 .. times also not supported, right?

Yes, that's right. If someone tries to do so, they get a `NotImplementedError` with the message `"Resampling the same column multiple times is not yet supported in Snowpark pandas."`

PR #3103, opened 2025-03-01
SNOW-1904626: Add Resource usage to api telemetry

sfc-gh-aalam:

   

   Fixes SNOW-1904626

   This PR collects the following resource usage data to dataframe APIs:
   - `wall_time`: time it takes to finish an API call
   - `cpu_time`: number of cpu cycles spend to finish the API call
   - `network_bytes_send_kib` : kilobyes of network IO send during API call
   - `network_bytes_recv_kib` : kilobyes of network IO received during the API call
   - `memory_rss_kib` : kilobytes of additional memory the API called required to execute


src/snowflake/snowpark/_internal/telemetry.py
```diff
@@ -120,13 +133,99 @@ class TelemetryField(Enum):
...
+            if PS_UTIL_AVAILABLE:
+                network_sent = (
+                    self._end_net_io_counters.bytes_sent
+                    - self._start_net_io_counters.bytes_sent
+                ) / 1024
+                network_recv = (
+                    self._end_net_io_counters.bytes_recv
+                    - self._start_net_io_counters.bytes_recv
+                ) / 1024
+                memory_rss = (self._end_rss - self._start_rss) / 1024
```
sfc-gh-oplaton, line 198: Not sure how relevant this is, I'll leave it up to you to decide:

* Divide by `1024.0`? Are the outputs intended as integers, or floats?
* Why KiB and not MiB or bytes?
* If sticking with KiB (a reasonable decision, I don't oppose it), would it make sense to put the unit in the key name? (It's already in the key, consider putting it in the Python constant name, too, e.g. `KEY_NETWORK_SENT_KIB`).

src/snowflake/snowpark/dataframe.py
```diff
@@ -2343,11 +2344,18 @@ def distinct(
...
+                    df = self._with_plan(
+                        self._select_statement.distinct(), _ast_stmt=stmt
+                    )
+                else:
+                    df = self._with_plan(Distinct(self._plan), _ast_stmt=stmt)
+            add_api_call(
+                df,
+                "DataFrame.distinct[select]",
+                resource_usage=resource_usage_collector.get_resource_usage(),
+            )
```
sfc-gh-oplaton, line 2358: If this is meant to refer to the `ResourceUsageCollector` on line 2347, indentation seems incorrect.

sfc-gh-oplaton: Nevermind. I forgot that `with` is not a naming scope.

sfc-gh-oplaton (approved):

src/snowflake/snowpark/_internal/telemetry.py
sfc-gh-aalam: updated based on suggestion from point 1, 3.

2) I have no strong reason to pick KiB over MiB. My initial guess is for most api that are not network/memory intensive, KiB should be a good unit to cover it. We can also report just raw bites and adjust the granularity later on.

sfc-gh-jdu (commented):
When adding such telemetry, maybe we should consult legal team to understand whether it's ok to send such data to snowflake. In the past we want to add arg values to telemetry when working on pandas API, but it was disapproved by legal team

src/snowflake/snowpark/_internal/telemetry.py
```diff
@@ -63,6 +71,11 @@ class TelemetryField(Enum):
     KEY_DURATION = "duration"
     KEY_FUNC_NAME = "func_name"
     KEY_MSG = "msg"
+    KEY_WALL_TIME = "wall_time"
+    KEY_CPU_TIME = "cpu_time"
+    KEY_NETWORK_SENT_KIB = "network_bytes_sent_kib"
```
sfc-gh-jkew, line 76: This is lovely. We will use this for cost optimization in the pandas hybrid client.

src/snowflake/snowpark/_internal/telemetry.py
```diff
@@ -120,13 +133,99 @@ class TelemetryField(Enum):
...
+
+    def __init__(self) -> None:
+        pass
+
+    def __enter__(self):
+        try:
+            self._start_time = time.time()
+            self._start_cpu_time = time.process_time()
+            if PS_UTIL_AVAILABLE:
+                self._start_net_io_counters = psutil.net_io_counters()
```
sfc-gh-yuwang, line 159: I have a basic question about psutil, since it is a develop requirement, this telemetry data will not be sent when used by customers, what is this telemetry data meant for and why we don't need them from customer?

sfc-gh-jkew (approved):

src/snowflake/snowpark/_internal/telemetry.py
sfc-gh-jkew: psutil is often available by chance on the system; but because it is architecture specific it is sometimes not.

src/snowflake/snowpark/_internal/telemetry.py
```diff
@@ -120,13 +133,99 @@ class TelemetryField(Enum):
...
+            pass
+
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        try:
+            self._end_time = time.time()
+            self._end_cpu_time = time.process_time()
+            if PS_UTIL_AVAILABLE:
+                self._end_net_io_counters = psutil.net_io_counters()
```
sfc-gh-jkew, line 171: I would double check that this does not cause issues in a sandbox, as I recall we had a "stubbed" or faked version of psutil which simulated some things; but if you called an API that we didn't stub out it would fail.

https://github.com/snowflakedb/snowflake/blob/ef1bdddffd5073faae4c41a08fcdfce9adf27d90/ExecPlatform/src/coprocessor/python/executor/py/joblib_parallel_backend.py#L12

sfc-gh-aalam:
> When adding such telemetry, maybe we should consult legal team to understand whether it's ok to send such data to snowflake. In the past we want to add arg values to telemetry when working on pandas API, but it was disapproved by legal team

Thanks for the heads-up. I'll reach out to legal

src/snowflake/snowpark/_internal/telemetry.py
sfc-gh-aalam: I ran a test inside a sproc environment and we don't have any metrics that depend on PS util but we don't have any errors raised either.

example output for 
```
        [
            [1, 1],
            [1, 1],
            [2, 2],
            [3, 3],
            [4, 4],
        ]
    ).to_df("id", "v")
df = df.distinct().sort(["id", "v"])
```

```
[
   {
      "name":"Session.create_dataframe[values]"
   },
   {
      "name":"DataFrame.to_df",
      "subcalls":[
         {
            "name":"DataFrame.select",
            "wall_time":0.0006542205810546875,
            "cpu_time":0.0
         }
      ],
      "wall_time":0.0007646083831787109,
      "cpu_time":0.0
   },
   {
      "name":"DataFrame.distinct[select]",
      "wall_time":0.00047516822814941406,
      "cpu_time":0.0
   },
   {
      "name":"DataFrame.sort",
      "wall_time":0.0005409717559814453,
      "cpu_time":0.0
   }
]
```

sfc-gh-aalam: ^ this is true even what I added `psutil` as a package dependency.

sfc-gh-jdu (approved):

PR #3111, opened 2025-03-04
SNOW-1762416: Handle AST capture for copy APIs

sfc-gh-heshah:

    [SNOW-1762416](https://snowflakecomputing.atlassian.net/browse/SNOW-1762416)

    - Extracting `Dataframe` plan copying logic into a helper
    - Add separate internal copy methods for both `Dataframe` and `Table` classes which do not emit AST
    - Re-enable AST generation for public `__copy__` method for both `Dataframe` and `Table`
    - Add tests which utilize Python's builtin `copy.copy` methods which invoke the available `__copy__` methods

[SNOW-1762416]: https://snowflakecomputing.atlassian.net/browse/SNOW-1762416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -916,23 +916,36 @@ class DataFrame:
             **kwargs,
         )
 
-    def __copy__(self) -> "DataFrame":
-        """Implements shallow copy protocol for copy.copy(...)."""
+    def _copy_plan(self) -> LogicalPlan:
+        """Returns a shallow copy of the plan of the DataFrame."""
         if self._select_statement:
             new_plan = copy.copy(self._select_statement)
             new_plan.column_states = self._select_statement.column_states
             new_plan._projection_in_str = self._select_statement.projection_in_str
             new_plan._schema_query = self._select_statement.schema_query
             new_plan._query_params = self._select_statement.query_params
+            return new_plan
         else:
-            new_plan = copy.copy(self._plan)
+            return copy.copy(self._plan)
 
-        # TODO SNOW-1762416: Clarify copy-behavior in AST. For now, done as weak-copy always. Yet, we may want to consider
-        # a separate AST entity to model deep-copying. A deep-copy would generate here a new ID different from self._ast_id.
-        # We additionally need to consider behavior when a user calls copy.copy() on a Snowpark Dataframe. For now, consider
-        # all calls internal, and leave AST handling to the caller.
-        df = DataFrame(self._session, new_plan, _emit_ast=False)
-        return df
+    def _copy_without_ast(self) -> "DataFrame":
+        """Returns a shallow copy of the DataFrame without AST generation."""
+        return DataFrame(self._session, self._copy_plan(), _emit_ast=False)
+
+    def __copy__(self) -> "DataFrame":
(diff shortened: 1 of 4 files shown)


tests/ast/data/df_copy.test
```diff
@@ -0,0 +1,886 @@
...
+
+df10 = res7.union(df4)
+
+df11 = df3.union(res8)
+
+df12 = res7.union(res8)
+
+df13 = res7.union(df4)
+
+df14 = df3.union(res8)
```
sfc-gh-heshah, line 81: From the Unparser standpoint, this may reflect a shallower/weaker copy approach than the user desired by invoking `copy.copy`.  The `Table.__copy__` is a much more straightforward deep-copy implementation (only sharing the current `Session` between copies), which makes unparsing relatively simple as well (see previous lines)

Unfortunately, the `Dataframe.__copy__` implementation is neither a shallow or deep-copy approach due to the how the Snowflake plan attributes are copied either by invoking a `<PlanNode.__copy__>` method (also neither deep or shallow in most cases), or via copy-by-reference in the new `Dataframe` instance. The main takeaway is that the copy is shallower than I originally thought, but we still do get a new `Dataframe` instance, which this unparsed code does not reflect.

Generally, I think this should be fine (unless we have issues with users interpreting `new_df = copy.copy(df)` as being equivalent to `new_df = df`), as most of our Dataframe APIs don't have side effects (please comment I've missed any known to have side effects).

Of course, this issue is limited to Phase 0 and mostly the Unparser as far as I can tell. On the server side, we are free to interpret the `DataframeRef` in the AST as either a clone of the original Dataframe's AST or simply a reuse of the already evaluated original Dataframe.

sfc-gh-azwiegincew (approved):

sfc-gh-vbudati (approved):
Thanks Hemit!

src/snowflake/snowpark/dataframe.py
```diff
@@ -916,23 +916,36 @@ def to_local_iterator(
...
+            return copy.copy(self._plan)
 
-        # TODO SNOW-1762416: Clarify copy-behavior in AST. For now, done as weak-copy always. Yet, we may want to consider
-        # a separate AST entity to model deep-copying. A deep-copy would generate here a new ID different from self._ast_id.
-        # We additionally need to consider behavior when a user calls copy.copy() on a Snowpark Dataframe. For now, consider
-        # all calls internal, and leave AST handling to the caller.
-        df = DataFrame(self._session, new_plan, _emit_ast=False)
-        return df
+    def _copy_without_ast(self) -> "DataFrame":
+        """Returns a shallow copy of the DataFrame with the AST enabled."""
```
sfc-gh-aalam, line 932: should we remove `without`

sfc-gh-aalam, line 932:
```suggestion
    def _copy_without_ast(self) -> "DataFrame":
        """Returns a shallow copy of the DataFrame without the AST enabled."""
```
should we add without here?

sfc-gh-aalam (approved):

sfc-gh-heshah:
Changes folded into #3115

PR #3115, opened 2025-03-04
SNOW-1916231: Guard AST assertion behind global ast_enabled flag

sfc-gh-heshah:

   

   Fixes SNOW-NNNNNNN


sfc-gh-heshah:
* **#3115** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3115?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3115?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sfc-gh-aalam (approved):

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/_internal/ast/utils.py
```diff
@@ -66,12 +66,12 @@
...
 
 # TODO(SNOW-1491199) - This method is not covered by tests until the end of phase 0. Drop the pragma when it is covered.
-def debug_check_missing_ast(ast, container) -> None:  # type: ignore[no-untyped-def] # pragma: no cover
+def debug_check_missing_ast(ast, session, container) -> None:  # type: ignore[no-untyped-def] # pragma: no cover
     """
     Debug check for missing AST. This is invoked with various arguments that are expected to be non-NULL if the AST
     is emitted correctly.
     """
-    if ast is None and FAIL_ON_MISSING_AST:
+    if session.ast_enabled and FAIL_ON_MISSING_AST and ast is None:
```
sfc-gh-oplaton, line 74: Is the implication here that `debug_check_missing_ast` is intended only for Session-bound APIs, as opposed to "free" APIs? For a trivial example, a call to `col` doesn't require a session.

tests/ast/data/Session.call.test
```diff
@@ -17,11 +17,11 @@ my_sproc_sp = sproc("my_sproc", return_type=StringType(), input_types=[LongType(
 
 df = session.call("my_sproc", 1, "two", {"param1": 10, "param2": "twenty"}, True)
 
-my_sproc_sp(1, "two", {"param1": 10, "param2": "twenty"}, True)
+df
```
sfc-gh-oplaton, line 20: Is this working as intended?

src/snowflake/snowpark/_internal/ast/utils.py
sfc-gh-heshah: Yes, I believe the check was originally extracted into this helper function to avoid repetition in `Dataframe` APIs when checking if we had an `_ast_id` attribute set, our only front-end indicator of whether the Dataframe had an AST associated with it in the Session's AST Batch or not.

In all other cases, since the class instances themselves carry their AST as attributes, our AST generation code itself would fail on similar missing field errors, albeit raised by the protobuf library rather than explicitly by us.

tests/ast/data/Session.call.test
sfc-gh-heshah: I believe this is what is intended by the API. We need to eagerly evaluate the AST associated with the Dataframe itself, rather than eagerly invoke the sproc (the invocation should be handled on the server side). This also allows us to mark `df` as evaluated, and connect the symbol/ID to its results as well.

The user technically never invokes the sproc themselves on the client side explicitly, so the unparsing previously is incorrect from that perspective as well.

sfc-gh-oplaton (approved):

PR #3123, opened 2025-03-05
Adding support for creating literal values from datetime.timedelta

sfc-gh-dyadav:

   

   Fixes SNOW-1920942

Now snowpark lit function would recognize datetime.timedelta as value and assign a new type to it which will be processed in to_sql to create the correct query.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
+++ b/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
@@ -39,6 +39,7 @@ from snowflake.snowpark.types import (
     _FractionalType,
     _IntegralType,
     _NumericType,
+    _TimeDeltaType,
 )
 
 MILLIS_PER_DAY = 24 * 3600 * 1000
@@ -232,6 +233,10 @@ def to_sql(
 
     if isinstance(datatype, VectorType):
         return f"{value} :: VECTOR({datatype.element_type},{datatype.dimension})"
+
+    if isinstance(datatype, _TimeDeltaType):
+        assert value is not None
+        return f"INTERVAL '{value.total_seconds()} seconds'"
 
     if isinstance(datatype, FileType):
         # TODO: SNOW-1950688: Remove parsing workaround once the server is ready for accepting full stage URI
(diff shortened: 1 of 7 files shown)


src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -233,6 +234,10 @@ def to_sql(
     if isinstance(datatype, VectorType):
         return f"{value} :: VECTOR({datatype.element_type},{datatype.dimension})"
 
+    if isinstance(datatype, _TimeDeltaType):
+        assert value is not None and isinstance(value, timedelta)
```
sfc-gh-dryazanov, line 238: `value is not None` is not needed since `isinstance` will cover that case.

But I'm curious, can the user trigger this assert? I don't see asserts for other data types, so maybe it's not even needed?

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -233,6 +234,10 @@ def to_sql(
...
+        return f"INTERVAL '{value.total_seconds()} seconds'"
```
sfc-gh-dryazanov, line 239: When I try `datetime.timedelta(seconds=100).total_seconds()`, I get a "100.0", not "100". How does it work here? Does it not require a cast to `int`?

Also, do we need to support milliseconds / microseconds?

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
sfc-gh-dyadav: I wanted to do the if then I thought the exception raised is only for the datatype but it looks it is raised for if any of the value of datatype is wrong so changed it now. Thanks for catching.

sfc-gh-dryazanov (approved):

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
sfc-gh-dyadav: Thanks for catching this Dima, resolved it.

sfc-gh-aalam (commented):
Looks good to me. Maybe we need some local testing changes here. can you add any integ tests to show how this type is being used

PR #3136, opened 2025-03-08
SNOW-1899126: Fix malformed sql for functions that create internal alias

sfc-gh-aalam:

   

   Fixes SNOW-1899126

   In this PR, we do the following:
   - Add a new internal method `Column._alias` to mark aliases created internally by snowpark functions.
   - Add a new UnaryExpression `_InternalAlias` to differentiate between internal and user-created aliases.
   - Refactor the `analyze` method to ignore internal aliases when it is not top level expression.


src/snowflake/snowpark/functions.py
```diff
@@ -3837,7 +3837,7 @@ def array_remove_nulls(col: Column) -> Column:
         ),
         separator=lit(sep, _emit_ast=False),
         _emit_ast=False,
-    ).alias(f"CONCAT_WS_IGNORE_NULLS('{sep}', {names})", _emit_ast=False)
+    )._alias(f"CONCAT_WS_IGNORE_NULLS('{sep}', {names})")
```
sfc-gh-aling, line 3840: what's the criteria to use this internal `_alias`?
I'm thinking in the future, if other developers are implementing new functions or other functionality which uses internal alias:

1. how can they be aware of this _alias
1. how they know they should use _alias vs alias

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -592,6 +613,11 @@ def analyze(
             str(expr)
         )  # pragma: no cover
 
+    def internal_alias_extractor(self, expr: Expression) -> Expression:
+        if isinstance(expr, _InternalAlias):
+            return expr.child
+        return expr
+
```
sfc-gh-aling, line 620: do you think we can make it part of the `analyze` function like the following way:
```python
if isinstance(expr, _InternalAlias):
    return self.analyze(expr.child)
```

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 ### Snowpark Python API Updates
 
+#### Bug Fixes
+
+- Fixed a bug for the following functions that raised errors `.cast()` is applied to their output
+  - `from_json`
+  - `size`
```
sfc-gh-aling, line 11: I still don't understand why the internal alias would cause failure to cast.. is it related to the incorrect column name?

sfc-gh-aalam: When we apply any other function on top of an aliased column, for example cast, the sql generated is malformed. For example `session.sql("select 1 as A").select(col("A").alias("A_alias").cast(StringType()))` generates
```
SELECT  CAST ("A" AS "A_ALIAS" AS STRING) FROM (select 1 as A)
```

This raises error. Therefore for functions implemented internally like listed here which add an `_alias`, if any operation is applied to the resulting column, we must ignore the alias and apply the function. More details are in doc attached in the JIRA

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-aalam: I do want to simplify it but if we follow what you suggest, it would ignore all expressions. We still need to allow the alias to get passed through `unary_expression_extractor` so that when the function is top level, we still add the alias.

tests/integ/test_column.py
```diff
@@ -144,6 +155,33 @@ def test_contains(session):
...
+            )
+        ],
+    )
+
+    Utils.check_answer(
+        df.select(window(df.ts, "10 seconds").cast(StringType())),
+        [Row('{"end":"1970-01-01 00:00:10.000","start":"1970-01-01 00:00:00.000"}')],
+    )
+    Utils.check_answer(
+        df.select(second(to_timestamp(window(df.ts, "10 seconds")["end"]))),
```
sfc-gh-jrose, line 180: For this second example what does the final column name end up being?

sfc-gh-jrose (approved):

CHANGELOG.md
sfc-gh-aling: oh I see, this example is clear enough to understand.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -592,6 +613,11 @@ def analyze(
             str(expr)
         )  # pragma: no cover
 
+    def internal_alias_extractor(self, expr: Expression) -> Expression:
```
sfc-gh-aling, line 616: one last comment: can you add comment on when this function needs to called when analyzing expression to give some context to future us?

sfc-gh-aling (approved):

PR #3138, opened 2025-03-10
SNOW-1904626 fix df api calls

sfc-gh-aalam:

   

   Fixes SNOW-1904626


src/snowflake/snowpark/dataframe.py
```diff
@@ -2356,6 +2356,7 @@ def distinct(
                 ],
                 _emit_ast=False,
             ).agg(_emit_ast=False)
+            adjust_api_subcalls(df, "DataFrame.distinct[group_by]", len_subcalls=2)
```
sfc-gh-aling, line 2359: I have a dumb question what this `adjust_api_subcalls` is doing compared to `add_api_call` call and how you decide the `len_subcalls` value?

sfc-gh-aalam: when an internal api is implemented using other api's for example 
```
df = self.group_by(
                [
                    self.col(quote_name(f.name), _emit_ast=False)
                    for f in self.schema.fields
                ],
                _emit_ast=False,
            ).agg(_emit_ast=False)
```
In this case, `group_by` and `agg` adds two api calls to `df._plan`. Since the are called internally, we move them into subcalls so the resulting `df._plan.api_calls` is 
```
{
  "name": "DataFrame.distinct",
  "subcalls": [
    {"name": "DataFrame.group_by"},
    {"name": "RelationalGroupedDataFrame.agg"},
]}
```

`len_subcalls` is decided by how many of most recently added `df._plan.api_calls` needs to be moved to subcalls. It is usually the number of internal apis being called.

`add_api_call` is when we add an api call to a dataframe api without using a decorator like `df_api_usage` for any reason.

sfc-gh-yuwang (approved):

sfc-gh-jdu (approved):

PR #3144, opened 2025-03-11
SNOW-1900040: Estimate an upper bound for row counts between operations in OrderedDataFrame

sfc-gh-nkrishna:

   

   Fixes SNOW-1900040

   This PR adds a RowCountEstimator utility class to estimate the upper bound row count of an `OrderedDataFrame` after an operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Improvements
 
-- Added estimate for row count upper bound to `OrderedDataFrame`.
+- Added estimate for row count upper bound to `OrderedDataFrame` to enable better engine switching. This could potentially result in increased query counts.
 
 ### Snowpark Python API Updates
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 1.31.0 (YYYY-MM-DD)
 
+### Snowpark pandas API Updates
+
+### Improvements
+
+- Added estimate for row count upper bound to `OrderedDataFrame`.
+
 ### Snowpark Python API Updates
 
 #### New Features
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -1231,6 +1231,7 @@ class OrderedDataFrame:
         original_right_quoted_identifiers = (
             right.projected_column_snowflake_quoted_identifiers
         )
+        original_right_ordered_dataframe = right
         # De-duplicate the column identifiers of right against self (left), so that
         # Snowpark doesn't perform any de-duplication on the result dataframe during join.
         right = right._deduplicate_active_column_snowflake_quoted_identifiers(
@@ -1263,6 +1264,7 @@ class OrderedDataFrame:
                 self,
                 DataFrameOperation.JOIN,
                 args={
+                    "right": original_right_ordered_dataframe,
                     "left_on_cols": left_on_cols,
                     "right_on_cols": right_on_cols,
                     "how": how,
@@ -1352,6 +1354,7 @@ class OrderedDataFrame:
             self,
             DataFrameOperation.JOIN,
             args={
+                "right": original_right_ordered_dataframe,
                 "left_on_cols": left_on_cols,
                 "right_on_cols": right_on_cols,
                 "how": how,
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -432,16 +432,6 @@ class OrderedDataFrame:
             count("*").over().as_(row_count_snowflake_quoted_identifier),
         )
 
-        # Get the row count from the underlying Snowpark dataframe.
-        materialized_row_count = (
-            ordered_dataframe._dataframe_ref.snowpark_dataframe.select(
-                row_count_snowflake_quoted_identifier
-            ).first()[row_count_snowflake_quoted_identifier.strip('"')]
-        )
-        # Set the row count and upper bound to the materialized row count.
-        ordered_dataframe.row_count = materialized_row_count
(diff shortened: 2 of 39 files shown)


src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -31,6 +31,10 @@
     sum as sum_,
 )
 from snowflake.snowpark.modin.plugin._typing import AlignTypeLit, JoinTypeLit
+from snowflake.snowpark.modin.plugin._internal.row_count_estimation import (
```
sfc-gh-jkew, line 34: @sfc-gh-nkrishna Main thing missing is tests. I think the general approach makes sense; and while I haven't checked the math on all the cases we just need some actual tests.

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -0,0 +1,124 @@
...
+        # Unpivot creates a new row for each value in the column list
+        elif operation == DataFrameOperation.UNPIVOT:
+            column_list = args["column_list"]
+            return current * len(column_list)
+
+        # Agg aggregates the rows into a single row
+        elif operation == DataFrameOperation.AGG:
+            return 1
+
+        # TODO: Implement a better estimate by having cases for different join types
```
sfc-gh-helmeleegy, line 85: Join types differ in their behavior when there are unmatched rows on either side of the join. But when every row on one side matches every row on the other side (the cartesian product case), they all behave the same way. So their upper bound will be the same too.

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -0,0 +1,124 @@
...
+                raise ValueError(
+                    "Cannot estimate row count: right DataFrame has no row count information."
+                )
+            return current * other_bound
+
+        # Limit sets the upper bound to n rows
+        elif operation == DataFrameOperation.LIMIT:
+            return args["n"]
+
+        # Sample can cause the row count to be set to n or reduced by a fraction
```
sfc-gh-helmeleegy, line 111: frac can be >1 so it can cause the row count to both decrease and increase.

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -0,0 +1,124 @@
...
+            return None
+
+        # These operations preserve or reduce the row count, so we can use the current upper bound
+        if operation in {
+            DataFrameOperation.SELECT,
+            DataFrameOperation.DROPNA,
+            DataFrameOperation.GROUP_BY,
+            DataFrameOperation.SORT,
+            DataFrameOperation.PIVOT,
+            DataFrameOperation.FILTER,
```
sfc-gh-helmeleegy, line 62: Sometimes we can have enough information to be more precise about the size of the output of a filter, for example in some cases of `iloc`.

sfc-gh-jkew:
As an FYI; I may merge this into the modin-hybrid-branch to see if we can reduce the count calls in our hybrid approach. I'll let you know what I find.

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
sfc-gh-nkrishna: The current code performs `ceil(current * frac)`, so if frac > 1 the estimated row count should go up and behaves correctly. Let me update the comment though

sfc-gh-helmeleegy: Yes, I was referring to the comment :)

sfc-gh-helmeleegy:
Can the modin test failures be fixed before merging?

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -425,6 +431,17 @@ def ensure_row_count_column(self) -> "OrderedDataFrame":
             *self.projected_column_snowflake_quoted_identifiers,
             count("*").over().as_(row_count_snowflake_quoted_identifier),
         )
+
+        # Get the row count from the underlying Snowpark dataframe.
+        materialized_row_count = (
```
sfc-gh-jkew, line 436: I don't think we need to do an eager query here. We can get the real row count cheaply in a couple of ways:

- When we need num_rows anyways, like for a repl or anything like that
- When we initially create the data frame

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -0,0 +1,119 @@
...
+        # Get the current upper bound. If not set, return None
+        current = df.row_count_upper_bound
+        if current is None:
+            return None
+
+        # These operations preserve or reduce the row count, so we can use the current upper bound
+        if operation in {
+            DataFrameOperation.SELECT,
+            DataFrameOperation.DROPNA,
+            DataFrameOperation.GROUP_BY,
```
sfc-gh-joshi, line 59: I don't know if we should include GROUP_BY here, since there's some operations that will increase the row count. The first thing that comes to mind for me is an up-sampled `resample` operation (as in the [pandas documentation example](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.SeriesGroupBy.resample.html)), but I'm sure there's some pathological `apply` cases as well.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## 1.31.0 (YYYY-MM-DD)
 
+### Snowpark pandas API Updates
+
+### Improvements
+
+- Added estimate for row count upper bound to `OrderedDataFrame`.
```
sfc-gh-joshi, line 9: You should mention the potential for increased query counts.

Also, we're not using the row count estimates anywhere yet, correct? Isn't this strictly a performance regression for the time being?

sfc-gh-jkew (approved):

sfc-gh-helmeleegy (approved):

PR #3146, opened 2025-03-11
SNOW-1976643: Allow creating multi-session in sprocs if enabled

sfc-gh-smirzaei:
1. Description:
With dbt project, we are planning on supporting multiple session insides procs. So we have added new API to _snowflake to create a pristine cloned version of the default server session. Using this, we will create a new new Snowpark session (a new object with new connection obj) and we will set the session_id of the connection (will be done in C API).  Currently we have a check in Snowpark lib which disallows creating multiple sessions inside sprocs. With this check we will that check the corresponding param is enabled (this param will only be enabled for dbt sprocs).

For more info on the multi-session project see this design doc: https://docs.google.com/document/d/1lwqyykHo549Vx3UWpFLWot1LIdYrwUjGKtGbQGaCnd8/edit?tab=t.0#heading=h.sil75ywnlm6n

An example of how this new API is used:
```
create or replace procedure get_session_sp()
returns string
language python
packages=('snowflake-snowpark-python')
runtime_version=3.8
handler='handler'
as
$$import _snowflake

def handler(session):
  new_session = _snowflake.create_session()
  new_session.sql('select foo')
  return 'done'
$$;
```
3. Fill out the following pre-review checklist:

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -557,11 +557,12 @@ class Session:
         conn: Union[ServerConnection, MockServerConnection, NopConnection],
         options: Optional[Dict[str, Any]] = None,
     ) -> None:
-        if (len(_active_sessions) >= 1
-                and is_in_stored_procedure()
-                and not conn._get_client_side_session_parameter(
+        if (
+            len(_active_sessions) >= 1
+            and is_in_stored_procedure()
+            and not conn._get_client_side_session_parameter(
                 "ENABLE_CREATE_SESSION_IN_STORED_PROCS", False
-                )
+            )
         ):
             raise SnowparkClientExceptionMessages.DONT_CREATE_SESSION_IN_SP()
         self._conn = conn
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -557,7 +557,12 @@ class Session:
         conn: Union[ServerConnection, MockServerConnection, NopConnection],
         options: Optional[Dict[str, Any]] = None,
     ) -> None:
-        if len(_active_sessions) >= 1 and is_in_stored_procedure():
+        if (len(_active_sessions) >= 1
+                and is_in_stored_procedure()
+                and not conn._get_client_side_session_parameter(
+                "ENABLE_CREATE_SESSION_IN_STORED_PROCS", False
+                )
+        ):
             raise SnowparkClientExceptionMessages.DONT_CREATE_SESSION_IN_SP()
         self._conn = conn
         self._query_tag = None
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -557,7 +557,12 @@ def __init__(
         conn: Union[ServerConnection, MockServerConnection, NopConnection],
         options: Optional[Dict[str, Any]] = None,
     ) -> None:
-        if len(_active_sessions) >= 1 and is_in_stored_procedure():
+        if (len(_active_sessions) >= 1
+                and is_in_stored_procedure()
+                and not conn._get_client_side_session_parameter(
```
sfc-gh-smirzaei, line 562: This param will be enabled for internal dbt use only.

src/snowflake/snowpark/session.py
```diff
@@ -557,7 +557,13 @@ def __init__(
         conn: Union[ServerConnection, MockServerConnection, NopConnection],
         options: Optional[Dict[str, Any]] = None,
     ) -> None:
-        if len(_active_sessions) >= 1 and is_in_stored_procedure():
+        if (
+            len(_active_sessions) >= 1
+            and is_in_stored_procedure()
+            and not conn._get_client_side_session_parameter(
+                "ENABLE_CREATE_SESSION_IN_STORED_PROCS", False
```
sfc-gh-smirzaei, line 564: `ENABLE_CREATE_SESSION_IN_STORED_PROCS` is an internal param that will be only enabled for internal dbt runs.

tests/integ/test_session.py
```diff
@@ -284,9 +284,17 @@ def test_create_session_in_sp(session):
...
+            "ENABLE_CREATE_SESSION_IN_STORED_PROCS", False
+        ):
+            with pytest.raises(SnowparkSessionException) as exec_info:
+                Session(session._conn)
+            assert exec_info.value.error_code == "1410"
+        else:
+            try:
+                Session(session._conn)
+            except SnowparkSessionException as e:
+                pytest.fail(f"Unexpected exception {e} was raised")
```
sfc-gh-aling, line 297: you might also want a positive case to ensure when the session parameter is set to True, no error is raised

sfc-gh-smirzaei: Ummm I may be missing something, but isn't that what my change do? basically the else part of `if not session._conn._get_client_side_session_parameter(
            "ENABLE_CREATE_SESSION_IN_STORED_PROCS", False
        ):`

sfc-gh-aling: yes, a test case that shows with `ENABLE_CREATE_SESSION_IN_STORED_PROCS` being `true` we can create the Session(session._conn).

sfc-gh-aling (approved):

sfc-gh-jrose (approved):

PR #3162, opened 2025-03-14
Add a functionality in apply_in_pandas to support spark api 

sfc-gh-dyadav:

   

   Fixes SNOW-1800723

If the dataframe comes from spark the column names are not same as the spark column names but the function would be assuming the spark column names and operating like that, this change resolves that issue 
This adds a functionality of (key, dataframe) which can also be the type of function spark support

The test for this will be added in this PR https://github.com/snowflakedb/sas/pull/725/files, this is a fork introduced for the non public usecase of snowpark library

I am extracting the spark names from the column_map which will only be present if this is being sent from the accelerated spark layer.


src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -327,6 +329,7 @@ def apply_in_pandas(
         func: Callable,
         output_schema: StructType,
         _emit_ast: bool = True,
+        _is_called_from_snowpark_connect=False,
         **kwargs,
```
sfc-gh-aalam, line 333: I'm wondering if future APIs are going to behave differently from SAS, then should we use context to store this info?

see: https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/context.py

sfc-gh-dyadav: Done

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -403,8 +406,36 @@ def apply_in_pandas(
...
+                if key_columns is not None:
+                    import numpy as np
+
+                    key_list = [pdf[key].iloc[0] for key in key_columns]
+                    numpy_array = np.array(key_list)
+                    keys = tuple(numpy_array)
+                if original_columns is not None:
+                    pdf.columns = original_columns
+                if key_columns is not None:
+                    return func(keys, pdf)
```
sfc-gh-aalam, line 438:
```suggestion
                if original_columns is not None:
                    pdf.columns = original_columns
                if key_columns is not None:
                    import numpy as np

                    key_list = [pdf[key].iloc[0] for key in key_columns]
                    numpy_array = np.array(key_list)
                    keys = tuple(numpy_array)
                    return func(keys, pdf)
```
nit: can we restructure it this way? grouping the if statements together

sfc-gh-dyadav: It won't work because we are changing the pdf in 
`if original_columns is not None:
                    pdf.columns = original_columns`

src/snowflake/snowpark/context.py
```diff
@@ -26,6 +26,9 @@
 _use_structured_type_semantics = False
 _use_structured_type_semantics_lock = threading.RLock()
 
+# This is an internal-only global flag, used to determine whether the api is called from snowflake.snowpark_connect
+_is_called_from_snowpark_connect = False
```
sfc-gh-skumbham, line 30: How do you feel about using this `_is_snowpark_connect_compatible_mode` this?

sfc-gh-skumbham (approved):

src/snowflake/snowpark/context.py
sfc-gh-dyadav: Thanks for the review :)

PR #3164, opened 2025-03-15
SNOW-1962180: Support relaxed pandas in pd.read_snowflake for query data sources

sfc-gh-helmeleegy:

   

   Fixes SNOW-1962180

   Support relaxed pandas in pd.read_snowflake for query data sources.


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 #### Improvements
 
 - Support relaxed consistency and ordering guarantees in `pd.read_snowflake` for non-query data sources.
+- Support relaxed consistency and ordering guarantees in `pd.read_snowflake` for query data sources.
```
sfc-gh-joshi, line 20: Should these be a single CL entry since they're coming in the same release?

tests/integ/modin/io/test_read_snowflake_query_call.py
```diff
@@ -4,20 +4,35 @@
...
-    session.sql(
-        """
+@pytest.mark.parametrize(
+    "relaxed_ordering",
+    [
+        pytest.param(
+            True,
+            marks=pytest.mark.skip(
+                "Queries with CALL statements raise a SQL compilation "
+                "error when relaxed_ordering=True"
```
sfc-gh-joshi, line 25: Is this by design? Should we explicitly test for the error instead of skipping?

sfc-gh-joshi (commented):
Could you clarify the intended behavior when an ORDER BY clause is encountered in a query, both with and without relaxed pandas?

sfc-gh-nkrishna (approved):
LGTM once Jonathan's comments are addressed

CHANGELOG.md
sfc-gh-helmeleegy: Good point. Done.

tests/integ/modin/io/test_read_snowflake_query_call.py
sfc-gh-helmeleegy: Yes, it's intensional. Unfortunately it's hard for the client to check if a statement is a CALL statement rather than a SELECT statement. For example if we just use string search and pattern matching, we won't be able to recognize the cases where the word "call" is present in say string literal inside a select query.

sfc-gh-joshi: Got it. So basically, if there's a CALL statement with `relaxed_ordering=True` there may or may not be a SQL compilation error, but we have no way of knowing ahead of time with 100% reliability?

Even if it's not fully consistent, I think we should still test the error so we're aware if the behavior changes due to any other server or client-side changes.

sfc-gh-helmeleegy:
> Could you clarify the intended behavior when an ORDER BY clause is encountered in a query, both with and without relaxed pandas?

The order by can only guarantee that the input data is ordered in a certain way. But it does not guarantee how the output data (after applying some pandas operations) will be ordered. I updated the docstring accordingly.

sfc-gh-nkumar (approved):

sfc-gh-joshi (approved):

Issue #3163, opened 2025-03-14
Snowpark Local Testing `minus` removing rows that match values across multiple removed rows


matt-comity:
1. What version of Python are you using?

`Python 3.11.9 (main, Jun 24 2024, 14:49:51) [Clang 15.0.0 (clang-1500.3.9.4)]`

2. What operating system and processor architecture are you using?

`macOS-15.3.1-arm64-arm-64bit`

3. What are the component versions in the environment (`pip freeze`)?

   There are a lot, but relevant for the test example:
```
pandas==2.1.4
snowflake-snowpark-python==1.27.0
```

4. What did you do?

I was trying to use `subtract` / `minus` / `except_` and write some unit tests for my code using Snowpark but ran into some odd behavior. I've created a toy example that illustrates the problem below.

```python
from snowflake.snowpark import Session
from datetime import date
session = Session.builder.config("local_testing", True).create()
df1 = session.create_dataframe([[1, 2], [3, 4]])
df2 = session.create_dataframe([[1, 1], [2, 2]])
df1.subtract(df2).show()
```

5. What did you expect to see?

Expected:
```
---------------
|"_1"  |"_2"  |
---------------
|1     |2     |
|3     |4     |
---------------
```

Got:
```
---------------
|"_1"  |"_2"  |
---------------
|3     |4     |
---------------
```

As you can see, the row `[1, 2]` is getting filtered out despite not existing in the dataframe being subtracted. This is because both `1` and `2` show up as values among the rows. The bug is on this [line of code](https://github.com/snowflakedb/snowpark-python/blob/3a21a065f4a07bcebc1d294724ce058bdd3c804f/src/snowflake/snowpark/mock/_plan.py#L1065), as it is checking if all of the values in each row in `df1` show up in rows in `df2`, but not necessarily the same row. This is due to smushing all the `df2` values together via `cur_df.values.ravel()`, so we lose the row distinctions.

In Snowflake itself, an equivalent query does what you'd expect:

```sql
select * from (select * from values (1, 2), (3, 4)) minus (select * from values (1, 1), (2, 2));
```

```
COLUMN1 | COLUMN2
--      | --
3       | 4
1       | 2
```

6. Can you set logging to DEBUG and collect the logs?

N/A

sfc-gh-sghosh:
Hello @matt-comity ,

Thanks for raising the issue.
we are able to reproduce the issue with local session. 
[...]

PR #3167, opened 2025-03-17
SNOW-1984396: [Local Testing] Fix bug in Dataframe.except_

sfc-gh-jrose:

   

   Fixes SNOW-1984396

   This PR rewrites the local testing case for EXCEPT operator case for finding the difference between two dataframes. The original approach had edge cases that caused data in multiple rows to be included when dropping individual rows. The new approach takes an alternative approach by concatting the right side twice and then removing all rows that have more than one instance.

See https://github.com/snowflakedb/snowpark-python/issues/3163 for more information.


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 #### New Features
 
+#### Bug Fixes
+
+- Fixed a bug in Dataframe.except_ that would cause rows to be incorrectly dropped.
```
sfc-gh-aling, line 17: nit:
```suggestion
- Fixed a bug in `Dataframe.except_` that would cause rows to be incorrectly dropped.
```

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -1060,14 +1060,15 @@ def execute_mock_plan(
...
-                        | (
-                            ~any_null_rows_in_cur_df & null_rows_in_res_df.values
-                        )  # Rows that are all None/NaN only in LEFT
-                    ]
+                    # A side-effect of Snowflake difference is that duplicates are removed from the left side
+                    res_df = res_df.drop_duplicates()
+                    sf_types = res_df.sf_types
+
+                    # Two copies of the right side ensures that all rows present there are dropped when keep=False
+                    res_df = pd.concat([res_df, cur_df, cur_df]).drop_duplicates(
```
sfc-gh-aling, line 1068: why one copy of the right does not work?

sfc-gh-jrose: Say left is the group `(1, 2, 3)` and right is the group `(2, 3, 4)`.

If you concat just one copy you get `(1, 2, 3, 2, 3, 4)`. `drop_duplicates(keep=False)` would then see the duplicates and reduce to `(1, 4)`. We don't want any of the elements in the right side though so this is incorrect.

When two copies of right are in the concat there will always be a duplicate of each element so all right side elements are always dropped.

sfc-gh-jdu (approved):

sfc-gh-aling (approved):

PR #3171, opened 2025-03-18
Fix allowing multiple aggregates in Pivot fix

sfc-gh-dyadav:

   

   Fixes SNOW-1897130, SNOW-1883226

The code to allow multiple aggregates had a typo where it was always passing the first aggregate and the following code was wrong so I have made the new changes to support multiple aggregates when a group by clause is used.


tests/integ/scala/test_dataframe_aggregate_suite.py
```diff
@@ -419,39 +420,48 @@ class MonthlySales(NamedTuple):
...
-    # 1) SUM and AVG
+    with pytest.raises(SnowparkDataframeException) as ex_info:
+        TestData.monthly_sales(session).pivot(
+            "month", ["JAN", "FEB", "MAR", "APR"]
+        ).agg([sum(col("amount")), avg(col("amount"))]).sort(col("empid"))
+
+    assert (
+        "You can apply only one aggregate expression to a RelationalGroupedDataFrame returned by the pivot() method unless the pivot is applied with a groupby clause."
+        in str(ex_info)
+    )
```
sfc-gh-jrose, line 431:
```suggestion
    with pytest.raises(SnowparkDataframeException, match="You can apply only one aggregate expression to a RelationalGroupedDataFrame returned by the pivot() method unless the pivot is applied with a groupby clause."):
        TestData.monthly_sales(session).pivot(
            "month", ["JAN", "FEB", "MAR", "APR"]
        ).agg([sum(col("amount")), avg(col("amount"))]).sort(col("empid"))
```

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1258,10 +1261,23 @@ def pivot_statement(
...
+                quote_name(f"{unwrap_single_quote(value)}_{aggregate}")
+                for value in pivot_values
+            ]
+            aliased_string = [
+                f"{quoted_name} AS {aliased_name}"
+                for aliased_name, quoted_name in zip(aliased_names, quoted_names)
+            ]
+            exclude_str = COMMA.join(quoted_names)
+            alised_str = COMMA.join(aliased_string)
+            select_str = f"{STAR} EXCLUDE({exclude_str}), {alised_str}"
```
sfc-gh-aalam, line 1276:
```suggestion
            aliased_str = COMMA.join(aliased_string)
            select_str = f"{STAR} EXCLUDE({exclude_str}), {aliased_str}"
```
nit: type

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1258,10 +1261,23 @@ def pivot_statement(
...
+        if pivot_values is not None and should_alias_column_with_agg:
+            quoted_names = [quote_name(value) for value in pivot_values]
+            aliased_names = [
+                quote_name(f"{unwrap_single_quote(value)}_{aggregate}")
+                for value in pivot_values
+            ]
+            aliased_string = [
+                f"{quoted_name} AS {aliased_name}"
+                for aliased_name, quoted_name in zip(aliased_names, quoted_names)
+            ]
```
sfc-gh-aalam, line 1273: do we know this list will be non-empty?

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -221,6 +222,8 @@ def _to_df(
                 self._dataframe._select_statement or self._dataframe._plan,
             )
         elif isinstance(self._group_type, _PivotType):
+            if len(agg_exprs) != 1 and len(unaliased_grouping) == 0:
+                raise SnowparkClientExceptionMessages.DF_PIVOT_ONLY_SUPPORT_ONE_AGG_EXPR()
```
sfc-gh-aalam, line 226: what happens with we reach this stage today? Do we end up with a different error? Checking to see if this could be considered a breaking change

sfc-gh-aalam (commented):
Can you also add example where we are aliasing the output columns. It is not clear from the test cases added

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-dyadav: yes it is formed above right where I have checked "pivot_values is not None"

src/snowflake/snowpark/relational_grouped_dataframe.py
sfc-gh-dyadav: It used to be this but it was removed when the two agg was supported in this PR https://github.com/snowflakedb/snowpark-python/pull/2953/files#diff-dfe06bfc0c516428afad2c7819b8b9f36d6bf0701f277f392319bdad71311cc4L223 but the support was wrong because of this line 
https://github.com/snowflakedb/snowpark-python/pull/2953/files#diff-1e9ad2e762dfaa958bd9a3674d0faf337bac01b42e4f5e75996162b3ad53431cR1106, even the expectation test were asserting wrong output and the documentation still says we only support one agg so I think it should be fine

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-aalam: I see. can we also do an `isinstance(pivot_values, List)` check since the type-hint says it can be both `str` and `List[str]`

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -1190,10 +1198,15 @@ def do_resolve_with_resolved_children(
                 # https://docs.snowflake.com/en/sql-reference/constructs/pivot#dynamic-pivot-with-multiple-aggregations-using-union
                 if plan is None:
                     plan = pivot_plan
-                else:
-                    union_plan = UnionPlan(plan, pivot_plan, is_all=False)
-                    plan = self.plan_builder.set_operator(
-                        plan, pivot_plan, union_plan.sql, union_plan
+                elif join_columns is not None:
+                    plan = self.plan_builder.join(
```
sfc-gh-aalam, line 1202: is it possible that both conditions are not met.

sfc-gh-dyadav: It should not be because of the check in [relational_grouped_dataframe.py](https://github.com/snowflakedb/snowpark-python/pull/3171/files#diff-dfe06bfc0c516428afad2c7819b8b9f36d6bf0701f277f392319bdad71311cc4)..if it's not a grouping expression and the aggregates are more than one we throw an error earlier.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-dyadav: We already check in above if that if it is a string, this is the else part

sfc-gh-dyadav:
> Can you also add example where we are aliasing the output columns. It is not clear from the test cases added

Done

sfc-gh-jdu (commented):
why do we only allow multiple pivot when having group by? From https://docs.snowflake.com/en/sql-reference/constructs/pivot#dynamic-pivot-with-multiple-aggregations-using-union, it seems not a requirement

tests/integ/scala/test_dataframe_aggregate_suite.py
```diff
@@ -419,39 +421,81 @@ class MonthlySales(NamedTuple):
...
+    df = (
         TestData.monthly_sales(session)
+        .groupBy(col("empid"))
         .pivot("month", ["JAN", "FEB", "MAR", "APR"])
         .agg([sum(col("amount")), avg(col("amount"))])
-        .sort(col("empid")),
+        .sort(col("empid"))
+    )
+
+    print(df.schema.fields)
```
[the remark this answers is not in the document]
sfc-gh-dyadav: done

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1258,10 +1261,23 @@ def pivot_statement(
...
+            exclude_str = COMMA.join(quoted_names)
+            aliased_str = COMMA.join(aliased_string)
+            select_str = f"{STAR} EXCLUDE({exclude_str}), {aliased_str}"
```
sfc-gh-jdu, line 1276: can you make EXCLUDE also a variable like other constants in this file?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1258,10 +1261,23 @@ def pivot_statement(
...
             + RIGHT_PARENTHESIS
         )
+        if pivot_values is not None and should_alias_column_with_agg:
+            quoted_names = [quote_name(value) for value in pivot_values]
+            aliased_names = [
+                quote_name(f"{unwrap_single_quote(value)}_{aggregate}")
+                for value in pivot_values
+            ]
+            aliased_string = [
+                f"{quoted_name} AS {aliased_name}"
```
sfc-gh-jdu, line 1271: use the variable AS?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1258,10 +1261,23 @@ def pivot_statement(
             + (ANY if pivot_values is None else COMMA.join(pivot_values))
             + RIGHT_PARENTHESIS
         )
+        if pivot_values is not None and should_alias_column_with_agg:
+            quoted_names = [quote_name(value) for value in pivot_values]
+            aliased_names = [
+                quote_name(f"{unwrap_single_quote(value)}_{aggregate}")
```
sfc-gh-jdu, line 1267: can you add a comment about why unwrapping single quote?

sfc-gh-jdu (approved):

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-dyadav: done

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-dyadav: done

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-dyadav: done

sfc-gh-aalam (approved):

PR #3180, opened 2025-03-20
SNOW-1882075,SNOW-1990301,SNOW-1990302,SNOW-1821290: AST finalizations for stateless batches, and DataframeWriter,DataframeReader new APIs

sfc-gh-heshah:

   This PR includes all changes required to complete the following tickets.

   - [SNOW-1882075](https://snowflakecomputing.atlassian.net/browse/SNOW-1882075): AST for DataFrameWriter.insertInto/insert_into
   - [SNOW-1990301](https://snowflakecomputing.atlassian.net/browse/SNOW-1990301): AST for DataframerWriter.{format,save}
   - [SNOW-1990302](https://snowflakecomputing.atlassian.net/browse/SNOW-1990302): AST for DataframeReader.{format,load}
   - [SNOW-1821290](https://snowflakecomputing.atlassian.net/browse/SNOW-1821290): Finalize the AST schema v1

   It also includes changes in support of [SNOW-1997136](https://snowflakecomputing.atlassian.net/browse/SNOW-1997136): Ensure AST batch is stateless.

   AST generation for `DataframeReader` APIs which return the same `DataframeReader` instance have been updated to follow the same pattern as `DataframeWriter`. Since all such methods are setters, it is only necessary to have the last set attributes in the AST sent to the serveer. This is in support of the new `format` API in both Snowpark classes.

    AST generation was added for the new APIs `DataframeReader.{format,load}`, and `DataframeWriter.{format,read,insert_into}` along with expectation test updates.

    Changes were also made in support of AST batch statelessness in the IR, by changing fields typed `VarId` to be `Expr` typed. This allows more consistent usage of `_set_ast_ref` as a helper method across all Snowpark APIs.

    This PR is accompanied by a [server-side PR](https://github.com/snowflakedb/snowflake/pull/270350).

[SNOW-1882075]: https://snowflakecomputing.atlassian.net/browse/SNOW-1882075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-1990301]: https://snowflakecomputing.atlassian.net/browse/SNOW-1990301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-1990302]: https://snowflakecomputing.atlassian.net/browse/SNOW-1990302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-1821290]: https://snowflakecomputing.atlassian.net/browse/SNOW-1821290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snowflake/snowpark/_internal/ast/utils.py
+++ b/src/snowflake/snowpark/_internal/ast/utils.py
@@ -334,10 +334,7 @@ def build_expr_from_python_val(
             build_expr_from_python_val(ast.vs.add(), v)  # type: ignore[attr-defined] # TODO(SNOW-1491199) # "Expr" has no attribute "vs"
     elif isinstance(obj, snowflake.snowpark.dataframe.DataFrame):
         ast = with_src_position(expr_builder.dataframe_ref)  # type: ignore[arg-type] # TODO(SNOW-1491199) # Argument 1 to "with_src_position" has incompatible type "DataframeRef"; expected "Expr"
-        assert (
-            obj._ast_id is not None
-        ), "Dataframe object to encode as part of AST does not have an id assigned. Missing AST for object or previous operation?"
-        ast.id.bitfield1 = obj._ast_id  # type: ignore[attr-defined] # TODO(SNOW-1491199) # "Expr" has no attribute "id"
+        obj._set_ast_ref(expr_builder)
     elif isinstance(obj, snowflake.snowpark.table_function.TableFunctionCall):
         raise NotImplementedError(
             "TODO SNOW-1629946: Implement TableFunctionCall with args."
@@ -1090,21 +1087,6 @@ def fill_write_file(
             t = expr.copy_options.add()  # type: ignore[attr-defined] # TODO(SNOW-1491199) # "Expr" has no attribute "copy_options"
             t._1 = k
             build_expr_from_python_val(t._2, v)
-
-
-# TODO(SNOW-1491199) - This method is not covered by tests until the end of phase 0. Drop the pragma when it is covered.
-def build_proto_from_pivot_values(  # type: ignore[no-untyped-def] # TODO(SNOW-1491199) # Function is missing a return type annotation
-    expr_builder: proto.PivotValue,
-    values: Optional[Union[Iterable["LiteralType"], "DataFrame"]],  # type: ignore[name-defined] # noqa: F821 # TODO(SNOW-1491199) # Name "LiteralType" is not defined, Name "DataFrame" is not defined
-):  # pragma: no cover
-    """Helper function to encode Snowpark pivot values that are used in various pivot operations to AST."""
-    if not values:
-        return
-
-    if isinstance(values, snowflake.snowpark.dataframe.DataFrame):
-        expr_builder.pivot_value__dataframe.v.id.bitfield1 = values._ast_id
-    else:
-        build_expr_from_python_val(expr_builder.pivot_value__expr.v, values)
 
 
 # TODO(SNOW-1491199) - This method is not covered by tests until the end of phase 0. Drop the pragma when it is covered.
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -250,6 +250,7 @@ class Column:
     #       For example, running: df.filter(col("A").isin(1, 2, 3) & col("B")) would fail since the boolean operator
     #       '&' would try to construct an AST using that of the new col("A").isin(1, 2, 3) column (which we currently
     #       don't fill if the only argument provided in the Column constructor is 'expr1' of type Expression)
+    @publicapi
     def __init__(
         self,
         expr1: Union[str, Expression],
@@ -1523,6 +1524,7 @@ class CaseExpr(Column):
         [Row(CASE_WHEN_COLUMN=1), Row(CASE_WHEN_COLUMN=2), Row(CASE_WHEN_COLUMN=3)]
     """
 
+    @publicapi
     def __init__(
         self,
         expr: CaseWhen,
--- a/src/snowflake/snowpark/dataframe_writer.py
+++ b/src/snowflake/snowpark/dataframe_writer.py
@@ -90,14 +90,22 @@ class DataFrameWriter:
     def __init__(
         self,
         dataframe: "snowflake.snowpark.dataframe.DataFrame",
-        _ast_stmt: Optional[proto.Assign] = None,
+        _emit_ast: bool = True,
     ) -> None:
         self._dataframe = dataframe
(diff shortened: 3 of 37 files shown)


sfc-gh-heshah:
* **#3180** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3180?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3180?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sfc-gh-vbudati (approved):
Thanks Hemit!

src/snowflake/snowpark/column.py
```diff
@@ -250,6 +250,7 @@ class Column:
     #       For example, running: df.filter(col("A").isin(1, 2, 3) & col("B")) would fail since the boolean operator
     #       '&' would try to construct an AST using that of the new col("A").isin(1, 2, 3) column (which we currently
     #       don't fill if the only argument provided in the Column constructor is 'expr1' of type Expression)
```
sfc-gh-lspiegelberg, line 252: I recall there used to be some issue adding this to the constructor, do we have tests covering on/off for using `Column` objects (i.e. calling this API, and also `col`)?

src/snowflake/snowpark/_internal/ast/utils.py
```diff
@@ -1092,21 +1089,6 @@ def fill_write_file(
             build_expr_from_python_val(t._2, v)
```
sfc-gh-lspiegelberg, line 1091: what's the reason for removal here? Is this duplicate code?

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -225,6 +228,7 @@ def save_as_table(
         statement_params: Optional[Dict[str, str]] = None,
         block: bool = True,
         _emit_ast: bool = True,
```
sfc-gh-lspiegelberg, line 230: why do we need `_flush_ast` here?

sfc-gh-lspiegelberg (commented):
No questions re the minor changes, but confused about `_flush_ast`. What's the role of it? Why do we need it? Could we get away with not adding yet another parameter?

src/snowflake/snowpark/column.py
sfc-gh-heshah: Could you clarify, do you mean on/off in terms of our global AST enabled flag?

src/snowflake/snowpark/_internal/ast/utils.py
sfc-gh-heshah: We removed the `PivotValue` entity from the AST due to redundancy. It's simply either a Python constant or a Dataframe, both of which can be captured with a simple `Expr` field instead.

sfc-gh-heshah:
`save_as_table` and `copy_into_location` (to which the `_flush_ast` parameter was added) are both public APIs with eager evaluation and internal uses. However, most other `DataframeWriter` APIs rely on them to perform the eager evaluation (in terms of Snowpark execution), and to flush the AST.

Only using `_emit_ast` meant that internal callers can't ask `save_as_table` and `copy_into_location` to both suppress their AST statement generation and flush the AST batch, because the latter was still guarded by our `if _emit_ast` check.

Using the combination means that internal callers can set `_emit_ast` to False to suppress the AST generation from these APIs, while also setting `_flush_ast` to true to indicate that the eager request sent through the underlying connection should still include the AST batch.

It's a unique case: public APIs, eager (in terms of actually performing the execution within the function itself rather than in a helper), and used internally by other public APIs. I don't think combination exists anywhere else in the Snowpark codebase unless I'm mistaken. 

Elsewhere we seem to pass an `_ast` argument containing a pre-built `proto.Expr` itself to be inherited by the callee, so we do use another parameter. In this case I think `_flush_ast` makes more sense to indicate what's actually happening, however, we can follow the common pattern and just use an `_ast` parameter with the built `proto.Expr()` for `save_as_table` and `copy_into_table` to add to the `Request` before flushing.

sfc-gh-azwiegincew (approved):

src/snowflake/snowpark/column.py
sfc-gh-lspiegelberg: oh like with AST flag enabled/disabled.

sfc-gh-lspiegelberg:
> `save_as_table` and `copy_into_location` (to which the `_flush_ast` parameter was added) are both public APIs with eager evaluation and internal uses. However, most other `DataframeWriter` APIs rely on them to perform the eager evaluation (in terms of Snowpark execution), and to flush the AST.
> 
> Only using `_emit_ast` meant that internal callers can't ask `save_as_table` and `copy_into_location` to both suppress their AST statement generation and flush the AST batch, because the latter was still guarded by our `if _emit_ast` check.
> 
> Using the combination means that internal callers can set `_emit_ast` to False to suppress the AST generation from these APIs, while also setting `_flush_ast` to true to indicate that the eager request sent through the underlying connection should still include the AST batch.
> 
> It's a unique case: public APIs, eager (in terms of actually performing the execution within the function itself rather than in a helper), and used internally by other public APIs. I don't think combination exists anywhere else in the Snowpark codebase unless I'm mistaken.
> 
> Elsewhere we seem to pass an `_ast` argument containing a pre-built `proto.Expr` itself to be inherited by the callee, so we do use another parameter. In this case I think `_flush_ast` makes more sense to indicate what's actually happening, however, we can follow the common pattern and just use an `_ast` parameter with the built `proto.Expr()` for `save_as_table` and `copy_into_table` to add to the `Request` before flushing.

Hm, internal use of `save_as_table` should not generate AST/not flush it. My mind model here is that all of this moves to the server-side, i.e. what is right now internal will simply execute server-side.

src/snowflake/snowpark/column.py
sfc-gh-heshah: Oh, yes we do have tests in both cases then (public calls made by tests, and internal calls to the constructor) if I'm understanding correctly.

sfc-gh-lspiegelberg (approved):

PR #3185, opened 2025-03-21
SNOW-1980102: Support reading XML file with row tag

sfc-gh-jdu:

   

   Fixes SNOW-1980102

   Use Python UDTF to support reading XML files with a row tag.


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1285,6 +1288,48 @@ def process_list(list_property):
...
 
+    def _create_xml_query(
+        self, udtf: "UserDefinedTableFunction", file_path: str, options: Dict[str, str]
+    ) -> str:
+        """
+        Creates a DataFrame from a UserDefinedTableFunction that reads XML files.
+        """
+        from snowflake.snowpark.functions import lit, col, seq8, flatten
+
+        default_num_workers = 16
```
sfc-gh-jkew, line 1299: should the default number of workers be an option? I would think so - just because the SEs can then adjust things depending on the customer.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1285,6 +1288,48 @@ def process_list(list_property):
             )(setting["property_value"])
         return new_options
 
+    def _create_xml_query(
+        self, udtf: "UserDefinedTableFunction", file_path: str, options: Dict[str, str]
```
sfc-gh-jkew, line 1292: nit: "udtf" is too generic of a name; make it "xmlparser_udtf" or something.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1285,6 +1288,48 @@ def process_list(list_property):
...
+        # Flatten the Variant data to get the key-value pairs
+        df = df.select(
+            worker_column_name,
+            xml_row_number_column_name,
+            flatten(df[XML_ROW_DATA_COLUMN_NAME]),
+        ).select(worker_column_name, xml_row_number_column_name, "key", "value")
+        # Apply dynamic pivot to get the flat table with dynamic schema
+        df = (
+            df.pivot("key")
+            .max("value")
```
sfc-gh-jkew, line 1327: If I'm reading this properly; which I may not be - it looks like we split the XML file up into row-tag chunks; but then we run flatten on each chunk. This should handle even cases where we have rows with different schemas:

```
<books>
    <book><a>1234</a><b>5678</b></book>
    <book><a>1234</a><c>5678</c></book> 
    <book><b>5678</b><c>0987</c></book> 
</books>
```

But why do you apply a "max" here?

tests/resources/books.xml
```diff
@@ -0,0 +1,136 @@
+<?xml version="1.0"?>
```
sfc-gh-jkew, line 1: I would add a file with an XML Entity header to validate there's no processing of those components. This is to avoid the classic class of security issues

https://owasp.org/www-community/vulnerabilities/XML_External_Entity_(XXE)_Processing

sfc-gh-jkew (changes requested):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jkew: It's required, of course. And should be in a place where there are no duplicates at this point. I think I've answered my own questions.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: There is a scalability issue of UDTF where 16 seems the maximum number of partitions we can use now. I'm investigating it here https://snowflakecomputing.atlassian.net/browse/SNOW-1983360. I'll add a TODO comment here. Once it's resolved, I'll add it as an option.

tests/resources/books.xml
sfc-gh-jdu: Thanks for pointing it out. I'm wondering whether XXE really makes sense if the file is already on a Snowflake stage.
1) if the XML file references to a local file, it doesn't make sense because inside UDTF as there is no such a file. If users intends to read such an XML file, it must be done locally. This is not our P0 (but if the customer requests it, we can do it)
2) if the XML file references to remote file, it requires network access inside UDTF, which involves configurations of external access in Snowflake, which complicates our approach as it relates to some security policies. https://docs.snowflake.com/en/developer-guide/external-network-access/creating-using-external-network-access. 

So I'd prefer not to support it initially and clearly document it. We can consider the support if there are some requests from customers. WDYT?

sfc-gh-jdu: added such a test

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jdu: done

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -32,6 +32,9 @@ def replace_entity(match: re.Match) -> str:
     if entity in predefined:
         # Leave these untouched.
         return match.group(0)
+    elif entity == "xxe":
```
sfc-gh-jkew, line 35: This is an interesting solution; but if it works it works.

sfc-gh-jkew (approved):
Looks good

sfc-gh-aling (approved):
LGTM, btw the coverage report is failing, can you double check/add coverage before merging?

PR #3183, opened 2025-03-21
SNOW-1821290: Finalize AST schema v1 by refactoring Assign to Bind, VarId to BindId

sfc-gh-heshah:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   [SNOW-1821290](https://snowflakecomputing.atlassian.net/browse/SNOW-1821290): Finalize AST schema v1 by refactoring `Assign` to `Bind`, `VarId` to `BindId`

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.
   - [x] I acknowledge that I have ensured my changes to be thread-safe. Follow the link for more information: [Thread-safe Developer Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#thread-safe-development)
   - [x] If adding any arguments to public Snowpark APIs or creating new public Snowpark APIs, I acknowledge that I have ensured my changes include AST support.

3. Please describe how your code solves the related issue.

   To reflect the new semantics, `AstBatch.assign` was refactored to `AstBatch.bind` and its usage was updated across the client. Usage of `var_id` fields was replaced with `bind_id` fields as well.

   `AstBatch._init_batch` now provides the UUID to the newly created `Request` message  in its binary 16 byte representation provided by `UUID.bytes` under the `Request.request_id` field.

   `AstBatch.bind` will also take the current `AstBatch._request_id` and fill `Bind.first_request_id` with the same binary 16 byte representation from `UUID.bytes` to keep track of the first `Request` message which this statement will be flushed to the server in.

   This PR is accompanied by a [server-side PR](https://github.com/snowflakedb/snowflake/pull/270978).

[SNOW-1821290]: https://snowflakecomputing.atlassian.net/browse/SNOW-1821290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3201, opened 2025-03-26
SNOW-1997136: Ensure transitive closure in AST Request messages

sfc-gh-heshah:

   [SNOW-1997136](https://snowflakecomputing.atlassian.net/browse/SNOW-1997136): Ensure transitive closure in AST Request messages.

3. Fill out the following pre-review checklist:

4. Please describe how your code solves the related issue.

   This PR introduces an AST handler class `AstBuilder` which will be carried by all Snowpark objects that need to generate requests to be sent to the server side (`Table`, `Dataframe`, `UDxF/Sproc` registration, etc).

    Requesting a review of just the implementation for now, the refactor to remove `Session._ast_batch` and use this handler class across the client will be done in a separate PR.

    The dependency graph between `AstBuilder` instances will be created via the `_set_ast_ref` helpers we have introduced to Snowpark classes via the `add_required_by` or `add_dependency` methods in the new `AstBuilder` class.


sfc-gh-heshah:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3201?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#3201** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3183** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3183?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3180** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3180?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns:
+            An instance of AstBuilder with a bind statement initialized.
+        """
+        b = cls()
+        b._bind = proto.Bind()
+        b._bind.uid = next(AstBuilder.__id_gen)
+        b._bind.symbol.value = symbol if isinstance(symbol, str) else ""
+        return b
+
+    def add_dependency(
```
sfc-gh-lspiegelberg, line 70: nit: maybe `depends_on`?

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,176 @@
...
+        self, dependency: "AstBuilder", add_inverse: bool = False
```
sfc-gh-lspiegelberg, line 71: why is `add_inverse` needed? Can we remove it?

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Args:
+            dependency: An AstBuilder instance that is a dependency of the current instance.
+        """
+        if dependency not in self.__dependencies:
+            self.__dependencies.append(dependency)
+        if add_inverse:
+            dependency.add_required_by(self, add_inverse=False)
+
+    def add_required_by(
```
sfc-gh-lspiegelberg, line 85: what is the difference between "depends_on" and "required_by"?

sfc-gh-heshah: "depends_on" corresponds to child statements for this one (dependencies)
"required_by" corresponds to parent statements that use this one

src/snowflake/snowpark/_internal/ast/builder.py
sfc-gh-heshah: Not necessary unless we'd like to keep track of both parent and child relations between AST statements.
Just an easy way to avoid needing another call into an `AstBuilder` method.

sfc-gh-lspiegelberg: Oh, I was more referring to the interface. I actually like having the double-linked relationship (i.e., you can access children/parents flexible). Would always do it/enforce it, but remove the parameter `add_inverse` from the interface. There's no need for this to exist/be an option really imho.

src/snowflake/snowpark/_internal/ast/builder.py
sfc-gh-lspiegelberg: ah, I think you want to have a single mechanism for dependence. You can define a single interface, and manage both with it! I.e.,

Have object A depend on B: A.depends_on(B).  Now the question is, do you need directed depdence? I.e.,

A -> B, B->A, A <-> B? are these different cases?

I would argue, we mostly care about a DAG structure here, i.e. A -> B only. This will make a lot of things easier, because the resulting graph can be topologically ordered (which I believe is what we want for the closure).

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,159 @@
...
+            if assign_first_request_ids and not ast_handle._bind.first_request_id:
+                # Assign the first request IDs for bind statements if not already set.
+                ast_handle._bind.first_request_id = current_request_id.bytes
+
+            # Add the bind and eval statements to the request body.
+            request.body.append(ast_handle._bind)
+            if ast_handle.__eval is not None:
+                request.body.append(ast_handle.__eval)
+
+        if self.__eval is None:
```
sfc-gh-lspiegelberg, line 146: this makes the function non-const and have side effects. Could we maybe not overwrite internal `self.__eval`? I actually think you don't need to even add __eval to the class. Could be fine to add here (e.g., introduce param `eval=False` and then explicitly call with `eval=True` in the source code). Wdyt?

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,159 @@
...
+        closure = []
+
+        while q:
+            ast_builder = q.pop()
+            closure.append(ast_builder)
+            visited.add(ast_builder)
+            for dep in ast_builder.__dependencies:
+                if dep not in visited:
+                    q.append(dep)
+        closure.reverse()
```
sfc-gh-lspiegelberg, line 105: nit: this code is fine with the reverse, alternative is to prepend instead of append.

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Args:
+            dependency: An AstBuilder instance that is a dependency of the current instance.
+        """
+        if dependency not in self.__dependencies:
+            self.__dependencies.append(dependency)
+        if self not in dependency.__required_by:
+            dependency.__required_by.append(self)
+
+    @staticmethod
```
sfc-gh-lspiegelberg, line 83: why static? can we do `def _closure(self)` instead?

src/snowflake/snowpark/_internal/ast/builder.py
sfc-gh-heshah: I don't think `__eval` is necessary either in this class either since we are moving to globally unique Bind IDs, unless we'd like to include previously generated Eval statements in future batches as a hint to the server. Otherwise I agree we should move to introduce an `eval=False` param to this method.

sfc-gh-lspiegelberg: Ok, then let's remove the __eval here (or add with eval=False/True). Good to have stateless/not self modifying function here.

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,154 @@
...
+# The current AST version number (generated by the DSL).
+CLIENT_AST_VERSION = proto.__Version__.MAX_VERSION
+
+
+SerializedBatch = namedtuple("SerializedBatch", ["request_id", "batch"])
+
+
+class AstBuilder:
+    """
+    A handler for AST statements that enables a Snowpark oject (e.g. Dataframe, Table, etc.) to carry its own AST,
```
sfc-gh-oplaton, line 26:
```suggestion
    A handler for AST statements that enables a Snowpark object (e.g. Dataframe, Table, etc.) to carry its own AST,
```

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # Function used to generate request IDs. This is overridden in some tests.
+    __generate_request_id = uuid.uuid4
+
+    # Class variable for generating IDs for bind statements.
+    __id_gen = itertools.count(start=1)
+
+    # NOTE: itertools.count and its __next__ method are only thread-safe in CPython with GIL.
+    # In other implementations (Jython, PyPy), this may not be thread-safe, and once the GIL is removed from CPython, it will not be thread-safe either.
+    # Likely, it would be best to make all of the methods in this class thread-safe with threading.Rlock
```
sfc-gh-oplaton, line 42: I would suggest simplifying this comment to "This class is not thread safe."

In particular, if the class is not thread safe, then it's not thread safe in Python implementations with a GIL, either.

src/snowflake/snowpark/_internal/ast/builder.py
```diff
@@ -0,0 +1,154 @@
...
+            ast_builder = q.pop()
+            closure.appendleft(ast_builder)
+            visited.add(ast_builder)
+            for dep in ast_builder.__dependencies:
+                if dep not in visited:
+                    q.append(dep)
+
+        return closure
+
+    def to_request(
```
sfc-gh-lspiegelberg, line 106: nit: I think we may want to split this up into `to_request() -> proto.Request()` and `to_batch() -> SerializedBatch`. It's a bit weird that to_request returns a batch.
E.g., refactor into:
```
def to_request( self, eval: bool = False, assign_first_request_ids: bool = True
    ) -> proto.Request:
        """
        Builds and returns a new AST request with transitive closure of all dependencies.
        """
        request = proto.Request()

        # Generate a new unique ID and convert to bytes for the request message field.
        current_request_id: uuid.UUID = AstBuilder.__generate_request_id()
        request.id = current_request_id.bytes

        # Initialize the request with the current client version, Python version, and AST version.
        (major, minor, patch) = VERSION
        request.client_version.major = major
        request.client_version.minor = minor
        request.client_version.patch = patch

        (major, minor, micro, releaselevel, serial) = sys.version_info
        request.client_language.python_language.version.major = major
        request.client_language.python_language.version.minor = minor
        request.client_language.python_language.version.patch = micro
        request.client_language.python_language.version.label = releaselevel

        request.client_ast_version = CLIENT_AST_VERSION

        # Get the transitive closure of the current AST builder and its dependencies.
        # Add all statements in the closure to the request body with unique IDs.
        closure = self._closure()
        for ast_handle in closure:
            if assign_first_request_ids and not ast_handle._bind.first_request_id:
                # Assign the first request IDs for bind statements if not already set.
                ast_handle._bind.first_request_id = current_request_id.bytes

            # Add the bind and eval statements to the request body.
            request.body.append(ast_handle._bind)

        if eval:
            # Generate the eval statement for the current AstBuilder instance's Bind statement.
            stmt = request.body.add()
            stmt.eval.bind_id = self._bind.uid

        from snowflake.snowpark._internal.ast.utils import fill_interned_value_table

        fill_interned_value_table(request.interned_value_table)
        return request
        
def to_batch(self, eval: bool = False, assign_first_request_ids: bool=True) -> SerializedBatch:
        # Serialize the request to a base64-encoded string for transmission.
        batch = str(base64.b64encode(self.to_request(eval, first_request_ids).SerializeToString()), "utf-8")
        return SerializedBatch(current_request_id, batch)
```

sfc-gh-lspiegelberg (approved):

sfc-gh-oplaton (approved):

PR #3203, opened 2025-03-27
SNOW-1923632: [Local Testing] Add support for Intervals to range_between

sfc-gh-jrose:

   

   Fixes SNOW-1923632


src/snowflake/snowpark/mock/_plan.py
```diff
@@ -314,6 +316,80 @@ def handle_order_by_clause(
...
+        "NANOSECOND",
+    ]:
+        lower_part = lower.values_dict.get(part)
+        upper_part = upper.values_dict.get(part)
+
+        if lower_part or upper_part:
+            lower_part = lower_part or 0
+            upper_part = upper_part or 0
+            if (lower_part + upper_part) != 0:
+                return lower_part > upper_part
```
sfc-gh-aling, line 372: depending on the desc, shouldn't this be?
```suggestion
                return lower_part >= upper_part
```

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -353,22 +429,70 @@ def search_boundary_idx(idx, delta, _win):
...
+                ):
+                    if lower:
+                        lower = Interval(day=-lower)
+                    if upper:
+                        upper = Interval(day=-upper)
+                if not (validate_interval(lower) and validate_interval(upper)):
+                    raise SnowparkLocalTestingException(
+                        "Mixing numeric and interval frames is not supported"
+                    )
+                if not compare_intervals(lower, upper):
```
sfc-gh-aling, line 451: I get a bit confused here on the `if not compare_intervals(lower, upper)`:

1. when the input is DateType, we negate only the "day" part, what about "year", "month"?
2. we negate the interval when group_col is DateType, we expect lower <= upper
    we do not negate for TimestampType type, we still expect lower <= upper

 do you have an example to help me understand the logic here?

src/snowflake/snowpark/mock/_plan.py
sfc-gh-jrose: You are correct. Fixed an added a test case that ensure this logic is correct.

src/snowflake/snowpark/mock/_plan.py
sfc-gh-jrose: 1. When the input is a DateType column and the parameters for rows_between are integers then the integers represent days. Eg. For an integer column the following window would actually mean +/- 1, but the semantics for a date column make it +/- 1 day.
```
Window.order_by(col("a")).range_between(-1, 1)
```

2. For TimestampType the example window above makes on sense. Therefore timestamp windows will always either have bounds of one of these types: `{0, None, UnboundPreceding/Following, Interval}`. Those will have already been negated as appropriately around line 2644 in this file. Datetype is a special case because of the extra semantics it has for integers.

sfc-gh-aling: gotcha, it is much clear to me now. thanks for the context!

sfc-gh-aling (approved):

PR #3204, opened 2025-03-27
SNOW-2012666 lazy pandas import

sfc-gh-eretunsky:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/type_utils.py
+++ b/src/snowflake/snowpark/_internal/type_utils.py
@@ -35,7 +35,6 @@ import snowflake.snowpark.context as context
 import snowflake.snowpark.types  # type: ignore
 from snowflake.connector.constants import FIELD_ID_TO_NAME
 from snowflake.connector.cursor import ResultMetadata
-from snowflake.connector.options import installed_pandas, pandas
 from snowflake.snowpark._internal.utils import quote_name
 from snowflake.snowpark.row import Row
 from snowflake.snowpark.types import (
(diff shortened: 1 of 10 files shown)


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -77,6 +76,30 @@
     File,
 )
 
+imported_installed_pandas = None
+def lazy_import_pandas():
+    if imported_installed_pandas is None:
+        from snowflake.connector.options import installed_pandas, pandas
+        import numpy
+
+        imported_installed_pandas = installed_pandas
```
sfc-gh-yixie, line 85: This might have a problem because `lazy_import_pandas` may be used in multiple threads.
I think Python interpreter will import a package only once even if you call the import multiple times. The import itself is also thread-safe. So I guess a global variable isn't needed.

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -85,14 +108,6 @@
 except ImportError:
     from collections.abc import Iterable  # noqa: F401
 
-if installed_pandas:
```
sfc-gh-yixie, line 88: Is type_utils.py used when a session is created? If no I feel this can be kept at module level.

src/snowflake/snowpark/session.py
```diff
@@ -221,6 +219,18 @@
 if TYPE_CHECKING:
     import modin.pandas  # pragma: no cover
 
+def lazy_import_pyarrow():
+    from snowflake.connector.options import pyarrow
+
+imported_installed_pandas = None
+def lazy_import_pandas():
```
sfc-gh-yixie, line 226: Other modules also have similar or the same functions. How about using an internal module so it's written only once?

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-eretunsky: It's more to make this a no-op on subsequent calls. So it wouldn't hurt.

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-eretunsky: I'd like to separate the function of this PR. It's all about lazy import, no additional optimizations.

src/snowflake/snowpark/session.py
sfc-gh-eretunsky: The problem is that in every file I made this change the imports are slightly different. So I kept if file local for now, at least until we ensure it still works and the import time reduced. After that we can have a follow-up to optimize this.

PR #3207, opened 2025-03-28
SNOW-2014383: Add pandas tag to eager queries to account for their credit usage

sfc-gh-helmeleegy:

   

   Fixes SNOW-2014383

   Add pandas tag to eager queries to account for their credit usage.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
@@ -1758,7 +1758,7 @@ def groupby_apply_sort_method(
         sp_func.builtin("boolor_agg")(
             SnowparkColumn(original_row_position_quoted_identifier) == -1
         ).as_("is_transform")
-    ).collect()[0][0]
+    ).collect(statement_params=get_default_snowpark_pandas_statement_params())[0][0]
     return (
         GroupbyApplySortMethod.ORIGINAL_ROW_ORDER
         if is_transform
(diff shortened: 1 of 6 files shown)


sfc-gh-mvashishtha (commented):
@sfc-gh-helmeleegy why not make `collect()` add the pandas tag by default?

Also, can we add some test coverage?

sfc-gh-helmeleegy:
> @sfc-gh-helmeleegy why not make `collect()` add the pandas tag by default?
> 
> Also, can we add some test coverage?

Good point. Actually, it is already adding it by default. So I guess these changes won't be as effective as initially thought.

sfc-gh-nkumar:
> @sfc-gh-helmeleegy why not make `collect()` add the pandas tag by default?
> 
> Also, can we add some test coverage?

Agree, most of them can be solved by adding pandas tags to `OrderedDataFrame.collect`.  There are some cases where we are directly using SnowparkDataframe.collect which will still require adding these params at caller (this is anti-pattern, ideally we should not be accessing underlying snowpark df directly and always use ordered dataframe but fixing that maybe out of scope of this PR)

sfc-gh-helmeleegy:
> > @sfc-gh-helmeleegy why not make `collect()` add the pandas tag by default?
> > Also, can we add some test coverage?
> 
> Agree, most of them can be solved by adding pandas tags to `OrderedDataFrame.collect`. There are some cases where we are directly using SnowparkDataframe.collect which will still require adding these params at caller (this is anti-pattern, ideally we should not be accessing underlying snowpark df directly and always use ordered dataframe but fixing that maybe out of scope of this PR)

Yes, I'll close this PR and create another one for the one mistakenly using SnowparkDataframe.

PR #3227, opened 2025-04-03
Propose cache_result for moving to Snowflake to avoid repeatedly inserting values into a temp table. This will make move_to snowflake eager, which I think is more sensible.

sfc-gh-mvashishtha:
Propose cache_result for moving to Snowflake to avoid repeatedly inserting values into a temp table. This will make move_to snowflake eager, which I think is more sensible.


sfc-gh-jkew (approved):
This is a lot more simple than I expected.

PR #3235, opened 2025-04-04
NOSNOW: Refactor functions.py for AST generation

sfc-gh-heshah:

   No JIRA ticket, but required for Phase 0 Client Closeout.

   Due to widespread internal use of the `functions.builtin` method (an alias for `functions.function`), as well as `functions.call_builtin` (an alias for `functions.call_function`), we had set up the propagation of the AST directly through these methods down to the eventual call to the internal helper `_call_function`.

   As a result, despite these being public APIs which are required to support Snowflake SQL functions not explicitly supported in Snowpark, we could not capture an AST to represent `call_function("foo", *args)` when no such `functions.foo` method exists in Snowpark.

   Public APIs should NOT be accepting `_ast` parameters to route generated ASTs through them as this complicates the scenario in which `_ast` is not provided but `_emit_ast` is `True`. If this Public API is used extensively internally, we then have an unresolvable question of whether to route `_ast=None` to the eventual internal helper (e.g. `_call_function`) which **will** eventually perform AST generation based on the other arguments, or whether to generate the Public API's own AST and pass it to the internal helper function.

   The only resolutions to the above issue are
     - To add **another** parameter to the extensively internally used public API to signal whether to pass down the provided input `_ast` parameter (regardless of whether it is `None` or not), or whether to confidently generate its own AST and assign it to the result.
     - To avoid the use of public APIs internally entirely.

   This PR takes the latter approach. I argue that using public APIs internally as helper functions is bad practice to motivate this refactor. Either we extract the common functionality into a separate internal helper function to be used by all Public APIs which rely on it, or (especially in this case) if all calls are eventually routed into the same internal helper (like `_call_function`), then we should avoid adding unnecessary calls to wrappers to the stack internally and call this helper directly.

   The refactor is quite simple overall but `functions.py` is large:
    - Replace all internal uses of `builtin/function` and `call_builtin/call_function` with direct calls to `_call_function`
    - Simplify AST generation wherever possible by using `build_function_expr`
    - Remove `_ast` parameters from public APIs and leave the assignment of the AST to the result as the caller's responsibility.

### DIFF
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -170,7 +170,6 @@ import snowflake.snowpark._internal.proto.generated.ast_pb2 as proto
 import snowflake.snowpark.table_function
 from snowflake.snowpark._internal.analyzer.expression import (
     CaseWhen,
-    Expression,
     FunctionExpression,
     Interval,
     ListAgg,
@@ -191,8 +190,6 @@ from snowflake.snowpark._internal.ast.utils import (
     build_call_table_function_apply,
     build_expr_from_snowpark_column_or_python_val,
     build_expr_from_snowpark_column_or_sql_str,
-    set_builtin_fn_alias,
-    snowpark_expression_to_ast,
     with_src_position,
     build_function_expr,
 )
@@ -489,13 +486,20 @@ def system_reference(
     privileges = privileges or []
     ast = (
         build_function_expr(
-            "system_reference", [object_type, object_identifier, scope, privileges]
+            "system_reference",
+            [object_type, object_identifier, scope] + privileges,
         )
         if _emit_ast
         else None
     )
-    return builtin("system$reference", _ast=ast, _emit_ast=_emit_ast)(
-        object_type, object_identifier, scope, *privileges
+    return _call_function(
+        "system$reference",
+        object_type,
+        object_identifier,
+        scope,
+        *privileges,
+        _ast=ast,
+        _emit_ast=_emit_ast,
     )
 
 
@@ -510,7 +514,7 @@ def current_session(_emit_ast: bool = True) -> Column:
         >>> result = session.create_dataframe([1]).select(current_session()).collect()
         >>> assert result[0]['CURRENT_SESSION()'] is not None
     """
-    return builtin("current_session", _emit_ast=_emit_ast)()
+    return _call_function("current_session", _emit_ast=_emit_ast)
 
 
 @publicapi
@@ -523,7 +527,7 @@ def current_statement(_emit_ast: bool = True) -> Column:
         >>> session.create_dataframe([1]).select(current_statement()).collect()
         [Row(CURRENT_STATEMENT()='SELECT current_statement() FROM ( SELECT "_1" FROM ( SELECT $1 AS "_1" FROM  VALUES (1 :: INT)))')]
     """
-    return builtin("current_statement", _emit_ast=_emit_ast)()
+    return _call_function("current_statement", _emit_ast=_emit_ast)
 
 
 @publicapi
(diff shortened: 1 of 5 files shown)


sfc-gh-heshah:
* **#3235** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3235?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3235?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sfc-gh-oplaton (approved):

tests/ast/data/functions1.test
```diff
@@ -362,9 +362,9 @@ df27 = df.select(bitshiftright("A", col("B")), bitshiftright("A", -10), bitshift
 
 df28 = df.select(bround("A", 10), bround("A", 2), bround(col("A"), col("B")))
 
-df29 = df.select(convert_timezone("A", col("B"), None), convert_timezone(col("A"), "B", None), convert_timezone("A", "B", None), convert_timezone(col("A"), col("B"), None), convert_timezone("A", "B", None), convert_timezone("A", "B", "A"))
```
sfc-gh-aalam, line 365: did we add `None` as the last pos arg here as a hack to make things work earlier? Does your refactor fix that?

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -666,13 +670,16 @@ def function(self, agg_name: str, _emit_ast: bool = True) -> Callable:
 
     builtin = function
 
+    @publicapi
     def _function(
```
sfc-gh-aalam, line 674: why is this a publicapi?

src/snowflake/snowpark/relational_grouped_dataframe.py
```diff
@@ -692,6 +699,7 @@ def _function(
 
         return df
 
+    @publicapi
     def _non_empty_argument_function(
```
sfc-gh-aalam, line 703: ditto

src/snowflake/snowpark/functions.py
```diff
@@ -11663,9 +11703,8 @@ def fl_get_content_type(e: ColumnOrName, _emit_ast: bool = True) -> Column:
         'text/csv'
     """
     function_name = "fl_get_content_type"
-    ast = build_function_expr(function_name, [e]) if _emit_ast else None
     col_input = _to_col_if_str(e, function_name)
-    return builtin(function_name, _ast=ast, _emit_ast=_emit_ast)(col_input)
+    return _call_function(function_name, col_input, _emit_ast=_emit_ast)
```
sfc-gh-aalam, line 11707: why are we dropping ast here and in some other places down below?

tests/ast/data/functions1.test
sfc-gh-heshah: This is actually because we were including a `None` arg in our protobuf message, so the Unparser was generating the `None` here. My refactor fixes this by removing the addition of any arguments that are `None` by default from the protobuf message.

src/snowflake/snowpark/relational_grouped_dataframe.py
sfc-gh-heshah: I mainly added the decorator as a safeguard that checks the global AST enabled flag before trying to generate the AST by default, rather than to signal that this method can be used publicly.

src/snowflake/snowpark/relational_grouped_dataframe.py
sfc-gh-heshah: Same here, just guards us by overriding `_emit_ast` to be false if the global AST enabled flag is off.

src/snowflake/snowpark/functions.py
sfc-gh-heshah: In all of these cases, the AST generation for the functions can be handled by `_call_function` directly as the inputs are simple (no conversions of constants to Snowpark `Literal`s), and the function name matches the SQL function name provided to `_call_function`.

PR #3248, opened 2025-04-09
SNOW-2014299: Query XML dataframe with dot notation

sfc-gh-jdu:

   

   Fixes SNOW-2014299

   This PR does three things: 1) add a private property to DataFrame class to indicate whether all columns are variant 2) When calling select() on a dataframe, if this property is true, it means all columns are variants and we should support query nested fields using dot notation 3) Handling double quotes when querying nested fields using dot notation

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -1403,7 +1403,7 @@ def split_dot_string(s: str) -> list:
           -> ['"a.b"', 'c', '"d.e.f"', 'g']
     """
     # ensures that dots inside quotes are not used for splitting.
-    parts = re.split(r'\.(?=(?:[^"]*"[^"]*")*[^"]*$)', s)
+    parts = re.compile(r'"(?:[^"]|"")*"|[^.]+').findall(s)
     return parts
 
 
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -1391,10 +1391,11 @@ def escape_quotes(unescaped: str) -> str:
     return unescaped.replace(DOUBLE_QUOTE, DOUBLE_QUOTE + DOUBLE_QUOTE)
 
 
-def split_dot_string(s: str) -> list:
+def split_snowflake_identifier_with_dot(s: str) -> list:
     """
-    Splits the string by dots that are not within double-quoted parts.
+    Splits the Snowflake identifier by dots that are not within double-quoted parts.
     Tokens that appear quoted in the input remain unchanged (quotes are kept).
+    See details in https://docs.snowflake.com/en/sql-reference/identifiers-syntax.
 
     Examples:
       'foo.bar."hello.world".baz'
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -1391,6 +1391,22 @@ def escape_quotes(unescaped: str) -> str:
     return unescaped.replace(DOUBLE_QUOTE, DOUBLE_QUOTE + DOUBLE_QUOTE)
 
 
+def split_dot_string(s: str) -> list:
+    """
+    Splits the string by dots that are not within double-quoted parts.
+    Tokens that appear quoted in the input remain unchanged (quotes are kept).
+
+    Examples:
+      'foo.bar."hello.world".baz'
+          -> ['foo', 'bar', '"hello.world"', 'baz']
+      '"a.b".c."d.e.f".g'
+          -> ['"a.b"', 'c', '"d.e.f"', 'g']
+    """
+    # ensures that dots inside quotes are not used for splitting.
+    parts = re.split(r'\.(?=(?:[^"]*"[^"]*")*[^"]*$)', s)
+    return parts
+
+
 # Define the full-width regex pattern, copied from Spark
 full_width_regex = re.compile(
     r"[\u1100-\u115F"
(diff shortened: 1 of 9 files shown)


sfc-gh-jkew (changes requested):

tests/unit/test_internal_utils.py
```diff
@@ -101,3 +102,23 @@ def test_generate_random_alphanumeric():
         futures = [executor.submit(generate_random_alphanumeric) for _ in range(5)]
         res = [f.result() for f in futures]
         assert len(set(res)) == 5  # no duplicate string
+
+
+@pytest.mark.parametrize(
+    "string, expected_result",
+    [
+        ('foo.bar."hello.world".baz', ["foo", "bar", '"hello.world"', "baz"]),
```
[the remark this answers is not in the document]
sfc-gh-jdu: thanks for the examples, added

src/snowflake/snowpark/dataframe.py
```diff
@@ -1541,7 +1556,14 @@ def select(
...
             if isinstance(e, Column):
-                names.append(e._named())
+                if self._all_variant_cols:
+                    names.append(
+                        Column(
+                            e._expr1, e._expr2, e._ast, _is_qualified_name=True
+                        )._named()
+                    )
+                else:
+                    names.append(e._named())
```
sfc-gh-aling, line 1566: can you help me remind the details -- do we parse xml and store content as object in VariantType column?

tests/unit/test_internal_utils.py
```diff
@@ -101,3 +102,31 @@ def test_generate_random_alphanumeric():
...
+            ['"quoted with ""embedded"" quotes"', "end"],
+        ),
+        (
+            '"quoted with ""embedded.and.some.dots"" quotes".end',
+            ['"quoted with ""embedded.and.some.dots"" quotes"', "end"],
+        ),
+        (
+            '"quoted with ""embedded.and.some.dots"" quotes escaped".end',
+            ['"quoted with ""embedded.and.some.dots"" quotes escaped"', "end"],
+        ),
```
sfc-gh-aalam, line 128: can we also add a case where we have open quotes
`'"open.quotes.end'`

src/snowflake/snowpark/dataframe.py
```diff
@@ -611,6 +612,20 @@ def __init__(
...
 
+        # Whether all columns are VARIANT data type,
+        # which support querying nested fields via dot notations
+        # If SQL simplifier is enabled, we need to get logical plan from plan.from_
+        if isinstance(plan, (SelectStatement, MockSelectStatement)):
+            self._all_variant_cols = (
+                isinstance(plan.from_, SelectSnowflakePlan)
+                and isinstance(plan.from_.snowflake_plan.source_plan, ReadFileNode)
+                and plan.from_.snowflake_plan.source_plan.xml_reader_udtf is not None
+            )
```
sfc-gh-aalam, line 623: wouldn't we be missing some cases here where dataframe might have all variant cols but it is not created from a read file operation? If we only care about the cases where source is a `ReadFileNode`, should we rename the internal variable to say that?

sfc-gh-jdu: for now, we only care about the dataframe generated by reading a XML file (see https://docs.google.com/document/d/11J2bL6QTpDs0DnMJZAq5EUNUCdipB9kfzojdlX_sCuU/edit?tab=t.0#heading=h.gky25x9oveeg), so querying nested fields in variant data using dot notation can be supported. In the future, we might also want to support such dot notation when all columns are variant (e.g., we can run a describe query to get types, which is an extra cost), so I had this variable name

src/snowflake/snowpark/dataframe.py
sfc-gh-jdu: yes

tests/unit/test_internal_utils.py
sfc-gh-jdu: done

sfc-gh-jkew (approved):

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1391,6 +1391,22 @@ def escape_quotes(unescaped: str) -> str:
...
+    Tokens that appear quoted in the input remain unchanged (quotes are kept).
+
+    Examples:
+      'foo.bar."hello.world".baz'
+          -> ['foo', 'bar', '"hello.world"', 'baz']
+      '"a.b".c."d.e.f".g'
+          -> ['"a.b"', 'c', '"d.e.f"', 'g']
+    """
+    # ensures that dots inside quotes are not used for splitting.
+    parts = re.split(r'\.(?=(?:[^"]*"[^"]*")*[^"]*$)', s)
```
sfc-gh-jrose, line 1406: The repetition of `[^"]*"` makes me wonder if this is susceptible to catastrophic backtracking.

sfc-gh-jdu: very good point, it should be fixed now

sfc-gh-jrose: That does look faster, but I think it might not cover all the same cases. I think this example should work like this, but the what manually split as expected does not match the actual result.
eg:
```
example = 'a."b."c".d."e.f.g".h".i.j."k"".""l."'
expected = ['a', '"b."c".d."e', 'f', 'g".h"', 'i', 'j', '"k"".""l."']

actual = ['a', '"b."', 'c"', 'd', '"e.f.g"', 'h"', 'i', 'j', '"k"".""l."']
```

sfc-gh-jrose (approved):

PR #3250, opened 2025-04-10
SNOW-2034858: to_pandas support non-select statement

sfc-gh-yuwang:

   

   Fixes SNOW-2034858


sfc-gh-jrose (commented):
Are there any cases when the conversion to padnas dataframe could fail? It might be worthwhile to still raise the error if the result is not some collection of row objects.

sfc-gh-yuwang:
> Are there any cases when the conversion to padnas dataframe could fail? It might be worthwhile to still raise the error if the result is not some collection of row objects.

it is possible, but I think that is already guarded by the code in server_connection: https://github.com/snowflakedb/snowpark-python/blob/47ee32c51b6ea1914ffcd6a8a541949918a18cd0/src/snowflake/snowpark/_internal/server_connection.py#L584

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1175,17 +1175,6 @@ def get_temp_type_for_object(use_scoped_temp_objects: bool, is_generated: bool)
...
-    if not isinstance(result, pandas.DataFrame):
-        raise SnowparkClientExceptionMessages.SERVER_FAILED_FETCH_PANDAS(
-            "to_pandas() did not return a pandas DataFrame. "
-            "If you use session.sql(...).to_pandas(), the input query can only be a "
-            "SELECT statement. Or you can use session.sql(...).collect() to get a "
-            "list of Row objects for a non-SELECT statement, then convert it to a "
-            "pandas DataFrame."
-        )
-
-
```
sfc-gh-aling, line 1188: can we understand why there was such a check at the very beginning?
I'm thinking removing this check would have unexpected side effect

src/snowflake/snowpark/dataframe.py
```diff
@@ -1040,7 +1041,8 @@ def to_pandas(
         # this might happen when calling this method with non-select commands
         # e.g., session.sql("create ...").to_pandas()
         if block:
-            check_is_pandas_dataframe_in_to_pandas(result)
+            if not isinstance(result, pandas.DataFrame):
+                return pd.DataFrame(result)
```
sfc-gh-aling, line 1045: what's the context of the change? 

also can you check the connector that is the result guaranted to be a dict or something that can be put into a pandas DF?

sfc-gh-jdu (approved):
How about to_pandas_batches?

src/snowflake/snowpark/dataframe.py
sfc-gh-yuwang: it is guaranteed to be data or iter or just raise error as it is guarded by the code in server_connection:
https://github.com/snowflakedb/snowpark-python/blob/47ee32c51b6ea1914ffcd6a8a541949918a18cd0/src/snowflake/snowpark/_internal/server_connection.py#L578
context is here: https://snowflake.slack.com/archives/C01BD4CSEFN/p1744278063299269

src/snowflake/snowpark/_internal/utils.py
sfc-gh-yuwang: this check is meant to make sure user do not call to_pandas on non-select statements.
As for reason, I think @sfc-gh-jdu  can answer

sfc-gh-yuwang:
> How about to_pandas_batches?

to_pandas_batches have a different code path so I think we should change it in this PR

src/snowflake/snowpark/dataframe.py
```diff
@@ -26,6 +26,8 @@
     overload,
 )
 
+import pandas as pd
```
sfc-gh-aling, line 29: this will cause error when pandas is not installed -- pandas is not a required dependency to snowpark

you should use the pandas from `snowflake.connector.options` which is already imported

sfc-gh-aling (approved):

PR #3264, opened 2025-04-15
SNOW-2042660: Update modin dependency to modin 0.32.0

sfc-gh-joshi:

   

   Fixes SNOW-2042660

To our surprise, modin 0.32.0 showed up in the Anaconda Snowflake channel today, and because the Snowsight package picker dropdown chooses the latest version of a package (ignoring solver constraints) by default, customers must currently manually select modin 0.30.1 to get the desired behavior.

Most changes are pretty minor, only thing worth noting is this version of modin officially includes support for numpy 2.0+, which causes some errors in tests using deprecated items like `np.NaN` (rather than `np.nan`).

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -530,8 +530,12 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
     this class is best explained by looking at https://github.com/modin-project/modin/blob/a8be482e644519f2823668210cec5cf1564deb7e/modin/experimental/core/storage_formats/hdk/query_compiler.py
     """
 
-    # When lazy_execution=True, upstream Modin elides some length checks that would incur queries.
-    lazy_execution = True
+    # When these laziness flags are set, upstream Modin elides some length checks that would incur queries.
+    lazy_row_labels = True
+    lazy_row_count = True
+    lazy_column_types = False
+    lazy_column_labels = False
+    lazy_column_count = False
 
     def __init__(self, frame: InternalFrame) -> None:
         """this stores internally a local pandas object (refactor this)"""
--- a/src/snowflake/snowpark/modin/plugin/docstrings/series.py
+++ b/src/snowflake/snowpark/modin/plugin/docstrings/series.py
@@ -406,7 +406,7 @@ class Series(BasePandasDataset):
         pass
 
     __iadd__ = __add__
-    __imul__ = __add__
+    __imul__ = __mul__
     __ipow__ = __pow__
     __isub__ = __sub__
     __itruediv__ = __truediv__
--- a/tests/integ/modin/index/test_monotonic.py
+++ b/tests/integ/modin/index/test_monotonic.py
@@ -12,7 +12,7 @@ from tests.integ.utils.sql_counter import sql_count_checker
 
 
 @pytest.mark.parametrize(
-    "values", [[], [1], [3, 2], [1, 3, 2], [1, 2, 2], [1, np.NaN, 3]]
+    "values", [[], [1], [3, 2], [1, 3, 2], [1, 2, 2], [1, np.nan, 3]]
 )
 @sql_count_checker(query_count=1)
 def test_monotonic_increasing_numbers(values):
@@ -23,7 +23,7 @@ def test_monotonic_increasing_numbers(values):
 
 
 @pytest.mark.parametrize(
-    "values", [[], [3], [1, 2], [3, 1, 2], [2, 2, 1], [3, np.NaN, 1]]
+    "values", [[], [3], [1, 2], [3, 1, 2], [2, 2, 1], [3, np.nan, 1]]
 )
 @sql_count_checker(query_count=1)
 def test_monotonic_decreasing_numbers(values):
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ cd snowpark-python
 
 - Create a new Python virtual environment with any Python version that we support.
   - The Snowpark Python API supports **Python 3.9, Python 3.10, Python 3.11 and Python 3.12**.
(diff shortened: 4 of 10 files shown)


src/snowflake/snowpark/modin/plugin/docstrings/series.py
```diff
@@ -406,7 +406,7 @@ def __rtruediv__():
         pass
 
     __iadd__ = __add__
-    __imul__ = __add__
```
sfc-gh-jjiao, line 409: was this mistake been there all the time? Good catch

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -530,8 +530,12 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
...
     """
 
-    # When lazy_execution=True, upstream Modin elides some length checks that would incur queries.
-    lazy_execution = True
+    # When these laziness flags are set, upstream Modin elides some length checks that would incur queries.
+    lazy_row_labels = True
+    lazy_row_count = True
+    lazy_column_types = False
+    lazy_column_labels = False
+    lazy_column_count = False
```
sfc-gh-jjiao, line 538: are those flags only effective since 0.32.0?

sfc-gh-jjiao:
I saw there are a few changes (update) besides pinning the modin version to 0.32.0, such as deleting the function `def case_when(self, caselist) -> Series: # noqa: PR01, RT01, D200`,   I am curious how we know we need to do those change ? is that by manual exhaustive searching or by seeing test failures and fix test by test?

sfc-gh-jjiao (approved):

src/snowflake/snowpark/modin/plugin/docstrings/series.py
sfc-gh-joshi: yes, this was a mistake upstream that we copied into the doc class by mistake

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: yes

sfc-gh-joshi:
> I saw there are a few changes (update) besides pinning the modin version to 0.32.0, such as deleting the function def case_when(self, caselist) -> Series: # noqa: PR01, RT01, D200,

We implemented `case_when` (a relatively new pandas function) before upstream modin did, and modin 0.32.0 has the correct implementation so it's safe to delete the override. No functionality should be changed as a result of this.

> I am curious how we know we need to do those change ? is that by manual exhaustive searching or by seeing test failures and fix test by test?

For most overrides, I left a comment explaining the reason (or asked the implementor to leave one) the override exists, as well as a modin version where the fix is available if it was upstreamed. `case_when` in particular was fixed in `0.31.0`, so I did a grep for "0.32.0" and "0.31.0" to see if there were any upstreamed changes we no longer needed extensions for.

Some stuff (like `np.NaN` being deprecated in newer versions of numpy that are supported by modin 0.32.0) was found by manual testing. For upstreamed bugfixes that aren't yet in a modin release, I have a spreadsheet I periodically update.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Why are some set to True and others set to False? How are they set in upstream Modin?

tests/integ/modin/index/test_monotonic.py
```diff
@@ -12,7 +12,7 @@
 
 
 @pytest.mark.parametrize(
-    "values", [[], [1], [3, 2], [1, 3, 2], [1, 2, 2], [1, np.NaN, 3]]
```
sfc-gh-helmeleegy, line 15: I'm assuming we still need to keep `test_old_numpy_aliases.py::test_old_nan`, which specifically checks that `np.NaN` still works, correct?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: The ray and dask backends have extra logic to cache index/columns objects after their first materialization, so the `PandasQueryCompiler` class by default determines these properties based on whether those have been materialized already (for example, `lazy_row_count` is False when the index has already been materialized, and True if it has not yet been).

These are used by frontend methods upstream to guard size and index checks, like here in `__getitem__`:
https://github.com/modin-project/modin/blob/735735ed98f81f5ad92a1bc93df5a4287379141a/modin/pandas/base.py#L3994

The previous `lazy_execution` flag was too coarse for cases like this. We could be using these more aggressively to remove unnecessary extensions in the Snowpark pandas codebase that exist only to remove query counts from length/label validation, but for now I'm leaving them in for simplicity.

tests/integ/modin/index/test_monotonic.py
sfc-gh-joshi: The `np.NaN` alias was outright removed in numpy 2.0; modin 0.30.1 had a ceiling on the nump<2 that is now removed. `test_old_nan` is only run by a job in the daily test suite that explicitly installs numpy 1.26 and uses the old `np.NaN` alias.

sfc-gh-helmeleegy: Do we know when we'll stop supporting those older versions of numpy?

sfc-gh-joshi: I'm not sure, it would depend on the minimum version that pandas allows. numpy 1.26.4 (the last version before 2.0) was released 2/5/24, and 1.22.4 (the minimum version supported by pandas 2.2.3) was released 5/20/22.

sfc-gh-joshi:
Daily test failures in https://github.com/snowflakedb/snowpark-python/actions/runs/14479320997 also occur on `main`; filed SNOW-2046103 to investigate later. Kicking off a run of the Snowpark pandas API daily jobs to make sure everything is fine.

sfc-gh-jrose (approved):

PR #3269, opened 2025-04-16
SNOW-1997136: Ensure transitive closure of AST Requests

sfc-gh-heshah:

   [SNOW-1997136](https://snowflakecomputing.atlassian.net/browse/SNOW-1997136)

   - [  ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.

   This PR implements post-processing of statements in `AstBatch` to ensure the transitive closure of all statements and their dependencies are included in every `Request` message sent to the server.

   Most of the logic is ported from `tests/ast/ast_test_utils::create_full_ast_request`, but modified to reflect several updates that have been made to the AST since this method's implementation.


sfc-gh-heshah:
* **#3269** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -79,14 +107,15 @@ def bind(self, symbol: Optional[str] = None) -> proto.Bind:
...
         """
-        stmt = self._request.body.add()
-        # TODO: extended BindingId spec from the branch snowpark-ir.
-        stmt.bind.first_request_id = (
-            self._request_id.bytes
-        )  # Set the first batch ID this stmt will be flushed with.
-        stmt.bind.uid = self._get_next_id()
-        stmt.bind.symbol.value = symbol if isinstance(symbol, str) else ""
-        return stmt.bind
+        with self._lock:
```
sfc-gh-lspiegelberg, line 110: put the lock only on 
```
stmt.bind.first_request_id = (
                self._request_id.bytes
            )  # Set the first batch ID this stmt will be flushed with.
stmt.bind.uid = next(AstBatch.__id_gen)
self._cur_stmts.append(stmt)
```
This is what only needs to be guarded.

A more elegant solution may be to simply use a helper function with a lock, e.g.
```
return self._append_stmt(stmt)
```
and then have
```
def _append_stmt(self, stmt: proto.Stmt) -> proto.Bind:
    with self._lock:
         stmt.bind.first_request_id = self._request_id.bytes
         stmt.bind.uid = next(AstBatch.__id_gen)
         self._cut_stmt.append(stmt)
         return stmt.bind
```

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -95,55 +124,142 @@ def eval(self, target: proto.Bind) -> None:
         Args:
             target: The variable to evaluate.
         """
-        stmt = self._request.body.add()
-        stmt.eval.bind_id = target.uid
+        with self._lock:
+            stmt = proto.Stmt()
```
sfc-gh-lspiegelberg, line 128: with the helper function, you can rese code here potentially.

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -95,55 +124,142 @@ def eval(self, target: proto.Bind) -> None:
...
+
+        def get_dependent_bind_ids(ast: Any) -> set[int]:
+            """Retrieve the dependent AST IDs required for this AST object."""
+            dependent_ids = set()
+            if isinstance(ast, Iterable) and not isinstance(ast, str):
+                for c in ast:
+                    dependent_ids.update(get_dependent_bind_ids(c))
+            elif hasattr(ast, "DESCRIPTOR"):
+                descriptor = ast.DESCRIPTOR
+                if descriptor.name in (
```
sfc-gh-lspiegelberg, line 147: Define a constant: DESCRIPTOR_NAMES={...} (set for faster lookup)

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -95,55 +124,142 @@ def eval(self, target: proto.Bind) -> None:
...
+        for stmt in self._cur_stmts:
+            if stmt.eval.bind_id != 0:
+                eval_stmt_bind_ids[
+                    stmt.eval.bind_id
+                ] = stmt  # stores the eval statement
+            else:
+                self._prev_bind_stmts[stmt.bind.uid] = stmt  # stores the bind statement
+                cur_request_bind_ids.append(stmt.bind.uid)
+
+        # Keep track of all the bind IDs (including dependencies) that we need to send over.
```
sfc-gh-lspiegelberg, line 179: this function is very long, can you split up into parts? similar to ASTBuilder design?

tests/unit/ast/test_batch.py
```diff
@@ -16,16 +16,16 @@ def fake_session():
```
sfc-gh-lspiegelberg, line 17: can we add some tests for the transitive closure? Just in terms of what to expect? That'd be great.

sfc-gh-heshah: Yes, I think @sfc-gh-oplaton's recently added smoke tests are a great candidate for testing the transitive closure work.

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-heshah: I think we should just maintain our own `RLock` instance. I'm weary of issues even if thread safe sessions aren't enabled due to the possibility that multiple users acquire the same `Session` instance through whatever means, and additionally reusing the `Session` instance's own lock seems like we're inviting future trouble as more and more attributes make use of it.

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -21,6 +24,37 @@
...
+    "RelationalGroupedDataframeRef",
+    "FnIdRefExpr",
+    "FnRef",
+    "IndirectTableFnIdRef",
+}
+
+
+def get_dependent_bind_ids(ast: Any) -> set[int]:
+    """Retrieve the dependent AST IDs required for this AST object."""
+    dependent_ids = set()
```
sfc-gh-lspiegelberg, line 39: how is potentially an infinite loop avoided here? I.e., when recursively calling get_dependent_ids(...), maybe have some visited check?

sfc-gh-lspiegelberg: will this case never happen because ids are unique?

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-lspiegelberg: ok, sgtm.

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -62,88 +100,195 @@ def __init__(
...
+            # Keep track of bind IDs we have already visited.
+            visited_bind_ids = set()
+
+            # Depth first search (FIFO) of all statements in the current request.
+            queue_bind_ids = list(self._cur_request_bind_ids)
+            while queue_bind_ids:
+                # queue_bind_ids is a copy of self._cur_request_bind_ids, which is maintained as a min heap.
+                # We pop the minimum bind ID from the queue with heappop to process the bind statements in the order they were generated.
+                bind_id = heapq.heappop(queue_bind_ids)
+                visited_bind_ids.add(bind_id)
```
sfc-gh-lspiegelberg, line 195: If I am not mistkane, the heapq.<cmd> APIs make `queue_bind_ids` a queue and not a stack? then I think we're doing BFS here.

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -62,88 +100,195 @@ def __init__(
...
+
+                # Add the new dependent bind IDs to the queue if we have not visited them yet.
+                new_dependent_bind_ids = dependent_bind_ids.difference(visited_bind_ids)
+                # Maintain the heap invariant by pushing new bind IDs to the queue one by one.
+                for id in new_dependent_bind_ids:
+                    heapq.heappush(queue_bind_ids, id)
+
+            return full_request_bind_ids
+
+    def create_full_ast_request(
```
sfc-gh-lspiegelberg, line 220: nit: `to_request`?

tests/ast/data/DataFrame.to_pandas_batch.test
```diff
@@ -16,10 +16,16 @@ df = session.table("table1")
 
 df.to_pandas_batches()
```
sfc-gh-lspiegelberg, line 18: these are an artifact of the transitive closure/unparser, right?

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-heshah: It can happen actually. Any dataframe APIs that accept another dataframe as a parameter would cause an infinite loop if the same dataframe is used.

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -62,88 +100,195 @@ def __init__(
...
+        """
+        with self._lock:
+            # Keep track of all the bind IDs (including dependencies) that we need to send over.
+            # This list should be maintained as min heap to ensure that we unparse bind statements in the order they were generated.
+            full_request_bind_ids: list[int] = []
+
+            # Keep track of bind IDs we have already visited.
+            visited_bind_ids = set()
+
+            # Depth first search (FIFO) of all statements in the current request.
```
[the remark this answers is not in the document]
sfc-gh-heshah: Ah this comment is actually outdated. Its neither now, as we're maintaining `queue_bind_ids` as a priority queue by bind ID. Maybe "best first search" is more apt name

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-heshah: Technically `queue_bind_ids` (and `self._cur_request_bind_ids`) are maintained as list representations of the min-heap binary tree, so I don't think the node expansions follow any consistent order as they do in BFS or DFS.

tests/ast/data/DataFrame.to_pandas_batch.test
sfc-gh-heshah: Yes, the Unparser isn't updated to only print statements whose `first_request_id` fields match the current `Request`'s ID field.

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-heshah: Sorry I misinterpreted your comment, we can never loop infinitely here because we are traversing the generated AST only (its populated fields), which is guaranteed to be finite and strictly acyclic.

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -62,88 +100,195 @@ def __init__(
...
         self._init_batch()
 
         # Track callables in this dict (memory id -> TrackedCallable).
         self._callables: dict[int, TrackedCallable] = {}
 
-    def reset_id_gen(self) -> None:
-        """Resets the ID generator."""
-        self._id_gen = itertools.count(start=1)
+        # Track previously flushed statements by their UIDs.
+        self._prev_bind_stmts: dict[int, proto.Stmt] = {}
```
sfc-gh-oplaton, line 111: Does this dictionary ever decrease in size?

src/snowflake/snowpark/_internal/ast/batch.py
```diff
@@ -62,88 +100,195 @@ def __init__(
...
+
+        # Cache the dependencies of each statement.
+        self._dependency_cache: dict[int, set[int]] = {}
```
sfc-gh-oplaton, line 114: Same question here, does this dictionary ever get flushed, under any circumstances?

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-heshah: It should never decrease in size during the lifetime of an instance of `AstBatch`.

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-lspiegelberg: We could explicitly clear when a session is closed via session.close() in case. This helps avoid issues with cleanup if there were still active references to the object.

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-lspiegelberg: oh, it's much simpler. If you use a data structure that has a FIFO property (first in, first out), you're doing BFS. If you have a data structure that has a LIFO property (last-in, first-out), you're doing DFS.

typical data structures are:
queue -> FIFO
stack -> LIFO

sfc-gh-lspiegelberg (approved):
lgtm, great job @sfc-gh-heshah !

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-heshah: Since an `AstBatch` instance will only ever be held by a `Session` I don't think we have a need for explicitly clearing. Once the `Session`'s reference to the `AstBatch` is dropped, these should just be cleaned up as well as there should be no additional active references. Did we still want to add an explicit clear method to be called in `Session.close()`?

sfc-gh-oplaton (approved):

src/snowflake/snowpark/_internal/ast/batch.py
sfc-gh-lspiegelberg: Might be a good idea. Wouldn't overly rely on Python's automatic reference counting too much.

PR #3282, opened 2025-04-17
NOSNOW: Enable installation from source with one line

sfc-gh-heshah:

   NOSNOW: Fix issues and allow installation from source via `pip install -e "[.development]"

   - Include generated `ast_pb2.py(i)` files so customers don't have to run `python -m tox -e protoc`
   - Include `protoc-wheel-0==21.1`, `mypy-protobuf` in dev dependencies


sfc-gh-heshah:
* **#3282** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3282?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowpark-python/3282?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sfc-gh-jdu (approved):

.gitignore
```diff
@@ -26,7 +26,6 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
-**/generated/
```
sfc-gh-oplaton, line 29: This doesn't seem right. We really don't want generated sources to accidentally make it into feature branches or `main` and go out of sync with the AST definition.

What is the goal of this change?

sfc-gh-lspiegelberg:
Why do we need to add the protobuf/mypy dependency? According to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#writing-your-pyproject-toml, the place should be pyproject.toml - shouldn't this suffice?

.gitignore
sfc-gh-heshah: The ask is for users to be able to install from source directly via `pip install git: ...` without having to run any additional commands.

Currently they would have to run `python -m tox -e protoc` to avoid running into the "Cannot import" errors due to the lack of the `ast_pb2.py` file in the GitHub repository. The protobuf compiler script would only work in bash environments as well, causing the extra command to fail when customers are using Windows.

sfc-gh-lspiegelberg: left a few comments below. Custom command seems like a sensible approach here https://github.com/pypa/setuptools/discussions/3762

sfc-gh-heshah:
My main concern with a custom command in the setup system is running into issues similar to those which the customer who raised this faced (`python -m tox -e protoc` failed on their environment). We are taking an additional dependency in our build tools (on customer environments) on the `protoc` library to run the custom command successfully.

FWIW, we are already running the risk of the generated files going out of sync with the true AST definitions on the server-side that generate `ast.proto`. IMO it is safe to add a merge gate that checks that the generated files are consistent with 
the definitions in `ast.proto` if we are okay with checking in these generated files.

I do think it is necessary to add the protobuf compiler/mypy packages as dependencies in the developer installation of the repository however, because for AST changes on the server-side we ask developers to run `pip install tox protoc-wheel-0==21.1 mypy-protobuf` in our AST/IR Howto so that they can locally generate `ast_pb2.py(i)` using an updated `ast.proto` from the server-side.

PR #3284, opened 2025-04-18
SNOW-2049277: Fix XML reader when an XML record has nested self-closing tag

sfc-gh-jdu:

   

   Fixes SNOW-2049277

   For such an XML record:
   ```
   <a>
      <b .../>
   </a>
   ```
   The original implementation will fail because we look for the self-closing tag right away. This PR fixes this bug by looking for a ">" directly after the opening tag.

### DIFF
--- a/src/snowflake/snowpark/_internal/xml_reader.py
+++ b/src/snowflake/snowpark/_internal/xml_reader.py
@@ -57,9 +57,11 @@ def tag_is_self_closing(
     chunk_size: int = DEFAULT_CHUNK_SIZE,
 ) -> Tuple[bool, int]:
     """
-    Return ``(is_self_closing, end_pos)`` for the opening tag at `start_pos`.
+    Return ``(is_self_closing, end_pos)`` after searching the terminating ``>`` .
     ``end_pos`` is the byte offset one past the terminating ``>`` of that
     same tag.
+    This method is quote-aware and will not consider a ``>`` inside a quote, e.g.,
+    ``<book title="a<b>c">``.
     """
     in_quote = False
     last_byte = b""
--- a/src/snowflake/snowpark/_internal/xml_reader.py
+++ b/src/snowflake/snowpark/_internal/xml_reader.py
@@ -64,6 +64,7 @@ def tag_is_self_closing(
     ``<book title="a<b>c">``.
     """
     in_quote = False
+    quote_char = None
     last_byte = b""
 
     while True:
@@ -74,8 +75,12 @@ def tag_is_self_closing(
 
         for idx, b in enumerate(struct.unpack(f"{len(chunk)}c", chunk)):
             # '>' inside quote should not be considered as the end of the tag
-            if b == b'"':
-                in_quote = not in_quote
+            if not in_quote and b in [b'"', b"'"]:
+                in_quote = True
+                quote_char = b
+            elif in_quote and b == quote_char:
+                in_quote = False
+                quote_char = None
             # '>' outside quotes
             elif b == b">" and not in_quote:
                 is_self = last_byte == b"/"
@@ -331,7 +336,7 @@ def process_xml_range(
 
             # decide whether the row element is self‑closing
             try:
-                is_self_close, tag_end = tag_is_self_closing(f, record_start)
+                is_self_close, tag_end = tag_is_self_closing(f)
             except EOFError:
                 # malformed XML record
                 break
--- a/src/snowflake/snowpark/_internal/xml_reader.py
+++ b/src/snowflake/snowpark/_internal/xml_reader.py
@@ -6,12 +6,12 @@ import os
 import re
 import html.entities
 import logging
+import struct
 import xml.etree.ElementTree as ET
-from typing import Optional, Dict, Any, Iterator, BinaryIO, Union
+from typing import Optional, Dict, Any, Iterator, BinaryIO, Union, Tuple
 from snowflake.snowpark.files import SnowflakeFile
 
 
-SELF_CLOSING_TAG: bytes = b"/>"
 DEFAULT_CHUNK_SIZE: int = 1024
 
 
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -52,6 +52,36 @@ def get_file_size(filename: str) -> Optional[int]:
         return file_obj.tell()
 
 
+def tag_is_self_closing(
+    file_obj: Union[BinaryIO, SnowflakeFile],
+    chunk_size: int = DEFAULT_CHUNK_SIZE,
+) -> Tuple[bool, int]:
+    """
+    Return ``(is_self_closing, end_pos)`` for the opening tag at `start_pos`.
+    ``end_pos`` is the byte offset one past the terminating ``>`` of that
```
sfc-gh-aalam, line 61: from the description it seems that the function should also take `start_pos` as an input arg but that is not the case. Does the docstring need to be updated?

sfc-gh-jdu: good catch, fixed

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -52,6 +52,38 @@ def get_file_size(filename: str) -> Optional[int]:
...
+        chunk_start_pos = file_obj.tell()
+        chunk = file_obj.read(chunk_size)
+        if not chunk:
+            raise EOFError("EOF reached before end of opening tag")
+
+        for idx, b in enumerate(struct.unpack(f"{len(chunk)}c", chunk)):
+            # '>' inside quote should not be considered as the end of the tag
+            if b == b'"':
+                in_quote = not in_quote
+            # '>' outside quotes
```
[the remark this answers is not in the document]
sfc-gh-jdu: This is a good point. Actually there is no back‑slash escaping (\") inside XML attribute values. If we want to embed a double‑quote inside a double‑quoted attribute, we must use the entity ``&quot;``. But I forgot single quote case, so fixed it here.

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -52,6 +52,45 @@ def get_file_size(filename: str) -> Optional[int]:
         return file_obj.tell()
 
 
+def tag_is_self_closing(
+    file_obj: Union[BinaryIO, SnowflakeFile],
```
sfc-gh-aalam, line 56: If I understand correctly, the file pointer may not be at the beginning when this function is called, right? But this function has a side-effect of incrementing the pointer. Would it be better if we go back to the older approach with has start file position as input? Seems like that approach will be easier to have a no side-effect implementation

sfc-gh-jdu: Actually we expect to increment the pointer here, similar to `find_next_closing_tag_pos`. Because later we will start here to find the next closing tag:
```
  try:
                is_self_close, tag_end = tag_is_self_closing(f)
            except EOFError:
                # malformed XML record
                break

            if is_self_close:
                record_end = tag_end
            else:
                f.seek(tag_end)
                try:
                    record_end = find_next_closing_tag_pos(f, closing_tag, chunk_size)
                except EOFError:
                    # incomplete XML record
                    break
```

sfc-gh-jdu: I can add a comment for this function that will increment the pointer.

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -87,33 +125,13 @@ def find_next_closing_tag_pos(
...
-        elif self_close_pos != -1:
-            chosen = self_close_pos
-            tag_len = len(SELF_CLOSING_TAG)
-        elif close_pos != -1:
-            chosen = close_pos
-            tag_len = len(closing_tag)
-
-        if chosen is not None:
-            absolute_pos = file_obj.tell() - len(data) + chosen + tag_len
+        idx = data.find(closing_tag)
```
sfc-gh-jkew, line 132: Hypothesis: the reason why `xml.etree.ElementTree.iterparse` will not work in this case is that we cannot use a traditional SAX XML parser for this problem because the search for matching tags needs to result in a byte range rather than an element range.

I just want to take a quick step back and confirm that hypothesis, because to an outside observer these types of modifications might indicate a problem in the approach. I think the hypothesis is true, but maybe we should mention that in comments.

sfc-gh-jkew (changes requested):
Add one comment to the top of `find_next_opening_tag_pos` which explains why a traditional streaming XML parser will not work ( ie. you need to return byte ranges for chunking rather than handle elements incrementally in a single process.).

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-jdu: done. Thanks for bringing it up!

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -301,6 +319,11 @@ def process_xml_range(
     tag_start_2 = f"<{tag_name} ".encode()
     closing_tag = f"</{tag_name}>".encode()
 
+    # We perform raw byte‑level scanning here because we must split the file into independent
```
sfc-gh-jdu, line 322: I added here in `process_xml_range` before calling `find_next_opening_tag`, which is more clear IMO. @sfc-gh-jkew

sfc-gh-jkew (approved):

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-jdu: done

sfc-gh-aalam (approved):

PR #3285, opened 2025-04-19
SNOW-2049599 Fix missing publicapi decorators and don't send dataframeAST if ast_enabled is set.

sfc-gh-evandenberg:

   

   Fixes SNOW-2049599 Fix missing publicapi decorators and don't send dataframeAST if ast_enabled is set

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -57,6 +57,7 @@ from snowflake.snowpark._internal.utils import (
     create_rlock,
     create_thread_local,
     escape_quotes,
+    is_ast_enabled,
     get_application_name,
     get_version,
     is_in_stored_procedure,
@@ -430,7 +431,7 @@ class ServerConnection:
         self, query: str, **kwargs: Any
     ) -> SnowflakeCursor:
         notify_kwargs = {}
-        if DATAFRAME_AST_PARAMETER in kwargs:
+        if DATAFRAME_AST_PARAMETER in kwargs and is_ast_enabled():
             notify_kwargs["dataframeAst"] = kwargs[DATAFRAME_AST_PARAMETER]
 
         try:
--- a/src/snowflake/snowpark/column.py
+++ b/src/snowflake/snowpark/column.py
@@ -826,6 +826,7 @@ class Column:
 
     # Note: For the operator overrides we always emit ast, it simply gets ignored in a call chain.
 
+    @publicapi
     def __neg__(self) -> "Column":
         """Unary minus."""
 
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -4732,6 +4732,7 @@ class DataFrame:
             + line
         )
 
+    @publicapi
     def _show_string_spark(
         self,
         num_rows: int = 20,
--- a/src/snowflake/snowpark/dataframe_writer.py
+++ b/src/snowflake/snowpark/dataframe_writer.py
@@ -611,6 +611,7 @@ class DataFrameWriter:
             **copy_options,
         )
 
+
     def _internal_copy_into_location(
         self,
         caller: str,
--- a/src/snowflake/snowpark/dataframe_writer.py
+++ b/src/snowflake/snowpark/dataframe_writer.py
@@ -611,7 +611,6 @@ class DataFrameWriter:
             **copy_options,
         )
 
-
     def _internal_copy_into_location(
         self,
         caller: str,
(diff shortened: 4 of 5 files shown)


src/snowflake/snowpark/column.py
```diff
@@ -826,6 +826,7 @@ def bitxor(
 
     # Note: For the operator overrides we always emit ast, it simply gets ignored in a call chain.
 
+    @publicapi
```
sfc-gh-heshah, line 829: Dunder/magic methods won't accept additional arguments afaik, so I don't think we should decorate them. Is this working in our tests?

src/snowflake/snowpark/dataframe.py
```diff
@@ -4732,6 +4732,7 @@ def row_to_string(row: List[str]) -> str:
             + line
         )
 
+    @publicapi
```
sfc-gh-heshah, line 4735: I understand why this has been added here, I did so previously myself in an earlier PR before removing it. The semantics of decorating an internal method with `@publicapi` is incorrect imho. I do think we need to protect internal methods with our decorator, but we should discuss renaming it if there is a strong case for using the decorator on many internal methods.

sfc-gh-heshah (changes requested):

src/snowflake/snowpark/dataframe.py
sfc-gh-evandenberg: Yes, I agree the naming is bad.  Perhaps we should have an @internalapi which alias onto @publicapi.  Would that be better?  Either way, this is a bug and we can also address the naming separately.

src/snowflake/snowpark/column.py
sfc-gh-lspiegelberg: +1, no need to decorate dunder methods.

src/snowflake/snowpark/dataframe.py
sfc-gh-lspiegelberg: this should be a public API. It came up many, many times. A good name would be `as_spark_string` or so. Let's do a proper job here, no issue in making this a public API. There's only upside. Let's document it and make it maintainable.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -430,7 +431,7 @@ def execute_and_notify_query_listener(
         self, query: str, **kwargs: Any
     ) -> SnowflakeCursor:
         notify_kwargs = {}
-        if DATAFRAME_AST_PARAMETER in kwargs:
+        if DATAFRAME_AST_PARAMETER in kwargs and is_ast_enabled():
```
sfc-gh-lspiegelberg, line 434: why do we need `is_ast_enabled()` here? It feels a bit odd to have a check here. Much simpler to pass a parameter or not. If the parameter is not there, it won't be used. Callers are responsible to provide meaningful input to internal APIs.

sfc-gh-evandenberg: cc: @sfc-gh-azhan 

@sfc-gh-lspiegelberg and I had a discussion on what to do in the case when dataframe (fragment or otherwise) AST is provided and ast_enabled=false.  Based on the current design we don't expect this situation to occur except as a result of a bug in our code likely around the top level APIs improperly defaulting _emit_ast parameter.  There are couple options we discussed based on whether we treat precondition violation as a hard failure or not.

1. Add an assertion (i.e., assert not (DATAFRAME_AST_PARAMETER in kwargs) or is_ast_enabled()) that raises a hard failure if the precondition is violated. The benefit here is an immediate failure occurs which propagates back to the original pyspark client (actually @sfc-gh-azhan can you confirm if this is the case as it would need to get propagated back through several layers).  Presumably customers would need to notify or we would monitor for this failure.  While in PrPr it would be better to have hard failures so we don't miss and can fix the underlying issue, avoid the risk of this becoming a silent failure.

2. Add a check that only requests the dataframe AST if ast_enabled=true.  This effectively forces the server connection to execute as sql and avoids treating as a hard failure.  The benefit is that we don't fail the upstream callers (pyspark client) and the workload can make progress uninterrupted.  We would need to augment this with a way to report the precondition violation to we can know to fix the bug.  @sfc-gh-azhan Do you know if there's a good way to do this, i.e. snowflake non-fatal incident log from python code?  Also, do we know how to monitor and create bugs from these reports.

sfc-gh-evandenberg: Yes, the @publicapi changes can be cleaned up (avoid using for internal methods).

I think it would be less disruptive to ensure we don't unnecessarily make into a hard failure when there's a very safe alternative action here (execute sql query.). Anyone else would be good to chime in?  @sfc-gh-oplaton @sfc-gh-azwiegincew

sfc-gh-evandenberg:
Follow up in offline discussion thread.

PR #3306, opened 2025-04-25
SNOW-2012567: Skip describe for SYSTEM$ procedures

sfc-gh-aalam:

   

   Fixes SNOW-2012567

  For `session.call` snowpark client runs a `describe procedure` query on the given procedure to determine if the return type is scalar or a table. However, running a describe on `SYSTEM$` procedures shows up as a failed query in query history. In this PR, we skip submitting a describe for `SYSTEM$` procedures since all system procedure's return type is scalar.

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -4072,7 +4072,7 @@ class Session:
     def _infer_is_return_table(
         self, sproc_name: str, *args: Any, log_on_exception: bool = False
     ) -> bool:
-        if sproc_name.upper().startswith("SYSTEM$"):
+        if sproc_name.strip().upper().startswith("SYSTEM$"):
             # Built-in stored procedures do not have schema and cannot be described
             # Currently all SYSTEM$ stored procedures are scalar so return false.
             return False
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -4072,7 +4072,7 @@ class Session:
     def _infer_is_return_table(
         self, sproc_name: str, *args: Any, log_on_exception: bool = False
     ) -> bool:
-        if "SYSTEM$" in sproc_name.upper():
+        if sproc_name.upper().startswith("SYSTEM$"):
             # Built-in stored procedures do not have schema and cannot be described
             # Currently all SYSTEM$ stored procedures are scalar so return false.
             return False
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -4072,6 +4072,10 @@ def stored_procedure_profiler(self) -> StoredProcedureProfiler:
     def _infer_is_return_table(
         self, sproc_name: str, *args: Any, log_on_exception: bool = False
     ) -> bool:
+        if "SYSTEM$" in sproc_name.upper():
```
sfc-gh-jrose, line 4075: nit:
```suggestion
        if sproc_name.upper().startswith("SYSTEM$"):
```

sfc-gh-jrose (approved):

src/snowflake/snowpark/session.py
```diff
@@ -4072,6 +4072,10 @@ def stored_procedure_profiler(self) -> StoredProcedureProfiler:
     def _infer_is_return_table(
         self, sproc_name: str, *args: Any, log_on_exception: bool = False
     ) -> bool:
+        if sproc_name.upper().startswith("SYSTEM$"):
+            # Built-in stored procedures do not have schema and cannot be described
```
sfc-gh-aalam, line 4076:
```suggestion
        if sproc_name.strip().upper().startswith("SYSTEM$"):
            # Built-in stored procedures do not have schema and cannot be described
```

src/snowflake/snowpark/session.py
```diff
@@ -4072,6 +4072,10 @@ def stored_procedure_profiler(self) -> StoredProcedureProfiler:
     def _infer_is_return_table(
         self, sproc_name: str, *args: Any, log_on_exception: bool = False
     ) -> bool:
+        if sproc_name.strip().upper().startswith("SYSTEM$"):
+            # Built-in stored procedures do not have schema and cannot be described
+            # Currently all SYSTEM$ stored procedures are scalar so return false.
+            return False
```
sfc-gh-aling, line 4078: is it always illegal to call desc on a system proc?
my major concern is whether this is a breaking change

sfc-gh-aalam: Calling `describe procedure SYSTEM$...` will fail but the function `_infer_is_return_table` has a try-finally protection, so the return value is always for in this case False. However, the failure does show up in query history and is a bad user experience (see JIRA).

Here what we are doing is skipping the describe for system$ procedures so there is no change in behavior and query history stays clean.

sfc-gh-aling (approved):

PR #3316, opened 2025-04-29
SNOW-807303 Use SELECT * EXCLUDE for DataFrame.drop()

sfc-gh-aalam:

   

   Fixes SNOW-807303

   This PR adds the ability to generate sql using `SELECT * EXCLUDE (exclude_cols...)` with `DataFrame.drop` API call.


src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -171,7 +171,7 @@ def infer_metadata(
                 ]
             # When source_plan doesn't have a projection, it's a simple `SELECT * from ...`,
             # which means source_plan has the same metadata as its child plan, we can use it directly
-            if source_plan.projection is None:
+            if not source_plan.has_projection:
```
sfc-gh-jrose, line 174: Do you have an example scenario when source_plan.has_projection is Falsey, but not None?

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-yuwang: yes, I have the same question `not source_plan.has_projection` meaning that if `source_plan.has_projection` is None or for example a empty list [], the statement would be true, is that desired behavior?

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
```diff
@@ -231,7 +231,7 @@ def cache_metadata_if_selectable(
             # we should cache it on the child plan too.
             # This is necessary SelectStatement.select() will need the column states of the child plan
             # (check the implementation of derive_column_states_from_subquery().
-            if source_plan.projection is None:
+            if not source_plan.has_projection:
```
sfc-gh-yuwang, line 234: yes, I have the same question not source_plan.has_projection meaning that if source_plan.has_projection is None or for example a empty list [], the statement would be true, is that desired behavior?

src/snowflake/snowpark/dataframe.py
```diff
@@ -1782,48 +1781,83 @@ def drop(
...
+                    )
+
+            normalized_names = {quote_name(n) for n in names}
+            existing_names = [attr.name for attr in self._output]
+            keep_col_names = [c for c in existing_names if c not in normalized_names]
+            if not keep_col_names:
+                raise SnowparkClientExceptionMessages.DF_CANNOT_DROP_ALL_COLUMNS()
+
+            if self._select_statement and self._session.conf.get(
+                "use_simplified_query_generation"
```
sfc-gh-jdu, line 1826: is there other functionality that will use this flag?

src/snowflake/snowpark/dataframe.py
```diff
@@ -1782,48 +1781,83 @@ def drop(
...
+                    c._expression, NamedExpression
+                ):
+                    names.append(c._expression.name)
+                else:
+                    raise SnowparkClientExceptionMessages.DF_CANNOT_DROP_COLUMN_NAME(
+                        str(c)
+                    )
+
+            normalized_names = {quote_name(n) for n in names}
+            existing_names = [attr.name for attr in self._output]
```
sfc-gh-jdu, line 1820: is it possible to not check existing names if using exclude? It will trigger a describe query

sfc-gh-aalam: as discussed offline, if we don't do this describe query, it may lead to a BCR since `SELECT * EXCLUDE (non_existing_column)` raises error SQL error while it raises no error today.

src/snowflake/snowpark/dataframe.py
sfc-gh-aalam: yeah. all new query generation improvement for dataframe APIs are protected with this parameter. .distinct(), .random_split(), .stat.sample_by(), .drop() and planning to add .rename() under this.

src/snowflake/snowpark/_internal/analyzer/metadata_utils.py
sfc-gh-aalam: `has_projection` is a new property I added for `SelectStatement` which is a check to know if we have added any projection on top of the child Selectable. Before this change, there was only one way to add projection which was with select method that updates `projection` attribute. Now, both existence of `projection` and `exclude_cols` mean we have added projection on top of child selectable.

https://github.com/snowflakedb/snowpark-python/pull/3316/files#diff-ba9463440d935e1d5667a2909f7d2b8061f56de7bdabd8207a29903e9eae8f59R859-R864

I can update this to ensure that `has_projection` is True when
- `projection is None`
- `projection == []`
- `exclude_cols is None`

sfc-gh-jdu (approved):

sfc-gh-yuwang (approved):

PR #3339, opened 2025-05-07
SNOW-2084165 Add dataframe operation lineage on SnowparkSQLException

sfc-gh-aalam:

   

   Fixes SNOW-2084165

   see [doc](https://docs.google.com/document/d/1sMeepp3hVXFAiBofz_GwCreim0FMcozERUmqqpkBcCw/edit?tab=t.0#heading=h.i394h1jd79wy).


tests/integ/scala/test_column_suite.py
```diff
@@ -508,10 +505,10 @@ def test_column_constructors_select(session):
 
     with pytest.raises(SnowparkSQLException) as ex_info:
         df.select('"Col"').collect()
-    assert "invalid identifier" in str(ex_info)
+    assert "invalid identifier" in str(ex_info.value)
```
sfc-gh-aalam, line 508: example str(ex_info) output
```
<ExceptionInfo SnowparkSQLException('snowflake.snowpark.exceptions.SnowparkSQLException: (1304): 01bc4ee1-0a10-0849-000c-a909c7a27113...s/snowpark-python/tests/integ/test_function.py|787:9-787:52:  session.create_dataframe([4], schema=["a"])', None, None) tblen=2>
```

example str(ex_info.value) output
```
snowflake.snowpark.exceptions.SnowparkSQLException: (1304): 01bc4ee1-0a10-0849-000c-a909c7a27113: 100044 (22P01): Invalid floating point operation: sqrt(-1)


--- Additional Debug Information ---


Trace of the dataframe operations that could have caused the error (total 2):

/Users/aalam/Projects/snowpark-python/tests/integ/test_function.py|793:8-793:32:  df.select(sqrt(lit(-1))).collec
/Users/aalam/Projects/snowpark-python/tests/integ/test_function.py|787:9-787:52:  session.create_dataframe([4], schema=["a"])
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -659,6 +801,7 @@ def stat(self) -> DataFrameStatFunctions:
     def analytics(self) -> DataFrameAnalyticsFunctions:
         return self._analytics
 
+    @dataframe_exception_handler
```
sfc-gh-aling, line 804: From code maintainability perspective, do you think it would make sense to integrate the lineage functionality into the existing `@publicapi`, or would it be better to keep them separate?

sfc-gh-aalam: I thought about this but it is only applicate to dataframe API and not all public apis. @sfc-gh-lspiegelberg do you think it makes sense to add this in @publicapi?

sfc-gh-aalam: we could also do something like this

```
from snowflake.snowpark._internal.utils import publicapi as common_publicapi


def publicapi(func):
    @dataframe_exception_handler
    @common_publicapi
    def combined_wrapper(*args, **kwargs):
        return func(*args, **kwargs)
    return combined_wrapper
```

This will reduce the amount of overall change but it also increases the congitive load on being aware that the publicapi in dataframe has an attached `dataframe_exception_handler` attached. I do not think is the best approach here.

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,101 @@
...
+                if end_line == 0 and i == start_line:
+                    code_lines.append(line)
+                elif end_line != 0 and start_line <= i <= end_line:
+                    code_lines.append(line)
+
+            if end_line != 0:
+                # Should we make this inference based on python version?
+                # If the end line is not 0, we need to trim the start and end columns
+                code_lines[0] = code_lines[0][start_column - 1 :]
+                code_lines[-1] = code_lines[-1][:end_column]
```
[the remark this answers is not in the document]
sfc-gh-aalam: `start_column` is actually 0-indexed. I discovered this when I had this test case

```
data = [
    ("a", "metric1", 10),
    ("a", "metric2", 20),
    ("b", "metric1", 30),
    ("b", "metric2", 40)
]
df = session.create_dataframe(data, schema=["id", "metric", "value"])

METRIC1 = "metric1"
METRIC2 = "metric2"
# Pivot the table — this will generate quoted column names like "METRIC1", "METRIC2"
pivoted_df = df.pivot("metric", [METRIC1, METRIC2]).sum("value")

# Trigger an error by trying to select a column with unquoted lowercase name
pivoted_df.select(METRIC1).show()
```
the debug logs produced by this are

```
--- Additional Debug Information ---


Trace of the most recent dataframe operations associated with the error (total 4):

/Users/aalam/Projects/testing/triage.py|44:0-44:26: pivoted_df.select(METRIC1)
/Users/aalam/Projects/testing/triage.py|40:13-40:64: df.pivot("metric", [METRIC1, METRIC2]).sum("value")
/Users/aalam/Projects/testing/triage.py|40:13-40:51: df.pivot("metric", [METRIC1, METRIC2])
/Users/aalam/Projects/testing/triage.py|34:5-34:69: session.create_dataframe(data, schema=["id", "metric", "value"])
```

src/snowflake/snowpark/dataframe.py
sfc-gh-jdu: We already handle some errors here https://github.com/snowflakedb/snowpark-python/blob/dc8ac301bc66568041b08a04505bd200629c7fcf/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L148. Is it possible to combine?

src/snowflake/snowpark/dataframe.py
```diff
@@ -343,6 +349,143 @@ def _disambiguate(
...
+                    continue
+                visited_batch_id.add(child_id)
+                next.append(DataFrameLineageNode(child_id, node.stmt_cache))
+
+        curr = next
+
+    return lineage
+
+
+def dataframe_exception_handler(func):
```
sfc-gh-jdu, line 403: should it also be applied to Table class, which is a subclass of dataframe?

src/snowflake/snowpark/dataframe.py
```diff
@@ -343,6 +349,143 @@ def _disambiguate(
     return lhs_remapped, rhs_remapped
```
sfc-gh-jdu, line 351: Should we add docstring to introduce this improvement? I think it will be very helpful if users can find it in documentation.

sfc-gh-aalam: Yeah. I'll add that. thanks for calling it out

src/snowflake/snowpark/dataframe.py
sfc-gh-aalam: this is a better place. It would also capture other classes as well.

src/snowflake/snowpark/dataframe.py
```diff
@@ -343,6 +349,143 @@ def _disambiguate(
...
+                    if isinstance(arg, DataFrame):
+                        dataframes_involved.append(arg)
+
+                df_lineage = _get_df_lineage(dataframes_involved)
+                lineage_trace_len = len(df_lineage)
+                show_lineage_len = int(
+                    os.environ.get(
+                        "SNOWPARK_PYTHON_DATAFRAME_LINEAGE_LENGTH_ON_ERROR", 5
+                    )
+                )
```
sfc-gh-aling, line 449: in the doc you mentioned this is turned off by default, but from the code I think it's turned on by default?

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        self, filename, start_line, end_line, start_column, end_column
+    ) -> str:
+        """Read the relevant code snippets of where the DataFrame was created. The filename given here
+        must have read permissions for the executing user."""
+        with open(filename) as f:
+            code_lines = []
+            for i, line in enumerate(f, 1):
+                if end_line == 0:
+                    if i == start_line:
+                        code_lines.append(line)
```
sfc-gh-aling, line 46: I just want to clarify that if end_line is 0 (I'm assuming 0 means first line), then start_line and i must also be 0 right?

sfc-gh-aalam: I should improve the comments here. end_line = 0 in python 3.9/3.10 environments. In those cases, we only know that start line. Therefore we have a distinction based on end_line. I think I should redo the if statement using 
```
if sys.version >= (3, 11):
    # condition 1
else:
    # condition 2
```

src/snowflake/snowpark/dataframe.py
sfc-gh-aalam: I don't understand what you mean by "turned off by default". Do you mean the process of us adding debug logs or the environment variable to get the df transformation trace length?

The additional debug logs is turned off by default because ast_id is None when AST is turned off. And when ast_id for dataframe is None, we cannot collect any trace. Which is why this feature is turned off by default. Once AST is enabled, we also enable the additional debug logs.

The environment variable can be used to control the trace length. The default trace length is 5.

tests/integ/test_df_lineage.py
```diff
@@ -0,0 +1,252 @@
...
+from tests.resources.test_df_debug_dir.dataframe_generator1 import DataFrameGenerator1
+from tests.resources.test_df_debug_dir.dataframe_generator2 import DataFrameGenerator2
+
+from snowflake.snowpark._internal.utils import set_ast_state, AstFlagSource
+
+pytestmark = [
+    pytest.mark.skipif(
+        "config.getoption('local_testing_mode', default=False)",
+        reason="CTE is a SQL feature",
+        run=False,
```
sfc-gh-aalam, line 22:
```suggestion
        reason="SnowparkSQLException is not raised in localtesting mode",
        run=False,
```

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,203 @@
...
+                    if start_line <= i <= end_line:
+                        code_lines.append(line)
+                    elif i > end_line:
+                        break
+                else:
+                    # For python 3.9/3.10, we do not extract the end line from the source code
+                    # so we just read the start line and return.
+                    if i == start_line:
+                        code_lines.append(line)
+                        break
```
sfc-gh-jrose, line 60: You could use an iterator slice to skip checking most lines:
https://docs.python.org/3/library/itertools.html#itertools.islice

sfc-gh-jrose (commented):
It might be worthwhile to have a flag that disables this feature. Some customers may not want their source code showing up in error logs.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -157,9 +164,28 @@ def wrap(*args, **kwargs):
...
+                    df_transform_debug_trace = None
+                    try:
+                        if df_ast_id is not None and stmt_cache is not None:
+                            df_transform_debug_trace = get_df_transform_trace_message(
+                                df_ast_id, stmt_cache
+                            )
+                    except Exception:
+                        # If we encounter an error when getting the df_transform_debug_trace,
+                        # we will ignore the error and not add the debug trace to the error message.
+                        pass
```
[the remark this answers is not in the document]
sfc-gh-aalam: I'm not sure to add a new telemetry for this but I could add a debug line here

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -157,9 +164,28 @@ def wrap(*args, **kwargs):
...
                     tb = sys.exc_info()[2]
                     assert e.msg is not None
+
+                    # extract df_ast_id, stmt_cache from args
+                    df_ast_id, stmt_cache = None, None
+                    for arg in args:
+                        if isinstance(arg, SnowflakePlan):
+                            df_ast_id = arg.df_ast_id
+                            stmt_cache = arg.session._ast_batch._bind_stmt_cache
+                            break
```
sfc-gh-aling, line 174: is it possible that there are multiple plans in the args? and if so, is the left most plan guaranteed to be the most recent plan?

sfc-gh-jdu: we have some code below:
```
children = [
                            arg for arg in args if isinstance(arg, SnowflakePlan)
                        ]
```                        
which means there will be multiple plans ?

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,203 @@
...
+    def __init__(self, batch_id: int, stmt_cache: Dict[int, proto.Stmt]) -> None:
+        self.batch_id = batch_id
+        self.stmt_cache = stmt_cache
+
+    @cached_property
+    def children(self) -> set[int]:
+        """Returns the batch_ids of the children of this node."""
+        return get_dependent_bind_ids(self.stmt_cache[self.batch_id])
+
+    def get_src(self):
```
sfc-gh-aling, line 32: nit: type hint?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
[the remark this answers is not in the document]
sfc-gh-jdu: yea let's add a log line here

src/snowflake/snowpark/_internal/debug_utils.py
sfc-gh-jdu, line 32: return type hint

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,203 @@
...
+        stmt = self.stmt_cache[self.batch_id]
+        api_call = stmt.bind.expr.WhichOneof("variant")
+        return (
+            getattr(stmt.bind.expr, api_call).src
+            if api_call and getattr(stmt.bind.expr, api_call).HasField("src")
+            else None
+        )
+
+    def _read_file(
+        self, filename, start_line, end_line, start_column, end_column
```
sfc-gh-jdu, line 43: type hint

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,203 @@
...
+                if start_line == end_line:
+                    code_lines[0] = code_lines[0][start_column:end_column]
+                else:
+                    code_lines[0] = code_lines[0][start_column:]
+                    code_lines[-1] = code_lines[-1][:end_column]
+
+            code_lines = [line.rstrip() for line in code_lines]
+            return "\n".join(code_lines)
+
+    def get_source_id(self) -> str:
```
sfc-gh-aling, line 72: I feel this can also be cached property, or could get_src return dynamic result?

tests/integ/test_df_debug_trace.py
sfc-gh-aling, line 1: this is adding a new integ test file, do you want to add it to the snowfort test?

sfc-gh-aling:
I would be echoing Jamison's comment on introducing a separate parameter to control this, to decouple from ast enable.
what if a customer enables ast but wants to opt out this feature

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: It is possible for args to have multiple `SnowflakePlans` but only the first arg is relevant to the failure. This is not obvious and we have to see the usage of this decorator to be sure of this.

The decorator is used for the following functions
1. `SnowflakePlan._analyze_attributes` [link](https://github.com/snowflakedb/snowpark-python/blob/ede1f244ba0951f74d23a8edc17f6b17644d519c/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L491)
2. `Selectable._analyze_attributes` [link](https://github.com/snowflakedb/snowpark-python/blob/ede1f244ba0951f74d23a8edc17f6b17644d519c/src/snowflake/snowpark/_internal/analyzer/select_statement.py#L565)
3. `SnowflakePlanBuilder.build` [link](https://github.com/snowflakedb/snowpark-python/blob/ede1f244ba0951f74d23a8edc17f6b17644d519c/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L690)
4. `SnowflakePlanBuilder.build_binary` [link](https://github.com/snowflakedb/snowpark-python/blob/ede1f244ba0951f74d23a8edc17f6b17644d519c/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L730)
5. `ServerConnection.get_result_set` [link](https://github.com/snowflakedb/snowpark-python/blob/ede1f244ba0951f74d23a8edc17f6b17644d519c/src/snowflake/snowpark/_internal/server_connection.py#L650)

Out of these only `build` and `build_binary` can have multiple snowflake plans. For each of them, the source of failure can only possible come from describe queries. But we have already wrapped `SnowflakePlan._analyze_attributes` and `Selectable._analyze_attributes` and they can have at most one SnowflakePlan. Outside of describe queries, the wrapper will be triggered to wrap programming exception on getting full result for example for `.show()` or `.collect()`. For these cases, we have wrapped `get_result_set` which only has one `SnowflakePlan` arg. IMO, the wrapper of `build` and `build_binary` are redundant. @sfc-gh-jdu did you notice any other case where parsing plans in these functions is actually giving you more information?

src/snowflake/snowpark/_internal/ast/utils.py
```diff
@@ -742,8 +742,7 @@ def with_src_position(
         # Once we've stepped out of the snowpark package, we should be in the code of interest.
         # However, the code of interest may execute in an environment that is not accessible via the filesystem.
         # e.g. Jupyter notebooks, REPLs, calls to exec, etc.
-        filename = frame.f_code.co_filename if frame is not None else ""
-        if frame is None or not Path(filename).is_file():
+        if frame is None:
```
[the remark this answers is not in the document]
sfc-gh-aalam: This is for reading line number in REPL environments. In absence of this, we don't get any information in REPL/notebooks like environments. I do have an improvement JIRA logging in where reading files for source code might not be necessary.

src/snowflake/snowpark/_internal/debug_utils.py
sfc-gh-aalam: It can be cached_property as well. let me update that

sfc-gh-aalam:
> I would be echoing Jamison's comment on introducing a separate parameter to control this, to decouple from ast enable. what if a customer enables ast but wants to opt out this feature

Added. they can do 
```
import snowflake.snowpark.context as context
context.disable_extract_debug_trace()
```

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(self, batch_id: int, stmt_cache: Dict[int, proto.Stmt]) -> None:
+        self.batch_id = batch_id
+        self.stmt_cache = stmt_cache
+
+    @cached_property
+    def children(self) -> set[int]:
+        """Returns the batch_ids of the children of this node."""
+        return get_dependent_bind_ids(self.stmt_cache[self.batch_id])
+
+    def get_src(self) -> Optional[proto.SrcPosition]:
```
sfc-gh-jkew, line 33: In the hybrid client prototype we are using a slightly different method to get the source location; by just using inspect to walk the stack to the appropriate source location. We have to do this because modin is not using any of the AST stuff, but it's also relatively straight forward.

I sort of want to use your debugging tool for snowpandas as well; but we may want to refactor this so we don't require any of the protobuf work.

https://github.com/snowflakedb/snowpark-python/blob/685fc313f27b43dfd23d608ffaf57129e374381b/src/snowflake/snowpark/modin/plugin/_internal/telemetry.py#L625

src/snowflake/snowpark/context.py
```diff
@@ -29,6 +29,17 @@
 # This is an internal-only global flag, used to determine whether the api code which will be executed is compatible with snowflake.snowpark_connect
 _is_snowpark_connect_compatible_mode = False
 
+# Following are internal-only global flags, used to enable development features.
+_enable_dataframe_trace_on_error = False
+
+
+def configure_development_features(
```
sfc-gh-aling, line 36: see my comment on: https://github.com/snowflakedb/snowpark-python/pull/3380#discussion_r2114583745

I'm thinking of the following to provide unified debug config experience.

@experimental(version="1.33.0")
def debug_config(
    *
    enable_eager_schema_validation=False,
    enable_dataframe_trace_on_error=False,
)
```


when users want to enable, they do

```python
import snowflake.snowpark.context
snowflake.snowpark.context.debug_config(enable_eager_schema_validation=True)
# or
snowflake.snowpark.context.debug_config(
    enable_eager_schema_validation=True,
    enable_dataframe_trace_on_error=True
)
```

sfc-gh-aalam: good idea. @sfc-gh-jrose and I are aligned on the name. Let me add the `@experimental` decorator as well.

src/snowflake/snowpark/context.py
```diff
@@ -29,6 +29,17 @@
...
+    *,
+    enable_dataframe_trace_on_error: bool = False,
```
sfc-gh-jrose, line 38: I think we should default to `True`. That way users that want a basic development mode can call this function without any parameters.

src/snowflake/snowpark/context.py
sfc-gh-aalam: could not import decorator due to circular import issues but added a warning there

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-aalam: https://github.com/snowflakedb/snowpark-python/pull/3437

src/snowflake/snowpark/_internal/debug_utils.py
sfc-gh-jkew: What about using this function? https://github.com/snowflakedb/snowpark-python/blob/f415c733c86971cc410491820fcf9aea88e12dec/src/snowflake/snowpark/_internal/open_telemetry.py#L116 

Essentially we seem to have three approaches to this problem. I'm /less/ of a fan of the AST because it doesn't help pandas for this type of debugging. but it seems like we might be able to consolidate w/ the open telemetry approach.

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,200 @@
...
+SNOWPARK_PYTHON_DATAFRAME_TRANSFORM_TRACE_LENGTH = (
+    "SNOWPARK_PYTHON_DATAFRAME_TRANSFORM_TRACE_LENGTH"
+)
+
+
+class DataFrameTraceNode:
+    """A node representing a dataframe operation in the DAG that represents the lineage of a DataFrame."""
+
+    def __init__(self, batch_id: int, stmt_cache: Dict[int, proto.Stmt]) -> None:
+        self.batch_id = batch_id
```
sfc-gh-heshah, line 25: Nit: I would argue that this isn't meant to be batch ID anymore. Within each Python session that imports the Snowpark module, each AST ID for a Table or Dataframe will be a UID.

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -0,0 +1,200 @@
...
+        self.stmt_cache = stmt_cache
+
+    @cached_property
+    def children(self) -> set[int]:
+        """Returns the batch_ids of the children of this node."""
+        return get_dependent_bind_ids(self.stmt_cache[self.batch_id])
```
sfc-gh-heshah, line 31: This helper method will return all dependencies of a given statement (anything that requires an AST ID including UDxFs and Sprocs), not just the direct dependencies. I see later you're expecting `node.children` to just be the direct dependencies to get the lineage.

sfc-gh-heshah (approved):

PR #3367, opened 2025-05-14
SNOW-2097941: Support reading XML files with namespace

sfc-gh-jdu:

   

   Fixes SNOW-2097941

   Use lxml library to enable namespace support

### DIFF
--- a/src/snowflake/snowpark/_internal/xml_reader.py
+++ b/src/snowflake/snowpark/_internal/xml_reader.py
@@ -6,9 +6,17 @@ import os
 import re
 import html.entities
 import struct
-import xml.etree.ElementTree as ET
 from typing import Optional, Dict, Any, Iterator, BinaryIO, Union, Tuple
 from snowflake.snowpark.files import SnowflakeFile
+
+try:
+    import lxml.etree as ET
+
+    lxml_installed = True
+except ImportError:
+    import xml.etree.ElementTree as ET
+
+    lxml_installed = False
 
 
 DEFAULT_CHUNK_SIZE: int = 1024
@@ -224,7 +232,7 @@ def find_next_opening_tag_pos(
         # Update the overlap from the end of the combined data.
         overlap = data[-overlap_size:] if len(data) >= overlap_size else data
 
-        # Otherwise, rewind by the length of the overlap so that a tag spanning the boundary isn’t missed.
+        # Otherwise, rewind by the length of the overlap so that a tag spanning the boundary isn't missed.
         file_obj.seek(-len(overlap), 1)
 
         # Check that progress is being made to avoid infinite loops.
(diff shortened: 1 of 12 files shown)


src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -399,8 +411,14 @@ def process_xml_range(
...
-                element = ET.fromstring(record_str)
-                yield element_to_dict(strip_namespaces(element))
+                if lxml_installed:
+                    parser = ET.XMLParser(recover=True, ns_clean=True)
+                    element = ET.fromstring(record_str, parser)
+                else:
+                    element = ET.fromstring(record_str)
+                if strip_namespaces:
+                    element = strip_xml_namespaces(element)
+                yield element_to_dict(element)
```
sfc-gh-aling, line 421: I'm a little confused。
would there be any functional difference between when lxml is installed and lxml is not installed?

also is this part of code executed in UDTF? I saw you added 
```
packages=["snowflake-snowpark-python", "lxml"],
```

in that case, wouldn't lxml be guaranteed available?

sfc-gh-jdu: Yeah for the UDTF, lxml will always be there; but this is also a file in snowpark python, and lxml is only added as a dev dependency. Although it's very unlikely that users will import this module locally, if they do, it will fail due to import error. So, just added it for this reason.

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-aling: I see, I feel it's worth adding the reason as comments in the code
otherwise I would think the code would always work regardless of whether lxml_installed
```python
                if lxml_installed:
                    parser = ET.XMLParser(recover=True, ns_clean=True)
                    element = ET.fromstring(record_str, parser)
                else:
                    element = ET.fromstring(record_str)
```

sfc-gh-aling (approved):

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1144,7 +1153,7 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
                 handler_name,
                 output_schema=output_schema,
                 input_types=input_types,
-                packages=["snowflake-snowpark-python"],
+                packages=["snowflake-snowpark-python", "lxml"],
```
sfc-gh-aling, line 1156: do we need a cap here? like 4 < lxml < 6?
or you have tested that all versions are compatibile?

sfc-gh-jdu: good point, done

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-jdu: done, added comments

PR #3379, opened 2025-05-20
SNOW-2060749: add databricks udtf support

sfc-gh-aling:

   

   Fixes SNOW-2060749

Add databricks UDTF ingestion support


sfc-gh-yuwang (approved):

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1353,6 +1353,7 @@ def create_oracledb_connection():
                 fetch_size=fetch_size,
                 imports=udtf_configs.get("imports", None),
                 packages=udtf_configs.get("packages", None),
```
sfc-gh-aalam, line 1355: How does ast work for this feature?

src/snowflake/snowpark/_internal/data_source/drivers/base_driver.py
```diff
@@ -166,6 +163,18 @@ def data_source_data_to_pandas_df(
 
     @staticmethod
     def to_result_snowpark_df(
-        session: "Session", table_name, schema, _emit_ast: bool = True
+        session: "Session", table_name: str, schema: StructType, _emit_ast: bool = True
```
sfc-gh-aalam, line 166: we have `_emit_ast = True` but we do not have `@publicapi`

src/snowflake/snowpark/_internal/data_source/drivers/base_driver.py
```diff
@@ -166,6 +163,18 @@ def data_source_data_to_pandas_df(
...
     ) -> "DataFrame":
         return session.table(table_name, _emit_ast=_emit_ast)
+
+    @staticmethod
+    def to_result_snowpark_df_udtf(
+        res_df: "DataFrame",
+        schema: StructType,
+        _emit_ast: bool = True,
```
sfc-gh-aalam, line 174: same

src/snowflake/snowpark/_internal/data_source/drivers/databricks_driver.py
```diff
@@ -63,29 +63,55 @@ def to_snow_type(self, schema: List[Any]) -> StructType:
...
+    def udtf_class_builder(self, fetch_size: int = 1000) -> type:
+        create_connection = self.create_connection
+
+        class UDTFIngestion:
+            def process(self, query: str):
+                conn = create_connection()
+                cursor = conn.cursor()
+
+                # First get schema information
+                describe_query = f"DESCRIBE QUERY SELECT * FROM ({query})"
```
sfc-gh-aalam, line 81: Is the query here going to be generated so that processing can be partitioned into different UDTF worker nodes?

sfc-gh-aling: correct, the query here is an already-partitioned one
we generate a table with partitioned queries first and let each worker process the query

but for schema inference within a single worker, whether partitioned or not does not make a difference

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aling: dumping our discussion here:

dbapi is public api, which makes use of the ast and we emit ast right before returning the result dataframe back to users.

there are two ingestion modes and we emit ast differently:

1. local parquet file based approach
in this approach, we only emit ast for last operation which returns session.table() to the user, all internal operations won't emit ast. we expect the operations to take place on client env.

2. udtf
in this approach, we emit ast for udtf registration, udtf call, and resulting dataframe column selection.

src/snowflake/snowpark/_internal/data_source/drivers/base_driver.py
sfc-gh-aling: see my comment above

src/snowflake/snowpark/_internal/data_source/drivers/base_driver.py
sfc-gh-aling: see my comment above

src/snowflake/snowpark/_internal/data_source/drivers/databricks_driver.py
sfc-gh-aalam: sweet

PR #3380, opened 2025-05-20
SNOW-2097586: Add Debug mode for eager evaluation

sfc-gh-jrose:

   

   Fixes SNOW-2097586

   This PR adds a new debug mode to sessions. Currently the only difference in debug mode is that dataframe attributes are populated on Dataframe creation instead of lazily, but in the future other features may be added. This change is done as a separate mode in order to avoid the additional describe queries for production pipelines.

The side effect of this change is that errors related to the dataframe schema become apparent on the line of code that caused them rather than on whatever line evaluates the dataframe. I've also modified the exception wrapper for column identifiers to not modify the error message if no additional context is available.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -236,43 +236,44 @@ class SnowflakePlan(LogicalPlan):
                                         quoted_identifiers.extend(
                                             node.quoted_identifiers
                                         )
+                            if quoted_identifiers:
 
-                            def add_single_quote(string: str) -> str:
-                                return f"'{string}'"
+                                def add_single_quote(string: str) -> str:
+                                    return f"'{string}'"
 
-                            # We can't display all column identifiers in the error message
-                            if len(quoted_identifiers) > 10:
-                                quoted_identifiers_str = f"[{', '.join(add_single_quote(q) for q in quoted_identifiers[:10])}, ...]"
-                            else:
-                                quoted_identifiers_str = f"[{', '.join(add_single_quote(q) for q in quoted_identifiers)}]"
+                                # We can't display all column identifiers in the error message
+                                if len(quoted_identifiers) > 10:
+                                    quoted_identifiers_str = f"[{', '.join(add_single_quote(q) for q in quoted_identifiers[:10])}, ...]"
+                                else:
+                                    quoted_identifiers_str = f"[{', '.join(add_single_quote(q) for q in quoted_identifiers)}]"
 
-                            msg = (
-                                f"There are existing quoted column identifiers: {quoted_identifiers_str}. "
-                                f"Please use one of them to reference the column. See more details on Snowflake identifier requirements "
-                                f"https://docs.snowflake.com/en/sql-reference/identifiers-syntax"
-                            )
-
-                            # Currently, when Snowpark user a Python string as identifier to access a column:
-                            # 1) if a column name is unquoted and
-                            #   a) contains no special characters, it is automatically uppercased and quoted in SQL, or
-                            #   b) if it includes special characters, it is simply quoted without uppercasing.
-                            # 2) If the name is explicitly quoted by the user, Snowpark preserves it as-is.
-                            # Therefore, if `col` is an invalid identifier, it is most likely due to 1a) above.
-                            # We attempt to provide a more helpful error message by suggesting the closest valid identifier.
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -27,6 +27,7 @@
 )
 
 import snowflake.snowpark
+import snowflake.snowpark.context
```
sfc-gh-aalam, line 30: context is imported but not used?

src/snowflake/snowpark/session.py
```diff
@@ -1127,6 +1128,15 @@ def custom_package_usage_config(self, config: Dict) -> None:
                 k.lower(): v for k, v in config.items()
             }
 
+    @experimental(version="1.32.0")
+    def enable_debug_mode(self, enabled=True):
+        """
```
sfc-gh-aalam, line 1133: we sort of have two modes right now for debug improvements

1. AST based which populates debug logs.
2. Eager schema query evaluation

Should we introduce two separate modes and reserve `session.enable_debug_mode` to enable both?

sfc-gh-aalam (approved):

src/snowflake/snowpark/context.py
```diff
@@ -37,6 +42,18 @@ def _should_use_structured_type_semantics():
         return _use_structured_type_semantics
 
 
+@experimental(version="1.32.0")
```
sfc-gh-aling, line 45: nit: 1.33.0

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -177,6 +177,11 @@ def wrap(*args, **kwargs):
                         children = [
                             arg for arg in args if isinstance(arg, SnowflakePlan)
                         ]
+
+                        if not children:
+                            # No context available to enhance error message
+                            raise e
```
sfc-gh-aling, line 183: is this behavior equivalent to the previous error handling, I mean same error type/message.
error is also part of our public API

src/snowflake/snowpark/dataframe.py
```diff
@@ -643,6 +644,9 @@ def __init__(
 
         self._alias: Optional[str] = None
 
+        if context._debug_eager_schema_validation:
+            self._plan.attributes
```
sfc-gh-aling, line 648: nit: I think we can leave a comment on calling `attribute` eagerly evaluates the plan.

src/snowflake/snowpark/context.py
```diff
@@ -37,6 +42,18 @@ def _should_use_structured_type_semantics():
...
+def enable_debug_mode(enable_eager_schema_validation=True):
```
sfc-gh-aling, line 46: how would this work with Afroz's debug error message?
will Afroz's introduce another parameter for this function?

sfc-gh-jrose: Yes

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-jrose: It is not. After some discussion I'm going to modify this part and move it into a separate PR here:
https://github.com/snowflakedb/snowpark-python/pull/3382

sfc-gh-aling: is it just error message update? or the error class would be changed.
I think error class change would be a BCR

sfc-gh-jrose: The error class the user would see would have remained the same because it would have been rewrapped further up in the stack. The solution in the new PR doesn't have to rewrap so it will be the same type, but a shorter message. The additional message was not giving useful context and may have caused confusion.

src/snowflake/snowpark/context.py
sfc-gh-yixie: I feel a keyword argument is better like `enable_debug_mode(*, enable_eager_schema_validation=True)`. We may add other debug options. Another thing is the wording options like `development_mode` or `debug_mode`.

sfc-gh-aling: just for API design, I'm wondering if it's better to call it debug_config/development_config?

I got the idea from python built-in logging module: https://docs.python.org/3/library/logging.html#logging.basicConfig

sfc-gh-daviwang (approved):

PR #3406, opened 2025-05-29
SNOW-2128261: Allow PATTERN when INFER_SCHEMA is enabled

sfc-gh-jrose:

   

   Fixes SNOW-2128261

   This PR fixes an issue that DataFrameReader can run into when inferring schema. Before this change the `_infer_schema_for_file_format` would generate a query that used all files in the stage location in order to infer the schema. This is brittle because a stage location could contain files of different types or structures.

We already provide support for specifying `PATTERN` when loading files with a predefined schema. This change ensures that the same `PATTERN` is used to filter out files that should not affect the inferred schema. It does this by running a `list @stage_location pattern='PATTERN'` query and using the list of files returned to infer the schema. 

While writing the tests for this change I found that the `convert_value_to_sql_option` helper function was not correctly handling tuples of size 1. The SQL generated for tuples of size 1 would have an incorrect trailing comma. This hasn't been an issue in the past because every example I can find has two or more items in the tuple, but the FILES parameter can have just one.

### DIFF
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -3,7 +3,6 @@
 #
 import multiprocessing as mp
 import os
-import pathlib
 import sys
 import time
 
@@ -59,13 +58,14 @@ from snowflake.snowpark._internal.utils import (
     TempObjectType,
     get_aliased_option_name,
     get_copy_into_table_options,
-    parse_positional_args_to_list_variadic,
-    publicapi,
+    get_stage_parts,
     get_temp_type_for_object,
+    is_in_stored_procedure,
+    parse_positional_args_to_list_variadic,
     private_preview,
+    publicapi,
     random_name_for_temp_object,
     warning,
-    is_in_stored_procedure,
 )
 from snowflake.snowpark.column import METADATA_COLUMN_TYPES, Column, _to_col_if_str
 from snowflake.snowpark.dataframe import DataFrame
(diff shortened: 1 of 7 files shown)


tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -1660,6 +1663,60 @@ def test_pattern(session, mode):
     )
 
 
+# @pytest.mark.parametrize("mode", ["select", "copy"])
+@pytest.mark.parametrize("mode", ["select"])
```
sfc-gh-aalam, line 1667: nit

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -967,10 +969,25 @@ def _infer_schema_for_file_format(
...
+        if (
+            pattern := self._cur_options.get("PATTERN", None)
+        ) and "FILES" not in infer_schema_options:
+            matches = self._session._conn.run_query(
+                f"list {path} pattern = '{pattern}'"
+            )["data"]
+            if len(matches):
+                infer_schema_options["FILES"] = tuple(
+                    m[0].split("/")[-1] for m in matches
+                )
```
sfc-gh-aalam, line 985: can we add a comment on what the return value is and what the tuples will look like

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -970,10 +973,35 @@ def _infer_schema_for_file_format(
         drop_tmp_file_format_if_exists_query: Optional[str] = None
         use_temp_file_format = "FORMAT_NAME" not in self._cur_options
         file_format_name = self._cur_options.get("FORMAT_NAME", temp_file_format_name)
-        infer_schema_options = self._cur_options.get("INFER_SCHEMA_OPTIONS", None)
+        infer_schema_options = self._cur_options.get("INFER_SCHEMA_OPTIONS", {})
+
+        # Client side has no context for if the path is a directory or a file
+        # It should be safe to assume it is a file if there is a parent directory which means that it's
+        # not the top level stage and there's a suffix which means it has a file extension.
+        parsed_path = pathlib.PurePosixPath(path)
```
sfc-gh-aling, line 981: does this class work on windows?

sfc-gh-jrose: It should for the purpose I'm using it for. I think it's unneeded though. I'll switch back to Path.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -970,10 +973,35 @@ def _infer_schema_for_file_format(
...
         use_temp_file_format = "FORMAT_NAME" not in self._cur_options
         file_format_name = self._cur_options.get("FORMAT_NAME", temp_file_format_name)
-        infer_schema_options = self._cur_options.get("INFER_SCHEMA_OPTIONS", None)
+        infer_schema_options = self._cur_options.get("INFER_SCHEMA_OPTIONS", {})
+
+        # Client side has no context for if the path is a directory or a file
+        # It should be safe to assume it is a file if there is a parent directory which means that it's
+        # not the top level stage and there's a suffix which means it has a file extension.
+        parsed_path = pathlib.Path(path)
+        path_is_file = bool(parsed_path.parent and parsed_path.suffix)
```
sfc-gh-aling: I feel there could be files that don't have a suffix.

what would happen if I have such a path `/abc/somefile`, "somefile" is just a normal csv file but without a suffix.
are we able to handle this case?

sfc-gh-jrose: I think the file/directory distinction was a mistake. Stages don't actually have directories, just prefixes. This is confusing because we can specify both the prefix and the pattern which are both just filters on the total list of files in the stage. I'll refactor this.

PR #3413, opened 2025-05-30
SNOW-2104775: Improve error messages for unsupported modin or pandas versions when used in Notebooks

sfc-gh-helmeleegy:

   

   Fixes SNOW-2104775

   Improve error messages for unsupported modin or pandas versions when used in Notebooks.


tests/unit/modin/test_python_version.py
```diff
@@ -23,3 +25,77 @@ def test_error_on_import():
         # Should not error
         import snowflake.snowpark  # noqa: F401
         import snowflake.snowpark.modin.plugin  # noqa: F401
+
+
+def test_pandas_version_error_basic():
+    with patch.dict("sys.modules", {"pandas": MagicMock()}):
+        try:
```
sfc-gh-mvashishtha, line 32: instead of try/except can we use

```python
with pytest.raises(expected_exception=RuntimeError, match=re.escape("Run `pip install")):
```

This is easier to read, and it makes sure that we do raise an exception.

ditto for other test cases.

tests/unit/modin/test_python_version.py
```diff
@@ -23,3 +25,77 @@ def test_error_on_import():
...
+            # Without the following two lines, this error is raised:
+            # TypeError: metaclass conflict: the metaclass of a derived class must be
+            # a (non-strict) subclass of the metaclasses of all its bases
+            pandas.Series = abc.ABCMeta
+            pandas.DataFrame = abc.ABCMeta
+
+            import snowflake.snowpark  # noqa: F401
+            import snowflake.snowpark.modin.plugin  # noqa: F401
+        except RuntimeError as ex:
+            # We will only get here when we run pytest with --noconftest
```
sfc-gh-mvashishtha, line 47: Why is that? Will we run with `--noconftest` in CI? If not, is there way to run this test case in CI? ditto for below.

tests/unit/modin/test_python_version.py
```diff
@@ -23,3 +25,77 @@ def test_error_on_import():
...
+            pandas.__version__ = "1.0"
+
+            # Without the following two lines, this error is raised:
+            # TypeError: metaclass conflict: the metaclass of a derived class must be
+            # a (non-strict) subclass of the metaclasses of all its bases
+            pandas.Series = abc.ABCMeta
+            pandas.DataFrame = abc.ABCMeta
+
+            import snowflake.snowpark  # noqa: F401
+            import snowflake.snowpark.modin.plugin  # noqa: F401
```
sfc-gh-mvashishtha, line 45: Does this statement cause python to rerun `__init__.py`, even if it's already been imported?

sfc-gh-joshi: I believe it does not, and you need to explicitly call `importlib.reload` to rerun the init file. `__init__.py` was not made to be idempotent though, and rerunning it may cause unexpected behaviors like doubly-overwriting extensions.

`test_python_version.py` was written to be explicitly invoked in a clean environment:

https://github.com/snowflakedb/snowpark-python/blob/aa0453c751e54b4e4f671614a1a1f18880bc67b4/tox.ini#L211-L214

The CI job that runs this was removed a while ago since py38 is deprecated now, but I think you would need to add a separate pytest/tox command for each of these tests in one of the GH actions workflow files to get the intended behavior. I'm not aware of a better way to force a test case to run in a separate Python interpreter every time.

sfc-gh-joshi: Is the error branch being run though? Even with `--noconftest`, the relevant modules should be installed, and after `snowflake.snowpark.modin.plugin` is imported successfully once, its init file shouldn't get run again.

sfc-gh-helmeleegy: The new error messages were tested manually, and they worked as expected with and without `--noconftest`. I believe manual testing was also what was done in this similar [PR](https://github.com/snowflakedb/snowpark-python/pull/1419). cc: @sfc-gh-joshi

sfc-gh-mvashishtha: > I believe manual testing was also what was done in this similar PR.

It doesn't necessarily follow that manual testing is always acceptable for changes to our warnings about modin and pandas versions.

I still think we should add the tox commands as @sfc-gh-joshi suggests and make sure that these tests pass in CI.

If we can't guarantee that we raise the expected errors in the tests, then the tests don't work. If you would like to merge your change without working tests, please don't add the test cases in this PR, so that people reading the tests don't expect that the tests are working as expected in CI.

sfc-gh-joshi: I believe the linked PR where I did manual testing previously ran the tests in a daily job through a separate tox environment, it's just no longer run because we no longer support Python 3.8.

sfc-gh-helmeleegy: @sfc-gh-mvashishtha @sfc-gh-joshi I added the tests to the GH action responsible for our daily modin tests. Please take another look.

src/snowflake/snowpark/modin/plugin/__init__.py
```diff
@@ -25,29 +20,43 @@
...
+install_pandas_msg = (
+    f"Please set the pandas version as {supported_pandas_major_version}.{supported_pandas_minor_version}.x in the Packages menu at the top of your notebook."
+    if "snowbook" in sys.modules  # this indicates the environment is Snowflake Notebook
+    else 'Run `pip install "snowflake-snowpark-python[modin]"` to resolve.'
+)
+
 if (
     actual_pandas_version.major != supported_pandas_major_version
     and actual_pandas_version.minor != supported_pandas_minor_version
 ):
```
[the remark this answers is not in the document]
sfc-gh-joshi: @sfc-gh-helmeleegy This comment looks correct, could you include this change in this PR?

tests/unit/modin/test_python_version.py
sfc-gh-joshi: Thanks. It looks like the test you added is failing: https://github.com/snowflakedb/snowpark-python/actions/runs/15574204185/job/43856050917

Could you take a look at what's going on, and try to reproduce the failure locally with the same tox command?

sfc-gh-helmeleegy: The 4 newly added tests are passing. The one that fails is the pre-existing python 3.8 test.

src/snowflake/snowpark/modin/plugin/__init__.py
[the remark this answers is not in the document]
sfc-gh-helmeleegy: Yeah, this is a bug we had. I think we can fix this bug separately, wdyt?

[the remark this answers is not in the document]
sfc-gh-joshi: It's related to the error message for incorrect pandas versions, and it's a one-line change so assuming it doesn't break any tests I think it may as well get bundled in this PR so we don't have to do another set of CI runs.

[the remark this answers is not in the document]
sfc-gh-helmeleegy: Done.

tests/unit/modin/test_python_version.py
sfc-gh-helmeleegy: Yes, we will now run it in the daily job.

tests/unit/modin/test_python_version.py
sfc-gh-helmeleegy: try/except was used to support both with and without `--noconftest`

src/snowflake/snowpark/modin/plugin/__init__.py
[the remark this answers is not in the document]
sfc-gh-helmeleegy: This has caused a bunch of tests to start failing with the following error, particularly `test_df_to_snowpark_pandas.py` when run in either windows or ubunto.
```
RuntimeError: The pandas version installed (2.3.0) does not match the supported pandas version in Snowpark pandas (2.2.x). Run `pip install "snowflake-snowpark-python[modin]"` to resolve.
```
Example:
https://github.com/snowflakedb/snowpark-python/actions/runs/15596145126/job/43927073455?pr=3413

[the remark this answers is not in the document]
sfc-gh-helmeleegy: I will revert this change for now, and we can address it along with the `test_df_to_snowpark_pandas.py` test later.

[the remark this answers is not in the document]
sfc-gh-joshi: This sounds like an issue on our part with version constraints rather than the check being incorrect, since pandas recently released version 2.3.0, which modin does not yet support. Addressing it separately makes sense though.

tests/unit/modin/test_python_version.py
sfc-gh-mvashishtha: Thank you for adding the conftest unit test! Since we are still using try/except instead of `pytest.raises`, we can't count on the tests working in the new daily unit test. I ran the entire test file locally with `noconftest` on python 3.9 and none of the new test cases cause the exceptions that we expect them to.

Could we add the test to a separate file with this fixture, and add `TEST_INCORRECT_MODIN_PANDAS_VERSIONS` to the new pytest command in tox?

```python
@pytest.fixture(autouse=True)
def skip_if_not_testing_incorrect_modin_or_pandas_versions():
    import os
    if not os.environ.get("TEST_INCORRECT_MODIN_PANDAS_VERSIONS"):
        pytest.skip("Skipping tests for incorrect modin/pandas versions")
```

Normally, we could use a custom pytest flag for this, but we need to add those in a conftest file.

tests/unit/modin/test_python_version.py
```diff
@@ -23,3 +25,77 @@ def test_error_on_import():
...
+        except ImportError as ex:
+            # We will only get here when we run pytest with --noconftest
+            assert "Run `pip install" in str(ex)
+
+
+def test_modin_version_error_in_notebooks():
+    with patch.dict("sys.modules", {"snowbook": MagicMock()}):
+        with patch.dict("sys.modules", {"modin": MagicMock()}):
```
sfc-gh-mvashishtha: Ditto: you can combine the dicts so you have just one context manager here.

src/snowflake/snowpark/modin/plugin/__init__.py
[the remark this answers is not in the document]
sfc-gh-mvashishtha: @sfc-gh-helmeleegy could you file a Snowpark pandas issue about fixing the version check here?

tests/unit/modin/test_python_version.py
```diff
@@ -23,3 +25,77 @@ def test_error_on_import():
...
+            import snowflake.snowpark  # noqa: F401
+            import snowflake.snowpark.modin.plugin  # noqa: F401
+        except RuntimeError as ex:
+            # We will only get here when we run pytest with --noconftest
+            assert "Run `pip install" in str(ex)
+
+
+def test_pandas_version_error_in_notebooks():
+    with patch.dict("sys.modules", {"snowbook": MagicMock()}):
+        with patch.dict("sys.modules", {"pandas": MagicMock()}):
```
sfc-gh-mvashishtha: you should be able to patch with a single dict.

tests/unit/modin/test_python_version.py
sfc-gh-helmeleegy: That's weird. The tests pass for me with and without `noconftest` on python 3.9. The new tests also passed when running the GH action manually.

sfc-gh-mvashishtha: Are you sure that the tests are all raising the exceptions that we expect them to? When I put an `else` clause in each  try/except blocks and raise an exception inside there, all the tests fail. I suspect that the tests pass when run individually but fail when run as a group, because `snowlfake.snowpark.modin.plugin` doesn't get re-imported.

sfc-gh-helmeleegy: You're actually right. I added the fixture and the env variable.

tests/unit/modin/test_modin_pandas_version.py
```diff
@@ -0,0 +1,88 @@
...
+
+import abc
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def skip_if_not_testing_incorrect_modin_or_pandas_versions():
+    import os
```
sfc-gh-mvashishtha: nit: move import to the top of the file

```suggestion

import os
import pytest


@pytest.fixture(autouse=True)
def skip_if_not_testing_incorrect_modin_or_pandas_versions():
```

tests/unit/modin/test_modin_pandas_version.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if not os.environ.get("TEST_INCORRECT_MODIN_PANDAS_VERSIONS"):
+        pytest.skip("Skipping tests for incorrect modin/pandas versions")
+
+
+def test_pandas_version_error_basic():
+    with patch.dict("sys.modules", {"pandas": MagicMock()}):
+        try:
```
sfc-gh-mvashishtha: can we use 

pytest.raises(RuntimeError, match=re.escape("Run `pip install"))

here and below?

tests/unit/modin/test_python_version.py
sfc-gh-helmeleegy: Tests still did not pass. Given the amount of trouble I've had so far with getting them to pass as one test file, I'm now trying splitting each test in its own file. Let's see if this will go through.

sfc-gh-helmeleegy: Passed!
https://github.com/snowflakedb/snowpark-python/actions/runs/15624549796/job/44016293061

sfc-gh-mvashishtha: Thank you for adding these tests!

sfc-gh-mvashishtha (approved):

PR #3416, opened 2025-06-02
 SNOW-2096130 : Adding sql query line tracking to SnowflakePlan and Selectable

sfc-gh-daviwang:

   Fixes SNOW-2096130 and SNOW-2096139

We add the field query_line_intervals to Query as a way to track which plans are responsible for line in a SQL query. For example for the query:
```
SELECT  *  FROM (
 SELECT 
    "_1" AS "ID", 
    "_2" AS "NAME", 
    "_3" AS "VALUE"
 FROM (
 SELECT $1 AS "_1", $2 AS "_2", $3 AS "_3" FROM  VALUES (1 :: INT, 'A' :: STRING, 100 :: INT), (2 :: INT, 'B' :: STRING, 200 :: INT)
)
)
 PIVOT (
    sum("VALUE") FOR "NAME" IN ( ANY )
)
```
The parent Snowflake plan's query will have the field query_line_intervals with the value:
```
[[QueryLineInterval(start=0, end=0, uuid=e63f7a07-d08b-4d6b-8606-da7511fca60e), QueryLineInterval(start=1, end=7, uuid=baa71276-1127-42dc-8f0e-cee6ec4539d4), QueryLineInterval(start=8, end=11, uuid=e63f7a07-d08b-4d6b-8606-da7511fca60e)]]
```
Note that each parent only includes its immediate children in this field. To find the exact plan that is responsible for a line, we traverse the tree in the function get_plan_from_line_numbers included in src/snowflake/snowpark/_internal/utils.py. A more detailed description is in the design doc: https://docs.google.com/document/d/1xX1O5VJ-uKOMLAInwhV6lB4HkbpHewBZqx57lONsONQ/edit?tab=t.0 

We disable this functionality for modin tests because it requires multiline sql queries. However, if multiline sql queries are enabled, we pass the modin tests: https://github.com/snowflakedb/snowpark-python/actions/runs/15423685680/job/43405371508?pr=3416. One other small change we add is removing unnecessary redundant white space after a left parenthesis or before a right parenthesis in the normalize_sql function. This ends up being a decent chunk of the lines changed.


src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -386,16 +386,21 @@ def flatten_expression(
     )
 
 
-def lateral_statement(lateral_expression: str, child: str) -> str:
+def lateral_statement(
+    lateral_expression: str, child: str, uuid: Optional[str] = None
+) -> str:
```
sfc-gh-aalam: nit: since we are use the uuid of the child to wrap the which sql, naming `child_uuid` might be a better choice

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -425,6 +431,7 @@ def join_table_function_statement(
     left_cols = [f"{LEFT_ALIAS}.{col}" for col in left_cols]
     right_cols = [f"{RIGHT_ALIAS}.{col}" for col in right_cols]
     select_cols = (COMMA + NEW_LINE + TAB).join(left_cols + right_cols)
+    UUID = f"\n-- {uuid}\n" if uuid else EMPTY_STRING
```
sfc-gh-aalam: Let's introduce `UUID_FORMAT = "\n-- {}\n"` at the top and replace this with

```
UUID = UUID_FORMAT.format(uuid) if uuid is not None else EMPTY_STRING
```

for modin tests, we can set `UUID_FORMAT = ""` and their query generation will not have any new lines as well

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -467,11 +476,17 @@ def case_when_expression(branches: List[Tuple[str, str]], else_value: str) -> st
...
+    project: List[str],
+    child: str,
+    is_distinct: bool = False,
+    uuid: Optional[str] = None,
+) -> str:
     if not project:
         columns = STAR
     else:
         columns = NEW_LINE + TAB + (COMMA + NEW_LINE + TAB).join(project)
+    UUID = f"-- {uuid}\n" if uuid else EMPTY_STRING
```
sfc-gh-aalam: this one is different from others since it does not start with `\n`. Is that intentional. Maybe UUID_FORMAT will make it uniform

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -262,6 +275,11 @@ def __init__(
         self._cumulative_node_complexity: Optional[Dict[PlanNodeCategory, int]] = None
         self._encoded_node_id_with_query: Optional[str] = None
         self.df_ast_ids: Optional[List[int]] = None
+        self._uuid = str(uuid.uuid4())
+        self.query_line_intervals: Optional[List[List["QueryLineInterval"]]] = None
+        self.pre_actions_query_line_intervals: Optional[
+            List[List["QueryLineInterval"]]
```
sfc-gh-aalam: as discussed offline, is it possible to embed this into `Query` class? Will make things easier to maintain longterm

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -317,6 +335,18 @@ def sql_in_subquery(self) -> str:
...
         )
 
+    @property
+    def sql_in_subquery_with_uuid(self) -> str:
+        return (
+            f"{analyzer_utils.LEFT_PARENTHESIS}"
+            f"{analyzer_utils.NEW_LINE}"
+            f"-- {self.uuid}{analyzer_utils.NEW_LINE}"
+            f"{self.sql_query}"
+            f"{analyzer_utils.NEW_LINE}-- {self.uuid}"
```
sfc-gh-aalam: shall we use formatter here as well?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -366,11 +408,13 @@ def get_snowflake_plan(self, skip_schema_query) -> SnowflakePlan:
                 df_aliased_col_name_to_real_col_name=self.df_aliased_col_name_to_real_col_name,
                 source_plan=self,
                 referenced_ctes=self.referenced_ctes,
+                query_line_intervals=query_line_intervals,
             )
             # set api_calls to self._snowflake_plan outside of the above constructor
             # because the constructor copy api_calls.
             # We want Selectable and SnowflakePlan to share the same api_calls.
             self._snowflake_plan.api_calls = self.api_calls
+            self._snowflake_plan._uuid = self._uuid
```
sfc-gh-aalam: If this is not necessary, let's have `SelectSnowflakePlan` it own uuid and not be inherited from `SnowflakePlan`. I can't think of a reason now but it does make it easier to differentiate between the two in case we are debugging some issue.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1663,12 +1808,40 @@ def __deepcopy__(self, memodict={}) -> "SetStatement":  # noqa: B006
...
     def sql_query(self) -> str:
         if not self._sql_query:
-            sql = f"({self.set_operands[0].selectable.sql_query})"
-            for i in range(1, len(self.set_operands)):
-                sql = f"{sql}{self.set_operands[i].operator}({self.set_operands[i].selectable.sql_query})"
-            self._sql_query = sql
+            sql_query = (
+                self._generate_sql_with_comments()
+                if self._session._generate_multiline_queries
+                else self._generate_sql_without_comments()
```
sfc-gh-aalam: great that we are param protecting this. Let's make this param False by default when reading from session

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-daviwang: we need to inherit the uuid because the query_line_intervals we inherit from the snowflake plan will still have the old uuids, and its parent will also reference that uuid

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: let's also create a separate param to control this. We can add it to context

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1831,3 +1835,137 @@ def get_sorted_key_for_version(version_str):
...
+
+
+def get_line_numbers(
+    sql_query: str,
+    child_uuids: List[Tuple[str, int]],
+    parent_uuid: str,
+    parent_query_idx: int,
+) -> List["QueryLineInterval"]:
+    """Helper function to get line numbers associated with child uuids in a query"""
+    """We first find the line numbers associated with each child uuid, and associate the remaining lines with the parent uuid"""
```
sfc-gh-aalam: let's have a more detailed description here

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1831,3 +1835,137 @@ def get_sorted_key_for_version(version_str):
...
+    return new_intervals
+
+
+def get_plan_from_line_numbers(
+    plan_node: Union["SnowflakePlan", "Selectable"],
+    line_number: int,
+    query_idx: int = -1,
+) -> Union["SnowflakePlan", "Selectable"]:
+    """Given a parent plan node, and a line number, return the plan node that contains the line number"""
+    from snowflake.snowpark._internal.analyzer.snowflake_plan import SnowflakePlan
```
sfc-gh-aalam: same. also in the description, add the fact that this function may raise ValueErrors

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -286,6 +295,12 @@ def sql_query(self) -> str:
         """Returns the sql query of this Selectable logical plan."""
         pass
 
+    @property
+    @abstractmethod
+    def commented_sql(self) -> str:
+        """Returns the sql query of this Selectable logical plan with uuids."""
+        pass
```
sfc-gh-aalam: can we make this more descriptive

```suggestion
        """
        Returns the sql query of this Selectable logical plan commented with uuids
        of children nodes.
            SELECT COL1, COL2 FROM 
            -- child_uuid
            (
                <child subquery>
            )
            -- child_uuid
        """
        pass
```

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -354,7 +382,12 @@ def get_snowflake_plan(self, skip_schema_query) -> SnowflakePlan:
             )
             self.analyzer.alias_maps_to_use = self.expr_to_alias.copy()
 
-            query = Query(self.sql_query, params=self.query_params)
+            query = Query(
+                self.commented_sql
+                if self._session._generate_multiline_queries
+                else self.sql_query,
```
sfc-gh-aalam: please create a new param to protect this change. See how we do it for df transform trace

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -532,12 +569,21 @@ def __deepcopy__(self, memodict={}) -> "SelectableEntity":  # noqa: B006
 
     @property
     def sql_query(self) -> str:
-        return f"{analyzer_utils.SELECT}{analyzer_utils.STAR}{analyzer_utils.FROM}{self.entity.name}"
+        sql = f"{analyzer_utils.SELECT}{analyzer_utils.STAR}{analyzer_utils.FROM}{self.entity.name}"
+        return sql
```
sfc-gh-aalam: this change is not necessary

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -532,12 +569,21 @@ def __deepcopy__(self, memodict={}) -> "SelectableEntity":  # noqa: B006
...
+    def commented_sql(self) -> str:
+        return self.sql_query
 
     @property
     def sql_in_subquery(self) -> str:
         return self.entity.name
 
+    @property
+    def sql_in_subquery_with_uuid(self) -> str:
+        return f"\n-- {self.uuid}\n{self.entity.name}\n-- {self.uuid}"
```
sfc-gh-aalam: let's use `UUID_FORMAT` to say consistent

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -597,6 +647,9 @@ def __init__(
         is_select = is_sql_select_statement(sql)
         if not is_select and convert_to_select:
             self.pre_actions = [Query(sql, params=params)]
+            self.pre_actions[-1].query_line_intervals = [
+                QueryLineInterval(0, sql.count("\n"), self.uuid)
+            ]
```
sfc-gh-aalam, line 673: let's add a comment on what this change is doing.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -665,6 +724,10 @@ def referenced_ctes(self) -> Dict[WithQueryBlock, int]:
         # auto created CTE tables referenced
         return dict()
 
+    @property
+    def children_plan_nodes(self) -> List[Union["Selectable", SnowflakePlan]]:
+        return []
```
sfc-gh-aalam: let's add a comment saying `SelectSQL` is a leaf node so it cannot have children

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -567,6 +613,10 @@ def attributes(self, value: Optional[List[Attribute]]):
         if self._session.reduce_describe_query_enabled and value is not None:
             self._schema_query = analyzer_utils.schema_value_statement(value)
 
+    @property
+    def children_plan_nodes(self) -> List[Union["Selectable", SnowflakePlan]]:
+        return []
```
sfc-gh-aalam: a comment saying it is a leaf node therefore its children list is empty

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -940,8 +1012,36 @@ def sql_query(self) -> str:
             return self._sql_query
         if not self.has_clause and not self.has_projection:
             self._sql_query = self.from_.sql_query
+            self._commented_sql = (
+                f"-- {self.from_.uuid}\n{self.from_.sql_query}\n-- {self.from_.uuid}"
```
sfc-gh-aalam: use `UUID_FORMAT`

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -940,8 +1012,36 @@ def sql_query(self) -> str:
...
+            )
             return self._sql_query
-        from_clause = self.from_.sql_in_subquery
+        self._commented_sql = self._generate_sql()
+        self._sql_query = remove_comments(self._commented_sql, [self.from_.uuid])
```
sfc-gh-aalam: let's not update `._commented_sql` here

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -940,8 +1012,36 @@ def sql_query(self) -> str:
...
+        if self._commented_sql:
+            return self._commented_sql
+        if not self.has_clause and not self.has_projection:
+            self._sql_query = self.from_.sql_query
+            self._commented_sql = (
+                f"-- {self.from_.uuid}\n{self.from_.sql_query}\n-- {self.from_.uuid}"
+            )
+            return self._commented_sql
+        self._commented_sql = self._generate_sql()
+        self._sql_query = remove_comments(self._commented_sql, [self.from_.uuid])
```
sfc-gh-aalam: similarly, let's not update `._sql_query` here. This function should only be touching `_commented_sql`

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -633,6 +688,10 @@ def __deepcopy__(self, memodict={}) -> "SelectSQL":  # noqa: B006
     def sql_query(self) -> str:
         return self._sql_query
 
+    @property
+    def commented_sql(self) -> str:
+        return self._commented_sql
```
sfc-gh-aalam, line 714: this doesn't really need to be made into a property using @property here. We can just get rid of `_commented_sql` here and replace it with `commented_sql` for SelectSQL

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -708,6 +771,10 @@ def snowflake_plan(self):
     def sql_query(self) -> str:
         return self._snowflake_plan.queries[-1].sql
 
+    @property
+    def commented_sql(self) -> str:
+        return self._snowflake_plan.queries[-1].sql
```
sfc-gh-aalam: In this case will the returned sql be actually commented?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1663,12 +1771,47 @@ def __deepcopy__(self, memodict={}) -> "SetStatement":  # noqa: B006
...
+
         return self._sql_query
 
+    @property
+    def commented_sql(self) -> str:
+        if not self._commented_sql:
+            self._commented_sql = (
+                self._generate_sql_with_comments()
+                if self._session._generate_multiline_queries
+                else self._generate_sql_without_comments()
```
sfc-gh-aalam: since we have a separate property for commented sql, it should generate commented sql irrespective of parameter.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1663,12 +1771,47 @@ def __deepcopy__(self, memodict={}) -> "SetStatement":  # noqa: B006
     @property
     def sql_query(self) -> str:
         if not self._sql_query:
-            sql = f"({self.set_operands[0].selectable.sql_query})"
-            for i in range(1, len(self.set_operands)):
-                sql = f"{sql}{self.set_operands[i].operator}({self.set_operands[i].selectable.sql_query})"
-            self._sql_query = sql
+            self._commented_sql = (
+                self._generate_sql_with_comments()
+                if self._session._generate_multiline_queries
```
sfc-gh-aalam: same comment. `sql_query` should not update `commented_sql`

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1663,12 +1771,47 @@ def __deepcopy__(self, memodict={}) -> "SetStatement":  # noqa: B006
...
+    def _generate_sql_without_comments(self) -> str:
+        sql = f"({self.set_operands[0].selectable.sql_query})"
+        for i in range(1, len(self.set_operands)):
+            sql = f"{sql}{self.set_operands[i].operator}({self.set_operands[i].selectable.sql_query})"
+        return sql
+
+    def _generate_sql_with_comments(self) -> str:
+        """Generate SQL query with UUID comments for child plans."""
+        sql = f"(\n-- {self.set_operands[0].selectable.uuid}\n{self.set_operands[0].selectable.sql_query}\n-- {self.set_operands[0].selectable.uuid}\n)"
+        for i in range(1, len(self.set_operands)):
```
sfc-gh-aalam: `UUID_FORMAT`

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -430,6 +433,25 @@ def __init__(
...
+        self._uuid = from_selectable_uuid if from_selectable_uuid else str(uuid.uuid4())
+        from snowflake.snowpark._internal.analyzer.select_statement import (
+            Selectable,
+        )
+
+        last_query = self.queries[-1]
+        child_uuids = []
+        for child in self.children_plan_nodes:
+            if isinstance(child, Selectable) or isinstance(child, SnowflakePlan):
+                child_uuids.append(child.uuid)
```
sfc-gh-aalam:
```suggestion
            if isinstance(child, (Selectable, SnowflakePlan)):
                child_uuids.append(child.uuid)
```

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -745,9 +766,10 @@ def build(
         is_ddl_on_temp_object: bool = False,
     ) -> SnowflakePlan:
         select_child = self.add_result_scan_if_not_select(child)
+        commented_sql = sql_generator(select_child.queries[-1].sql)
         queries = select_child.queries[:-1] + [
             Query(
-                sql_generator(select_child.queries[-1].sql),
+                commented_sql,
                 query_id_place_holder="",
```
sfc-gh-aalam: is this change still needed?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -761,7 +783,6 @@ def build(
                 child.schema_query is not None
             ), "No schema query is available in child SnowflakePlan"
             new_schema_query = schema_query or sql_generator(child.schema_query)
-
```
sfc-gh-aalam: undo

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -835,18 +856,18 @@ def build_binary(
...
-                    select_left.queries[-1].sql, select_right.queries[-1].sql
-                ),
+                commented_sql,
                 params=[
                     *select_left.queries[-1].params,
                     *select_right.queries[-1].params,
                 ],
             )
         ]
-
```
sfc-gh-aalam: undo

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -835,18 +856,18 @@ def build_binary(
...
 
+        commented_sql = sql_generator(
+            select_left.queries[-1].sql, select_right.queries[-1].sql
+        )
         queries = merged_queries + [
             Query(
-                sql_generator(
-                    select_left.queries[-1].sql, select_right.queries[-1].sql
-                ),
+                commented_sql,
```
sfc-gh-aalam: same

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -430,6 +433,25 @@ def __init__(
...
             )
+        self._uuid = from_selectable_uuid if from_selectable_uuid else str(uuid.uuid4())
+        from snowflake.snowpark._internal.analyzer.select_statement import (
+            Selectable,
+        )
+
+        last_query = self.queries[-1]
+        child_uuids = []
+        for child in self.children_plan_nodes:
+            if isinstance(child, Selectable) or isinstance(child, SnowflakePlan):
```
sfc-gh-aalam: can we add this change in a new SnowflakePlan method

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1900,3 +1904,144 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+def remove_comments(sql_query: str, uuids: List[str]) -> str:
+    """Removes comments associated with child uuids in a query"""
+    comment_placeholders = {f"-- {uuid}" for uuid in uuids}
+    return "\n".join(
```
sfc-gh-aalam: UUID_FORMAT here as well

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1900,3 +1904,144 @@ def wrapper(*args, **kwargs):
...
+        - (4, 4, "child-uuid-2")
+
+        To do so, we split the commented_sql_query into lines and check for comments associated with child uuids. We find all
+        intervals associated with child uuids and associate the remaining lines with the parent uuid.
+    """
+    from snowflake.snowpark._internal.analyzer.select_statement import QueryLineInterval
+
+    sql_lines = commented_sql_query.split("\n")
+    comment_placeholders = {f"-- {uuid}": uuid for uuid in child_uuids}
+
```
sfc-gh-aalam: same

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1900,3 +1904,144 @@ def wrapper(*args, **kwargs):
...
+                return mid
+            elif line_number < start:
+                right = mid - 1
+            else:
+                left = mid + 1
+        return -1
+
+    # traverse the plan tree to find the plan that contains the line number
+    stack = [(plan_node, line_number)]
+    while stack:
```
sfc-gh-aalam, line 2031: this is an interesting way of using stack. very cool

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-daviwang: We don't want a new line following the comment at the end of this string. I added another UUID formatter without the newline character to cover this case (which is also the case for sql_in_subquery_with_uuid)

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-daviwang: the function name is kind of misleading because the generate_sql_without_comments also doesn't add new lines to the sql query. i'll rename the functions so that is more clear

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-daviwang: nope, the snowflake_plan this SelectSnowflakePlan is derived from should already have constructed query_line_intervals so there is no need for commented_sql. I'll delete this field.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-daviwang: since we made commented_sql an abstract method i still need to keep this, but i can make it just commented_sql

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -791,7 +831,11 @@ def snowflake_supported_join_statement(
     condition: str,
     match_condition: str,
     use_constant_subquery_alias: bool,
+    left_uuid: Optional[str] = None,
+    right_uuid: Optional[str] = None,
 ) -> str:
+    LEFT_UUID = UUID_FORMAT.format(left_uuid) if left_uuid else EMPTY_STRING
+    RIGHT_UUID = UUID_FORMAT.format(right_uuid) if right_uuid else EMPTY_STRING
```
sfc-gh-jrose: This pattern is repeated often enough that it might be useful to have a helper:
```
def format_uuid(uuid: str) -> str:
    return UUID_FORMAT.format(uuid) if left_uuid else EMPTY_STRING
```

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1030,10 +1086,14 @@ def create_table_as_select_statement(
...
-    row_count: str, offset: str, child: str, on_top_of_order_by: bool
+    row_count: str,
+    offset: str,
+    child: str,
+    on_top_of_order_by: bool,
+    child_uuid: Optional[str] = None,
 ) -> str:
     return (
-        f"{child if on_top_of_order_by else project_statement([], child)}"
+        f"{child if on_top_of_order_by else project_statement([], child, child_uuid= child_uuid)}"
```
sfc-gh-jrose:
```suggestion
        f"{child if on_top_of_order_by else project_statement([], child, child_uuid=child_uuid)}"
```

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -286,6 +295,26 @@ def sql_query(self) -> str:
...
+        (
+            <child subquery>
+        )
+        -- child_uuid
+        This is used in get_snowflake_plan to generate query_line_intervals for the last query.
+        In the case of SelectSnowflakePlan and SelectTableFunction, the snowflake plan is either
+        passed in the constructor or resolved by the analyzer, so we do not need commented_sql.
+        For these classes, we just return the sql_query.
+        """
+        pass
```
sfc-gh-jrose, line 317: Is this WIP?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -969,7 +1086,7 @@ def sql_query(self) -> str:
             if self.distinct_
             else snowflake.snowpark._internal.utils.EMPTY_STRING
         )
-        self._sql_query = (
+        return (
```
sfc-gh-jrose, line 1089: Why remove this cache?

src/snowflake/snowpark/context.py
```diff
@@ -34,10 +34,14 @@
 # Following are internal-only global flags, used to enable development features.
 _enable_dataframe_trace_on_error = False
 
+# This is an internal-only global flag, used to determine whether to generate multiline queries and enable query line tracking.
+_generate_multiline_queries = False
+
```
sfc-gh-aalam: lets rename this to `_enable_trace_sql_errors_to_dataframe` else we would confuse it with https://github.com/snowflakedb/snowpark-python/commit/890889cf7cdc00b5af1979603613dfb447703e08

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -2138,6 +2235,18 @@ class PlanQueryType(Enum):
...
+class QueryLineInterval:
+    def __init__(self, start: int, end: int, uuid: str) -> None:
+        self.start = start
+        self.end = end
+        self.uuid = uuid
+
+    def __repr__(self) -> str:
+        return (
+            f"QueryLineInterval(start={self.start}, end={self.end}, uuid={self.uuid})"
+        )
```
sfc-gh-jrose: If it fits in with the style of the file a dataclass would be more concise:
```suggestion
@dataclass
class QueryLineInterval:
    start: int
    end: int
    uuid: str
```

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -941,7 +1034,31 @@ def sql_query(self) -> str:
         if not self.has_clause and not self.has_projection:
             self._sql_query = self.from_.sql_query
             return self._sql_query
-        from_clause = self.from_.sql_in_subquery
+        commented_sql = self._generate_sql()
+        self._sql_query = remove_comments(commented_sql, [self.from_.uuid])
```
sfc-gh-aalam: how about we update the signature of `_generate_sql` to `_generate_sql(self, generate_uuid_comments:bool) -> str`. That way we can call `self._generate_sql(False)` here and `self._generate_sql(True)` below. Not need for `remove_comments` here.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -1663,12 +1788,45 @@ def __deepcopy__(self, memodict={}) -> "SetStatement":  # noqa: B006
...
-            for i in range(1, len(self.set_operands)):
-                sql = f"{sql}{self.set_operands[i].operator}({self.set_operands[i].selectable.sql_query})"
-            self._sql_query = sql
+            from snowflake.snowpark.context import _generate_multiline_queries
+
+            commented_sql = (
+                self._generate_sql_with_comments_and_new_lines()
+                if _generate_multiline_queries
+                else self._generate_sql_without_comments_and_new_lines()
+            )
```
sfc-gh-aalam: sql_query always generates queries without comments so no need for this check here. Instead just use the uncommented queries.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -649,6 +654,34 @@ def reset_cumulative_node_complexity(self) -> None:
...
+        Sets the query line intervals for the last query in the queries list. When multiline
+        queries are enabled, we expect the last query to have an SQL query that is commented with
+        the uuids of its children. We use these uuids to construct the query line intervals,
+        and then remove the comments to get the final SQL query. When multiline queries is not enabled,
+        this function will assign responsibility for the single line to the current plan, and
+        make no adjustments to the SQL query.
+        """
+        from snowflake.snowpark._internal.analyzer.select_statement import (
+            Selectable,
+        )
```
sfc-gh-aalam, line 672: let's add param protection here as well. something like 
```
if _enable_trace_sql_errors_to_dataframe:
    return
```

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -203,6 +203,8 @@
 ADD_FIELDS = " ADD FIELDS"
 NEW_LINE = "\n"
 TAB = "    "
+UUID_FORMAT = "-- {}\n"
+UUID_COMMENT = "-- {}"
```
sfc-gh-aalam: do you think we can just use one? say, `UUID_COMMENT_FORMAT`?

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-daviwang: no it's an abstract method implemented by the other classes. I'll add some clarification in the comments

sfc-gh-jrose: Ah, I missed the abstractmethod decorator.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-daviwang: it's cached above in the sql_query property

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-daviwang: i'll just use UUID_COMMENT - it's probably better to be explicit about new lines, i was just being lazy

tests/integ/test_session.py
```diff
@@ -416,6 +416,7 @@ def test_create_session_from_connection(
     session_builder = Session.builder.configs({"connection": connection})
     new_session = session_builder.create()
     new_session.sql_simplifier_enabled = sql_simplifier_enabled
+    print("FUCK ME", new_session._generate_multiline_queries)
```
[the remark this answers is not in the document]
sfc-gh-joshi: @sfc-gh-daviwang consider pushing the change in question before marking it as resolved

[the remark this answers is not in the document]
sfc-gh-daviwang: my bad, i had two of these debug messages and i deleted a different one (which was test_session_suite and not test_session) and thought this was the same one

tests/integ/modin/conftest.py
```diff
@@ -284,6 +285,8 @@ def indices_dict():
 @pytest.fixture(scope="module", autouse=True)
 def session(session):
     session._disable_multiline_queries()
+    # We disable trace_sql_errors because it requires multiline queries to be enabled
+    context.configure_development_features(enable_trace_sql_errors_to_dataframe=False)
```
sfc-gh-joshi: Instead of disabling this manually in tests, should we check `_PYTHON_SNOWPARK_GENERATE_MULTILINE_QUERIES` before using `enable_trace_sql_errors_to_dataframe`? If it doesn't make sense for the latter to be enabled without the former, then the code reading that variable should reflect that behavior accordingly.

sfc-gh-daviwang: Yeah that makes sense, i can do that instead

sfc-gh-daviwang: I ended up not doing this, but now enabling tracing sql errors should not depend on multiline queries

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -941,7 +1037,28 @@ def sql_query(self) -> str:
...
+            return self._commented_sql
+        self._commented_sql = self._generate_sql(generate_uuid_comments=True)
+        return self._commented_sql
+
+    def _generate_sql(self, generate_uuid_comments: bool) -> str:
+        """Generate SQL query with UUID comments for child plans if multiline queries are enabled.
+        Otherwise, generate SQL query without comments."""
+        from_clause = (
+            (f"{self.from_.sql_in_subquery_with_uuid}")
+            if generate_uuid_comments
```
sfc-gh-aalam:
```suggestion
            self.from_.sql_in_subquery_with_uuid
            if generate_uuid_comments
```

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -203,10 +203,22 @@
 ADD_FIELDS = " ADD FIELDS"
 NEW_LINE = "\n"
 TAB = "    "
+UUID_COMMENT = "-- {}"
 
 TEMPORARY_STRING_SET = frozenset(["temporary", "temp"])
 
 
+def format_uuid(uuid: Optional[str], with_new_line: bool = False) -> str:
+    """
```
sfc-gh-aalam: lets make it True by default

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -649,6 +655,37 @@ def reset_cumulative_node_complexity(self) -> None:
...
+        and then remove the comments to get the final SQL query. When tracing sql errors is not enabled,
+        this function will do nothing.
+        """
+        import snowflake.snowpark.context as context
+
+        if not context._enable_trace_sql_errors_to_dataframe:
+            return
+        from snowflake.snowpark._internal.analyzer.select_statement import (
+            Selectable,
+        )
```
sfc-gh-aalam:
```suggestion
        if not context._enable_trace_sql_errors_to_dataframe:
            # No action needed if feature is not enabled
            return
            
        from snowflake.snowpark._internal.analyzer.select_statement import Selectable
```

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1831,3 +1835,144 @@ def get_sorted_key_for_version(version_str):
...
+        For this query, we want to return a list of QueryLineInterval objects with the following intervals:
+        - (0, 1, "parent-uuid")
+        - (2, 2, "child-uuid-1")
+        - (3, 3, "parent-uuid")
+        - (4, 4, "child-uuid-2")
+
+        To do so, we split the commented_sql_query into lines and check for comments associated with child uuids. We find all
+        intervals associated with child uuids and associate the remaining lines with the parent uuid.
+    """
+    from snowflake.snowpark._internal.analyzer.select_statement import QueryLineInterval
```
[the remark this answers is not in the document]
sfc-gh-aalam: Isn't this suggestion correct?

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/utils.py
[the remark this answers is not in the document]
sfc-gh-daviwang: yeah i can change it

sfc-gh-heshah (approved):

PR #3418, opened 2025-06-02
SNOW-2126855: Allow table existence to be specified

sfc-gh-jrose:

   

   Fixes SNOW-2126855


src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -335,6 +336,8 @@ def save_as_table(
                 * storage_serialization_policy: specifies the storage serialization policy for the table
 
                 * iceberg_version: Overrides the version of iceberg to use. Defaults to 2 when unset.
+            table_exists: This parameter allows skipping a potentially expensive table lookup if the existence of the table is known. When
```
sfc-gh-aling: what would happen if users accidentally set this to True, but the table does not exist?

sfc-gh-jrose: They will get a SQLException. There is a test case that covers such a situation.

sfc-gh-aling: I feel the name could be more explicit -- just name it "skip_table_exists_check"? the current name makes me think this is a parameter related to overwrite existing table

sfc-gh-jrose: With the current parameter you have these definitions:
```
True: table does exist
False: table does not exist
None: need to look up the table to figure it out.
```

With the proposed name change it sounds like this:
```
True: Skip the check.
False: Do the check
```
What value should we assume though when skipping?

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - Added a new argument to `Dataframe.describe` called `strings_include_math_stats` that triggers `stddev` and `mean` to be calculated for String columns.
 - Improved the error message for `Session.write_pandas()` and `Session.create_dataframe()` when the input pandas DataFrame does not have a column.
 - Added support for retrieving `Edge.properties` when retrieving lineage from `DGQL` in `DataFrame.lineage.trace`.
+- Added a parameter `table_exists` to `DataFrameWriter.save_as_table` that allows specifying if a table already exists. This allows skipping a table lookup that can be expensive.
```
sfc-gh-jdu, line 37: why is a table lookup expensive?

sfc-gh-jrose: I'm not sure, but we received a report from a solutions engineer that they profiled `show tables` to take almost 1 minute in an account containing many tables.

src/snowflake/snowpark/dataframe_writer.py
sfc-gh-aalam: If we go with `skip_table_exists_check`, the the behavior should be

```
True: skip table_exists check, assume table exists and append data into table.
False: do the check to determine if table exists, create if it does not exists, and then append to table.
```

sfc-gh-aalam (approved):

sfc-gh-jdu (approved):

PR #3429, opened 2025-06-03
SNOW-2062444: Add ttl cache to describe queries

sfc-gh-jrose:

   

   Fixes SNOW-2062444

   This PR adds a ttl cache to describe queries. Describe queries are cached for up to 15 seconds and the attributes are returned from the cache if the query is repeated within that interval. This matches behavior used by SAS. Any non-select query invalidates the cache in case those queries modify schemas in any way. 

Due to this change several tests on both the snowpark and pandas side that test for a certain number of describe queries would become flaky. I've added a patch to those tests that bypasses the cache so that they describe results as expected.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Added support for PostgreSQL in `DataFrameReader.dbapi` (PrPr) for both Parquet and UDTF-based ingestion.
 - Added support for Databricks in `DataFrameWriter.dbapi` (PrPr) for UDTF-based ingestion.
 - Added support to `DataFrameReader` to enable use of `PATTERN` when reading files with `INFER_SCHEMA` enabled.
+- Added a ttl cache to describe queries. Repeated queries in a 15 second interval will used the cached value rather than requery Snowflake.
```
sfc-gh-jdu: How is this number determined?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -501,6 +502,24 @@ def run_query(
                 if not kwargs.get("_statement_params"):
                     kwargs["_statement_params"] = {}
                 kwargs["_statement_params"]["SNOWPARK_SKIP_TXN_COMMIT_IN_DDL"] = True
+            if (
+                query.strip()
```
sfc-gh-jdu: why clear cache?

sfc-gh-aalam: how about we just support select and anything else is cleared?
https://github.com/snowflakedb/snowpark-python/blob/af8073837280a85438daf1b572e02e937d8b664e/src/snowflake/snowpark/_internal/utils.py#L388

sfc-gh-aalam (commented):
we should also protect this with reduce-describe query param

CHANGELOG.md
sfc-gh-jrose: This is the amount that SAS uses for their queries today. I could make it configurable if desired.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-jrose: > why clear cache?

Whenever a DDL statement is executed we can no longer guarantee that the cached results are valid. There might be a way to only invalidate only cached queries that reference modified resources, but such an approach would be much higher overhead I think.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -135,6 +136,13 @@
     from collections.abc import Iterable
 
 
+@ttl_cache(ttl_seconds=15)
+def _analyze_attributes(
+    sql: str, session: "snowflake.snowpark.session.Session"  # type: ignore
```
sfc-gh-aalam: can we rename this to cached_analyze_attributes so it doesn't get confused with `SnowflakePlan._analyze_attributes`

PR #3438, opened 2025-06-04
SNOW-2110972: Allow user override for AST collection

sfc-gh-aalam:

   

   Fixes SNOW-2110972

   Allow users to enable/disable AST collection.


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1003,11 +1009,22 @@ def set_state(self, source: AstFlagSource, enable: bool) -> None:
...
+                    source,
                 )
                 return
+
+            if source == AstFlagSource.USER:
+                # User is allowed to override the server setting if it is not
+                # finalized by test.
+                self._ast_enabled = enable
+                self._state = _AstFlagState.USER_SET
+                return
```
sfc-gh-oplaton, line 1026: I think this would allow user code to perform the problematic disabled -> enabled transition. We don't want that.

What happens if you write a simple test that does roughly:

* Create a session.
* Explicitly set `ast_enabled` to `False`.
* Create a trivial expression `col("foo") + 1`.
* Explicitly set `ast_enabled` to `True`.
* Evaluate a `Dataframe` expression that involves the trivial expression created earlier, and tries to `collect` or `show`.

I think the code should simplify the expression for `safe_transition`, below: `safe_transition: bool = self._ast_enabled or not enable`. (Dropped the state check and pushed `not` inside the remaining expression). This code needs to move after line 1032, after computing `safe_transition`, and would also check the value of `safe_transition`.

sfc-gh-aalam: as discussed offline, this is intended to allow the unsafe transition. I'll add a comment explicitly mentioning this.

sfc-gh-oplaton (approved):

sfc-gh-yuwang (approved):

PR #3469, opened 2025-06-16
SNOW-2108634: Adding python source trace to debug context

sfc-gh-daviwang:

   Fixes SNOW-2108634

 I added a new debug util function to map SQL compilation errors to the offending Python line numbers. This additional information is added to debug context. As an example: for this erroneous piece of code:
```
session = Session.builder.config("connection_name", "azure").create()
df = session.create_dataframe([[1, 2], [3, 4]], schema=["a", "b"])
df1 = df.select(col("a").alias("x"), col("b").alias("y"))
df2 = df1.select(col("x") + col("a"))  # ERROR: 'a' no longer exists after aliasing
df2.collect()
```
The statement:
"SQL compilation error corresponds to Python source at lines 19, columns 10-41"
is added to debug context.


src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -198,3 +199,55 @@ def get_df_transform_trace_message(
...
+    )
+    match = sql_compilation_error_regex.match(error_msg)
+    if not match:
+        return None
+
+    sql_line_number = int(match.group(1)) - 1
+    from snowflake.snowpark._internal.analyzer.snowflake_plan import SnowflakePlan
+    from snowflake.snowpark._internal.analyzer.select_statement import (
+        Selectable,
+    )
```
sfc-gh-aalam:
```suggestion
    from snowflake.snowpark._internal.analyzer.select_statement import Selectable
```
nit

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -198,3 +199,55 @@ def get_df_transform_trace_message(
...
+
+    for arg in args:
+        if isinstance(arg, SnowflakePlan) or isinstance(arg, Selectable):
+            try:
```
sfc-gh-aalam: these args should be processed in SnowflakePlan wrapper. We should pass a SnowflakePlan as an arg here

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -198,3 +199,55 @@ def get_df_transform_trace_message(
...
+                from snowflake.snowpark._internal.utils import (
+                    get_plan_from_line_numbers,
+                )
+
+                plan = get_plan_from_line_numbers(arg, sql_line_number)
+                ast_id = None
+                bind_stmt = None
+                if isinstance(plan, Selectable) and plan.df_ast_ids is not None:
+                    ast_id = plan.df_ast_ids[-1]
+                    bind_stmt = plan._session._ast_batch._bind_stmt_cache.get(ast_id)
```
sfc-gh-aalam: I think in this case, when there are multiple ast_ids associated, we should list sources of all dataframes as possible source of error

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -198,3 +199,55 @@ def get_df_transform_trace_message(
...
+                elif isinstance(plan, SnowflakePlan) and plan.df_ast_id is not None:
+                    ast_id = plan.df_ast_id
+                    bind_stmt = plan.session._ast_batch._bind_stmt_cache.get(ast_id)
+                if not ast_id:
+                    continue
+                if not bind_stmt:
+                    continue
```
sfc-gh-aalam:
```suggestion
                if ast_id is None or bind_stmt is None:
                    continue
```

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -2031,7 +2031,7 @@ def find_interval_containing_line(intervals, line_number):
     while stack:
         node, line_number = stack.pop()
         if isinstance(node, Selectable):
-            node = node.get_snowflake_plan(skip_schema_query=True)
+            node = node.get_snowflake_plan(skip_schema_query=False)
```
sfc-gh-aalam, line 2034: curious to know why is this change needed?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -187,16 +196,19 @@ def wrap(*args, **kwargs):
...
-                    df_transform_debug_trace = None
+                    df_transform_debug_trace = ""
                     try:
                         if (
                             _enable_dataframe_trace_on_error
                             and df_ast_id is not None
                             and stmt_cache is not None
                         ):
-                            df_transform_debug_trace = get_df_transform_trace_message(
+                            trace_result = get_df_transform_trace_message(
```
sfc-gh-aalam: I think it is easier to return empty string here https://github.com/snowflakedb/snowpark-python/blob/b75739c6b999b1b689d0652b7928d64c08583498/src/snowflake/snowpark/_internal/debug_utils.py#L181 to minimize the changes

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -179,6 +180,14 @@ def wrap(*args, **kwargs):
                     query = getattr(e, "query", None)
                     tb = sys.exc_info()[2]
                     assert e.msg is not None
+                    python_debug_context = ""
```
sfc-gh-aalam: let's name it `error_source_context`

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -205,9 +217,16 @@ def wrap(*args, **kwargs):
                         )
                         pass
 
+                    debug_header = "\n\n--- Additional Debug Information ---\n"
+                    debug_context = (
+                        debug_header + df_transform_debug_trace + python_debug_context
+                    )
```
sfc-gh-aalam: let's also handle the case when both these traces are empty strings. In that case, we shouldn't add debug_header without a debug info

src/snowflake/snowpark/_internal/ast/utils.py
```diff
@@ -1645,3 +1645,23 @@ def clear_line_no_in_request(request: proto.Request) -> None:
...
+        return expr.src
+
+    #  recursively search for src
+    for field, value in expr.ListFields():
+        if field.name == "src":
+            return value
+        else:
+            result = extract_src_from_expr(value)
+            if result:
+                return result
```
sfc-gh-jrose:
```suggestion
            return extract_src_from_expr(value)
```

sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/utils.py
sfc-gh-daviwang: i was failing this test: tests/integ/test_stored_procedure.py::test_stored_procedure_with_basic_column_datatype with error No schema query provided.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-daviwang: i check for this case in lines 224-225

src/snowflake/snowpark/dataframe.py
```diff
@@ -670,9 +670,24 @@ def _ast_id(self, value: Optional[int]) -> None:
...
+            self._propagate_ast_id_to_select_sql(self._plan, value)
         if self._select_statement is not None:
             self._select_statement.add_df_ast_id(value)
 
+    def _propagate_ast_id_to_select_sql(self, plan, ast_id: Optional[int]) -> None:
+        """
+        Propagate df_ast_id to associated SelectSQL instances.
+        """
+        from snowflake.snowpark._internal.analyzer.select_statement import SelectSQL
+
```
sfc-gh-aalam: I think we can extract this change out of this PR into a smaller PR to fix ast_id propagation. Also, let's handle this propagation within SnowflakePlan method instead of DataFrame class

src/snowflake/snowpark/_internal/ast/utils.py
```diff
@@ -1645,3 +1645,23 @@ def clear_line_no_in_request(request: proto.Request) -> None:
     fixes in determining better line_no info for chained python code, etc."""
     for stmt in request.body:
         clear_line_no_in_ast(stmt)
+
+
+def extract_src_from_expr(expr: proto.Expr) -> Optional[Any]:
```
sfc-gh-aalam, line 1650: @sfc-gh-heshah could you take a look at this and give your feedback. I don't have much context if this covers all cases.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -179,15 +180,24 @@ def wrap(*args, **kwargs):
...
                             break
-                    df_transform_debug_trace = None
+                    error_source_context = ""
+                    if (
+                        "SQL compilation error:" in e.msg
+                        and "error line" in e.msg
+                        and top_plan is not None
+                    ):
+                        error_source_context = get_python_source_from_sql_error(
+                            top_plan, e.msg
```
sfc-gh-aalam: lets move this part into the try catch below so we log any errors using `_logger.info`

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -198,3 +202,71 @@ def get_df_transform_trace_message(
...
+    match = sql_compilation_error_regex.match(error_msg)
+    if not match:
+        return ""
+
+    sql_line_number = int(match.group(1)) - 1
+
+    try:
+        from snowflake.snowpark._internal.utils import (
+            get_plan_from_line_numbers,
+        )
```
sfc-gh-aalam: remove this try catch since we will wrap this in try-except inside snowflake plan

src/snowflake/snowpark/_internal/ast/utils.py
sfc-gh-heshah: This should cover all cases. 

From what I gather from this work and the PR context, when we get a SQL compilation error, we provide the trace of dataframe operations that were involved in the debug info. If we are only extracting the source information from the topmost Dataframe API calls in the chain of operations, how can we be certain that these correspond to the SQL compilation error? As opposed to a nested/earlier Dataframe API call in the same line?

sfc-gh-daviwang: get_python_source_from_sql_error uses get_plan_from_line_numbers to traverse through the topmost Dataframe API call's children to find which plan is responsible for the line that causes the error

sfc-gh-heshah: Ah I see, that's pretty clever. So the mapping is going from: SQL query lines --> `SnowflakePlan` --> `dataframe._ast_id` --> source info.

In that case, was there a specific need for recursing into the proto to extract source information you found? If this is only ever called on `bind` statements, they should always carry source information. No harm in keeping the code there as a precaution, just curious.

sfc-gh-heshah (approved):

src/snowflake/snowpark/_internal/ast/utils.py
sfc-gh-daviwang: yeah that's correct. I don't think i found a case where src wasn't 2 levels down (eg. something like expr.dataframe_select.src) but I included the recursion just in case

PR #3473, opened 2025-06-17
SNOW-2150552: Add TRY_CAST and USE_RELAXED_TYPES options to DataFrameReader

sfc-gh-jrose:

   

   Fixes SNOW-2150552

   This PR adds two new options to file loading in order to improve user experience.
   
   `TRY_CAST` - Today if you try to load data that does not match the expected schema you will either get an error, or if `ON_ERROR = SKIP_FILE` the file will be ignored altogether. This can be undesirable because a single line can invalidate an entire file. With `TRY_CAST` enabled each column uses a TRY_CAST instead of a `::` cast. This allows invalid values to be replaced with NULL instead of failing.

    `USE_RELAXED_TYPES` - Schema inference can lead to very narrow types being returned. EG `1.2` would get inferred to `NUMBER(2, 1)`. If other files have data of other precisions this can lead to data truncation. This option causes inferred schemas to use max length strings for all string data and Double for all numeric data.

I'm not particularly tied to either option name if there's something better we want to call them.


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -1255,6 +1256,53 @@ def type_string_to_type_object(type_str: str) -> DataType:
...
+    elif isinstance(datatype, MapType):
+        return MapType(
+            None
+            if datatype.key_type is None
+            else most_permissive_type(datatype.key_type),
+            None
+            if datatype.value_type is None
+            else most_permissive_type(datatype.value_type),
+            structured=datatype.structured,
+            value_contains_null=datatype.value_contains_null,
```
sfc-gh-daviwang: do we want to keep the contains_null value of the original datatype? imo if a field is as permissive as possible, it should always be nullable

PR #3476, opened 2025-06-20
SNOW-1473621: Add support for mapInPandas

sfc-gh-jrose:

   

   Fixes SNOW-1473621

   This PR adds a new function that allows users to map a pandas function across a dataframe. See similar functionality in pyspark: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.mapInPandas.html

### DIFF
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -89,7 +89,9 @@ def pytest_collection_modifyitems(config, items):
     This function is used to skip some doctests on certain conditions.
     For example, some functions in PrPr might not be available on Azure and GCP.
     """
-    host = conn_params.get("host") or conn_params.get("HOST")
+    host = (
+        conn_params.get("host") or conn_params.get("HOST") or conn_params.get("account")
+    )
     if any(platform in host.split(".") for platform in ["gcp", "azure"]):
         skip = pytest.mark.skip(reason="Skipping doctest for Azure and GCP deployment")
         disabled_doctests = ["ai_classify"]  # Add any test names that should be skipped
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -136,6 +136,7 @@ from snowflake.snowpark._internal.type_utils import (
     ColumnOrSqlExpr,
     LiteralType,
     snow_type_to_dtype_str,
+    type_string_to_type_object,
 )
 from snowflake.snowpark._internal.udf_utils import add_package_to_existing_packages
 from snowflake.snowpark._internal.utils import (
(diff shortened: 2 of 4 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -6679,3 +6680,83 @@ def process(self, *argv):
     return dataframe.join_table_function(
         map_udtf(*df_columns).over(partition_by=partition_by)
     ).select(*output_columns)
+
+
+def map_in_pandas(
```
sfc-gh-aling, line 6692: why do we introduce this function, is this for migration? if so I see in pyspark this is a method on dataframe instead of a module function (or I get the wrong doc): https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.mapInPandas.html

do we need to align with them?

sfc-gh-jrose: This is for compatibility. 

Ideally this would be a dataframe function, but any addition to DataFrame is considered a bcr due to possibly shadowing column names. I think there was a discussion when we added `map` that we should add them at module level and then in the next BCR move them into the DataFrame api.

sfc-gh-aalam: can we also support/expose the following kwargs that are also supported in `map`. We originally added them in map for giving users ability to distribute their

```
    output_column_names: Optional[List[str]] = None,
    imports: Optional[List[Union[str, Tuple[str, str]]]] = None,
    packages: Optional[List[Union[str, ModuleType]]] = None,
    immutable: bool = False,
    partition_by: Optional[Union[ColumnOrName, List[ColumnOrName]]] = None,
    vectorized: bool = False,
    max_batch_size: Optional[int] = None,
```

sfc-gh-jrose: I've added a `partition_by` parameter as well as passed kwargs through to the udtf registration to cover some of the other parameters. `output_column_names` is still left out as I think it's already covered by `schema`.

sfc-gh-aalam: I remember having this discussion about just exposing kwargs for map with yijun and he suggested we add only a few of the kwargs explicitly. Since this is mapInPandas, vectorized is already `True`. can we add imports, packages, and max_batch_size.

src/conftest.py
```diff
@@ -89,7 +89,9 @@ def pytest_collection_modifyitems(config, items):
     This function is used to skip some doctests on certain conditions.
     For example, some functions in PrPr might not be available on Azure and GCP.
     """
-    host = conn_params.get("host") or conn_params.get("HOST")
+    host = (
+        conn_params.get("host") or conn_params.get("HOST") or conn_params.get("account")
```
sfc-gh-aling, line 93: nit: what is this change for, seems unrelated to the code?

sfc-gh-jrose: My local doctests fail because I use `account` config rather than `host` config so I updated this to be able to run the doctests I added in this change.

sfc-gh-aalam (approved):

PR #3386, opened 2025-05-23
SNOW-1285387: add local testing and support for snowurl get/put

sfc-gh-lju:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-1285387

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.
   - [ ] I acknowledge that I have ensured my changes to be thread-safe. Follow the link for more information: [Thread-safe Developer Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#thread-safe-development)
   - [ ] If adding any arguments to public Snowpark APIs or creating new public Snowpark APIs, I acknowledge that I have ensured my changes include AST support. Follow the link for more information: [AST Support Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#ast-abstract-syntax-tree-support-in-snowpark)

3. Please describe how your code solves the related issue.

  Added tests to verify that snowurls work with session.file.get and .put in local testing. Removed unnecessary error check in stage_registry to allow for snowurls and make get and read_file consistent with put.

PR #3479, opened 2025-06-22
Lju snow 2110578 snowflake file mocked stage support

sfc-gh-lju:

   

   Fixes SNOW-NNNNNNN


CHANGELOG.md
```diff
@@ -55,6 +56,7 @@
 - Added support for PostgreSQL in `DataFrameReader.dbapi` (PrPr) for both Parquet and UDTF-based ingestion.
 - Added support for Databricks in `DataFrameWriter.dbapi` (PrPr) for UDTF-based ingestion.
 - Added support to `DataFrameReader` to enable use of `PATTERN` when reading files with `INFER_SCHEMA` enabled.
+- Added local testing support for reading files with `SnowflakeFile` using local file paths.
```
sfc-gh-aissa: Needs to be removed

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Added a new option `TRY_CAST` to `DataFrameReader`. When `TRY_CAST` is True columns are wrapped in a `TRY_CAST` statement rather than a hard cast when loading data.
 - Added a new option `USE_RELAXED_TYPES` to the `INFER_SCHEMA_OPTIONS` of `DataFrameReader`. When set to True this option casts all strings to max length strings and all numeric types to `DoubleType`.
 - Added debuggability improvements to eagerly validate dataframe schema metadata. Enable it using `snowflake.snowpark.context.configure_development_features()`.
+- Added local testing support for reading files with `SnowflakeFile` using stage paths and snowurls.
```
sfc-gh-aissa: Move to snowpark local testing section, probably just append the existing one to say:

Added local testing support for reading files with `SnowflakeFile` using local file paths, stage paths, and snowurls.

src/snowflake/snowpark/files.py
```diff
@@ -112,21 +116,49 @@ def __init__(
...
+        if self._is_local_file and mode in READ_MODES:
+            # Buffered Reader used to support BufferedIOBase methods such as read1 and readinto1
+            self._file_stream = BufferedReader(
+                open(self._file_location, self._mode), _DEFAULT_READ_BUFFER_SIZE
+            )
+
+            # SEEK_CUR and SEEK_END are only supported in binary mode for Python IO so we must obtain the size of the file
+            # in a different way. We can use the raw stream to get the size of the file.
+            temp_file = open(self._file_location, "rb")
+            self._file_size = temp_file.seek(0, SEEK_END)
```
sfc-gh-aissa, line 152: Is there no cleaner way to do this? If no, does seek return the current position, or do you need tell to set the size?

Would it make sense to use the actual file stream, then seek to the end and back to the beginning after? I guess if customer opens in read mode, then we can't do that. Just wanting to make sure I understand the necessity of this.

src/snowflake/snowpark/files.py
```diff
@@ -112,21 +116,49 @@ def __init__(
...
+            temp_file.close()
+        elif self._is_stage_file and mode in READ_MODES:
+            self._file_stream = get_active_session().file.get_stream(
+                self._file_location
+            )
+            self._file_size = self._file_stream.seek(0, SEEK_END)
+            self._file_stream.seek(0, SEEK_SET)
```
sfc-gh-aissa, line 159: This is what I was saying above, is this doable here because the file stream is always in rb mode?

src/snowflake/snowpark/files.py
```diff
@@ -271,9 +303,12 @@ def read(self, size: int = -1) -> Sequence:
         self._raise_if_not_read()
         if self._is_local_file:
             content = self._file_stream.raw.read(size)
-            self._pos += len(content)
-            return content
-        raise NotImplementedError(_NON_LOCAL_PATH_ERR_MSG)
+        elif self._is_stage_file:
```
sfc-gh-aissa, line 307: Since we already validate it's a local or stage file in the init, do we need elif or can we just say else? Same question applies below, it's a nit

src/snowflake/snowpark/files.py
```diff
@@ -271,9 +303,12 @@ def read(self, size: int = -1) -> Sequence:
         self._raise_if_not_read()
         if self._is_local_file:
             content = self._file_stream.raw.read(size)
-            self._pos += len(content)
-            return content
-        raise NotImplementedError(_NON_LOCAL_PATH_ERR_MSG)
```
sfc-gh-aissa, line 276: Can _NON_LOCAL_PATH_ERR_MSG be removed now, or where do we still use this? It should be gone right?

tests/mock/test_udf_files.py
```diff
@@ -262,22 +369,36 @@ def is_writable_read(file_location: str, mode: str) -> bool:
...
     [
-        ("r", "w", 0),
-        ("r", "w", 1),
-        ("r", "w", 5),
-        ("r", "w", 10),
-        ("rb", "wb", 0),
-        ("rb", "wb", 1),
-        ("rb", "wb", 5),
-        ("rb", "wb", 10),
+        ("r", "w", 0, True),
```
sfc-gh-aissa, line 345: Do we need all these sizes? If yes can you add a comment differentiating what they are testing. I can get 0 and 1, but 5 and 10? Is 10 the full/past the file size? 5 is somewhere in the middle?

There's a bunch of unit tests where my questions apply as well. Can you add comments on why the values are needed, or if they are redundant, reduce them? 

I am trying to make sure our testing is thorough but not excessive :D

sfc-gh-aissa (commented):
Looks good, few comments, should be close.

src/snowflake/snowpark/files.py
```diff
@@ -17,8 +16,12 @@
     SEEK_END,
     SEEK_CUR,
 )
-from snowflake.snowpark._internal.utils import RELATIVE_PATH_PREFIX
+from snowflake.snowpark._internal.utils import (
```
sfc-gh-aissa, line 19: General comment for this file, are there any places where we need to update the method description to explain local testing with a stage specifically? Or are our explanations sufficient?

sfc-gh-aissa:
Go ahead and publish the PR, it doesn't need to be in draft mode. Also check the precommit please, there are some failing tests

src/snowflake/snowpark/files.py
sfc-gh-lju: I don't think so. The behavior for stages in local testing should be the same as if we passed in a relative path.

src/snowflake/snowpark/files.py
sfc-gh-lju: Yes, exactly!

src/snowflake/snowpark/files.py
sfc-gh-lju: I think this is the cleanest way to do this. Just passing in SEEK_END to seek will return the last position/the size of the file. As mentioned in the comment, since SEEK_END is only supported in binary mode (i.e., "rb" or "wb"), we can't use the actual file stream here as the user might open in read mode.

src/snowflake/snowpark/files.py
sfc-gh-lju: Yes, I will remove it!

src/snowflake/snowpark/files.py
sfc-gh-lju: Good question. In the previous PR for local testing support, one of the comments was about adding in an elif to make it clear one path is for read and one is for write. I was thinking that for readability, using elif would be better. Let me know your thoughts though!

tests/mock/test_udf_files.py
sfc-gh-lju: Makes sense, I think you're right that the tests are a bit excessive (although they do test different code paths, i.e., 5 is the same size of the buffer, 10 is past the buffer size, etc). I'll remove some of them to reduce the overhead though!

sfc-gh-lju (commented):
Thanks for the comments! I hope I provided some clarification to your questions

sfc-gh-jrose (approved):

src/snowflake/snowpark/files.py
sfc-gh-aissa: But this elif will never be hit for write, it's read() api. We already check self._raise_if_not_read()? Seems redundant to me for this specific API (and a couple others bellow), but idk the context for the previous PR

tests/mock/test_udf_files.py
sfc-gh-aissa: You should keep them then and explain that instead then, it's not excessive to test the different major code paths, especially with unit tests. I'm not worried about overhead for testing different code paths, we need the coverage. 

The overhead is low anyways, it's not like we are spinning up actual warehouses to run these tests, it's just to our local machines and some release jobs, at least afaik, as I assume the public package doesn't include our tests, only source code

edit: confirmed with Jamison tests aren't added to the release wheels

src/snowflake/snowpark/files.py
sfc-gh-lju: I meant that the comment on the previous PR was about making it readable and explicitly calling out the elif path (i.e., calling out the elif write path), not about using write here. I was thinking that the same logic should apply here. I can remove the elif and replace it with else if you want though

sfc-gh-aissa: Ah I see, I am fine either way, whatever folks feel is best/follows Python best practices. We can let other folks chime in

CHANGELOG.md
```diff
@@ -27,7 +27,7 @@
 
 ### Snowpark Local Testing Updates
 
-- Added local testing support for reading files with `SnowflakeFile` using local file paths.
+- Added local testing support for reading files with `SnowflakeFile` using local file paths, stage paths, and snowurls.
```
sfc-gh-lschmidt: Can we put examples of the types in the names. E.g. stage paths (@stage/file), snow url (snow://).

Also I am not sure that snow url is the formal name, but I am not sure. Can we look that up.

What about scoped urls or stage urls?

src/snowflake/snowpark/files.py
```diff
@@ -76,8 +79,7 @@ class SnowflakeFile(RawIOBase):
 
     These examples are using the client, but this same pattern can be used inside SQL-defined UDFs.
 
-    We provide a local implementation of SnowflakeFile to aid in local testing. This currently only supports using read APIs on relative paths.
-    Local testing to Snowflake stages is not yet supported.
+    We provide a local implementation of SnowflakeFile to aid in local testing. This currently only supports using read APIs on relative paths and mocked stages.
```
sfc-gh-lschmidt: Should we add a note or reference to what a mocked stage is?

src/snowflake/snowpark/files.py
```diff
@@ -112,21 +114,48 @@ def __init__(
...
+                open(self._file_location, self._mode), _DEFAULT_READ_BUFFER_SIZE
+            )
+
+            # SEEK_CUR and SEEK_END are only supported in binary mode for Python IO so we must obtain the size of the file
+            # in a different way. We can use the raw stream to get the size of the file.
+            temp_file = open(self._file_location, "rb")
+            self._file_size = temp_file.seek(0, SEEK_END)
+            temp_file.close()
+        elif self._is_stage_file and mode in _READ_MODES:
+            self._file_stream = get_active_session().file.get_stream(
```
sfc-gh-lschmidt, line 155: This works with mocked and non-mocked session, right? So it would handle any kind of stage?

src/snowflake/snowpark/files.py
sfc-gh-lschmidt: I don't care really but the elif is slightly more maintainable if someone adds another mode.

src/snowflake/snowpark/files.py
```diff
@@ -288,11 +320,12 @@ def read1(self, size: int = -1) -> Sequence:
...
             if self._mode == "r":
                 content = self.read(size).encode()
-            else:
+            elif self._mode == "rb":
                 content = self._file_stream.read1(size)
-            self._pos += len(content)
-            return content
-        raise NotImplementedError(_NON_LOCAL_PATH_ERR_MSG)
+        elif self._is_stage_file:
+            content = self._file_stream.read1(size)
```
sfc-gh-lschmidt, line 330: Do we not need to decode this one? Is that because read1 does it?

src/snowflake/snowpark/files.py
```diff
@@ -375,9 +413,12 @@ def readlines(self, hint: int = -1) -> list[Sequence]:
...
         if self._is_local_file:
             content = self._file_stream.raw.readlines(hint)
-            self._pos += sum(len(line) for line in content)
-            return content
-        raise NotImplementedError(_NON_LOCAL_PATH_ERR_MSG)
+        elif self._is_stage_file:
+            content = self._file_stream.readlines(hint)
+            if self._mode == "r":
+                content = [line.decode() for line in content]
+        self._pos += sum(len(line) for line in content)
```
sfc-gh-lschmidt, line 424: Do we properly handle newlines here?
Can we add some tests for unicode characters? We've had bugs with those and readlines in the past. There's some examples with a emoji in I think unicode.sql file in our tests.

tests/mock/conftest.py
```diff
@@ -7,6 +7,7 @@
 
 from snowflake.snowpark import Session
 from snowflake.snowpark.mock._connection import MockServerConnection
+from tests.utils import Utils
```
sfc-gh-lschmidt, line 10: This is a lot of tests so I will mostly leave to Alex to review.
But can we make sure we cover encryption and r/rb for all methods.
Also can we test this with real stages or where do those tests go?

src/snowflake/snowpark/files.py
sfc-gh-lju: Yes, it should handle any kind of stage!

src/snowflake/snowpark/files.py
sfc-gh-lju: Read1 should only return bytes, and as such we don't need to decode it:
image

src/snowflake/snowpark/files.py
sfc-gh-lju: I'll add some tests for unicode characters! From my tests we do handle newlines.

tests/mock/conftest.py
sfc-gh-lju: r/rb are covered for all methods! What did you mean by encryption? Also, for the real stage tests, we'll be adding that in the next PR. I should call out that the changes we made to files.py should already enable using SnowflakeFile with real stages, but we decided to separate the PRs.

src/snowflake/snowpark/files.py
sfc-gh-lju: I used the unicode characters from our test_databricks.py.
image

src/snowflake/snowpark/files.py
sfc-gh-lju: We should, I'll do that!

CHANGELOG.md
sfc-gh-lju: I'll add some examples! Also for snow url, it seems like the formal name is Snow URI? I also can't find much formal documentation, but I was looking at some slack discussions where people refer to it as the Snow URI semantic. We don't support scoped urls or stage urls, @sfc-gh-aissa do you know the reason for this?

sfc-gh-lju (commented):
Thanks for the comments Lara! I hope I addressed them

CHANGELOG.md
sfc-gh-aissa: I think "snow urls" with appropriate capitalization is fine, we discussed this with folks when you added local testing support for it as a ramp up Lawrence, it's not really documented anywhere publicly. In the [PR](https://github.com/snowflakedb/snowpark-python/pull/3386), we said snow urls.

We never planned on supporting scoped or stage urls, see: https://docs.google.com/document/d/1zR7wlQpGCPrEblnZeysQ91LxoQgzOCVJNaVkMfUCEO0/edit?tab=t.0#heading=h.emlafc49wrk7

They aren't supported at all in local testing mode, it's non-trivial to build fake URLs pointing to local files, which is why we are doing @stage, snow://, or local file path. Recall if it's a real session, scoped and stage urls will work out of box because everything is running server side, and the outputs are just passed as a df table.

tests/mock/conftest.py
sfc-gh-aissa: Discussing offline with Lawrence, the only concern for this PR we are looking into is file decryption with mocked stages. It's unclear to me if it's supported or not, but if it is, it should be handled by the mocked stage plus mocked GET file call existing logic.

For real stages, I think it will be handled by the stage interface plus GET call, which should save the decrypted version locally (pending confirmation). I found some code/tests but Lawrence will add references once we confirm with folks.

We will add tests for mocked stages if it is supported, if not, then I view it as a non-issue since the support would need to be in mocked stage+mocked GET logic, which we didn't scope building.

sfc-gh-lschmidt: If this is not yet with real stages then the encryption is probably not required. I didn't realize it didn't have that yet. 
For the mocked stage it should be okay as long as we test binary and etc.

sfc-gh-aissa:
Discussed offline, lgtm less:
- Add back removed necessary test cases, but add comment explaining their necessity (only for the 16+ combination ones, the 1/2/4/8 combos are fine)
- Change snow URI to URL

Test coverage otherwise lgtm, re-use is also good

sfc-gh-lschmidt:
Let me know when Alex approves and I can take a final pass but everything looked good to me. :)

tests/mock/test_udf_files.py
```diff
@@ -16,42 +16,87 @@
...
+    return test_msg, file_location
+
+
+def _write_test_msg_to_stage(
+    write_mode: str,
+    file_location: str,
+    tmp_stage: str,
+    session: Session,
+    test_msg: str = None,
+) -> tuple[Union[str, bytes], str]:
```
sfc-gh-lschmidt: Can we get function comments for this?

tests/mock/test_udf_files.py
```diff
@@ -16,42 +16,87 @@
 
 def _write_test_msg(
     write_mode: str, file_location: str, test_msg: str = None
-) -> Union[str, bytes]:
+) -> tuple[Union[str, bytes], str]:
```
sfc-gh-lschmidt: update comments with what this returns?

tests/mock/test_udf_files.py
```diff
@@ -16,42 +16,87 @@
...
 def _generate_and_write_lines(
-    num_lines: int, write_mode: str, file_location: str
-) -> list[Union[str, bytes]]:
+    num_lines: int,
+    write_mode: str,
+    file_location: str,
+    msg: Union[str, bytes] = None,
+) -> tuple[list[Union[str, bytes]], str]:
     """
     Generates a list of test messages and writes them to the specified file location.
```
sfc-gh-lschmidt: ditto here and below

tests/unit/conftest.py
```diff
@@ -89,3 +90,15 @@ def mock_session(mock_analyzer) -> Session:
     mock_analyzer.session = fake_session
     fake_session._enable_multiline_queries()
     return fake_session
+
+
+@pytest.fixture(scope="function")
+def session():
```
sfc-gh-lschmidt, line 96: Any comments useful here?

tests/unit/test_files.py
```diff
@@ -13,50 +13,98 @@
...
+    return test_msg, file_location
+
+
+def _write_test_msg_to_stage(
+    write_mode: str,
+    file_location: str,
+    tmp_stage: str,
+    session: Session,
+    test_msg: str = None,
+) -> tuple[Union[str, bytes], str]:
```
sfc-gh-lschmidt: Are these duplicated? Would it be worth moving to some share somewhere?

tests/unit/test_files.py
```diff
@@ -485,78 +696,148 @@ def sf_readline_with_size(
...
     else:
-        assert content == lines[0][:size]
+        assert content == lines[0][:size] or content == windows_lines[0][:size]
 
 
-@pytest.mark.parametrize(["read_mode", "write_mode"], [("r", "w"), ("rb", "wb")])
-def test_readlines_snowflakefile_local(read_mode, write_mode, tmp_path):
-    temp_file = os.path.join(tmp_path, "test.txt")
+@pytest.mark.parametrize(
+    ["read_mode", "write_mode", "use_stage"],
```
sfc-gh-lschmidt: Would it be worth putting these into a constant since you repeat the args in every test?

sfc-gh-lschmidt (approved):
Just some nits on the tests, otherwise looks good. Thanks!

tests/unit/conftest.py
sfc-gh-lju: I feel like for this one, it's pretty self explanatory, so imo no comments needed.

tests/unit/test_files.py
sfc-gh-lju: That's a good idea, i'll move it to tests/utils.py.

tests/unit/test_files.py
sfc-gh-lju: Good idea! I'll do just that

tests/mock/conftest.py
sfc-gh-aissa: Confirmed with Adam and Jamison that mocked stages don't have encryption support, and real stages it's built in (which we will show in the final PR when we add tests for real stages)

PR #3480, opened 2025-06-23
SNOW-2094325: Added fix for explode to support complex inputs

sfc-gh-aalam:

   

   Fixes SNOW-2094325

   This PR enables explode to select column expressions that can be complex expressions. Earlier only simple column expressions were accepted.

### DIFF
--- a/tests/integ/test_dataframe.py
+++ b/tests/integ/test_dataframe.py
@@ -806,14 +806,16 @@ def test_explode(session):
         df.select(df.strs, explode(df.maps).as_("primo", "secundo")), expected_result
     )
 
+    if not session.sql_simplifier_enabled:
+        return
+
     # with input as array construct
     Utils.check_answer(
         session.range(1).select(explode(array_construct(lit(1), lit(2), lit(3)))),
         [Row(VALUE="1"), Row(VALUE="2"), Row(VALUE="3")],
     )
 
-    try:
-        context._use_structured_type_semantics = True
+    with mock.patch.object(context, "_use_structured_type_semantics", True):
         # with input as array construct and cast
         Utils.check_answer(
             session.range(1).select(
@@ -821,8 +823,6 @@ def test_explode(session):
             ),
             [Row(VALUE=1), Row(VALUE=2)],
         )
-    finally:
-        context._use_structured_type_semantics = False
 
 
 @pytest.mark.skipif(


tests/integ/test_dataframe.py
```diff
@@ -804,6 +806,24 @@ def test_explode(session):
...
     )
 
+    # with input as array construct
+    Utils.check_answer(
+        session.range(1).select(explode(array_construct(lit(1), lit(2), lit(3)))),
+        [Row(VALUE="1"), Row(VALUE="2"), Row(VALUE="3")],
+    )
+
+    try:
+        context._use_structured_type_semantics = True
```
sfc-gh-jrose: I'd recommend setting this with a patch context. I've seen other tests fail due to various context flags being set when running tests in parallel.
```suggestion
        with patch.object(context, "_use_structured_type_semantics", True):
```

src/snowflake/snowpark/table_function.py
```diff
@@ -321,9 +319,16 @@ def get_column_names_from_plan(plan: SnowflakePlan) -> List[str]:
...
+    plan: SnowflakePlan,
+    select_statement: Optional[SelectStatement],
 ) -> Tuple[List, List]:
-    explode_col_type = plan.output_dict.get(func.col, [None])[0]
+    if select_statement:
+        plan = select_statement.select([func.col._named()]).snowflake_plan
+        explode_col_type = plan.output[0].datatype
+    else:
+        col_name = func.col._expression.name
+        explode_col_type = plan.output_dict.get(col_name, [None])[0]
```
sfc-gh-aling, line 331: previously the key used to retrieve info is of `Column` type and now it's a string, are they equivalent?

I searched the code base and it looks like output_dict is just using strings as keys, does that mean the previous `explode_col_type` could be wrong?

sfc-gh-aalam: previously, `func.col` was stored as string (see line 211-214 in the deleted lines above) which would represent the selected column, but now we store `func.col` as `Column` object. This is done because we can to support more complex cases.

The if branch on line 326 covers the case when sql simplifier is enabled and `select_statement is not None`. In this case, we use the `SelectStatement` object to select only a single column i.e. `func.col`. We get the `SnowflakePlan` for this new `Selectable` and get the output type using `plan.output[0].datatype`.

The else branch on line 329 reverts back to old logic where we first extract the column name from column expression using `func.col._expression.name` and then extract the column type using the plan's output_dict which has string keys.

We have slightly different approaches in both cases.

sfc-gh-aling: I see, thanks for the explanation!

sfc-gh-aling (approved):

PR #3484, opened 2025-06-24
SNOW-2130723 - Set enforce_ordering=True when using DML/DDL changes

sfc-gh-jkew:
This is a behavior change, but when users are reading from parquet
files or another file using snowpark and then converting immediately
to snowpark we previously would print a not-implemented error. Since
the default is for enforce_ordering to be False this would introduce
confusion. In this case we simply switch to enforce_ordering=True
if there are multiple queries in the Snowpark dataframe.

Fixes SNOW-2130723

   - [X] I acknowledge that I have ensured my changes to be thread-safe. Follow the link for more information: [Thread-safe Developer Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#thread-safe-development)

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -1421,7 +1421,7 @@ class DataFrame:
             has_existing_ddl_dml_queries = True
             warning(
                 "enforce_ordering_ddl",
-                "enforce_ordering is disabled when using DML/DDL operations",
+                "enforce_ordering is enabled when using DML/DDL operations regardless of user setting",
                 warning_times=1,
             )
 


src/snowflake/snowpark/dataframe.py
```diff
@@ -1414,8 +1415,11 @@ def to_snowpark_pandas(
                 )
             if columns is not None:
                 ast.columns.extend(columns if isinstance(columns, list) else [columns])
+        has_existing_ddl_dml_queries = False
+        if not enforce_ordering and len(self.queries["queries"]) > 1:
+            has_existing_ddl_dml_queries = True
```
sfc-gh-jrose, line 1421: Maybe emit a warning here that the default behavior is automatically being changed.

sfc-gh-helmeleegy (approved):

PR #3493, opened 2025-06-26
SNOW-2171823: Reuse UDFs/UDTFs that share a common signature

sfc-gh-joshi:

   

   Fixes SNOW-2171823

## Background
While investigating a possible implementation for a customer request to support persistent UDxFs in apply (SNOW-2157873), I decided to check whether a persistent UDxF had already been created via a local cache storing references to Snowpark UDF/UDTF object constructed during the current session. During this process, I noticed the potential for large performance improvements from caching even with the temporary UDF/UDTFs that are created today.

The control flow behind UDF/UDTF creation in Snowpark pandas is a little complicated, but at its core are 2 mechanisms:
- For series-like apply operations (`Series.apply` or `DataFrame.map`), a new UDF is created to apply the function to the single column if the input is a Series, or every column in the DF if the input is a DF.
- For other apply operations (`DataFrame.apply(axis=0)` or `DataFrame.apply(axis=1)`), multiple UDTFs may be created for each column in the frame, with additional pre/post-processing JOIN and PIVOT operations depending on the exact parameters of the operation.

In many cases, this can cause the creation of many redundant UDF/UDTFs, which can cause significant overhead. Consider this very simple operation on a 10x10 dataframe (linked in the parent ticket SNOW-2171823):

```python
import modin.pandas as pd

def f(column):
    return column + 2 


df = pd.DataFrame([[20] * 10] * 10) 
result2 = df.map(f)
```

When run on the `main` branch (07feb45) it issued the following behemoth of query text:

<details>
<summary>queries</summary>
<pre>
INFO:snowflake.snowpark:AST state has not been set explicitly. Defaulting to ast_enabled = True.
DEBUG:snowflake.snowpark:Setting AST state. Current state: ast_enabled = True, state = _AstFlagState.TENTATIVE. Request: source = AstFlagSource.SERVER, enable = False.
INFO:snowflake.snowpark.session:Snowpark Session information: 
"version" : 1.33.0,
"python.version" : 3.9.21,
"python.connector.version" : 3.14.0,
"python.connector.session.id" : 3563546421454907,
"os.name" : Darwin

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8b39-000c-a90a7c43cd83] SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE' IN SESSION
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8ba3-000c-a90a7c440a03] SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE' IN SESSION
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-86d6-000c-a90a7c44433f] create SCOPED TEMPORARY                     stage if not exists "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-882b-000c-a90a7c4435cf] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8b9d-000c-a90a7c4461a3]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-882b-000c-a90a7c4435cf')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8aa6-000c-a90a7c439f43] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_E9QW7E8FEY(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8aa5-000c-a90a7c43ae5b] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8a9f-000c-a90a7c43f93b]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-8aa5-000c-a90a7c43ae5b')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8b39-000c-a90a7c43cd9f] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_LKH1GF0U55(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8ba3-000c-a90a7c440a23] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8a71-000c-a90a7c43e947]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-8ba3-000c-a90a7c440a23')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-86d6-000c-a90a7c444353] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_0W28QRGDRN(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-882b-000c-a90a7c4435eb] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8b9d-000c-a90a7c4461d7]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-882b-000c-a90a7c4435eb')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8aa6-000c-a90a7c439f6b] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_ANY42XXAZU(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8aa5-000c-a90a7c43ae7f] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8a9f-000c-a90a7c43f95f]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-8aa5-000c-a90a7c43ae7f')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8b39-000c-a90a7c43cdc3] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_9MUXPJSDAU(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8ba3-000c-a90a7c440a57] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8a71-000c-a90a7c43e96b]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-8ba3-000c-a90a7c440a57')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-86d6-000c-a90a7c444373] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_2AVCNHHRM9(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-882b-000c-a90a7c44360b] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8b9d-000c-a90a7c4461f3]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-882b-000c-a90a7c44360b')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8aa6-000c-a90a7c439f93] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_8D859TF5JG(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8aa5-000c-a90a7c43ae9f] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8a9f-000c-a90a7c43f977]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-8aa5-000c-a90a7c43ae9f')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8ba3-000c-a90a7c440a6f] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_KPK3UVWB8T(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8a71-000c-a90a7c43e997] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-86d6-000c-a90a7c444397]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-8a71-000c-a90a7c43e997')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-882b-000c-a90a7c44363b] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_US0A3O5DMV(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8b9d-000c-a90a7c44620f] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_12S78G63EB'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8aa6-000c-a90a7c439fab]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d0-0810-8b9d-000c-a90a7c44620f')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8aa5-000c-a90a7c43aec3] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_2JKSLHABKI(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('800595e3030000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d9b04430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f94680f754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681e295294681e2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468247d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468346802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683b754e4e4e74945294682768437d947d9428681a683c682a683c682c7d94682e4e682f4e6830684168314e68324e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468347d9485945294879468475d9468497d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

INFO:snowflake.snowpark.session:Closing session: 3563546421454907
INFO:snowflake.snowpark.session:Canceling all running queries
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d0-0810-8a9f-000c-a90a7c43f9a3] select system$cancel_all_queries(3563546421454907)
INFO:snowflake.snowpark.session:Closed session: 3563546421454907
</pre>
</details>

Running this code on my local machine (slightly abridged for session setup) completes in **9.399s**.

Run on this PR, the same code produces the following queries:

<details>
<summary>
queries
</summary>
<pre>
INFO:snowflake.snowpark:AST state has not been set explicitly. Defaulting to ast_enabled = True.
DEBUG:snowflake.snowpark:Setting AST state. Current state: ast_enabled = True, state = _AstFlagState.TENTATIVE. Request: source = AstFlagSource.SERVER, enable = False.
INFO:snowflake.snowpark.session:Snowpark Session information: 
"version" : 1.33.0,
"python.version" : 3.9.21,
"python.connector.version" : 3.14.0,
"python.connector.session.id" : 3563546421688247,
"os.name" : Darwin

DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d5-0810-8b39-000c-a90a7c44ba6f] SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE' IN SESSION
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d5-0810-8aa6-000c-a90a7c448afb] SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE' IN SESSION
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d5-0810-8a9f-000c-a90a7c44f62b] create SCOPED TEMPORARY                     stage if not exists "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_7AEDP16D89
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d5-0810-86d6-000c-a90a7c444f47] ls '@"TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_STAGE_7AEDP16D89'
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d5-0810-8aa5-000c-a90a7c44ab3b]  SELECT "name" FROM ( SELECT  *  FROM  TABLE ( RESULT_SCAN('01bd46d5-0810-86d6-000c-a90a7c444f47')))
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d5-0810-8ba3-000c-a90a7c45073b] 
CREATE
TEMPORARY  FUNCTION  "TESTDB_SNOWPANDAS"."PUBLIC".SNOWPARK_TEMP_FUNCTION_KB48E646FQ(arg1 BIGINT)

RETURNS BIGINT
LANGUAGE PYTHON 
VOLATILE
RUNTIME_VERSION=3.9

PACKAGES=('cloudpickle==2.2.1','pandas')

HANDLER='compute'

AS $$
import pickle

func = pickle.loads(bytes.fromhex('8005953d040000000000008c17636c6f75647069636b6c652e636c6f75647069636b6c65948c0e5f6d616b655f66756e6374696f6e9493942868008c0d5f6275696c74696e5f747970659493948c08436f6465547970659485945294284b014b004b004b014b044b1343167c006a00880166016401880069018802a4018e015300944e8c04617267739486948c056170706c799485948c01789485948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e7079948c0a6170706c795f66756e63944d4f05430200039468098c0466756e63948c066b776172677394879429749452947d94288c0b5f5f7061636b6167655f5f948c29736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c948c085f5f6e616d655f5f948c35736e6f77666c616b652e736e6f777061726b2e6d6f64696e2e706c7567696e2e5f696e7465726e616c2e6170706c795f7574696c73948c085f5f66696c655f5f948c592f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f7372632f736e6f77666c616b652f736e6f777061726b2f6d6f64696e2f706c7567696e2f5f696e7465726e616c2f6170706c795f7574696c732e707994754e4e68008c105f6d616b655f656d7074795f63656c6c949394295294681f295294681f2952948794749452948c1c636c6f75647069636b6c652e636c6f75647069636b6c655f66617374948c125f66756e6374696f6e5f736574737461746594939468257d947d9428681a68108c0c5f5f7175616c6e616d655f5f948c2f6372656174655f7564665f666f725f7365726965735f6170706c792e3c6c6f63616c733e2e6170706c795f66756e63948c0f5f5f616e6e6f746174696f6e735f5f947d948c0e5f5f6b7764656661756c74735f5f944e8c0c5f5f64656661756c74735f5f944e8c0a5f5f6d6f64756c655f5f94681b8c075f5f646f635f5f944e8c0b5f5f636c6f737572655f5f9468008c0a5f6d616b655f63656c6c949394298594529468356802286807284b014b004b004b014b024b4343087c00640117005300944e4b028694298c06636f6c756d6e9485948c412f55736572732f6a6f7368692f636f64652f736e6f7770616e6461732f2e2e2f736e6f7770616e6461735f64656d6f732f6170706c795f706572736973742e7079948c0166944b0343020001942929749452947d942868184e681a8c085f5f6d61696e5f5f94681c683c754e4e4e74945294682868447d947d9428681a683d682b683d682d7d94682f4e68304e6831684268324e68334e8c175f636c6f75647069636b6c655f7375626d6f64756c6573945d948c0b5f5f676c6f62616c735f5f947d947586948652308594529468357d9485945294879468485d94684a7d947586948652302e'))
# The following comment contains the source code generated by snowpark-python for explanatory purposes.
# from __main__ import f as func
# args  # variable of type <class 'tuple'>
# kwargs  # variable of type <class 'dict'>
# def apply_func(x):  # type: ignore[no-untyped-def] # pragma: no cover
#     # TODO SNOW-1874779: Add verification here to ensure inferred type matches
#     #  actual type.
#     return x.apply(func, args=args, **kwargs)
#
# func = apply_func

from threading import RLock

lock = RLock()

class InvokedFlag:
    def __init__(self):
        self.invoked = False

def lock_function_once(f, flag):
    def wrapper(*args, **kwargs):
        if not flag.invoked:
            with lock:
                if not flag.invoked:
                    result = f(*args, **kwargs)
                    flag.invoked = True
                    return result
                return f(*args, **kwargs)
        return f(*args, **kwargs)
    return wrapper

invoked = InvokedFlag()

def compute(df):
    return lock_function_once(func, invoked)(*[df[idx] for idx in range(df.shape[1])])

import pandas

compute._sf_vectorized_input = pandas.DataFrame
$$

INFO:snowflake.snowpark.session:Closing session: 3563546421688247
INFO:snowflake.snowpark.session:Canceling all running queries
DEBUG:snowflake.snowpark._internal.server_connection:Execute query [queryID: 01bd46d5-0810-882b-000c-a90a7c45432f] select system$cancel_all_queries(3563546421688247)
INFO:snowflake.snowpark.session:Closed session: 3563546421688247
</pre>
</details>

This completes in just **3.265s**, almost a **3x** improvement in runtime for this very simple example code. This improvement occurs for one very simple reason: the new implementation issues only a single UDF creation query, while the `main` branch creates a new UDF for _every column in the frame_. I have not performed benchmarking on other types of `apply`/`map` calls, but this proof of concept, along with the changes to query counts in the test base, should serve as proof enough of the performance gains in this PR.

## Caching Implementation
All UDF/UDTF registration is handled within `src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py`. This PR introduces 2 new classes `UDFCacheKey` and `UDTFCacheKey` that uniquely identify the parameters involved in UDF/UDTF construction, and let us reuse UDF/UDTF objects when an `apply`/`map` call is performed that matches the parameters from a previous operation. To avoid possible scoping issues, these maps are also unique to each session.

For UDFs, caching is primarily based on the following (with a few metadata properties elided for simplicity):
- The pickled bytes of the wrapper function submitted to the UDF/UDTF (obtained from `cloudpickle.dumps(func)`),
- the inferred output type of the column,
- and the inferred input type of the column.

~~Note the choice of pointers (obtained via built-in `id`) rather than a hash for many of these properties. The cache key object must be hashable, but many args/kwargs passed to a function are not hashable. As such, we use pointers to guarantee uniqueness. This may create cache misses in certain scenarios that a robust hash function would be able to cover, such as a function called twice with a list argument with the same values but different container object, but I preferred a conservative approach for simplicity and correctness.~~

My initial approach used pointers from functions and args/kwargs (obtained by `id`), but these may not necessarily remain unique if memory is freed, and become unreliable if a property of an argument/kwarg or even the function itself is mutated. When `cloudpickle` serializes a function, it includes the contents of all closure-captured variables, so this new approach is sound in the face of mutation of arguments between apply/map calls. The performance overhead of pickling is negligible compared to that of the apply/map call itself.

UDTFs are similar, with data types as tuples rather than a single item, and some additional information for grouping operations and index column names included in the pickled blob. The presence of index and column labels in UDTF creation means cache hits will be less common than for UDFs, which have less specific parameters, so there is potential for future caching improvements by renaming input/output columns before UDTF operations so they are less dependent on specific column/index label values. This PR remained relatively conservative in determining cache keys for the sake of simplicity.

UDTFs are constructed by 3 separate functions in `apply_utils.py`: `create_udtf_for_apply_axis_1`, `create_udtf_for_groupby_no_pivot`, and `create_udtf_for_groupby_apply`. Because each of these helper functions constructs UDTFs in slightly different ways, each of these code paths uses a separate cache.

## Areas of Improvement
The UDF/UDTF caches are cleared between test cases to ensure the SQL count checker receives correct results.

Relatively few existing test cases explicitly perform the same function call twice, but such behavior is not uncommon in interactive environments where code in a cell may be rerun, or an apply is performed repeatedly in a loop.

Most query count reductions occur in `test_applymap.py`. These improvements occur for the same reason as detailed in the sample code above; a new UDF was previously created for each column in the frame, but caching now allows reuse of UDFs when the input/output types and args/kwargs match.

`test_apply.py` does not have any direct query count improvements, but I added a new Jenkins-only test case to exercise the cache. I added some review comments indicating locations where the query count is reduced compared to running the same test on `main`.


tests/integ/modin/frame/test_apply.py
```diff
@@ -1038,3 +1040,178 @@ def test_udfs_and_udtfs_with_snowpark_object_error_msg():
...
+    # axis=0 application
+    with SqlCounter(**axis_0_no_cache_kwargs):
+        eval_snowpark_pandas_result(
+            snow_df,
+            native_df,
+            lambda df: df.apply(f),
+        )
+    # second application should trigger cache hit, even with explicit axis argument or via transform call
+    # unclear why SQL counter still parses UDTF creations here despite having lower query counts
+    with SqlCounter(query_count=7, udtf_count=3):
```
sfc-gh-joshi, line 1076: 16 queries on main

tests/integ/modin/frame/test_apply.py
```diff
@@ -1038,3 +1040,178 @@ def test_udfs_and_udtfs_with_snowpark_object_error_msg():
...
+        eval_snowpark_pandas_result(
+            snow_df,
+            native_df,
+            lambda df: df.transform(f),
+        )
+    with SqlCounter(query_count=7, udtf_count=3):
```
sfc-gh-joshi, line 1082: 16 queries on main

tests/integ/modin/frame/test_apply.py
```diff
@@ -1038,3 +1040,178 @@ def test_udfs_and_udtfs_with_snowpark_object_error_msg():
...
+        )
+    # application with a different argument value should trigger a new UDTF creation
+    with SqlCounter(**axis_0_no_cache_kwargs):
+        eval_snowpark_pandas_result(
+            snow_df,
+            native_df,
+            lambda df: df.apply(f, default_arg=3),
+        )
+    # application on a new dataframe with the same schema should hit the cache
+    with SqlCounter(query_count=7, udtf_count=3):
```
sfc-gh-joshi, line 1096: 16 queries on main

tests/integ/modin/frame/test_apply.py
```diff
@@ -1038,3 +1040,178 @@ def test_udfs_and_udtfs_with_snowpark_object_error_msg():
...
+        )
+    # calling on axis=1 creates a new UDTF
+    with SqlCounter(query_count=5, udtf_count=1):
+        eval_snowpark_pandas_result(
+            snow_df,
+            native_df,
+            lambda df: df.apply(f, axis=1),
+        )
+    # a second call with axis=1 does hit the cache (not sure why SQL counter registers a udtf creation)
+    with SqlCounter(query_count=2, udtf_count=1):
```
sfc-gh-joshi, line 1109: 5 queries on main

tests/integ/modin/frame/test_apply.py
```diff
@@ -1038,3 +1040,178 @@ def test_udfs_and_udtfs_with_snowpark_object_error_msg():
...
+        high_count_expected=True,
+        high_count_reason="UDTF creation on multiple columns",
+    ):
+        eval_snowpark_pandas_result(
+            snow_df,
+            native_df,
+            lambda df: df.apply(f, axis=0),
+        )
+    # A second call hits the cache.
+    with SqlCounter(query_count=5, udtf_count=2):
```
sfc-gh-joshi, line 1194: 11 queries on main

tests/integ/modin/frame/test_apply.py
```diff
@@ -1038,3 +1040,178 @@ def test_udfs_and_udtfs_with_snowpark_object_error_msg():
...
+        )
+    # The same rules apply with a different axis argument.
+    with SqlCounter(query_count=5, udtf_count=1):
+        eval_snowpark_pandas_result(
+            snow_df,
+            native_df,
+            lambda df: df.apply(f, axis=1),
+        )
+    # A second call hits the cache.
+    with SqlCounter(query_count=2, udtf_count=1):
```
sfc-gh-joshi, line 1208: 5 queries on main

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -130,6 +132,75 @@
     ALL_SNOWFLAKE_CORTEX_FUNCTIONS = tuple()
 
 
+# Reuse UDFs that have already been constructed in each session.
+@dataclass(eq=True, frozen=True)
+class UDFCacheKey:
+    func_id: int  # The result of calling id() on the function passed to apply
+    return_type: DataType  # The inferred return type of the column
```
sfc-gh-jkew, line 145: The function is not immutable, correct, that is why you are also checking arguments and such?

I think I would not use function pointers for this; but I might use a hash of the pickled function itself as the key. That would also allow for server-side caching between sessions if we want to do that in the future.

We can chat about it though.

sfc-gh-jkew (changes requested):

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-joshi: Apparently Python functions are indeed mutable: https://stackoverflow.com/questions/31599376/why-is-code-for-a-functionpython-mutable (example is python 2, but equivalent code works in python 3). It turns out function objects are actually inherently hashable, but are based on the function's `id` and don't check the `__code__` attribute either:
```
def f():
    print("in f")

def g():
    print("in g")

print(hash(f))  # 271948328
f.__code__ = g.__code__
print(hash(f))  # 271948328
```

It seems like the body of a `__code__` attribute is itself immutable, so we could key on that instead (assuming the apply function doesn't rely on other metadata like `func.__name__`).

I think you're right that to truly determine uniqueness, we need to pickle the functions and check the hashes of the pickled blobs. I'm a little wary of potential performance implications, but given that we're doing this for apply (an already-slow operation) it should be fine. Ideally we would be able to cache the client-side pickled content, but the abstraction layer responsible for doing that seems to be buried pretty deep within Snowpark code.

sfc-gh-helmeleegy: I wouldn't worry too much about the performance overhead here because this is not dependent on the data size. The real slowness of apply becomes more and more apparent as the data size increases. This is not the case here.

sfc-gh-joshi: Makes sense.

That said, mutation in function arguments is definitely more likely than mutation of the function itself, so perhaps I should be pickling all args/kwargs for the cache key as well. I'm a little worried about doing so (rather than using `id`), since apparently you can't pickle Snowpark pandas objects:
```
import modin.pandas as pd
import cloudpickle
cloudpickle.dumps(pd.DataFrame())  # NotImplementedError: Snowpark pandas does not yet support the method DataFrame.__reduce__
```

There are likely other edge cases where an arg/kwarg of the function cannot be pickled. I suspect such cases would error out when Snowpark internals trying to pickle the function to upload, but I'm still concerned that this caching mechanism would break some other edge cases.

WDYT @sfc-gh-helmeleegy @sfc-gh-jkew ?

sfc-gh-joshi: I decided to circumvent this problem entirely by pickling the method used in the apply call rather than the input function alone. This simplifies logic a fair bit because it naturally captures all closure variables used by the functions, and I added a couple jenkins-only tests to make sure the pickled output changes when a mutable value is changed.

sfc-gh-helmeleegy: This makes sense to me!

tests/integ/modin/frame/test_apply.py
sfc-gh-jkew: wow

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -130,6 +135,62 @@
...
+
+
+# Reuse UDTFs that have already been constructed in each session
+@dataclass(eq=True, frozen=True)
+class UDTFCacheKey:
+    # The result of calling pickle_function(func), where func is the end_partition method of the
+    # wrapper class is used in the UDTF. We cannot pickle the wrapper class directly because pickle
+    # will serialize each class differently even if they have the same body and closure variables
+    # because they are technically different classes.
+    func_pickle_blob: bytes
```
sfc-gh-jkew, line 163: I think this will help a lot, it would be interesting if we could have this blob be passed directly into the bowels of snowpark so it does not need to be pickled again but that is an optimization. This is already great work.

sfc-gh-jkew (approved):
Awesome

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-joshi: Worth noting that this blob is actually not the same as the content that Snowpark pickles. Snowpark internally pickles the entire `ApplyFunc` class, but we only pickle `ApplyFunc.end_partition` because pickling a re-defined class will give different bytes even if the contents are technically the same. Example:
```
import cloudpickle

def f():
    class A:
        def f():
            pass
    return cloudpickle.dumps(A)

f() == f() # returns False
```
This makes sense, since from Python's POV each class constructed in this way is unique, but it will prevent us from easily reusing the pickled contents for the foreseeable future.

PR #3498, opened 2025-06-27
SNOW-2173644, SNOW-2157873: Fix __array_function__ hybrid switching and native Series constructor switching bugs

sfc-gh-joshi:

   

   Fixes SNOW-2173644 and SNOW-2157873

SNOW-2157873 occurs because upstream modin does not implement `__array_function__`, instead converting to numpy ndarrays via `__array__` when a numpy function is called on it. The presence of the extension wrapper for `__array_function__` introduced by Snowpark pandas confuses numpy dispatch, causing unexpected AttributeErrors. This is fixed upstream with https://github.com/modin-project/modin/pull/7617, and will presumably become available in the next modin release. On the Snowpark side, this PR adds relevant tests, and adds a version-guarded flag to remove the extension function and push it down to the query compiler.

SNOW-2173644 occurs in specific circumstances when determining switching conditions for the DataFrame constructor. Series objects are treated as dict-like, but `Series.values` is a property rather than a function. We thus skip over native_pd.Series objects in the dict-like check in move_to_me_cost.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
@@ -63,7 +63,10 @@ from pandas.util._validators import (
     validate_percentile,
 )
 
-from snowflake.snowpark.modin.plugin._internal.utils import MODIN_IS_AT_LEAST_0_33_0
+from snowflake.snowpark.modin.plugin._internal.utils import (
+    MODIN_IS_AT_LEAST_0_33_0,
+    MODIN_IS_AT_LEAST_0_34_0,
+)
 from snowflake.snowpark.modin.plugin._typing import ListLike
 from snowflake.snowpark.modin.plugin.extensions.utils import (
     ensure_index,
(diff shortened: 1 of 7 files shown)


src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -2372,45 +2375,48 @@ def f(x):
 
 
 # Snowpark pandas has custom dispatch logic for ufuncs, while modin defaults to pandas.
-@register_base_override("__array_ufunc__")
-def __array_ufunc__(self, ufunc: np.ufunc, method: str, *inputs, **kwargs):
-    """
-    Apply the `ufunc` to the `BasePandasDataset`.
+# TODO delete this method once 0.33.x is no longer supported
+if MODIN_IS_AT_LEAST_0_34_0:
```
sfc-gh-mvashishtha: Why is this condition `if MODIN_IS_AT_LEAST_0_34_0`, while the other is `if not MODIN_IS_AT_LEAST_0_34_0`?

sfc-gh-joshi: This one should also be flipped, thanks for the catch.

sfc-gh-helmeleegy (approved):

PR #3501, opened 2025-06-30
SNOW-2177453 - Fix issue with df.melt when there are duplicate internal column names

sfc-gh-jkew:
SNOW-2177453 - Fix issue with df.melt when there are duplicate internal column names
    
The issue here occurs when you have performed a merge on a data frame prior to
using melt. Internally the columns in the Dataframe include both the original
column names and the new ones from the merge, which happen to contain a four
digit random suffix.
    
Normally the Dataframe can map the currently projected pandas labels to the correct
internal snowflake columns, but melt is a very violent operation. We mangle the names
of the columns during the unpivot operation and then remap them back later using the
"suffix_to_unpivot_map" dict which is passed to the snowpark unpivot call.
    
 In this case the bug is that we were setting the value of this map to an intermediate
 column name rather than the quoted, external pandas name.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/unpivot_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/unpivot_utils.py
@@ -704,6 +704,7 @@ def _simple_unpivot(
 
     # create the initial set of columns to be retained as identifiers and those
     # which will be unpivoted. Collect data type information.
+    unpivot_pandas_columns = []
     unpivot_quoted_columns = []
     unpivot_quoted_column_types = []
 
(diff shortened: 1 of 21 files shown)


src/snowflake/snowpark/modin/plugin/_internal/unpivot_utils.py
```diff
@@ -767,7 +769,9 @@ def _simple_unpivot(
         )
         renamed_quoted_unpivot_cols.append(renamed_quoted_unpivot_col)
         # create the column name mapper which is passed to unpivot
-        suffix_to_unpivot_map[renamed_quoted_unpivot_col] = c
+        suffix_to_unpivot_map[renamed_quoted_unpivot_col] = (
+            '"' + unpivot_pandas_columns[i] + '"'
```
sfc-gh-joshi: consider using `snowflake.snowpark._internal.analyzer.quote_name_without_upper_casing` instead

sfc-gh-joshi (approved):

src/snowflake/snowpark/modin/plugin/_internal/unpivot_utils.py
sfc-gh-jkew: done

sfc-gh-helmeleegy (approved):

PR #3511, opened 2025-07-03
NOSNOW - Update precommit to trim the PR test matrix

sfc-gh-jkew:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -134,6 +134,12 @@ jobs:
           - os: windows-latest-64-cores
             python-version: "3.11"
             cloud-provider: azure
+          - os: windows-latest-64-cores
+            python-version: "3.11"
+            cloud-provider: gcp
+          - os: windows-latest-64-cores
+            python-version: "3.12"
+            cloud-provider: gcp
           # limit python 3.12 to one os per csp
           - os: macos-latest
             python-version: "3.12"
@@ -240,7 +246,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12" ]
+        python-version: ["3.9", "3.12" ]
         cloud-provider: [aws]
         exclude:
           # only run each version on one os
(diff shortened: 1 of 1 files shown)


sfc-gh-jrose (approved):

.github/workflows/precommit.yml
```diff
@@ -94,65 +94,36 @@ jobs:
...
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.12"] # only test oldest and latest versions
         cloud-provider: [aws, gcp, azure]
         exclude:
-          # only run macos with aws py3.9 for doctest
-          - os: macos-latest
-            python-version: "3.10"
-          - os: macos-latest
-            python-version: "3.11"
+          # only run py3.9 tests on ubuntu+aws to isolate 3.9 failures
```
sfc-gh-joshi, line 99: Since the python version/CSP matrix is now relatively small, should we use an `include` key instead of `exclude` so this file is easier to read? We should also trim the os listings in the matrix if we're going to only run tests on ubuntu.

[example](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations)

.github/workflows/precommit.yml
```diff
@@ -240,29 +211,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12" ]
+        python-version: ["3.9", "3.12" ]
```
sfc-gh-joshi: As above--can we trim the os matrix, and use `include` instead of `exclude`?

sfc-gh-joshi (changes requested):

.github/workflows/precommit.yml
sfc-gh-jkew: Yeah. that's a good idea. getting all the exclusions in place is tedious.

sfc-gh-helmeleegy (approved):

.github/workflows/precommit.yml
sfc-gh-jkew: ok. Take a look if you have a chance. I switched to using include, and it looks like all the expected tests are triggered.

.github/workflows/precommit.yml
```diff
@@ -239,31 +204,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12" ]
+        os: [ubuntu-latest]
+        python-version: ["3.12" ]
```
sfc-gh-joshi:
```suggestion
        python-version: ["3.12"]
```

sfc-gh-joshi (approved):
Thanks!

PR #3531, opened 2025-07-09
Adding enriched error message for table already exists

sfc-gh-daviwang:

   Fixes SNOW-2151536

 We add an enriched error message for SQL compilation errors of the form "SQL compilation error:
Object 'DUPLICATE_TABLE' already exists." We return the place in the code where the table is first referenced. As an example, for the code snippet: 
```
df = session.create_dataframe([[1, "a"], [2, "b"], [3, "c"]], schema=["id", "name"])

table_name = "DUPLICATE_TABLE"

df1.write.save_as_table(table_name, mode="overwrite")
df2.write.save_as_table(table_name, mode="errorifexists")
```
we have the debug context:
--- Additional Debug Information ---

Table 'DUPLICATE_TABLE' was first referenced at /Users/daviwang/snowpark-python/test_table.py: line 27.


tests/integ/scala/test_object_already_exists.py
```diff
@@ -0,0 +1,74 @@
...
+    """Setup fixture to enable AST for all tests in this module."""
+    original = session.ast_enabled
+    set_ast_state(AstFlagSource.TEST, True)
+    yield
+    set_ast_state(AstFlagSource.TEST, original)
+
+
+def test_existing_table_with_save_as_table(session):
+    table_name = "DUPLICATE_TABLE"
+    df = session.create_dataframe([{"a": 1, "b": 2}])
```
sfc-gh-aalam: generate table name using random table name util to avoid conflits and flakiness here and other tests.

tests/integ/scala/test_object_already_exists.py
```diff
@@ -0,0 +1,74 @@
...
+    df.write.save_as_table(table_name, mode="overwrite")
+
+    try:
+        with pytest.raises(SnowparkSQLException) as ex:
+            df.write.save_as_table(table_name)
+
+        assert f"Table '{table_name}' was first referenced" in str(
+            ex.value.debug_context
+        )
```
sfc-gh-aalam: Is this always enabled?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -188,13 +189,20 @@ def wrap(*args, **kwargs):
                             df_ast_id = arg.df_ast_ids[-1]
                             stmt_cache = arg.session._ast_batch._bind_stmt_cache
                             break
+                        # If we can't find a plan with df_ast_ids, we'll take any SnowflakePlan for
+                        # getting existing_object_context
+                        if isinstance(arg, SnowflakePlan):
+                            top_plan = arg
```
sfc-gh-aalam: I think we can re-write this section as
```
for arg in args:
  if isinstance(arg, SnowflakePlan):
    top_plan = arg
    break
if top_plan is not None and top_plan.df_ast_ids:
  df_ast_id = ...
  stmt_cache = ...
```

tests/integ/scala/test_object_already_exists.py
sfc-gh-daviwang: it is dependent on ast (which i enable in the test suite), and also the enable_trace_sql_errors_to_dataframe flag in context

tests/integ/scala/test_python_source_error.py
sfc-gh-daviwang, line 1: moved the tests in this file to tests/integ/test_trace_sql_errors_to_df.py

src/snowflake/snowpark/_internal/debug_utils.py
```diff
@@ -267,3 +267,126 @@ def get_python_source_from_sql_error(top_plan: "SnowflakePlan", error_msg: str)
...
+        # target_object could be "DB.SCHEMA.TABLE" or just "TABLE"
+        table_name_parts = target_upper.split(".")
+        table_name = table_name_parts[-1]  # Get the last part (table name)
+
+        # Pattern to match SQL identifiers that can be quoted or unquoted
+        # Matches: identifier, "identifier", 'identifier'
+        identifier_pattern = r'(?:["\']?[^"\'\s.]+["\']?)'
+
+        create_patterns = [
+            # Simple table name: CREATE TABLE table_name
```
sfc-gh-jkew, line 363: I feel like there should be a better way of doing this than parsing our own sql... something like:

1. The sql executed had > 1 queries ( this is often how we check for DML changes today ) 
2. An error occurred
3. The sql state is 42710 (duplicate object)
4. Then ideally we should have been able to extract the table name from the original call rather than the generated SQL.

sfc-gh-daviwang: This code isn't being used to determine which sql query is throwing the error. We do that in lines 285-290, where we extract the name of the table from the error for the original call like you describe. This code determines where the duplicate table was first referenced/create, and as a result we might need to account for a lot of different cases

sfc-gh-jkew: Ahh of course. I guess we do not have a record of that right now.

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -219,6 +223,14 @@ def wrap(*args, **kwargs):
                             missing_object_context = get_missing_object_context(
                                 top_plan, e.msg
                             )
+                        if (
+                            ("already exists" in e.msg)
```
sfc-gh-aling, line 227: I'm just curious that
1. would creating any kind of duplicate object (table, view, function, database etc.) trigger "already exists" error?
2. your implementation seems to be error message based, does it mean other apis (like session.sql, session.create_dynamic table etc.) can take free rides?

e.g. 
```python
session.sql("create schema abc").collect()
session.sql("create schema abc").collect()
```

does your code also help with these cases?

sfc-gh-daviwang: 1. My PR covers tables and views b/c that's what the JIRA ticket outlined, but I can check if I can cover other duplicate objects. 
2. Yeah, it should work for all those cases. I check the various apis in my integration tests.

sfc-gh-daviwang: 1. My PR covers tables and views b/c that's what the JIRA ticket outlined, but I can check if I can cover other duplicate objects. Specifically for the example you provide, I  already cover it b/c I pattern match for create statements in session.sql calls.
2. Yeah, it should work for all those cases. I check the various apis in my integration tests.

sfc-gh-aling: this is really helpful, thanks!

PR #3538, opened 2025-07-11
SNOW-2196517: Merge io tests into frame and series tests for to_view/to_dynamic_table/to_iceberg

sfc-gh-helmeleegy:

   

   Fixes SNOW-2196517

   Merge io tests into frame and series tests for to_view/to_dynamic_table/to_iceberg.


tests/integ/modin/frame/test_to_dynamic_table.py
```diff
@@ -14,7 +14,7 @@
 from tests.utils import Utils
 
 
-@sql_count_checker(query_count=7)
+@sql_count_checker(query_count=5)
```
sfc-gh-joshi, line 30: why did this go down?

sfc-gh-joshi (approved):
One question: should we consolidate these all into a single test file, or is there a reason to continue keeping these split across frame/test_to_*, io/test_to_*, and series/test_to_*?

tests/integ/modin/frame/test_to_dynamic_table.py
sfc-gh-helmeleegy: Because we removed the parentheses which resulted in treating the passed query as a table (we have an optimization for treating `select * from T` queries as simply passing the table `T`). In this case, one less temp table gets created.

tests/integ/modin/frame/test_to_dynamic_table.py
sfc-gh-mvashishtha, line 1: should we add some tests for a different `method_type` in this file?

tests/integ/modin/frame/test_to_dynamic_table.py
sfc-gh-joshi: Interesting. Do you know why we had the parentheses there in the first place?

sfc-gh-helmeleegy: I think the example was copied from another test where we wanted to test the query vs table behavior.

sfc-gh-mvashishtha (commented):
rather than having the string parameter `method_type`, can we have a fixture called e.g.  `to_dynamic_table` that will provide the two callables to test? It would look like


```python
@pytest.fixture(params=[
	param(
		lambda obj, *args, **kwargs: obj.to_dynamic_table(*args, **kwargs)
		id="method",
    ),
   param(
		pd.to_dynamic_table,
        id="function"
    ),]
)
def to_dynamic_table(request):
	return request.param
```

tests/integ/modin/frame/test_to_dynamic_table.py
sfc-gh-helmeleegy: Yeah, updated now.

sfc-gh-helmeleegy:
> rather than having the string parameter `method_type`, can we have a fixture called e.g. `to_dynamic_table` that will provide the two callables to test? It would look like
> 
> ```python
> @pytest.fixture(params=[
> 	param(
> 		lambda obj, *args, **kwargs: obj.to_dynamic_table(*args, **kwargs)
> 		id="method",
>     ),
>    param(
> 		pd.to_dynamic_table,
>         id="function"
>     ),]
> )
> def to_dynamic_table(request):
> 	return request.param
> ```

Good idea. Done. PTAL.

PR #3541, opened 2025-07-11
SNOW-2201361:fix dbapi does not work in python stored proc

sfc-gh-yuwang:

   

   Fixes SNOW-2201361


src/snowflake/snowpark/_internal/data_source/utils.py
```diff
@@ -326,6 +327,8 @@ def worker_process(
                 stop_event,
             )
         except queue.Empty:
+            # indicate whether a process is exit gracefully
+            process_or_thread_error_indicator.put((os.getpid(),))
```
sfc-gh-aling: wouldn't getpid return the same value in the multi thread approach?

src/snowflake/snowpark/_internal/data_source/utils.py
```diff
@@ -309,6 +309,7 @@ def _retry_run(func: Callable, *args, **kwargs) -> Any:
 def worker_process(
     partition_queue: Union[mp.Queue, queue.Queue],
     parquet_queue: Union[mp.Queue, queue.Queue],
+    process_or_thread_error_indicator: Union[mp.Queue, queue.Queue],
```
sfc-gh-aling, line 312: do we need another queue, can we reuse the existing parquet_queue, using another constant to signal that the process has exit because of running out of tasks?

src/snowflake/snowpark/_internal/data_source/utils.py
```diff
@@ -406,6 +412,10 @@ def process_parquet_queue_with_threads(
         while len(completed_partitions) < total_partitions or thread_futures:
             # Process any completed futures and handle errors
             process_completed_futures(thread_futures)
+            while not process_or_thread_error_indicator.empty():
+                gracefully_exited_processes.add(
+                    process_or_thread_error_indicator.get(block=False)[0]
+                )
```
sfc-gh-aling: if we use the same queue, we don't need the while here, we can handle in the following `elif parquet_id = ...` logic

src/snowflake/snowpark/_internal/data_source/utils.py
```diff
@@ -448,7 +458,10 @@ def process_parquet_queue_with_threads(
                 if fetch_with_process:
                     # Check if any processes have failed
                     for i, process in enumerate(workers):
-                        if not process.is_alive() and process.exitcode != 0:
+                        if (
+                            not process.is_alive()
+                            and process.pid not in gracefully_exited_processes
+                        ):
```
sfc-gh-aling: there is another exitcode usage on line 487, you can do a global search to see if there're any other places

src/snowflake/snowpark/_internal/data_source/utils.py
```diff
@@ -448,7 +458,10 @@ def process_parquet_queue_with_threads(
                 if fetch_with_process:
                     # Check if any processes have failed
```
sfc-gh-aling, line 467: I think we need another check here to avoid inf loop:

if all thread futures re done, all processes are not alive, but len(completed_partitions) < total_partitions still holds, then it means some process has silently failed. we need to raise the error out

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1471,6 +1471,7 @@ def create_oracledb_connection():
             # Determine the number of processes or threads to use
             max_workers = max_workers or os.cpu_count()
             queue_class = mp.Queue if fetch_with_process else queue.Queue
+            process_or_thread_error_indicator = queue_class()
```
sfc-gh-aling, line 1494: if reusing the q, we don't need to touch this file

src/snowflake/snowpark/_internal/data_source/utils.py
sfc-gh-yuwang: i modified the script to reuse existing queue, I also apply this logic only when using multi process

src/snowflake/snowpark/_internal/data_source/utils.py
```diff
@@ -326,6 +327,9 @@ def worker_process(
                 stop_event,
             )
         except queue.Empty:
+            # indicate whether a process is exit gracefully, not needed in multithread approach
+            if isinstance(parquet_queue, mp.queues.Queue):
```
sfc-gh-aling: you can use `stop_event` which is only set when using multihreading
```suggestion
            if stop_event is None:  # this is process based ingestion
```

src/snowflake/snowpark/_internal/data_source/utils.py
```diff
@@ -469,11 +480,18 @@ def process_parquet_queue_with_threads(
 
     if fetch_with_process:
         # Wait for all processes to complete
-        for idx, process in enumerate(workers):
+        for process in workers:
             process.join()
-            if process.exitcode != 0:
+        # empty parquet queue to get all signals after each process ends
+        while not parquet_queue.empty():
```
sfc-gh-aling: queue.empty is unreliable, https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.empty

if you want to get all elements out of queue, use the 
```python
while True:
    try:
        q.get_nowait or q.get(Flase)
    except queue.Empty:
        break
```

src/snowflake/snowpark/_internal/data_source/utils.py
```diff
@@ -469,11 +480,18 @@ def process_parquet_queue_with_threads(
...
+            parquet_id, parquet_buffer = parquet_queue.get(block=False)
+            if parquet_id == PROCESS_EXIT_GRACEFULLY_SIGNAL:
```
sfc-gh-aling: assert parquet_id == PROCESS_EXIT_GRACEFULLY_SIGNAL
gracefully_exited_processes.add(parquet_buffer)

sfc-gh-aling (approved):
LGTM, please address my minor comment before merging

sfc-gh-jdu (approved):

PR #3547, opened 2025-07-14
SNOW-2185699: Support filtering after grouping by and aggregation 

sfc-gh-jdu:

   

   Fixes SNOW-2185699

   Adding an internal api `_filter_group_by_rows` for SAS workaround


tests/integ/test_df_aggregate.py
```diff
@@ -634,3 +634,16 @@ def test_agg_on_empty_df(session):
     Utils.check_answer(
         agged_no_group_by, [Row(None, 0, 0, "", None, None, None, None, None, None)]
     )
+
+
+def test_agg_having(session):
+    df = session.create_dataframe([(1, 2, 3), (3, 2, 1), (3, 2, 1)], ["a", "b", "c"])
+    df1 = df.group_by("a").agg(sum_("c"))._filter_group_by_rows(count("a") > 1)
+    assert "HAVING" in df1.queries["queries"][0]
+    Utils.check_answer(df1, [Row(3, 2)])
```
sfc-gh-aalam: In terms of nesting, how does this query look like?

sfc-gh-jdu: it's not nested query:
```
SELECT 
    "A", 
    sum("C") AS "SUM(C)"
 FROM (
 SELECT 
    "A", 
    "B", 
    "C"
 FROM (
 SELECT $1 AS "A", $2 AS "B", $3 AS "C" FROM  VALUES (1 :: INT, 2 :: INT, 3 :: INT), (3 :: INT, 2 :: INT, 1 :: INT), (3 :: INT, 2 :: INT, 1 :: INT)
)
)
 GROUP BY 
    "A"
 HAVING (count("A") > 1)
```
we can only call this internal method after RelationalGroupedDataFrame.agg.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1976,6 +1978,32 @@ def filter(
             _ast_stmt=stmt,
         )
 
+    def _filter_group_by_rows(
```
sfc-gh-aling: I assume you will need to change all the corresponding code in sas to make usage of this new api?

I'm wondering if it would be easier to make it part of the existing group by implementation, and decide the behavior based upon the context parameter spark_connect_mode, such way you can minimize the code change in the sas project

sfc-gh-jdu: yeah it's a good point, let me check with SAS team

sfc-gh-jdu: done

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -530,14 +531,17 @@ def project_statement(
 
 
 def filter_statement(
-    condition: str, child: str, child_uuid: Optional[str] = None
+    condition: str, is_having: bool, child: str, child_uuid: Optional[str] = None
```
sfc-gh-aling, line 534: nit: this is just related to coding style, I feel the new parameter could be added to the end of signature with a default value `False`:
`condition: str, child: str, child_uuid: Optional[str] = None, is_having: bool = False`

this helps:
1. code style consistency -- the change you made in unary_plan_node.py is using this pattern
2. reduce some code change in the caller where we expliclity set this value to False. in this PR the help might be little, just 2 lines of code as far as I can see


this comment applies to other places where the new parameter is introduced

sfc-gh-jdu: The reason I added here is I think child and child_uuid seems to be together. Yeah but I can make the change

src/snowflake/snowpark/dataframe.py
```diff
@@ -1963,18 +1964,41 @@ def filter(
...
 
-        if self._select_statement:
+        # In snowpark_connect_compatible mode, we need to handle
+        # the filtering for dataframe after aggregation without nesting using HAVING
+        if (
+            context._is_snowpark_connect_compatible_mode
+            and self._is_grouped_by_and_aggregated
+        ):
+            having_plan = Filter(filter_col_expr, self._plan, is_having=True)
+            if self._select_statement:
```
sfc-gh-aalam, line 1975: shouldn't you be using `Having` node here?

src/snowflake/snowpark/dataframe.py
```diff
@@ -2104,16 +2128,40 @@ def sort(
...
 
-        df = (
-            self._with_plan(self._select_statement.sort(sort_exprs))
-            if self._select_statement
-            else self._with_plan(Sort(sort_exprs, self._plan))
-        )
+        # In snowpark_connect_compatible mode, we need to handle
+        # the sorting for dataframe after aggregation without nesting
+        if (
+            context._is_snowpark_connect_compatible_mode
```
sfc-gh-aalam, line 2135: why only enable it in snowpark_connect_compatible mode? Does this addition break anything? If not, let's make it the default behavior and call out it as an improvement

src/snowflake/snowpark/dataframe.py
sfc-gh-jdu: Actually I decided to add an attribute in `Filter` instead of creating `Having`

src/snowflake/snowpark/dataframe.py
sfc-gh-jdu: Yeah it will be a breaking change. Consider the following code:
```
df = session.create_dataframe([(1, 2, 3), (3, 2, 1)], ["a", "b", "c"])
df.group_by("a").agg(sum("c")).filter(col("a") > 0).show()
```
which is currently working in Snowpark, though filters rows from the entire dataframe instead of individual group.

src/snowflake/snowpark/dataframe.py
sfc-gh-aalam: I see, then you should remove `Having` from unary_plan_node.py?

src/snowflake/snowpark/dataframe.py
sfc-gh-aalam: sorry, I don't understand this example. Shouldn't they be equivalent. I modified filter in your example.

```
df = session.create_dataframe([(1, 2, 3), (3, 2, 1)], ["a", "b", "c"])
df.group_by("a").agg(sum("c")).show()
```

| A | SUM(C)|
|---|---|
|3|1|
|1|3|

```
df = session.create_dataframe([(1, 2, 3), (3, 2, 1)], ["a", "b", "c"])
df.group_by("a").agg(sum("c")).filter(col("a") > 1).show()
```
| A | SUM(C)|
|---|---|
|3|1|

If I run the flattened query with HAVING I have
```
SELECT "A",
   sum("C") AS "SUM(C)"
FROM (
      SELECT "A",
         "B",
         "C"
      FROM (
            SELECT $1 AS "A",
               $2 AS "B",
               $3 AS "C"
            FROM
            VALUES (1::INT, 2::INT, 3::INT),
               (3::INT, 2::INT, 1::INT)
         )
   )
GROUP BY "A"
HAVING ("A" > 1);
```

| A | SUM(C)|
|---|---|
|3|1|

sfc-gh-jdu: Yeah you are right, using HAVING should cover all cases WHERE can support.

sfc-gh-aalam (approved):

PR #3533, opened 2025-07-10
[SNOW-2111917]: Add real stage tests for SnowflakeFile

sfc-gh-lju:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-2111917.

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.
   - [x] I acknowledge that I have ensured my changes to be thread-safe. Follow the link for more information: [Thread-safe Developer Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#thread-safe-development)
   - [x] If adding any arguments to public Snowpark APIs or creating new public Snowpark APIs, I acknowledge that I have ensured my changes include AST support. Follow the link for more information: [AST Support Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#ast-abstract-syntax-tree-support-in-snowpark)

3. Please describe how your code solves the related issue.

   Adds in tests for calling SnowflakeFile methods while interacting with real stages.

PR #3571, opened 2025-07-21
[SNOW-2138082]: Add relative path support for SnowflakeFile

sfc-gh-lju:

   

   Fixes SNOW-2138082

Removes enforcement that the path passed to SnowflakeFile has to be an absolute path.

### DIFF
--- a/src/snowflake/snowpark/files.py
+++ b/src/snowflake/snowpark/files.py
@@ -81,6 +81,7 @@ class SnowflakeFile(RawIOBase):
     We provide a local implementation of SnowflakeFile to aid in local testing.
     This currently supports using read APIs on relative paths, mocked stages
     (sessions in local testing mode that aren't connected to a real stage), and Snowflake stages.
+    Scoped and Stage URLs (https://) are not yet supported.
 
     Note:
         1. All of the implementation in this file is for local testing purposes.
@@ -118,6 +119,9 @@ class SnowflakeFile(RawIOBase):
         # Attributes required for local testing functionality
         _DEFAULT_READ_BUFFER_SIZE = 32 * 1024
         self._pos = 0
+        if self._file_location.startswith("https://"):
+            raise ValueError("Scoped and Stage URLs are not yet supported.")
+
         self._is_stage_file = (
             True
             if self._file_location.startswith(tuple(SNOWFLAKE_PATH_PREFIXES))
--- a/src/snowflake/snowpark/files.py
+++ b/src/snowflake/snowpark/files.py
@@ -17,7 +17,6 @@ from io import (
     SEEK_CUR,
 )
 from snowflake.snowpark._internal.utils import (
-    RELATIVE_PATH_PREFIX,
     SNOWFLAKE_PATH_PREFIXES,
 )
 from typing import Sequence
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/files.py
```diff
@@ -17,7 +17,6 @@
     SEEK_CUR,
 )
 from snowflake.snowpark._internal.utils import (
-    RELATIVE_PATH_PREFIX,
     SNOWFLAKE_PATH_PREFIXES,
```
sfc-gh-lschmidt, line 20: Can you add https:// to this so scoped urls and stage urls work too?

tests/unit/test_files.py
```diff
@@ -352,6 +352,18 @@ def sf_open(file_location: str) -> None:
         sf_open(temp_file)
 
 
+@pytest.mark.parametrize(["read_mode", "write_mode"], [("r", "w"), ("rb", "wb")])
+def test_read_relative_path_snowflakefile(read_mode, write_mode, tmp_path):
```
sfc-gh-aissa: Can we add a test with ~ and rel path where the path is just cur dir. ie SnowflakeFile.open("file.txt", ...) and "~/dir/file.txt".

Let's also add an integ test for one of the cases.

src/snowflake/snowpark/files.py
sfc-gh-aissa: Good callout, this would eventually be needed when we add write support, and has use cases today. Things where customers generate a file on Snowflake and get a URL, then want to read the file locally.

Let's add an integ test for this, like the ones in https://github.com/snowflakedb/snowpark-python/pull/3533.

https://docs.snowflake.com/en/sql-reference/functions/build_scoped_file_url and https://docs.snowflake.com/en/sql-reference/functions/build_stage_file_url should help make the test feasible (ie you can use PUT, then generate a URL for the file with this)

sfc-gh-lju: image

I thought in the PRD we decided not to add scoped urls and stage urls.

EDIT: Just saw Alex's comment. I'll add an integ test for this.

sfc-gh-aissa: What I meant in that was more for local files and mocked stages (and for the write logic we didn't implement because scoped URLs are the output of SnowflakeFile.write()). I didn't want to try and support fake stage/scoped URLs for mocked stages. But we definitely should support it for reading a file locally from a real stage, that's a valid path a customer could use today.

tests/unit/test_files.py
sfc-gh-lju: I'll add in the ~ and cur dir rel path test case! I don't think we need an integ test though since this should only change the behavior for local file paths.

sfc-gh-lju: ~ might not be testable as it'll throw errors with the Windows precommit. I'll add in the cur dir rel path test case though.

src/snowflake/snowpark/files.py
sfc-gh-lju: Discussed offline. Adding in a new error message for not supporting scoped and stage urls as a GET query gets sent to GS and GET doesn't support scoped and stage urls.

tests/unit/test_files.py
```diff
@@ -352,6 +352,42 @@ def sf_open(file_location: str) -> None:
         sf_open(temp_file)
 
 
+@pytest.mark.parametrize(
+    ["read_mode", "write_mode", "use_tmp_path"],
+    [("r", "w", True), ("r", "w", False), ("rb", "wb", True), ("rb", "wb", False)],
+)
+def test_read_relative_path_snowflakefile(
```
sfc-gh-aissa, line 359: Did you add the test for ~/path/to/file.txt?

sfc-gh-aissa: Nvm saw comment above, Windows PowerShell supports ~ but default Windows terminal doesn't, TIL(de).

sfc-gh-aissa (approved):

sfc-gh-lschmidt (approved):

PR #3578, opened 2025-07-22
SNOW-2042703 - Upper bound for row estimates when resulting from a cartesian product

sfc-gh-jkew:
SNOW-2042703 - Upper bound for row estimates when resulting from a cartesian product
    
There was a significant performance regression in astronomer_attribution resulting from
the row estimatation. After some investigation the issue seems to occur after we estimate
rows from the cartesian product in an align ( and possibly join too ). When the result
of this estimation is very large set_frame_2d_labels seems to take a very long time as well.
    
We had added a "hack" to _get_rows in SnowflakeQueryCompiler to catch this case and
recalculate rows manually when we absolutely need it, but we apparently use this information
in other places as well.
    
Returning None when this product is very large (>1e34) reduces astronomer attribution from
940s to 9s on my machine, and returning None from the row estimator is a conservative and
reasonable thing to do ( we should never fully count on estimates ).
    
We still need to figure out why this value would cause such a problem, and potentially fix
that code as well. Even with a very high maximum estimated upper bound on rows we should
not cause a performance regression.
    
This should provide some relief from those workloads though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@
 
 ### Snowpark Local Testing Updates
 
-- Added local testing support for reading files with `SnowflakeFile` using local file paths, stage paths (@stage/file_path), and the Snow URL semantic (snow://...).
+- Added local testing support for reading files with `SnowflakeFile` using local file paths, the Snow URL semantic (snow://...), local testing framework stages, and Snowflake stages (@stage/file_path).
 
 #### New Features
 
--- a/src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
@@ -112,7 +112,7 @@ class RowCountEstimator:
                 return None
             # SNOW-2042703 - TODO: Performance regression in cartiesian products with row estimate
             # When the product becomes very large we return None conservatively, as this can have
-            # a negative performance impact on alignment. This is a similiar fix to what was added
+            # a negative performance impact on alignment. This is a similar fix to what was added
             # in SnowflakeQueryCompiler::_get_rows
             cartesian_result = current * other_bound
             if cartesian_result > MAX_ROW_COUNT_FOR_ESTIMATION:
--- a/src/snowflake/snowpark/_internal/data_source/utils.py
+++ b/src/snowflake/snowpark/_internal/data_source/utils.py
@@ -229,30 +229,25 @@ def _upload_and_copy_into_table(
     # Reset buffer position to beginning
     parquet_buffer.seek(0)
 
-    try:
-        # Upload BytesIO directly to stage using put_stream
-        stage_file_path = f"@{snowflake_stage_name}/{parquet_id}"
-        session.file.put_stream(
-            parquet_buffer,
-            stage_file_path,
-            overwrite=True,
-        )
+    # Upload BytesIO directly to stage using put_stream
+    stage_file_path = f"@{snowflake_stage_name}/{parquet_id}"
+    session.file.put_stream(
+        parquet_buffer,
+        stage_file_path,
+        overwrite=True,
+    )
 
-        # Copy into table
-        copy_into_table_query = f"""
-        COPY INTO {snowflake_table_name} FROM @{snowflake_stage_name}/{parquet_id}
-        FILE_FORMAT = (TYPE = PARQUET USE_VECTORIZED_SCANNER=TRUE)
-        MATCH_BY_COLUMN_NAME=CASE_SENSITIVE
-        PURGE=TRUE
-        ON_ERROR={on_error}
-        {DATA_SOURCE_SQL_COMMENT}
-        """
-        session.sql(copy_into_table_query).collect(statement_params=statements_params)
-        logger.debug(f"Successfully uploaded and copied BytesIO parquet: {parquet_id}")
-    finally:
-        # proactively close the buffer to release memory
-        parquet_buffer.close()
-        backpressure_semaphore.release()
+    # Copy into table
+    copy_into_table_query = f"""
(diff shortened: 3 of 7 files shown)


src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -89,7 +89,14 @@ def upper_bound(
             if right_bound is None:
                 # Cannot estimate row count: other DataFrame has no row count information
                 return None
-            return current * right_bound
+            # SNOW-2042703 - TODO: Performance regression in cartiesian products with row estimate
+            # When the product becomes very large we return None conservatively, as this can have
+            # a negative performance impact on alignment. This is a similiar fix to what was added
+            # in SnowflakeQueryCompiler::_get_rows
+            cartesian_result = current * right_bound
+            if cartesian_result > 1e34:
```
sfc-gh-mvashishtha: Is there a convenient way to avoid using this magic number here, or at least use a constant for it? I see it elsewhere in snowflake_query_compiler.py

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -99,7 +106,14 @@ def upper_bound(
             if other_bound is None:
                 # Cannot estimate row count: other DataFrame has no row count information
                 return None
-            return current * other_bound
+            # SNOW-2042703 - TODO: Performance regression in cartiesian products with row estimate
+            # When the product becomes very large we return None conservatively, as this can have
+            # a negative performance impact on alignment. This is a similiar fix to what was added
+            # in SnowflakeQueryCompiler::_get_rows
+            cartesian_result = current * other_bound
+            if cartesian_result > 1e34:
```
sfc-gh-mvashishtha: ditto about the constant

CHANGELOG.md
```diff
@@ -100,7 +103,7 @@ local ingestion. By default, local ingestion uses multithreading. Multiprocessin
 - Show more helpful error messages to Snowflake Notebook users when the `modin` or `pandas` version does not match our requirements.
 - Added a data type guard to the cost functions for hybrid execution mode (PrPr) which checks for data type compatibility.
 - Added automatic switching to the pandas backend in hybrid execution mode (PrPr) for many methods that are not directly implemented in Snowpark pandas.
-- Set the 'type' and other standard fields for Snowpark pandas telemetry.
+- Set the 'type' and other standard fields for modin telemetry.
```
sfc-gh-joshi: is this change intentional?

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
sfc-gh-joshi: 1e34 also seems like a very high threshold. Do you know if it can be reduced?

Did you also get a chance to take a look at the other single-API benchmarks that had regressions around the same time as the astronomer one? https://snowflakecomputing.atlassian.net/browse/SNOW-2042703?focusedCommentId=7472259

tests/unit/modin/test_ordered_dataframe.py
```diff
@@ -137,3 +141,26 @@ def test_ordered_dataframe_missing_row_position_column_negative(
...
+    df1 = mock.create_autospec(OrderedDataFrame)
+    df1.row_count = 1e20
+    df1.row_count_upper_bound = None
+
+    df2 = mock.create_autospec(OrderedDataFrame)
+    df2.row_count = 1e20
+    df2.row_count_upper_bound = None
+
+    # Verify that the RowCountEstimator returns None for a JOIN operation
+    # that would overflow standard float precision.
```
sfc-gh-joshi: Can you clarify what "standard float precision" is here? I find it hard to believe that even a threshold of 1e34 is getting close to causing overflow, given that `sys.float_info.max` is `1.797e308` on my machine. If it's a literal overflow that we're worried about (in the sense of a value being too large to represent in a 64-bit double), then shouldn't we be checking against `float("inf")` instead?

CHANGELOG.md
sfc-gh-jkew: No, this was a bad merge. Thanks

tests/unit/modin/test_ordered_dataframe.py
sfc-gh-jkew: lol. That comment was written by AI. My fault. It isn't an overflow at all.

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
sfc-gh-jkew: You know 1e34 rows is "ten decillion" rows. I think we might get away with setting this to "1e15"

tests/unit/modin/test_ordered_dataframe.py
```diff
@@ -143,18 +143,18 @@ def test_ordered_dataframe_missing_row_position_column_negative(
...
     df1.row_count_upper_bound = None
 
     df2 = mock.create_autospec(OrderedDataFrame)
-    df2.row_count = 1e20
+    df2.row_count = 1e10
     df2.row_count_upper_bound = None
 
     # Verify that the RowCountEstimator returns None for a JOIN operation
-    # that would overflow standard float precision.
+    # which is "large"
```
sfc-gh-joshi: Do we have a positive test verifying that the estimator still produces an estimate for smaller values?

sfc-gh-jkew: Yeah, we test this value in a few other tests.

sfc-gh-helmeleegy:
Curious if we tested this change on other workloads - just to make sure that it is not over-fitted for astronomer?

sfc-gh-helmeleegy:
Otherwise, it looks good to me.

sfc-gh-helmeleegy (approved):

PR #3588, opened 2025-07-24
WIP Performance

sfc-gh-jkew:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
@@ -52,7 +52,7 @@ class RowCountEstimator:
             int: The estimated upper bound on the number of rows in the resulting dataframe
         """
         # Get the current upper bound. If not set, return None
-        current = df.row_count_upper_bound
+        current = df.row_count_upper_bound or df.row_count
         if current is None:
             return None
 
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -783,9 +783,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
                 or ordered_dataframe.row_count_upper_bound
                 > MAX_ROW_COUNT_FOR_ESTIMATION
             ):
-                num_rows = query_compiler.get_axis_len(0)
-            if num_rows is None:
-                return 1000000000
+                return MAX_ROW_COUNT_FOR_ESTIMATION
         else:
             num_rows = query_compiler.get_axis_len(0)
         return num_rows
@@ -799,9 +797,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
             ordered_dataframe.row_count_upper_bound is None
             or ordered_dataframe.row_count_upper_bound > MAX_ROW_COUNT_FOR_ESTIMATION
         ):
-            num_rows = self.get_axis_len(0)
-        if num_rows is None:
-            num_rows = 10_000_000_000
+            return MAX_ROW_COUNT_FOR_ESTIMATION, num_columns
         return num_rows, num_columns
 
     @classmethod
@@ -13616,7 +13612,8 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
             row_count = 0 if len(pandas_frame) == 0 else pandas_frame.iat[0, -1]
             pandas_frame = pandas_frame.iloc[:, :-1]
         col_count = len(pandas_frame.columns)
-
+        # update upper bound
+        frame.ordered_dataframe.row_count_upper_bound = row_count
         return row_count, col_count, pandas_frame
 
     def quantiles_along_axis0(
--- a/tests/integ/modin/hybrid/test_switch_operations.py
+++ b/tests/integ/modin/hybrid/test_switch_operations.py
@@ -12,6 +12,7 @@ import pandas as native_pd
 import numpy as np
 from numpy.testing import assert_array_equal
 from modin.config import context as config_context
+from modin.config import AutoSwitchBackend
 import modin.pandas as pd
 import snowflake.snowpark.modin.plugin  # noqa: F401
 from snowflake.snowpark.modin.plugin._internal.utils import (
@@ -75,12 +76,14 @@ def test_merge(init_transaction_tables, us_holidays_data):
     assert combined.get_backend() == "Snowflake"
 
 
-@sql_count_checker(query_count=7)
+@sql_count_checker(query_count=4)
 def test_filtered_data(init_transaction_tables):
     # When data is filtered, the engine should change when it is sufficiently small.
     df_transactions = pd.read_snowflake("REVENUE_TRANSACTIONS")
     assert df_transactions.get_backend() == "Snowflake"
     # in-place operations that do not change the backend
+    # TODO: the following will result in a cartesian align which will grow the
+    # size of the row estimate
     df_transactions["DATE"] = pd.to_datetime(df_transactions["DATE"])
     assert df_transactions.get_backend() == "Snowflake"
     base_date = pd.Timestamp("2025-06-09").date()
@@ -91,7 +94,8 @@ def test_filtered_data(init_transaction_tables):
     assert df_transactions_filter1.get_backend() == "Snowflake"
     # The smaller dataframe does operations in pandas
     df_transactions_filter1 = df_transactions_filter1.groupby("DATE").sum()["REVENUE"]
-    assert df_transactions_filter1.get_backend() == "Pandas"
+    # We still operate in Snowflake because we cannot properly estimate the rows
+    assert df_transactions_filter1.get_backend() == "Snowflake"
     df_transactions_filter2 = pd.read_snowflake(
         "SELECT * FROM revenue_transactions WHERE Date >= DATEADD( 'days', -7, '2025-06-09' ) and Date < '2025-06-09'"
     )
@@ -106,8 +110,10 @@ def test_filtered_data(init_transaction_tables):
 @sql_count_checker(query_count=8)
 def test_apply(init_transaction_tables, us_holidays_data):
     df_transactions = pd.read_snowflake("REVENUE_TRANSACTIONS")
+    assert df_transactions.get_backend() == "Snowflake"
     df_us_holidays = pd.DataFrame(us_holidays_data, columns=["Holiday", "Date"])
     df_us_holidays["Date"] = pd.to_datetime(df_us_holidays["Date"])
+    assert df_us_holidays.get_backend() == "Pandas"
 
     def forecast_revenue(df, start_date, end_date):
         # Filter data from last year
@@ -136,8 +142,10 @@ def test_apply(init_transaction_tables, us_holidays_data):
 
     start_date = pd.Timestamp("2025-10-01")
     end_date = pd.Timestamp("2025-10-31")
+
+    assert df_transactions._query_compiler._modin_frame.ordered_dataframe.row_count_upper_bound == 10000000
     df_forecast = forecast_revenue(df_transactions, start_date, end_date)
-    assert df_forecast.get_backend() == "Pandas"
+    assert df_forecast.get_backend() == "Snowflake"
 
     def adjust_for_holiday_weekend(row):
         # For national holidays, revenue down 5% since stores are closed.
@@ -197,7 +205,7 @@ def test_groupby_agg_post_op_switch(operation, small_snow_df):
 
 
 @sql_count_checker(query_count=1)
-def test_explain_switch(us_holidays_data):
+def test_explain_switch(init_transaction_tables, us_holidays_data):
     from snowflake.snowpark.modin.plugin._internal.telemetry import (
         clear_hybrid_switch_log,
     )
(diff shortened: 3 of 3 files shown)


src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -93,14 +93,29 @@ def upper_bound(
...
+                if current < right_bound:
+                    return current
+                else:
+                    return right_bound
+            if how == "left" or how == "asof":
+                return current
+            if how == "right":
+                return right_bound
+            if how == "outer":
+                return current + right_bound
```
sfc-gh-mvashishtha: can't left, right, inner, and outer joins increase row count if there are duplicate labels in one of the frames? e.g.

```python
import pandas as pd
pd
pd.DataFrame([1, 1]).merge(pd.DataFrame([1, 1]), how='inner', on=0)
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -75,12 +76,14 @@ def test_merge(init_transaction_tables, us_holidays_data):
...
 
 
-@sql_count_checker(query_count=7)
+@sql_count_checker(query_count=4)
 def test_filtered_data(init_transaction_tables):
     # When data is filtered, the engine should change when it is sufficiently small.
     df_transactions = pd.read_snowflake("REVENUE_TRANSACTIONS")
     assert df_transactions.get_backend() == "Snowflake"
     # in-place operations that do not change the backend
+    # TODO: the following will result in a cartesian align which will grow the
```
sfc-gh-mvashishtha: does it still?

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -103,11 +107,13 @@ def test_filtered_data(init_transaction_tables):
     )
 
 
-@sql_count_checker(query_count=8)
+@sql_count_checker(query_count=5, join_count=1)
```
sfc-gh-mvashishtha: Why is there a new join?

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -296,3 +307,38 @@ def test_unimplemented_autoswitches(class_name, method_name, f_args):
...
+)
+def test_query_count_no_switch(init_transaction_tables):
+    def inner_test(df_in):
+        df_result = df_in[(df_in["REVENUE"] > 123) & (df_in["REVENUE"] < 200)]
+        df_result["REVENUE_DUPE"] = df_result["REVENUE"]
+        df_result["COUNT"] = df_result.groupby("DATE")["REVENUE"].transform("count")
+        return df_result
+
+    df_transactions = pd.read_snowflake("REVENUE_TRANSACTIONS")
+    df_transactions.to_csv("revenue_transactions.csv")
```
sfc-gh-mvashishtha: Why is there a `to_csv` here?

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -296,3 +307,38 @@ def test_unimplemented_autoswitches(class_name, method_name, f_args):
...
+
+
+@sql_count_checker(
+    query_count=13,
+    join_count=6,
+    udtf_count=2,
+    high_count_expected=True,
+    high_count_reason="tests queries across different execution modes",
+)
+def test_query_count_no_switch(init_transaction_tables):
```
sfc-gh-mvashishtha, line 350: Could you add a docstring explaining what this test is checking? I think I understood it when we were debugging the queries two days ago, but now I've forgotten.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -296,3 +307,38 @@ def test_unimplemented_autoswitches(class_name, method_name, f_args):
...
+        return df_result
+
+    df_transactions = pd.read_snowflake("REVENUE_TRANSACTIONS")
+    df_transactions.to_csv("revenue_transactions.csv")
+    inner_test(df_transactions)
+    orig_len = None
+    hybrid_len = None
+    with pd.session.query_history() as query_history_orig:
+        with config_context(AutoSwitchBackend=False, NativePandasMaxRows=10):
+            assert AutoSwitchBackend.get() is False
```
sfc-gh-joshi: Do you need this assert within the context manager?

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
sfc-gh-jkew: good point; it looks like only the align operations might be simplified - everything else can potentially result in a product

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jkew: The behavior changed; but as far as I can tell it's not wrong - essentially the function is now using Snowflake for df_future and daily_avg as the backend.

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jkew: Removed for now; I was paranoid that the configuration was not working as expected when working through this issue.

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jkew: For join yes; but I don't think so for align.

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -51,8 +51,13 @@ def upper_bound(
         Returns:
             int: The estimated upper bound on the number of rows in the resulting dataframe
         """
+
+        if df.row_count and df.row_count_upper_bound < df.row_count:
```
sfc-gh-mvashishtha:
```suggestion
        if df.row_count is not None and df.row_count_upper_bound < df.row_count:
```

Could we clarify whether we are looking for `None` or 0? Both are falsy, but I think `None` makes sense.

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -110,14 +118,17 @@ def upper_bound(
...
-            if cartesian_result > MAX_ROW_COUNT_FOR_ESTIMATION:
-                return None
-            return cartesian_result
+            how = args["how"]
+            if how == "inner":
+                # choose smallest
+                if current < other_bound:
+                    return current
+                else:
+                    return other_bound
```
sfc-gh-mvashishtha, line 127: return min(current, other_bound)

sfc-gh-mvashishtha:
```suggestion
                return min(current, other_bound)
```

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -51,10 +51,15 @@ def upper_bound(
...
             int: The estimated upper bound on the number of rows in the resulting dataframe
         """
-        # Get the current upper bound. If not set, return None
-        current = df.row_count_upper_bound
-        if current is None:
-            return None
+
+        # Get the current upper bound. If current is None that may still be valid for
+        # some operations like limit or agg
+        current = df.row_count_upper_bound or df.row_count
```
sfc-gh-joshi: Is it possible to only read `row_count_upper_bound`, and instead modify the setter of the precise `row_count` field to ensure the row_count_upper_bound is also set when it's updated?

sfc-gh-jkew: You mean set the upper bound here. I think I can do that.

tests/integ/modin/hybrid/test_df_creation_backend.py
```diff
@@ -73,10 +73,10 @@ def test_from_list(us_holidays_data):
 
 
 def test_move_threshold_setting():
-    with config_context(NativePandasMaxRows=10):
+    with config_context(AutoSwitchBackend=True, NativePandasMaxRows=10):
```
sfc-gh-mvashishtha:
```suggestion
    with config_context(NativePandasMaxRows=10):
```

I don't think we need to set `AutoSwitchBackend=True` in the `hybrid` directory.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -14,17 +14,51 @@
...
 from tests.integ.utils.sql_counter import sql_count_checker
 
 
+@sql_count_checker(query_count=0)
+def test_get_rows_with_large_and_none_upper_bound():
+    """
+    Tests that _get_rows returns a large default value when row_count_upper_bound
+    is None or very large.
+    """
+    from modin.core.storage_formats.base.query_compiler import QCCoercionCost
```
sfc-gh-mvashishtha: nit: move to top of file

tests/unit/modin/test_ordered_dataframe.py
```diff
@@ -156,11 +156,43 @@ def test_row_count_estimator_join_big():
...
         is None
     )
 
     assert (
-        RowCountEstimator.upper_bound(df1, DataFrameOperation.ALIGN, {"right": df2})
-        is None
+        RowCountEstimator.upper_bound(
+            df1, DataFrameOperation.ALIGN, {"right": df2, "how": "left"}
+        )
+        == 20000000000
```
sfc-gh-mvashishtha:
```suggestion
        == 2e10
```

nit: for readability since I think this comes from `1e10 + 1e10`

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
sfc-gh-joshi: I forgot that `row_count` is an actual field rather than a property, but I meant doing something like this in the ordered dataframe code:

https://github.com/snowflakedb/snowpark-python/blob/ca7e49a3cadccac8edc885d0b43bb6ce9acd7c02/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py#L239-L242

```
_row_count: Optional[int]
@property
def row_count(self) -> Optional[int]:
    return self._row_count

@row_count.setter
def row_count(self, value):
    self._row_count = value
    self.row_count_upper_bound = value
```

sfc-gh-jkew: done

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13616,7 +13612,8 @@ def build_repr_df(
             row_count = 0 if len(pandas_frame) == 0 else pandas_frame.iat[0, -1]
             pandas_frame = pandas_frame.iloc[:, :-1]
         col_count = len(pandas_frame.columns)
-
+        # update upper bound
+        frame.ordered_dataframe.row_count_upper_bound = row_count
```
sfc-gh-joshi: Will this be used anywhere? I think the reference to `frame` is always discarded after this function.

Also, since writes to `row_count` now affect `row_count_upper_bound` as well, we should just set that directly.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -14,15 +14,49 @@
...
 import modin.pandas as pd
 import snowflake.snowpark.modin.plugin  # noqa: F401
+from snowflake.snowpark.modin.plugin._internal.row_count_estimation import (
+    MAX_ROW_COUNT_FOR_ESTIMATION,
+)
 from snowflake.snowpark.modin.plugin._internal.utils import (
     MODIN_IS_AT_LEAST_0_34_0,
 )
 
+
```
sfc-gh-joshi: nit: delete newline
```suggestion
```

tests/unit/modin/test_ordered_dataframe.py
```diff
@@ -156,11 +156,48 @@ def test_row_count_estimator_join_big():
...
     )
+
+
+def test_row_count_estimator_invariants():
+    # Verify we raise a runtime error if estimate < count
+    df1 = mock.create_autospec(OrderedDataFrame)
+    df1.row_count = 100
+    df1.row_count_upper_bound = 50
+
+    with pytest.raises(RuntimeError):
```
sfc-gh-joshi: Can you add a match clause on the error messages for this test case? RuntimeError and ValueError are both pretty broad, and may accidentally suppress genuine bugs.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: Good point; it is not needed with the property accessors.

tests/unit/modin/test_ordered_dataframe.py
sfc-gh-jkew: Done

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jkew: Done

sfc-gh-mvashishtha (approved):
Thank you @sfc-gh-jkew !

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
 
 #### New Features
 
+#### Improvements
+- Hybrid execution row estimate improvements and a reduction of eager calls.
```
sfc-gh-joshi, line 20:
```suggestion
- Improved row estimates and reduced eager queries in hybrid execution mode (PrPr).
```

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
```diff
@@ -51,10 +51,15 @@ def upper_bound(
         Returns:
             int: The estimated upper bound on the number of rows in the resulting dataframe
         """
-        # Get the current upper bound. If not set, return None
+
+        # Get the current upper bound. If current is None that may still be valid for
+        # some operations like limit or agg
         current = df.row_count_upper_bound
-        if current is None:
-            return None
```
sfc-gh-joshi, line 57: Why not leave this immediate return here? Is there some following branch where we can still return an estimate when `current is None`?

sfc-gh-joshi (approved):

src/snowflake/snowpark/modin/plugin/_internal/row_count_estimation.py
sfc-gh-jkew: Yeah, so it /turns/ out that current can be None and that's totally valid if a LIMIT or agg operation is applied where we know the output size explicitly based on the input arguments.

PR #3591, opened 2025-07-24
SNOW-2166197: add support for jdbc

sfc-gh-yuwang:

   

   Fixes SNOW-2166197

### DIFF
--- a/src/snowflake/snowpark/_internal/data_source/jdbc.py
+++ b/src/snowflake/snowpark/_internal/data_source/jdbc.py
@@ -1,0 +1,117 @@
+#
+# Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
+#
+
+JDBC_UDTF = """
+CREATE OR REPLACE FUNCTION jdbc(query VARCHAR)
+RETURNS TABLE (A VARIANT)
+LANGUAGE JAVA
+RUNTIME_VERSION = '11'
+EXTERNAL_ACCESS_INTEGRATIONS=(snowpark_jdbc_influxdb_test_integration)
+packages=('com.snowflake:snowpark:latest')
+IMPORTS = ('@test_stage/influxdb-java-2.25.jar','@test_stage/amazon-timestream-jdbc-2.0.0.jar','@test_stage/amazon-timestream-jdbc-2.0.0-shaded.jar')
+HANDLER = 'DataLoader'
+as
+$$
+
+import java.sql.*;
+import java.util.stream.Stream;
+
+import com.snowflake.snowpark_java.types.*;
+import com.snowflake.snowpark_java.udtf.*;
+import com.snowflake.snowpark_java.types.SnowflakeSecrets;
+import com.snowflake.snowpark_java.types.UsernamePassword;
+
+class OutputRow {{
+
+    public String A;
+
+    public OutputRow(String A) {{
+        this.A = A;
+    }}
+
+}}
+
+
+
+public class DataLoader{{
+    private final Connection conn;
+
+    private static String escapeJson(String str) {{
+        return str.replace("\\", "\\\\")
+                  .replace("\"", "\\\"")
+                  .replace("\b", "\\b")
+                  .replace("\f", "\\f")
+                  .replace("\n", "\\n")
+                  .replace("\r", "\\r")
+                  .replace("\t", "\\t");
+    }}
+
+
+    public StructType outputSchema() {{
+        return StructType.create(new StructField("A", DataTypes.VariantType));
+    }}
+    public static Class getOutputClass() {{
+      return OutputRow.class;
+    }}
+
+    private static Connection createConnection() {{
+        try {{
+            Class.forName("com.wisecoders.dbschema.influxdb.JdbcDriver");
+            //Class.forName("software.amazon.timestream.jdbc.TimestreamDriver");
+            String url = "";
+            return DriverManager.getConnection(url);
+        }} catch (Exception e) {{
+            throw new RuntimeException("Failed to create JDBC connection: " + e.getMessage());
+        }}
+    }}
+
+    public DataLoader() {{
+        this.conn = createConnection();
+    }}
+
+    public Stream<OutputRow> process(String query) {{
+        try {{
+            ResultSet result = this.conn.createStatement().executeQuery(query);
+            if (result.next()) {{
+                ResultSetMetaData meta = result.getMetaData();
+                int columnCount = meta.getColumnCount();
+                StringBuilder json = new StringBuilder();
+                json.append("{{");
+
+                for (int i = 1; i <= columnCount; i++) {{
+                    String name = meta.getColumnLabel(i);
+                    Object value = result.getObject(i);
+
+                    json.append("\"").append(escapeJson(name)).append("\":");
+
+                    if (value == null) {{
+                        json.append("null");
+                    }} else if (value instanceof Number || value instanceof Boolean) {{
+                        json.append(value.toString());
+                    }} else {{
+                        json.append("\"").append(escapeJson(value.toString())).append("\"");
+                    }}
+
+                    if (i < columnCount) {{
+                        json.append(",");
+                    }}
+                }}
+
+                json.append("}}");
+                return Stream.of(new OutputRow(json.toString()));
+            }}
+
+            return Stream.empty();
+        }} catch (Exception e) {{
+            throw new RuntimeException("Failed to load data: " + e.getMessage());
+        }}
+    }}
+
+    public Stream<OutputRow> endPartition() {{
+        return Stream.empty();
+    }}
+}}
+$$
+;
+"""


src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,496 @@
...
+    "NUMERIC": DecimalType,
+    "VARCHAR": StringType,
+    "CHAR": StringType,
+    "CLOB": StringType,
+    "NCHAR": StringType,
+    "NVARCHAR": StringType,
+    "NCLOB": StringType,
+    "TIMESTAMP": TimestampType,
+    "VARBINARY": StringType,
+}
```
sfc-gh-aling: I see there are more types in JDBC spec like double/float/bool/time, ref: https://docs.oracle.com/javase/8/docs/api/java/sql/JDBCType.html, is there a reason why we just pick a small subset of list?

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,496 @@
...
+        )
+        self.secret_sql = f"SECRETS = ('cred' = {self.secret})"
+        self._infer_schema_successful = True
+        self.secret_detector()
+
+    @cached_property
+    def schema(self) -> StructType:
+        infer_schema_udtf_name = random_name_for_temp_object(TempObjectType.FUNCTION)
+        infer_schema_udtf_registration = f"""
+            CREATE OR REPLACE FUNCTION {infer_schema_udtf_name}(query VARCHAR)
```
sfc-gh-aling: is it creating a permanent function? do we need a TEMPORARY keyword?
```suggestion
            CREATE OR REPLACE TEMPORARY FUNCTION {infer_schema_udtf_name}(query VARCHAR)
```

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,496 @@
...
+        )
+
+    def read(self, partition_table: str) -> "DataFrame":
+        jdbc_ingestion_name = random_name_for_temp_object(TempObjectType.FUNCTION)
+        udtf_table_return_type = ", ".join(
+            [f"{field.name} VARCHAR" for field in self.schema.fields]
+        )
+
+        jdbc_udtf_registration = f"""
+            CREATE OR REPLACE FUNCTION {jdbc_ingestion_name}(query VARCHAR)
```
sfc-gh-aling: same comment, should this be a temp function?

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,496 @@
...
+        if self.properties is not None:
+            user_properties_overwrite = "\n".join(
+                [
+                    f'properties.put("{key}", "{value}");'
+                    for key, value in self.properties.items()
+                ]
+            )
+        get_secret = """
+                    SnowflakeSecrets secrets = SnowflakeSecrets.newInstance();
+                    UsernamePassword up = secrets.getUsernamePassword("cred");
```
sfc-gh-aling, line 478: does this throw error if "cred" is not available?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1298,156 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
             set_api_call_source(df, f"DataFrameReader.{format.lower()}")
         return df
 
+    @private_preview(version="1.36.0")
```
sfc-gh-aling: nit:
```suggestion
    @private_preview(version="1.38.0")
```

src/snowflake/snowpark/_internal/data_source/jdbc.py
sfc-gh-yuwang: yes

src/snowflake/snowpark/_internal/data_source/jdbc.py
sfc-gh-yuwang: that is a good point, I will add most of jdbc type to our support

src/snowflake/snowpark/_internal/data_source/jdbc.py
sfc-gh-yuwang: In python code, secret name was fetched and assigned to 'cred' in java stored procedure, so this only raise an error when secret is not correctly setup in EAI

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -570,6 +580,7 @@ def format(
 
         Args:
             format: The format of the file(s) to load. Supported formats are csv, json, avro, parquet, orc, and xml.
+                Snowpark data source type is also supported, currently support jdbc.
```
sfc-gh-aling: nice catch, should we also update line 576 to include "dbapi" and "jdbc"?

sfc-gh-aling:
```suggestion
                Snowpark data source type is also supported, currently support dbapi and jdbc.
```

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1298,156 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+            predicates: A list of expressions suitable for inclusion in WHERE clauses, where each expression defines a partition.
+                Partitions will be retrieved in parallel.
+                If both `column` and `predicates` are specified, `column` takes precedence.
+            session_init_statement: One or more SQL statements executed before fetching data from
+                the external data source.
+                This can be used for session initialization tasks such as setting configurations.
+                For example, `"SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED"` can be used in SQL Server
+                to avoid row locks and improve read performance.
+                The `session_init_statement` is executed only once at the beginning of each partition read.
+        """
```
sfc-gh-aling, line 1603: nit: sample code?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1298,156 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+                The dictionary may include the following keys:
+
+                - external_access_integration (str, required): The name of the external access integration,
+                    which allows the UDTF to access external endpoints.
+
+                - imports (List[str], required): A list of stage file names to import into the UDTF.
+                    Please include Jar file of jdbc driver to establish connection to external data source.
+
+                - java_version (int, optional): A integer that indicate the java runtime version of udtf.
+                    By default we use java 11.
```
sfc-gh-aling: emmm why not default to 17? usually the newer version is faster than the older one.
btw is our own JDBC code compatible with both 11 and 17?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1298,156 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+        secret = None
+        for r in eai_info:
+            if "ALLOWED_AUTHENTICATION_SECRETS" == r[0]:
+                secret = r[2][1:-1].split(",")[0].strip()
+                break
+
+        if external_access_integration is None or imports is None:
+            raise ValueError(
+                "external_access_integration and imports must be specified in udtf configs"
+            )
```
sfc-gh-aling: I think we can move this validation right after getting external_access_integration and imports. if EAI is None we do not need to send query to describe it

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1298,156 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+        imports = udtf_configs.get("imports", None)
+        packages = udtf_configs.get("packages", ["com.snowflake:snowpark:latest"])
+        java_version = udtf_configs.get("java_version", 11)
+        eai_info = self._session.sql(
+            f"describe external access integration {external_access_integration}"
+        ).collect()
+        secret = None
+        for r in eai_info:
+            if "ALLOWED_AUTHENTICATION_SECRETS" == r[0]:
+                secret = r[2][1:-1].split(",")[0].strip()
```
sfc-gh-aling: can you give some reference or specification in the comment why secret is parsed this way to get what?

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,504 @@
...
+    "java.lang.String": StringType,
+    "java.sql.Timestamp": TimestampType,
+    "java.sql.Date": DateType,
+    "java.sql.Array": ArrayType,
+    "java.lang.String[]": ArrayType,
+    "java.lang.Long": DecimalType,
+    "java.lang.Integer": DecimalType,
+    "java.lang.Short": DecimalType,
+}
+JDBC_TYPE_TO_SNOWFLAKE_TYPE = {
```
sfc-gh-aling, line 106: can we have a complete list here, looks like there are still some primitives we miss here: https://docs.oracle.com/javase/8/docs/api/java/sql/Types.html -- there are some primitive types like Date -> DateType, Time -> TimeType

or do you use another newer reference doc?


ideally we should have similar structure like the following:
 https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/data_source/drivers/psycopg2_driver.py#L41-L116

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,504 @@
...
+                        f" {field_name} of type {java_type} with precision {precision} and scale {scale}. "
+                        "The default Numeric precision and scale will be used."
+                    )
+                    precision, scale = None, None
+                data_type = snow_type(
+                    precision if precision is not None else 38,
+                    scale if scale is not None else 0,
+                )
+            elif snow_type == TimestampType:
+                data_type = snow_type(TimestampTimeZone.TZ)
```
sfc-gh-aling, line 555: why we default to TZ here?

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,504 @@
...
+        return StructType(fields)
+
+    def generate_select_sql(self):
+        select_sql_alias = (
+            f"SNOWPARK_JDBC_SELECT_SQL_ALIAS_{generate_random_alphanumeric(5)}"
+        )
+        cols = (
+            [col[0] for col in self.raw_schema]
+            if self.raw_schema is not None
+            else ["*"]
```
sfc-gh-aling: when would self.raw_schema be None?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1298,156 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+        upper_bound: Optional[Union[str, int]] = None,
+        num_partitions: Optional[int] = None,
+        query_timeout: Optional[int] = 0,
+        fetch_size: Optional[int] = 0,
+        custom_schema: Optional[Union[str, StructType]] = None,
+        predicates: Optional[List[str]] = None,
+        session_init_statement: Optional[List[str]] = None,
+        _emit_ast: bool = True,
+    ) -> DataFrame:
+        """
```
sfc-gh-aling, line 1503: we should call out the credential part
e.g., password in url will be ignored, eai with secret is required

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1298,156 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+        for r in eai_info:
+            if "ALLOWED_AUTHENTICATION_SECRETS" == r[0]:
+                secret = r[2][1:-1].split(",")[0].strip()
+                break
+
+        if external_access_integration is None or imports is None:
+            raise ValueError(
+                "external_access_integration and imports must be specified in udtf configs"
+            )
+        jdbc_client = JDBC(
```
sfc-gh-aling, line 1621: should we also throw error if secret is None?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1298,161 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+            "external_access_integration", None
+        )
+        imports = udtf_configs.get("imports", None)
+        packages = udtf_configs.get("packages", ["com.snowflake:snowpark:latest"])
+        java_version = udtf_configs.get("java_version", 11)
+        eai_info = self._session.sql(
+            f"describe external access integration {external_access_integration}"
+        ).collect()
+        secret = None
+        for r in eai_info:
```
sfc-gh-aling: 1. secret = None, raise error
2. len(secret) = 1, we just use this one, 
3. len(secret) > 1, we need extra user info, we require users to give us the secret name

----

case 2:
solution 1: secret name is required, we validate

sfc-gh-aling: check with Qinyi

src/snowflake/snowpark/_internal/data_source/jdbc.py
sfc-gh-aling: after experience this should be NTZ

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: Aligned with Qinyi: Secret will be a required field in udtf_configs

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1285,6 +1300,180 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
+                    which allows the UDTF to access external endpoints.
+
+                - secret (str, required): The name of the snowflake secret that contain username and password of
+                    external data source.
+
+                - imports (List[str], required): A list of stage file names to import into the UDTF.
+                    Please include Jar file of jdbc driver to establish connection to external data source.
+
+                - java_version (int, optional): A integer that indicate the java runtime version of udtf.
+                    By default we use java 11.
```
sfc-gh-aling: nit: 
```suggestion
                    By default we use java 17.
```

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,582 @@
...
+    from snowflake.snowpark.dataframe import DataFrame
+
+logger = logging.getLogger(__name__)
+
+# reference for java type to snowflake type:
+# https://docs.snowflake.com/en/developer-guide/udf-stored-procedure-data-type-mapping
+
+
+class JDBCType(Enum):
+    # ARRAY = "ARRAY"  # not supported in snowflake
```
sfc-gh-aling, line 48: let's create a JIRA as followups on these unsupported types? I feel some types can be supported with additional efforts (e.g., BLOB -> bytes -> hexstring -> snowflake binary)

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,582 @@
...
+                                row.scale = meta.getScale(i);
+                                row.nullable = meta.isNullable(i) == ResultSetMetaData.columnNullable;
+                                list.add(row);
+                            }}
+                        }}
+                    }} catch (SQLException e) {{
+                        throw new RuntimeException("SQL error: " + e.getMessage(), e);
+                    }}
+
+                    return list.stream();
```
sfc-gh-aling, line 281: IIUC, this loads all data into local first, which is likely to cause OOM if the data is large.

I think we should use Stream.generate like: https://medium.com/@orellabac/ingest-external-data-into-snowflake-with-snowpark-and-jdbc-eb487b61078c

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,582 @@
...
+                            OutputRow row = new OutputRow();
+                            {self.create_output_row_java_code()}
+                            list.add(row);
+                        }}
+                    }}
+                }} catch (SQLException e) {{
+                    throw new RuntimeException("SQL error: " + e.getMessage(), e);
+                }}
+
+                return list.stream();
```
sfc-gh-aling: IIUC, this loads all data into local first and then return the whole list, which is likely to cause OOM if the data is large.

I think we should use Stream.generate like: https://medium.com/@orellabac/ingest-external-data-into-snowflake-with-snowpark-and-jdbc-eb487b61078c

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,582 @@
...
+            self.schema,
+            _emit_ast=self._emit_ast,
+        )
+
+    def generate_create_connection(self):
+        user_properties_overwrite = ""
+        if self.properties is not None:
+            user_properties_overwrite = "\n".join(
+                [
+                    f'properties.put("{key}", "{value}");'
```
sfc-gh-aling, line 472: this converts both key and value to string, could there be non-string value?

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,582 @@
...
+        self.session = session
+        self.url = url
+        self.table_or_query = table_or_query
+
+        self.external_access_integration = external_access_integration
+        self.imports = imports
+        self.packages = packages
+        self.java_version = java_version
+        self.secret = secret
+        self.properties = properties
```
sfc-gh-aling: let's make a copy the properties dict, otherwise it's a reference to the original user input.
code like below alters user's original dict.

```python
                    del self.properties[key]
```

src/snowflake/snowpark/_internal/data_source/jdbc.py
sfc-gh-yuwang: No, all jdbc properties are stored as a string key-value pair

src/snowflake/snowpark/_internal/data_source/jdbc.py
sfc-gh-yuwang: that is a good point ,let me see how to do it

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,606 @@
...
+        self._emit_ast = _emit_ast
+        self.imports_sql = (
+            f"""IMPORTS =({",".join([f"'{imp}'" for imp in self.imports])})"""
+            if self.imports is not None
+            else ""
+        )
+        self.packages_sql = (
+            f"""PACKAGES=({','.join([f"'{pack}'" for pack in self.packages])})"""
+            if self.packages is not None
+            else "PACKAGES=('com.snowflake:snowpark:latest')"
```
sfc-gh-aling, line 201: two questions:

1. what's this `com.snowflake:snowpark:latest` for?
2. why it's an if-else logic

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,606 @@
...
+            """
+
+        self.session.sql(
+            infer_schema_udtf_registration, _emit_ast=self._emit_ast
+        ).collect()
+        self.raw_schema = self.session.sql(
+            f"SELECT * FROM TABLE({infer_schema_udtf_name}('{self.infer_schema_sql()}'))",
+            _emit_ast=self._emit_ast,
+        ).collect()
+
```
sfc-gh-aling, line 318: could you remind me what would happen if schema infer failed (self.raw_schema being None)

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,606 @@
...
+                                }} else {{
+                                    rs.close();
+                                    stmt.close();
+                                    this.conn.close();
+                                    return null;
+                                }}
+                            }} catch (SQLException e) {{
+                                e.printStackTrace();
+                                return null;
+                            }}
```
sfc-gh-aling: let's remove / just throw error

src/snowflake/snowpark/_internal/data_source/jdbc.py
sfc-gh-aling: we will raise error

src/snowflake/snowpark/_internal/data_source/jdbc.py
```diff
@@ -0,0 +1,606 @@
...
+                            }} else {{
+                                rs.close();
+                                stmt.close();
+                                this.conn.close();
+                                return null;
+                            }}
+                        }} catch (SQLException e) {{
+                            e.printStackTrace();
+                            return null;
+                        }}
```
sfc-gh-aling: let's raise error here / remove the except handle

src/snowflake/snowpark/_internal/data_source/jdbc.py
sfc-gh-aling: let's align with dbapi that we default add the package

sfc-gh-aling (approved):

PR #3601, opened 2025-07-29
DRAFT Jkew/conservative hybrid configuration

sfc-gh-jkew:
1. Which Jira issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   

   Fixes SNOW-NNNNNNN

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
      - [ ] If this test skips Local Testing mode, I'm requesting review from @snowflakedb/local-testing
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency
   - [ ] If this is a new feature/behavior, I'm adding the Local Testing parity changes.
   - [ ] I acknowledge that I have ensured my changes to be thread-safe. Follow the link for more information: [Thread-safe Developer Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#thread-safe-development)
   - [ ] If adding any arguments to public Snowpark APIs or creating new public Snowpark APIs, I acknowledge that I have ensured my changes include AST support. Follow the link for more information: [AST Support Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#ast-abstract-syntax-tree-support-in-snowpark)

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

PR #3602, opened 2025-07-29
SNOW-2230971: Support repr, joins, loc, reset_index, and binary ops in faster pandas

sfc-gh-helmeleegy:

   

   Fixes SNOW-2230971

   Support repr, joins, loc, reset_index, and binary ops in faster pandas.
   The env variable `SNOWPARK_PANDAS_DUMMY_ROW_POS_OPTIMIZATION_ENABLED` can be used to turn off the optimization. It's turned on by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,10 @@
 - Fixed a bug in hybrid execution mode (PrPr) where calling numpy operations like `np.where` on modin objects with the Pandas backend would raise an `AttributeError`. This fix requires `modin` version 0.34.0 or newer.
 - Fixed issue in `df.melt` where the resulting values have an additional suffix applied.
 
+#### Improvements
+
+- Improved performance by deferring row position computation.
+
 ## 1.33.0 (2025-06-19)
 
 ### Snowpark Python API Updates
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@
 
 - Added support for creating permanent and immutable UDFs/UDTFs with `DataFrame/Series/GroupBy.apply`, `map`, and `transform` by passing the `snowflake_udf_params` keyword argument. See documentation for details.
 
+#### Improvements
+
+- Improved performance by deferring row position computation. 
+  - The following operations are currently supported and can benefit from the optimization: `read_snowflake`, `repr`, `loc`, `reset_index`, `merge`, and binary operations.
+  - If a lazy object (e.g., DataFrame or Series) depends on a mix of supported and unsupported operations, the optimization will not be used.
+
 #### Bug Fixes
 
 - Fixed an issue where Snowpark pandas plugin would unconditionally disable `AutoSwitchBackend` even when users had explicitly configured it via environment variables or programmatically.
@@ -63,10 +69,6 @@
 
 - Reduced the number of UDFs/UDTFs created by repeated calls to `apply` or `map` with the same arguments on Snowpark pandas objects.
 - Added an example for reading a file from a stage in the docstring for `pd.read_excel`.
-- Improved performance by deferring row position computation. 
-  - The following operations are currently supported and can benefit from the optimization: `read_snowflake`, `repr`, `loc`, `reset_index`, `merge`, and binary operations.
-  - If a lazy object (e.g., DataFrame or Series) depends on a mix of supported and unsupported operations, the optimization will not be used.
-
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,10 @@
 
 - Reduced the number of UDFs/UDTFs created by repeated calls to `apply` or `map` with the same arguments on Snowpark pandas objects.
 - Added an example for reading a file from a stage in the docstring for `pd.read_excel`.
+- Improved performance by deferring row position computation. 
+  - The following operations are currently supported and can benefit from the optimization: `read_snowflake`, `repr`, `loc`, `reset_index`, `merge`, and binary operations.
+  - If a lazy object (e.g., DataFrame or Series) depends on a mix of supported and unsupported operations, the optimization will not be used.
+
 
 #### Bug Fixes
 
@@ -121,10 +125,6 @@
 - Fixed a bug in hybrid execution mode (PrPr) where certain Series operations would raise `TypeError: numpy.ndarray object is not callable`.
 - Fixed a bug in hybrid execution mode (PrPr) where calling numpy operations like `np.where` on modin objects with the Pandas backend would raise an `AttributeError`. This fix requires `modin` version 0.34.0 or newer.
 - Fixed issue in `df.melt` where the resulting values have an additional suffix applied.
-
-#### Improvements
-
-- Improved performance by deferring row position computation.
 
 ## 1.33.0 (2025-06-19)
 
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -332,7 +332,7 @@ class OrderedDataFrame:
         """
         return [col.snowflake_quoted_identifier for col in self.ordering_columns]
 
-    def _row_position_snowpark_column(self) -> Column:
+    def _row_position_snowpark_column(self, dummy_row_pos_mode: bool) -> Column:
         """
         Returns a row position Snowpark column for the dataframe.
         If row position column already exist in the current dataframe, it will be directly returned.
@@ -345,7 +345,15 @@ class OrderedDataFrame:
         if self.row_position_snowflake_quoted_identifier is not None:
             return Column(self.row_position_snowflake_quoted_identifier)
 
-        return row_number().over(Window.order_by(self._ordering_snowpark_columns())) - 1
+        if dummy_row_pos_mode:
+            from snowflake.snowpark.modin.plugin._internal.utils import pandas_lit
+
+            return pandas_lit(1)
+        else:
+            return (
+                row_number().over(Window.order_by(self._ordering_snowpark_columns()))
+                - 1
+            )
 
     @property
     def projected_column_snowflake_quoted_identifiers(self) -> list[str]:
@@ -355,7 +363,9 @@ class OrderedDataFrame:
         """
         return list(self._projected_column_snowflake_quoted_identifiers_tuple)
 
-    def ensure_row_position_column(self) -> "OrderedDataFrame":
+    def ensure_row_position_column(
+        self, dummy_row_pos_mode: bool = False
+    ) -> "OrderedDataFrame":
         """
         Returns an OrderedDataFrame with a row position column, and the row position column is guaranteed to
         be part of the projected column.
@@ -389,7 +399,7 @@ class OrderedDataFrame:
         )
         ordered_dataframe = self.select(
             *self.projected_column_snowflake_quoted_identifiers,
-            self._row_position_snowpark_column().as_(
+            self._row_position_snowpark_column(dummy_row_pos_mode).as_(
                 row_position_snowflake_quoted_identifier
             ),
         )
@@ -1133,6 +1143,7 @@ class OrderedDataFrame:
             "MatchComparator"  # noqa: F821
         ] = None,
         how: JoinTypeLit = "inner",
+        dummy_row_pos_mode: bool = False,
     ) -> "OrderedDataFrame":
         """
         Performs equi join of the specified type (``how``) with the current
@@ -1182,6 +1193,59 @@ class OrderedDataFrame:
                     For other join methods, the ordering columns preserves the left order, followed by right order.
 
         """
+        left = self
+        if False and dummy_row_pos_mode and how not in ["asof", "cross"]:
+            if (
+                left_on_cols is not None
+                and len(left_on_cols) == 1
+                and "row_position" in left_on_cols[0]
+            ):
+                new_col = (
+                    row_number().over(
+                        Window.order_by(left._ordering_snowpark_columns())
+                    )
+                    - 1
+                )
+                new_identifier = left.generate_snowflake_quoted_identifiers(
+                    pandas_labels=["row_position_left"],
+                    wrap_double_underscore=True,
+                )[0]
+                new_col = new_col.as_(new_identifier)
+                new_ordered_dataframe = left.select("*", new_col)
+                new_ordered_dataframe.row_position_snowflake_quoted_identifier = (
+                    new_identifier
+                )
+                new_ordered_dataframe = new_ordered_dataframe.sort(
+                    OrderingColumn(new_identifier)
+                )
+                left = new_ordered_dataframe
+                left_on_cols = [new_identifier]
+            if (
+                right_on_cols is not None
+                and len(right_on_cols) == 1
+                and "row_position" in right_on_cols[0]
+            ):
+                new_col = (
+                    row_number().over(
+                        Window.order_by(right._ordering_snowpark_columns())
+                    )
+                    - 1
+                )
+                new_identifier = right.generate_snowflake_quoted_identifiers(
+                    pandas_labels=["row_position_right"],
+                    wrap_double_underscore=True,
+                )[0]
+                new_col = new_col.as_(new_identifier)
+                new_ordered_dataframe = right.select("*", new_col)
+                new_ordered_dataframe.row_position_snowflake_quoted_identifier = (
+                    new_identifier
+                )
+                new_ordered_dataframe = new_ordered_dataframe.sort(
+                    OrderingColumn(new_identifier)
+                )
+                right = new_ordered_dataframe
+                right_on_cols = [new_identifier]
+
         left_on_cols = left_on_cols or []
         right_on_cols = right_on_cols or []
         assert len(left_on_cols) == len(
@@ -1189,7 +1253,7 @@ class OrderedDataFrame:
         ), "left_on_cols and right_on_cols must be of same length"
         _raise_if_identifier_not_exists(
             left_on_cols,
-            self.projected_column_snowflake_quoted_identifiers,
+            left.projected_column_snowflake_quoted_identifiers,
             "join left_on_cols",
         )
 
@@ -1204,7 +1268,7 @@ class OrderedDataFrame:
             assert right_match_col, "right_match_col was not provided to ASOF Join"
             _raise_if_identifier_not_exists(
                 [left_match_col],
-                self.projected_column_snowflake_quoted_identifiers,
+                left.projected_column_snowflake_quoted_identifiers,
                 "join left_match_col",
             )
             _raise_if_identifier_not_exists(
@@ -1221,7 +1285,7 @@ class OrderedDataFrame:
             "right",
             "inner",
             "outer",
-        ] and self._is_self_join_on_row_position_column(
+        ] and left._is_self_join_on_row_position_column(
             left_on_cols, right, right_on_cols
         ):
             is_join_needed = False
@@ -1241,9 +1305,9 @@ class OrderedDataFrame:
             right.projected_column_snowflake_quoted_identifiers
         )
 
-        # the new projected columns are set to self._projected_columns + right._projected_column after de-conflict
+        # the new projected columns are set to left._projected_columns + right._projected_column after de-conflict
         projected_column_snowflake_quoted_identifiers = (
-            self.projected_column_snowflake_quoted_identifiers
+            left.projected_column_snowflake_quoted_identifiers
             + right.projected_column_snowflake_quoted_identifiers
         )
 
@@ -1254,9 +1318,9 @@ class OrderedDataFrame:
             new_df = OrderedDataFrame(
                 right._dataframe_ref,
                 projected_column_snowflake_quoted_identifiers=projected_column_snowflake_quoted_identifiers,
-                ordering_columns=self.ordering_columns,
-                row_position_snowflake_quoted_identifier=self.row_position_snowflake_quoted_identifier,
-                row_count_snowflake_quoted_identifier=self.row_count_snowflake_quoted_identifier,
+                ordering_columns=left.ordering_columns,
+                row_position_snowflake_quoted_identifier=left.row_position_snowflake_quoted_identifier,
+                row_count_snowflake_quoted_identifier=left.row_count_snowflake_quoted_identifier,
             )
             new_df.row_count_upper_bound = RowCountEstimator.upper_bound(
                 self,
@@ -1274,7 +1338,7 @@ class OrderedDataFrame:
             return new_df
 
         # reproject the snowpark dataframe with only necessary columns
-        left_snowpark_dataframe_ref = self._to_projected_snowpark_dataframe_reference(
+        left_snowpark_dataframe_ref = left._to_projected_snowpark_dataframe_reference(
             include_ordering_columns=True
         )
         right_snowpark_dataframe_ref = right._to_projected_snowpark_dataframe_reference(
@@ -1334,9 +1398,9 @@ class OrderedDataFrame:
         # for right join, we preserve the right order first, then left order.
         # for all join type, left order is preserved first, then right order.
         if how == "right":
-            ordering_columns = right.ordering_columns + self.ordering_columns
+            ordering_columns = right.ordering_columns + left.ordering_columns
         else:
-            ordering_columns = self.ordering_columns + right.ordering_columns
+            ordering_columns = left.ordering_columns + right.ordering_columns
 
         new_df = OrderedDataFrame(
             DataFrameReference(
@@ -1429,6 +1493,7 @@ class OrderedDataFrame:
         right_on_cols: list[str],
         how: AlignTypeLit = "outer",
         enable_default_sort: bool = True,
+        dummy_row_pos_mode: bool = False,
     ) -> "OrderedDataFrame":
         """
         Performs align operation of the specified join type (``how``) with the current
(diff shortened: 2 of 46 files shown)


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8136,6 +8259,28 @@ def merge(
             )
 
         left = self
+        relaxed_query_compiler = None
+        if (
```
sfc-gh-jkew, line 8270: I'm still concerned that maintaining two query compilers is going to be very complicated. This is going to make things really hard to reason about if we have to maintain them simultaneously.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1367,11 +1369,23 @@ def from_snowflake(
         index_col: Optional[Union[str, list[str]]] = None,
         columns: Optional[list[str]] = None,
         enforce_ordering: bool = False,
+        dummy_row_pos_mode: bool = False,
     ) -> "SnowflakeQueryCompiler":
         """
         See detailed docstring and examples in ``read_snowflake`` in frontend layer:
         src/snowflake/snowpark/modin/plugin/pd_extensions.py
         """
+        relaxed_query_compiler = None
```
sfc-gh-jkew, line 1381: I have an example where I tried to add a modin-style config here, but I cannot yet use it with a config context:

https://github.com/snowflakedb/snowpark-python/pull/3601/files

sfc-gh-helmeleegy: It should be easy to use an environment variable for this. I just found that there are still some failing tests. So I wanted to focus on them.

sfc-gh-helmeleegy: I added an environment variable (`SNOWPARK_PANDAS_DUMMY_ROW_POS_OPTIMIZATION_ENABLED `) to turn the optimization on or off as needed. It's on by default.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1182,14 +1193,67 @@ def join(
                     For other join methods, the ordering columns preserves the left order, followed by right order.
 
         """
+        left = self
+        if False and dummy_row_pos_mode and how not in ["asof", "cross"]:
```
sfc-gh-jkew: I know this is a draft; but is "if False" just temporary? As an aside; most of the *utils changes are pretty simple with the notable exception of this block; which would never be executed.

sfc-gh-helmeleegy: You're right. This code block was disabled during testing. Re-enabled now. It is meant to handle the case of joining two different df's on their row positions. It's not a case that's needed by CBOE because they only need self-joins on row positions when using filters, where the join is optimized away altogether. But the general case still needs to be handled because we cannot join on dummy row positions. The rest of the *utils changes mainly plumb the `dummy_row_pos_mode` flag from the qc layer down to the ordered dataframe layer.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: I think the modin approach is a lot nicer for adding environment variables: https://github.com/snowflakedb/snowpark-python/pull/3601/files#diff-644ecac66746749550c23f6a19cb2ea505c2208548e4606d8ee9209c0c174baeR81

I like that it uses the existing modin configuration bits; even though it currently requires monkey patching modin and we need to clean up the forked version of envvars.py

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2630,6 +2672,24 @@ def binary_op(
 
         # Native pandas does not support binary operations between a Series and a list-like object.
 
+        relaxed_query_compiler = None
```
sfc-gh-jkew, line 2669: Could we write this with a context manager? The following is not quite right because `qc` can change; but the fact that you essentially bookend most of the qc operations with synchronization on the relaxed_qc makes me think we could take advantage of a context_manager-like approach here:
```
with relaxed_sync(self, self.binary_op, locals()):
     ... normal logic
```
which would roughly translate into something like the following:
```
relaxed_qc = self.relaxed_query_compiler
operation = 
try:
    if relaxed_qc is not None:
         relaxed_qc = relaxed_qc=operation(locals)
    < inner of context manager >
finally:
    if relaxed_qc is not None:
          qc._relaxed_query_compiler = relaxed_qc
```

sfc-gh-helmeleegy: Hmm, I'm not sure exactly how to go about this at the moment. But making the process of onboarding more qc operations more streamlined can certainly help. For now, I'm most interested in making sure that the required operations are supported under this faster pandas framework to unlock customer workloads.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: This sounds good. So you're suggesting that we later come back to use a modin based config instead of the env var?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: I'm also concerned about the maintenance cost of having two query compilers. Also, we do need more SQL queries in some cases with this approach. @sfc-gh-helmeleegy are you sure that we can't start with a single dummy-ordered dataframe from which we project a real ordering column whenever we need it? Like I suggested [here](https://docs.google.com/document/d/1yGe8c1W3HGOOlsCe58qLrO5gG0I8pKR_iKp-6q4vp4k/edit?disco=AAABjebTAAk).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: Yeah, I think it's easier to integrate into code than having to use the os.* libraries in client code. I sort of like modin's config system even though it could use some cleanup.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Yes, I've already tried using a single query compiler using dummy row positions and the amount of test failures resulting from this did not feel like they could all be addressed in a reasonable amount of time. What you're suggesting in the comment is effectively what we have now. It's just that we added a boolean parameter to distinguish between the two uses instead of having two distinct methods.

In general, when introducing a new approach that breaks one of the fundamental assumptions of the entire code base, I prefer that we take incremental and careful steps with the option to fallback to the original approach whenever necessary.

sfc-gh-helmeleegy: @sfc-gh-mvashishtha Sorry I forgot to answer your other question about the number of queries. Yes, we do need to increase the number of queries mainly because of the eager queries triggered by the two query compilers The good news though is that one of the guaranteed eager queries, which computes the row count at read_snowflake time, is now going to use the dummy mode making it much faster. This added speed will apply to both query compilers. So despite issuing more eager queries, the total time might improve even for workloads whose operations are not fully onboarded on Faster pandas yet.

tests/integ/modin/series/test_str_accessor.py
```diff
@@ -725,7 +725,7 @@ def test_str_len_list_coin_base(session, enable_sql_simplifier):
     expected_udf_count = 2
     if session.sql_simplifier_enabled:
         expected_udf_count = 1
-    with SqlCounter(query_count=9, udf_count=expected_udf_count):
+    with SqlCounter(query_count=5):
```
sfc-gh-jkew, line 728: this drop is a bit surprising. Any idea why?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: Can we have a quick benchmark on this; on maybe astronomer-attribution or astronomer-jaffle to see what the difference is?

tests/integ/modin/series/test_str_accessor.py
sfc-gh-helmeleegy: It's actually not a drop. It's the usual increase by 1 from 9 to 10. Going to 10 or higher though requires justification. So I split the block of code into two blocks with 5 queries each.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: For the perf improvements made by the new approach, we already have numbers for the CBOE workload which is most relevant. https://docs.google.com/document/d/1DXb-WduSRX9bTFUaju--8vRrhb7XWSKb3hnLSEpCt_w/edit?tab=t.0#bookmark=id.y2s7lllozkso

If you're referring to the 2 QCs vs 1 QC, it will depend on the data size and what other eager queries are triggered by each QC.

Do you know how to run one of those benchmarks using this branch with and without the flag? cc: @sfc-gh-joshi

sfc-gh-helmeleegy: @sfc-gh-jkew I ran the astronomer-attribution benchmark with and without the dummy mode and the results were almost identical (slightly better with the dummy mode):

[with the dummy mode](https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkPandasPerfBenchmarkRunner/197635/console): 26secs (from 2025-07-31 16:12:31 to 2025-07-31 16:12:57)
[without the dummy mode](https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkPandasPerfBenchmarkRunner/197636/console): 28secs (from 2025-07-31 16:35:53 to 2025-07-31 16:36:21)

Runtimes were extracted from the console logs as the time diff between the log lines `run workload astronomer_attribution 1/1` and `teardown workload astronomer_attribution 1/1`.

(Of course we're only comparing here the overhead of eager queries in both approaches. The actual perf improvements from the dummy mode optimization is likely not applicable to the astronomer workload yet because many of the pandas operations it uses are not supported in Faster pandas yet.)

sfc-gh-mvashishtha: @sfc-gh-helmeleegy 

> What you're suggesting in the comment is effectively what we have now. 

Would the approach I suggested also increase query counts? If not, would it be worth it to look further into that approach?

>  when introducing a new approach that breaks one of the fundamental assumptions of the entire code base, I prefer that we take incremental and careful steps with the option to fallback to the original approach whenever necessary.

I agree, but couldn't the other solution be applied incrementally? We would replace all instances of get_row_position_column() with get_dummy_row_position_column, except for cases where this version sets `dummy_row_pos_mode=True`.

sfc-gh-helmeleegy: In a single qc solution, you cannot start by using a dummy row pos column (because the first few operations happened to be currently supported) and then simply switch back to a real row pos column (because an unsupported operation is encountered). All those currently unsupported operations are expecting their input dataframe to have a real row pos column. So passing to them a dataframe with a dummy row pos column and expecting them to work normally results in so many test failures.

If I'm not mistaken the approach you suggested is to have two copies of get_row_position_column() into a real one and a dummy one, and then to incrementally have some of the QC methods use the dummy copy instead of the real one. All the others will keep using the real one, correct? If so then the splitting part is really not any different than reusing the same method while passing a new param, which is what we do, agree? The other part of the suggestion is to maintain one QC instead of two. That is what I have already started with. And it didn't work as you expect because of what I explained above.

CHANGELOG.md
```diff
@@ -122,6 +122,10 @@
 - Fixed a bug in hybrid execution mode (PrPr) where calling numpy operations like `np.where` on modin objects with the Pandas backend would raise an `AttributeError`. This fix requires `modin` version 0.34.0 or newer.
 - Fixed issue in `df.melt` where the resulting values have an additional suffix applied.
 
+#### Improvements
```
sfc-gh-mvashishtha: I think this note is in the section for the previous Snowpark version.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1221,11 +1233,63 @@ def join(
...
             left_on_cols, right, right_on_cols
         ):
             is_join_needed = False
 
+        if is_join_needed and dummy_row_pos_mode and how not in ["asof", "cross"]:
+            if (
+                left_on_cols is not None
+                and len(left_on_cols) == 1
+                and "row_position" in left_on_cols[0]
+            ):
```
sfc-gh-mvashishtha:
```suggestion
		if left_on_cols == [left.row_position_snowflake_quoted_identifier]:
```

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1221,11 +1233,63 @@ def join(
             "right",
             "inner",
             "outer",
-        ] and self._is_self_join_on_row_position_column(
+        ] and left._is_self_join_on_row_position_column(
             left_on_cols, right, right_on_cols
         ):
             is_join_needed = False
 
+        if is_join_needed and dummy_row_pos_mode and how not in ["asof", "cross"]:
```
sfc-gh-mvashishtha, line 1249: Can we please duplicate this and the next if block?

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -434,22 +435,39 @@ def create_initial_ordered_dataframe(
         # if no snowflake_quoted_identifiers is specified, all columns will be selected
         if enforce_ordering:
             row_position_column_str = f"{METADATA_ROW_POSITION_COLUMN} as {row_position_snowflake_quoted_identifier}"
+            row_position_column_str_for_count = row_position_column_str
```
sfc-gh-mvashishtha: Why does the dummy row position not suffice? Doesn't SnowparkDataFrame.count() generate `COUNT(1)` to compute row count anyway?

src/snowflake/snowpark/modin/plugin/io/snow_io.py
```diff
@@ -251,7 +251,10 @@ def read_snowflake(
         src/snowflake/snowpark/modin/plugin/pd_extensions.py
         """
         return cls.query_compiler_cls.from_snowflake(
-            name_or_query, index_col, columns, enforce_ordering=enforce_ordering
+            name_or_query,
+            index_col,
+            columns,
+            enforce_ordering=enforce_ordering,
```
sfc-gh-mvashishtha, line 257:
```suggestion
            name_or_query, index_col, columns, enforce_ordering=enforce_ordering,
```

sfc-gh-mvashishtha (commented):
Here's one batch of comments. I reviewed everything except snowflake_query_compiler.py.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: @sfc-gh-mvashishtha @sfc-gh-jkew I reran the two benchmarking jobs with all 15 benchmarks instead of just one:

[with dummy mode](https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkPandasPerfBenchmarkRunner/197640/console): 22mins 43secs (from 2025-08-01 11:31:23 to 2025-08-01 11:54:06)
[without dummy mode](https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkPandasPerfBenchmarkRunner/197641/console): 22mins 46secs (from 2025-08-01 12:13:47 to 2025-08-01 12:36:33)

Runtimes were extracted from the console logs as the time diff between the log lines `setup workload astronomer_attribution 1/15` and `teardown workload tpcxbb_q06 15/15`.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-helmeleegy: This is to ensure that the inner query of the count(1) query does not compute that row_number() window function. You can find in the second branch of the if.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-helmeleegy: You mean "deduplicate", right? Well, the few lines between the two if blocks use the left and right ordered data frames, which are updated in the first if block.

sfc-gh-mvashishtha: Yes, deduplicate. I think you can write a helper function to force ordering and return the new ordered frame + `left_on_cols`.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1239,11 +1239,7 @@ def join(
             is_join_needed = False
 
         if is_join_needed and dummy_row_pos_mode and how not in ["asof", "cross"]:
```
sfc-gh-mvashishtha, line 1249: Would this code be correct if the left join columns were `[dummy_row_position, data_column_0]`?

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-mvashishtha: I understand that we need a different `row_position_column_str`, but I don't understand why we also need `row_position_column_str_for_count` and a separate ordered dataframe, `ordered_dataframe_for_count`.

sfc-gh-mvashishtha (commented):
still have to review query compiler

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2058,7 +2087,9 @@ def _shift_values(
             return self._shift_values_axis_1(periods, fill_value)
 
     def _shift_values_axis_0(
-        self, periods: int, fill_value: Hashable
+        self,
+        periods: int,
+        fill_value: Hashable,
```
sfc-gh-mvashishtha:
```suggestion
        self, periods: int, fill_value: Hashable,
```

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-helmeleegy: The idea was to make sure that it's used even with the normal query compiler (that doesn't use dummy row pos). This way (with the 2-query-compiler approach), all queries and eager queries will be free of the row_number() window function at the lower levels.

sfc-gh-mvashishtha: Would it suffice to instead do `materialized_row_count = ordered_dataframe._dataframe_ref.snowpark_dataframe.select(lit(None)).count(
        statement_params=get_default_snowpark_pandas_statement_params(), _emit_ast=False)`

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -3535,6 +3629,18 @@ def reset_index(
         Returns:
             A new SnowflakeQueryCompiler instance with updated index.
         """
+        relaxed_query_compiler = None
+        if self._relaxed_query_compiler is not None and not drop:
```
sfc-gh-mvashishtha, line 3627: add a comment that if `drop`, we have to compute a new row number?

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-helmeleegy: Good point. But I don't feel this is a normal scenario for users to mix the row position with data columns in their join keys. The row position is supposed to be unique and so joining on it should be enough. However, I guess we can change the equality check into an inclusion check to account for this.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13585,25 +13817,93 @@ def build_repr_df(
...
+                data_column_pandas_index_names=frame.data_column_pandas_index_names,
+                data_column_pandas_labels=frame.data_column_pandas_labels,
+                data_column_snowflake_quoted_identifiers=frame.data_column_snowflake_quoted_identifiers,
+                index_column_pandas_labels=frame.index_column_pandas_labels,
+                index_column_snowflake_quoted_identifiers=frame.index_column_snowflake_quoted_identifiers,
+                data_column_types=frame.cached_data_column_snowpark_pandas_types,
+                index_column_types=frame.cached_index_column_snowpark_pandas_types,
             )
-        )
+            if len(new_frame.index_column_snowflake_quoted_identifiers) == 1 and (
```
sfc-gh-mvashishtha, line 13858: I don't understand this condition or what this block is doing. Could you add a comment?

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-helmeleegy: Tried it but it didn't work. I think the reason is that when `dummy_row_pos_mode` is False, `ordered_dataframe` would have the `row_number()` call embedded in it.

sfc-gh-mvashishtha: Try doing `materialized_row_count = initial_ordered_dataframe._dataframe_ref.snowpark_dataframe.select(lit(None)).count( statement_params=get_default_snowpark_pandas_statement_params(), _emit_ast=False)`?

sfc-gh-helmeleegy: Nice. Worked and we don't need the `select(pandas_lit(None))` either.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: Ok. The astronomer results are encouraging

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Added a comment.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Added a comment.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: I did it, but it seems black keeps changing it back!

sfc-gh-mvashishtha: Have you tried turning off auto-format by black in your IDE?

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-mvashishtha: Just a comment is ok.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: No. I'm using vscode. Do you know where it's done there?

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-helmeleegy: It's already using an inclusion check now.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: 1. command-shift-p
2. search for "open workspace settings (JSON)"
3. add `    "editor.formatOnSave": true` to the top level of the json object. make sure there's no override of `formatOnSave` in the JSON.
4. repeat but for "open user settings (JSON)"

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-helmeleegy: I ended up using the existing `ensure_row_position_column` instead.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-helmeleegy: Added the comment.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Done.

sfc-gh-mvashishtha (approved):
Thanks for addressing my comments!

PR #3604, opened 2025-07-29
SNOW-2171718: Do not set AutoSwitchBackend to False if it's True

sfc-gh-mvashishtha:
Stop disabling AutoSwitchBackend when the user has set it to True.

Add tests for AutoSwitchBackend configuration scenarios:
  - User explicitly sets to True
  - User hasn't changed default
  - Environment variable enabled/disabled
- Added tox environment to run each new test case separately with `--noconftest`
- Updated CI to run AutoSwitchBackend tests in precommit workflow

Fixes SNOW-2171718


sfc-gh-joshi (approved):

.github/workflows/precommit.yml
```diff
@@ -621,6 +621,29 @@ jobs:
             .tox/.coverage
             .tox/coverage.xml
 
+              test-unsupported-modin-or-pandas:
+
+  test-snowpark-pandas-auto-switch-backend:
```
sfc-gh-jkew: This seems... extraordinarily expensive testing for this type of problem. Is there a way to avoid having a new job?

tests/integ/modin/hybrid/auto_switch_backend/test_environment_variable_disabled.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def test_auto_switch_backend_environment_variable_disabled_SNOW_2171718(monkeypatch):
+    """Test that importing snowflake.snowpark.modin.plugin respects AutoSwitchBackend=False set via environment variable"""
+
+    monkeypatch.setenv("MODIN_AUTO_SWITCH_BACKENDS", "False")
+
+    from modin.config import AutoSwitchBackend
+
+    assert AutoSwitchBackend.get() is False
```
sfc-gh-joshi, line 13: should we replace the `is False` and `is True` assertions with just `not AutoSwitchBackend.get()` and `AutoSwitchBackend.get()`?

.github/workflows/precommit.yml
sfc-gh-joshi: Right now our testing suite disables this via an explicit `AutoSwitchBackend.switch()` call in `tests/integ/modin/conftest.py`. If we convert that to an auto-used fixture that creates a `config_context(AutoSwitchBackend=False)` and yields before each test + move the environment variable tests to a separate folder, we could potentially avoid running these in a separate environment.

sfc-gh-jkew: Yeah, I think that might make sense to me. This is also one of those things which may be partially temporary as well until it is defaulted to True.

.github/workflows/precommit.yml
```diff
@@ -621,6 +621,29 @@ jobs:
...
+    name: Test Snowpark pandas AutoSwitchBackend
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
```
sfc-gh-yuwang: I don't have much context on this test, is it designed to only run on 3.9?

.github/workflows/precommit.yml
```diff
@@ -621,6 +621,29 @@ jobs:
             .tox/.coverage
             .tox/coverage.xml
 
+              test-unsupported-modin-or-pandas:
```
sfc-gh-mvashishtha:
```suggestion
```

.github/workflows/precommit.yml
sfc-gh-joshi: We're changing how Snowpark pandas interacts with a configuration variable supplied by Modin, so we're pretty confident that the behavior will be the same across all Python versions. The one-off task here is to make sure it's tested in isolation from the rest of our tests, which explicitly modify this configuration variable's value.

We're discussing in other comments on whether we can avoid creating an extra tox environment + task for this test.

tests/integ/modin/hybrid/auto_switch_backend/test_environment_variable_disabled.py
sfc-gh-mvashishtha: I don't think it matters much either way. We do only care whether the variable is truthy or falsey, but also it shouldn't be a non-bool value.

.github/workflows/precommit.yml
sfc-gh-mvashishtha: I've changed the snowparkpandasnotdoctest tox command so it now runs each of these tests with a separate `pytest` invocation. We can't run these tests along with the rest of the test suite because they are testing imports.

.github/workflows/precommit.yml
sfc-gh-mvashishtha: Resolved; I reverted the workflow changes.

sfc-gh-helmeleegy (approved):

sfc-gh-jkew (approved):

PR #3606, opened 2025-07-29
Trace SQL error to DF bug fixes

sfc-gh-daviwang:

   

   Fixes NO-SNOW

   Fixing small bug with eager schema validation enabled for tracing SQL error to DF, and adding line numbers to integration tests

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -679,9 +679,9 @@ class DataFrame:
     @_ast_id.setter
     def _ast_id(self, value: Optional[int]) -> None:
         self.__ast_id = value
-        if self._plan is not None:
+        if self._plan is not None and value is not None:
             self._plan.add_df_ast_id(value)
-        if self._select_statement is not None:
+        if self._select_statement is not None and value is not None:
             self._select_statement.add_df_ast_id(value)
 
     @publicapi
@@ -1737,6 +1737,7 @@ class DataFrame:
                 )
 
             return self._with_plan(self._select_statement.select(names), _ast_stmt=stmt)
+
         return self._with_plan(Project(names, join_plan or self._plan), _ast_stmt=stmt)
 
     @df_api_usage
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -6415,6 +6415,7 @@ Query List:
 
         if _ast_stmt is not None:
             df._ast_id = _ast_stmt.uid
+
         return df
 
     def _get_column_names_from_column_or_name_list(
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -1737,7 +1737,6 @@ class DataFrame:
                 )
 
             return self._with_plan(self._select_statement.select(names), _ast_stmt=stmt)
-
         return self._with_plan(Project(names, join_plan or self._plan), _ast_stmt=stmt)
 
     @df_api_usage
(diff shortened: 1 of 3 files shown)


tests/integ/test_trace_sql_errors_to_df.py
```diff
@@ -41,7 +42,7 @@ def test_python_source_location_in_sql_error(session):
         df2.collect()
     assert "SQL compilation error corresponds to Python source" in str(
         ex.value.debug_context
-    )
+    ) and "line 40" in str(ex.value.debug_context)
```
sfc-gh-aling: is the line number in these tests stable?

if I add new tests/new imports/new pytestsmark, would those line number be affected?
I'm concerned that any change to this test file causing line number change will fail the tests

tests/integ/test_trace_sql_errors_to_df.py
```diff
@@ -55,6 +56,10 @@ def test_python_source_location_in_session_sql(session):
     assert "SQL compilation error corresponds to Python source" in str(
         ex.value.debug_context
     )
+    if sys.version_info[:2] == (3, 9):
+        assert "line 51" in str(ex.value.debug_context)
```
sfc-gh-aalam: btw, this works for me

```
    pytest.mark.skipif(
        sys.version_info < (3, 10),
        reason="...",
        run=False,
    ),
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -6410,12 +6410,11 @@ def _with_plan(self, plan, _ast_stmt=None) -> "DataFrame":
         """
         :param proto.Bind ast_stmt: The AST statement protobuf corresponding to this value.
         """
-        df = DataFrame(self._session, plan, _ast_stmt=_ast_stmt, _emit_ast=False)
+        df = DataFrame(self._session, plan, _ast_stmt=_ast_stmt)
         df._statement_params = self._statement_params
 
         if _ast_stmt is not None:
             df._ast_id = _ast_stmt.uid
-
```
sfc-gh-aalam: nit: undo

sfc-gh-aalam (approved):

tests/integ/test_trace_sql_errors_to_df.py
sfc-gh-aalam: this is a good point and I was going to make a similar comment. I think if new tests are added, they must be added at the end of file. We can make a note at the top of this file about this.

src/snowflake/snowpark/dataframe.py
```diff
@@ -6410,12 +6410,11 @@ def _with_plan(self, plan, _ast_stmt=None) -> "DataFrame":
         """
         :param proto.Bind ast_stmt: The AST statement protobuf corresponding to this value.
         """
-        df = DataFrame(self._session, plan, _ast_stmt=_ast_stmt, _emit_ast=False)
+        df = DataFrame(self._session, plan, _ast_stmt=_ast_stmt)
```
sfc-gh-aling, line 6413: can you help me understand the change here: what it means for _with_plan to emit ast = True/why _emit_ast=False would not work

tests/integ/test_trace_sql_errors_to_df.py
sfc-gh-daviwang: The line number is only affected when changing lines inside this test file

src/snowflake/snowpark/dataframe.py
sfc-gh-daviwang: If we don't emit_ast here and there is an error during eager schema validation, we don't update the plan for the dataframe with the ast expressions they are responsible for.  This means that the debug trace we output won't include the most recent line that causes the error.

tests/integ/test_trace_sql_errors_to_df.py
sfc-gh-daviwang: Yeah I realized there's an issue with only doing sys.version_info == (3, 9). I can do this instead

tests/integ/test_trace_sql_errors_to_df.py
sfc-gh-aling: I still feel this could be cumbersome for developers who's going to work on this part in the future.
can we dynamically get the line number and assert using relative location? see my following code:

```python
import sys

def get_current_line_number_sys():
    """Returns the current line number of the caller using sys._getframe()."""
    # sys._getframe(1) gets the frame of the caller (index 0 is the current frame)
    return sys._getframe(1).f_lineno

print(f"This is line {get_current_line_number_sys()}")
y = 20
print(f"This line is {get_current_line_number_sys()}")
```

PR #3610, opened 2025-07-31
SNOW-2230553 - Reduce Telemetry Overhead when running w/ a Pandas Engine

sfc-gh-jkew:
SNOW-2230553 - Reduce Telemetry Overhead when running w/ a Pandas Engine

### Problem
We have a different way of sending telemetry from the modin layer when using the eager pandas backend. These metrics are sent individually for a number of events, but prior to this change they were not pre-aggregated on the client. This would result in a significant number of telemetry messages in the TelemetryClient, which would quickly exhaust the queue and flush all messages more frequently than desired.

### Solution
We select only hybrid decision metrics and query compiler api metrics to process, and then further pre-aggregate them on the client into 5 second windows. This will mean that we may loose some telemetry from the client, which is OK - because telemetry is advisory only. For each metric we produce some summary statistics for the aggregated results.

### Summary
- Send fewer modin metrics, focusing on query compiler timings and hybrid decision making metrics.
- Reduces the total number of telemetry messages from snowpandas to 12% of previous versions.
- Aggregate modin metrics for every 5 second window using count, median, and mean.
- Refactored telemetry integration tests to use mocks instead of pulling logs from the connector telemetry client; because of intermittent failures during tests.


src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
+      modin.query-compiler.snowflakequerycompiler.value_counts.stat.mean
+      modin.query-compiler.snowflakequerycompiler.value_counts.stat.median
+      modin.query-compiler.snowflakequerycompiler.value_counts.stat.count
+      modin.hybrid.auto.decision.Pandas.count
+      modin.hybrid.auto.decision.Snowflake.mean
+      ...
+    """
+    global _modin_event_log
+    try:
+        summary_stat_names = ["count", "median", "mean"]
```
sfc-gh-helmeleegy, line 711: Do we need other aggregations like, say, sum?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
-                event=metric_name,
-                value=metric_value,
-                aggregatable=False,
-            )
-        except Exception:
-            pass
+    simplified_metric = metric_name
+
+    metric_valid = False
+    # ignore telemetry from dunder and internal metrics
```
sfc-gh-helmeleegy, line 740: what does dunder mean?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
+    if metric_name.startswith(MODIN_PERFORMANCE_METRICS):
+        parts = metric_name.split(".")
+        if parts[3].startswith("_"):
```
sfc-gh-helmeleegy, line 743: Do we want to check if parts will always have more than 3 elements?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
         return type.__new__(cls, name, bases, attrs)
 
 
+_modin_event_log: list = [[]]
+_next_modin_metric_flush: float = 0
+_modin_metric_flush_interval = 0
+
+MODIN_SWITCH_DECISION_METRICS = (
```
sfc-gh-helmeleegy: Perhaps call this `MODIN_SWITCH_DECISION_METRIC_PREFIXES`?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
+    "modin.hybrid.merge.decision",
+    "modin.hybrid.auto.decision",
+)
+MODIN_PERFORMANCE_METRICS = ("modin.query-compiler",)
```
sfc-gh-helmeleegy: Same

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
+MODIN_ALL_COLLECTED_METRICS = MODIN_SWITCH_DECISION_METRICS + MODIN_PERFORMANCE_METRICS
```
sfc-gh-helmeleegy: Where is this used?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
+                    event=stat_specific_metric,
+                    value=row[1][stat],
+                    aggregatable=stat == ("value", "count"),
+                )
+    except Exception:
+        pass
+    _modin_event_log = []
+
+
 def modin_telemetry_watcher(metric_name: str, metric_value: Union[int, float]) -> None:
```
sfc-gh-helmeleegy, line 732: For a metric like, say, `modin.query-compiler.snowflakequerycompiler.value_counts`, how is the `metric_value` currently populated and what does it represent -- before any time window aggregation? Does it represent the number of times `value_counts` was called since the last time `modin_telemetry_watcher` is called? Or is `metric_value` always equal to 1 in this case?

sfc-gh-helmeleegy (commented):
Are there dashboards that are currently dependent on this telemetry? If so, do we know how they will be impacted? I'm seeing each old metric replaced by three new metrics, right?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
+      modin.hybrid.auto.decision.Snowflake.mean
+      ...
+    """
+    global _modin_event_log
+    try:
+        summary_stat_names = ["count", "median", "mean"]
+        processing_df = native_pd.DataFrame(
+            _modin_event_log, columns=["metric", "value"]
+        )
+        summary_stats = processing_df.groupby("metric").agg(summary_stat_names)
```
sfc-gh-lmukhopadhyay, line 715: Where do we make the distinction here between aggregating modin.query-compiler and modin.hybrid stats?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-joshi: methods starting with __

src/snowflake/snowpark/modin/plugin/_internal/__init__.py
```diff
@@ -1,3 +1,6 @@
 #
 # Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
 #
+from snowflake.snowpark.modin.plugin._internal.telemetry import ModinTelemetrySender
+
+__all__ = ["ModinTelemetrySender"]
```
sfc-gh-joshi, line 7: why is this necessary?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
+
+MODIN_SWITCH_DECISION_METRICS = (
+    "modin.hybrid.merge.decision",
+    "modin.hybrid.auto.decision",
+)
+MODIN_PERFORMANCE_METRICS = ("modin.query-compiler",)
+MODIN_ALL_COLLECTED_METRICS = MODIN_SWITCH_DECISION_METRICS + MODIN_PERFORMANCE_METRICS
+
+
+def _check_and_reset_metric_flush_time() -> bool:
```
sfc-gh-joshi, line 676: Should we have this function directly call flush instead of returning a flag asking the caller to do so?

src/snowflake/snowpark/modin/config/envvars.py
```diff
@@ -78,6 +78,34 @@ def get_help(cls) -> str:
         return help
 
 
+class SnowflakeModinTelemetryFlushInterval(EnvironmentVariable, type=int):
+    """
+    Minimum number of seconds between a flush of telemetry to snowflake
+    from metrics generated in the client modin layer.
```
sfc-gh-joshi, line 84: There's no background thread or anything running at this interval, right? It only checks this interval when an operation that generates metrics is performed?

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-jkew: Yeah, the terminology is weird to me too. Maybe in the future we don't want to ignore a set of dunder methods either; but I wanted to sort of blanket cull a bunch of weird ones.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-jkew: The metric which is fired in modin is the execution time of that api call; when we perform the aggregation ( count, median, mean ) we collect the total count of all calls as well as the median/mean from the metric_value.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-jkew: good point; an earlier version used that as another filter. Not sure why it didn't get caught my the automated tools though...

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-jkew: Not sure if it is necessary here. There are some assumptions on the format here for these metrics which are aligned with modin, but I think we would catch those if they diverged.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-jkew: This is a good question. "sum" would represent the sum-total time of an api call like "value_counts" across the five-minute window. From that we could compute the mean certainly and probably do some other descriptive statistics inside of snowhouse.

We can always add it later if we need to, but I wanted to keep the list very small. I don't even like "mean" as a metric because averages are dumb, but other people like it so I reluctantly included it. What I would really like to add is a p95; but I worry that the sample size is too low to make that meaningful in any way.

With "median" I can do something called a "median of medians" or a "ninther" which is an approximation of median useful for a large data. It's not exactly a median, but it's close enough without being distracted by outliers.

Anyways; we can add it later if we want.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-jkew: I did this for testing reasons; to essentially allow me to just mock this single function where I actually call time()

src/snowflake/snowpark/modin/config/envvars.py
sfc-gh-jkew: Right. No background thread. I didn't want to introduce an additional context switch.

It will even loose the last metrics generated before the process ends, which is OK for now. Maybe in the future I will refactor it to capture those - but mainly I'm not interested in individual timings but the distribution of timings.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-jkew: I don't. Technically we do not need to aggregate modin.hybrid stats with median and mean, which will always be exactly "1"... but it was nicer to pull all of this into the same pandas dataframe for aggregation rather than maintain a separate accumulator.

Technically a separate accumulator would be faster, but I think there are other places to get more efficiencies ( like in the connector Telemetry class itself ).

src/snowflake/snowpark/modin/plugin/_internal/__init__.py
sfc-gh-jkew: This is so I could do @patch.object(ModinTelemetrySender, "_send_telemetry") reliably. One of the interesting things I realized was that patch actually patches the caller; not the underlying object - and so I was getting some non-deterministic patching behavior when I patched a specific method without a class. The use of the module was a way to expose this to the integration test more reliably.

sfc-gh-helmeleegy:
> > Are there dashboards that are currently dependent on this telemetry? If so, do we know how they will be impacted? I'm seeing each old metric replaced by three new metrics, right?
> 
> No dashboards rely on this directly - just a few widgets in our existing dashboards.

So we won't have any downstream issues now that the old metrics will be gone and new ones will be introduced?

sfc-gh-jkew:
> > > Are there dashboards that are currently dependent on this telemetry? If so, do we know how they will be impacted? I'm seeing each old metric replaced by three new metrics, right?
> > 
> > 
> > No dashboards rely on this directly - just a few widgets in our existing dashboards.
> 
> So we won't have any downstream issues now that the old metrics will be gone and new ones will be introduced?

The old metrics are not gone. They are still there when the engine is "Snowflake"; these metrics are a the modin layer though, so they cover all engines.

sfc-gh-helmeleegy:
> > > > Are there dashboards that are currently dependent on this telemetry? If so, do we know how they will be impacted? I'm seeing each old metric replaced by three new metrics, right?
> > > 
> > > 
> > > No dashboards rely on this directly - just a few widgets in our existing dashboards.
> > 
> > 
> > So we won't have any downstream issues now that the old metrics will be gone and new ones will be introduced?
> 
> The old metrics are not gone. They are still there when the engine is "Snowflake"; these metrics are a the modin layer though, so they cover all engines.

Great!

sfc-gh-helmeleegy (approved):

src/snowflake/snowpark/modin/plugin/_internal/__init__.py
sfc-gh-joshi: Can you add a code comment explaining this? Otherwise it might confuse people that this is the only public export from the module.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,108 @@ def __new__(
...
+    global _modin_metric_flush_interval
+
+    # Check whether the settings for the flush interval have changed, and
+    # if so, reset the timer.
+    current_flush_interval = SnowflakeModinTelemetryFlushInterval.get()
+    if _modin_metric_flush_interval != current_flush_interval:
+        _modin_metric_flush_interval = current_flush_interval
+        _next_modin_metric_flush = (
+            time.time() + SnowflakeModinTelemetryFlushInterval.get()
+        )
```
sfc-gh-joshi: This doesn't _really_ matter for the purposes of what we're trying to do with telemetry, but this branch makes it so the first flush attempt after `SnowflakeModinTelemetryFlushInterval` is updated won't succeed unless the new interval is very small, since `_next_modin_metric_flush` would always get updated to be `> time.time()`.

Instead, can we track a single `_last_modin_metric_flush` timestamp, and on every flush attempt check `time.time() > _last_modin_metric_flush + SnowflakeModinTelemetryFlushInterval.get()`? Again, I don't expect this to matter since we won't update the variable very often, but it makes the code a bit cleaner.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -644,28 +662,109 @@ def __new__(
...
+        session = snowflake.snowpark.session._get_active_session()
+        for row in summary_stats.iterrows():
+            for stat in summary_stats:
+                stat_specific_metric = f"{row[0]}.stat.{stat[1]}"
+
+                _send_modin_api_telemetry(
+                    session=session,
+                    event=stat_specific_metric,
+                    value=row[1][stat],
+                    aggregatable=stat == ("value", "count"),
```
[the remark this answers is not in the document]
sfc-gh-joshi: Is the `("value", "count")` label being generated from a `value_counts` call?

sfc-gh-joshi (approved):
Left a couple minor comments, otherwise LGTM.

src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
sfc-gh-jkew: Yeah. I'll change this. Originally I was not concerning myself with updates to the environment variable so this approach actually /avoided/ some complexity and binary operations; but now it's better to go back to the version you suggest.

sfc-gh-jkew: updated

PR #3614, opened 2025-07-31
Adding tabbing for nested queries

sfc-gh-daviwang:

   

   Fixes NO-SNOW

Tabbing inner queries. For example:


sfc-gh-jrose (approved):

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -892,7 +895,6 @@ def build_binary(
             left_schema_query = schema_value_statement(select_left.attributes)
             right_schema_query = schema_value_statement(select_right.attributes)
             schema_query = sql_generator(left_schema_query, right_schema_query)
-
```
sfc-gh-aalam: nit: we should keep this line here

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1966,7 +1966,7 @@ def get_line_numbers(
     current_child_uuid = None
 
     for line in sql_lines:
-        if line in comment_placeholders:
```
sfc-gh-aalam: we can do line = line.strip()

tests/integ/compiler/test_query_generator.py
```diff
@@ -254,7 +254,7 @@ def test_table_create_from_large_query_breakdown(session, plan_source_generator)
     assert Utils.normalize_sql(
         queries[PlanQueryType.QUERIES][0].sql
     ) == Utils.normalize_sql(
-        f"CREATE SCOPED TEMPORARY TABLE {table_name} AS SELECT * FROM ( select 1 as a, 2 as b )"
+        f"CREATE SCOPED TEMPORARY TABLE {table_name} AS SELECT *  FROM ( select 1 as a, 2 as b )"
```
sfc-gh-aalam: shouldn't normalize be able to handle this?

tests/unit/test_analyzer_util_suite.py
```diff
@@ -74,6 +77,90 @@ def test_format_uuid():
...
+        (
+            "basic_indentation",
+            "SELECT col1\nFROM table1",
+            "    SELECT col1\n    FROM table1",
+        ),
+        (
+            "single_quotes_with_newlines",
+            "SELECT 'line1\nline2' as col1\nFROM table1",
+            "    SELECT 'line1\nline2' as col1\n    FROM table1",
+        ),
```
sfc-gh-aalam: `CREATE temp TABLE "quote""andun\nquote""" AS \nSELECT 1 as A`

does this work?

tests/integ/compiler/test_query_generator.py
sfc-gh-daviwang: yeah not sure why i did this

tests/unit/test_analyzer_util_suite.py
sfc-gh-daviwang: i didn't realize you could also escape quotes this way, i added handling for this case.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -218,7 +218,10 @@ def format_uuid(uuid: Optional[str], with_new_line: bool = True) -> str:
     if not uuid:
         return EMPTY_STRING
     if with_new_line:
-        return f"\n{UUID_COMMENT.format(uuid)}\n"
+        if NEW_LINE == "":
+            return f"\n{UUID_COMMENT.format(uuid)}\n"
```
sfc-gh-aalam: curious to know, what values outside of `""`, `"\n"` can `NEW_LINE` take?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -310,6 +313,60 @@ def table_function_partition_spec(
...
+            result.append(NEW_LINE)
+            if not in_single_quote and not in_double_quote:
+                result.append(TAB)
+            i += 1
+            continue
+        char = child[i]
+        if char == "'" and not in_double_quote:
+            num_backslashes = 0
+            j = i - 1
+            while j >= 0 and child[j] == "\\":
```
sfc-gh-aalam: can you add a comment on what this is doing for example, here we are counting number of consecutive backslashes before a single quote

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -310,6 +313,60 @@ def table_function_partition_spec(
...
+                    i += 1
+                    continue
+                else:
+                    in_double_quote = not in_double_quote
+
+        result.append(char)
+        i += 1
+
+    final_result = "".join(result)
+    return TAB + final_result
```
sfc-gh-aalam: nit: we could initialize result = [TAB] in the beginning and just return "".join(result) here.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1497,7 +1555,7 @@ def unpivot_statement(
         + LEFT_PARENTHESIS
         + NEW_LINE
         + UUID
-        + child
+        + indent_child_query(child)
```
sfc-gh-aalam: In this PR, can you also add the change to add new lines between rows of each snowflake values statement?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-daviwang: oh whoops this was a remnant from when i was doing the other approach

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -310,6 +310,67 @@ def table_function_partition_spec(
...
+        char = child[i]
+        if char == "'" and not in_double_quote:
+            num_backslashes = 0
+            j = i - 1
+            while j >= 0 and child[j] == "\\":
+                num_backslashes += 1
+                j -= 1
+            # check for the case we escape a single quote using \
+            if num_backslashes % 2 == 0:
+                # check for the case we escape a single quote using ''
```
sfc-gh-aalam: are you sure we can escape single quote using a single quote? I think it only applied to double quotes. For example

```
select 1 as "My ''Identifier'";
```

Produces

Image

Can you double-check the rules here: https://docs.snowflake.com/en/sql-reference/identifiers-syntax

sfc-gh-daviwang: The identifiers syntax doesn't explicitly mention it, but I think you can escape single quotes inside single quotes. In your example, you use double quotes for the outer quotes b/c it's an identifier. This is an example where you can escape single quotes with single quotes:
```
df1 = session.create_dataframe([[1, "A"], [2, "B"]], schema=["id", "name"])
df1.write.save_as_table("TEST_TABLE", mode="overwrite", table_type="temp")
session.sql("INSERT INTO TEST_TABLE (id, name) VALUES (1, 'A''\nB'''), (2, 'B')").collect()
df2 = session.sql("SELECT * FROM TEST_TABLE").collect()
print(df2)
```
This gives the output:
```
[Row(ID=1, NAME='A'), Row(ID=2, NAME='B'), Row(ID=1, NAME="A'\nB'"), Row(ID=2, NAME='B')]
```

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -310,6 +310,67 @@ def table_function_partition_spec(
     )
 
 
+def indent_child_query(child: str) -> str:
```
sfc-gh-aling, line 313: To clarify, this is just a formatter that beautifies the SQL without changing any semantic right and it's mostly helpful for local debugging purpose?

this question might be too late but have we thought about using existing libraries to do formatting:
https://github.com/tobymao/sqlglot
https://github.com/andialbrecht/sqlparse

(add dependency or vendor their source code depending on their license)

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -310,6 +310,67 @@ def table_function_partition_spec(
...
+    if NEW_LINE == "":
+        return child
+    result = []
+    if child and child[0] != NEW_LINE:
+        result.append(TAB)
+    i = 0
+    in_single_quote = False
+    in_double_quote = False
+
+    while i < len(child):
```
sfc-gh-aling, line 323: this seems to be called on the critical path of our sql generation, or not? what's the performance overhead compared to when there is no such indent?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-daviwang: Yeah it shouldn't change the semantic meaning of any sql as we are only adding tabs and new line characters. I think there are two main reasons we don't want to use existing libraries:
- Some Snowflake specific syntax (eg. UDFs) are not properly formatted. 
- If anything in the existing library breaks, our library can also break, and we would like to be able to control that

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-daviwang: It is on the critical path, but it shouldn't add significant performance overhead. When we have child queries, we already do something like:
SELECT + LEFT_PARENTHESIS +  child_query + RIGHT_PARENTHESIS, which requires us to iterate through all the characters in the child_query. With this indent_child_query function, we just iterate through all the characters in the child_query twice.

PR #3615, opened 2025-07-31
SNOW-2237247: Update interface to enable developer improvement features

sfc-gh-aalam:

   

   Fixes SNOW-2237247

   Made the following changes

  i. Each feature must be explicitely enabled.
  ii. For the features that depend on AST, they enable AST for the session automatically instead of relies it to be done by the user.
  iii. Minor formatting updates to make the debug context look better.

### DIFF
--- a/src/snowflake/snowpark/context.py
+++ b/src/snowflake/snowpark/context.py
@@ -42,32 +42,46 @@ _enable_trace_sql_errors_to_dataframe = False
 
 def configure_development_features(
     *,
-    enable_dataframe_trace_on_error: bool = True,
-    enable_eager_schema_validation: bool = True,
-    enable_trace_sql_errors_to_dataframe: bool = True,
+    enable_eager_schema_validation: bool = False,
+    enable_dataframe_trace_on_error: bool = False,
+    enable_trace_sql_errors_to_dataframe: bool = False,
 ) -> None:
     """
     Configure development features for the session.
 
     Args:
-        enable_dataframe_trace_on_error: If True, upon failure, we will add most recent dataframe
-            operations to the error trace. This requires AST collection to be enabled in the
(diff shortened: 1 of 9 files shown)


src/snowflake/snowpark/context.py
```diff
@@ -42,33 +42,47 @@
...
     _enable_trace_sql_errors_to_dataframe = enable_trace_sql_errors_to_dataframe
 
+    if _enable_dataframe_trace_on_error or _enable_trace_sql_errors_to_dataframe:
+        try:
+            session = get_active_session()
+            if session is None:
+                _logger.warning(
+                    "No active session found. Please create a session first to enable development features.",
+                )
+                return
```
sfc-gh-aling: if there is no active session and I did `configure_development_features(enable_trace_sql_errors_to_dataframe =True)`, and later on I create session, would `enable_trace_sql_errors_to_dataframe` work? or do I need to call this `configure_development_features` again to enable it

src/snowflake/snowpark/context.py
sfc-gh-daviwang, line 1: Should we provide some api to enable and disable all features at once? I feel like it would be annoying to have to type out something like enable_eager_schema_validation = false as opposed to doing something like disable_development_features()

src/snowflake/snowpark/context.py
sfc-gh-aalam: Users will have to call `configure_development_features` to enable it again if session does not exist. Let me update the warning message to reflect this fact

src/snowflake/snowpark/context.py
sfc-gh-aalam: I think this is a good suggestion. I want to wait for some customer feedback to see if they find the current way of enable features too verbose

sfc-gh-aling (approved):

PR #3625, opened 2025-08-04
SNOW-2246517: Support YearMonthIntervalType

sfc-gh-fhe:

   

   Fixes SNOW-NNNNNNN


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -196,6 +200,8 @@ def convert_sf_to_sp_type(
     scale: int,
     internal_size: int,
     max_string_size: int,
+    start_field: int,
+    end_field: int,
```
sfc-gh-fhe: @sfc-gh-jdu @sfc-gh-aling Not sure if this is correct here, can either of you guys confirm?

sfc-gh-nkrishna: Is `start_field` and `end_field` here corresponding to years and months. For Day-Time this is going to need day, hour, minute, second fields. Perhaps naming this `year_field` and `month_field` and nothing that is is not needed for other types is better?

src/snowflake/snowpark/types.py
```diff
@@ -230,6 +230,64 @@ def _fill_ast(self, ast: proto.DataType) -> None:
...
+
+    pass
+
+
+class YearMonthIntervalType(AnsiIntervalType):
+    """YearMonthIntervalType data type. This maps to the INTERVAL YEAR TO MONTH data type in Snowflake.
+
+    Args:
+        start_field: The start field of the interval (0=YEAR, 1=MONTH)
+        end_field: The end field of the interval (0=YEAR, 1=MONTH)
```
sfc-gh-nkrishna, line 248: Related to earlier comment but for Year-Month start will always be year and end with always be month. Sometime in case of subtype `INTERVAL YEAR` month_field will be None and vise-versa. MONTH TO YEAR is invalid, so can we rename to `year_field` and `month_field`?

src/snowflake/snowpark/types.py
```diff
@@ -982,6 +1047,7 @@ def _fill_ast(self, ast: proto.DataType) -> None:
 
 _FIXED_VECTOR_PATTERN = re.compile(r"vector\(\s*(int|float)\s*,\s*(\d+)\s*\)")
 _FIXED_DECIMAL_PATTERN = re.compile(r"decimal\(\s*(\d+)\s*,\s*(\d+)\s*\)")
+_INTERVAL_YEARMONTH_PATTERN = re.compile(r"interval (year|month)( to (year|month))?")
```
sfc-gh-nkrishna, line 1056:
```suggestion
_INTERVAL_YEARMONTH_PATTERN = re.compile(r"interval (year|month)( to month)?")
```

We are always guaranteed that is it one of the following formats (haven't included precision or the actual value itself):

INTERVAL YEAR
INTERVAL MONTH
INTERVAL YEAR TO MONTH

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-fhe: So for DayTime, it would still be start_field and end_field. For DayTime these 2 values would be between 0-3 representing:
  DAY = 0
  HOUR = 1
  MINUTE = 2
  SECOND = 3
So our constructor would still be 
`    def __init__(self, startField: Optional[int] = None, endField: Optional[int] = None):`, we'd just allow more values for startField and endField if that makes sense. (This is exactly how Spark does it)

sfc-gh-nkrishna: Ah that makes a lot of sense, thanks!

src/snowflake/snowpark/types.py
sfc-gh-fhe: So INTERVAL MONTH TO YEAR is not supported? I noticed that I could do this in PySpark but admittedly didn't do much further testing

```>>> a = YearMonthIntervalType(1, 0)
>>> a
YearMonthIntervalType(1, 0)
>>> a.__repr__()
'YearMonthIntervalType(1, 0)'
>>> >>> a.simpleString()
'interval month to year'
```

sfc-gh-nkrishna: Yes if the leading field is a lesser unit that the trailing field (e.g. INTERVAL MONTH TO YEAR), that violates ANSI SQL standard so isn't supported in most engines (Oracle, Teradata, Snowflake, Trino etc.). Not sure of the PySpark behavior (let me look closer at it) but it is definitely invalid in Snowflake.

sfc-gh-fhe: Yeah, seems like you're able to create an interval as such but can't really use it for anything. It doesn't work using Spark SQL either.

```>>> spark.sql("SELECT INTERVAL '1-0' MONTH TO YEAR as ym").show()
  File "/Users/fhe/Desktop/test_spark/.venv/lib/python3.11/site-packages/pyspark/sql/session.py", line 1631, in sql
    return DataFrame(self._jsparkSession.sql(sqlQuery, litArgs), self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/fhe/Desktop/test_spark/.venv/lib/python3.11/site-packages/py4j/java_gateway.py", line 1322, in __call__
    return_value = get_return_value(
                   ^^^^^^^^^^^^^^^^^
  File "/Users/fhe/Desktop/test_spark/.venv/lib/python3.11/site-packages/pyspark/errors/exceptions/captured.py", line 185, in deco
    raise converted from None
pyspark.errors.exceptions.captured.ParseException: 
Intervals FROM month TO year are not supported.(line 1, pos 16)

== SQL ==
SELECT INTERVAL '1-0' MONTH TO YEAR as ym
----------------^^^
```

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-aling: I have two questions:

1. where does this information come from, does it come with the metadata?
I believe the existing five positional arguments come from metadata, if start_field/end_field also come from the metadata, then it can be added here.

2. another thing is you're adding two required positional argument to this api, it means all the callers will need to update to set the additional argument. does sever always return the start_field/end_field for whatever type it is?

sfc-gh-fhe: Just discussed with Naren and Naresh, it seems like we can get all the information from the scale field so I'll roll this change back

tests/integ/scala/test_datatype_suite.py
```diff
@@ -285,6 +289,7 @@ def test_verify_datatypes_reference(session):
             StructField("FLOAT", DoubleType()),
             StructField("DOUBLE", DoubleType()),
             StructField("DECIMAL", DecimalType(10, 2)),
+            StructField("YEARMONTHINTERVAL", YearMonthIntervalType()),
```
sfc-gh-fhe, line 291: This test is waiting on Python Connector support for YearMonthInterval column datatypes.
https://github.com/snowflakedb/snowflake-connector-python/pull/2467

src/snowflake/snowpark/types.py
```diff
@@ -230,6 +230,64 @@ def _fill_ast(self, ast: proto.DataType) -> None:
...
+class YearMonthIntervalType(AnsiIntervalType):
+    """YearMonthIntervalType data type. This maps to the INTERVAL YEAR TO MONTH data type in Snowflake.
+
+    Args:
+        start_field: The start field of the interval (0=YEAR, 1=MONTH)
+        end_field: The end field of the interval (0=YEAR, 1=MONTH)
+    """
+
+    YEAR = 0
+    MONTH = 1
```
sfc-gh-fhe: I will support precision in a different PR as this one is getting too large.

src/snowflake/snowpark/types.py
```diff
@@ -230,6 +230,64 @@ def _fill_ast(self, ast: proto.DataType) -> None:
...
+    ) -> None:
+        if start_field is None and end_field is None:
+            start_field = YearMonthIntervalType.YEAR
+            end_field = YearMonthIntervalType.MONTH
+        if start_field is not None and end_field is None:
+            end_field = start_field
+
+        fields = self._FIELD_NAMES.keys()
+        if start_field not in fields or end_field not in fields:
+            raise RuntimeError(f"interval {start_field} to {end_field} is invalid")
```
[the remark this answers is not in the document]
sfc-gh-fhe: For anyone reading this: I could make this change, although Spark does support creating YearMonthIntervalTypes with start_field > end_field (though you can't do much with it).

src/snowflake/snowpark/types.py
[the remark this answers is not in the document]
sfc-gh-fhe: I took this regex straight from PySpark. I can improve this but I feel that it is likely sufficient for now.

sfc-gh-heshah (approved):

CHANGELOG.md
```diff
@@ -142,6 +142,7 @@
 - Added a ttl cache to describe queries. Repeated queries in a 15 second interval will use the cached value rather than requery Snowflake.
 - Added a parameter `fetch_with_process` to `DataFrameReader.dbapi` (PrPr) to enable multiprocessing for parallel data fetching in local ingestion. By default, local ingestion uses multithreading. Multiprocessing may improve performance for CPU-bound tasks like Parquet file generation.
 - Added a new function `snowflake.snowpark.functions.model` that allows users to call methods of a model.
+- Added a new datatype `YearMonthIntervalType` that allows users to create intervals for datetime operations.
```
sfc-gh-aling: nit: this shall be moved to 1.38.0

tests/integ/scala/test_dataframe_suite.py
```diff
@@ -3227,3 +3228,56 @@ def test_limit(session):
 |     |     |
 -------------\n""".lstrip()
     )
+
+
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="FEAT: session.sql is not supported and interval is in PrPr",
+)
+def test_yearmonth_interval_type_dataframe(session):
```
sfc-gh-aling: can we add test which perform operation like date + interval as new date, want to check if it works well with operators

src/snowflake/snowpark/types.py
```diff
@@ -230,6 +230,64 @@ def _fill_ast(self, ast: proto.DataType) -> None:
...
+
+    YEAR = 0
+    MONTH = 1
+
+    _FIELD_NAMES = {YEAR: "year", MONTH: "month"}
+
+    _FIELD_VALUES_TO_NAMES = {value: key for key, value in _FIELD_NAMES.items()}
+
+    def __init__(
+        self, start_field: Optional[int] = None, end_field: Optional[int] = None
```
sfc-gh-aling, line 264: what does it mean start/end for interval type? I see your tests testing combination being (0,0) (0,1) (1,1), what about 1,0?

is this following spark design?

tests/integ/scala/test_dataframe_suite.py
```diff
@@ -3227,3 +3228,56 @@ def test_limit(session):
...
+    session.sql("alter session set feature_interval_types=enabled;").collect()
+
+    schema = StructType(
+        [
+            StructField("ID", LongType(), nullable=False),
+            StructField("YM_INTERVAL", YearMonthIntervalType(), nullable=False),
```
sfc-gh-aling, line 3241: can we test different combination of the input for YearMonthIntervalType?

src/snowflake/snowpark/types.py
sfc-gh-fhe: It follows spark design.

startField: This is the highest-order unit included in the interval. 
endField: This is the lowest-order unit included.

They can be either 0 for Year or 1 for Month. 
startField = 0, endField = 1 gives a year month interval
startField = 0, endField = 0 gives a year interval
startField = 1, endField = 1 gives a month interval
startField = 1, endField = 0 is invalid. You can't exactly do anything with it. The only thing is that Spark allows you to create the interval type itself but there isn't much else that can be done.

tests/integ/scala/test_dataframe_suite.py
sfc-gh-fhe: Good point, will do.

tests/integ/scala/test_dataframe_suite.py
sfc-gh-fhe: Not sure what you mean by this, could you clarify?

src/snowflake/snowpark/types.py
sfc-gh-jdu: Do we want to expose this constant to user? If yes, add docstring; if not, prefix with underscore

src/snowflake/snowpark/types.py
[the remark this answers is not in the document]
sfc-gh-jdu: Should we use ValueError instead of RuntimeError

src/snowflake/snowpark/types.py
```diff
@@ -230,6 +230,64 @@ def _fill_ast(self, ast: proto.DataType) -> None:
...
+            start_field = YearMonthIntervalType.YEAR
+            end_field = YearMonthIntervalType.MONTH
+        if start_field is not None and end_field is None:
+            end_field = start_field
+
+        fields = self._FIELD_NAMES.keys()
+        if start_field not in fields or end_field not in fields:
+            raise RuntimeError(f"interval {start_field} to {end_field} is invalid")
+
+        self.start_field = start_field
```
sfc-gh-jdu, line 280: same here, add docstring

src/snowflake/snowpark/types.py
```diff
@@ -1022,6 +1098,9 @@ def _parse_datatype_json_value(json_value: Union[dict, str]) -> DataType:
 #: The type hint for annotating Geometry data when registering UDFs.
 File = TypeVar("File")
 
+#: The type hint for annotating YearMonthInterval data when registering UDFs.
+YearMonthInterval = TypeVar("YearMonthInterval")
```
sfc-gh-jdu, line 1106: any udf test?

sfc-gh-jdu (commented):
Is this type supported in UDF/Stored Proc? If yes, we should add such tests; if not, we should mention it in docstring

src/snowflake/snowpark/types.py
sfc-gh-fhe: It isn't supported in UDF yet (serverside), I will make a follow up PR for it once it gets support

sfc-gh-fhe: cc @sfc-gh-nkrishna

src/snowflake/snowpark/types.py
sfc-gh-fhe: Do I need to add docstrings here? I have docstrings in the class definition specifying what start_field and end_field are

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -51,6 +52,24 @@ def str_to_sql(value: str) -> str:
...
 
+def str_to_sql_for_year_month_interval(
+    value: str, datatype: YearMonthIntervalType
+) -> str:
+    extracted_values = value.split(" ")[1]
+    start_field = datatype.start_field if datatype.start_field is not None else 0
+    end_field = datatype.end_field if datatype.end_field is not None else 1
+    if datatype.start_field == datatype.end_field:
+        extracted_values = extracted_values.split("-")
+        extracted_value = extracted_values[0]
```
sfc-gh-aalam, line 84: can you add a comment here to show what the extract values look like.

src/snowflake/snowpark/types.py
sfc-gh-jdu: ah ok, no worries

src/snowflake/snowpark/types.py
sfc-gh-jdu: sounds good

src/snowflake/snowpark/types.py
```diff
@@ -230,6 +230,67 @@ def _fill_ast(self, ast: proto.DataType) -> None:
         ast.time_type = True
 
 
+class AnsiIntervalType(_AtomicType):
+    """The interval type which conforms to the ANSI SQL standard."""
+
+    pass
```
sfc-gh-aalam: Is this also set as a public type intentionally?

tests/integ/scala/test_dataframe_suite.py
```diff
@@ -3227,3 +3228,56 @@ def test_limit(session):
...
 -------------\n""".lstrip()
     )
+
+
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="FEAT: session.sql is not supported and interval is in PrPr",
+)
+def test_yearmonth_interval_type_dataframe(session):
+    session.sql("alter session set feature_interval_types=enabled;").collect()
```
sfc-gh-jdu: you need to skip it in the stored proc test 

```
@pytest.mark.skipif(
    IS_IN_STORED_PROC,
    reason=...,
)
```
alter session is not supported in stored proc, same for all other places calling alter session

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -327,6 +361,8 @@ def schema_expression(data_type: DataType, is_nullable: bool) -> str:
             "'STAGE_FILE_URL', 'some_new_file.jpeg', 'SIZE', 123, 'ETAG', 'xxx', 'CONTENT_TYPE', 'image/jpeg', "
             "'LAST_MODIFIED', '2025-01-01'))"
         )
+    if isinstance(data_type, YearMonthIntervalType):
+        return "INTERVAL '1-0' YEAR TO MONTH"
```
sfc-gh-nkrishna, line 386: I don't follow what this is doing, can you maybe add a comment?

sfc-gh-nkrishna (approved):
Left one question, but overall LGTM. @sfc-gh-nkumar Do you want to take a pass as well?

tests/integ/scala/test_dataframe_suite.py
sfc-gh-aling: sorry I didn't made it clear, I mean the input parameters for YearMonthIntervalType.
YearMonthIntervalType(0,1)
YearMonthIntervalType(0,0)
YearMonthIntervalType(1,0)
YearMonthIntervalType(1,1)

I'm thinking from the angle of negative test case -- e.g., users specify YearMonthIntervalType(1,0), what error is raised

sfc-gh-aling (approved):

src/snowflake/snowpark/types.py
sfc-gh-fhe: Nope, will make it private. Thanks!

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
sfc-gh-fhe: I believe this method simply gives an example of what this datatype looks like schema-wise. Let's see what Afroz has to say

cc @sfc-gh-aalam

tests/integ/scala/test_dataframe_suite.py
sfc-gh-fhe: Got it. will test that. I intentionally made no error on the YearMonthIntervalType(1, 0) as PySpark doesn't error. But I'll test creating dataframes in this test.

sfc-gh-fhe: I've set it to error since PySpark in theory shouldn't support it either

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
sfc-gh-aalam: This is used to construct a dummy sql that has the same schema as original sql. This sql is used to build the sql for schema queries aka describe queries

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -51,6 +52,31 @@ def str_to_sql(value: str) -> str:
...
 
+def str_to_sql_for_year_month_interval(
+    value: str, datatype: YearMonthIntervalType
+) -> str:
+    # Extract interval value from strings like "INTERVAL 1-7 YEAR TO MONTH"
+    # Examples: "1-7" -> INTERVAL '1-7' YEAR TO MONTH, "1" -> INTERVAL '1' YEAR
+    # "1-7" -> INTERVAL '7' MONTH
+    parts = value.split(" ")
+    if len(parts) < 2:
+        raise ValueError(f"Invalid interval format: {value}")
```
sfc-gh-aalam, line 72: As a reader, this is still hard to understand. How do we expect value to look like. Should it be `INTERVAL YY-MM [YEAR TO MONTH | YEAR | MONTH]`

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -51,6 +52,31 @@ def str_to_sql(value: str) -> str:
     return f"'{sql_str}'"
 
 
+def str_to_sql_for_year_month_interval(
+    value: str, datatype: YearMonthIntervalType
+) -> str:
+    # Extract interval value from strings like "INTERVAL 1-7 YEAR TO MONTH"
+    # Examples: "1-7" -> INTERVAL '1-7' YEAR TO MONTH, "1" -> INTERVAL '1' YEAR
+    # "1-7" -> INTERVAL '7' MONTH
```
sfc-gh-aalam: how is "1-7" mapped to both `INTERVAL '1-7' YEAR TO MONTH` and `INTERVAL '7' MONTH`

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -51,6 +52,38 @@ def str_to_sql(value: str) -> str:
...
+) -> str:
+    """
+    Converts "INTERVAL YY-MM [YEAR TO MONTH | YEAR | MONTH]" to quoted format, optimizing based on datatype:
+    - Same start/end field: extracts specific value (YEAR or MONTH only)
+    - Different fields: uses full "YEAR TO MONTH" format
+
+    Examples:
+        "INTERVAL 1-2 YEAR TO MONTH" -> "INTERVAL '1-2' YEAR TO MONTH"
+        "INTERVAL 5-0 YEAR TO MONTH" + YearMonthIntervalType(0,0) -> "INTERVAL '5' YEAR"
+        "INTERVAL 0-3 YEAR TO MONTH" + YearMonthIntervalType(1,1) -> "INTERVAL '3' MONTH"
```
sfc-gh-aalam: This example can be a bit misleading since this function accepts string and datatype. It doesn't really handle additions, right. Maybe a better example would be some different kind of acceptable inputs.

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -51,6 +52,38 @@ def str_to_sql(value: str) -> str:
...
+    """
+    parts = value.split(" ")
+    if len(parts) < 2:
+        raise ValueError(f"Invalid interval format: {value}")
+    extracted_values = parts[1]
+    start_field = datatype.start_field if datatype.start_field is not None else 0
+    end_field = datatype.end_field if datatype.end_field is not None else 1
+    if datatype.start_field == datatype.end_field:
```
sfc-gh-aalam, line 82: what does start_field == end_field imply? Let's add that context here as a comment as well.

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -51,6 +52,38 @@ def str_to_sql(value: str) -> str:
...
+        extracted_values = extracted_values.split("-")
+        extracted_value = extracted_values[0]
+        if datatype.start_field == 1 and len(extracted_values) == 2:
+            # Extract the second value if we only have a MONTH interval.
+            extracted_value = extracted_values[1]
+        return (
+            f"INTERVAL '{extracted_value}' {datatype._FIELD_NAMES[start_field].upper()}"
+        )
+    else:
+        return f"INTERVAL '{extracted_values}' {datatype._FIELD_NAMES[start_field].upper()} TO {datatype._FIELD_NAMES[end_field].upper()}"
```
sfc-gh-aalam: nit: this else is redundant because the if branch always returns.

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
sfc-gh-fhe: Good point. The plus isn't the indicate addition, it was meant to represent the inputs. I'll update it to a comma so it's easier to read

sfc-gh-aalam: I think I misunderstood this example. Here left of `->` are inputs value, datatype

sfc-gh-fhe: Correct, will make adjustment for it to be clearer, thanks!

sfc-gh-aalam: another clarifying question: we strictly do not support input value `INTERVAL '1-2' YEAR TO MONTH` right? There must be no quotes around YY-MM

sfc-gh-fhe: In SQL it requires the single quotations around YY-MM

SELECT INTERVAL 1-11 YEAR TO MONTH; -> Syntax error: unexpected '1'. (line 25)
SELECT INTERVAL '1-11' YEAR TO MONTH; -> results in correct result

sfc-gh-aalam: That's not what I meant to ask. let me rephrase. 

Currently `str_to_sql_for_year_month_interval("INTERVAL '1-11' YEAR TO MONTH", datatype)` is invalid. This function expects no quotes around YY-MM. Is that the desired behavior?

sfc-gh-fhe: Got it. I think it makes more sense for us to simply passthrough when we get input as you suggested so I've added it to passthrough after detecting quotations.

src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
```diff
@@ -51,6 +52,45 @@ def str_to_sql(value: str) -> str:
...
+        "INTERVAL 5-0 YEAR TO MONTH", YearMonthIntervalType(0,0) -> "INTERVAL '5' YEAR"
+        "INTERVAL 0-3 YEAR TO MONTH", YearMonthIntervalType(1,1) -> "INTERVAL '3' MONTH"
+        "INTERVAL '1-2' YEAR TO MONTH", YearMonthIntervalType(0,1) -> "INTERVAL '1-2' YEAR TO MONTH" (passthrough)
+    """
+    parts = value.split(" ")
+    if len(parts) < 2:
+        raise ValueError(f"Invalid interval format: {value}")
+
+    if parts[1].startswith("'") and parts[1].endswith("'"):
+        return value
```
sfc-gh-aalam:
```suggestion
    if parts[1].startswith("'") and parts[1].endswith("'"):
        return value  # passthrough
```
nit

tests/integ/scala/test_datatype_suite.py
```diff
@@ -341,7 +350,11 @@ def test_verify_datatypes_reference_vector(session):
     "config.getoption('local_testing_mode', default=False)",
     reason="FEAT: function to_geography not supported",
 )
+@pytest.mark.skipif(
+    IS_IN_STORED_PROC, reason="Alter Session not supported in stored procedure."
+)
 def test_dtypes(session):
+    session.sql("alter session set feature_interval_types=enabled;").collect()
```
sfc-gh-aalam: I guess this can be enabled at session level instead of enabling and then disabling for each test.

https://github.com/snowflakedb/snowpark-python/blob/784c3d5c449d7ee94c0447a652e29bab69d80760/tests/integ/conftest.py#L272-L276

```
session = (
    Session.builder.configs(db_parameters)
    .config("local_testing", local_testing_mode)
    .config("session_parameters": {"feature_interval_types": "ENABLED"})
    .create()
)
```

tests/integ/test_dataframe.py
```diff
@@ -1689,6 +1690,116 @@ def test_create_dataframe_with_basic_data_types(session):
...
+    assert subtraction_result[0][0] == datetime.date(2021, 7, 15)
+    assert subtraction_result[1][0] == datetime.date(2022, 9, 30)
+
+    interval_arithmetic_df = session.sql(
+        """
+        SELECT
+            DATE '2023-01-01' + INTERVAL '1-2' YEAR TO MONTH as addition_result,
+            DATE '2023-12-31' - INTERVAL '0-6' YEAR TO MONTH as subtraction_result
+    """
+    )
```
sfc-gh-aalam, line 1741:
```suggestion
    interval_arithmetic_df = session.sql("""
        SELECT
            DATE '2023-01-01' + INTERVAL '1-2' YEAR TO MONTH as addition_result,
            DATE '2023-12-31' - INTERVAL '0-6' YEAR TO MONTH as subtraction_result
    """)
```
nit

tests/integ/test_dataframe.py
```diff
@@ -1689,6 +1690,116 @@ def test_create_dataframe_with_basic_data_types(session):
...
+    assert interval_subtraction_result[2][0] == "-3-06"
+    assert interval_subtraction_result[3][0] == "-3-00"
+
+    interval_sql_df = session.sql(
+        """
+        SELECT
+            INTERVAL '2-6' YEAR TO MONTH + INTERVAL '1-3' YEAR TO MONTH as interval_addition,
+            INTERVAL '3-0' YEAR TO MONTH - INTERVAL '1-6' YEAR TO MONTH as interval_subtraction,
+            INTERVAL '1-6' YEAR TO MONTH - INTERVAL '3-0' YEAR TO MONTH as interval_subtraction_2,
+    """
```
sfc-gh-aalam, line 1785: same

.gitignore
```diff
@@ -106,6 +106,7 @@ env/
 venv*/
 ENV/
 env.bak/
+snowpark-venv/
```
sfc-gh-aalam: remove

sfc-gh-aalam (approved):
Just clean-up some minor nits. Otherwise looks good

tests/integ/test_dataframe.py
sfc-gh-fhe: It's the autoformatter that isn't letting me fix these formats

PR #3628, opened 2025-08-05
SNOW-2179320: Support snowpark and cortex functions on pandas backend.

sfc-gh-mvashishtha:
Fixes SNOW-2179320


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -563,6 +564,27 @@ def wrap(self, *args, **kwargs):  # type: ignore
...
+    Check if the function passed to an apply-like method is a Snowpark or Cortex function.
+
+    Args:
+        func: The function to check. Could be a single function, a list-like
+              of functions, or a dict-like of functions.
+
+    Returns:
+        True if the function is a Snowpark or Cortex function, False otherwise.
+    """
+    if is_dict_like(func):
```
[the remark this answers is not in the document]
sfc-gh-mvashishtha: Which comment? This is not typescript. What is `x` here?

sfc-gh-joshi (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -567,6 +568,27 @@ def wrap(self, *args, **kwargs):  # type: ignore
...
+        return any(_apply_func_has_snowpark_function(func[key]) for key in func.keys())
+    if is_list_like(func):
+        return any(_apply_func_has_snowpark_function(each_item) for each_item in func)
+    return (
+        func in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY
+        or func in SUPPORTED_SNOWPARK_PYTHON_FUNCTIONS_IN_APPLY
```
sfc-gh-jkew, line 587: I don't think we should map the SUPPORTED_SNOWPARK_PYTHON_FUNCTIONS_IN_APPLY functions. All of those have numpy equivalents, and forcing the data to use snowflake just for something like floor ( which can be replaced with np.floor seems very weird.

I suppose the way users would use this; they would have to import snowpark.functions.floor first; but it still seems a little... odd.  I think with numpy ufuncs we translate to the snowflake function; I almost wonder if we should translate bidirectionally and choose the function which is most appropriate for the engine.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -567,6 +568,27 @@ def wrap(self, *args, **kwargs):  # type: ignore
...
+
+    Returns:
+        True if the function is a Snowpark or Cortex function, False otherwise.
+    """
+    if is_dict_like(func):
+        return any(_apply_func_has_snowpark_function(func[key]) for key in func.keys())
+    if is_list_like(func):
+        return any(_apply_func_has_snowpark_function(each_item) for each_item in func)
+    return (
+        func in SUPPORTED_SNOWFLAKE_CORTEX_FUNCTIONS_IN_APPLY
```
sfc-gh-jkew, line 586: Can we add snowflake.snowpark.functions.model to this list possibly? I'm not sure what it woudl take or if it is that wildly different.

sfc-gh-mvashishtha: I think `model` would work somewhat like other snowpark functions, but I think we should add support for it in a separate PR because we don't support `model` on either backend currently.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: > I almost wonder if we should translate bidirectionally and choose the function which is most appropriate for the engine.

That's a great point, and I agree! I tried changing this branch so that it would translate snowpark functions  to numpy equivalents for the pandas backend, but it turns out that modin [doesn't support](https://github.com/modin-project/modin/issues/7645) applying numpy functions! I added a TODO here to keep the data in pandas once the pandas backend can apply numpy functions.

sfc-gh-jkew (approved):

PR #3633, opened 2025-08-05
SNOW-2129018 async sproc

sfc-gh-mayliu:

   

   Fixes SNOW-2129018

Adds asynchronous stored procedure execution support. `Session.call()` and `StoredProcedure.__call__()` now accept a `block: bool = True` parameter. When set to False, these methods submit the stored procedure query and immediately return an AsyncJob handle, allowing non-blocking execution. A convenience method `Session.call_nowait()` is also introduced as shorthand for `Session.call(block=False)`.


sfc-gh-aalam (approved):
Looks great!

src/snowflake/snowpark/session.py
```diff
@@ -4513,3 +4576,53 @@ def _explain_query(self, query: str) -> Optional[str]:
...
+                results_cursor = self._conn.execute_async_and_notify_query_listener(
+                    query, _statement_params=statement_params
+                )
+                return AsyncJob(
+                    results_cursor["queryId"],
+                    query,
+                    self,
+                    _AsyncResultType.COUNT,
+                    log_on_exception=log_on_exception,
+                )
```
sfc-gh-aling: nit: I feel there's some redundancy in this part of code, for the async part, as far as I can see the only difference is _AsyncResultType.COUNT vs _AsyncResultType.ROW

e.g.,

the returning can just be the following for async job

```python
                results_cursor = self._conn.execute_async_and_notify_query_listener(
                    query, _statement_params=statement_params
                )
                return AsyncJob(
                    results_cursor["queryId"],
                    query,
                    self,
                    _AsyncResultType. ROW if is_return_table else _AsyncResultType.COUNT,
                    log_on_exception=log_on_exception,
                )
```

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -153,6 +154,15 @@ def __call__(
...
+                    query=query,
+                    is_return_table=self._is_return_table,
+                    block=block,
+                    statement_params=statement_params,
+                    ast_stmt=self._ast_stmt,
+                )
+
             if self._is_return_table:
                 qid = session._conn.execute_and_get_sfqid(
                     query, statement_params=statement_params
```
sfc-gh-aling: is the logic 166-170 encapsulated into `_execute_sproc_internal`? is it possible to do the following?

```python
ret = _execute_sproc_internal(...block=block, return_table=self._is_return_table)
if not block:
    return ret
else:
    ret._ast = sproc_expr
    return ret
```

tests/integ/test_stored_procedure.py
```diff
@@ -1282,6 +1283,297 @@ def select_star_decorator_sp(session_: Session, name: str) -> DataFrame:
...
+            input_types=[IntegerType()],
+            return_type=StringType(),
+            replace=True,
+        )
+
+        # 1a: session.call(block=False)
+        async_job_scalar = session.call(scalar_sp_name, 17, block=False)
+        assert isinstance(async_job_scalar, AsyncJob)
+        while not async_job_scalar.is_done():
+            time.sleep(1)
```
sfc-gh-aling: do we need the while loop here?
`AsyncJob.result()` blocks until it completes

(same comments to other sleep)

tests/integ/test_stored_procedure.py
```diff
@@ -1282,6 +1283,297 @@ def select_star_decorator_sp(session_: Session, name: str) -> DataFrame:
...
+        async_job_sproc_table = table_sproc(block=False)
+        while not async_job_sproc_table.is_done():
+            time.sleep(1)
+        result_sproc_table = async_job_sproc_table.result()
+        assert (
+            len(result_sproc_table) == 4
+        ), "StoredProcedure async call should return same data"
+        assert result_sproc_table[1]["TITLE"] == "Movie Two", "Should contain test data"
+
+        # ================ Test 3: None return sproc ================
```
sfc-gh-aling, line 1393: is this testing something other than `is_returning_table`?
i'm wondering if test 3 is needed here

tests/integ/test_stored_procedure.py
sfc-gh-mayliu: good catch, removed the `sleep` where applicable. For `test_async_stored_procedure_failure` though, I left `sleep` unchanged since `is_failed` is non-blocking and need completion to return failure

tests/integ/test_stored_procedure.py
sfc-gh-mayliu: IMHO, test 3 provides better coverage with `None` return type for sproc. Async sproc uses either `_AsyncResultType.ROW or _AsyncResultType.COUNT`, so test 3 covers the edge case when sproc doesn't return and verify that `AsyncJob.result()` can empty results well.

src/snowflake/snowpark/stored_procedure.py
```diff
@@ -153,22 +153,29 @@ def __call__(
...
+                statement_params=statement_params,
+                ast_stmt=self._ast_stmt,
+            )
+
+            if not block:
+                return res
+            else:
+                if self._is_return_table:
+                    res._ast = sproc_expr
+                    return res
```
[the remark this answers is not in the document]
sfc-gh-mayliu: AST-registered synchronous scalar-returning sproc should wait until `session._ast_batch.eval(self._ast_stmt)` instead of early returning here

sfc-gh-aling: do we need this "else" here?
I think lin 181-189 already handles it?
```suggestion
```

PR #3636, opened 2025-08-05
SNOW-2221407: Support xpath functions

sfc-gh-jdu:

   

   Fixes SNOW-2221407

    Similar to how to support reading XML data with row tag using UDTF, now we support xpath functions using UDF.


src/snowflake/snowpark/session.py
```diff
@@ -745,7 +750,17 @@ def __init__(
         # of SnowflakePlan or Selectable objects
         self._plan_lock = create_rlock(self._conn._thread_safe_session_enabled)
 
+        # this lock is used to protect XPath UDF cache to avoid race conditions during
+        # UDF registration and retrieval
+        self._xpath_udf_cache_lock = create_rlock(
+            self._conn._thread_safe_session_enabled
+        )
```
sfc-gh-jrose, line 757: I wonder if all udf registration should be protected by a lock.

sfc-gh-jrose (approved):
Do you think it would be worthwhile to have a sproc test so that this gets validated before hitting Snowfort?

src/snowflake/snowpark/functions.py
```diff
@@ -11764,6 +11764,170 @@ def randn(
     return ans
 
 
+@publicapi
```
sfc-gh-aling, line 11767: I'm curious did you use llm to generate them?

sfc-gh-jdu: Yeah, most are all from AI, as I completed  session._get_or_register_xpath_udf and this part is easy. But AI still failed to write the correct AST code actually.

sfc-gh-jdu:
> Do you think it would be worthwhile to have a sproc test so that this gets validated before hitting Snowfort?

Yes, I'll test it with sproc. Note that a new test file is not automatically added to precommit sproc test, and we need to manually add it in monorepo.

src/snowflake/snowpark/session.py
sfc-gh-jdu: I don't think so? Because we only care about creating and getting xpath udf in the cache, which might have race condition?

PR #3640, opened 2025-08-06
SNOW-2252162: Add tests for relaxed pandas regressions.

sfc-gh-mvashishtha:
Add tests for relaxed pandas regressions SNOW-2250244 and /SNOW-2252101, so that when we add relaxed pandas again, we don't reintroduce those regressions.

I confirmed that these tests both fail as expected on commit 5503abb70521243208695ee8b6c46014f35c64b6, the last commit before commit bb3d445ed269f936f36dc011521cdb56e53d9491 was reverted.

Fixes SNOW-2252162.

### DIFF
--- a/tests/integ/modin/binary/test_binary_op.py
+++ b/tests/integ/modin/binary/test_binary_op.py
@@ -6,6 +6,7 @@ import datetime
 import math
 import operator
 import random
+from typing import Generator
 
 import modin.pandas as pd
 import numpy as np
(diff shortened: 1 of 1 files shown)


sfc-gh-jkew (approved):

tests/integ/modin/binary/test_binary_op.py
```diff
@@ -548,6 +550,55 @@ def test_binary_add_between_series(native_df):
...
+    pandas_df = request.param
+    table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
+    session.write_pandas(
+        df=pandas_df,
+        table_name=table_name,
+        overwrite=True,
+        table_type="temp",
+        auto_create_table=True,
+    )
+    yield (pd.read_snowflake(table_name), pandas_df)
```
sfc-gh-nkrishna, line 566: Just curious, why the `yield` statement here?

sfc-gh-nkrishna (approved):

tests/integ/modin/binary/test_binary_op.py
sfc-gh-mvashishtha: We yield the table and then [clean it up](https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended) after the yield.

PR #3653, opened 2025-08-12
SNOW-2262562: Return empty DF instead of None for pd.explain_switch(simple=True)

sfc-gh-joshi:

   

   Fixes SNOW-2262562

`pd.explain_switch(simple=True)` would return `None` if no operations were performed, which looks weird in an interactive environment because nothing is printed.

The bug cited in the ticket occurs because the previous implementation of `explain_switch` tripped an uncommon code path in telemetry that checked if a function returned None and if its first argument was a Snowpark pandas DF/Series. This PR adds a defensive length check to telemetry to avoid the IndexError, and also changes `explain_switch(simple=True)` to always return an empty DF instead of None. When `simple=False`, it already returned an empty frame, though column labels are left unset.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
@@ -420,6 +420,7 @@ def _telemetry_helper(
     elif (
         result is None
         and func.__name__ not in ["to_snowflake", "to_iceberg"]
+        and len(args) > 0
         and is_snowpark_pandas_dataframe_or_series_type(args[0])
     ):
         args[0]._query_compiler.snowpark_pandas_api_calls = (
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/modin/plugin/_internal/telemetry.py
```diff
@@ -420,6 +420,7 @@ def _telemetry_helper(
     elif (
         result is None
         and func.__name__ not in ["to_snowflake", "to_iceberg"]
+        and len(args) > 0
```
sfc-gh-jkew, line 423: Why do we add this?

sfc-gh-joshi: Before this PR, `explain_switch` would return `None`, and check this branch condition that checks for in-place lazy APIs that returned `None`. The line right below this checks if `args[0]` (normally `self` for a method) is a Snowpark pandas Series/DF, but since `explain_switch` is 0-arity, this raised the `IndexError` seen in the ticket.

Since `explain_switch` no longer returns `None`, this branch is no longer run, but I added the defensive length check anyway in case we add future APIs that would hit it.

sfc-gh-joshi:
> Should we add a CHANGELOG note? Otherwise LGTM.

Added.

sfc-gh-mvashishtha (approved):

PR #3654, opened 2025-08-12
SNOW-2250267: Prevent switching within extension Index constructor

sfc-gh-joshi:

   

   Fixes SNOW-2250267

`pd.to_datetime`/`pd.to_timedelta` have code paths to create `pd.Index` objects (defined as extensions in Snowpark pandas) when their arguments are non-pandas objects. Internally, the `pd.Index` constructor may create a modin DF/Series from this data, which then may incorrectly use a `NativeQueryCompiler` when a `SnowflakeQueryCompiler` one is expected instead. This PR forces the extension `pd.Index` constructor to always use the Snowflake backend, as any switching decisions should have been made before the constructor was called.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/extensions/index.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/index.py
@@ -28,6 +28,7 @@ from functools import cached_property
 from typing import Any, Callable, Hashable, Iterable, Iterator, Literal
 
 import modin
+from modin.config import context as config_context
 import numpy as np
 import numpy.typing as npt
 import pandas as native_pd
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/modin/plugin/extensions/index.py
```diff
@@ -172,32 +173,38 @@ def _init_query_compiler(
...
-                index=cls._NATIVE_INDEX_TYPE(data=data, **kwargs)
-            )._query_compiler
-
-        if len(query_compiler.columns):
-            query_compiler = query_compiler.drop(columns=query_compiler.columns)
-        return query_compiler
+        # Keep the backend as Snowflake within this method to ensure we always return an SFQC object.
+        # In hybrid mode, the DataFrame constructor may inappropriately create a native QC object
+        # that causes difficult-to-find errors further down the line.
+        with config_context(Backend="Snowflake", AutoSwitchBackend=False):
```
sfc-gh-jkew, line 179: This is interesting. I think it's the first use of the config_context w/in the internals of snowpandas.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -397,6 +398,14 @@ def test_unimplemented_autoswitches(class_name, method_name, f_args):
             assert snow_result == pandas_result
 
 
+@sql_count_checker(query_count=0)
+def test_to_datetime():
+    assert Backend.get() == "Snowflake"
```
sfc-gh-jkew, line 403: What would happen in the following? Would this snippet of code pass?
```
with config_context(Backend="Pandas"):
        result = pd.to_datetime([3, 4, 5], unit="Y")
        assert isinstance(result, modin.pandas.DatetimeIndex)
```

https://github.com/modin-project/modin/blob/cf9590b28aef82241d49b75c81953f0a61f67acc/modin/pandas/__init__.py#L55

sfc-gh-mvashishtha (approved):

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-joshi: It would resolve to the re-exported `<class 'pandas.core.indexes.datetimes.DatetimeIndex'>` type. The global namespace is sensitive to the active backend, so it only would use the extension when Snowflake is the active backend. I believe we made an exception for stuff like `read_snowflake` that's only defined in one backend but not the others.

PR #3656, opened 2025-08-13
SNOW-2265266: Add interval_year_month_from_parts function

sfc-gh-fhe:

   

   Fixes SNOW-2265266

   - [X] I acknowledge that I have ensured my changes to be thread-safe. Follow the link for more information: [Thread-safe Developer Guidelines](https://github.com/snowflakedb/snowpark-python/blob/main/CONTRIBUTING.md#thread-safe-development)

### DIFF
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -10915,7 +10915,7 @@ def make_interval(
 
 @private_preview(
     version="1.38.0",
-    extra_doc_string="The interval_year_month_from_parts function creates a column of type YearMonthIntervalType. This type is currently in private preview and needs to be enabled by setting parameter `FEATURE_INTERVAL_TYPES` to `enabled`.",
+    extra_doc_string="Type YearMonthIntervalType is currently in private preview and needs to be enabled by setting parameter `FEATURE_INTERVAL_TYPES` to `ENABLED`.",
 )
 @publicapi
 def interval_year_month_from_parts(
@@ -10950,23 +10950,42 @@ def interval_year_month_from_parts(
         <BLANKLINE>
 
     """
+    ast = None
+    if _emit_ast:
+        args = []
+        if years is not None:
+            args.append(years)
+        if months is not None:
+            args.append(months)
+        ast = build_function_expr("interval_year_month_from_parts", args)
+
     years_col = (
-        lit(0)
+        lit(0, _emit_ast=False)
         if years is None
         else _to_col_if_str(years, "interval_year_month_from_parts")
     )
     months_col = (
-        lit(0)
+        lit(0, _emit_ast=False)
         if months is None
         else _to_col_if_str(months, "interval_year_month_from_parts")
     )
 
-    total_months = years_col * lit(12) + months_col
+    total_months = years_col * lit(12, _emit_ast=False) + months_col
 
-    normalized_years = cast(cast(floor(abs(total_months) / lit(12)), "int"), "str")
-    normalized_months = cast(cast(floor(abs(total_months) % lit(12)), "int"), "str")
-    sign_prefix = iff(total_months < lit(0), lit("-"), lit(""))
-    interval_string = concat(sign_prefix, normalized_years, lit("-"), normalized_months)
+    normalized_years = cast(
+        cast(floor(abs(total_months) / lit(12, _emit_ast=False)), "int"), "str"
+    )
+    normalized_months = cast(
+        cast(floor(abs(total_months) % lit(12, _emit_ast=False)), "int"), "str"
+    )
+    sign_prefix = iff(
+        total_months < lit(0, _emit_ast=False),
+        lit("-", _emit_ast=False),
+        lit("", _emit_ast=False),
+    )
+    interval_string = concat(
+        sign_prefix, normalized_years, lit("-", _emit_ast=False), normalized_months
+    )
 
     def get_col_name(col):
         if isinstance(col._expr1, Literal):
@@ -10976,7 +10995,9 @@ def interval_year_month_from_parts(
 
     alias_name = f"interval_year_month_from_parts({get_col_name(years_col)}, {get_col_name(months_col)})"
 
-    return cast(interval_string, "INTERVAL YEAR TO MONTH").alias(alias_name)
+    res = cast(interval_string, "INTERVAL YEAR TO MONTH").alias(alias_name)
+    res._ast = ast
+    return res
 
 
 @publicapi
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -10915,7 +10915,7 @@ def make_interval(
 
 @private_preview(
     version="1.38.0",
-    extra_doc_string="The interval_year_month_from_parts function creates a column of type YearMonthIntervalType. This type is currently in PrPr and needs to be enabled by setting this parameter to `enabled` feature_interval_types.",
+    extra_doc_string="The interval_year_month_from_parts function creates a column of type YearMonthIntervalType. This type is currently in private preview and needs to be enabled by setting parameter `FEATURE_INTERVAL_TYPES` to `enabled`.",
 )
 @publicapi
 def interval_year_month_from_parts(
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -10964,32 +10964,26 @@ def interval_year_month_from_parts(
         ast = build_function_expr("interval_year_month_from_parts", args)
 
     years_col = (
-        lit(0, _emit_ast=False)
+        lit(0)
         if years is None
         else _to_col_if_str(years, "interval_year_month_from_parts")
     )
     months_col = (
-        lit(0, _emit_ast=False)
+        lit(0)
         if months is None
         else _to_col_if_str(months, "interval_year_month_from_parts")
     )
 
-    total_months = years_col * lit(12, _emit_ast=False) + months_col
+    total_months = years_col * lit(12) + months_col
 
-    normalized_years = cast(
-        cast(floor(abs(total_months) / lit(12, _emit_ast=False)), "int"), "str"
-    )
-    normalized_months = cast(
-        cast(floor(abs(total_months) % lit(12, _emit_ast=False)), "int"), "str"
-    )
+    normalized_years = cast(cast(floor(abs(total_months) / lit(12)), "int"), "str")
+    normalized_months = cast(cast(floor(abs(total_months) % lit(12)), "int"), "str")
     sign_prefix = iff(
-        total_months < lit(0, _emit_ast=False),
-        lit("-", _emit_ast=False),
-        lit("", _emit_ast=False),
+        total_months < lit(0),
+        lit("-"),
+        lit(""),
     )
-    interval_string = concat(
-        sign_prefix, normalized_years, lit("-", _emit_ast=False), normalized_months
-    )
+    interval_string = concat(sign_prefix, normalized_years, lit("-"), normalized_months)
 
     def get_col_name(col):
         if isinstance(col._expr1, Literal):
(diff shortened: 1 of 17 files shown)


src/snowflake/snowpark/functions.py
```diff
@@ -10987,6 +10987,79 @@ def make_interval(
...
+    Returns:
+        A Column representing a year-month interval
+
+    Example::
+
+        >>> from snowflake.snowpark.functions import make_ym_interval
+        >>> df = session.create_dataframe([[1, 2]], ["years", "months"])
+        >>> df.select(make_ym_interval(col("years"), col("months")).alias("interval")).show()
+        >>> # make_ym_interval(2014, 12) returns INTERVAL '2015-0' YEAR TO MONTH
+        >>> # make_ym_interval(1, 14) returns INTERVAL '2-2' YEAR TO MONTH
```
sfc-gh-aalam: shouldn't this include the result of show?

sfc-gh-aalam (commented):
+ add a change in functions.rst

src/snowflake/snowpark/functions.py
sfc-gh-fhe: Will do

src/snowflake/snowpark/functions.py
```diff
@@ -10987,6 +10987,88 @@ def make_interval(
...
+        if total_months >= 0:
+            normalized_years, normalized_months = divmod(total_months, 12)
+        else:
+            normalized_years = total_months * -1 // 12
+            normalized_months = (12 * normalized_years + total_months) * -1
+
+        return sql_expr(
+            f"""INTERVAL '{'-' if total_months < 0 else ''}{normalized_years}-{normalized_months}' YEAR TO MONTH""",
+            _emit_ast=_emit_ast,
+        )
```
sfc-gh-fhe: This case isn't valid in PySpark, not sure if we want to support it or not in Snowpark

src/snowflake/snowpark/functions.py
```diff
@@ -10987,6 +10987,84 @@ def make_interval(
...
+        total_months >= lit(0),
+        cast(floor(total_months / lit(12)), "int"),
+        cast(floor((total_months * lit(-1)) / lit(12)), "int"),
+    )
+
+    normalized_months = iff(
+        total_months >= lit(0),
+        cast(total_months % lit(12), "int"),
+        cast((lit(12) * normalized_years + total_months) * lit(-1), "int"),
+    )
```
sfc-gh-aalam: Ideally shouldn't this logic be moved to server-side? Is it possible to move this logic there?

sfc-gh-fhe: Yes, ideally it would be server-side. I think @sfc-gh-nkrishna mentioned it might be a post-ga priority. Although we had anticipated that it would be a little bit easier to implement than this.

@sfc-gh-nkrishna @sfc-gh-nkumar wdyt? The conversions are needed as we're able to pass any value into `make_ym_interval` but `interval x-x year to month` have certain value ranges.

`260012 (2200U): Year-Month Interval '1-13' is invalid, required that 0 <= MONTH <= 11`

src/snowflake/snowpark/functions.py
```diff
@@ -10987,6 +10987,84 @@ def make_interval(
     return res
 
 
+@publicapi
+def make_ym_interval(
```
sfc-gh-nkrishna: Can we call this `INTERVAL_YEAR_MONTH_FROM_PARTS(<year>, <month>)` to match Snowflake server-side functions `DATE_FROM_PARTS`, `TIME_FROM_PARTS`, etc.: https://docs.snowflake.com/en/sql-reference/functions/date_from_parts.

src/snowflake/snowpark/functions.py
sfc-gh-nkrishna: Ideally it would be server-side (request from the SnowConvert team as well), but we settled on it as a post-PuPr priority. My vote is for now to keep it client side just for time constraint reasons and I don't think there will be a large amount of work that gets thrown away. Behavior-wise, this is what we want my only comment is re: the function signature.

sfc-gh-aalam: I see. Let's also tag this function with `@private_preview(version="1.38.0")` to protect us from any surprises that may come after we move to server side

src/snowflake/snowpark/functions.py
```diff
@@ -10987,6 +10987,84 @@ def make_interval(
...
+        lit(0) if months is None else _to_col_if_str(months, "make_ym_interval")
+    )
+
+    total_months = years_col * lit(12) + months_col
+
+    normalized_years = iff(
+        total_months >= lit(0),
+        cast(floor(total_months / lit(12)), "int"),
+        cast(floor((total_months * lit(-1)) / lit(12)), "int"),
+    )
```
sfc-gh-nkumar: This logic can be simplified using abs. 
```
normalized_years = cast(cast(floor(abs(total_months) / lit(12)), "int"), "str")
normalized_months = cast(cast(floor(abs(total_months) % lit(12)), "int"), "str")
sign_prefix = iff(total_months < lit(0), lit("-"), lit(""))
interval_string = concat(sign_prefix, normalized_years, lit("-"), normalized_months)
```

tests/integ/scala/test_function_suite.py
```diff
@@ -5613,3 +5615,132 @@ def test_any_value(session):
...
+    "config.getoption('local_testing_mode', default=False)",
+    reason="FEAT: Alter Session not supported in local testing",
+)
+@pytest.mark.skipif(
+    IS_IN_STORED_PROC, reason="Alter Session not supported in stored procedure."
+)
+def test_make_ym_interval(session):
+    session.sql("alter session set feature_interval_types=enabled;").collect()
+
+    df = session.create_dataframe(
```
sfc-gh-nkumar, line 5625: nit: can you please add a column to df with expected values. It's easier to review/verify when they are side by side.

tests/integ/scala/test_function_suite.py
```diff
@@ -5613,3 +5615,132 @@ def test_any_value(session):
...
+        [
+            (0, 0),
```
sfc-gh-nkumar: Can you please also include some cases with mixed signs. year > 0 and month < 0 and vice-a-versa.

src/snowflake/snowpark/functions.py
sfc-gh-fhe: yup, good point. I simplified it on make_dt_interval but forgot to mirror the changes here. Thanks!

tests/integ/scala/test_function_suite.py
sfc-gh-fhe: done

sfc-gh-nkumar (approved):

src/snowflake/snowpark/functions.py
```diff
@@ -10987,6 +10988,72 @@ def make_interval(
     return res
 
 
+@private_preview(
+    version="1.38.0",
+    extra_doc_string="The interval_year_month_from_parts function creates a column of type YearMonthIntervalType. This type is currently in PrPr and needs to be enabled by setting this parameter to `enabled` feature_interval_types.",
```
sfc-gh-aalam:
```suggestion
    extra_doc_string="The interval_year_month_from_parts function creates a column of type YearMonthIntervalType. This type is currently in private preview and needs to be enabled by setting parameter `FEATURE_INTERVAL_TYPES` to `enabled`.",
```

src/snowflake/snowpark/functions.py
```diff
@@ -10912,6 +10913,72 @@ def make_interval(
     return res
 
 
+@private_preview(
+    version="1.38.0",
+    extra_doc_string="The interval_year_month_from_parts function creates a column of type YearMonthIntervalType. This type is currently in private preview and needs to be enabled by setting parameter `FEATURE_INTERVAL_TYPES` to `enabled`.",
+)
```
sfc-gh-aalam:
```suggestion
    extra_doc_string="Type YearMonthIntervalType is currently in private preview and needs to be enabled by setting parameter `FEATURE_INTERVAL_TYPES` to `ENABLED`.",
)
```

src/snowflake/snowpark/functions.py
```diff
@@ -10912,6 +10913,72 @@ def make_interval(
...
+        if months is None
+        else _to_col_if_str(months, "interval_year_month_from_parts")
+    )
+
+    total_months = years_col * lit(12) + months_col
+
+    normalized_years = cast(cast(floor(abs(total_months) / lit(12)), "int"), "str")
+    normalized_months = cast(cast(floor(abs(total_months) % lit(12)), "int"), "str")
+    sign_prefix = iff(total_months < lit(0), lit("-"), lit(""))
+    interval_string = concat(sign_prefix, normalized_years, lit("-"), normalized_months)
```
sfc-gh-aalam, line 10984: need to build ast here as well.

src/snowflake/snowpark/functions.py
```diff
@@ -10912,6 +10913,93 @@ def make_interval(
...
+
+    """
+    ast = None
+    if _emit_ast:
+        args = []
+        if years is not None:
+            args.append(years)
+        if months is not None:
+            args.append(months)
+        ast = build_function_expr("interval_year_month_from_parts", args)
```
sfc-gh-fhe, line 10962: @sfc-gh-heshah not sure if this is the correct way for ast

src/snowflake/snowpark/functions.py
```diff
@@ -10912,6 +10913,93 @@ def make_interval(
...
+        if isinstance(col._expr1, Literal):
+            return str(col._expr1.value)
+        else:
+            return col._expression.name
+
+    alias_name = f"interval_year_month_from_parts({get_col_name(years_col)}, {get_col_name(months_col)})"
+
+    res = cast(interval_string, "INTERVAL YEAR TO MONTH").alias(alias_name)
+    res._ast = ast
+    return res
```
sfc-gh-fhe, line 10996: @sfc-gh-heshah  +1 here

tests/integ/scala/test_function_suite.py
```diff
@@ -5613,3 +5615,106 @@ def test_any_value(session):
...
+            (-2, -11, "-2-11"),
+            (-1, -12, "-2-00"),
+            (0, -13, "-1-01"),
+            (-1, -24, "-3-00"),
+            (10, 5, "+10-05"),
+            (-10, -5, "-10-05"),
+            (999, 999, "+1082-03"),
+            (-999, -999, "-1082-03"),
+            (0, 999, "+83-03"),
+            (0, -999, "-83-03"),
```
sfc-gh-nkrishna, line 5647: Can we test the max and min as well? `+999999999-11` and `-999999999-11`

src/snowflake/snowpark/functions.py
sfc-gh-heshah: This is correct.

src/snowflake/snowpark/functions.py
sfc-gh-heshah: Yes, this is correct.

src/snowflake/snowpark/functions.py
```diff
@@ -10912,6 +10913,97 @@ def make_interval(
...
+    months_col = (
+        lit(0, _emit_ast=False)
+        if months is None
+        else _to_col_if_str(months, "interval_year_month_from_parts")
+    )
+
+    total_months = years_col * lit(12, _emit_ast=False) + months_col
+
+    normalized_years = cast(
+        cast(floor(abs(total_months) / lit(12, _emit_ast=False)), "int"), "str"
```
sfc-gh-heshah: Lots of uses of `_emit_ast=False` are missing here, and below. Since you are replacing the final result's AST manually, its probably better to leave out all uses of `_emit_ast=False`.

sfc-gh-heshah (commented):
Please add an AST test for the new function to `tests/ast/data/functions2.test`

src/snowflake/snowpark/functions.py
sfc-gh-fhe: Got it, will do. Thanks! Will also add the test.

PR #3673, opened 2025-08-18
DOC-8461 Reproducibility warning for random_split

sfc-gh-stramer:

   

   Fixes DOC-8461

   - [x] I am modifying a docstring

   It turns out that the implementation of `DataFrame.random_split()` relies on the behavior of [SnowSQL RANDOM](https://docs.snowflake.com/en/sql-reference/functions/random) under the hood. This method of generating random partitions means that using a consistent `seed` value doesn't guarantee reproducibility, which might otherwise be expected.
   
   Note that this is not (currently) considered a bug in the random seed behavior, as SnowSQL is executed in a distributed environment and is subject to the restrictions of irreproducible seeding under multithreading/distributed conditions.


sfc-gh-jrose (approved):

src/snowflake/snowpark/dataframe.py
```diff
@@ -6245,6 +6245,8 @@ def random_split(
                 weight is specified, the returned DataFrame list only includes
                 the current DataFrame.
             seed: The seed for sampling.
+
+                .. caution:: Reusing a seed value doesn't guarantee reproducible results.
```
sfc-gh-aalam: There is a way to get random_split seed work with reproducible results but it is parameter protected right now. You can set `session.conf.set("use_simplified_query_generation", True)` to enable the improved query generation. Can you also update the docstring to reflect that.

sfc-gh-stramer: Does using this parameter have any other consequences, and is there reference documentation for session configuration parameters?

**EDIT:** Did a little digging on this one, it's a discussion for another day. I'll see about filing an internal issue. For now, I'll amend the notes to include this remark on reproducible seeds.

sfc-gh-mayliu (approved):

PR #3674, opened 2025-08-19
SNOW-1791191: time travel support v2

sfc-gh-mayliu:

   

   Fixes SNOW-1791191

Support time travel:
```
# session.py::table()
df = session.table("my_table", time_travel_mode="before", offset=-3600)
df = session.table("my_table", time_travel_mode="at",timestamp=datetime.now())

# dataframe_reader.py::table()
df = session.read.table("my_table", time_travel_mode="before", offset=-3600)

# dataframe_reader.py::option()
df = session.read.option("time-travel-mode", "at/before").option("timestamp", datetime.now()).table("my_table")

# timestamp: str that specifies timezone -> Respect user's choice to invoke timestamp_type; otherwise no cast
df = session.read.option("time-travel-mode", "before").option("timestamp", '2025-08-25T10:00:00-13:00').option("timestamp_type", "LTZ").table("my_table") # cast to LTZ

# timestamp: datetime with tzinfo without specifying timestamp_type -> Default to TO_TIMESTAMP_TZ
ts_with_tz = datetime(2023, 6, 15, 14, 30, 0, tzinfo=timezone(timedelta(hours=-8)))  # -08:00 (PST)
df = session.read.option("time-travel-mode", "before").option("timestamp", ts_with_tz).table("my_table")

# timestamp: datetime without tzinfo nor timestamp_type -> no cast
ts_with_tz = datetime(2023, 6, 15, 14, 30, 0)
df = session.table("my_table", time-travel-mode="at", timestamp=ts_with_tz) # no cast

# dataframe_reader.py::option() with as-of-timestamp
df = session.read.option("as-of-timestamp", datetime.now()).table("my_table")

# PySpark's as-of-timestamp is NOT allowed to be used with before
df = session.read.option("time-travel-mode", "before").option("as-of-timestamp", datetime.now()).table("my_table")
```


src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -588,20 +589,41 @@ def __deepcopy__(self, memodict={}) -> "SelectableEntity":  # noqa: B006
 
     @property
     def sql_query(self) -> str:
-        return f"{analyzer_utils.SELECT}{analyzer_utils.STAR}{analyzer_utils.FROM}{self.entity.name}"
+        table_reference = self.entity.name
+
+        if self.entity.time_travel_config is not None:
+            table_reference += generate_time_travel_sql_clause(
+                self.entity.time_travel_config
+            )
```
sfc-gh-aling: can this be calculated in the constructor so that we avoid recompute each time we call the property?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1925,141 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+    timezone: str = "NTZ"
+    stream: Optional[str] = None
+
+
+def validate_and_normalize_time_travel_params(
+    time_travel_mode: Optional[str] = None,
+    statement: Optional[str] = None,
+    offset: Optional[int] = None,
+    timestamp: Optional[Union[str, datetime.datetime]] = None,
+    timezone: Optional[str] = "NTZ",
```
sfc-gh-aling: I'm wondering if we can also allow users to use `types.TimestampTimeZone` to specify timezone

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1925,141 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+        try:
+            parsed_timestamp = datetime.datetime.strptime(
+                timestamp_str, "%Y-%m-%d %H:%M:%S.%f"
+            )
+        except ValueError:
+            # Fall back to the original format without fractional seconds
+            parsed_timestamp = datetime.datetime.strptime(
+                timestamp_str, "%Y-%m-%d %H:%M:%S"
+            )
+        return str(parsed_timestamp)
```
sfc-gh-aling: if users pass us string format timestamp, why we do this str -> datetime -> str conversion for user?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1925,141 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+    )
+
+    # Validate mode
+    if time_travel_mode is None:
+        if time_travel_arg_count == 0:
+            return None
+        else:
+            raise ValueError(
+                "Must specify time travel mode 'at' or 'before' if any other time travel parameter is provided."
+            )
```
sfc-gh-aling: I'm curious what's the error message if there is a SQL having multiple parameters set?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1925,141 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+
+    if time_travel_mode.lower() not in ["at", "before"]:
+        raise ValueError(
+            f"Invalid time travel mode: {time_travel_mode}. Must be 'at' or 'before'."
+        )
+
+    # Validate exactly one parameter is provided
+    if time_travel_arg_count != 1:
+        raise ValueError(
```
sfc-gh-aling: is this the actual error message when we run sql having multiple parameters?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1925,141 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+    elif config.timestamp is not None:
+        if config.timezone.upper() == "NTZ":
+            func_name = "TO_TIMESTAMP_NTZ"
+        elif config.timezone.upper() == "LTZ":
+            func_name = "TO_TIMESTAMP_LTZ"
+        elif config.timezone.upper() == "TZ":
+            func_name = "TO_TIMESTAMP_TZ"
+        else:
+            func_name = "TO_TIMESTAMP_NTZ"  # default fallback
+        clause += f"(TIMESTAMP => {func_name}('{config.timestamp}'))"
```
sfc-gh-aling: I'm curious why we do a transformation here instead of string concatenation?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -101,6 +102,17 @@
 logger = getLogger(__name__)
 
 LOCAL_TESTING_SUPPORTED_FILE_FORMAT = ("JSON",)
+
+TIME_TRAVEL_OPTIONS_PARAMS_MAP = {
```
sfc-gh-aling: this would be exposed as part of public api.
if you think this is just an internal implementation detail, we can make it private by prepending an underscore `_TIME_TRAVEL_OPTIONS_PARAMS_MAP`

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -113,6 +125,7 @@
     "NULLVALUE": "NULL_IF",
     "DATEFORMAT": "DATE_FORMAT",
     "TIMESTAMPFORMAT": "TIMESTAMP_FORMAT",
+    **{k: v.upper() for k, v in TIME_TRAVEL_OPTIONS_PARAMS_MAP.items()},
```
sfc-gh-aling: I think there is only one alias `AS-OF-TIMESTAMP`, no alias in the rest right?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -490,7 +556,23 @@ def table(self, name: Union[str, Iterable[str]], _emit_ast: bool = True) -> Tabl
...
+                statement,
+                offset,
+                timestamp,
+                timezone,
+                stream,
+            ]
+            time_travel_params = dict(
+                zip(TIME_TRAVEL_OPTIONS_PARAMS_MAP.values(), direct_param_values)
+            )
+            time_travel_params = {k: v for k, v in time_travel_params.items()}
```
sfc-gh-aling: I don't think you need this line. it's basically deep copying a time_travel_params
```suggestion
```

src/snowflake/snowpark/table.py
```diff
@@ -314,26 +335,55 @@ def __init__(
...
+                    "stream": self._time_travel_config.stream,
+                }
+            )
+
         return Table(
             self.table_name,
             session=self._session,
             is_temp_table_for_cleanup=self._is_temp_table_for_cleanup,
             _emit_ast=False,
+            **kwargs,
```
sfc-gh-aling: I think you can try using `_asdict()` method on a namedtuple which should simplify your code
https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aling, line 1: @sfc-gh-aalam I remember we default to sql simplifier right? I'm wondering we should considering adding support for sql simplifier disabled case any more

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: Correct, only `timestamp` has an extra alias `as-of-timestamp `to align with the iceberg version

src/snowflake/snowpark/_internal/utils.py
sfc-gh-mayliu: No, SQL only reports a generic message for improper usage.
`Syntax error: unexpected '=>'. (line 16)`
Same msg for the below 3 bad use cases:
```
SELECT  id  FROM timetraveltable BEFORE ( offset => -60*1 and statement => '01be8213-0e11-47d4-000c-a90b9a62a653');

SELECT  id  FROM timetraveltable BEFORE ( offset => -60*1) BEFORE (statement => '01be8213-0e11-47d4-000c-a90b9a62a653');

SELECT  * 
 FROM (
 SELECT  id  FROM timetraveltable
) BEFORE ( offset => -60*1);
```

src/snowflake/snowpark/_internal/utils.py
sfc-gh-mayliu: using `strptime` here mainly to parse and validate timestamp str format in case of user typo or unsupported format, which could be common since timestamp is strict on format and people have different habits in reporting date & time. would it be better to throw a detailed error here or let the SQL engine handle it in case of str format error?

src/snowflake/snowpark/_internal/utils.py
sfc-gh-mayliu: The time travel doc mentions
> The value must be explicitly cast to a TIMESTAMP, TIMESTAMP_LTZ, TIMESTAMP_NTZ, or TIMESTAMP_TZ data type.

I think it'd be clearer to explicitly cast the timestamp to the according type here

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-aalam: Simplifier enabled is the default but we haven't deprecated the disabled path yet. We could start as a team to not support the disabled path and eventually remove all disabled paths from our code.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1926,144 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+    mode: str  # 'at' or 'before'
+    statement: Optional[str] = None
+    offset: Optional[int] = None
+    timestamp: Optional[str] = None
+    timezone: str = "NTZ"
+    stream: Optional[str] = None
+
+
+def validate_and_normalize_time_travel_params(
+    time_travel_mode: Optional[str] = None,
```
sfc-gh-aalam: you can use Literal here with options `'at'` and `'before'`.
example: https://github.com/snowflakedb/snowpark-python/blob/6cd8f0728a067f8a5c8e9daee4c9fafec132843c/src/snowflake/snowpark/dataframe_writer.py#L227

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1926,144 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+
+    mode: str  # 'at' or 'before'
+    statement: Optional[str] = None
+    offset: Optional[int] = None
+    timestamp: Optional[str] = None
+    timezone: str = "NTZ"
+    stream: Optional[str] = None
+
+
+def validate_and_normalize_time_travel_params(
```
sfc-gh-aalam: have you though about making this a static method of `TimeTravelConfig` class? seems like they are closely related.

src/snowflake/snowpark/table.py
```diff
@@ -314,26 +336,41 @@ def __init__(
...
         )
 
     def __copy__(self) -> "Table":
+        kwargs = {}
+        if self._time_travel_config:
+            time_travel_dict = self._time_travel_config._asdict()
+            time_travel_dict["time_travel_mode"] = time_travel_dict.pop("mode")
```
sfc-gh-aalam: can you override the _asdict method in `TimeTravelConfig` since it is repeated here

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -101,6 +103,17 @@
...
+
+_TIME_TRAVEL_OPTIONS_PARAMS_MAP = {
+    "TIME_TRAVEL_MODE": "time_travel_mode",
+    "STATEMENT": "statement",
+    "OFFSET": "offset",
+    "TIMESTAMP": "timestamp",
+    "TIMEZONE": "timezone",
+    "STREAM": "stream",
+    "AS-OF-TIMESTAMP": "timestamp",
+}
```
sfc-gh-aalam: any particular reason why we chose `AS-OF-TIMESTAMP` instead of `AS_OF_TIMESTAMP`?

sfc-gh-mayliu: Iceberg uses `AS-OF-TIMESTAMP`, so it'd be more compatible with it to make migration easier:
https://iceberg.apache.org/docs/latest/spark-queries/#dataframe
```
// time travel to October 26, 1986 at 01:21:00
spark.read
    .option("as-of-timestamp", "499162860000")
    .format("iceberg")
    .load("path/to/table")
```

src/snowflake/snowpark/_internal/utils.py
sfc-gh-mayliu: Good point, `Literal` is a lot better at type checking than a mere `Optional[str]`

src/snowflake/snowpark/_internal/utils.py
sfc-gh-mayliu: sounds good, I put `generate_sql_clause` under `TimeTravelConfig` as well

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1926,149 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+        elif self.timestamp is not None:
+            if self.timezone.upper() == "NTZ":
+                func_name = "TO_TIMESTAMP_NTZ"
+            elif self.timezone.upper() == "LTZ":
+                func_name = "TO_TIMESTAMP_LTZ"
+            elif self.timezone.upper() == "TZ":
+                func_name = "TO_TIMESTAMP_TZ"
+            else:
+                func_name = "TO_TIMESTAMP_NTZ"  # default fallback
+            clause += f"(TIMESTAMP => {func_name}('{self.timestamp}'))"
```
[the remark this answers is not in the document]
sfc-gh-aalam: This is a good point. Do we know what kind of values are allowed here in sql generation @sfc-gh-mayliu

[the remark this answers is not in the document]
sfc-gh-mayliu: There can be 4 types of params:
- `statement` has to be a query ID, maybe we can provide some regex parsing to verify the length of the input string.
- `timestamp` if provided as string type is parsed and validated in utils.py::_normalize_timestamp
- `offset` is only accepted as an int so we're good here
- `stream` is a name identifier which can be anything...

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1926,149 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+        return TimeTravelConfig(
+            mode=time_travel_mode,
+            statement=statement,
+            offset=offset,
+            timestamp=normalized_timestamp,
+            timezone=timezone,
+            stream=stream,
+        )
+
+    def generate_sql_clause(self) -> str:
```
sfc-gh-aling, line 2040: TimeTravelConfig instance is immutable (a named tuple) right?

I think we can make generate_sql_clause a cached_property?
https://docs.python.org/3/library/functools.html#functools.cached_property

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1926,149 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+            elif self.timezone.upper() == "TZ":
+                func_name = "TO_TIMESTAMP_TZ"
+            else:
+                func_name = "TO_TIMESTAMP_NTZ"  # default fallback
+            clause += f"(TIMESTAMP => {func_name}('{self.timestamp}'))"
+
+        return clause
+
+
+def _normalize_timestamp(timestamp: Union[str, datetime.datetime]) -> str:
```
sfc-gh-aling: is this function used to validate user input timestamp? for string input, snowflake supports datetime format auto detection -- not just format ""%Y-%m-%d %H:%M:%S.%f": https://docs.snowflake.com/en/sql-reference/date-time-input-output#label-date-time-input-output-supported-formats-for-auto-detection

btw there is a lib python-dateutil which we already depends on, you can check if this library helps validate datetime

if we think too much complexity will be involved in validating, we should let the server handle it then

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -101,6 +103,17 @@
...
 LOCAL_TESTING_SUPPORTED_FILE_FORMAT = ("JSON",)
+
+_TIME_TRAVEL_OPTIONS_PARAMS_MAP = {
+    "TIME_TRAVEL_MODE": "time_travel_mode",
+    "STATEMENT": "statement",
+    "OFFSET": "offset",
+    "TIMESTAMP": "timestamp",
+    "TIMEZONE": "timezone",
+    "STREAM": "stream",
+    "AS-OF-TIMESTAMP": "timestamp",
```
sfc-gh-aling: I'm reconsidering this option from migration perspective:

snowpark, "time_travel_mode" is required
session.read.option("time_travel_mode", "at").option("as-of-timestamp", ...)

pyspark 
session.read.option("as-of-timestamp")

if we want to go on par with pyspark that users can just call `session.read.option("as-of-timestamp")` without inputing `time_travel_mode`

 then "as-of-timestamp" is not just alias of "timestamp", it also means using time-travel-mode "at" (auto set mode to "at"), and it can't work with time-travel-mode "before" (raise error)

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1926,149 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
     )
 
 
+class TimeTravelConfig(NamedTuple):
+    """Configuration for time travel operations."""
+
+    mode: str  # 'at' or 'before'
```
sfc-gh-aling: nit: you can just rename this to "time_travel_mode" for consistency, such that you do not need to overwrite the `asdict_for_table`

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -490,7 +557,19 @@ def table(self, name: Union[str, Iterable[str]], _emit_ast: bool = True) -> Tabl
...
+            time_travel_params = {
+                "time_travel_mode": time_travel_mode,
+                "statement": statement,
+                "offset": offset,
+                "timestamp": timestamp,
+                "timezone": timezone,
+                "stream": stream,
+            }
+        else:
+            time_travel_params = _extract_time_travel_from_options(self._cur_options)
```
sfc-gh-aling, line 612: what would happen if users call 
session.read.json/csv...option("time-travel-mode"), these time travel parameters are ignored?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1924,6 +1926,149 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+            )
+
+        # Validate offset
+        if offset is not None and offset > 0:
+            raise ValueError("'offset' must be a negative integer (seconds).")
+
+        # Normalize timestamp
+        normalized_timestamp = None
+        if timestamp is not None:
+            normalized_timestamp = _normalize_timestamp(timestamp)
```
sfc-gh-aling: as per discussion: local validation can be removed

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aling: session.read.option(as-of-timestamp)
session.read.option(time_tavel_mode, "at").option(as-of-timestamp)
session.read.option(time_tavel_mode, "before").option(as-of-timestamp)

src/snowflake/snowpark/_internal/utils.py
sfc-gh-mayliu: Yes, `NamedTuple` is immutable and good point about making it a `cached_property`. However, there's a limitation of using `@cached_property` because `NamedTuple` does not support `__dict__`, so there's no place for cached_property to store the cached result.

As per our discussion, since `generate_sql_clause` doesn't involve heavy computation and it's mostly invoked in `SelectableEntity` constructor (via the simplified query generation path), it should be fine to leave it as is. Thanks!

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: users shouldn't be able to call `session.read.json/csv...option("time-travel-mode")`. `session.read.json/csv` returns a `DataFrame`, but `option` only works on `DataFrameReader`.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -114,6 +126,8 @@
     "NULLVALUE": "NULL_IF",
     "DATEFORMAT": "DATE_FORMAT",
     "TIMESTAMPFORMAT": "TIMESTAMP_FORMAT",
+    **{k: v.upper() for k, v in _TIME_TRAVEL_OPTIONS_PARAMS_MAP.items()},
+    "AS-OF-TIMESTAMP": "AS-OF-TIMESTAMP",
```
sfc-gh-aling: nit: do we need these two lines?


if you check where this dict is used, we also call `get_aliased_option_name` that by default returns uppered key

tests/integ/test_dataframe.py
```diff
@@ -5910,3 +5910,400 @@ def test_order_by_col(session, sql_simplifier_enabled, local_testing_mode):
...
+        df_before_ts_after_update = session.table(
+            table_name,
+            time_travel_mode="before",
+            timestamp=ts_after_update,
+            timezone="LTZ",
+        )
+        Utils.check_answer(df_before_ts_after_update, expected_after_update)
+
+        df_at_ts_after_update = session.table(
+            table_name, time_travel_mode="at", timestamp=ts_after_update, timezone="LTZ"
```
sfc-gh-aling: have you tested passing TimestampTimezone.LTZ as timezone?

tests/integ/test_simplifier_suite.py
```diff
@@ -1399,6 +1400,141 @@ def test_col_ilike(session, use_simplified_query_generation):
         session.conf.set("use_simplified_query_generation", original)
 
 
+@pytest.mark.parametrize("use_simplified_query_generation", [True, False])
```
sfc-gh-aling, line 1403: you can use Utils.check_answer to compare df.collect() with result

tests/integ/test_dataframe.py
sfc-gh-mayliu: good catch!

src/snowflake/snowpark/session.py
```diff
@@ -2519,6 +2526,15 @@ def table(
 
             _emit_ast: Whether to emit AST statements.
 
+            time_travel_mode: Time travel mode, either 'at' or 'before'. When specified,
+                returns a DataFrame with time travel query instead of a Table.
+                Exactly one of statement, offset, timestamp, or stream must be provided when time_travel_mode is set.
+            statement: Query ID for time travel.
+            offset: Negative integer representing seconds in the past for time travel.
+            timestamp: Timestamp string in 'YYYY-MM-DD HH:MM:SS' format or datetime object.
+            timezone: Timezone for timestamp conversion ('NTZ', 'LTZ', or 'TZ'). Defaults to 'NTZ'.
```
sfc-gh-yixie: The name timezone may be misleading. People may think the value is like `America/Los_Angeles`, `GMT-7`, etc.
I suggest `timestamp_type`. 

What's the behavior if the timestamp already has a timezone and the timestamp_type is NTZ?

sfc-gh-mayliu: good points! renamed `timezone` to `timestamp_type`. 

for the original behavior, if user provides` datetime.tzinfo`, it would’ve overwritten the tzinfo with NTZ if no `timestamp_type` was specified. to better handle both `str` and `datetime` inputs for `timestamp` and `timestamp_type`:

- `datetime` with timezone + timestamp_type NOT provided → auto-sets to 'TZ'
- `datetime` with timezone + timestamp_type explicitly provided → uses provided timestamp_type
- `datetime` without timezone (naive) → uses provided timestamp_type or defaults to 'NTZ'
- `string` timestamps → uses provided timestamp_type or defaults to 'NTZ'

this way, we prioritize user-provided `timestamp_type` to respect their choices, but still preserve `tzinfo` if not specified. for string timestamps, since there are too many possible formats, we avoid heavy parsing and pass the raw string to snowflake server for built-in auto-detection.

CHANGELOG.md
```diff
@@ -5,7 +5,10 @@
 ### Snowpark Python API Updates
 
 #### New Features
-
+- Added time travel support for querying historical data:
+  - `Session.table()` now supports time travel parameters: `time_travel_mode`, `statement`, `offset`, `timestamp`, `timezone`, and `stream`.
```
sfc-gh-aling: nit: since we change the param name for timezone, let's update here?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1939,6 +1941,151 @@ def remove_comments(sql_query: str, uuids: List[str]) -> str:
...
+        # Validate exactly one parameter is provided
+        if time_travel_arg_count != 1:
+            raise ValueError(
+                "Exactly one of 'statement', 'offset', 'timestamp', or 'stream' must be provided."
+            )
+
+        # Normalize timestamp
+        normalized_timestamp = None
+        if timestamp is not None:
+            normalized_timestamp = _normalize_timestamp(timestamp)
```
sfc-gh-aling: nit: do we need _normalize_timestamp any more?
I think the logic can be simplified as str(timestamp).strip() which works for both datetime and str

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -128,6 +140,42 @@ def _validate_stage_path(path: str) -> str:
...
+    Special handling for 'AS-OF-TIMESTAMP' (PySpark compatibility):
+    - Automatically sets time_travel_mode to 'at'
+    - Cannot be used with time_travel_mode='before' (raises error)
+    - Cannot be mixed with regular 'timestamp' option (raises error)
+    """
+    result = {}
+    excluded_keys = set()
+
+    # Handle PySpark as-of-timestamp special case
+    if "AS-OF-TIMESTAMP" in options:
```
sfc-gh-aling, line 156: options would only have upper cased keys?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -926,14 +1056,35 @@ def option(self, key: str, value: Any, _emit_ast: bool = True) -> "DataFrameRead
...
+            - ``timestamp_type``: Type of timestamp interpretation ('NTZ', 'LTZ', or 'TZ').
+
+              Smart timestamp_type handling:
+
+              - datetime with timezone + timestamp_type NOT provided → auto-sets to 'TZ'
+              - datetime with timezone + timestamp_type explicitly provided → uses provided timestamp_type
+              - datetime without timezone (naive) + timestamp_type NOT provided → no casting, raw string (timestamp_type=None)
+              - datetime without timezone (naive) + timestamp_type explicitly provided → uses provided timestamp_type
+              - string timestamps + timestamp_type NOT provided → no casting, raw string (timestamp_type=None, Snowflake native handling)
+              - string timestamps + timestamp_type explicitly provided → uses TO_TIMESTAMP_XXX casting
```
sfc-gh-aling: this looks like internal implementation details, I think we can exclude it first and add back based upon user feedback

src/snowflake/snowpark/table.py
```diff
@@ -288,6 +293,15 @@ def __init__(
...
         _ast_stmt: Optional[proto.Bind] = None,
         _emit_ast: bool = True,
+        *,
+        time_travel_mode: Optional[Literal["at", "before"]] = None,
+        statement: Optional[str] = None,
+        offset: Optional[int] = None,
+        timestamp: Optional[Union[str, datetime.datetime]] = None,
+        timestamp_type: Optional[
+            Union[str, TimestampTimeZone]
+        ] = TimestampTimeZone.DEFAULT,
```
sfc-gh-aling: this should be None

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: yes, `def option` calls a helper to convert all` self._cur_options `keys to uppercase.
`aliased_key = get_aliased_option_name(key, READER_OPTIONS_ALIAS_MAP)`

tests/integ/test_dataframe.py
```diff
@@ -5910,3 +5910,403 @@ def test_order_by_col(session, sql_simplifier_enabled, local_testing_mode):
...
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="Time travel is not supported in Local Testing",
+)
+def test_time_travel_core_functionality(session):
+    table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
+    data = [(1, "product_a", 100), (2, "product_b", 200), (3, "product_c", 300)]
+    df = session.create_dataframe(data, schema=["id", "name", "price"])
+    df.write.save_as_table(table_name, table_type="temporary")
+    session.table(table_name).collect()
```
sfc-gh-aling: nit: I don't think we need this line?

tests/integ/test_dataframe.py
```diff
@@ -5910,3 +5910,403 @@ def test_order_by_col(session, sql_simplifier_enabled, local_testing_mode):
...
+    ]
+    df_1 = session.create_dataframe(data_1, schema=["id", "name", "price", "active"])
+    df_1.write.save_as_table(table1_name, table_type="temporary")
+
+    data_2 = [(1, "electronics"), (2, "clothing"), (3, "automotive")]
+    df_2 = session.create_dataframe(data_2, schema=["id", "category"])
+    df_2.write.save_as_table(table2_name, table_type="temporary")
+
+    # Sleep to ensure stable time travel window
+    import time
```
sfc-gh-aling: you can move import to the top of the file to avoid importing twice in test functions

tests/integ/test_dataframe.py
```diff
@@ -5910,3 +5910,403 @@ def test_order_by_col(session, sql_simplifier_enabled, local_testing_mode):
...
+            timestamp=ts_before_update,
+            timestamp_type=TimestampTimeZone.LTZ,
+        )
+        Utils.check_answer(df_before_ts, expected_before_update)
+
+        df_at_ts = session.table(
+            table_name,
+            time_travel_mode="at",
+            timestamp=ts_before_update,
+            timestamp_type="LTZ",
```
sfc-gh-aling, line 5996: I forget what's the reason we have to use LTZ in test, probably worth leaving a comment.

sfc-gh-aling (approved):

PR #3687, opened 2025-08-21
SNOW-2230971: Re-introduce the support of repr, joins, loc, reset_index, and binary ops in Faster pandas (Two-QCs)

sfc-gh-helmeleegy:
This reverts commit 371d3a4fd6246ed692d3f9c64c6b2cd1daa45c5f.

   Fixes SNOW-2230971

   Re-introduce the support of repr, joins, loc, reset_index, and binary ops in Faster pandas using the two QueryCompilers approach.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -2730,10 +2730,14 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
             or other._query_compiler._relaxed_query_compiler is not None
         ):
             if isinstance(other, (Series, DataFrame)):
-                new_other = copy.copy(other)
-                new_other._query_compiler = (
-                    other._query_compiler._relaxed_query_compiler
-                )
+                if isinstance(other, Series):
+                    new_other = Series(
+                        query_compiler=other._query_compiler._relaxed_query_compiler
+                    )
+                else:  # DataFrame
+                    new_other = DataFrame(
+                        query_compiler=other._query_compiler._relaxed_query_compiler
+                    )
             else:
                 new_other = other
             relaxed_query_compiler = self._relaxed_query_compiler._binary_op_internal(


tests/integ/modin/test_faster_pandas.py
```diff
@@ -0,0 +1,66 @@
...
+        df2[df2["D"] == 22], left_on="A", right_on="C"
+    )
+
+    # create pandas dataframes
+    native_df1 = df1.to_pandas()
+    native_df2 = df2.to_pandas()
+    native_result = native_df1[native_df1["B"] > 11].merge(
+        native_df2[native_df2["D"] == 22], left_on="A", right_on="C"
+    )
+
```
sfc-gh-jkew, line 53: Can we have a test which transitions from relaxed mode to ordered mode? Just test the transition point.

sfc-gh-helmeleegy: The chain in the second test will be using both the relaxed and ordered modes side by side until the filtering step. Then, once the groupby/agg step is introduced, the relaxed mode is dropped and only the ordered mode is left.

For the chain in the second test, all operations are supported and hence both the relaxed and ordered modes are maintained until the end.

sfc-gh-helmeleegy: Note that in the 2-QC approach, there are no "transitions" -- just a decision as to how many QCs to keep maintaining as we receive each new operation.

sfc-gh-mvashishtha (approved):

tests/integ/modin/test_faster_pandas.py
sfc-gh-jkew: Do you mean the first test (test_read_filter_join) transitions from two qc to one qc at line 32? I just want a test to see that we have collapsed to a single qc past that point.

The second test (test_read_filter_groupby_agg) doesn't look like it would transition at all afaict.

tests/integ/modin/hybrid/test_extensions.py
```diff
@@ -104,7 +104,7 @@ def test_to_snowpark(object_backend, to_snowpark, global_backend):
         ),
     ],
 )
-@sql_count_checker(query_count=6)
+@sql_count_checker(query_count=8)
```
sfc-gh-jkew: Why do these increase?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2698,6 +2738,27 @@ def binary_op(
...
         # Native pandas does not support binary operations between a Series and a list-like object.
 
+        relaxed_query_compiler = None
+        if self._relaxed_query_compiler is not None and (
+            not isinstance(other, (Series, DataFrame))
+            or other._query_compiler._relaxed_query_compiler is not None
+        ):
+            relaxed_other = other
+            if isinstance(other, (Series, DataFrame)):
+                relaxed_other._query_compiler = copy.copy(
```
sfc-gh-jkew: Why do we need to make a copy of the relaxed query compiler - why not pass in the underlying relaxed query compiler to the binary operation?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2698,6 +2738,27 @@ def binary_op(
 
         # Native pandas does not support binary operations between a Series and a list-like object.
 
+        relaxed_query_compiler = None
+        if self._relaxed_query_compiler is not None and (
+            not isinstance(other, (Series, DataFrame))
+            or other._query_compiler._relaxed_query_compiler is not None
+        ):
+            relaxed_other = other
```
sfc-gh-jkew: relaxed_other can be a scalar here right, not just a qc. I see where you want to use relaxed_other as the other operand, but since it could be a scalar maybe it could be clearer if were just another operand.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2815,14 +2892,22 @@ def binary_op(
             # However, pandas allows to call Series <binop> DataFrame with axis=0 set. In this case, the parameter
             # axis=0 is ignored and the result works the same as if axis=1 is invoked.
             if not squeeze_self and axis == 0:
-                return self._binary_op_between_dataframe_and_series_along_axis_0(
+                qc = self._binary_op_between_dataframe_and_series_along_axis_0(
                     op, other._query_compiler, fill_value
                 )
+                qc._dummy_row_pos_mode = self._dummy_row_pos_mode
+                if relaxed_query_compiler is not None:
+                    qc._relaxed_query_compiler = relaxed_query_compiler
```
sfc-gh-jkew: What makes the 2QC approach so complicated is that every return point of a supported function needs to have special code to possibly set the relaxed_query_compiler. There's no clear way to separate out the concerns either without a complete rewrite of the function.

I still keep thinking we might be able to clean this up with a wrapper function.

So you might write a new method called:
```
def _binary_op_relaxed_outer(self,
        op: str,
        other: Union[Scalar, AnyArrayLike, "pd.Series", "pd.DataFrame"],
        ...
    ) -> "SnowflakeQueryCompiler":
    relaxed_query_compiler = None
    if ( we need to perform the relaxed qc equivalent ): 
         // set relaxed_other and run the same code on the underlying qc
         relaxed_query_compiler  = self._relaxed_query_compiler.binary_op(other_relaxed_qc,...)
    // run the normal function
    self.binary_op(...)
    if relaxed_query_compiler is not None:
       self._relaxed_query_compiler = relaxed_query_compiler
```

So in the prototype code above the function, `binary_op` never changes. The only thing left is to make sure that 
`_binary_op_relaxed_outer` is only called on the top-level qc; which can be handled with some modification of aliases.

My point is that instead of having many different places in `binary_op` to maybe set the relaxed query compiler we *separate the concerns of relaxed qc management from the underlying logic of the function*

sfc-gh-jkew (changes requested):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -8204,6 +8343,28 @@ def merge(
             )
 
         left = self
+        relaxed_query_compiler = None
+        if (
+            left._relaxed_query_compiler is not None
+            and right._relaxed_query_compiler is not None
+            and how not in ["asof", "cross"]
```
sfc-gh-jkew: Why would cross joins not work w/ relaxed?

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
```diff
@@ -165,6 +165,7 @@ def join(
     sort: Optional[bool] = False,
     join_key_coalesce_config: Optional[list[JoinKeyCoalesceConfig]] = None,
     inherit_join_index: InheritJoinIndex = InheritJoinIndex.FROM_LEFT,
+    dummy_row_pos_mode: bool = False,
```
sfc-gh-jkew, line 168: dummy_row_pos_mode has a default of False; but should it be a non-optional parameter given that it has to be worm-holed everywhere?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2719,15 +2782,27 @@ def binary_op(
...
-                return self._binary_op_list_like_rhs_axis_0(op, other, fill_value)
+                qc = self._binary_op_list_like_rhs_axis_0(op, other, fill_value)
+                qc._dummy_row_pos_mode = self._dummy_row_pos_mode
+                if relaxed_query_compiler is not None:
+                    qc._relaxed_query_compiler = relaxed_query_compiler
+                return qc
             else:  # axis=1
-                return self._binary_op_list_like_rhs_axis_1(op, other, fill_value)
+                qc = self._binary_op_list_like_rhs_axis_1(op, other, fill_value)
+                qc._dummy_row_pos_mode = self._dummy_row_pos_mode
```
sfc-gh-jkew: Maybe cleaning this up is as simple as creating a function like:

```
def _maybe_set_relaxed_and_return(qc, relaxed_query_compiler):
    qc._dummy_row_pos_mode = self._dummy_row_pos_mode
    if relaxed_query_compiler is not None:
        qc._relaxed_query_compiler = relaxed_query_compiler
    return qc
```

You would then put, at each return point:
```
return _maybe_set_relaxed_and_return(
      self._binary_op_list_like_rhs_axis_0(op, other, fill_value),
      relaxed_query_compiler
      )
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -3752,7 +3865,10 @@ def reset_index(
             index_column_types=index_column_types_remained,
         )
 
-        return SnowflakeQueryCompiler(internal_frame)
+        qc = SnowflakeQueryCompiler(internal_frame)
+        if relaxed_query_compiler is not None:
```
sfc-gh-jkew: Do we need to set `qc._dummy_row_pos_mode=self._dummy_row_pos_mode` here?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13621,7 +13849,16 @@ def build_repr_df(
         # 2. retrieve all columns
         # 3. filter on rows with recursive count
 
-        frame = self._modin_frame.ensure_row_position_column()
+        if self._relaxed_query_compiler is None:
+            frame = self._modin_frame.ensure_row_position_column(
+                dummy_row_pos_mode=self._dummy_row_pos_mode
```
sfc-gh-jkew, line 13948: Would dummy_row_pos_mode ever be true if there is no relaxed_query_compiler?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -13653,25 +13890,97 @@ def build_repr_df(
...
+                )
+                | (
+                    col(row_position_snowflake_quoted_identifier)
+                    >= row_count_expr - num_rows_for_head_and_tail
+                )
             )
-            | (
-                col(row_position_snowflake_quoted_identifier)
-                >= row_count_expr - num_rows_for_head_and_tail
+        else:
```
sfc-gh-jkew, line 14008: Using the limit clause here doesn't seem dependent on whether there is a relaxed_query compiler or not... This seems like a more general optimization if relaxed mode is enabled. I'm wondering if we could use that boolean rather than the existence of a second qc to do this.

sfc-gh-jkew: And given how this part of `build_repr_df` is pretty separate I wonder if we can refactor it out to it's own function. Maybe that's not super important; but it may help with readability.

sfc-gh-jkew (changes requested):

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -1229,11 +1241,25 @@ def join(
...
-        ] and self._is_self_join_on_row_position_column(
+        ] and left._is_self_join_on_row_position_column(
             left_on_cols, right, right_on_cols
         ):
             is_join_needed = False
 
+        if is_join_needed and dummy_row_pos_mode and how not in ["asof", "cross"]:
+            # Replace the dummy row position with a real one before performing a join on the row position
+            # This currently does not handle the unlikely case of joining on both the row position and a data column
+            if left_on_cols == [left.row_position_snowflake_quoted_identifier]:
```
sfc-gh-jkew: This section doesn't have any code coverage; I would like to make sure most of the big, non-trivial code is covered by tests.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1435,11 +1438,33 @@ def from_snowflake(
...
         columns: Optional[list[str]] = None,
         enforce_ordering: bool = False,
+        dummy_row_pos_mode: bool = False,
     ) -> "SnowflakeQueryCompiler":
         """
         See detailed docstring and examples in ``read_snowflake`` in frontend layer:
         src/snowflake/snowpark/modin/plugin/pd_extensions.py
         """
+        dummy_row_pos_optimization_enabled = (
+            os.environ.get(
```
sfc-gh-jkew, line 1458: I would use a modin config variable here. Users should not have to `import os` to turn this on or off.

I'm also not totally sure it would work within warehouse notebooks - which would execute as a separate stored proc per cell. That's a bit of a weird case.

sfc-gh-jkew (changes requested):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -611,6 +612,8 @@ def __init__(self, frame: InternalFrame) -> None:
             frame, InternalFrame
         ), "frame is None or not a InternalFrame"
         self._modin_frame = frame
+        self._dummy_row_pos_mode = False
+        self._relaxed_query_compiler: Optional[SnowflakeQueryCompiler] = None
```
sfc-gh-jkew, line 616: I would create a new query compiler class, which is more or less a dummy query compiler class to help differentiate this from other query compilers in telemetry. It would also let us make sure that hybrid mode is not applied to the relaxed query compiler in any way.

We should make sure, under hybrid mode, that we would not use the query_compiler_caster.

tests/integ/modin/test_faster_pandas.py
```diff
@@ -0,0 +1,66 @@
...
+import pandas as native_pd
+
+from snowflake.snowpark._internal.utils import TempObjectType
+import snowflake.snowpark.modin.plugin  # noqa: F401
+from tests.integ.modin.utils import assert_frame_equal
+from tests.integ.utils.sql_counter import sql_count_checker
+from tests.utils import Utils
+
+
+@sql_count_checker(query_count=5, join_count=1)
```
sfc-gh-jkew, line 15: We also need to make sure we have some tests for this under hybrid mode; maybe in the test_switch_operations. In that test we should verify that telemetry is disabled and that we would not invoke the query_compiler_caster.

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-jkew: Also - Is this the fix to the join optimizer issue tested by `test_comparing_series_from_same_read_snowflake_result_does_not_require_join_SNOW_2250244` ?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: As an alternative to the solution above I keep thinking there must be a way to use a derived version of the SnowflakeQueryCompiler to isolate the relaxed changes by calling `super.[op]`, but I don't think that's possible.

There really isn't a clean way to reduce the cyclomatic complexity.

sfc-gh-jkew: This issue BTW; is similar to the comment I left on the original review, which was not addressed: https://github.com/snowflakedb/snowpark-python/pull/3602#discussion_r2244168333

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: This comment is similar to the comment I left on the original review here: https://github.com/snowflakedb/snowpark-python/pull/3602#discussion_r2240373927

tests/integ/modin/test_faster_pandas.py
sfc-gh-helmeleegy: I expanded on the tests to also check that the relaxed query compiler is populated until the end in the case of read_filter_join since all operations are supported. But it gets set to none when the groupby/agg is applied in the case of read_filter_groupby_agg since those operations are not supported yet.

tests/integ/modin/hybrid/test_extensions.py
sfc-gh-helmeleegy: This was before you merge your change to check the metadata for counts instead of issuing select queries. Now these query counts are not altered.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Updated.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Both scalar and non-scalar versions of `other` are handled. But I changed the variable name to `used_other` instead of `relaxed_other` to avoid confusion, especially in the case of scalar where it stays the same.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: I was just focusing on the common cases of join for now.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-helmeleegy: The default value is meant for consistency with most other parameters (and other methods taking the same argument at the same layer of code) and also for safety in case one call site does not set the parameter. If you feel strongly about it, I can make it non-optional for this method.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Good idea. Done.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: This logic has changed with the introduction of `_maybe_set_relaxed_qc`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: `dummy_row_pos_mode` would only be true inside a relaxed qc. This has been made simpler after the introduction of `_maybe_set_relaxed_qc`.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: No, `limit` is only applicable when the row position is dummy; i.e., we are free to choose the ordering of rows. So we can assume that the rows returned by the limit query represent the first and last few rows. But if the row order was pre-determined, then we have to use a where clause instead of limit to retrieve the specific rows occurring at the top and bottom of the df.

sfc-gh-helmeleegy: I'm not sure it's necessary to break up the `build_repr_df` into two methods. There is no reusability argument here and the method is not excessively long now. It just handles two different cases. But let me know if it's something you feel strongly about.

tests/integ/modin/test_faster_pandas.py
sfc-gh-helmeleegy: I think in the offline discussion we realized that the existing test_switch_operations and telemetry tests do use the 2QC approach while the hybrid mode is enabled, and they are all passing. Hence, no more tests are needed there, correct?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Just to confirm, in our offline discussion, we arrived at the conclusion that we won't be creating a separate QC class, right?

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-helmeleegy: Added a new test for that: `test_read_filter_join_on_index`. And no, this was not related to the join elimination optimization. The fix for that issue was added in the binary_op implementation.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: A better option is to make it a snowpark parameter similar to the many other parameters snowpark has. Also, this is how users are used to make parameter changes to snowpark. I suggest to do this in a followup PR since it will require a different set of reviewers who don't need to be aware of all the details in this PR.

sfc-gh-helmeleegy: I think modin config variables are more useful when they are applicable to Snowflake as well as other backends, which is not the case here. This is specific to Snowflake/Snowpark pandas.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: I didn't notice your first comment in this thread, let me try that. But I guess this would mean that each QC method will be split to two methods, right? For now, I'm keeping it as one method but while using `_may_set_relaxed_qc` whenever we are returning from the method, which made it much more concise than before.

sfc-gh-helmeleegy: Updated. This was a good suggestion. Looks much better to me now.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: Yeah, I don't think it is necessary because I don't *think* we are applying the wrappers.

src/snowflake/snowpark/modin/plugin/_internal/join_utils.py
sfc-gh-jkew: I think I'm just worried about not passing it in a place and then having unexpected behaviors; but we can make it non-optional later.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -2818,6 +2912,7 @@ def binary_op(
                 return self._binary_op_between_dataframe_and_series_along_axis_0(
                     op, other._query_compiler, fill_value
                 )
+                # return self._maybe_set_relaxed_qc(qc, relaxed_query_compiler)
```
sfc-gh-jkew: Is this left over from an edit?

sfc-gh-helmeleegy: yup

PR #3691, opened 2025-08-21
SNOW-2275424: fixudtf_ingestion does not work in stored proc because function name is out of spec

sfc-gh-yuwang:

   

   Fixes SNOW-2275424


tests/integ/test_stored_procedure.py
```diff
@@ -2501,3 +2511,71 @@ def worker_process(parquet_queue):
 
     ingestion = sproc(core_ingestion_logic, return_type=StringType())
     assert ingestion() == "success"
+
+
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="data source is not supported in local testing",
+    run=False,
+)
```
sfc-gh-aling, line 2513: this test only works on github right? then we should have the decorator to skip for non-GH envs

tests/integ/test_stored_procedure.py
```diff
@@ -16,6 +16,14 @@
 
 import pytest
 
+from tests.integ.datasource.test_oracledb import (
+    ORACLEDB_TEST_EXTERNAL_ACCESS_INTEGRATION,
+)
+from tests.parameters import ORACLEDB_CONNECTION_PARAMETERS
```
sfc-gh-aling: these two imports may cause failure in our snowfort tests, I don't remember seeing `tests.parameters ` in our snowfort setup.

you can move them into your test method and add a skip in non GH env to your test case

tests/integ/test_stored_procedure.py
```diff
@@ -2501,3 +2511,72 @@ def worker_process(parquet_queue):
...
+    "config.getoption('local_testing_mode', default=False)",
+    reason="data source is not supported in local testing",
+    run=False,
+)
+def test_data_source_udtf_ingestion(session, db_parameters):
+    with Session.builder.configs(db_parameters).create() as new_session:
+        new_session.custom_package_usage_config["enabled"] = True
+        new_session.custom_package_usage_config["force_push"] = True
+        new_session.add_packages("oracledb")
+        new_session.add_packages("snowflake-snowpark-python")
```
sfc-gh-aling, line 2528: do we need to create a new session? can we use the fixture session

sfc-gh-yuwang: i am afraid that these session settings would affect other test on this session, so create a new one would be better

sfc-gh-aling: what would happen if you don't do these operations on the new session, would the test fail?

sfc-gh-yuwang: yes, I basically set these settings with the reminder in the error message if I don't set them

PR #3708, opened 2025-08-26
SNOW-2296598: Estimate row count only when Hybrid is enabled

sfc-gh-helmeleegy:

   

   Fixes SNOW-2296598

   Estimate row count only when Hybrid is enabled.


tests/integ/modin/test_ordered_dataframe.py
```diff
@@ -1119,6 +1119,10 @@ def test_snowpark_pandas_statement_params(session, df1):
 @sql_count_checker(query_count=7)
 @pytest.mark.parametrize("columns", [["A", "b", "C"]])
 def test_ordered_dataframe_row_count(session, columns):
+    from modin.config import AutoSwitchBackend
```
sfc-gh-jkew: Instead of turning on Hybrid; call count_row on the dataframe to set the row_count to the correct value.

sfc-gh-helmeleegy: Done.

sfc-gh-jkew (approved):

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
 #### Improvements
 - Set the default transfer limit in hybrid execution for data leaving Snowflake to 100k, which can be overridden with the SnowflakePandasTransferThreshold environment variable. This configuration is appropriate for scenarios with two available engines, "Pandas" and "Snowflake" on relational workloads.
 - Improve import error message by adding '--upgrade' to 'pip install "snowflake-snowpark-python[modin]"' in the error message.
+  - Only perform row count estimation when Hybrid Execution is enabled.
```
sfc-gh-joshi:
```suggestion
- Only perform row count estimation when hybrid execution is enabled. This reduces the number of queries issued for many workloads.
```

tests/integ/modin/test_ordered_dataframe.py
```diff
@@ -1134,6 +1135,12 @@ def test_ordered_dataframe_row_count(session, columns):
         row_position_quoted_identifier,
     ) = create_initial_ordered_dataframe(test_table_name, enforce_ordering=True)
 
+    # Manualy initialize row_count and row_count_upper_bound
```
sfc-gh-joshi, line 1138: Why are these explicit calls needed? Doesn't this defeat the purpose of the test if we're doing explicit sets like this?

If we're trying to test the hybrid behavior, then you should do something like

```python
from modin.config import context as config_context

...

with config_context(AutoSwitchBackend=True):
    (ordered_df1, row_position_quoted_identifier) = create_initial_ordered_dataframe(...)
```

etc. to ensure the estimation code is run when we expect it to be.

tests/integ/modin/test_df_to_snowpark_pandas.py
```diff
@@ -141,9 +141,10 @@ def test_to_snowpark_pandas_with_operations(
 def test_to_snowpark_pandas_duplicated_columns_raises(
     session, tmp_table_basic, enforce_ordering
 ) -> None:
-    sql_simplifier_enabled_original = session.sql_simplifier_enabled
-    # Error is raised only when SQL simplifier is enabled.
-    session.sql_simplifier_enabled = True
+    with SqlCounter(query_count=0):
```
sfc-gh-joshi: Is this extra block necessary? The counter for the whole function is already 0.

sfc-gh-joshi (changes requested):

tests/integ/modin/test_ordered_dataframe.py
sfc-gh-helmeleegy: This is similar to how the initial change looked like. @sfc-gh-jkew preferred to have the test not dependent on hybrid. I'm fine either way.

That said, calling `count_rows` and enabling hybrid will both have the same effect of triggering an eager query to compute the exact initial count. The test then verifies that the followup operations result in expected count estimates and upper bounds.

sfc-gh-joshi: IMO the question this test is asking is "does initializing the ordered dataframe [in hybrid mode] produce the expected row estimates?" Explicitly setting `row_count` and `row_count_upper_bound` breaks the abstraction barrier of the OrderedDataFrame constructor, and is basically asking "does the `count_rows` function return the expected result?" I can imagine a world where we would want to change the implementation of count estimation in the future and this would matter.

WDYT @sfc-gh-jkew ?

sfc-gh-helmeleegy: And the followup question this test asks is: "And are the row estimates maintained as expected as various operations are applied to the initial ordered dataframe?"

sfc-gh-jkew: upper bound row estimates can be useful outside of hybrid, specifically for join cardinality estimations and query tree pruning. Specifically this can help us identify when GS query costs might become greater than XP execution costs.

The test only tests the estimator; which can be initialized in a number of different ways.

sfc-gh-jkew: Also; I'm basically following up with a PR that will simply call "SHOW OBJECT.." to get the initial row count; and that will be universally applicable whether hybrid is enabled or not.

tests/integ/modin/test_df_to_snowpark_pandas.py
sfc-gh-helmeleegy: Fixed.

CHANGELOG.md
sfc-gh-jkew: We still always perform row count estimation and on aggregates and limit clauses we can set it to a known value. Maybe it should be "Set the initial row count only when hybrid is enabled"

tests/integ/modin/test_ordered_dataframe.py
sfc-gh-jkew: Here's the WIP change to fetch the row count from metadata, but I need to add more tests for views, materialized views, incomplete schema information, etc. https://github.com/snowflakedb/snowpark-python/pull/3711

sfc-gh-joshi (approved):

PR #3709, opened 2025-08-27
SNOW-641374: support directory

sfc-gh-aling:
   Fixes SNOW-641374

support https://docs.snowflake.com/en/user-guide/data-load-dirtables-query

a thin wrapper of `SELECT * FROM DIRECTORY( @<stage_name> )`

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -4735,17 +4735,15 @@ class Session:
             - ``FILE_URL``: Snowflake file URL to access the file
 
         Examples::
-            >>> session.sql("create or replace temp stage test_stage")
-
             >>> # Get all file metadata from a stage named 'mystage'
-            >>> df = session.directory('test_stage')
-            >>> df.show()
+            >>> df = session.directory('test_stage') # doctest: +SKIP
+            >>> df.show() # doctest: +SKIP
 
             >>> # Get file URLs for CSV files only
-            >>> csv_files = session.directory('test_stage').filter(
+            >>> csv_files = session.directory('test_stage').filter( # doctest: +SKIP
             ...     col('RELATIVE_PATH').like('%.csv%')
             ... ).select('FILE_URL')
-            >>> csv_files.show()
+            >>> csv_files.show() # doctest: +SKIP
 
         For details, see the Snowflake documentation on
         `Snowflake Directory Tables Documentation <https://docs.snowflake.com/en/user-guide/data-load-dirtables-query>`_
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -4744,7 +4744,7 @@ class Session:
             >>> # List all files in the stage
             >>> df = session.directory('test_stage')
             >>> df.count()
-            3
+            2
 
             >>> # Get file URLs for CSV files only
             >>> csv_files = session.directory('test_stage').filter(
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -4707,3 +4707,58 @@ class Session:
             set_api_call_source(df, "Session.call")
             # Note the collect is implicit within the stored procedure call, so should not emit_ast here.
             return df.collect(statement_params=statement_params, _emit_ast=False)[0][0]
+
+    def directory(self, stage_name: str, _emit_ast: bool = True) -> DataFrame:
+        """
+        Returns a DataFrame representing the results of a directory table query on the specified stage.
+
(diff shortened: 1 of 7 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -4707,3 +4707,68 @@ def _execute_sproc_internal(
...
+            >>> csv_files.show()
+            [ROW(FILE_URL='tests/resources/testCSV.csv')]
+
+        For details, see the Snowflake documentation on
+        `Snowflake Directory Tables Documentation <https://docs.snowflake.com/en/user-guide/data-load-dirtables-query>`_
+        """
+        stage_name = (
+            stage_name
+            if stage_name.startswith(STAGE_PREFIX)
+            else f"{STAGE_PREFIX}{stage_name}"
```
sfc-gh-jrose, line 4767: Should this work with `snow://` type prefixes?

sfc-gh-aling: good call, I tried the following combinations but none of them worked:

```sql
select * from directory('SNOW://TEST_STAGE'); -- Argument 1 to function 'DIRECTORY' does not start with '@'. Please specify stage name as '@<stage_name>'.
select * from directory(SNOW://TEST_STAGE); -- Syntax error: unexpected 'SNOW'. (line 28)
select * from directory(@SNOW://TEST_STAGE); -- Syntax error: unexpected '@SNOW://TEST_STAGE'. (line 29)
select * from directory('@SNOW://TEST_STAGE'); -- Argument 1 to function 'DIRECTORY' should only contain the stage name and not a path.
```

the directory table doc doesn't say it supports snow:// prefix, or you know any other form that might work?

sfc-gh-jrose: Those are all the forms I would have tried. It might require the stage be created as a dataset rather then as a stage. Someone from the ML team might be more familiar with this use case.

sfc-gh-aling: this looks like another rabbit hole -- I think we can delay the support until there is an actual request? should be fairly easy to add

sfc-gh-heshah (approved):

sfc-gh-mayliu (approved):

tests/integ/test_session.py
```diff
@@ -919,3 +929,90 @@ def test_session_atexit(db_parameters):
...
+        session.sql(f"ALTER STAGE {temp_stage} REFRESH").collect()
+        session.sql(f"ALTER STAGE {temp_stage_special} REFRESH").collect()
+
+        EXPECTED_SCHEMA = StructType(
+            [
+                StructField("RELATIVE_PATH", StringType(), nullable=True),
+                StructField("SIZE", LongType(), nullable=True),
+                StructField(
+                    "LAST_MODIFIED",
+                    TimestampType(timezone=TimestampTimeZone.TZ),
```
sfc-gh-mayliu, line 975: The doc mentioned `LAST_MODIFIED` would be returned as `LTZ` (https://docs.snowflake.com/en/user-guide/data-load-dirtables-query#output)

> LAST_MODIFIED | TIMESTAMP_LTZ | Timestamp when the file was last updated in the stage.


would TZ here cause potential discrepancy?

tests/integ/test_session.py
```diff
@@ -919,3 +929,90 @@ def test_session_atexit(db_parameters):
...
+        df_filtered = directory_df.where(
+            col("RELATIVE_PATH").startswith("file1")
+        ).select(col("SIZE"))
+        assert len(df_filtered.collect()) == 1 and df_filtered.schema == StructType(
+            [StructField("SIZE", LongType(), nullable=True)]
+        )
+
+        # Test directory() method on special stage name (with quotes)
+        directory_df_special = session.directory(temp_stage_special)
+        assert directory_df_special.schema == EXPECTED_SCHEMA
```
sfc-gh-mayliu, line 998: since `directory` outputs columns matching `EXPECTED_SCHEMA`, would the column order always preserves? i.e. if the returned columns start with `FILE_URL` instead of `RELATIVE_PATH`, would it fail this test?

tests/integ/test_session.py
sfc-gh-aling: I tested with snowsight (pure SQL) on aws and gcp and the type of the column is TZ not LTZ. also I confirmed in the python client that server is returning timestamp TZ for this column.

this might be a doc issue, I will create a jira for the storage team to take a look

tests/integ/test_session.py
sfc-gh-aling: yes, the order should preserve. changing the order of columns would be a breaking change

PR #3711, opened 2025-08-28
[WIP] Get the row count from metadata directly rather than through a count(*) call

sfc-gh-jkew:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/utils.py
@@ -31,7 +31,7 @@ from snowflake.snowpark._internal.analyzer.analyzer_utils import (
     quote_name_without_upper_casing,
 )
 from snowflake.snowpark._internal.analyzer.expression import Literal
-from snowflake.snowpark._internal.type_utils import LiteralType as SnowparkLiteralType
+from snowflake.snowpark._internal.type_utils import LiteralType
 from snowflake.snowpark._internal.utils import (
     SNOWFLAKE_OBJECT_RE_PATTERN,
     TempObjectType,
@@ -1946,8 +1946,9 @@ def get_object_metadata_row_count(table_name: str) -> Optional[int]:
     # quote db and schema to handle special characters.
     quoted_db = quote_name_without_upper_casing(db)
     quoted_schema = quote_name_without_upper_casing(schema)
-
-    res = session.sql(f"SHOW OBJECTS LIKE '{table}' IN SCHEMA {quoted_db}.{quoted_schema} LIMIT 1").collect(
+    query = f"SHOW OBJECTS LIKE '{table}' IN SCHEMA {db}.{schema} LIMIT 1"
+    #breakpoint()
+    res = session.sql(query).collect(
         statement_params=get_default_snowpark_pandas_statement_params()
     )
     if len(res) != 1:
--- a/src/snowflake/snowpark/modin/plugin/_internal/utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/utils.py
@@ -8,7 +8,7 @@ import re
 import traceback
 from collections.abc import Hashable, Iterable, Sequence
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Optional, Union
+from typing import TYPE_CHECKING, Any, Optional, Union, TypeAlias
 from packaging import version  # noqa: E402,F401
 
 import modin.pandas as pd
@@ -31,12 +31,13 @@ from snowflake.snowpark._internal.analyzer.analyzer_utils import (
     quote_name_without_upper_casing,
 )
 from snowflake.snowpark._internal.analyzer.expression import Literal
-from snowflake.snowpark._internal.type_utils import LiteralType
+from snowflake.snowpark._internal.type_utils import LiteralType as SnowparkLiteralType
 from snowflake.snowpark._internal.utils import (
     SNOWFLAKE_OBJECT_RE_PATTERN,
     TempObjectType,
     generate_random_alphanumeric,
     get_temp_type_for_object,
+    parse_table_name,
     random_name_for_temp_object,
 )
 from snowflake.snowpark.column import Column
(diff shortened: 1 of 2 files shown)


tests/integ/modin/test_utils.py
```diff
@@ -163,3 +165,77 @@ def test_ensure_index(data):
...
+    num_rows = 5
+    # Use temporary table to ensure it's cleaned up after session.
+    session.create_dataframe(
+        [[i] for i in range(num_rows)], schema=["a"]
+    ).write.save_as_table(
+        table_name_with_rows, mode="overwrite", table_type="temporary"
+    )
+
+    # Test with different qualifications
+    current_db = session.get_current_database()
```
sfc-gh-helmeleegy, line 186: Can we also add tests where the table/view's db/schema are different from the current db/schema?

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1914,6 +1915,62 @@ def count_rows(df: OrderedDataFrame) -> int:
     return row_count
 
 
+def get_object_metadata_row_count(object_name: str) -> Optional[int]:
```
sfc-gh-joshi, line 1911: Is this function used anywhere in this PR?

sfc-gh-jkew: Not yet, no - I would be using it as a replacement for count(*)

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1914,6 +1915,62 @@ def count_rows(df: OrderedDataFrame) -> int:
...
+        db, schema, table = parts[0], parts[1], parts[2]
+    else:
+        return None
+
+    if not db:
+        return None
+    if not schema:
+        return None
+
+    query = f"SHOW OBJECTS LIKE '{table}' IN SCHEMA {db}.{schema} LIMIT 1"
```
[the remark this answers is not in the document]
sfc-gh-jkew: SQL Injection is prevented in few ways here:
* The input string has been parsed with parse_table_name, which also performs a verification of the format.
* The underlying snowflake-connector does not allow for execution of multiple statements in a single execution
* The API in which this will be used also accepts free-form queries as well as table names by design
* This method would be called specifically from a branch of code which is unique to table names.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-joshi: I think it would be good to have the call site for this function get included in the same PR, since there could be some additional nuance to its usage since we sometimes need to get row count on queries that aren't just a simple table name.

I just remembered that there's actually an old PR from Rehan that did something similar, though it performed COUNT queries instead of reading metadata: https://github.com/snowflakedb/snowpark-python/pull/2760/files

I'm not sure about the SQL internals here, but I would expect a `SELECT COUNT(*) FROM table_name` to be optimized to read from the relevant metadata field if present, which would eliminate the need to parse the schema name from a table.

sfc-gh-jkew: The issue we are solving is to avoid doing a count(*) on a view; which would have the same syntax as above but it would issue a long-running query. So we have to differentiate between the types of object.

sfc-gh-joshi: Makes sense. I still think it would be good to include the code to call this function in the same PR, since I suspect it might cause some changes.

sfc-gh-jkew: @sfc-gh-joshi Ok. made use of the function. I also made a few modifications to make sure that the SHOW OBJECT call itself would not cause any issues if there is a permissions issue. One of the weirder impacts of this is that if you have permissions to the metadata but not the rows in the table itself len(df) could return the table size but df.head() would print zero rows.

I don't think that is something we need to worry about too much, but it's interesting and maybe a little weird.

CHANGELOG.md
```diff
@@ -116,6 +116,7 @@
 - Set the default transfer limit in hybrid execution for data leaving Snowflake to 100k, which can be overridden with the SnowflakePandasTransferThreshold environment variable. This configuration is appropriate for scenarios with two available engines, "Pandas" and "Snowflake" on relational workloads.
 - Improve import error message by adding '--upgrade' to 'pip install "snowflake-snowpark-python[modin]"' in the error message.
 - Reduce the telemetry messages from the modin client by pre-aggregating into 5 second windows and only keeping a narrow band of metrics which are useful for tracking hybrid execution and native pandas performance.
+- Add internal functionality for retrieving object metadata through SHOW OBJECTS
```
sfc-gh-joshi: move to section for next release (1.39)

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -1914,6 +1908,69 @@ def count_rows(df: OrderedDataFrame) -> int:
...
+        return None  # pragma: no cover
+    if not schema:
+        return None  # pragma: no cover
+
+    try:
+        query = f"SHOW OBJECTS LIKE '{table}' IN SCHEMA {db}.{schema} LIMIT 1"
+        res = session.sql(query).collect(
+            statement_params=get_default_snowpark_pandas_statement_params()
+        )
+    except Exception:
```
sfc-gh-joshi: Should we make the exception class more specific so we fail loudly on unexpected errors?

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -499,16 +500,9 @@ def create_initial_ordered_dataframe(
...
-        # Set the materialized row count
-        materialized_row_count = (
-            initial_ordered_dataframe._dataframe_ref.snowpark_dataframe.count(
-                statement_params=get_default_snowpark_pandas_statement_params(),
-                _emit_ast=False,
-            )
-        )
+    materialized_row_count = None
+    if not is_query:
+        materialized_row_count = get_object_metadata_row_count(table_name_or_query)
```
sfc-gh-joshi, line 505: If this fails, do we want a fallback if hybrid is enabled that performs an explicit COUNT query?

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-joshi: > One of the weirder impacts of this is that if you have permissions to the metadata but not the rows in the table itself len(df) could return the table size but df.head() would print zero rows.

Pretty weird indeed. I guess it's fine since it doesn't leak any information (if the user has metadata access but not row access, they'd be able to find the row count via SQL anyway), so I agree it's not an issue.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-jkew: No, I don't think so - because the behavior for queries really should be changed as we discussed this morning. This code always assumed that getting the dataset size should be cheap, and I think we want to keep it that way.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-jkew: Narrowed it to SnowflakeSQLException; which is the one I expect the most

CHANGELOG.md
sfc-gh-jkew: I always forget this. I wish there was a more... linear way of managing the changelog.

sfc-gh-joshi (approved):

sfc-gh-jrose (approved):

PR #3715, opened 2025-08-28
SNOW-2304052 - Create a test parameter to optionally enable hybrid tests for the integration module

sfc-gh-jkew:
SNOW-2304052 - Create a test parameter to optionally enable hybrid tests for the integration module

The new test parameter is called '--enable_modin_hybrid_mode' which is only applied to the integ
modin module. This is not used yet; but it allows for enabling hybrid in an adhoc way. Eventually there
will be a new pre-commit test which enables hybrid just for the integration modin module.

This change also disables the sql_counter when running under hybrid mode; because virtually no sql queries are issued.


sfc-gh-jrose (approved):

tests/integ/modin/conftest.py
```diff
@@ -115,6 +121,7 @@ def check_sql_counter_invoked(request):
         # on Jenkins.
         and not running_on_jenkins()
         and not SKIP_SQL_COUNT_CHECK
+        and not AutoSwitchBackend.get()  # disable sql_counter for hybrid mode tests
```
sfc-gh-joshi: Can we have this read the pytest flag instead of the modin config variable? I think we still want the SQL counter to run in `tests/integ/modin/hybrid/`, where query counts would be helpful in ensuring we switch exactly where we mean to switch.

tests/integ/utils/sql_counter.py
```diff
@@ -148,7 +149,12 @@ def __init__(
         # Bypassing sql counter since
         #   1. it is an unnecessary metric for tests running in stored procedures
         #   2. pytest-assume package is not available in conda
-        self._no_check = no_check or IS_IN_STORED_PROC or SKIP_SQL_COUNT_CHECK
+        self._no_check = (
+            no_check
+            or IS_IN_STORED_PROC
+            or SKIP_SQL_COUNT_CHECK
+            or AutoSwitchBackend.get()
```
sfc-gh-joshi: as above, can we use the pytest flag instead?

tests/integ/modin/conftest.py
sfc-gh-helmeleegy: +1

sfc-gh-jkew: We can't directly use the pytest flag here; but I added a global variable similar to SKIP_SQL_COUNT_CHECK which essentially does the same thing.

sfc-gh-joshi (approved):

tests/integ/utils/sql_counter.py
```diff
@@ -139,6 +139,7 @@ def __init__(
         **kwargs,
     ) -> "SqlCounter":
         from tests.conftest import SKIP_SQL_COUNT_CHECK
+        from tests.integ.modin.conftest import MODIN_HYBRID_TEST_MODE_ENABLED
```
sfc-gh-jkew: This eventually fails in CI because of a modin import issue on certain pre-commit runs. So I need to move it a bit.

PR #3741, opened 2025-09-04
SNOW-2273496: Enable Cortex Complete function for apply

sfc-gh-jenzhang:

   

   Fixes [SNOW-2273496](https://snowflakecomputing.atlassian.net/browse/SNOW-2273496)

   My code solved the issue by adding the cortex complete function in the supported list and making logic change the parameter mapping to ensure kwargs are supported as well for the complete function. I added unit tests to ensure it is working.

[SNOW-2273496]: https://snowflakecomputing.atlassian.net/browse/SNOW-2273496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CHANGELOG.md
```diff
@@ -610,6 +610,7 @@
 - Added support for list values in `Series.str.slice`.
 - Added support for applying Snowflake Cortex functions `ClassifyText`, `Translate`, and `ExtractAnswer`.
 - Added support for `Series.hist`.
+- Added support for applying Snowflake Cortex functions `Complete`.
```
sfc-gh-nkrishna: We actually want to put this line at the top of the CHANGELOG, I believe it's currently `1.39.0` under `Snowpark pandas API Updates, New Features` as that is the log for the release and this code would only be in the next release not in the last one.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -244,3 +234,136 @@ def test_apply_snowflake_cortex_negative(session, is_series, operation):
...
+
+        print("response: ", response)
+
+        assert isinstance(response, str)
+        assert "2012" in response
+
+
+@pytest.mark.skipif(
+    running_on_jenkins(),
+    reason="TODO: SNOW-1859087 snowflake.cortex.complete SSL error",
```
sfc-gh-nkrishna: @sfc-gh-lmukhopadhyay Do you know if the SSL error is a result of it being a preview feature or do we actually need to go fix the cert in Jenkins?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9393,37 +9393,53 @@ def sf_function(col: SnowparkColumn) -> SnowparkColumn:
...
+                        # Use default value for optional positional params
+                        if not is_kw_only:
+                            resolved_positional.append(param.default)
+                        # For keyword-only with default, omit since default will be used
                     else:
+                        # Missing required param (non-keyword-only)
                         ErrorMessage.not_implemented(
-                            f"Unspecified Argument: {arg} - when using apply with kwargs, all function arguments should be specified except the single column reference (if applicable)."
+                            f"Unspecified Argument: {arg_name} - "
+                            "all function arguments should be specified except the single column reference (if applicable)."
```
sfc-gh-nkrishna: I'd actually keep the `when using apply` bit because this code will be executed lazily by the user and sometimes it can be hard to track what line the original error is propagating from. The Sparkle project (AST resolution) should help with this, but until specifying the operation in the error message is a working hack.

CHANGELOG.md
sfc-gh-jenzhang: Yup, I moved it to the newest release

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: Got it! I added it back.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9393,37 +9393,52 @@ def sf_function(col: SnowparkColumn) -> SnowparkColumn:
...
+            # Track keys in kwargs that are not yet processed
+            unprocessed_keys = set(kwargs.keys())
+
+            for arg_name, param in params.items():
+                # Skip special internal Snowpark params always
+                if arg_name in ("_emit_ast", "_ast"):
+                    # Always remove internal params from unprocessed_keys (safe even if not present)
+                    unprocessed_keys.discard(arg_name)
+                    continue
+
```
sfc-gh-jenzhang, line 9411: Hey @sfc-gh-lmukhopadhyay! Wanted to confirm the expected behaviour of "_emit_ast" and "_ast" parameters . I saw the old code skipped them when not found but also appended them to the positional arguments list. Also saw the old comment ([ticket](https://snowflakecomputing.atlassian.net/browse/SNOW-1927811)) saying that we shouldn't be adding them to the positional arguments since they will be passed by Snowpark Python. I have skipped them completely but just wanted to confirm what would be the expected behaviour here since you have more context :))

tests/integ/modin/test_apply_snowflake_cortex_functions.py
sfc-gh-lmukhopadhyay: IIRC the cortex functions were not supported in certain deployments like preprod but I'm not sure if that's still the case. Maybe we can remove the jenkins skip for the tests temporarily and use this branch to run the prod and preprod Jenkins tests @sfc-gh-jenzhang?
[prod](https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkPythonSnowPandasDailyRegressRunner/build?delay=0sec)
[preprod](https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkPythonSnowPandasDailyRegressRunner/build?delay=0sec)

CHANGELOG.md
```diff
@@ -37,6 +37,8 @@
 
 #### Improvements
 
+- Added support for applying Snowflake Cortex functions `Complete`.
```
sfc-gh-lmukhopadhyay: nit
```suggestion
- Added support for applying Snowflake Cortex function `Complete`.
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-lmukhopadhyay: Yes this should be the correct behavior, thanks for updating! The _emit_ast positional arg issue hasn't been tackled yet as far as I know

sfc-gh-lmukhopadhyay (changes requested):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: Got it, thanks!

tests/integ/modin/test_apply_snowflake_cortex_functions.py
sfc-gh-joshi: It might be easier to run the test locally with preprod account credentials. The preprod jenkins runs have also been failing for quite a while now, so if these tests fail in those environments it's unlikely anybody would notice.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -244,3 +234,136 @@ def test_apply_snowflake_cortex_negative(session, is_series, operation):
...
+):
+    content = "When was Snowflake founded?"
+    modin_input = (pd.Series if is_series else pd.DataFrame)([content])
+
+    with SqlCounter(query_count=query_count):
+        result = modin_input.apply(Complete, **kwargs)
+        if is_series:
+            response = result.iloc[0]
+        else:
+            response = result[0][0]
```
sfc-gh-joshi:
```suggestion
            response = result.iloc[0, 0]
```

This will take 1 query instead of 2, so you won't need to parametrize the tests on query_count anymore.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
```diff
@@ -244,3 +234,136 @@ def test_apply_snowflake_cortex_negative(session, is_series, operation):
...
+        ),
+    ],
+)
+def test_apply_snowflake_cortex_complete_error_cases(
+    session, is_series, test_case, kwargs, expected_error
+):
+    content = "When was Snowflake founded?"
+    modin_input = (pd.Series if is_series else pd.DataFrame)([content])
+
+    with pytest.raises((NotImplementedError, ValueError, TypeError)) as exc_info:
```
sfc-gh-joshi: Can you make the expected failure type into a test parameter? That makes it easier to tell what the expected behavior is, and if the cortex API ever changes it'll also be easier to update our tests to match.

Example:
https://github.com/snowflakedb/snowpark-python/blob/f71dfa80fbbbddbcdd92625ef0a4add6888fdda4/tests/integ/modin/frame/test_loc.py#L2241-L2255

sfc-gh-jenzhang: Yup that makes sense. I've updated the test params.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
sfc-gh-jenzhang: Ahh I see! I've removed the query count.

tests/integ/modin/test_apply_snowflake_cortex_functions.py
sfc-gh-jenzhang: I removed all the ssl error skipifs and ran the tests locally with preprod credentials and they all passed. Seems like we can safely remove all of them? Will run the jenkins job to double check again

sfc-gh-jenzhang: The [preprod](https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkPythonSnowPandasDailyRegressRunner/198189/) and [prod](https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkPythonSnowPandasDailyRegressRunner/198188/) run were successful with all the SSL errors removed. So I think we can safely remove all of them?

sfc-gh-nkrishna: Yes let's do that please, thanks!

sfc-gh-lmukhopadhyay: @sfc-gh-jenzhang Thanks for checking via the jenkins runs! One thing, would you be able to run the preprod jenkins run with this connection_param_file? `scripts/parameters.py.gpg` I think the one used in the run is the prod parameters even though deployment_type was set as preprod

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9393,37 +9393,52 @@ def sf_function(col: SnowparkColumn) -> SnowparkColumn:
...
+            # Track keys in kwargs that are not yet processed
+            unprocessed_keys = set(kwargs.keys())
+
+            for arg_name, param in params.items():
+                # Skip special internal Snowpark params always
+                if arg_name in ("_emit_ast", "_ast"):
+                    # Always remove internal params from unprocessed_keys (safe even if not present)
+                    unprocessed_keys.discard(arg_name)
+                    continue
```
sfc-gh-heshah, line 9410: This seems a little risky to me at a glance. From what I understand, the previous version of this method would still inject `_emit_ast` and `_ast` into the eventual function call if the caller of `_apply_snowflake_function_to_columns` included the parameters in the `kwargs` argument, and the `snowflake_function` also had those parameters in the function signature.

The new behavior is to skip `_emit_ast` and `_ast` even when provided to `_apply_snowflake_function_to_columns`, which I believe @sfc-gh-lmukhopadhyay agreed is the correct behavior in a separate comment. 

As long as we're confident `_emit_ast` and `_ast` are never provided when calling `_apply_snowflake_function_to_columns`, this should be fine, but we're also making the assumption that they will never be provided in the future. These are pretty important internal parameters that we pass around to avoid breaking AST generation when enabled, so I hesitate to say its okay to discard them entirely.

Since Snowpark Pandas is not planned to have AST support for the time being, I think it would be safer to always pass `_emit_ast=False` and `_ast=None` whenever the `snowflake_function` accepts these parameters. cc: @sfc-gh-nkrishna @sfc-gh-jkew

sfc-gh-nkrishna: Makes sense to to me to pass them/set them to False/None cc @sfc-gh-jenzhang

sfc-gh-jenzhang: Thanks for taking a look @sfc-gh-heshah!  That makes sense. I've made the change you suggested.

PR #3746, opened 2025-09-04
SNOW-2318100: Downgrade the log message about DataFrame reference updates.

sfc-gh-mvashishtha:
e.g. prior to this commit, try running `pytest tests/integ/modin/frame/test_quantile.py` locally and you get 152 messages like  "The Snowpark DataFrame in DataFrameReference with id=[...] is updated"  for 24 test cases. I don’t think everyone who works on pandas even know what this message means. I know what it means, but I have never used it. It makes it difficult to debug many test failures because it often appears many times in a single test case.

This commit changes the level of the message to `DEBUG - 1` so it doesn't appear by default when we run `pytest`.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
@@ -680,6 +680,11 @@ class OrderedDataFrame:
             dataframe_ref.cached_snowflake_quoted_identifiers_tuple = tuple(
                 new_column_identifiers
             )
+            _logger.log(
+                # log at level DEBUG - 1 because we make this update very often.
+                level=logging.DEBUG - 1,
+                msg=f"The Snowpark DataFrame in DataFrameReference with id={dataframe_ref._id} is updated",
+            )
 
         new_df = OrderedDataFrame(
             dataframe_ref,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,8 +36,9 @@
 #### New Features
 
 #### Improvements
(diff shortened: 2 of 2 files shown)


src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
```diff
@@ -680,8 +680,10 @@ def select(
             dataframe_ref.cached_snowflake_quoted_identifiers_tuple = tuple(
                 new_column_identifiers
             )
-            _logger.debug(
-                f"The Snowpark DataFrame in DataFrameReference with id={dataframe_ref._id} is updated"
+            _logger.log(
+                # log at level DEBUG - 1 because we make this update very often.
+                level=logging.DEBUG - 1,
```
sfc-gh-jkew, line 685: DEBUG-1 is weird to me; because I think that means "9"
> https://docs.python.org/3/library/logging.html#logging.DEBUG

Is this common? Why not remove the message entirely?

Really python needs a TRACE level.

sfc-gh-helmeleegy (approved):

src/snowflake/snowpark/modin/plugin/_internal/ordered_dataframe.py
sfc-gh-mvashishtha: > Is this common?

I don't think so.

> Why not remove the message entirely?

In case someone needs to use this info, they can change the logging level during a pytest run.

> Really python needs a TRACE level.

That sounds like a good idea! As you guessed, this is me rolling my own TRACE level.

PR #3766, opened 2025-09-11
Fix drop_duplicates issue due to mismatching row positions in dfs based on same data source

sfc-gh-helmeleegy:

   

   Fixes SNOW-2328061

   Fix drop_duplicates issue due to mismatching row positions in dfs based on same data source. Enable CTE optimization to make sure data source is read once.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@
 
 #### Bug Fixes
 
+- Fixed an issue with drop_duplicates where the same data source could be read multiple times in the same query but in a different order each time, resulting in missing rows in the final result. The fix ensures that the data source is read only once.
+
 ### Snowpark Local Testing Updates
 
 #### New Features


tests/integ/modin/frame/test_drop_duplicates.py
```diff
@@ -79,6 +81,29 @@ def test_drop_duplicates(subset, keep, ignore_index):
...
+def test_drop_duplicates_after_read_snowflake(session):
+    pandas_df = native_pd.DataFrame(
+        {"A": [0, 1, 1, 2, 0], "B": ["a", "b", "c", "b", "a"]}
+    )
+    table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
+    session.create_dataframe(pandas_df).write.save_as_table(
+        table_name, table_type="temp"
+    )
+    # Simulate random order each time read_snowflake is run.
+    snow_df = pd.read_snowflake(
```
sfc-gh-jkew, line 93: +1 this is a really interesting way of testing the problem

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -1514,6 +1514,17 @@ def drop_duplicates(
...
+
+    # Make sure CTE optimization is enabled.
+    # The reason this is required for drop_duplicates is that two dataframes need to be joined
+    # on their row position (one is used as a filter for the other) and while not identical,
+    # they both originate from the same source.
+    # Since read_snowflake can result in assigning row positions differently each time it's run,
+    # then if we compute the two dataframes independently, their row positions may not match.
+    # With the CTE optimization, we are guaranteed that reading the input source will only happen
+    # once in the finally generated query, and hence no mismatch in row positions will take place.
+    pd.session.cte_optimization_enabled = True
```
sfc-gh-jkew, line 1526: Would this impact the existing snowpark session - and do you have an idea on when the optimization will be a default?

sfc-gh-helmeleegy: It should be default soon. But it seems the team needs to do a bit more testing and they are now busy with the sparkle project. So it may get delayed because of that.

sfc-gh-joshi: Should we reset the flag back to its original value at the end of the function? I think we have some tests in the daily suite that explicitly toggle this flag, and if this test runs before other tests in the same session it may cause unexpected behavior.

sfc-gh-helmeleegy: It needs to remain after the function returns because the actual CTE optimization only kicks in at the time of query generation which can happen after several other operations are applied beyond just drop_duplicates.
For the daily tests, I think it should be the responsibility of each test to first preserve the existing setting of the flag, toggle it as needed, and then finally set it back to its original setting. Perhaps the only difference I can imagine is if there was a test testing some pandas pipeline that happens to use drop_duplicates and assumes that the optimization is disabled, it will get re-enabled as part of the execution. So we are currently forcing the optimization to be enabled for any pipeline using drop_duplicates.

PR #3768, opened 2025-09-11
SNOW-2331021: Wrap internal Series/DF creation

sfc-gh-joshi:

   

   Fixes SNOW-2331021

Many internal functions create an instance of `pd.Series` or `pd.DataFrame`, then retrieve a query compiler from the object. With hybrid execution enabled, this query compiler is sometimes an instance of `NativeQueryCompiler`, which may break invariants when a `SnowflakeQueryCompiler` is expected instead. The new helper function `new_snow_series` and `new_snow_df` disable hybrid mode when creating a Series/DF for internal use.

This PR replaces constructor calls within the SFQC and indexing_overrides, which were responsible for the bug in the original ticket. Other constructor calls remain, but I have not yet triaged them.

This removes a few hundred test failures in `modin_hybrid_integ_results.csv`.

### DIFF
--- a/.github/workflows/daily_precommit.yml
+++ b/.github/workflows/daily_precommit.yml
@@ -103,15 +103,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - image_name: ubuntu-latest-64-cores
-            download_name: linux
-          - image_name: macos-latest
-            download_name: macos
-          - image_name: windows-latest-64-cores
-            download_name: windows
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
-        cloud-provider: [aws, azure, gcp]
+        include:
+          # Ubuntu + rotating cloud providers
+          - { os: {image_name: ubuntu-latest-64-cores, download_name: linux}, python-version: "3.9", cloud-provider: aws }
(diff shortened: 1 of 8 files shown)


tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -591,3 +592,19 @@ def test_applying_cortex_function_causes_backend_switch(self, data_class, method
...
+
+@sql_count_checker(query_count=1, join_count=2)
+def test_switch_then_iloc():
+    # Switching backends then calling iloc should be valid.
+    # Prior to fixing SNOW-2331021, discrepancies with the index class caused an AssertionError.
+    df = pd.DataFrame([[0] * 10] * 10)
+    assert df.get_backend() == "Pandas"
+    # Should not error
+    assert_snowpark_pandas_equal_to_pandas(
+        df.move_to("Snowflake").iloc[[1, 3, 9], 1],
```
sfc-gh-helmeleegy, line 605: Is it worth it to test the switching in the opposite direction too?

sfc-gh-joshi: I added this test case because this line of code was failing specifically after switching native -> Snowflake. For completeness we should eventually test both directions, but for now I'm keeping the test to be explicitly the behavior that was failing.

PR #3770, opened 2025-09-11
SNOW-2241913: non-retryable error handling

sfc-gh-yuwang:

   

   Fixes SNOW-2241913

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,63 +19,80 @@
   - `get_cloud_provider_token`
 
 - Added support for the following scalar functions in `functions.py`:
-  - `array_remove_at`
-  - `as_boolean`
-  - `booland`
-  - `boolnot`
-  - `boolor`
-  - `boolor_agg`
-  - `boolxor`
-  - `chr`
-  - `decode`
-  - `div0null`
-  - `dp_interval_high`
-  - `dp_interval_low`
-  - `greatest_ignore_nulls`
-  - `h3_cell_to_boundary`
-  - `h3_cell_to_children`
-  - `h3_cell_to_children_string`
-  - `h3_cell_to_parent`
-  - `h3_cell_to_point`
-  - `h3_compact_cells`
-  - `h3_compact_cells_strings`
-  - `h3_coverage`
-  - `h3_coverage_strings`
-  - `h3_get_resolution`
-  - `h3_grid_disk`
-  - `h3_grid_distance`
-  - `h3_int_to_string`
-  - `h3_polygon_to_cells`
-  - `h3_polygon_to_cells_strings`
-  - `h3_string_to_int`
-  - `h3_try_grid_path`
-  - `h3_try_polygon_to_cells`
-  - `h3_try_polygon_to_cells_strings`
-  - `h3_uncompact_cells`
-  - `h3_uncompact_cells_strings`
-  - `haversine`
-  - `h3_grid_path`
-  - `h3_is_pentagon`
-  - `h3_is_valid_cell`
-  - `h3_latlng_to_cell`
-  - `h3_latlng_to_cell_string`
-  - `h3_point_to_cell`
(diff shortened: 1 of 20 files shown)


CHANGELOG.md
```diff
@@ -67,6 +67,7 @@
 - Hybrid execution mode is now enabled by default. Certain operations on smaller data will now automatically execute in native pandas in-memory. Use `from modin.config import AutoSwitchBackend; AutoSwitchBackend.disable()` to turn this off and force all execution to occur in Snowflake.
 - Added a session parameter `pandas_hybrid_execution_enabled` to enable/disable hybrid execution as an alternative to using `AutoSwitchBackend`.
 - Removed an unnecessary `SHOW OBJECTS` query issued from `read_snowflake` under certain conditions.
+- Add non-retryable failure in `DataFrameReader.dbapi`(PuPr), such as syntax error in external data source SQL.
```
sfc-gh-aling: nit:

1. this should be moved to snowpark python api updates
2. the description is a little confusing, you could describe it like improve dbapi not to retry on non-retryable error

src/snowflake/snowpark/exceptions.py
```diff
@@ -283,3 +283,24 @@ class SnowparkInvalidObjectNameException(SnowparkGeneralException):
     """
 
     pass
+
+
+class SnowparkDataSourceNonRetryableException(SnowparkGeneralException):
```
sfc-gh-aling: this makes the exception class public api. is this class for internal usage only to differentiate from other error? if so we can make it private by `_ SnowparkDataSourceNonRetryableException`.

another way to is set a private memvar on SnowparkDataframeReaderException to differentiate retry from non-retry

the name sound general and more for internal usage, and if we want something for public customer facing exception, I personally prefer exception classes with concrete meanings which I think would take more time to design

src/snowflake/snowpark/_internal/data_source/datasource_reader.py
sfc-gh-aling, line 1: what about ingestion (copy into), if we have SnowparkSQLException, should we retry?

sfc-gh-yuwang: I think non-retryable exception is for error that we know that it will not work even if we retry.
I think SnowparkSQLException is a very general failure that not only exception like SQL syntax error could trigger it.
For example, a udtf could fail with SnowparkSQLException because of python error that is retryable.

src/snowflake/snowpark/_internal/data_source/datasource_reader.py
```diff
@@ -85,6 +88,11 @@ def read(self, partition: str) -> Iterator[List[Any]]:
                         batch = []
             else:
                 raise ValueError("fetch size cannot be smaller than 0")
+        except Exception as exc:
+            if self.driver.non_retryable_error_checker(exc):
+                raise _SnowparkDataSourceNonRetryableException(exc)
```
sfc-gh-aling: as per our discussion, we do not need the private class and can just raise the existing reader exception here

tests/integ/datasource/test_oracledb.py
```diff
@@ -254,6 +254,18 @@ def test_unsupported_type():
...
+def test_oracledb_non_retryable_error(session):
+    with pytest.raises(
+        SnowparkDataframeReaderException,
+        match="ORA-00920: invalid relational operator",
+    ):
+        session.read.dbapi(
+            create_connection_oracledb,
+            table=ORACLEDB_TABLE_NAME,
+            predicates=["invalid syntax"],
+        ).collect()
```
sfc-gh-mayliu, line 267: Curious about why some tests use `.collect()` while others don't, my impression is that `dbapi()` is not lazily evaluated?

sfc-gh-mayliu (approved):

PR #3782, opened 2025-09-16
SNOW-2331792: Support DataFrame.eval for single-level index.

sfc-gh-mvashishtha:
Support all the parameters of `DataFrame.eval` via the Modin implementation. Only support single-level row index for now because the MultiIndex is currently a pandas object, so we have to get resolvers for its levels differently.


src/snowflake/snowpark/modin/plugin/utils/warning_message.py
```diff
@@ -131,3 +133,11 @@ def lost_type_warning(cls, operation: str, type: str) -> None:
         cls.single_warning(
             f"`{type}` may be lost in `{operation}`'s result, please use `astype` to convert the result type back."
         )
+
+
+@pytest.fixture
+def clear_printed_warnings() -> Generator[None, None, None]:
+    warnings = WarningMessage.printed_warnings
+    WarningMessage.printed_warnings.clear()
```
sfc-gh-joshi:
```suggestion
    warnings = set(WarningMessage.printed_warnings)
    WarningMessage.printed_warnings.clear()
```

I think these would point to the same set object, so you need to make a copy to actually save old warnings across the yield.

tests/integ/modin/frame/test_eval.py
```diff
@@ -0,0 +1,519 @@
...
+    "engine_kwargs",
+    [
+        param({"engine": "python"}, id="engine_python"),
+        param({"engine": "numexpr"}, id="engine_numexpr"),
+        param({}, id="default_engine"),
+    ],
+)
+
+
+def python_eval(self, expr, *, inplace=False, **kwargs):
```
sfc-gh-joshi: IMO it would be better to use a name other than `self` since this isn't a method.

tests/integ/modin/frame/test_eval.py
```diff
@@ -0,0 +1,519 @@
...
+@pytest.mark.parametrize(
+    "expr",
+    [
+        "CUSTOMER_KEY + 1",
+        "CUSTOMER_KEY + @global_int",
+        "CUSTOMER_KEY < @local_int",
+        "~(CUSTOMER_KEY < @global_int)",
+        "MARKET_SEGMENT == 'MACHINERY' or ACCOUNT_BALANCE > @global_int",
+        "AUTOMOBILE_OR_MACHINERY = MARKET_SEGMENT.isin(('AUTOMOBILE', 'MACHINERY'))",
+        """
```
sfc-gh-joshi, line 156: Should the block strings be wrapped in a `textwrap.dedent`?

tests/integ/modin/frame/test_eval.py
sfc-gh-joshi, line 1: Can you add some negative tests for very simple errors, like referencing a non-existent variable or having invalid syntax?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
```diff
@@ -226,9 +227,63 @@ def dot(self, other):  # noqa: PR01, RT01, D200
...
+    # eval() lets the user reference variables according to dynamic scope
+    # at `level` stack frames below the stack frame that called eval(). The
+    # eval() implementation is 4 stack frames above the frame where we execute
+    # the _eval() implementation:
+    # 1) query_compiler_caster wrapper dispatches to snowflake implementation
+    # 2) telemetry wrapper 1
+    # 3) telemetry wrapper 2 calls this implementation.
+    # 4) This method implementation calls the _eval() implementation
+    # so we add 4 to the `level` param.
+    kwargs["level"] = kwargs.get("level", 0) + 4
```
sfc-gh-jkew, line 273: Is it possible to call a nested eval? Or another internal function which then calls eval? I am wondering about assumptions for stack depth.

tests/integ/modin/frame/test_eval.py
```diff
@@ -0,0 +1,519 @@
...
+):
+    def level_0(df):
+        local_var = 0  # noqa: F841
+        return python_eval(df, "@local_var", level=level, **engine_kwargs)
+
+    def level_1(df):
+        local_var = 1  # noqa: F841
+        return level_0(df)
+
+    def level_2(df):
```
sfc-gh-jkew, line 377: cool. tests on stack depth

tests/integ/modin/frame/test_eval.py
sfc-gh-mvashishtha: I had a test for referencing a non-existent variable already in `test_user_error`, but I've now added a test case for invalid syntax.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
sfc-gh-mvashishtha: I think it's intentionally part of the API of `eval()` that it can only access variables in the frame that called it. `level` is the escape hatch for that behavior.

tests/integ/modin/frame/test_eval.py
sfc-gh-mvashishtha: Ideally, yes, but I think this is ok.

tests/integ/modin/frame/test_eval.py
sfc-gh-mvashishtha: Done.

src/snowflake/snowpark/modin/plugin/utils/warning_message.py
sfc-gh-mvashishtha: Done.

src/snowflake/snowpark/modin/plugin/_internal/utils.py
```diff
@@ -119,6 +119,11 @@
 
 _logger = logging.getLogger(__name__)
 
+# Flag guarding certain features available only in newer modin versions.
+# Snowpark pandas supports the newest two released versions of modin; update this flag and remove legacy
+# code as needed when we bump dependency versions.
+MODIN_IS_AT_LEAST_0_36_0 = version.parse(pd.__version__) >= version.parse("0.36.0")
```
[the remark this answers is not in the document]
sfc-gh-mvashishtha: they both work

sfc-gh-joshi (approved):

sfc-gh-jkew (approved):

src/snowflake/snowpark/modin/plugin/_internal/utils.py
sfc-gh-lmukhopadhyay, line 125: Is eval the only API affected by modin version >= 0.36? Maybe we could add a TODO in other places if not

sfc-gh-mvashishtha: It's the only one so far.

PR #3789, opened 2025-09-17
SNOW-2338542: Add a session parameter pandas_hybrid_execution_enabled

sfc-gh-helmeleegy:

   

   Fixes SNOW-2338542

   Add a session parameter pandas_hybrid_execution_enabled.

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -304,6 +304,7 @@ _PYTHON_SNOWPARK_INTERNAL_TELEMETRY_ENABLED = "ENABLE_SNOWPARK_FIRST_PARTY_TELEM
 _SNOWPARK_PANDAS_DUMMY_ROW_POS_OPTIMIZATION_ENABLED = (
     "SNOWPARK_PANDAS_DUMMY_ROW_POS_OPTIMIZATION_ENABLED"
 )
+_SNOWPARK_PANDAS_HYBRID_EXECUTION_ENABLED = "SNOWPARK_PANDAS_HYBRID_EXECUTION_ENABLED"
 
 # AST encoding.
 _PYTHON_SNOWPARK_USE_AST = "PYTHON_SNOWPARK_USE_AST"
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -1200,6 +1228,25 @@ def dummy_row_pos_optimization_enabled(self, value: bool) -> None:
...
+            # If modin is not installed, treat this method as a no-op.
+            return
+
+        from modin.config import AutoSwitchBackend
+
+        if value in [True, False]:
+            if value:
+                AutoSwitchBackend.enable()
+            else:
+                AutoSwitchBackend.disable()
```
sfc-gh-joshi:
```suggestion
            AutoSwitchBackend.put(value)
```

tests/integ/modin/hybrid/test_data_movement.py
sfc-gh-joshi, line 1: Can you put these tests in a different file? These tests are meant to test implementations of `move_to`/`move_from` methods, but the session parameter you added should have no bearing on them since it only affects the `AutoSwitchBackend` configuration.

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-joshi, line 1: Can you make these tests pytest parameters instead so we're not duplicating entire test cases?

You should also wrap `pd.session.pandas_hybrid_execution_enabled` sets with the `config_context(AutoSwitchBackend=True/False)` context manager. Otherwise, I'm not sure if the value of the config variable will be properly reset between tests.

src/snowflake/snowpark/session.py
```diff
@@ -1200,6 +1228,25 @@ def dummy_row_pos_optimization_enabled(self, value: bool) -> None:
                 "value for dummy_row_pos_optimization_enabled must be True or False!"
             )
 
+    @pandas_hybrid_execution_enabled.setter
+    def pandas_hybrid_execution_enabled(self, value: bool) -> None:
+        """Set the value for pandas_hybrid_execution_enabled"""
+        if not importlib.util.find_spec("modin"):
+            # If modin is not installed, treat this method as a no-op.
```
sfc-gh-joshi: Should it be an error instead? I can't think of a scenario where a user would intentionally set this flag without having modin installed.

tests/integ/modin/hybrid/test_data_movement.py
sfc-gh-helmeleegy: I removed those tests and kept the ones in test_switch_operations. I think they should be sufficient.

PR #3791, opened 2025-09-17
SNOW-2345238: Re-use row count from relaxed query compiler

sfc-gh-joshi:

   

   Fixes SNOW-2345238

SNOW-2230971 (#3687) introduced a second "relaxed" query compiler with placeholder row positions that exists in parallel to the existing Snowflake QC. When the Snowflake QC calls `create_initial_ordered_dataframe`, it may issue a metadata query to retrieve a row count if the frame is created from a Snowflake query on a table. However, after the addition of the relaxed QC, this code path was hit twice ([here](https://github.com/snowflakedb/snowpark-python/blob/3da00126cb8fe5d4198cadc5bcb03cfa825d1a9d/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py#L1458-L1482)), and a second redundant row count query was issued. This PR re-uses the result of the first row count query when possible to do so.

This regression was not caught by testing because the SQL counter [explicitly filters out](https://github.com/snowflakedb/snowpark-python/blob/3da00126cb8fe5d4198cadc5bcb03cfa825d1a9d/tests/integ/utils/sql_counter.py#L88-L89) row count metadata queries. Accordingly, this PR makes no testing changes because the row count reduction is not visible to the SQL counter as currently configured. Moreover, the regression was not caught by the benchmarking job (which does NOT filter row count queries) because it stopped running due to crashes caused by enabling hybrid execution on main. I verified the removal of the extra row count query manually, and its impact will be visible in the benchmark dashboard.

The removal of this query has a non-trivial impact on operations with short runtimes, where network latency becomes a meaningful factor. This will address the regression of single-API benchmarks like `pd.read_snowflake`, which went from ~0.2s to ~0.6s for a 1e6x10 table between commits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 #### New Features
 
+### Snowpark pandas API Updates
+
+#### New Features
+
+#### Improvements
+- Hybrid execution mode is now enabled by default. Certain operations on smaller data will now automatically execute in native pandas in-memory. Use `from modin.config import AutoSwitchBackend; AutoSwitchBackend.disable()` to turn this off and force all execution to occur in Snowflake.
+- Removed an unnecessary `SHOW OBJECTS` query issued from `read_snowflake` under certain conditions.
+
 ## 1.39.0 (YYYY-MM-DD)
 
 ### Snowpark Python API Updates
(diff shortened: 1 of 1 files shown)


sfc-gh-mvashishtha (approved):
thanks!

PR #3795, opened 2025-09-19
SNOW-800922: Support DataFrame.query for single-level index.

sfc-gh-mvashishtha:
Support all the parameters of `DataFrame.query` via the Modin implementation. Only support single-level row index for now because the MultiIndex is currently a pandas object, so we have to get resolvers for its levels differently.


src/snowflake/snowpark/modin/plugin/docstrings/dataframe.py
```diff
@@ -3294,7 +3294,133 @@ def quantile():
...
+            or punctuations (besides underscores) or starting with digits must be
+            surrounded by backticks. (For example, a column named "Area (cm^2)" would
+            be referenced as ```Area (cm^2)```). Column names which are Python keywords
+            (like "list", "for", "import", etc) cannot be used.
+
+            For example, if one of your columns is called ``a a`` and you want
+            to sum it with ``b``, your query should be ```a a` + b``.
+
+        inplace : bool
+            Whether to modify the DataFrame rather than creating a new one.
```
sfc-gh-joshi, line 3319:
```suggestion
            Whether to modify the DataFrame rather than creating a new one.

```

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
```diff
@@ -327,9 +318,37 @@ def prod(
...
+
+    key = self.eval(
+        expr,
+        inplace=False,
+        **(kwargs | {"level": kwargs.get("level", 0) + EVAL_QUERY_EXTRA_STACK_LEVELS}),
+    )
+
+    try:
+        result = self.loc[key]
+    except ValueError:
```
sfc-gh-joshi, line 348: Can you add a comment for what error we expect to catch?

tests/integ/modin/window/__init__.py
sfc-gh-joshi, line 1: empty file?

sfc-gh-joshi (changes requested):

src/snowflake/snowpark/modin/plugin/extensions/dataframe_overrides.py
sfc-gh-mvashishtha: Whoops, I forgot to carry over the comment from the modin implementation. Done.

tests/integ/modin/window/__init__.py
sfc-gh-mvashishtha: It seems that we [need](https://stackoverflow.com/a/51408950) an __init__.py in either this directory or in the new test_eval_and_query/ directory, so I added __init__.py to both places.

sfc-gh-joshi (approved):

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 ### Snowpark pandas API Updates
 
 #### New Features
+- Added support for `DataFrame.query()` for dataframes with single-level indexes.
```
sfc-gh-helmeleegy:
```suggestion
- Added support for `DataFrame.query` for dataframes with single-level indexes.
```

PR #3804, opened 2025-09-24
SNOW-2360274: Fix schema query sql generation for structured types

sfc-gh-jrose:

   

   Fixes SNOW-2360274

  The sql generation for describe queries has an edge case that can cause invalid sql to be generated when using nested structured types. The schema_expression function does it's best to respect nullability when generating schema queries. Snowflake does not support nullability inside of structured types so this should not be needed for structured types. This fix short circuits the sql generation by replacing the recursive call with NULLs. I have not found any cases in which this would break, but to be extra sure it's not a BCR I am gating it behind a context variable. In order to use this bugfix you will have to set `context. _enable_fix_2360274` to be True.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Bug Fixes
 
 - Fixed a bug that `DataFrame.limit()` fail if there is parameter binding in the executed SQL.
+- Fixed a bug in schema query generation that could cause invalid sql to be genrated when using nested structured types.
 
 #### New Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 #### Bug Fixes
 
 - Fixed a bug that `DataFrame.limit()` fail if there is parameter binding in the executed SQL.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 #### Bug Fixes
 
 - Fixed a bug that `DataFrame.limit()` fail if there is parameter binding in the executed SQL.
+- Fixed a bug in schema query generation that could cause invalid sql to be genrated when using nested structured types.
```
sfc-gh-aling: this is a private fix, I don't think we need to call it out in the changelog.
also shall we create a follow up task to see if the fix can be the default behavior after thorough testing.

sfc-gh-jrose: Without a changelog note there won't be anything in the release notes for 1.39.1. Do you think that's alright?

sfc-gh-mayliu (approved):

CHANGELOG.md
sfc-gh-aling: you are right..
explicitness is better.

PR #3820, opened 2025-09-29
SNOW-2372686: Write large_pandas_backend_df.to_snowflake() via parquet.

sfc-gh-mvashishtha:
To implement to_snowflake() for a Snowpark pandas dataframe on the pandas backend with large enough data, upload data via a parquet file instead of via a Snowpark dataframe. The Snowpark dataframe typically inserts values through parametrized SQL queries.

Benchmarking showed that a good threshold to switch to parquet was roughly 3 MB, so I've set that as the configurable default switching threshold. Performance of this approach seems to improve with dataset size. Exporting an 800 MB dataframe took about 55 seconds via parquet versus about 429 seconds via the old method, so we get over 7x speedup.

This work took me much longer than expected, so I am sending out this PR with just the changes for DataFrame.to_snowflake(). We can address Series.to_snowflake() separately.

We can take a similar approach to speed up pandas_backend_df.move_to('snowflake').

<details>
<summary>benchmark script with a 3XL warehouse</summary>

```python
import pandas as native_pd
import modin.pandas as pd
import numpy as np
import datetime
from tests.utils import Utils
from multiprocessing import Pool
import psutil



from time import perf_counter

def get_benchmark_result(input_data: native_pd.DataFrame, method: str):
    if method == "parquet":
        pd.session.pandas_to_snowflake_row_limit = 0
    else:
        pd.session.pandas_to_snowflake_row_limit = int(1e99)

    df = pd.DataFrame(input_data).set_backend('pandas').pin_backend()
    table = Utils.random_table_name()
    start = perf_counter()
    df.to_snowflake(table)
    end = perf_counter()
    Utils.drop_table(pd.session, table)
    return {
        "time": datetime.timedelta(seconds=end-start),
        "memory_usage": input_data.memory_usage().sum(),
        "num_rows": input_data.shape[0],
        "num_columns": input_data.shape[1],
        "method": method
    }



# num_processes = psutil.cpu_count(logical=False)
# for 
rows = []
for input_data in (
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e0), 1))),     
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e3), 1))),    
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e3), 3))),     
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e4), 1))),
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e4), 3))),    
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e5), 1))),    
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e5), 3))),     
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e6), 1))),
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e6), 3))),    
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e6), 10))),
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e7), 1))),
    native_pd.DataFrame(np.random.randint(low=0,high=1000,size=(int(1e7), 10))),    
):
    for method in ("parquet", "snowpark"):
        rows.append(get_benchmark_result(input_data, method=method))

native_pd.DataFrame(rows).to_parquet('to_snowflake_timing_2', index=False)
```

</details>

to_snowflake_timing_2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,7 @@
 - Hybrid execution mode is now enabled by default. Certain operations on smaller data will now automatically execute in native pandas in-memory. Use `from modin.config import AutoSwitchBackend; AutoSwitchBackend.disable()` to turn this off and force all execution to occur in Snowflake.
 - Added a session parameter `pandas_hybrid_execution_enabled` to enable/disable hybrid execution as an alternative to using `AutoSwitchBackend`.
 - Removed an unnecessary `SHOW OBJECTS` query issued from `read_snowflake` under certain conditions.
+- Improved performance of `DataFrame.to_snowflake` and `pd.to_snowflake(dataframe)` for large data by uploading data via a parquet file.
 
 ## 1.39.0 (2025-09-17)
 
--- a/src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
@@ -98,7 +98,7 @@ def to_snowflake(  # noqa: F811
         self.memory_usage(deep=False).sum()
         <= PandasToSnowflakeParquetThresholdBytes.get()
     ):
-        return self.set_backend("snowflake").to_snowflake(
+        return self.set_backend("Snowflake").to_snowflake(
             name=name,
             if_exists=if_exists,
             index=index,
--- a/src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
@@ -11,6 +11,14 @@ from collections.abc import Iterable
 import functools
 from typing import Any, List, Literal, Optional, Union
 
+from snowflake.snowpark._internal.analyzer.analyzer_utils import (
+    quote_name_without_upper_casing,
+)
+from snowflake.snowpark.modin.config.envvars import (
+    PandasToSnowflakeParquetThresholdBytes,
+)
+from snowflake.snowpark.modin.plugin.utils.warning_message import WarningMessage
+
 import modin.pandas as pd
 from modin.pandas.api.extensions import (
     register_dataframe_accessor as _register_dataframe_accessor,
@@ -21,6 +29,15 @@ from pandas._typing import IndexLabel
 from snowflake.snowpark._internal.type_utils import ColumnOrName
 from snowflake.snowpark.async_job import AsyncJob
 from snowflake.snowpark.dataframe import DataFrame as SnowparkDataFrame
+from snowflake.snowpark.modin.plugin._internal.snowpark_pandas_types import (
+    SnowparkPandasType,
+)
+from snowflake.snowpark.modin.plugin._internal.utils import (
+    extract_and_validate_index_labels_for_to_snowflake,
+    handle_if_exists_for_to_snowflake,
+    is_all_label_components_none,
+    validate_column_labels_for_to_snowflake,
+)
 from snowflake.snowpark.modin.plugin.extensions.utils import (
     add_cache_result_docstring,
     register_non_snowflake_accessors,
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -119,6 +119,7 @@
 - Hybrid execution mode is now enabled by default. Certain operations on smaller data will now automatically execute in native pandas in-memory. Use `from modin.config import AutoSwitchBackend; AutoSwitchBackend.disable()` to turn this off and force all execution to occur in Snowflake.
 - Added a session parameter `pandas_hybrid_execution_enabled` to enable/disable hybrid execution as an alternative to using `AutoSwitchBackend`.
 - Removed an unnecessary `SHOW OBJECTS` query issued from `read_snowflake` under certain conditions.
+- Improved performance of `DataFrame.to_snowflake` and `pd.to_snowflake(dataframe)` for large data by uploading data via a parquet file.
```
sfc-gh-joshi: Should mention the `PandasToSnowflakeParquetThresholdBytes` variable.

Also, should we make it a session parameter rather than a modin flag? This would only be encountered in Snowpark pandas.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -37,12 +53,123 @@
...
+        # Shallow memory usage may underestimate the memory usage of the
+        # dataframe, but deep memory usage can be expensive to compute.
+        # Since this threshold uses shallow memory usage, it may sometimes
+        # not use the parquet method in cases where the parquet method would be
+        # faster, especially if this frame contains deep data like Python
+        # strings.
+        self.memory_usage(deep=False).sum()
+        <= PandasToSnowflakeParquetThresholdBytes.get()
+    ):
+        return self.set_backend("snowflake").to_snowflake(
```
sfc-gh-jkew: "Snowflake" - does not really matter; but the convention is still to capitalize

sfc-gh-joshi:
```suggestion
        return self.set_backend("Snowflake").to_snowflake(
```

I know it's case-insensitive, but we should try to be consistent.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -37,12 +53,123 @@
...
+        table_type=table_type,
+    )
+
+    return None
+
+
 # Implementation note: Arguments names and types are kept consistent with pandas.DataFrame.to_sql
 @register_dataframe_accessor("to_snowflake")
-def to_snowflake(
+def to_snowflake(  # noqa: F811
```
sfc-gh-joshi: Is this because we're redefining `to_snowflake` here? Should we rename the new pandas-backend method to something like `pandas_to_snowflake` (while keeping the extension name as `to_snowflake`) to make sure we don't run into any possible confusion with this method called directly?

src/snowflake/snowpark/modin/plugin/extensions/pd_extensions.py
```diff
@@ -486,36 +515,42 @@ def to_snowflake(
...
-        - :func:`DataFrame.to_snowflake <modin.pandas.DataFrame.to_snowflake>`
-        - :func:`Series.to_snowflake <modin.pandas.Series.to_snowflake>`
-        - :func:`read_snowflake <modin.pandas.read_snowflake>`
-    """
-    return _check_obj_and_set_backend_to_snowflake(obj)._query_compiler.to_snowflake(
-        name, if_exists, index, index_label, table_type
-    )
+@_register_pd_accessor("to_snowflake", backend="Snowflake")
+@doc(_TO_SNOWFLAKE_DOC)
+def to_snowflake(  # noqa: F811
```
sfc-gh-joshi: as for DF: consider de-duplicating the method name

src/snowflake/snowpark/modin/plugin/extensions/pd_extensions.py
```diff
@@ -486,36 +515,42 @@ def to_snowflake(
...
+    if_exists: Optional[Literal["fail", "replace", "append"]] = "fail",
+    index: bool = True,
+    index_label: Optional[IndexLabel] = None,
+    table_type: Literal["", "temp", "temporary", "transient"] = "",
+) -> None:
+    _snowpark_pandas_obj_check(obj)
+    return obj.to_snowflake(name, if_exists, index, index_label, table_type)
+
+
+@_register_pd_accessor("to_snowflake", backend="Ray")
```
sfc-gh-joshi: If all these methods use the same body, should we just call the decorator manually to avoid repeating the function definition?

```python
def to_snowflake(
    ...
) -> None:
    ...

_register_pd_accessor("to_snowflake", backend="Pandas")(to_snowflake)
_register_pd_accessor("to_snowflake", backend="Snowflake")(to_snowflake)
_register_pd_accessor("to_snowflake", backend="Ray")(to_snowflake)
```

tests/integ/modin/frame/test_to_snowflake.py
```diff
@@ -3,29 +3,117 @@
...
+
+
+@pytest.fixture
+def valid_unquoted_identifier_table_name(session):
+    name = _generate_lower_case_table_name_unquoted() + "$n"
+    yield name
+    Utils.drop_table(session, name)
+
+
+@pytest.fixture(params=("pandas", "snowflake"), autouse=True)
```
sfc-gh-joshi: This doubles the number of tests run in the file, correct? Do we know how much runtime this adds?

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
```diff
@@ -37,12 +53,123 @@
...
+    validate_column_labels_for_to_snowflake(
+        index_column_labels=index_column_labels,
+        data_column_labels=list(pandas_frame.columns),
+    )
+    if index:
+        # write_pandas() will always drop the index, so we move the index into
+        # the data columns with reset_index().
+        pandas_frame = pandas_frame.reset_index(drop=False, names=index_column_labels)
+
+    unsupported_types = list(
```
sfc-gh-jkew, line 141: Should checks like these occur earlier in the function? I'm surprised at the complexity, and I'm wondering why we do not see these with the regular to_snowflake.

sfc-gh-joshi (commented):
Very cool! That's an impressive perf graph.

Left a few comments.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
sfc-gh-jkew: It's a bit confusing to me to have to_snowflake called from within to_snowflake without it being recursive. Is there a better way to differentiate between this higher-level wrapper function and the underlying implementation?

CHANGELOG.md
sfc-gh-mvashishtha: > Should mention the PandasToSnowflakeParquetThresholdBytes variable.

Done.

> Also, should we make it a session parameter rather than a modin flag? This would only be encountered in Snowpark pandas.

I was wondering about that, since I've seen we've used both options. I decided to stay consistent with `SnowflakePandasTransferThreshold`.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
sfc-gh-mvashishtha: Done.

sfc-gh-mvashishtha: I agree that this looks confusing , but the extension system is supposed to work this way... we could try to use the snowflake query compiler's `to_snowflake()` implementation, but this extension for the pandas backend should not assume that the snowflake query compiler has a `to_snowflake()` method.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
sfc-gh-mvashishtha: > Should checks like these occur earlier in the function?

We have to check `index` to determine whether we will lose the index types, so doing the check here means we don't need to add an extra step that's gated on `index`.

> I'm surprised at the complexity, and I'm wondering why we do not see these with the regular to_snowflake.

We do these checks for to_snowflake and to_snowpark, but via `SnowflakeQueryCompiler._warn_lost_snowpark_pandas_type`. I've now refactored the code so that all 3 checks share some of their code.

tests/integ/modin/frame/test_to_snowflake.py
sfc-gh-mvashishtha: In fact this PR also adds `parquet_threshold`, which further doubles the number of test cases. Only 3 of the 4 new parameter combinations are meaningful, so I've now changed the parameter set to just those.

On my laptop, with 12 physical cores, I get 53.56 seconds to run this test suite on main (+ a bug fix so the tests don't fail). This PR now takes the time up to 180 seconds, so it adds about 2 minutes to the total test time.  The CI runners seem to have 64 cores, although we are using `-n 36` right now. to_snowflake() is slow and the parquet approach is especially slow on this small test data, but I want to test all the cases in this suite for correctness.

I'll try using more cores in CI separately.

src/snowflake/snowpark/modin/plugin/extensions/pd_extensions.py
sfc-gh-mvashishtha: Great idea! It works. Done.

src/snowflake/snowpark/modin/plugin/extensions/pd_extensions.py
sfc-gh-mvashishtha: I may have resolved this issue in https://github.com/snowflakedb/snowpark-python/pull/3820/files#r2392311567, but let me know if not.

src/snowflake/snowpark/modin/plugin/extensions/dataframe_extensions.py
sfc-gh-mvashishtha: I prefer to use different names, but I thought there was an expectation once in Modin's dispatching code that all extension instances would use the same function name, but that would be a bug, so I've gone ahead and used separate names now.

tests/integ/modin/frame/test_to_snowflake.py
```diff
@@ -94,38 +186,44 @@ def test_to_snowflake_if_exists(session, test_table_name):
...
     # Verify existing table is replaced with new data
+    if_exists = "replace"
     df = pd.DataFrame({"a": [1, 2, 3], "c": [4, 5, 6]})
-    with SqlCounter(query_count=2):
-        df.to_snowflake(test_table_name, if_exists="replace", index=False)
-        verify_columns(test_table_name, ["a", "c"])
-        verify_num_rows(session, test_table_name, 3)
+    with to_snowflake_counter(df=df, if_exists=if_exists):
+        df.to_snowflake(test_table_name, if_exists=if_exists, index=False)
+    verify_columns(test_table_name, ["a", "c"])
```
sfc-gh-joshi, line 211: These `verify_` functions may cause a query, right? Do we need to make sure they're covered in a SQL counter?

sfc-gh-mvashishtha: The test suite doesn't require that we cover them with a SQL counter, and I don't think the tests should care about their query counts.

sfc-gh-jkew (approved):

PR #3821, opened 2025-09-30
[1.40.0 RELEASE ONLY]: Disable hybrid execution by default

sfc-gh-joshi:
Delaying the release of hybrid execution again so we can be certain of performance improvements first.


sfc-gh-joshi:
After further discussion, this was NOT cherry-picked into 1.40.0.

PR #3824, opened 2025-09-30
SNOW-2380741: Use all logical cores in snowpark tests.

sfc-gh-mvashishtha:
Formerly, we had limited the test parallelism to not overwhelm our Snowflake deployments, but now that we are using multi-cluster warehouses, we should be able to use all 64 cores.

This change speeds up Modin CI jobs according to a sample size of 1. Other long-running CI jobs seem to take about as long before and after. [before](https://github.com/snowflakedb/snowpark-python/actions/runs/18150423863) vs [after](https://github.com/snowflakedb/snowpark-python/actions/runs/18150450772?pr=3824):

<table>
<tr>
 <td> test
 <td> before (minutes)
 <td> after (minutes)
<tr>
 <td> modin-macos-latest-3.9-aws
 <td> 30
 <td> 28
<tr>
 <td>  modin-ubuntu-latest-64-cores-3.10-azure
 <td> 42
 <td> 31
<tr>
 <td>  modin-ubuntu-latest-64-cores-3.12-aws
 <td> 48
 <td> 30
<tr>
 <td>  modin-windows-latest-64-cores-3.11-gcp
 <td> 71
 <td> 42
</table>


sfc-gh-joshi (approved):

sfc-gh-joshi:
Is it possible to use `pytest -n auto` instead in case we change CI runners in the future?

sfc-gh-mvashishtha:
> Is it possible to use pytest -n auto instead in case we change CI runners in the future?

Done.

sfc-gh-mvashishtha:
I'm marking this PR as a draft because:

- it may cause OOMs for AST tests
- I think we can go even faster with `-n 128` since modin tests are network-bound.

sfc-gh-mvashishtha:
I found that using 128 cores seemed to push the GitHub Actions runners beyond their capacity so that they failed with exit code 143: https://github.com/snowflakedb/snowpark-python/pull/3824

I used the previous level of parallelism for AST tests, which seem to use too many resources when we use `-n logical`.

sfc-gh-aling (approved):

PR #3830, opened 2025-10-01
SNOW-2381033: Deflake hybrid tests by using separate tables.

sfc-gh-mvashishtha:


sfc-gh-joshi (approved):
I do remember the initialization fixture taking a while, and I see you made the fixture generate fewer rows. As long as it's still above whatever switching thresholds we're testing it should be OK.

The "one table per test approach" is fine, but I think we ideally would find a way to initialize the table once during setup instead of re-creating it every time the fixture is initialized (maybe a single CTAS if not exists?), since none of the tests mutate it. I thought that's how module-scoped fixtures work, but I guess that's not the case.

sfc-gh-mvashishtha:
@sfc-gh-joshi never mind; the problem with the pytest-module-level fixture is that we could get the race condition again if different test files use the fixture. the `session` fixture is at the pytest module level, so I prefer the function-scoped version. It's not too slow, for now, either.

tests/integ/modin/hybrid/conftest.py
```diff
@@ -28,25 +28,35 @@ def enable_autoswitch():
...
+
+
+@pytest.fixture(scope="module")
+def revenue_transactions(module_scoped_test_table_name):
     session = pd.session
     session.sql(
-        """
-    CREATE OR REPLACE TABLE revenue_transactions (
+        f"""
+    CREATE OR REPLACE TEMP TABLE {module_scoped_test_table_name} (
```
sfc-gh-mvashishtha, line 44: TEMP is not necessary now that we're using a different table name in each pytest module, but that's ok.

PR #3849, opened 2025-10-03
SNOW-2325411 Enable autoswitching for unsupported arguments

sfc-gh-jenzhang:

   

   Fixes [SNOW-2359398](https://snowflakecomputing.atlassian.net/browse/SNOW-2359398), [SNOW-2359399](https://snowflakecomputing.atlassian.net/browse/SNOW-2359399), [SNOW-2359400](https://snowflakecomputing.atlassian.net/browse/SNOW-2359400), [SNOW-2359402](https://snowflakecomputing.atlassian.net/browse/SNOW-2359402), [SNOW-2359405](https://snowflakecomputing.atlassian.net/browse/SNOW-2359405)

   - [] I am adding a new telemetry message

4. 
   This PR adds backend implementation logic for enabling autoswitching on unsupported arguments in Snowflake. Please see [project design doc](https://docs.google.com/document/d/1dna1xOg7wuMLLTOtEqY4AZSemfmC9vBp4RF8l6bKSAA/edit?tab=t.0#heading=h.rlnpynspkowp) for design considerations. This PR enables autoswitching for:

    - cumsum when axis = 1
    - cummin when axis = 1
    - cummax when axis = 1
    - skew when axis = 1 or numeric_only = False
    - get_dummies when dummy_na = True or drop_first = True or dtype is not None

[Current Progress](https://docs.google.com/document/d/1dna1xOg7wuMLLTOtEqY4AZSemfmC9vBp4RF8l6bKSAA/edit?tab=t.0#heading=h.8g78ug38cpy8).

[SNOW-2359398]: https://snowflakecomputing.atlassian.net/browse/SNOW-2359398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-2359399]: https://snowflakecomputing.atlassian.net/browse/SNOW-2359399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-2359400]: https://snowflakecomputing.atlassian.net/browse/SNOW-2359400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-2359402]: https://snowflakecomputing.atlassian.net/browse/SNOW-2359402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-2359405]: https://snowflakecomputing.atlassian.net/browse/SNOW-2359405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -1060,7 +1060,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         cls,
         api_cls_name: Optional[str],
         operation: str,
-        args: Optional[MappingProxyType[Any, Any]],
+        args: MappingProxyType[Any, Any],
     ) -> bool:
         """
         Check if method call contains unsupported args that can only be run on the native pandas backend.
@@ -1077,8 +1077,6 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         Returns:
             True if unsupported args are detected and an auto-switch should occur
         """
-        if args is None:
-            return False
 
         method_key = MethodKey(api_cls_name, operation)
         if method_key not in HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS:
@@ -1160,7 +1158,7 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         other_qc: BaseQueryCompiler,
         api_cls_name: Optional[str],
         operation: str,
-        arguments: Optional[MappingProxyType[str, Any]] = None,
+        arguments: MappingProxyType[str, Any],
     ) -> int:
         """
         Return the coercion costs from other_qc to this qc type.


CHANGELOG.md
```diff
@@ -140,6 +140,11 @@
 - When hybrid execution is enabled, `pd.merge`, `pd.concat`, `DataFrame.merge`, and `DataFrame.join` may now move arguments to backends other than those among the function arguments.
 - Improved performance of `DataFrame.to_snowflake` and `pd.to_snowflake(dataframe)` for large data by uploading data via a parquet file. You can control the dataset size at which Snowpark pandas switches to parquet with the variable `modin.config.PandasToSnowflakeParquetThresholdBytes`.
 - Improved performance of `Series.to_snowflake` and `pd.to_snowflake(series)` for large data by uploading data via a parquet file. You can control the dataset size at which Snowpark pandas switches to parquet with the variable `modin.config.PandasToSnowflakeParquetThresholdBytes`.
+- Enhanced autoswitching functionality from Snowflake to native Pandas for methods with unsupported argument combinations:
+  - `get_dummies()` with `dummy_na=True`, `drop_first=True`, or custom `dtype` parameters
+  - `cumsum()`, `cummin()`, `cummax()` with `axis=1` (column-wise operations)
+  - `skew()` with `axis=1` or `numeric_only=False` parameters
+  - `round()` with `decimals` parameter as a Series
```
sfc-gh-nkrishna, line 186: AFAIK the parameters that are unsupported are already marked in the documentation, so we can just list the method names in the CHANGELOG (e.g. `get_dummies(), cumsum()`) without the parameters themselves

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -501,9 +503,114 @@
...
+    """
+    Decorator for SnowflakeQueryCompiler methods with args-based auto-switching.
+
+    Registers pre-op switching and raises a NotImplementedError if any unsupported-parameter predicate evaluates True.
+
+    Args:
+        api_cls_name: Frontend class name (e.g., "BasePandasDataset", "Series", "DataFrame", "None").
+        method_name: Method name to register.
+        unsupported_args: UnsupportedArgsRule for args-based auto-switching.
+                            If None, method is treated as completely unimplemented.
```
```suggestion
                            If None, method is treated as completely unimplemented.
```

fix indentation

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -501,9 +503,114 @@
...
+
+def register_query_compiler_method_not_implemented(
+    api_cls_name: Optional[str],
+    method_name: str,
+    unsupported_args: Optional["UnsupportedArgsRule"] = None,
+) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
+    """
+    Decorator for SnowflakeQueryCompiler methods with args-based auto-switching.
+
+    Registers pre-op switching and raises a NotImplementedError if any unsupported-parameter predicate evaluates True.
```
```suggestion
    Registers pre-op switching for the specified API-layer method, replacing the decorated query compiler method with a version that raises a NotImplementedError if any unsupported parameter predicate evaluates True.
```

It would also be helpful to add a bit in this docstring explaining the design reasoning for why the annotation is on the query compiler instead of on frontend (DataFrame/Series) methods.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -501,9 +503,114 @@
...
+
+    # register the method in the hybrid switch for unsupported args
+    if unsupported_args is None:
+        HYBRID_SWITCH_FOR_UNIMPLEMENTED_METHODS.add(reg_key)
+    else:
+        HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS[reg_key] = unsupported_args
+
+    from modin.core.storage_formats.pandas.query_compiler_caster import (
+        register_function_for_pre_op_switch,
+    )
```
sfc-gh-joshi: move import to top level of file

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -501,9 +503,114 @@
...
+        @functools.wraps(query_compiler_method)
+        def wrapper(self: "SnowflakeQueryCompiler", *args: Any, **kwargs: Any) -> Any:
+            bound_arguments = inspect.signature(query_compiler_method).bind(
+                self, *args, **kwargs
+            )
+            bound_arguments.apply_defaults()
+            # Extract parameters excluding 'self'
+            arguments = MappingProxyType(
+                {k: v for k, v in bound_arguments.arguments.items() if k != "self"}
+            )
```
sfc-gh-joshi, line 666:
```suggestion
            # Extract parameters excluding 'self'
            arguments = bound_arguments.arguments
            del arguments["self"]
```
nit: a little easier to read (assuming `bound_arguments` isn't needed anywhere later)

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
 
+    @classmethod
+    def _has_unsupported_args(
+        cls,
+        api_cls_name: Optional[str],
+        operation: Optional[str],
+        args: Optional[MappingProxyType[Any, Any]],
+    ) -> bool:
+        """
+        Check if method call contains unsupported args that require pandas backend.
```
```suggestion
        Check if method call contains unsupported args that can only be run on the native pandas backend.
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+
+        This method evaluates registered UnsupportedArgsRule conditions to determine
+        if the provided arguments are not supported by Snowpark pandas and should
+        trigger an auto-switch to pandas backend.
```
```suggestion
        trigger an auto-switch to the native pandas backend.
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+
+        Args:
+            api_cls_name: Class name (DataFrame, Series, BasePandasDataset, None for top-level functions)
+            operation: Method name
+            args: Method arguments including args
```
```suggestion
            args: Method arguments, including *args and **kwargs
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+
+        Returns:
+            True if unsupported args detected and auto-switch should occur
```
```suggestion
            True if unsupported args are detected and an auto-switch should occur.
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+            return False
+
+        if args is None:
+            return False
+
+        rule = HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS.get(
+            MethodKey(api_cls_name, operation)
+        )
+        if rule is None:
+            return False
```
```suggestion
        if not operation or args is None or rule not in HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS:
            return False

        rule = HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS[
            MethodKey(api_cls_name, operation)
        ]
```

nit

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+            try:
+                # Validate condition format
+                if not isinstance(condition, tuple) or len(condition) != 2:
+                    logging.warning(
+                        f"Skipping malformed condition for {api_cls_name}.{operation}: "
+                        f"expected tuple of length 2, got {type(condition).__name__} "
+                        f"of length {len(condition) if hasattr(condition, '__len__') else 'unknown'}. "
+                        f"Condition: {condition}"
+                    )
+                    continue
```
sfc-gh-joshi: IMO it would be better to 
1) fail loudly with an exception if this condition occurs
2) do the check after all decorators have been run, rather than at runtime during method invocation, OR use a helper method to insert entries into the rules dict that raises an exception for any malformed entries when they're added

The rule dictionary is initialized after all the query compiler methods are declared and decorators are run, and never mutated again. As such, we can afford to do this validation either once after the decorators are all run, or while we're in the process of populating the dict in the first place. Raising a warning rather than failing loudly is harder to detect and triage.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+            return False
+
+        rule = HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS.get(
+            MethodKey(api_cls_name, operation)
+        )
+        if rule is None:
+            return False
+
+        # Check custom conditions
+        for condition in rule.unsupported_conditions:
```
sfc-gh-joshi: Consider turning this loop into a method on the `UnsupportedArgsRule` class, like `def validate(args: Optional[MappingProxyType[Any, Any]]) -> bool`. IMO it would also be easier to read if it were structured like `return any(lambda condition: ...)` instead of an explicit loop, but that's personal preference (see also my comment below about moving out validation code).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+            return None
+
+        rule = HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS.get(
+            MethodKey(api_cls_name, operation)
+        )
+        if rule is None:
+            return None
+
+        # Check custom conditions and return the specific reason
+        for condition in rule.unsupported_conditions:
```
sfc-gh-joshi: As above, this loop should probably be a method on the rule class instead.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+                logging.warning(
+                    f"Exception occurred while evaluating unsupported args condition for "
+                    f"{api_cls_name}.{operation}: {e}. Condition: {condition}. Continuing to next condition."
+                )
+                continue
+
+        return False
+
+    @classmethod
+    def _get_unsupported_args_reason(
```
sfc-gh-joshi: Does this need to be a query compiler method, or can it just be a standalone function/method on the rule class?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+                        and args.get(arg_name) == unsupported_value
+                    ):
+                        # Auto-generate error message for simple value checks
+                        return f"{arg_name}={repr(unsupported_value)} is not supported"
+            except Exception as e:
+                logging.warning(
+                    f"Exception occurred while evaluating unsupported args reason for "
+                    f"{api_cls_name}.{operation}: {e}. Condition: {condition}. Continuing to next condition."
+                )
+                continue
```
sfc-gh-joshi: does this `continue` do anything?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -916,11 +1158,28 @@ def stay_cost(
         operation: str,
         arguments: MappingProxyType[str, Any],
     ) -> Optional[int]:
+
```
sfc-gh-joshi, line 1130:
```suggestion
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -916,11 +1158,28 @@ def stay_cost(
...
         if (
             self._is_in_memory_init(api_cls_name, operation, arguments)
-            or (api_cls_name, operation) in HYBRID_SWITCH_FOR_UNIMPLEMENTED_METHODS
+            or MethodKey(api_cls_name, operation)
```
sfc-gh-joshi: nit: initialize this `MethodKey` into a local variable since it's reused later

sfc-gh-joshi (changes requested):
reviewed query compiler changes, will review tests later

CHANGELOG.md
sfc-gh-joshi: I believe we usually call out parameter-specific behavior explicitly in the changelog like this.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,311 @@ def test_set_index():
     # SNOW-2333472: Switching backends then setting the index should be valid.
     s.index = ["a"]
     assert_snowpark_pandas_equal_to_pandas(s, native_pd.Series([0], index=["a"]))
+
+
+def _test_stay_cost(data_obj, api_cls_name, method_name, args, expected_cost):
```
sfc-gh-mvashishtha:
```suggestion
def _test_stay_cost(*, data_obj, api_cls_name, method_name, args, expected_cost):
```

There are a lot of parameters here, so I think it's clearer to make them all keyword arguments.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,311 @@ def test_set_index():
...
+    stay_cost = data_obj._query_compiler.stay_cost(
+        api_cls_name, method_name, MappingProxyType(args)
+    )
+    assert stay_cost == expected_cost
+
+
+def _test_move_to_me_cost(pandas_qc, api_cls_name, method_name, args, expected_cost):
```
sfc-gh-mvashishtha:
```suggestion
def _test_move_to_me_cost(*, pandas_qc, api_cls_name, method_name, args, expected_cost):
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,311 @@ def test_set_index():
...
+    move_to_me_cost = SnowflakeQueryCompiler.move_to_me_cost(
+        pandas_qc, api_cls_name, method_name, MappingProxyType(args)
+    )
+    assert move_to_me_cost == expected_cost
+
+
+def _test_expected_backend(
+    data_obj, method_name, args, expected_backend, is_top_level=False
```
sfc-gh-mvashishtha:
```suggestion
    *, data_obj, method_name, args, expected_backend, is_top_level=False
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,311 @@ def test_set_index():
...
+
+    _test_expected_backend(df, "round", args, "Pandas", False)
+    _test_expected_backend(series, "round", args, "Pandas", False)
+
+
+@sql_count_checker(query_count=0)
+def test_error_handling_when_auto_switch_disabled():
+    # Test that unsupported args raise NotImplementedError when auto-switch is disabled.
+    with config_context(AutoSwitchBackend=False):
+        for case in UNSUPPORTED_CASES:
```
sfc-gh-mvashishtha: We should parametrize the test instead of looping over cases within the test body. The main benefit of doing that is that someone who's running all the tests in the file will see all the parameters that fail instead of just seeing that the first parameter in the loop failed.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,311 @@ def test_set_index():
...
+        is_supported=False,
+        test_series=False,
+    ),
+]
+
+
+@pytest.mark.parametrize(
+    "case", SUPPORTED_CASES + UNSUPPORTED_CASES, ids=lambda c: f"{c.method}({c.args})"
+)
+def test_auto_switch(case: AutoSwitchCase):
```
sfc-gh-mvashishtha: Generally, we should [keep logic to a minimum](https://dev.to/ethancarlsson/dont-write-conditional-logic-in-tests-5aa) in our tests. There are a lot of conditionals in this test. Could you try to simplify this test, even if that means that you have to write out more parameter combinations and/or more test cases? You might consider having separate test functions for dataframes and series.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,311 @@ def test_set_index():
...
+
+            if hasattr(actual, "get_backend"):
+                assert_snowpark_pandas_equal_to_pandas(actual, expected)
+            else:
+                assert actual == expected
+
+
+@sql_count_checker(query_count=2)
+def test_round_with_series_decimals():
+    # Test round with Series decimals (unsupported case). round requires a different implementation so we make a separate test.
```
sfc-gh-mvashishtha: Different from what? Which implementation?

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,311 @@ def test_set_index():
...
+    round_series_decimals = pd.Series([1, 2], index=["A", "B"])
+    test_data = {"A": [1.234, 2.567], "B": [3.891, 4.123]}
+
+    df = pd.DataFrame(test_data).move_to("Snowflake")
```
sfc-gh-mvashishtha: let's split the dataframe and series into separate test cases

sfc-gh-mvashishtha (commented):
I just reviewed tests/integ/modin/hybrid/test_switch_operations.py

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -873,6 +980,141 @@ def _are_dtypes_compatible_with_snowflake(cls, compiler: BaseQueryCompiler) -> b
...
+                )
+                continue
+
+        return False
+
+    @classmethod
+    def _get_unsupported_args_reason(
+        cls,
+        api_cls_name: Optional[str],
+        operation: Optional[str],
```
sfc-gh-mvashishtha: Is there a need to make the interface accept `None` operation? If not I would just say it's a `str`. I would probably not even bother with validating that it's a `str`.

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+                ("dummy_na", True),
+                ("drop_first", True),
+                (
+                    lambda args: args.get("dtype") is not None,
+                    "get_dummies with non-default dtype parameter is not supported yet in Snowpark pandas.",
+                ),
+            ]
+        )
+
+    def test_basic_functionality(self):
```
sfc-gh-mvashishtha: Can we reduce the number of assertions in this test case, either by parametrizing or using separate test functions? Generally we would like each test case to test a single behavior.

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+            "DataFrame", "method", None
+        )
+
+        # No rule registered
+        args = MappingProxyType({"param": "value"})
+        assert not SnowflakeQueryCompiler._has_unsupported_args(
+            "DataFrame", "unknown_method", args
+        )
+
+    def test_simple_conditions(self):
```
sfc-gh-mvashishtha: ditto about reducing the number of assertions per test case.

sfc-gh-mvashishtha (commented):
through line 145 of tests/unit/modin/test_unsupported_args_auto_switch.py

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,311 @@ def test_set_index():
...
+    api_cls_name: Optional[str]
+    method: str
+    args: Dict[str, Any]
+    test_data: Dict[str, Any]
+    expected_query_count: int
+    is_supported: bool
+    skip_backend_check: bool = (
+        False  # skip backend behavior checks due to post-operation switching
+    )
+    test_series: bool = True  # some methods/args are not supported on Series
```
sfc-gh-joshi: Is this meant to determine whether a `BasePandasDataset` test is run on `Series` objects as well? The naming is a bit confusing here, similar to Mahesh's point about keeping logic in tests to a minimum. Separating DF/Series tests would help with that.

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+    UnsupportedArgsRule,
+    MethodKey,
+    HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS,
+    HYBRID_SWITCH_FOR_UNIMPLEMENTED_METHODS,
+    register_query_compiler_method_not_implemented,
+)
+
+
+class TestHasUnsupportedArgs:
+    def setup_method(self):
```
sfc-gh-joshi: You should restore `HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS` at the end of each test in case other tests in the system depend on their values getting initialized properly.

We also usually use pytest fixtures instead of `setup_method`/`teardown_method`. That way you can also share common setup across classes, like

```python
@pytest.fixture(scope="function", autouse=True)
def reset_unsupported_args_dict():
    old_args = dict(HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS)
    HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS.clear()
    HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS.update({
        # whatever initialization code
    })
    yield
    HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS.clear()
    HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS.update(old_args)
```

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+    def test_decorator_functionality(self):
+        rule = UnsupportedArgsRule(unsupported_conditions=[("bad_param", True)])
+
+        @register_query_compiler_method_not_implemented(
+            "DataFrame", "test_method", rule
+        )
+        def test_method(self, param1=None, bad_param=False):
+            return "original_result"
+
+        mock_qc = MagicMock()
```
sfc-gh-joshi: We usually avoid mocking objects in tests because a lot of invariants can end up getting broken by accident. Can you find an actual QC method that you've applied the decorator to, and call it with the appropriate parameters here? If we end up implementing those parameters in the future we can change the test later.

If this makes the test redundant then I'd say it's fine to remove.

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+        )
+        def test_method(self, bad_param=False):
+            return "original_result"
+
+        mock_qc = MagicMock()
+        test_method(mock_qc, bad_param=True)
+
+        mock_error.assert_called_once_with(
+            "test_method", "specific reason for bad_param"
+        )
```
sfc-gh-joshi: As above: prefer to not mock QCs, instead call an actual method with the relevant behavior and wrap in `pytest.raises`. In this case, `ErrorMessage.not_implemented_with_reason` is also an implementation detail, and what we want to test is that an error was raised, not that this specific internal helper method was called.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: make sense, i made it to be a static method on the rule class

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: sure, i changed this loop to be a rule instance method

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: removed!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: sure

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+
+        args = MappingProxyType({"valid_param": True})
+
+        with patch("logging.warning") as mock_warning:
+            result = SnowflakeQueryCompiler._has_unsupported_args(
+                "DataFrame", "test_method", args
+            )
+
+            # Should return True because the valid condition matches
```
sfc-gh-mvashishtha: I prefer that we raise the original exception in this case. Generally we shouldn't catch exceptions unless there's a particular reason to do so.

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+                "DataFrame", None, MappingProxyType({})
+            )
+            is None
+        )
+        assert (
+            SnowflakeQueryCompiler._get_unsupported_args_reason(
+                "DataFrame", "", MappingProxyType({})
+            )
+            is None
+        )
```
sfc-gh-mvashishtha: ditto about checking `None` / empty methods-- this should probably raise an exception. I wouldn't check these cases in this test function.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -501,9 +506,229 @@
...
+                raise ValueError(
+                    f"Invalid condition at index {i}: first element must be callable or string, "
+                    f"got {type(condition[0]).__name__}. Condition: {condition}"
+                )
+
+    @staticmethod
+    def get_unsupported_args_reason(
+        api_cls_name: Optional[str],
+        operation: str,
+        args: Optional[MappingProxyType[Any, Any]],
```
sfc-gh-mvashishtha: As with `operation`, I don't think there's a reason to allow `args` to be `None`.

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+
+        args = MappingProxyType({"param1": "bad"})
+        reason = SnowflakeQueryCompiler._get_unsupported_args_reason(
+            "DataFrame", "test_method", args
+        )
+        assert reason == "first reason"
+
+    def test_exception_handling(self):
+        def failing_condition(args):
+            raise ValueError("Test exception")
```
sfc-gh-mvashishtha: ditto from the other test file; we should not catch exceptions from rules.

tests/unit/modin/test_unsupported_args_auto_switch.py
```diff
@@ -0,0 +1,602 @@
...
+            return "original_result"
+
+        mock_qc = MagicMock()
+        test_method(mock_qc, bad_param=True)
+
+        mock_error.assert_called_once_with(
+            "test_method", "specific reason for bad_param"
+        )
+
+    def test_fallback_error_case(self):
```
sfc-gh-mvashishtha: is there a real case like this that doesn't require mocks?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: yup i fixed the indentation

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: yup, I added more explanation on why doing it in the qc

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: moved :)

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: that makes sense. I've changed the signature and made them all kwargs

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: right... I was thinking of defining this list once and reusing these test cases in other test functions. But I guess it is clearer to define separated parameterization for each test case. I've added the changes to parametrize the tests

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: Sure! The reason why i put them in one test case is they shared similar logic (checking costs, backend, etc.), but I guess it's more readable with separate test functions. I've made the change :)

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: This case is special because round checks to see if decimals is a pd.Series (which we need a session to create), pytest initializes the parameters at module load time and a session is not yet available at that time.

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: sure! I've split the test cases to handle both scenarios

tests/unit/modin/test_unsupported_args_auto_switch.py
sfc-gh-jenzhang: Sure, I've changed the original implementation to raise an exception when dealing with this case.

tests/unit/modin/test_unsupported_args_auto_switch.py
sfc-gh-jenzhang: The fallback was introduced if the rule is not valid. But since we are now validating the rules in __post_init__,the fallback is not needed anymore. It will raise an error immediately if the unsupportedRules are malformed

tests/unit/modin/test_unsupported_args_auto_switch.py
sfc-gh-jenzhang: removed!

tests/unit/modin/test_unsupported_args_auto_switch.py
sfc-gh-jenzhang: Sure, I removed the mocks and instead moved the reason checking in the integration tests.

tests/unit/modin/test_unsupported_args_auto_switch.py
sfc-gh-jenzhang: Had a discussion with @sfc-gh-joshi offline, and I removed most of the unit tests since they are already covered by the integration tests and some unit tests are not applicable anymore.

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: Sure! I separated out the DF and series case

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: Do we still want to make it MappingProxyType?

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: added!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: true, I changed it to str

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: Sure.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -501,9 +506,175 @@
...
+        for i, condition in enumerate(self.unsupported_conditions):
+            if not isinstance(condition, tuple) or len(condition) != 2:
+                raise ValueError(
+                    f"Invalid condition at index {i}: expected tuple of length 2, "
+                    f"got {type(condition).__name__} of length "
+                    f"{len(condition) if hasattr(condition, '__len__') else 'unknown'}. "
+                    f"Condition: {condition}"
+                )
+
+            if not (callable(condition[0]) or isinstance(condition[0], str)):
```
sfc-gh-joshi, line 544: Can you also validate that `condition[1]` is a str for the callable case here?

sfc-gh-jenzhang: sure, I added the check

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: I just want to make sure I’m following — were you suggesting something like this? I could be mistaken, but this feels a bit awkward to me — do you think this version’s better than what we had before?

```python
arguments = bound_arguments.arguments
del arguments["self"]
arguments = MappingProxyType(arguments)
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -501,9 +506,182 @@
...
+
+
+# Named tuple for method registry keys.
+class MethodKey(NamedTuple):
+    api_cls_name: Optional[str]
+    method_name: str
+
+
+# Rule defining which args should trigger auto-switching.
+@dataclass
```
sfc-gh-mvashishtha:
```suggestion
@dataclass(frozen=True)
```

Does this work?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -501,9 +506,182 @@
...
+            if callable(condition[0]):
+                # tuple[Callable, str]: (condition_function, reason)
+                condition_func, reason = condition
+                if condition_func(args):
+                    return reason
+            else:
+                # tuple[str, Any]: (argument_name, unsupported_value)
+                arg_name, unsupported_value = condition
+                if args.get(arg_name) == unsupported_value:
+                    return f"{arg_name} = {repr(unsupported_value)} is not supported"
```
sfc-gh-mvashishtha:
```suggestion
                    return f"{arg_name} = {unsupported_value} is not supported"
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -955,7 +1183,12 @@ def move_to_me_cost(
         if (
             cls._is_in_memory_init(api_cls_name, operation, arguments)
             or not cls._are_dtypes_compatible_with_snowflake(other_qc)
-            or (api_cls_name, operation) in HYBRID_SWITCH_FOR_UNIMPLEMENTED_METHODS
+            or (
+                operation
+                and MethodKey(api_cls_name, operation)
```
sfc-gh-mvashishtha:
```suggestion
                MethodKey(api_cls_name, operation)
```

I don't think `operation` can be `None`. Could you fix it in the docstring and remove this condition?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -955,7 +1183,12 @@ def move_to_me_cost(
...
+                in HYBRID_SWITCH_FOR_UNIMPLEMENTED_METHODS
+            )
+            or cls._has_unsupported_args(api_cls_name, operation, arguments)
```
sfc-gh-mvashishtha, line 1186: could you move the comment " # in-memory intialization should not move to Snowflake" above the relevant condition instead of above all the conditions? I think that was a problem even prior to your PR, but we can fix it now since we're adding more conditions.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+
+def _test_expected_backend(
+    *, data_obj, method_name, args, expected_backend, is_top_level=False
+):
+    if is_top_level:
+        result = getattr(pd, method_name)(data_obj, **args)
+    else:
+        result = getattr(data_obj, method_name)(**args)
+
+    if hasattr(result, "get_backend"):
```
sfc-gh-mvashishtha: We should assume that the input has a backend since the function is called `_test_expected_backend`.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+
+
+@pytest.mark.parametrize(
+    "method,args,test_data,expected_query_count",
+    [
+        ("get_dummies", {}, {"A": ["x", "y", "z"], "B": [1, 2, 3]}, 1),
+    ],
+)
+def test_auto_switch_supported_top_level_functions(
+    method, args, test_data, expected_query_count
```
sfc-gh-mvashishtha: nit: can we call `args` `kwargs` instead? Same for other test functions.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        # Test cost - should be zero for supported operations
+        _test_stay_cost(
+            data_obj=df,
+            api_cls_name=None,
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_ZERO,
+        )
+
+        # Test expected backend
```
sfc-gh-mvashishtha:
```suggestion
```

nit: The helper function name says the same thing as the comment.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        # Test cost - should be zero for supported operations
+        _test_stay_cost(
+            data_obj=df,
+            api_cls_name="BasePandasDataset",
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_ZERO,
+        )
+
+        # Test expected backend
```
sfc-gh-mvashishtha:
```suggestion
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        )
+
+        # Test result equality
+        actual = getattr(df, method)(**args)
+        expected = getattr(native_pd.DataFrame(test_data), method)(**args)
+
+        if hasattr(actual, "get_backend"):
+            assert_snowpark_pandas_equal_to_pandas(actual, expected)
+        else:
+            assert actual == expected
```
sfc-gh-mvashishtha:
```suggestion
        eval_snowpark_pandas_result(df, native_pd.DataFrame(test_data), lambda df: getattr(df, method)(**args))
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        # Test cost - should be zero for supported operations
+        _test_stay_cost(
+            data_obj=series,
+            api_cls_name="BasePandasDataset",
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_ZERO,
+        )
+
+        # Test expected backend
```
sfc-gh-mvashishtha:
```suggestion
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        _test_expected_backend(
+            data_obj=series,
+            method_name=method,
+            args=args,
+            expected_backend="Snowflake",
+            is_top_level=False,
+        )
+
+        # Test result equality
+        actual = getattr(series, method)(**args)
```
sfc-gh-mvashishtha: We can use `eval_snowpark_pandas_result` here.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        expected = getattr(native_pd.Series(series_data), method)(**args)
+
+        if hasattr(actual, "get_backend"):
```
sfc-gh-mvashishtha: Instead of using this conditional to determine whether to check `get_backend`, we can add an extra parameter for `comparator` and pass that to `eval_snowpark_pandas_result`. The comparator parameters can be `[assert_snowpark_pandas_equal_to_pandas, assert_scalars_equal]` where `assert_scalars_equal` is a function defined in tests/integ/modin/utils.py like

```python
def assert_scalars_equal(snow, expected_pandas):
    assert snow == expected_pandas
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+            api_cls_name="BasePandasDataset",
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_ZERO,
+        )
+
+        # Test result equality - don't check backend as it switches post-operation
+        df_actual = getattr(df, method)(**args)
+        df_expected = getattr(native_pd.DataFrame(test_data), method)(**args)
+        assert_snowpark_pandas_equal_to_pandas(df_actual, df_expected)
```
sfc-gh-mvashishtha:
```suggestion
        # Test result equality - don't check backend as it switches post-operation
        eval_snowpark_pandas_result(df, native_pd.DataFrame(test_data), lambda df: getattr(df, method)(**args))
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+    return result
+
+
+@pytest.mark.parametrize(
+    "method,args,test_data,expected_query_count",
+    [
+        ("get_dummies", {}, {"A": ["x", "y", "z"], "B": [1, 2, 3]}, 1),
+    ],
+)
+def test_auto_switch_supported_top_level_functions(
```
sfc-gh-mvashishtha: query count is always 1 so you can remove the parametrization and use `sql_count_checker`. ditto for some other test cases.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        _test_stay_cost(
+            data_obj=series,
+            api_cls_name="BasePandasDataset",
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_ZERO,
+        )
+
+        # Test result equality - don't check backend as it switches post-operation
+        series_actual = getattr(series, method)(**args)
```
sfc-gh-mvashishtha: ditto; can use eval_snowpark_pandas_result

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        series_expected = getattr(native_pd.Series(series_data), method)(**args)
+        assert_snowpark_pandas_equal_to_pandas(series_actual, series_expected)
+
+
+@pytest.mark.parametrize(
```
sfc-gh-mvashishtha, line 877: can remove the parametrization on test data since it's all the same

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+            expected_cost=QCCoercionCost.COST_ZERO,
+        )
+
+        # Test result equality - don't check backend as it switches post-operation
+        df_actual = getattr(df, method)(**args)
+        df_expected = getattr(native_pd.DataFrame(test_data), method)(**args)
+        assert_snowpark_pandas_equal_to_pandas(df_actual, df_expected)
+
+
+@pytest.mark.parametrize(
```
sfc-gh-mvashishtha, line 845: can use the same data for all test cases and remove the parametrization on data

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+    "method,args,test_data,expected_query_count",
+    [
+        ("cumsum", {"axis": 0}, {"A": [1, 2, 3], "B": [4, 5, 6]}, 1),
+        ("cummin", {"axis": 0}, {"A": [1, 2, 3], "B": [4, 5, 6]}, 1),
+        ("cummax", {"axis": 0}, {"A": [1, 2, 3], "B": [4, 5, 6]}, 1),
+    ],
+)
+def test_auto_switch_supported_post_op_switch_point_dataframe(
```
sfc-gh-mvashishtha: can use the same data for all test cases and remove the parametrization on data

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        # Test cost - should be impossible for unsupported operations
+        _test_stay_cost(
+            data_obj=df,
+            api_cls_name=None,
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_IMPOSSIBLE,
+        )
+
+        # Test move_to_me cost
```
sfc-gh-mvashishtha:
```suggestion
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        pandas_df = pd.DataFrame(test_data)
+        _test_move_to_me_cost(
+            pandas_qc=pandas_df._query_compiler,
+            api_cls_name=None,
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_IMPOSSIBLE,
+        )
+
+        # Test expected backend
```
sfc-gh-mvashishtha:
```suggestion
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        result = getattr(pd, method_name)(data_obj, **args)
+    else:
+        result = getattr(data_obj, method_name)(**args)
+
+    if hasattr(result, "get_backend"):
+        assert (
+            result.get_backend() == expected_backend
+        ), f"Expected {expected_backend} backend, but got {result.get_backend()}"
+
+    return result
```
sfc-gh-mvashishtha:
```suggestion
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+    "method,args,test_data,expected_query_count",
+    [
+        ("skew", {"axis": 1}, {"A": [1, 2, 3], "B": [4, 5, 6]}, 1),
+        ("skew", {"numeric_only": False}, {"A": [1, 2, 3], "B": [4, 5, 6]}, 1),
+        ("cumsum", {"axis": 1}, {"A": [1, 2, 3], "B": [4, 5, 6]}, 1),
+        ("cummin", {"axis": 1}, {"A": [1, 2, 3], "B": [4, 5, 6]}, 1),
+        ("cummax", {"axis": 1}, {"A": [1, 2, 3], "B": [4, 5, 6]}, 1),
+    ],
+)
+def test_auto_switch_unsupported_dataframe(
```
sfc-gh-mvashishtha: ditto about not making test data a parameter.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        pandas_df = pd.DataFrame(test_data)
+        _test_move_to_me_cost(
+            pandas_qc=pandas_df._query_compiler,
+            api_cls_name="BasePandasDataset",
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_IMPOSSIBLE,
+        )
+
+        # Test expected backend
```
sfc-gh-mvashishtha:
```suggestion
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        _test_expected_backend(
+            data_obj=df,
+            method_name=method,
+            args=args,
+            expected_backend="Pandas",
+            is_top_level=False,
+        )
+
+        # Test result equality
+        actual = getattr(df, method)(**args)
```
sfc-gh-mvashishtha: ditto; can use eval_snowpark_pandas_result and parametrize the comparator.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        # Test cost - should be impossible for unsupported operations
+        _test_stay_cost(
+            data_obj=series,
+            api_cls_name="BasePandasDataset",
+            method_name=method,
+            args=args,
+            expected_cost=QCCoercionCost.COST_IMPOSSIBLE,
+        )
+
+        # Test expected backend
```
sfc-gh-mvashishtha:
```suggestion
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        _test_expected_backend(
+            data_obj=series,
+            method_name=method,
+            args=args,
+            expected_backend="Pandas",
+            is_top_level=False,
+        )
+
+        # Test result equality
```
sfc-gh-mvashishtha: assert_snowpark_pandas_equal_to_pandas

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: yes it does

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        (
+            "get_dummies",
+            {"dtype": int},
+            {"A": ["x", "y", "z"], "B": [1, 2, 3]},
+            "get_dummies with non-default dtype parameter is not supported yet in Snowpark pandas.",
+        ),
+    ],
+)
+@sql_count_checker(query_count=0)
+def test_error_handling_top_level_functions_when_auto_switch_disabled(
```
sfc-gh-mvashishtha, line 1046: remove the parametrization by data.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+    method, args, test_data, expected_reason
+):
+    # Test that unsupported top-level function args raise NotImplementedError when auto-switch is disabled.
+    with config_context(AutoSwitchBackend=False):
+        df = pd.DataFrame(test_data).move_to("Snowflake")
+
+        with pytest.raises(NotImplementedError) as exc_info:
```
sfc-gh-mvashishtha: you can use the `match` parameter, but be sure to escape the match pattern with `re.escape`.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        (
+            "cummax",
+            {"axis": 1},
+            {"A": [1, 2, 3], "B": [4, 5, 6]},
+            "axis = 1 is not supported",
+        ),
+    ],
+)
+@sql_count_checker(query_count=0)
+def test_error_handling_dataframe_when_auto_switch_disabled(
```
sfc-gh-mvashishtha, line 1093: ditto; remove parametrization by data

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+    method, args, test_data, expected_reason
+):
+    # Test that unsupported DataFrame args raise NotImplementedError when auto-switch is disabled.
+    with config_context(AutoSwitchBackend=False):
+        df = pd.DataFrame(test_data).move_to("Snowflake")
+
+        with pytest.raises(NotImplementedError) as exc_info:
```
sfc-gh-mvashishtha: ditto; use `match` parameter

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+)
+@sql_count_checker(query_count=0)
+def test_error_handling_series_when_auto_switch_disabled(
+    method, args, series_data, expected_reason
+):
+    # Test that unsupported Series args raise NotImplementedError when auto-switch is disabled.
+    with config_context(AutoSwitchBackend=False):
+        series = pd.Series(series_data).move_to("Snowflake")
+
+        with pytest.raises(NotImplementedError) as exc_info:
```
sfc-gh-mvashishtha: ditto ; use `match`

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+    [
+        (
+            "skew",
+            {"numeric_only": False},
+            [1, 2, 3, 4, 5, 6],
+            "numeric_only = False argument not supported for skew",
+        ),
+    ],
+)
+@sql_count_checker(query_count=0)
```
sfc-gh-mvashishtha, line 1119: remove parametrization by data

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        # Assert the specific expected reason is in the error message
+        error_msg = str(exc_info.value)
+        assert (
+            f"Snowpark pandas {method} does not yet support the parameter combination because {expected_reason}"
+            in error_msg
+        )
+
+
+@sql_count_checker(query_count=1)
+def test_auto_switch_unsupported_round_dataframe():
```
sfc-gh-mvashishtha: why is there a separate test function for this? why is it impossible to move to pandas but also impossible to stay on Snowflake?

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,635 @@ def test_set_index():
...
+        data_obj=df,
+        method_name="round",
+        args=args,
+        expected_backend="Pandas",
+        is_top_level=False,
+    )
+
+
+@sql_count_checker(query_count=1)
+def test_auto_switch_unsupported_round_series():
```
sfc-gh-mvashishtha: ditto. why is there a separate test case case for this?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: I think that's gonna make the type incompatible because `MethodKey` expects a `operation: str` and `move_to_me_cost` takes `operation: Optional[str]`. Do we want to change the type for operation in `move_to_me_cost`?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: sure, I've moved the comment above the first condition

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: This is to deal with the case where the result is a scalar (i.e. skew) instead of a df/series. If that's the case, then the backends shouldn't be checked. Would it be better if we create separate test functions for the scalar case?

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: It is impossible to stay on snowflake and it is also impossible to move from pandas to snowflake because it is an unsupported case. This case is special because round checks to see if decimals is a pd.Series (which we need a session to create), pytest initializes the parameters at module load time and a session is not yet available at that time, so thats why I made a separate test case for this.

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: This case is special because round checks to see if decimals is a pd.Series (which we need a session to create), pytest initializes the parameters at module load time and a session is not yet available at that time, so thats why I made a separate test case for this.

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: sure!

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: okay, changed to use `match` parameter

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: Keeping the original is fine.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-joshi: I think the type of move_to_me_cost in Snowpark pandas is wrong, and Modin declares it as a non-nullable `str` and it wasn't caught because mypy checking is disabled for the query compiler.

Modin:
https://github.com/modin-project/modin/blob/99037af07717cba5a53155595ad28202e59aba90/modin/core/storage_formats/base/query_compiler.py#L459-L465

Snowpark pandas:
https://github.com/snowflakedb/snowpark-python/blob/2f09d5ef06f3143a72fa37220a616a3b97b5ced6/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py#L930-L936

Feel free to fix it in this PR.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,631 @@ def test_set_index():
...
     assert_snowpark_pandas_equal_to_pandas(s, native_pd.Series([0], index=["a"]))
+
+
+def _test_stay_cost(*, data_obj, api_cls_name, method_name, args, expected_cost):
+    stay_cost = data_obj._query_compiler.stay_cost(
+        api_cls_name, method_name, MappingProxyType(args)
+    )
+    assert (
+        stay_cost == expected_cost
+    ), f"Expected {expected_cost} stay cost, but got {stay_cost}"
```
sfc-gh-joshi:
```suggestion
    assert stay_cost == expected_cost
```
nit: no need for the message here since pytest will show the values of `stay_cost` and `expected_cost` already, and it's clear which value is expected and which is actual by convention

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,631 @@ def test_set_index():
...
+
+
+def _test_move_to_me_cost(*, pandas_qc, api_cls_name, method_name, args, expected_cost):
+    move_to_me_cost = SnowflakeQueryCompiler.move_to_me_cost(
+        pandas_qc, api_cls_name, method_name, MappingProxyType(args)
+    )
+    assert (
+        move_to_me_cost == expected_cost
+    ), f"Expected {expected_cost} move to me cost, but got {move_to_me_cost}"
```
sfc-gh-joshi:
```suggestion
    assert move_to_me_cost == expected_cost
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,631 @@ def test_set_index():
...
+):
+    if is_top_level:
+        result = getattr(pd, method_name)(data_obj, **args)
+    else:
+        result = getattr(data_obj, method_name)(**args)
+
+    if hasattr(result, "get_backend"):
+        assert (
+            result.get_backend() == expected_backend
+        ), f"Expected {expected_backend} backend, but got {result.get_backend()}"
```
sfc-gh-joshi:
```suggestion
        assert result.get_backend() == expected_backend
```

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -661,3 +664,631 @@ def test_set_index():
...
+            api_cls_name="BasePandasDataset",
+            method_name=method,
+            args=kwargs,
+            expected_cost=QCCoercionCost.COST_ZERO,
+        )
+
+        # Test result equality - don't check backend as it switches post-operation
+        df_actual = getattr(df, method)(**kwargs)
+        df_expected = getattr(native_pd.DataFrame(test_data), method)(**kwargs)
+        assert_snowpark_pandas_equal_to_pandas(df_actual, df_expected)
```
sfc-gh-joshi: You can replace these equality checks with this helper function that does the assertion for you:

```
eval_snowpark_pandas_result(
    df,
    native_pd.DataFrame(test_data),
    lambda df: getattr(df, method)(**kwargs)
)
```

sfc-gh-jenzhang: sure!

tests/integ/modin/hybrid/test_switch_operations.py
sfc-gh-jenzhang: okok, I removed all of them

tests/integ/modin/utils.py
```diff
@@ -298,6 +298,22 @@ def assert_snowpark_pandas_equal_to_pandas(
             assert [str(dt) for dt in snow_to_native.dtypes.tolist()] == expected_dtypes
 
 
+def assert_scalars_equal(snow, expected_pandas):
```
sfc-gh-joshi: You can directly use [numpy.testing.assert_allclose](https://numpy.org/doc/2.3/reference/generated/numpy.testing.assert_allclose.html) instead of this custom function.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -1075,9 +1075,9 @@ def _has_unsupported_args(
             args: Method arguments, including *args and **kwargs
 
         Returns:
-            True if unsupported args detected and auto-switch should occur
+            True if unsupported args are detected and an auto-switch should occur
         """
-        if not operation or args is None:
+        if args is None:
```
sfc-gh-mvashishtha: I don't think `args` can be `None` either... can we fix that too? I think there is also a pre-existing bug in the `move_to_me_cost` signature.

tests/integ/modin/hybrid/test_switch_operations.py
```diff
@@ -692,538 +692,447 @@ def _test_expected_backend(
...
         )
 
-        # Test expected backend
-        _test_expected_backend(
-            data_obj=series,
-            method_name=method,
-            args=args,
-            expected_backend="Snowflake",
-            is_top_level=False,
+        # We only test expected backend for non-scalar results
```
sfc-gh-mvashishtha:
```suggestion
```

nit: The code is clear enough

sfc-gh-jenzhang: sure! I removed the comment

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: Yup, I removed the optional

PR #3859, opened 2025-10-07
SNOW-2397758: Add parameter for aliasing interval function names

sfc-gh-fhe:

   

   Fixes SNOW-NNNNNNN

Currently we alias the column name for the interval_day_time_from_parts and interval_year_month_from_parts functions. However, it runs into a bug when we try to use it in a when statement. So, I'm introducing a parameter to allow us to use / not use the alias at will.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@
     - `st_geometryfromwkt`
     - `try_to_geography`
     - `try_to_geometry`
-
+- Added a parameter to enable and disable automatic column name aliasing for `interval_day_time_from_parts` and `interval_year_month_from_parts` functions.
 
 #### Bug Fixes
 
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -11095,6 +11095,8 @@ def interval_year_month_from_parts(
 
     res = cast(interval_string, "INTERVAL YEAR TO MONTH")
     if _alias_column_name:
+        # Aliasing column names when using this in a case when will throw an error. This allows us to only alias
+        # when necessary.
 
         def get_col_name(col):
             if isinstance(col._expr1, Literal):
@@ -11248,6 +11250,8 @@ def interval_day_time_from_parts(
 
     res = cast(interval_value, "INTERVAL DAY TO SECOND")
     if _alias_column_name:
+        # Aliasing column names when using this in a case when will throw an error. This allows us to only alias
+        # when necessary.
 
         def get_col_name(col):
             if isinstance(col._expr1, Literal):


src/snowflake/snowpark/functions.py
```diff
@@ -11091,15 +11093,19 @@ def interval_year_month_from_parts(
...
+    res = cast(interval_string, "INTERVAL YEAR TO MONTH")
+    if _alias_column_name:
+
+        def get_col_name(col):
+            if isinstance(col._expr1, Literal):
+                return str(col._expr1.value)
+            else:
+                return col._expression.name
+
+        alias_name = f"interval_year_month_from_parts({get_col_name(years_col)}, {get_col_name(months_col)})"
```
sfc-gh-aling, line 11107: let's add some context in the comments that alias doesn't work in CASE WHEN

src/snowflake/snowpark/functions.py
```diff
@@ -11031,6 +11031,7 @@ def make_interval(
 def interval_year_month_from_parts(
```
sfc-gh-aling, line 11031: can you mention the change in the changelog that we no longer alias the column name

CHANGELOG.md
```diff
@@ -7,43 +7,18 @@
...
-      - `st_distance`
-      - `st_dwithin`
-      - `st_endpoint`
-      - `st_envelope`
-      - `st_geohash`
-      - `st_geomfromgeohash`
-      - `st_geompointfromgeohash`
-      - `st_hausdorffdistance`
-      - `st_makepoint`
+- Added a parameter to enable and disable automatic column name aliasing for `interval_day_time_from_parts` and `interval_year_month_from_parts` functions.
```
sfc-gh-aling, line 61: after giving a second thought, this could be a BCR (changing output column name) to users whose code depends on column name. is it okay? I feel since internal type is an experimental feature so we shall be good, but still we need to call out in the changelog.

and my second question if Q1 is okay:

if we are going to do a BCR anyway, shall we consider removing the alias completely? or do you think there is still value in providing an alias?

sfc-gh-fhe: It's not a BCR I think, it's defaulting to the original behavior. The parameter allows the other behavior

src/snowflake/snowpark/functions.py
```diff
@@ -11091,15 +11093,21 @@ def interval_year_month_from_parts(
     )
     interval_string = concat(sign_prefix, normalized_years, lit("-"), normalized_months)
 
-    def get_col_name(col):
-        if isinstance(col._expr1, Literal):
-            return str(col._expr1.value)
-        else:
-            return col._expression.name
+    res = cast(interval_string, "INTERVAL YEAR TO MONTH")
+    if _alias_column_name:
```
sfc-gh-aalam, line 11097: I think this shows the need for a server side implementation for this function.

PR #3861, opened 2025-10-07
SNOW-2398189: Suppress warnings due to enabling CTE optimization at Snowpark pandas initialization

sfc-gh-helmeleegy:

   

   Fixes SNOW-2398189

   Only enable CTE optimization if not already enabled.

### DIFF
--- a/tests/integ/modin/test_session_cte_optimization.py
+++ b/tests/integ/modin/test_session_cte_optimization.py
@@ -9,6 +9,9 @@ import threading
 from contextlib import contextmanager
 from typing import Any, Generator
 
+import modin.pandas as pd
+import snowflake.snowpark.modin.plugin  # noqa: F401
+
 from snowflake.snowpark.session import _get_active_session
 
 from tests.integ.utils.sql_counter import sql_count_checker
@@ -45,11 +48,7 @@ def test_cte_optimization_for_snowpark_pandas(db_parameters, caplog):
 
         caplog.clear()
 
-        # Import Snowpark pandas
-        import modin.pandas as pd
-        import snowflake.snowpark.modin.plugin  # noqa: F401
-
-        # Verify that cte optimization will now be enabled
+        # Verify that cte optimization will be enabled once snowpark pandas is used
         assert pd.session == session
         assert pd.session.cte_optimization_enabled is True
 


sfc-gh-joshi (approved):

sfc-gh-mvashishtha (changes requested):
This needs some changes. I will follow up with a comment.

src/snowflake/snowpark/modin/plugin/_internal/session.py
```diff
@@ -81,7 +81,8 @@ def _get_active_session(self) -> Session:
             self._warn_if_possible_when_quoted_identifiers_ignore_case_is_set(
                 self._session
             )
-            self._session.cte_optimization_enabled = True
+            if not self._session.cte_optimization_enabled:
```
sfc-gh-mvashishtha, line 118: This parameter is `False` by default, so I think most users will still see the warning. The fix does work for me on user test_snowpark_python of account SFCTEST0, but I think that we've enabled the parameter at some high level there. If I switch to my snowhouse account, for example, I still get the warning on your branch with this reproducer:

```python
import modin.pandas as pd
df = pd.DataFrame([['a', 'b'], ['c', 'd']])
```

I don't think that this check really respects the warning, since there could still be a race condition when two threads try set different values of `cte_optimization_enabled`. There is a Session._lock, but I don't think it's meant for access outside the Session class.

But I don't understand why there are multiple threads active, both in an ipython session and in pytest. `threading.enumerate()` returns a list of length 2 for me in either case, hence the warning.

sfc-gh-mvashishtha: There is already a lot of code in modin that uses global objects in a way that's not thread-safe, so it may be ok to say that Snowpark pandas does not work with multithreading and ignore the warning.

sfc-gh-helmeleegy: Good point!

When I try in ipython (using account SFCTEST0), the two threads I get are the main thread and IPythonHistorySavingThread. Both of them are non-daemon threads. When I try in Notebooks (using account SNOWHOUSE), I see seven different threads including those two. The rest seem related to the notebook execution. But only the main thread is non-daemon and the other six are daemon threads. I believe that all those threads other than the main thread won't be updating the parameter. The only possibility for having another thread updating the parameter is when the user explicitly creates another thread and updates the parameter in it. In this case, that other user thread will generate the warning itself.

For pytest, not all tests generate the warning. And so not all tests have multiple threads. Looking further, it seems that creating a snowpark dataframe creates a new thread. So tests where this happens raise the warning. There might be other cases too.

sfc-gh-helmeleegy: I made changes to filter out the warning so that it does not show up at all because of how we set the parameter at the beginning of the snowpark pandas session.

tests/integ/modin/test_session_cte_optimization.py
```diff
@@ -3,22 +3,37 @@
...
 
 @sql_count_checker(query_count=0)
-def test_cte_optimization_for_snowpark_pandas(db_parameters):
+def test_cte_optimization_for_snowpark_pandas(db_parameters, caplog):
     session = _get_active_session()
 
+    caplog.set_level(logging.WARNING)
+
+    # Creating a dataframe creates a new thread which triggers the warning
+    assert len(threading.enumerate()) == 1
```
sfc-gh-mvashishtha: This assertion is failing in CI.

tests/integ/modin/test_session_cte_optimization.py
```diff
@@ -27,6 +42,12 @@ def test_cte_optimization_for_snowpark_pandas(db_parameters):
...
+    assert (
+        "You might have more than one threads sharing the Session object trying to update"
+        not in caplog.text
+    )
+
     # Revert to original setting
-    session.cte_optimizatin_enabled = cte_optimization_original
-    assert session.cte_optimizatin_enabled == cte_optimization_original
+    session.cte_optimization_enabled = cte_optimization_original
+    assert session.cte_optimization_enabled == cte_optimization_original
```
sfc-gh-mvashishtha: can we use a fixture to save the old value and reset it? That's a little bit safer.

sfc-gh-helmeleegy: Used a context manager instead.

tests/integ/modin/test_session_cte_optimization.py
sfc-gh-helmeleegy: Updated.

tests/integ/modin/test_session_cte_optimization.py
```diff
@@ -3,30 +3,58 @@
...
-    session.cte_optimization_enabled = False
-    assert session.cte_optimization_enabled is False
+        caplog.clear()
 
-    # Import Snowpark pandas
-    import modin.pandas as pd
-    import snowflake.snowpark.modin.plugin  # noqa: F401
+        # Import Snowpark pandas
+        import modin.pandas as pd
+        import snowflake.snowpark.modin.plugin  # noqa: F401
```
sfc-gh-mvashishtha:
```suggestion
```

You don't need to do these imports here. You can move the pd import to the top of the file.

tests/integ/modin/test_session_cte_optimization.py
```diff
@@ -3,30 +3,58 @@
...
+        # Verify that cte optimization will now be enabled
+        assert pd.session == session
+        assert pd.session.cte_optimization_enabled is True
 
-    # Revert to original setting
-    session.cte_optimizatin_enabled = cte_optimization_original
-    assert session.cte_optimizatin_enabled == cte_optimization_original
+        # Verify that the warning is filtered out
+        assert (
+            "You might have more than one threads sharing the Session object trying to update"
```
sfc-gh-mvashishtha, line 57: Can we use a shared constant to warn and to test here? If the warning changes, the filter may fail while the test passes.

tests/integ/modin/test_session_cte_optimization.py
sfc-gh-helmeleegy: Those imports are what caused the parameter to flip from False to True. The purpose of the test is to verify their impact by comparing their before and after.

tests/integ/modin/test_session_cte_optimization.py
sfc-gh-helmeleegy: We can do this in a separate PR because the warning code is not ours, and it also has its own tests which test for the same string as well.

tests/integ/modin/test_session_cte_optimization.py
sfc-gh-mvashishtha: > Those imports are what caused the parameter to flip from False to True.

Are you sure? The tests passes for me if I delete these imports.

sfc-gh-helmeleegy: I see. So it flips when snowpark pandas is used. Updated test.

PR #3869, opened 2025-10-09
SNOW-2403463: Fix xml reader with undeclared namespace

sfc-gh-jdu:

   

   Fixes SNOW-2403463

### DIFF
--- a/src/snowflake/snowpark/_internal/xml_reader.py
+++ b/src/snowflake/snowpark/_internal/xml_reader.py
@@ -474,10 +474,14 @@ def process_xml_range(
                         # when ignoring namespaces, strip attribute prefixes
                         # like xyz:id -> id so records with undeclared prefixes can still parse.
                         if ignore_namespace:
-                            cleaned_record = re.sub(
-                                r"\s+(\w+):(\w+)=", r" \2=", record_str
-                            )
-                            element = ET.fromstring(cleaned_record, parser)
+                            try:
+                                cleaned_record = re.sub(
+                                    r"\s+(\w+):(\w+)=", r" \2=", record_str
+                                )
+                                element = ET.fromstring(cleaned_record, parser)
+                            except Exception as inner_ex:
+                                # avoid chained exceptions
+                                raise inner_ex from None
                         else:
                             raise
                 else:


src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -468,7 +468,18 @@ def process_xml_range(
...
+                    try:
+                        element = ET.fromstring(record_str, parser)
+                    except ET.XMLSyntaxError:
+                        # when ignoring namespaces, strip attribute prefixes
+                        # like xyz:id -> id so records with undeclared prefixes can still parse.
+                        if ignore_namespace:
+                            cleaned_record = re.sub(
+                                r"\s+(\w+):(\w+)=", r" \2=", record_str
+                            )
+                            element = ET.fromstring(cleaned_record, parser)
```
sfc-gh-jrose: If this throws a further exception I think it will display as a chained exception which may get confusing.

sfc-gh-jdu: which do you mean by "chained exception", the stack trace will only throw one exception from this line?

sfc-gh-jrose: Here's a trivial example that can create some chained errors:
```
def test(val):
    try:
        if not isinstance(val, str):
            raise RuntimeError("not str")
        print(val)
    except Exception as e:
        if isinstance(val, list):
            raise RuntimeError("list")
        elif isinstance(val, int):
            print(val)
        else:
            raise
```
If the input value is not a string it throws an exception and then tries to recover if it is an int. If the type is a list though it throws another exception during the except clause. For a non-string and non-list input you get the outer error:
```
>>> test(dict())
  File "<stdin>", line 4, in test
RuntimeError: not str
```
For the list case though you get a chained error list this:
```
>>> test([1])
  File "<stdin>", line 4, in test
RuntimeError: not str

During handling of the above exception, another exception occurred:

  File "<stdin>", line 8, in test
RuntimeError: list
```
Here the `not str` exception isn't really what's failing so it just adds confusion to the errors.

sfc-gh-jdu: good catch. It should be fixed now

PR #3879, opened 2025-10-10
SNOW-2396077: Support dtype parameter in get_dummies.

sfc-gh-mvashishtha:
The "dtype" parameter controls the values of the indicator variables, e.g. dtype=int means that we use 1 and 0 instead of True and False, respectively.

### DIFF
--- a/tests/integ/modin/frame/test_get_dummies.py
+++ b/tests/integ/modin/frame/test_get_dummies.py
@@ -331,13 +331,13 @@ class TestDtypeParameter:
 
 
 @sql_count_checker(query_count=0)
-def test_dummy_na():
+def test_dummy_na_negative():
     with pytest.raises(NotImplementedError):
         pd.get_dummies(pd.DataFrame(["a", None]), dummy_na=True)
 
 
 @sql_count_checker(query_count=0)
-def test_drop_first():
+def test_drop_first_negative():
     with pytest.raises(NotImplementedError):
         pd.get_dummies(pd.DataFrame(["a", "b"]), drop_first=True)
 


tests/integ/modin/series/test_get_dummies.py
sfc-gh-mvashishtha, line 1: I wasn't able to get GitHub to recognize that I was moving this file, even though I used `git cp` on both this file and the frame test file.

tests/integ/modin/frame/test_get_dummies.py
```diff
@@ -246,25 +250,98 @@ def test_get_dummies_pandas_after_read_snowflake(session):
...
+@sql_count_checker(query_count=0)
+def test_dummy_na():
+    with pytest.raises(NotImplementedError):
+        pd.get_dummies(pd.DataFrame(["a", None]), dummy_na=True)
+
+
+@sql_count_checker(query_count=0)
+def test_drop_first():
+    with pytest.raises(NotImplementedError):
+        pd.get_dummies(pd.DataFrame(["a", "b"]), drop_first=True)
```
sfc-gh-mvashishtha, line 342: The previous test, `test_get_dummies_pandas_negative`, was checking `dummy_na`, `drop_first`, and `dtypes` in a single test case.

docs/source/modin/supported/general_supported.rst
```diff
@@ -32,8 +32,8 @@ Data manipulations
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``from_dummies``            | N                               |                                  |                                                    |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``get_dummies``             | P                               | ``sparse`` is ignored            | ``Y`` if params ``dummy_na``, ``drop_first``       |
-|                             |                                 |                                  | and ``dtype`` are default, otherwise ``N``         |
+| ``get_dummies``             | P                               | ``sparse`` is ignored            | ``Y`` if params ``dummy_na`` and ``drop_first``    |
+|                             |                                 |                                  | and default, otherwise ``N``                       |
```
sfc-gh-helmeleegy:
```suggestion
|                             |                                 |                                  | are default, otherwise ``N``                       |
```

tests/integ/modin/frame/test_get_dummies.py
```diff
@@ -246,25 +250,98 @@ def test_get_dummies_pandas_after_read_snowflake(session):
...
+            expect_exception=True,
+            expect_exception_type=ValueError,
+            expect_exception_match=re.escape(
+                "dtype=object is not a valid dtype for get_dummies"
+            ),
         )
 
 
+@sql_count_checker(query_count=0)
+def test_dummy_na():
```
sfc-gh-helmeleegy: Can you have "negative" or something similar in the test name to indicate it's a negative case?

tests/integ/modin/frame/test_get_dummies.py
```diff
@@ -246,25 +250,98 @@ def test_get_dummies_pandas_after_read_snowflake(session):
...
+    with pytest.raises(NotImplementedError):
+        pd.get_dummies(pd.DataFrame(["a", None]), dummy_na=True)
+
+
+@sql_count_checker(query_count=0)
+def test_drop_first():
```
sfc-gh-helmeleegy: Same for renaming the negative test

tests/integ/modin/series/test_get_dummies.py
sfc-gh-helmeleegy: This is weird!

tests/integ/modin/frame/test_get_dummies.py
sfc-gh-mvashishtha: Done.

tests/integ/modin/frame/test_get_dummies.py
sfc-gh-mvashishtha: Done.

tests/integ/modin/frame/test_get_dummies.py
```diff
@@ -246,25 +250,98 @@ def test_get_dummies_pandas_after_read_snowflake(session):
...
+            "float64",
+            str,
+            "str",
+            np.str_,
+            bool,
+            "bool",
+            np.bool_,
+            "datetime64[ns]",
+            param(
+                "timedelta64[ns]",
```
sfc-gh-lmukhopadhyay, line 271: Why is timedelta64[ns] unimplemented?

tests/integ/modin/frame/test_get_dummies.py
```diff
@@ -246,25 +250,98 @@ def test_get_dummies_pandas_after_read_snowflake(session):
...
+            expect_exception=True,
+            expect_exception_type=ValueError,
+            expect_exception_match=re.escape(
+                "dtype=object is not a valid dtype for get_dummies"
+            ),
         )
 
 
+@sql_count_checker(query_count=0)
+def test_dummy_na_negative():
```
sfc-gh-lmukhopadhyay, line 334: Can we add a test with all NaN values with different dtypes?

tests/integ/modin/frame/test_get_dummies.py
sfc-gh-mvashishtha: We still use the client-side representation for timedelta, so we have to do a bit of extra work to make timedelta work for this particular method. I don't think it's a worth a comment here since there a lot of cases like this.

tests/integ/modin/frame/test_get_dummies.py
sfc-gh-mvashishtha: The values shouldn't affect what the indicator variables look like, so I don't think we need to add a test like that.

PR #3897, opened 2025-10-14
SNOW-2430625:encapsulate local ingestion and udtf ingestion, get rid of driver reference on top level

sfc-gh-yuwang:

   

   Fixes SNOW-2430625

### DIFF
--- a/src/snowflake/snowpark/_internal/data_source/datasource_partitioner.py
+++ b/src/snowflake/snowpark/_internal/data_source/datasource_partitioner.py
@@ -30,7 +30,10 @@ from snowflake.snowpark.types import (
     DateType,
     DataType,
 )
-import snowflake
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import snowflake.snowpark
 
 logger = logging.getLogger(__name__)
 
@@ -172,7 +175,7 @@ class DataSourcePartitioner:
             self.num_partitions,
         )
 
-    def udtf_ingestion(
+    def _udtf_ingestion(
         self,
         session: "snowflake.snowpark.Session",
         schema: StructType,
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -1991,7 +1991,7 @@ class DataFrameReader:
                 table_type="temp",
                 statement_params=statements_params_for_telemetry,
             )
-            df = partitioner.udtf_ingestion(
+            df = partitioner._udtf_ingestion(
                 self._session,
                 struct_schema,
                 partitions_table,


src/snowflake/snowpark/_internal/data_source/datasource_partitioner.py
```diff
@@ -171,6 +180,32 @@ def partitions(self) -> List[str]:
             self.num_partitions,
         )
 
+    def udtf_ingestion(
```
sfc-gh-mayliu: wondering how this function will be used and tested?

src/snowflake/snowpark/_internal/data_source/datasource_partitioner.py
```diff
@@ -30,6 +31,7 @@
     DateType,
     DataType,
 )
+import snowflake
```
sfc-gh-aling: what's this import for?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1839,29 +1838,19 @@ def create_sqlite_connection(timeout=5.0, isolation_level=None, **kwargs):
             "udtf_ingestion" if udtf_configs is not None else "local_ingestion"
         )
 
-        table_or_query = table or query
-        is_query = True if table is None else False
+        # prepare parameter for data source partitioner
+        ds_par_params = locals()
```
sfc-gh-aling: the locals() will contain all the local variables not just the ones we require to instantiate DataSourcePartitioner.

besides it loses readability and explicitness, I think we should keep using the explicit dict approach

src/snowflake/snowpark/_internal/data_source/datasource_partitioner.py
```diff
@@ -87,6 +89,13 @@ def __init__(
             else None
         )
 
+    @classmethod
+    def from_dict(cls, init_dict: dict) -> "DataSourcePartitioner":
+        """Create safely from a dict, ignoring extra fields."""
+        params = inspect.signature(cls).parameters
+        filtered = {k: v for k, v in init_dict.items() if k in params}
```
sfc-gh-aling: if we switch to the normal dict approach instead of locals, you probably won't need the filter logic

src/snowflake/snowpark/_internal/data_source/datasource_partitioner.py
sfc-gh-yuwang: this is for the type hint of udtf_ingestion function added to partitioner

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: the from_dict function in partitioner will filter out any parameter that does not belong to it.
This change is meant for custom data source as custom data source would have a different init function that accept different parameters, but I think I can move this change to later PR if this change is necessary

src/snowflake/snowpark/_internal/data_source/datasource_partitioner.py
sfc-gh-aling: this may a side effect of loading all modules under the snowflake, which slows down the module loading (the infra team cares loading time)

a good practice practice is to:
```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from snowflake.snowpark....  # import the class you need
```

PR #3906, opened 2025-10-16
SNOW-2359402 Enabled autoswitching on some unsupported args

sfc-gh-jenzhang:

   

   Fixes [SNOW-2359402](https://snowflakecomputing.atlassian.net/browse/SNOW-2359402)

[SNOW-2359402]: https://snowflakecomputing.atlassian.net/browse/SNOW-2359402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -626,7 +626,7 @@ def get_unsupported_args_reason(
 
 
 def register_query_compiler_method_not_implemented(
-    api_cls_name: Optional[str],
+    api_cls_names: List[Optional[str]],
```
sfc-gh-joshi: You could consider turning this into `List[Optional[str]] | str` instead, then internally converting str -> [str] so devs don't have to specify a one-item list if they're just registering a single object.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -644,22 +644,24 @@ def register_query_compiler_method_not_implemented(
...
                           If None, method is treated as completely unimplemented.
     """
-    reg_key = MethodKey(api_cls_name, method_name)
 
-    # register the method in the hybrid switch for unsupported args
-    if unsupported_args is None:
-        HYBRID_SWITCH_FOR_UNIMPLEMENTED_METHODS.add(reg_key)
-    else:
-        HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS[reg_key] = unsupported_args
+    for api_cls_name in api_cls_names:
```
sfc-gh-joshi, line 659: you should add an assert that `api_cls_names` is non-empty

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4240,7 +4274,7 @@ def sort_index(
         1.0    c
         dtype: object
         """
-        if axis in (1, "index"):
+        if axis in (1, "columns"):
```
sfc-gh-joshi: Nice catch. You should be able to just check `axis == 1` here actually, since the frontend layer parses other column descriptors like "columns" into an `int` value.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10377,6 +10444,63 @@ def pivot(
...
 
+    @register_query_compiler_method_not_implemented(
+        [None],
+        "pivot_table",
+        UnsupportedArgsRule(
+            unsupported_conditions=[
+                ("sort", False),
+                (
+                    lambda args: (
+                        args.get("index")
```
```suggestion
                        args.get("index") is not None
```

This will otherwise pass over false-y values like `""` and `[]` (this may have been a bug in the original code).

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10377,6 +10444,63 @@ def pivot(
...
+                            not isinstance(args.get("index"), str)
+                            and not all([isinstance(v, str) for v in args.get("index")])
+                            and None not in args.get("index")
+                        )
+                    ),
+                    "non-string of list of string index is not yet supported for pivot_table",
+                ),
+                (
+                    lambda args: (
+                        args.get("columns")
```
```suggestion
                        args.get("columns") is not None
```

src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -1623,6 +1623,7 @@ def sort_values(
             na_position=na_position,
             ignore_index=ignore_index,
             key=key,
+            axis=axis,
```
sfc-gh-joshi: Is this addition necessary? The two branches are already meant to handle the two distinct axes cases.

sfc-gh-jenzhang: you are right. Since I added a default in `def sort_columns_by_row_values(`, this is redundant. I will remove it!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jenzhang: sure I added the assert.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -21239,7 +21392,7 @@ def _stack_helper(
         return qc
 
     @register_query_compiler_method_not_implemented(
-        api_cls_name="DataFrame",
+        api_cls_names=["DataFrame"],
```
```suggestion
        api_cls_names="DataFrame",
```

sfc-gh-joshi (approved):
Discussed with @sfc-gh-jenzhang offline about the added axis=1 argument to sort_columns_by_row_values, even though it's technically redundant because the QC method always operates on axis=1. This is necessary because when `stay_cost` is computed, the decorator compares against the frontend method's arguments to return `COST_IMPOSSIBLE`, but when the wrapped QC method is called, the decorated examines the arguments of the QC method instead.

Alternative options to enable switching for this case were:
- Don't add a QC argument, and use `lambda args: args.get("axis", 1) == 1` for the condition instead of just `("axis", 1)` -- works but is harder to understand
- Keep unsupported condition as `("axis", 1)` but don't add `axis=1` argument: will trigger an auto-switch, but when switching is disabled will not trigger the decorator's error message code path because the QC method lacks an `axis` parameter
- Set unsupported condition to be unconditional (e.g. `lambda args: True`): will trigger auto-switch even for supported use cases

We decided just adding a docstring comment explaining the addition of the parameter would be better.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -10377,6 +10447,63 @@ def pivot(
...
+        UnsupportedArgsRule(
+            unsupported_conditions=[
+                ("sort", False),
+                (
+                    lambda args: (
+                        args.get("index") is not None
+                        and (
+                            not isinstance(args.get("index"), str)
+                            and not all([isinstance(v, str) for v in args.get("index")])
+                            and None not in args.get("index")
```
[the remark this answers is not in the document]
sfc-gh-jenzhang: this is okay bc we are already checking that its not a string

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -4240,7 +4277,7 @@ def sort_index(
         1.0    c
         dtype: object
         """
-        if axis in (1, "index"):
+        if axis == 1:
```
sfc-gh-nkrishna, line 4294: Why remove `index`?

sfc-gh-jenzhang: This is redundant because the frontend layer already parses "index" to 1.

PR #3907, opened 2025-10-16
SNOW-2435161: Replace importlib guard by sys.modules check in session initialization

sfc-gh-joshi:

   

   Fixes SNOW-2435161

`importlib.util.find_spec` checks if a module _can_ be imported in the current environment, while `sys.modules` checks whether a module was _already_ imported. This check guards a session flag; if modin was already imported, then we need to `import modin.config` and set the value of the relevant configuration variable. The previous check would improperly import `modin` if it was present in the environment even if the user did not intend to do so; this triggered a bug in narwhals where the library checked whether `"modin" in sys.modules` was true instead of `"modin.pandas" in sys.modules` ([code](https://github.com/narwhals-dev/narwhals/blob/main/narwhals/dependencies.py#L57-L61)), causing an ImportError even if the user did not intend to use modin at all.

Relaxing our check to use `sys.modules` instead avoids unnecessarily polluting the namespace, avoiding this crash.

Reproducer script (raises ImportError with snowflake-snowpark-python==1.40.0, modin==0.37.1, and plotly>=6 installed, no automated test since this requires a pretty specific environment setup):
```python3
import importlib
session = Session.builder.getOrCreate()
df = session.sql("SELECT 0 AS X, 1 AS Y").to_pandas()
print("type", type(df))
print(importlib.util.find_spec("modin"))

import plotly.express as px
# Uncomment the below line and everything magically works
#import modin.pandas
px.line(df, x="X", y="Y")
```

### DIFF
--- a/src/snowflake/snowpark/_internal/data_source/drivers/base_driver.py
+++ b/src/snowflake/snowpark/_internal/data_source/drivers/base_driver.py
@@ -177,6 +177,7 @@ class BaseDriver:
                 packages=packages or UDTF_PACKAGE_MAP.get(self.dbms_type),
                 imports=imports,
                 statement_params=statement_params,
+                _emit_ast=_emit_ast,  # internal function call, _emit_ast will be set to False by the caller
             )
         logger.debug(f"register ingestion udtf takes: {udtf_register_time()} seconds")
         call_udtf_sql = f"""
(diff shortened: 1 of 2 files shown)


sfc-gh-eferguson:
> hey
> 
> > The previous check would improperly import modin if it was present in the environment even if the user did not intend to do so; this triggered a bug in narwhals where the library checked whether "modin" in sys.modules was true instead of "modin.pandas" in sys.modules ([code](https://github.com/narwhals-dev/narwhals/blob/main/narwhals/dependencies.py#L57-L61)), causing an ImportError even if the user did not intend to use modin at all.
> 
> just dropping by to say, if there's anything you'd like us to change in narwhals which can facilitate snowpark, please do let us know

@MarcoGorelli I put up a PR [here](https://github.com/narwhals-dev/narwhals/pull/3219) to change the import line. If you can review I would appreciate it.

PR #3908, opened 2025-10-16
SNOW-2435290: Add support for groupby.agg/min/max/count/sum/mean/median/std/var in faster pandas

sfc-gh-helmeleegy:

   

   Fixes SNOW-2435290

   Add support for groupby.agg/min/max/count/sum/mean/median/std/var in faster pandas.


src/snowflake/snowpark/modin/plugin/docstrings/series_utils.py
```diff
@@ -2642,7 +2642,7 @@ def seconds():
         0    1
         1    2
         2    3
-        dtype: int64
+        dtype: int8
```
sfc-gh-nkrishna: Why did data type change here?

sfc-gh-helmeleegy: I'm not 100% sure. But the sql queries look a bit different in faster pandas compared to the original snowpark pandas. So the new order of sql operations may have caused the typing system on the server side to infer a slightly different final type.

sfc-gh-jkew: Are you sure this is not related to the snowflake-connector upgrade? This feels like an arrow change.

tests/integ/modin/test_faster_pandas.py
```diff
@@ -123,9 +123,9 @@ def test_read_filter_join_on_index(session):
     )
 
 
-@sql_count_checker(query_count=3)
-def test_read_filter_groupby_agg(session):
-    # test a chain of operations that are not fully supported in faster pandas
+@sql_count_checker(query_count=3, join_count=2)
+def test_read_filter_iloc_index(session):
```
sfc-gh-jkew, line 128: This test change seems out of place. Did this get pulled in from a different branch?

src/snowflake/snowpark/modin/plugin/docstrings/series_utils.py
sfc-gh-joshi: Seems to be failing in main: https://github.com/snowflakedb/snowpark-python/actions/runs/18579579403/job/52971829830

tests/integ/modin/test_faster_pandas.py
sfc-gh-helmeleegy: No it's intentional. This test is meant to represent a scenario that is not supported in faster pandas. Groupby was used for that purpose. With this PR, groupby is now supported, so we're using iloc instead as an example of an unsupported operation. (Actually another PR is adding partial support to iloc. But this should remain to be unsupported.)

src/snowflake/snowpark/modin/plugin/docstrings/series_utils.py
sfc-gh-helmeleegy: I see. In this case, let's fix this in a separate PR since the root cause is not related to this PR.
Here is the separate fix PR: https://github.com/snowflakedb/snowpark-python/pull/3911
@sfc-gh-jkew @sfc-gh-joshi @sfc-gh-nkrishna

tests/integ/modin/test_faster_pandas.py
```diff
@@ -252,6 +252,61 @@ def test_duplicated(session):
...
+        "count",
+        "sum",
+        "mean",
+        "median",
+        "std",
+        "var",
+    ],
+)
+@sql_count_checker(query_count=6)
+def test_groupby_agg(session, func):
```
sfc-gh-joshi, line 325: It seems like a lot of the fasters pandas tests have some pretty similar setup/assert/teardown code. Could you add some of that logic to a shared logic/fixture in the future?

sfc-gh-jkew (approved):

sfc-gh-lmukhopadhyay (approved):

PR #3918, opened 2025-10-20
SNOW-2396205: Support random_state in sample.

sfc-gh-mvashishtha:
Implement the random_state parameter DataFrame.sample and Series.sample by sorting with a seeded random order and then selecting the top `n` or `frac * len(dataset)` rows. We use this solution because we can't use the built-in SAMPLE with SEED for this use case.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -626,7 +626,7 @@ HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS: dict[MethodKey, UnsupportedArgsRule] = {}
 
 
 def register_query_compiler_method_not_implemented(
-    api_cls_names: Union[list[Optional[str]], Optional[str]],
+    api_cls_name: Optional[str],
     method_name: str,
     unsupported_args: Optional["UnsupportedArgsRule"] = None,
 ) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
@@ -644,30 +644,22 @@ def register_query_compiler_method_not_implemented(
     without meaningful benefit.
 
     Args:
-        api_cls_names: Frontend class names (e.g. "BasePandasDataset", "Series", "DataFrame", or None). It can be a list if multiple api_cls_names are needed.
+        api_cls_name: Frontend class name (e.g., "BasePandasDataset", "Series", "DataFrame", "None").
         method_name: Method name to register.
         unsupported_args: UnsupportedArgsRule for args-based auto-switching.
                           If None, method is treated as completely unimplemented.
     """
+    reg_key = MethodKey(api_cls_name, method_name)
 
-    if isinstance(api_cls_names, str) or api_cls_names is None:
-        api_cls_names = [api_cls_names]
-    assert (
-        api_cls_names
-    ), "api_cls_names must be a string (e.g., 'DataFrame', 'Series') or a list of strings (e.g., ['DataFrame', 'Series']) or None for top-level functions"
+    # register the method in the hybrid switch for unsupported args
+    if unsupported_args is None:
+        HYBRID_SWITCH_FOR_UNIMPLEMENTED_METHODS.add(reg_key)
+    else:
+        HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS[reg_key] = unsupported_args
 
-    for api_cls_name in api_cls_names:
-        reg_key = MethodKey(api_cls_name, method_name)
-
-        # register the method in the hybrid switch for unsupported args
-        if unsupported_args is None:
-            HYBRID_SWITCH_FOR_UNIMPLEMENTED_METHODS.add(reg_key)
-        else:
-            HYBRID_SWITCH_FOR_UNSUPPORTED_ARGS[reg_key] = unsupported_args
-
-        register_function_for_pre_op_switch(
-            class_name=api_cls_name, backend="Snowflake", method=method_name
-        )
+    register_function_for_pre_op_switch(
+        class_name=api_cls_name, backend="Snowflake", method=method_name
+    )
 
     def decorator(query_compiler_method: Callable[..., Any]) -> Callable[..., Any]:
         @functools.wraps(query_compiler_method)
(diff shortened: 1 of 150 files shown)


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15950,59 +15950,117 @@ def sample(
...
+                np.random.Generator,
+            ),
+        ):
+            ErrorMessage.not_implemented("non-integer `random_state` is not supported.")
 
-        if random_state is not None:
-            ErrorMessage.not_implemented("`random_state` is not supported.")
-
+        if random_state is not None and not is_integer(random_state):
+            raise ValueError("random_state must be an integer or None.")
```
sfc-gh-joshi, line 16440: Why is this check separate from the above not_implemented branch?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15950,59 +15950,117 @@ def sample(
...
         assert n is not None or frac is not None
+        if not replace:
+            if frac is not None and frac > 1:
+                raise ValueError(
+                    "Replace has to be set to `True` when upsampling the population `frac` > 1."
+                )
```
```suggestion
        if not replace and frac is not None and frac > 1:
            raise ValueError(
                "Replace has to be set to `True` when upsampling the population `frac` > 1."
            )
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15950,59 +15950,117 @@ def sample(
...
-            )
 
+        # use builtin('random') instead of snowflake.snowpark.functions.random
+        # because the latter does not take Column inputs, but we want to use
+        # pandas_lit() to create the seed.
+        # if random_state is None, we have to call random() with no arguments.
+        # random(NULL) is not valid.
+        random_maybe_with_state = builtin("random")(
+            *(tuple() if random_state is None else (pandas_lit(random_state),))
+        )
```
sfc-gh-joshi: nit: IMO it would be easier to read as

```
random = builtin("random")
random() if random_state is None else random(pandas_lit(random_state))
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15950,59 +15950,117 @@ def sample(
...
-                        self._modin_frame.ordered_dataframe.row_position_snowflake_quoted_identifier
-                    ],
+            sampled_odf = sampled_row_positions_odf.join(
+                right=self._modin_frame.ordered_dataframe,
+                left_on_cols=[sampled_row_position_identifier],
+                right_on_cols=[
+                    self._modin_frame.ordered_dataframe.row_position_snowflake_quoted_identifier
+                ],
+            )
+            # if random_state is not None, the result is already deterministic.
```
```suggestion
            # if random_state is not None, the result is seeded and already deterministic.
```

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -15950,59 +15950,117 @@ def sample(
...
+            # table. A snowflake query compiler does not necessarily correspond
+            # to a particular snowflake table, and even though we could sample
+            # an intermediate table produce with cache_result(), we need to
+            # select a set of rows that is deterministic with respect to the
+            # table length rather than with respect to the query compiler or
+            # even the dataframe. For example, pd.DataFrame(list(range(1000))).sample(n=1, random_state=0) and
+            # pd.DataFrame(list(range(1000))[::-1]).sample(n=1, random_state=0)
+            # select the same row position.
+            # We use this alternate implementation rather than the generator
+            # one that we use for replace=True because we can avoid a join.
```
sfc-gh-joshi: Do you know how performance of your select + order by + limit approach compares to the builtin SAMPLE?

sfc-gh-joshi: Can this approach also be applied to the seeded case to eliminate the join?

I guess you would lose the guarantee that `df.sample(n=2, random_state=0)[0] == df.sample(n=1, random_state=0)[0]`, but is that something that's expected from the API? If so, we should add a test for it.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: We should raise a NotImplementedError for the types that are valid in pandas but not in Snowpark pandas, and a ValueError for types that are invalid in both.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: Done.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-mvashishtha: I was surprised to find that the built-in sample appears to be much, much faster for large data, e.g.  `repr(df.sample(frac=3/100))` on `SNOWPANDAS_PERF.TPCH_SF10.LINEITEM` on a large warehouse took 5.84 second with built-in sample + cache_result, but 29.6 seconds with the select + sort + limit approach.

I added a comment saying that the built-in sample was more efficient.

> Can this approach also be applied to the seeded case to eliminate the join?

Seeded for replace=False will not have a join. With replace=True I think we need the join.

> I guess you would lose the guarantee that df.sample(n=2, random_state=0)[0] == df.sample(n=1, random_state=0)[0], but is that something that's expected from the API? If so, we should add a test for it.

I don't think this is expected from the API.

docs/source/modin/supported/dataframe_supported.rst
```diff
@@ -384,8 +384,8 @@ Methods
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
 | ``rtruediv``                | P                               | ``level``                        |                                                    |
 +-----------------------------+---------------------------------+----------------------------------+----------------------------------------------------+
-| ``sample``                  | P                               |                                  | ``N`` if ``weights`` or ``random_state`` is        |
-|                             |                                 |                                  | specified when ``axis = 0``                        |
+| ``sample``                  | P                               |                                  | ``N`` if ``weights`` is  specified when            |
```
sfc-gh-jkew: Because this adds a join and a sort; you may want to mention that the random_state argument will not use Snowflake's underlying function which may have performance implications.

sfc-gh-mvashishtha: Done.

PR #3922, opened 2025-10-20
SNOW-2441438: Add modin dependency to import error tests.

sfc-gh-mvashishtha:
With this change, `tox -e snowpark_pandas_modin_pandas_import_error` passes instead of failing.


sfc-gh-joshi (commented):
I believe these tests were intended to run in an environment that specifically did not have modin or pandas installed, hence why the tests are mocking the modin/pandas imports. Is it possible to update the test expectations instead?

PR #3929, opened 2025-10-21
SNOW-2390752  - Series replace fails with KeyError ( Test Only )

sfc-gh-jkew:
Additional tests for a reported issue involving series replace. In the original issue; a `Series['some idx'].replace('str', 'str2')` failed with a KeyError,  but ultimately we were not able to replicate this behavior on any engine. This CL adds tests for the scenario in the expectation that there may be some other cause for the issue.

Related to: SNOW-2390752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -155,6 +155,7 @@
   - `groupby.var`
   - `groupby.nunique`
   - `groupby.size`
+  - `groupby.apply`
   - `drop_duplicates`
 - Reuse row count from the relaxed query compiler in `get_axis_len`.
 
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -4888,6 +4888,52 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
     def groupby_apply(
         self,
         by: Any,
+        agg_func: AggFuncType,
+        axis: int,
+        groupby_kwargs: dict[str, Any],
+        agg_args: Any,
+        agg_kwargs: dict[str, Any],
+        series_groupby: bool,
+        include_groups: bool,
+        force_single_group: bool = False,
+        force_list_like_to_series: bool = False,
+    ) -> "SnowflakeQueryCompiler":
+        """
+        Wrapper around _groupby_apply_internal to be supported in faster pandas.
+        """
+        relaxed_query_compiler = None
+        if self._relaxed_query_compiler is not None:
+            relaxed_query_compiler = (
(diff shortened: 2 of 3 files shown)


tests/integ/modin/frame/test_replace.py
```diff
@@ -23,6 +23,45 @@ def snow_df():
...
+    test_ndf = test_sdf.to_pandas()
+    test_mdf = test_sdf.move_to("Pandas")
+    assert test_mdf.get_backend() == "Pandas"
+
+    eval_snowpark_pandas_result(
+        test_sdf, test_ndf, lambda df: df["col1"].replace(["two"], ["a"])
+    )
+    snow_result = test_sdf["col1"].replace(["two"], ["a"])
+    modin_result = test_mdf["col1"].replace(["two"], ["a"])
+    assert modin_result.to_pandas().equals(snow_result.to_pandas())
```
sfc-gh-joshi: Is this any different from the `eval_snowpark_pandas_result` call?

tests/integ/modin/frame/test_replace.py
```diff
@@ -23,6 +23,45 @@ def snow_df():
...
+
+
+@sql_count_checker(query_count=4, join_count=2)
+def test_replace_mauricio_series():
+
+    input_dict = pd.Series({"a": "apple", "socks": "socks are darned", "b": "bee"})
+    input_dict_native = input_dict.to_pandas()
```
sfc-gh-joshi:
```suggestion
    input_dict_native = native_pd.Series({"a": "apple", "socks": "socks are darned", "b": "bee"})
    input_dict = pd.Series(input_dict_native)
```
reduces the query count

sfc-gh-mvashishtha (approved):

tests/integ/modin/frame/test_replace.py
sfc-gh-jkew: No; it's the same essentially, and the data types are simple enough that we do not need any special handling. I could not use eval here though because of set_attr handling and checking; which makes assumptions about how attributes on a dataframe are handled in the test code.

tests/integ/modin/frame/test_replace.py
sfc-gh-jkew: Good idea

tests/integ/modin/frame/test_replace.py
sfc-gh-joshi: Ah, I didn't realize that the above eval call used `test_ndf` instead of `test_mdf`. Can you rename the variables to be a little lexicographically further apart?

sfc-gh-jkew: Done

sfc-gh-joshi (approved):

PR #3930, opened 2025-10-21
SNOW-2362373: Fix bug with groupby.agg named aggregation path

sfc-gh-nkrishna:
Fixes SNOW-2362373. Fixes a bug with `groupby.agg` where tuples are treated as multi-index level using uses the [named aggregation](https://pandas.pydata.org/docs/user_guide/groupby.html#groupby-aggregate-named) code path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@
 - Fixed a bug where writing Snowpark pandas dataframes on the pandas backend with a column multiindex to Snowflake with `to_snowflake` would raise `KeyError`.
 - Fixed a bug that `DataFrameReader.dbapi` (PuPr) is not compatible with oracledb 3.4.0.
 - Fixed a bug where `modin` would unintentionally be imported during session initialization in some scenarios.
+- Fixed a bug where `session.udf|udtf|udaf|sproc.register` failed when an extra session argument was passed. These methods do not expect a session argument; please remove it if provided.
 - Fixed a bug in `DataFrameGroupBuy.agg` where func is a list of tuples used to set the names of the output columns.
 
 #### Improvements
@@ -128,6 +129,7 @@
   - `sort_values`
   - `loc` (setting columns)
   - `to_datetime`
+  - `rename`
   - `drop`
   - `invert`
   - `duplicated`
--- a/tests/integ/scala/test_function_suite.py
+++ b/tests/integ/scala/test_function_suite.py
@@ -5735,20 +5735,16 @@ def test_interval_year_month_from_parts(session):
         assert field.datatype == YearMonthIntervalType(0, 1)
 
     result_literals_no_alias = literals_schema_result_no_alias.collect()
-    assert (
-        result_literals_no_alias[0][
-            "CAST (CONCAT(IFF((((2 * 12) + 5) < 0), '-', ''),  CAST ( CAST (FLOOR((ABS(((2 * 12) + 5)) / 12)) AS INT) AS STRING), '-',  CAST ( CAST (FLOOR((ABS(((2 * 12) + 5)) % 12)) AS INT) AS STRING)) AS INTERVAL YEAR TO MONTH)"
-        ]
-        == "+2-05"
-    )
+    if session.eliminate_numeric_sql_value_cast_enabled:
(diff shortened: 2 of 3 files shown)


src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
```diff
@@ -899,7 +899,9 @@ def _is_supported_snowflake_agg_func(
     if isinstance(agg_func, tuple) and len(agg_func) == 2:
         # For named aggregations, like `df.agg(new_col=("old_col", "sum"))`,
         # take the second part of the named aggregation.
-        agg_func = agg_func[0]
+        agg_func = (
+            agg_func.func if isinstance(agg_func, AggFuncWithLabel) else agg_func[1]
```
sfc-gh-joshi, line 903: Can you add a comment here explaining what inputs this branch is meant to match?

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
```diff
@@ -1536,7 +1543,12 @@ def generate_column_agg_info(
         for func_info, label, identifier in zip(
             agg_func_list, agg_col_labels, agg_col_identifiers
         ):
-            func = func_info.func
+            # If func_info.func is a tuple, treat as named aggregation and return the aggregate function
```
sfc-gh-joshi, line 1546: Shouldn't this be changed at the level of where `func_info` is being parsed instead?
Also, is there any validation on `len(func_info.func)` here?

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
```diff
@@ -1378,13 +1380,18 @@ def get_agg_func_to_col_map(
     return agg_func_to_col_map
 
 
-def get_pandas_aggr_func_name(aggfunc: AggFuncTypeBase) -> str:
+def get_pandas_aggr_func_name(aggfunc: AggFuncTypeBase) -> Union[str, Any, None]:
```
sfc-gh-joshi: Why did this return type change?

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@
 - Fixed a bug where writing Snowpark pandas dataframes on the pandas backend with a column multiindex to Snowflake with `to_snowflake` would raise `KeyError`.
 - Fixed a bug that `DataFrameReader.dbapi` (PuPr) is not compatible with oracledb 3.4.0.
 - Fixed a bug where `modin` would unintentionally be imported during session initialization in some scenarios.
+- Fixed a bug with `DataFrameGroupBy.agg` where tuples are treated as multiindex level using the named aggregation code path.
```
sfc-gh-joshi: "named aggregation code path" is a little unclear to users. It might be better to frame this as a feature instead, like "Added support for `DataFrameGroupBuy.agg` where `func` is a list of tuples used to set the names of the output columns."

sfc-gh-helmeleegy: +1

tests/integ/modin/groupby/test_groupby_basic_agg.py
```diff
@@ -1321,3 +1321,23 @@ def test_groupby_timedelta_var(self):
...
+                {
+                    "team": ["A", "B"],
+                    "score": [10, 15],
+                }
+            ),
+            lambda df: df.groupby("team").agg({"score": [("total_score", "sum")]}),
+        )
+
+    @sql_count_checker(query_count=1)
+    def test_groupby_agg_named_aggregation_explicit(self):
```
sfc-gh-helmeleegy, line 1338: We already had support for this, right? I also believe we did have test coverage for it. But I guess it doesn't hurt to show the contrast here between implicit and explicit named aggregation.

sfc-gh-helmeleegy (approved):
Thanks!

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
sfc-gh-nkrishna: Good catch, didn't explicitly wrap the default value for `getattr` in a str, done.

src/snowflake/snowpark/modin/plugin/_internal/aggregation_utils.py
sfc-gh-nkrishna: I think this is the right place to get `func` since this is where the `ColumnAggregationInfo` is being constructed for each aggregation and the associated labels are being constructed just above on Line 1529.

PR #3946, opened 2025-10-23
SNOW-2443512: Add support for scalar string and binary functions ( part 2 )

sfc-gh-aherreraaguilar:

   

   Fixes [SNOW-2443512](https://snowflakecomputing.atlassian.net/browse/SNOW-2443512)

Add support for the scalar string and binary functions:

- `hex_decode_string`
- `jarowinkler_similarity`
- `parse_url`
- `regexp_instr`
- `regexp_like`
- `regexp_substr`
- `regexp_substr_all`
- `rtrimmed_length`
- `space`
- `split_part`

[SNOW-2443512]: https://snowflakecomputing.atlassian.net/browse/SNOW-2443512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### New Features
 
+- Added support for `DataFrame.lateral_join`
 - Added support for `Session.client_telemetry`.
 - Added support for `Session.udf_profiler`.
 - Added support for `functions.ai_translate`.
@@ -56,6 +57,7 @@
 
 - Enhanced `DataFrame.sort()` to support `ORDER BY ALL` when no columns are specified.
 - Catalog API now uses SQL commands instead of SnowAPI calls. This new implementation is more reliable now.
+- Removed experimental warning from `Session.cte_optimization_enabled`.
 
 #### Dependency Updates
 
@@ -134,6 +136,10 @@
   - `dt.month_name`
   - `dt.day_name`
   - `dt.strftime`  
+  - `dt.dayofweek`  
+  - `dt.weekday`  
+  - `dt.dayofyear`    
+  - `dt.isocalendar`  
   - `rolling.min`
   - `rolling.max`
   - `rolling.count`
@@ -154,6 +160,37 @@
   - `cumsum`
   - `cummin`
   - `cummax`
+  - `groupby.groups`
+  - `groupby.indices`
+  - `groupby.first`
+  - `groupby.last`
+  - `groupby.rank`
+  - `groupby.shift`
+  - `groupby.cumcount`
+  - `groupby.cumsum`
+  - `groupby.cummin`
+  - `groupby.cummax`
+  - `groupby.any`
+  - `groupby.all`
+  - `groupby.unique`
+  - `groupby.get_group`
+  - `groupby.rolling`
+  - `groupby.resample`
+  - `to_snowflake`
+  - `to_snowpark`
+  - `resample.min`
+  - `resample.max`
+  - `resample.count`
+  - `resample.sum`
+  - `resample.mean`
+  - `resample.median`
+  - `resample.std`
+  - `resample.var`
+  - `resample.size`
+  - `resample.first`
+  - `resample.last`
+  - `resample.quantile`
+  - `resample.nunique`
 - Make faster pandas disabled by default (opt-in instead of opt-out).
 - Improve performance of `drop_duplicates` by avoiding joins when `keep!=False` in faster pandas.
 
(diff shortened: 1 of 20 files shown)


src/snowflake/snowpark/_functions/scalar_functions.py
```diff
@@ -3971,3 +3971,747 @@ def try_to_geometry(
...
+        [Row(SKIP_TO_PARAMS_AND_GROUP=1)]
+    """
+    if position is None:
+        position = lit(1)
+    if occurrence is None:
+        occurrence = lit(1)
+    if option is None:
+        option = lit(0)
+    if regexp_parameters is None:
+        regexp_parameters = lit("c")
```
sfc-gh-aling, line 4914: I have two questions here:
1. do we need to check the dependency between the optional args? what happen if `occurrence` is set but `position` is not set. I feel this is not allowed in sql as in sql the arguments are positional, users have to specify `position` first then `occurrence` then possibly the rest.

3. I see there is a long note about
`Note
By default, REGEXP_INSTR returns the begin or end character offset for the entire matching part of the subject. However, if the e (for “extract”) parameter is specified, REGEXP_INSTR returns the begin or end character offset for the part of the subject that matches the first sub-expression in the pattern. If e is specified but a group_num is not also specified, then the group_num defaults to 1 (the first group). If there is no sub-expression in the pattern, REGEXP_INSTR behaves as if e was not set. For examples that use e, see [Examples](https://docs.snowflake.com/en/sql-reference/functions/regexp_instr#examples) in this topic.`

shall we not set the default value from client at all and let server decide the default rule? I feel this way it's simpler

src/snowflake/snowpark/_functions/scalar_functions.py
```diff
@@ -3971,3 +3971,747 @@ def try_to_geometry(
...
+
+@publicapi
+def regexp_substr(
+    subject: ColumnOrName,
+    pattern: ColumnOrName,
+    position: ColumnOrName = None,
+    occurrence: ColumnOrName = None,
+    regex_parameters: ColumnOrName = None,
+    group_num: ColumnOrName = None,
+    _emit_ast: bool = True,
```
sfc-gh-aling, line 4988: same comment as above

src/snowflake/snowpark/_functions/scalar_functions.py
```diff
@@ -3971,3 +3971,747 @@ def try_to_geometry(
...
+    ]
+
+    if group_num is not None:
+        args.append(_to_col_if_str(group_num, "regexp_substr"))
+
+    return builtin("regexp_substr", _emit_ast=_emit_ast)(*args)
+
+
+@publicapi
+def regexp_substr_all(
```
sfc-gh-aling, line 5080: same for this func

src/snowflake/snowpark/_functions/scalar_functions.py
```diff
@@ -3971,3 +3971,747 @@ def try_to_geometry(
...
+    string_col = _to_col_if_str(string, "split_part")
+    delimiter_col = _to_col_if_str(delimiter, "split_part")
+    part_number_col = _to_col_if_str(part_number, "split_part")
+    return builtin("split_part", _emit_ast=_emit_ast)(
+        string_col, delimiter_col, part_number_col
+    )
+
+
+@publicapi
+def strtok(
```
sfc-gh-aling: I see functions starting from this has been implemented in https://github.com/snowflakedb/snowpark-python/pull/3947

sfc-gh-aherreraaguilar: Ups, my bad, I removed them from here.

src/snowflake/snowpark/_functions/scalar_functions.py
sfc-gh-aherreraaguilar: Hello @sfc-gh-aling ,
regarding these questions, the reason why I am setting the default values is that when I call the function
`builtin("regexp_instr", _emit_ast=_emit_ast)(*args)`
, passing the *args list, it does not explicitly indicate the name for each parameter. So, for example, if I send regexp_instr(A,B,C) the order will be the same when calling the builtin, but when calling something like regexp_instr(A, C=C), the call in the builtin will be regexp_instr(A, C)

sfc-gh-aherreraaguilar:
@sfc-gh-mvegaalvarez @sfc-gh-aling , I tried to reopen this PR, but it is not possible, so I created this new [PR](https://github.com/snowflakedb/snowpark-python/pull/4027) to merge these changes.

PR #3955, opened 2025-10-24
SNOW-1051741: df.apply(axis=1) should preserve the original index

sfc-gh-jkew:
`df.apply(axis=1)` should preserve the original index. Previously we would return a RangeIndex regardless of the original index. This approach passes the index data into the underlying UDTF.

Mostly AI written approach, but with original tests for verification.

Fixes SNOW-1051741

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -16910,6 +16910,35 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         *args: Any,
         **kwargs: Any,
     ) -> "SnowflakeQueryCompiler":
+        """
+        Wrapper around _rolling_count_internal to be supported in faster pandas.
+        """
+        relaxed_query_compiler = None
+        if self._relaxed_query_compiler is not None:
+            relaxed_query_compiler = (
+                self._relaxed_query_compiler._rolling_count_internal(
+                    fold_axis=fold_axis,
+                    rolling_kwargs=rolling_kwargs,
+                    numeric_only=numeric_only,
+                    **kwargs,
+                )
+            )
+        qc = self._rolling_count_internal(
+            fold_axis=fold_axis,
+            rolling_kwargs=rolling_kwargs,
+            numeric_only=numeric_only,
+            **kwargs,
+        )
+        return self._maybe_set_relaxed_qc(qc, relaxed_query_compiler)
+
+    def _rolling_count_internal(
+        self,
+        fold_axis: Union[int, str],
+        rolling_kwargs: dict,
+        numeric_only: bool = False,
+        *args: Any,
+        **kwargs: Any,
+    ) -> "SnowflakeQueryCompiler":
         return self._window_agg(
             window_func=WindowFunction.ROLLING,
             agg_func="count",
@@ -16918,6 +16947,39 @@ class SnowflakeQueryCompiler(BaseQueryCompiler):
         )
 
     def rolling_sum(
+        self,
+        fold_axis: Union[int, str],
+        rolling_kwargs: dict,
+        numeric_only: bool = False,
+        engine: Optional[Literal["cython", "numba"]] = None,
+        engine_kwargs: Optional[dict[str, bool]] = None,
+        *args: Any,
+        **kwargs: Any,
+    ) -> "SnowflakeQueryCompiler":
+        """
+        Wrapper around _rolling_sum_internal to be supported in faster pandas.
+        """
+        relaxed_query_compiler = None
+        if self._relaxed_query_compiler is not None:
+            relaxed_query_compiler = self._relaxed_query_compiler._rolling_sum_internal(
+                fold_axis=fold_axis,
+                rolling_kwargs=rolling_kwargs,
+                numeric_only=numeric_only,
+                engine=engine,
+                engine_kwargs=engine_kwargs,
+                **kwargs,
+            )
+        qc = self._rolling_sum_internal(
+            fold_axis=fold_axis,
+            rolling_kwargs=rolling_kwargs,
+            numeric_only=numeric_only,
+            engine=engine,
+            engine_kwargs=engine_kwargs,
+            **kwargs,
+        )
+        return self._maybe_set_relaxed_qc(qc, relaxed_query_compiler)
+
+    def _rolling_sum_internal(
         self,
         fold_axis: Union[int, str],
         rolling_kwargs: dict,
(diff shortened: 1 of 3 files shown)


tests/integ/modin/frame/test_apply.py
```diff
@@ -148,6 +148,48 @@ def foo(row) -> str:
         eval_snowpark_pandas_result(snow_df, df, lambda x: x.apply(foo, axis=1))
 
 
+@pytest.mark.parametrize("index", [None, ["a", "b"], [100, 200]])
+@sql_count_checker(query_count=5, join_count=2, udtf_count=1)
+def test_apply_axis_1_index_preservation(index):
```
sfc-gh-helmeleegy, line 153: So these tests would fail prior to this PR?

sfc-gh-helmeleegy (approved):
LGTM, just had one question.

tests/integ/modin/frame/test_apply.py
sfc-gh-jkew: The second two parameters would fail where the index was non-default.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9618,6 +9618,23 @@ def _apply_with_udtf_and_dynamic_pivot_along_axis_1(
         )
 
         # The apply function is encapsulated in a UDTF and run as a stored procedure on the pandas dataframe.
+        # Determine if we should pass index columns to the UDTF
+        # We pass index columns when the index is not the row position itself
+        index_columns_for_udtf = (
+            new_internal_df.index_column_snowflake_quoted_identifiers
+        )
+        if row_position_snowflake_quoted_identifier in index_columns_for_udtf:
+            # The row position IS the index (e.g., RangeIndex), don't pass index columns
```
sfc-gh-jkew: This branch is not necessary.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -443,7 +445,9 @@ def create_udtf_for_apply_axis_1(
         result_type: pandas parameter controlling apply within the UDTF.
         args: pandas parameter controlling apply within the UDTF.
         column_index: The columns of the callee DataFrame, i.e. df.columns as pd.Index object.
-        input_types: Snowpark column types of the input data columns.
+        input_types: Snowpark column types of the input data columns (including index columns).
+        index_column_pandas_labels: The pandas labels for the index columns, if any.
```
sfc-gh-mvashishtha: There are always pandas labels for the index columns. If this is a special case for when the labels are also the row positions, can we say that explicitly?

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -458,8 +462,34 @@ def create_udtf_for_apply_axis_1(
...
+                    # Multi-index case
+                    index = native_pd.MultiIndex.from_arrays(
+                        [index_cols.iloc[:, i] for i in range(num_index_columns)],
+                        names=index_column_pandas_labels
+                        if index_column_pandas_labels
+                        else None,
+                    )
+                data_cols.index = index
+                df = data_cols
+            else:
```
sfc-gh-mvashishtha, line 488: can't we use set_index() in both cases?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9628,6 +9628,18 @@ def _apply_with_udtf_and_dynamic_pivot_along_axis_1(
         )
 
         # The apply function is encapsulated in a UDTF and run as a stored procedure on the pandas dataframe.
+        # Determine if we should pass index columns to the UDTF
+        # We pass index columns when the index is not the row position itself
+        index_columns_for_udtf = (
+            new_internal_df.index_column_snowflake_quoted_identifiers
+        )
+        if row_position_snowflake_quoted_identifier not in index_columns_for_udtf:
```
sfc-gh-mvashishtha: This doesn't look like the correct condition. Shouldn't we check whether `index_columns_for_udtf` is equal to `[internal_frame.row_position_snowflake_quoted_identifier]` or something like that?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9628,6 +9628,18 @@ def _apply_with_udtf_and_dynamic_pivot_along_axis_1(
...
+            # Pass the actual index columns to the UDTF
+            num_index_columns = len(index_columns_for_udtf)
```
sfc-gh-mvashishtha: Does `num_index_columns` get initialized if we don't enter this case?

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -443,7 +445,9 @@ def create_udtf_for_apply_axis_1(
         result_type: pandas parameter controlling apply within the UDTF.
         args: pandas parameter controlling apply within the UDTF.
         column_index: The columns of the callee DataFrame, i.e. df.columns as pd.Index object.
-        input_types: Snowpark column types of the input data columns.
+        input_types: Snowpark column types of the input data columns (including index columns).
+        index_column_pandas_labels: The pandas labels for the index columns, if any.
+        num_index_columns: Number of index columns being passed into the UDTF.
```
sfc-gh-mvashishtha: `num_index_columns` is redundant-- if the users passes `index_column_pandas_labels`, it's `len(index_column_pandas_labels)`; otherwise it's `0`. Can we remove this parameter?

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: Oddly enough; this is the one line I changed from the AI code as the previous condition was garbage. I'm actually not even sure if we need the if statement here at all though; just passing in the pandas labels is fine.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-jkew: removed `num_index_columns`

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-jkew: removed `num_index_columns`

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -443,7 +444,9 @@ def create_udtf_for_apply_axis_1(
         result_type: pandas parameter controlling apply within the UDTF.
         args: pandas parameter controlling apply within the UDTF.
         column_index: The columns of the callee DataFrame, i.e. df.columns as pd.Index object.
-        input_types: Snowpark column types of the input data columns.
+        input_types: Snowpark column types of the input data columns (including index columns).
+        index_column_pandas_labels: The pandas labels for the index columns, if any.
```
sfc-gh-mvashishtha:
```suggestion
        input_types: Snowpark column types of the input data columns (including index columns).
```

tests/integ/modin/frame/test_apply.py
```diff
@@ -148,6 +148,48 @@ def foo(row) -> str:
...
+    snow_df = snow_df.set_index("a")
+    native_df = native_df.set_index("a")
+
+    eval_snowpark_pandas_result(
+        snow_df, native_df, lambda x: x.apply(lambda row: row.name, axis=1)
+    )
+
+
+@sql_count_checker(query_count=5, join_count=2, udtf_count=1)
+def test_apply_axis_1_multiindex_preservation():
```
sfc-gh-mvashishtha, line 179: Could we also test 

1) `func` with return type annotations. We'll use vectorized UDFs instead of UDTFs.
2) `func` returning a series
3) apply() on series (with func typed, untyped, or returning a series)

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
sfc-gh-mvashishtha: I meant that you can replace most of the code here with set_index(). See https://github.com/snowflakedb/snowpark-python/pull/3979.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -9809,6 +9809,13 @@ def _apply_with_udtf_and_dynamic_pivot_along_axis_1(
         )
 
         # The apply function is encapsulated in a UDTF and run as a stored procedure on the pandas dataframe.
+        # Determine if we should pass index columns to the UDTF
+        # We pass index columns when the index is not the row position itself
```
sfc-gh-mvashishtha, line 9813: We always pass the index column names here. We can keep doing that, but we should update the comment and make the parameter required, since there don't seem to be any other invocations of that function.

src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py
```diff
@@ -424,6 +424,7 @@ def create_udtf_for_apply_axis_1(
     column_index: native_pd.Index,
     input_types: list[DataType],
     session: Session,
+    index_column_labels: list[Hashable] | None = None,
```
sfc-gh-mvashishtha, line 427: It turns out that just passing the number of index columns is enough: https://github.com/snowflakedb/snowpark-python/blob/849ba40d32e79b212fb08673df7bcfe8f25c2a26/src/snowflake/snowpark/modin/plugin/_internal/apply_utils.py#L471

PR #3963, opened 2025-10-28
SNOW-2442822: Fix np.asarray(datetime_with_timezone_index)

sfc-gh-mvashishtha:
The current implementation of `Index.__array__` passes `self.dtype` to `self.to_pandas().__array__()`, but we should instead propagate the `dtype` argument from the `__array__` call.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
+++ b/src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
@@ -962,7 +962,7 @@ def get_valid_col_pos_list_from_columns(
 
         # convert float like keys to integers
         elif not is_integer_dtype(pos_array.dtype):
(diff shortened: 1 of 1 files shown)


src/snowflake/snowpark/modin/plugin/_internal/indexing_utils.py
```diff
@@ -962,7 +962,7 @@ def get_valid_col_pos_list_from_columns(
 
         # convert float like keys to integers
         elif not is_integer_dtype(pos_array.dtype):
-            assert is_float_dtype(
+            assert pos_array.size == 0 or is_float_dtype(
```
sfc-gh-mvashishtha, line 965: Come iloc tests failed until I added this condition.

sfc-gh-helmeleegy (approved):

PR #3964, opened 2025-10-28
SNOW-2391351: Avoid joins for drop_duplicates when keep!=False in faster pandas

sfc-gh-helmeleegy:

   

   Fixes SNOW-2391351

   Avoid joins for drop_duplicates when keep!=False in faster pandas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,15 @@
   - `str.lstrip`  
   - `str.rstrip`  
   - `str.translate`  
+  - `dt.tz_localize`
+  - `dt.tz_convert`
+  - `dt.ceil`
+  - `dt.round`
+  - `dt.floor`
+  - `dt.normalize`
+  - `dt.month_name`
+  - `dt.day_name`
+  - `dt.strftime`  
 - Make faster pandas disabled by default (opt-in instead of opt-out).
 - Improve performance of `drop_duplicates` by avoiding joins when `keep!=False` in faster pandas.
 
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/modin/plugin/extensions/base_overrides.py
```diff
@@ -1523,8 +1523,12 @@ def drop_duplicates(
...
     else:
         df = self
-    duplicated = df.duplicated(keep=keep)
-    result = self[~duplicated]
+    if pd.session.dummy_row_pos_optimization_enabled and keep in ["first", "last"]:
+        result_qc = df._query_compiler.drop_duplicates()
+        result = self.__constructor__(query_compiler=result_qc)
+    else:
+        duplicated = df.duplicated(keep=keep)
+        result = self[~duplicated]
```
sfc-gh-joshi, line 1531: Is there a way to push this logic down to the QC instead? Checking a session flag at the frontend layer feels un-idiomatic.

sfc-gh-helmeleegy: I tried pushing down the logic into the QC. But the issue is that the operations in L1530 and L1531 are at the level of the frontend layer, where non-trivial logic is already happening outside of the QC. So simply using the corresponding operations at the QC level may miss important corner cases that are currently handled by the frontend.

So in theory, it's doable. But it doesn't seem to be that straightforward.

sfc-gh-joshi (approved):

PR #3973, opened 2025-10-29
SNOW-1524760: Fix Series.isin behavior

sfc-gh-joshi:

   

   Fixes SNOW-1524760

This PR fixes the behavior of `Series.isin(other_series)`, which ignores indices instead of joining on row/column labels. It also adds a fast path for `Series.isin(dataframe)`, which should always return false at every index.

Per @sfc-gh-mvashishtha's investigation in the linked ticket:

> It seems that pandas behavior is:
> - ignore row and column labels for Series.isin(series)
> - Series.isin(dataframe) always returns False, e.g. s = pandas.Series([1]); s.isin(s.to_frame())
> - DataFrame.isin(dataframe) joins on both row and column labels
> - DataFrame.isin(series) ignores column labels but not row labels, e.g. pandas.DataFrame({'A': [1, 2]}).isin(pandas.Series([1, 2], name='B', index=[0,1])) gives True values because even though the column name is different, the index matches, but pandas.DataFrame({'A': [1, 2]}).isin(pandas.Series([1, 2], name='B', index=[-1, -2])) gives False values.

### DIFF
--- a/tests/integ/modin/test_faster_pandas.py
+++ b/tests/integ/modin/test_faster_pandas.py
@@ -688,7 +688,7 @@ def test_isin_list(session):
         assert_frame_equal(snow_result, native_result, check_dtype=False)
 
 
-@sql_count_checker(query_count=3)
+@sql_count_checker(query_count=3, join_count=2)
 def test_isin_series(session):
     with session_parameter_override(
         session, "dummy_row_pos_optimization_enabled", True


tests/integ/modin/test_faster_pandas.py
```diff
@@ -688,7 +688,7 @@ def test_isin_list(session):
         assert_frame_equal(snow_result, native_result, check_dtype=False)
 
 
-@sql_count_checker(query_count=3)
+@sql_count_checker(query_count=3, join_count=2)
```
sfc-gh-jkew, line 691: I'm a little surprised this changed; because you didn't need to change the join counts on other isin tests.

sfc-gh-jkew (approved):

tests/integ/modin/test_faster_pandas.py
sfc-gh-joshi: It looks like the query generated in this test got significantly more complicated because of the added ARRAY_AGG operation, but I believe the previous version was incorrect for a lot of cases. The actual query text of the other isin tests have joins as well, but I guess they're not being parsed correctly by the SQL counter.

PR #3982, opened 2025-10-30
NO-SNOW: fix type check before AST in sort()

sfc-gh-mayliu:

   

   Fixes SNOW-NNNNNNN

Impacted AST test job: https://github.com/snowflakedb/snowpark-python/actions/runs/18924709047/job/54029335118?pr=3971
```
[gw1] linux -- Python 3.13.7 /home/runner/work/snowpark-python/snowpark-python/.tox/py313-notudfdoctest-ci/bin/python3
tests/integ/test_df_sort.py:71: in test_sort_invalid_inputs
    assert (
E   assert 'sort() only accepts str and Column objects, or a list containing str and Column objects' in '<ExceptionInfo TypeError("<class \'list\'> is not a valid type for Column or column name AST.") tblen=5>'
E    +  where '<ExceptionInfo TypeError("<class \'list\'> is not a valid type for Column or column name AST.") tblen=5>' = str(<ExceptionInfo TypeError("<class 'list'> is not a valid type for Column or column name AST.") tblen=5>)
```

Root cause is that type checking for `sort(...*cols)` occurs later than AST expr parsing and thus impacts AST before TypeError is thrown.

### DIFF
--- a/tests/integ/scala/test_dataframe_suite.py
+++ b/tests/integ/scala/test_dataframe_suite.py
@@ -1219,11 +1219,6 @@ def test_sort(session):
         for i in range(len(sorted_rows) - 1)
     ]
 
-    # Negative test: sort() needs at least one sort expression
-    with pytest.raises(ValueError) as ex_info:
-        df.sort([])
-    assert "sort() needs at least one sort expression" in ex_info.value.args[0]
-
 
 def test_select(session):
     df = session.create_dataframe([(1, "a", 10), (2, "b", 20), (3, "c", 30)]).to_df(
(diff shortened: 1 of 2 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -2183,6 +2183,13 @@ def sort(
         """
 
         is_order_by_all = not cols
+
+        # This code performs additional type checks, run first.
+        exprs = self._convert_cols_to_exprs("sort()", *cols)
+        if not is_order_by_all:
+            if not exprs:
+                raise ValueError("sort() needs at least one sort expression.")
```
sfc-gh-aling: do we need these three lines at all?
if is_order_by_all, it won't run
if not order_by_all, there must be exprs.

sfc-gh-mayliu: These lines were needed to raise an error when something like `df.sort([])` is called. But now it makes me think that the error message is contradictory to the default behavior of `sort()`, which runs `sort by all` if no cols are provided. What about changing to below?
```
        exprs = self._convert_cols_to_exprs("sort()", *cols)
        is_order_by_all = not cols or not exprs
```

PR #3984, opened 2025-11-03
SNOW-2643972: Add support for groupby properties (groupby.groups/indices) in faster pandas

sfc-gh-helmeleegy:

   

   Fixes SNOW-2643972

   Add support for groupby properties (groupby.groups/indices) in faster pandas.


src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -823,14 +823,17 @@ def __init__(self, frame: InternalFrame) -> None:
     storage_format = property(lambda self: "Snowflake")
 
     def _raise_not_implemented_error_for_timedelta(
-        self, frame: InternalFrame = None
+        self, frame: InternalFrame = None, stack_depth: int = 2
```
sfc-gh-joshi, line 826: How much effort would it be to directly pass a method name instead of doing stack frame inspection here? This feels very brittle.

sfc-gh-joshi (approved):
One minor comment, otherwise LGTM.

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-helmeleegy: Yes, that should be doable. It will require changes to all the call sites though. I think the original implementation used stack frame inspection to avoid passing the method name each time. But I agree that that would be less brittle. I can do it in a followup PR.

sfc-gh-lmukhopadhyay (approved):

PR #4021, opened 2025-12-04
SNOW-2866374: Enforce deterministic column ordering in infer_schema

sfc-gh-mayliu:

   

   Fixes SNOW-2866374

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -1302,13 +1302,11 @@ def infer_schema_statement(
     path: str,
     file_format_name: str,
     options: Optional[Dict[str, str]] = None,
-    ordered: bool = True,
 ) -> str:
     """
-    Note: If ordered is set to True (default), results are ordered by ORDER_ID and
-    COLUMN_NAME for deterministic column ordering. Set to False to disable ordering.
+    Note: Results are ordered by ORDER_ID and COLUMN_NAME for deterministic column ordering.
     """
-    base_query = (
+    return (
         SELECT
         + STAR
         + FROM
@@ -1335,12 +1333,9 @@ def infer_schema_statement(
         )
         + RIGHT_PARENTHESIS
         + RIGHT_PARENTHESIS
+        + ORDER_BY
+        + "ORDER_ID, COLUMN_NAME"
     )
-
-    if ordered:
-        base_query += ORDER_BY + "ORDER_ID, COLUMN_NAME"
-
-    return base_query
 
 
 def file_operation_statement(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 #### Improvements
 
+- Improved `DataFrameReader` to return columns in deterministic order when using `INFER_SCHEMA`.
+
 ### Snowpark pandas API Updates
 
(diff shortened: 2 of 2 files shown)


tests/integ/conftest.py
```diff
@@ -237,6 +237,7 @@ def connection(db_parameters, local_testing_mode):
             "user",
             "password",
             "private_key_file",
+            "private_key_file_pwd",
```
sfc-gh-joshi, line 240: What's this change for?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1299,9 +1299,16 @@ def create_file_format_statement(
 
 
 def infer_schema_statement(
-    path: str, file_format_name: str, options: Optional[Dict[str, str]] = None
+    path: str,
+    file_format_name: str,
+    options: Optional[Dict[str, str]] = None,
+    ordered: bool = True,
```
sfc-gh-joshi: Is this called with `ordered=False` anywhere?

sfc-gh-joshi (approved):
LGTM, just some minor questions.

tests/integ/conftest.py
sfc-gh-mayliu: It's a key-pair auth connection param for testing locally via `SFCTEST0` service accounts. Human accounts require MFA now to enter the rotating 6-digit code every 30 seconds, I found it too inconvenient so created a service role with key-pair auth. The change in this file merely allows us to locally run integ tests with [key-pair auth](https://docs.snowflake.com/en/user-guide/key-pair-auth#configuring-key-pair-authentication). Here's the link to [configure in Python client](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#using-key-pair-authentication-and-key-pair-rotation)

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-mayliu: Nope, it's currently not called anywhere. The reason for this additional param is to allow upper stream consumer to fall back to old behavior in case unexpected hiccup happens; but at the same time this change is low-risk, so I don't want to over-engineer by adding a client level feature flag

sfc-gh-aling: I understand the intention is to provide flexibility but looking at the existing callers, I don't think there is a way for users to pass the value through any public api.

based upon the test result, this seems to be a pretty safe change

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1299,9 +1299,16 @@ def create_file_format_statement(
...
 ) -> str:
-    return (
+    """
+    Note: If ordered is set to True (default), results are ordered by ORDER_ID and
+    COLUMN_NAME for deterministic column ordering. Set to False to disable ordering.
```
sfc-gh-aling: we can call this out as improvement in the changelog

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-mayliu: Makes sense, removed the `ordered` param

PR #4022, opened 2025-12-04
SNOW-2866776: add private var in numeric type

sfc-gh-yuwang:

   

   Fixes SNOW-2866776

https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkConnectRegressionRunner/220/
scos test


src/snowflake/snowpark/types.py
```diff
@@ -169,7 +169,9 @@ def _fill_ast(self, ast: proto.DataType) -> None:
 
 
 class _NumericType(_AtomicType):
-    pass
+    def __init__(self, **kwargs) -> None:
+        self._precision = kwargs.get("precision", None)
+        self._scale = kwargs.get("scale", None)
```
sfc-gh-joshi:
```suggestion
    def __init__(self, precision: int | None = None, scale: int | None = None) -> None:
        self._precision = precision
        self._scale = scale
```
Can we do this style instead to make it more explicit?

tests/integ/test_datatypes.py
```diff
@@ -408,3 +413,73 @@ def test_join_basic(session):
...
+
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="session.sql not supported by local testing mode",
+)
+def test_numeric_type_store_precision_and_scale(session):
+    table_name = Utils.random_table_name()
+    df = session.create_dataframe(
+        [Decimal("9" * 38)],
+        StructType([StructField("large_value", DecimalType(38, 0), True)]),
```
sfc-gh-joshi: Can you add test cases with other precision values?

tests/integ/test_datatypes.py
```diff
@@ -408,3 +413,73 @@ def test_join_basic(session):
...
+    )
+    df.write.save_as_table(table_name, mode="overwrite", table_type="temp")
```
sfc-gh-joshi: Should this table be cleaned up after the test?

tests/integ/test_datatypes.py
```diff
@@ -408,3 +413,73 @@ def test_join_basic(session):
...
+
+
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="relaxed_types not supported by local testing mode",
+)
+def test_numeric_type_store_precision_and_scale_read_file(session):
+    stage_name = Utils.random_stage_name()
+    header = ("BIG_NUM",)
+    test_data = [("9" * 38,)]
```
sfc-gh-joshi: As above--can we test more precisions?

sfc-gh-joshi (changes requested):

src/snowflake/snowpark/types.py
sfc-gh-yuwang: this is a internal only feature, I think we don't want to expose this information to customer?

tests/integ/test_datatypes.py
sfc-gh-yuwang: this is a temp table that should be auto cleaned, but i'll add some code to clean it

src/snowflake/snowpark/types.py
sfc-gh-joshi: OK, we can just keep it as kwargs then.

What do you think about moving the arguments onto `_IntegralType` instead of `_NumericType`? It looks like Snowflake only has a single precision for `FLOAT` and `DECIMAL` ([docs](https://docs.snowflake.com/en/sql-reference/data-types-numeric#data-types-for-floating-point-numbers)), so we should never have to expose precision information for those types.

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -307,7 +307,10 @@ def convert_sf_to_sp_type(
     if column_type_name == "REAL":
         return DoubleType()
     if (column_type_name == "FIXED" or column_type_name == "NUMBER") and scale == 0:
-        return LongType()
+        if context._store_precision_and_scale_in_numeric_type:
```
sfc-gh-aling: you can use the scos mode parameter
```suggestion
        if context._is_snowpark_connect_compatible_mode:
```

src/snowflake/snowpark/types.py
sfc-gh-aling: +1 on using _IntegralType, I don't think we need _scale as it would always be 0

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -133,7 +133,12 @@ def __init__(self) -> None:
         super().__init__()
 
     def __eq__(self, other: Any) -> bool:
-        return isinstance(other, self.__class__) and self.__dict__ == other.__dict__
+        def filtered(d: dict) -> dict:
+            return {k: v for k, v in d.items() if k not in ("_precision", "_scale")}
+
+        return isinstance(other, self.__class__) and filtered(
+            self.__dict__
+        ) == filtered(other.__dict__)
```
sfc-gh-joshi: Can we delete this method entirely? It looks identical to the version inherited from `_NumericType`.

sfc-gh-yuwang: removed

src/snowflake/snowpark/types.py
```diff
@@ -169,7 +169,17 @@ def _fill_ast(self, ast: proto.DataType) -> None:
...
+
+    def __eq__(self, other):
+        def filtered(d: dict) -> dict:
+            return {k: v for k, v in d.items() if k not in ("_precision", "_scale")}
+
+        return isinstance(other, self.__class__) and filtered(
+            self.__dict__
+        ) == filtered(other.__dict__)
```
sfc-gh-aling: why we overwrite the logic to exclude the new prop?

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -132,9 +132,6 @@ class TimedeltaType(SnowparkPandasType, LongType):
     def __init__(self) -> None:
         super().__init__()
 
-    def __eq__(self, other: Any) -> bool:
-        return isinstance(other, self.__class__) and self.__dict__ == other.__dict__
-
```
sfc-gh-aling, line 137: ideally we shall not update existing snowpark/pandas code.
I think this is mis-deleted? or is the new prop breaking anything?

sfc-gh-joshi: Yuyang discussed with me on slack. It looks like there are some tests that do equality checks between types, and this type inherits from `LongType`. Since `__dict__` includes the new `_precision` and `_scale` fields, these would cause errors if the fields are compared. The new superclass `__eq__` method now handles this filtering, and the implementation is otherwise equivalent.

src/snowflake/snowpark/types.py
sfc-gh-yuwang: here is a example test:
    expected_fields = [
        StructField("COL1", LongType(), nullable=True),
        StructField("COL2", LongType(), nullable=True),
    ]
    assert df.schema.fields == expected_fields
 we exclude the private var so that these existing test does not fail

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
sfc-gh-yuwang: This pandas type is inherenting from LongType(), which have the private var, so it should also exclude private vars we added during comparing.
@sfc-gh-joshi can you help confirm if this change is valid?

sfc-gh-joshi: I haven't worked with the TimedeltaType much before, but as far as I can tell the change makes sense.

src/snowflake/snowpark/types.py
sfc-gh-yuwang: i see, will fix this

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
```diff
@@ -121,7 +121,7 @@ class TimedeltaType(SnowparkPandasType, LongType):
     two times.
     """
 
-    snowpark_type: DataType = LongType()
+    snowpark_type: DataType = field(default_factory=LongType)
```
sfc-gh-joshi: Why this change? It looks like this variable isn't used anywhere, but is changing it necessary?

src/snowflake/snowpark/types.py
```diff
@@ -370,7 +370,19 @@ def _fill_ast(self, ast: proto.DataType) -> None:
...
+    def __eq__(self, other):
+        def filtered(d: dict) -> dict:
+            return {k: v for k, v in d.items() if k != "_precision"}
+
+        return isinstance(other, self.__class__) and filtered(
+            self.__dict__
+        ) == filtered(other.__dict__)
+
+    def __hash__(self):
+        return hash(repr(self))
```
sfc-gh-joshi, line 393: Why do we need a custom hash implementation for this class?

tests/integ/test_datatypes.py
```diff
@@ -408,3 +413,80 @@ def test_join_basic(session):
...
+    try:
+        df = session.create_dataframe(
+            [Decimal(massive_number)],
+            StructType([StructField("large_value", DecimalType(precision, 0), True)]),
+        )
+        df.write.save_as_table(table_name, mode="overwrite", table_type="temp")
+        result = session.sql(f"select * from {table_name}")
+        datatype = result.schema.fields[0].datatype
+        assert isinstance(datatype, LongType)
+        assert datatype._precision == 38
```
sfc-gh-joshi, line 443: Don't we expect the precision to change based on the number of digits in the input number?

src/snowflake/snowpark/types.py
```diff
@@ -370,7 +370,19 @@ def _fill_ast(self, ast: proto.DataType) -> None:
 
 # Numeric types
 class _IntegralType(_NumericType):
-    pass
+    def __init__(self, **kwargs) -> None:
+        self._precision = kwargs.get("precision", None)
```
sfc-gh-aling: let's add an underscore to precision marking it as private, and keep raising TypeError for mis-passed parameters

```suggestion
    def __init__(self, **kwargs) -> None:
        self._precision = kwargs.pop("_precision", None)
        if kwargs:  # for customers, they are not supposed to pass kwargs when constructing the type
            raise TypeError('...')
        
```

src/snowflake/snowpark/types.py
sfc-gh-yuwang: TL:DR: this is to fix some test failure
when you defined __eq__, you need to redefine __hash__ otherwise it is considered None

src/snowflake/snowpark/modin/plugin/_internal/snowpark_pandas_types.py
sfc-gh-yuwang: This is to pass the test,
otherwise it will complain:
E   ValueError: mutable default <class 'snowflake.snowpark.types.LongType'> for field snowpark_type is not allowed: use default_factory
what does this error mean
cause we changed the definition of LongType that it is mutable now

src/snowflake/snowpark/types.py
```diff
@@ -370,7 +370,19 @@ def _fill_ast(self, ast: proto.DataType) -> None:
...
+
+    def __eq__(self, other):
+        def filtered(d: dict) -> dict:
+            return {k: v for k, v in d.items() if k != "_precision"}
+
+        return isinstance(other, self.__class__) and filtered(
+            self.__dict__
+        ) == filtered(other.__dict__)
```
sfc-gh-aling: I think whether excluding precision or not should be based upon the context._is_snowpark_connect_mode.

scos they might want to distinguish LongType by precision

tests/integ/test_datatypes.py
sfc-gh-yuwang: I updated the test,
the behavior I observed is that after the df being save as table, the schema it being set to Number(38, 0) some how, so it is always 38 when read it from table.
I don't think this is the correct behavior, but should be fixed in another PR

src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -307,7 +307,7 @@ def convert_sf_to_sp_type(
     if column_type_name == "REAL":
         return DoubleType()
     if (column_type_name == "FIXED" or column_type_name == "NUMBER") and scale == 0:
-        return LongType()
+        return LongType(precision=precision)
```
sfc-gh-aling: if we make the argument private, it will be:
```suggestion
        return LongType(_precision=precision)
```

sfc-gh-joshi (approved):

sfc-gh-yuwang:
I still kept the __eq__ function in TimedeltaType class.
Otherwise it will treat LongType() == TimedeltaType() which is wrong.

PR #4024, opened 2025-12-08
SNOW-2887937: fix dataframewriter loses decimal precision info when writing back to snowflake table

sfc-gh-yuwang:

   

   Fixes SNOW-2887937

### DIFF
--- a/tests/integ/test_datatypes.py
+++ b/tests/integ/test_datatypes.py
@@ -554,7 +554,6 @@ def test_integral_type_default_precision(mock_default_precision):
     "config.getoption('local_testing_mode', default=False)",
     reason="session.sql not supported by local testing mode",
 )
-@pytest.mark.parametrize("precision", [38, 19, 5, 3])
 @pytest.mark.parametrize(
     "mock_default_precision",
     [
@@ -562,7 +561,7 @@ def test_integral_type_default_precision(mock_default_precision):
         {LongType: 19, IntegerType: 10},
     ],
 )
-def test_end_to_end_default_precision(session, precision, mock_default_precision):
+def test_end_to_end_default_precision(session, mock_default_precision):
     table_name = Utils.random_table_name()
 
     with mock.patch.object(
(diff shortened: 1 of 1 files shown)


tests/integ/test_datatypes.py
```diff
@@ -437,7 +440,6 @@ def test_numeric_type_store_precision_and_scale(session, massive_number, precisi
         # does not have precision information, thus set to default 38.
         df.write.save_as_table(table_name, mode="overwrite", table_type="temp")
         result = session.sql(f"select * from {table_name}")
-        session.sql(f"describe table {table_name}").show()
         datatype = result.schema.fields[0].datatype
         assert isinstance(datatype, LongType)
         assert datatype._precision == 38
```
sfc-gh-joshi, line 445: Shouldn't this now be the precision in the test parameter instead of the default 38?

sfc-gh-yuwang: this is the test of previous change, which does not turned on the default precision setting, so it is expected to be 38 in this test.
default precision is covered in the added test

tests/integ/test_datatypes.py
```diff
@@ -502,3 +504,162 @@ def write_csv(data):
 def test_illegal_argument_intergraltype():
     with pytest.raises(TypeError, match="takes 0 argument but 1 were given"):
         LongType(b=10)
+
+
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="session.sql not supported by local testing mode",
+)
+@pytest.mark.parametrize("precision", [38, 19, 5, 3])
```
sfc-gh-aling, line 513: super nit: I think you can just create a dataframe with 4 columns of different precision Decimal instead of parametrizing the tests, and verify the 4 columns

this would help save some test time and resources

tests/integ/test_datatypes.py
```diff
@@ -502,3 +504,162 @@ def write_csv(data):
...
+        )
+
+
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="relaxed_types not supported by local testing mode",
+)
+@pytest.mark.parametrize("massive_number", ["9" * 38, "5" * 19, "7" * 5])
+def test_default_precision_read_file(session, massive_number):
+    mock_default_precision = {"LongType": 19, "IntegerType": 10}
```
sfc-gh-aling: I'm wondering if we need set mock_default_precision here, looks like the test case won't hit the code path of looking up mock_default_precision at all.

I ask because I see the key type is incorrect while it doesn't impact the execution of the test

sfc-gh-joshi (approved):

PR #4031, opened 2025-12-10
SNOW-2872192: Support targeted delete-insert in save_as_table

sfc-gh-mayliu:

   

   Fixes SNOW-2872192

Added override_condition parameter to DataFrameWriter.save_as_table() that enables atomic targeted delete-insert operations when used with `mode="append"`. Delete and insert is wrapped in transaction to ensure atomicity and protect tables from entering a bad state.

This performs a similar operation to PySpark's [DataFrameWriterV2.overwrite(condition)](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriterV2.overwrite.html#pyspark.sql.DataFrameWriterV2.overwrite), where rows matching the condition are deleted from the target table before inserting all rows from the DataFrame.

For more details on this PR, refer to this [JIRA](https://snowflakecomputing.atlassian.net/browse/SNOW-2872192) that contains customer's code snippet.

Monorepo for AST: https://github.com/snowflake-eng/snowflake/pull/368680

### DIFF
--- a/src/snowflake/snowpark/dataframe_writer.py
+++ b/src/snowflake/snowpark/dataframe_writer.py
@@ -271,7 +271,9 @@ class DataFrameWriter:
 
                 "append": Append data of this DataFrame to the existing table. Creates a table if it does not exist.
 
-                "overwrite": Overwrite the existing table by dropping old table.
+                "overwrite": Overwrite the existing table. By default, drops and recreates the table.
+                    When ``overwrite_condition`` is specified, performs selective overwrite: deletes only
+                    rows matching the condition, then inserts new data.
 
                 "truncate": Overwrite the existing table by truncating old table.
 
@@ -333,8 +335,10 @@ class DataFrameWriter:
                 Set to ``True`` if table exists, ``False`` if it doesn't, or ``None`` (default) for automatic detection.
                 Primarily useful for "append", "truncate", and "overwrite" with overwrite_condition modes to avoid running query for automatic detection.
             overwrite_condition: Specifies the overwrite condition to perform atomic targeted delete-insert.
-                Can only be used when ``mode`` is "overwrite" and the table exists. Rows matching the
-                condition are deleted from the target table, then all rows from the DataFrame are inserted.
+                Can only be used when ``mode`` is "overwrite". When provided and the table exists, rows matching
+                the condition are atomically deleted and all rows from the DataFrame are inserted, preserving
+                non-matching rows. When not provided, the default "overwrite" behavior applies (drop and recreate table).
+                If the table does not exist, ``overwrite_condition`` is ignored and the table is created normally.
 
 
         Example 1::
--- a/tests/integ/test_dataframe.py
+++ b/tests/integ/test_dataframe.py
@@ -4568,7 +4568,7 @@ def test_write_table_with_clustering_keys_and_comment(
 
 @pytest.mark.xfail(
     "config.getoption('local_testing_mode', default=False)",
-    reason="overwrite_condition is a SQL feature",
+    reason="overwrite_condition is not supported in local testing mode",
     run=False,
 )
 def test_write_table_with_overwrite_condition(session):
@@ -4680,7 +4680,7 @@ def test_write_table_with_overwrite_condition(session):
 
 @pytest.mark.xfail(
     "config.getoption('local_testing_mode', default=False)",
-    reason="overwrite_condition is a SQL feature",
+    reason="overwrite_condition is a not supported in local testing mode",
     run=False,
 )
 @pytest.mark.parametrize(


src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -331,6 +332,9 @@ def save_as_table(
             table_exists: Optional parameter to specify if the table is known to exist or not.
                 Set to ``True`` if table exists, ``False`` if it doesn't, or ``None`` (default) for automatic detection.
                 Primarily useful for "append" and "truncate" modes to avoid running query for automatic detection.
+            override_condition: Specifies the override condition to perform atomic targeted delete-insert.
+                Can only be used when ``mode`` is "append" and the table exists. Rows matching the
```
sfc-gh-aling: my intuition is this shall be put under "overwrite" mode -- it's an advanced feature for selective "overwrite".

or maybe it can work for both `append` and `overwrite`

src/snowflake/snowpark/dataframe_writer.py
```diff
@@ -256,6 +256,7 @@ def save_as_table(
             Dict[str, Union[str, Iterable[ColumnOrSqlExpr]]]
         ] = None,
         table_exists: Optional[bool] = None,
+        override_condition: Optional[ColumnOrSqlExpr] = None,
```
sfc-gh-aling: I personally find the name `override` a bit confusing -- to me it is more of a OOP concept where we replace the old behavior with a new one, but here we actually overwrite data.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1417,19 +1418,65 @@ def get_create_and_insert_plan(child: SnowflakePlan, replace, error):
...
+                    Query(
+                        insert_into_statement(
+                            table_name=full_table_name,
+                            child=child.queries[-1].sql,
+                            column_names=column_names,
+                        ),
+                        params=child.queries[-1].params,
+                        is_ddl_on_temp_object=is_temp_table_type,
+                    ),
+                    Query("COMMIT"),
```
sfc-gh-aling, line 1451: I'm curious is it possible to achieve the same effect with one `merge into` query?

I asked for the following reasons:
- no need to deal with transaction
- will merge into give faster performance?

sfc-gh-mayliu: The `merge into` query would worsen performance due to a few reasons:
1. `merge into` does row-level comparison, and it's join-based
2. if the table does not have a perfect primary key, `merge into` risks a cartesian product which can significantly worsen performance

In comparison, the performance impact of the transaction wrapper is negligible, and `delete + insert into` are lighter operations.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1417,19 +1418,65 @@ def get_create_and_insert_plan(child: SnowflakePlan, replace, error):
...
+                    # Normal append without overwrite_condition
+                    return self.build(
+                        lambda x: insert_into_statement(
+                            table_name=full_table_name,
+                            child=x,
+                            column_names=column_names,
+                        ),
+                        child,
+                        source_plan,
+                    )
```
sfc-gh-aling: after giving it more thought, I still feel that conditional overwrite doesn’t conceptually align with append mode..
but I can live with it, not a blocker to this feature support

@sfc-gh-yixie wdyt? shall we make this new option only work for `overwrite` mode.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1417,19 +1418,65 @@ def get_create_and_insert_plan(child: SnowflakePlan, replace, error):
...
+
+            Deletes rows matching the overwrite_condition condition, then inserts
+            all rows from the source DataFrame. Wrapped in a transaction for atomicity.
+            """
+            child = self.add_result_scan_if_not_select(child)
+
+            return SnowflakePlan(
+                [
+                    *child.queries[:-1],
+                    Query("BEGIN TRANSACTION"),
```
sfc-gh-aling, line 1432: do we need explicit transaction control here?
this is a new pattern we haven't done before.

sfc-gh-joshi (commented):
I have two general questions about this PR:

1. In slack you mentioned this behavior is similar to Spark's `DataFrameWriterV2.overwrite` method. Why not implement that instead of adding a flag to `saveAsTable`? Doing so would keep code simpler, and avoid the potential semantic mismatch between the `mode` flag and `overwrite_condition` that @sfc-gh-aling mentioned.
2. What happens if the schema of the new dataframe differs from that of the original table? My guess from looking at the code is that the INSERT query would fail, which I assume would fail the whole transaction, but if `mode="overwrite"` we would expect to drop the original table and the operation should succeed, as per [our documentation](https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/latest/snowpark/api/snowflake.snowpark.DataFrameWriter.saveAsTable):
> ”overwrite”: Overwrite the existing table by dropping old table.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-mayliu: Reverted support for `append` mode, now targeted delete-insert only works with `overwrite`

sfc-gh-joshi (approved):

PR #4033, opened 2025-12-11
SNOW-2408964: support custom_schema when reading xml

sfc-gh-yuwang:

   

   Fixes SNOW-2408964

   Here is a doc to talk about the behavior of XML read with user schema in snowpark:
   https://docs.google.com/document/d/1W3os2SGbO_M89TjIa7xVuNrQ0RDDuGMHEOBHFuw0OYw/edit?tab=t.0

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -2004,7 +2004,7 @@ def attribute_to_schema_string(attributes: List[Attribute]) -> str:
 def attribute_to_schema_string_deep(attributes) -> str:
     from snowflake.snowpark.types import ArrayType, MapType, StructType
 
-    def dt_to_type_string(dt: DataType) -> str:
+    def dt_to_type_string(dt) -> str:
         if isinstance(dt, ArrayType):
             return f"array<{dt_to_type_string(dt.element_type)}>"
 


src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1371,8 +1382,14 @@ def _read_semi_structured_file(self, path: str, format: str) -> DataFrame:
...
 
-        if self._user_schema and format.lower() != "json":
+        if self._user_schema and format.lower() not in ["json", "xml"]:
             raise ValueError(f"Read {format} does not support user schema")
+        if (
+            self._user_schema
+            and format.lower() == "xml"
+            and XML_ROW_TAG_STRING not in self._cur_options
+        ):
+            raise ValueError("When read XML with user schema, rowtag must be set.")
```
sfc-gh-aalam:
```suggestion
            raise ValueError("When reading XML with user schema, rowtag must be set.")
```

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -52,6 +57,36 @@ def replace_entity(match: re.Match) -> str:
...
+    if isinstance(dt, ArrayType) and dt.element_type is not None:
+        return struct_type_to_result_template(dt.element_type)
+
+    if isinstance(dt, MapType) and dt.value_type is not None:
+        return struct_type_to_result_template(dt.value_type)
+
+    return None
+
+
+def generate_norm_column_name_to_ori_column_name_dict(result: dict):
```
sfc-gh-aalam: does this really need to be a function?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1069,7 +1070,17 @@ def xml(self, path: str, _emit_ast: bool = True) -> DataFrame:
             ast.reader.CopyFrom(self._ast)
             df._ast_id = stmt.uid
 
-        return df
+        # cast to input custom schema type
+        if self._user_schema:
+            cols = [
+                df[single_quote(field._name)]
```
sfc-gh-aalam, line 1077: why are we using single_quote here?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1069,7 +1070,17 @@ def xml(self, path: str, _emit_ast: bool = True) -> DataFrame:
...
+                .cast(field.datatype)
+                .alias(quote_name_without_upper_casing(field._name))
```
sfc-gh-aling, line 1079: does our JSON reader preserve case sensitivity for those column identifiers?
I am asking for the sake of reader behavior consistency

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-yuwang: you are right, I removed this function

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: currently all column name is produced with a single quote around, 
I am not sure why we are having this behavior but I think we should not change now as it would be a BCR
single quote is added here so that we can correctly read the column

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -52,6 +57,36 @@ def replace_entity(match: re.Match) -> str:
         return match.group(0)
 
 
+def schema_string_to_result_dict_and_struct_type(schema_string: str) -> Optional[dict]:
+    if schema_string == "":
+        return None
+    schema = type_string_to_type_object(schema_string)
+    if not isinstance(schema, StructType):
+        return None
```
sfc-gh-aling: two questions:
1. what's the case of the returned schema not being a StructType / what is the type returned?
2. should we error out in this case if non-StructType is returned?

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: yes it does,
this is also aligned with the behavior of pyspark json/xml reader

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-yuwang: 1. this case actually would not happened, it return a non-structtype only when input is not a valid schema string. While the upper stream is already making sure it would return a valid schema string
2. technically we could remove this line of code as it is a bit redundant

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aalam: Ahh, I think I remember Jianzhun mentioning we use PIVOT and it snowflake pivot adds single quote for string, and datetime columns. This could break since server side is planning to do a BCR to remove quotes.

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -270,10 +299,16 @@ def element_to_dict_or_str(
     value_tag: str = "_VALUE",
     null_value: str = "",
     ignore_surrounding_whitespace: bool = False,
+    result_template: Optional[dict] = None,
 ) -> Optional[Union[Dict[str, Any], str]]:
     """
     Recursively converts an XML Element to a dictionary.
     """
+    norm_name_to_ori_name = (
+        {key.lower(): key for key in result_template.keys()}
```
sfc-gh-aling, line 308: is this just process one layer?\
I'm wondering if we need to call `key.lower()` recursive to the nested dict.

or can we chain the lower() call in the `struct_type_to_result_template` directly?

tests/integ/test_xml_reader_row_tag.py
```diff
@@ -467,3 +474,154 @@ def test_read_xml_row_validation_xsd_path_failfast(session):
         session.read.option("rowTag", row_tag).option(
             "rowValidationXSDPath", f"@{tmp_stage_name}/{test_file_books_xsd}"
         ).option("mode", "failfast").xml(f"@{tmp_stage_name}/{test_file_books_xml}")
+
+
+def test_read_xml_with_custom_schema(session):
```
sfc-gh-aling, line 480: can we have one simple case for nested data? books2.xml seems to be a nested one

PR #4037, opened 2025-12-16
SNOW-2895675: Skip aliases when source/destination column are identical

sfc-gh-joshi:

   

   Fixes SNOW-2895675

When an alias clause would emit an alias that does not change an input column's name, as in `SELECT "A" AS "A"`, the alias is elided to just `SELECT "A"`. This PR also includes a fix from @sfc-gh-aling to emit `SELECT *` (rather than `SELECT "col1", "col2"`) in join operations if no aliasing is necessary.
As a result, the SQL for the final query emitted by `test_dataframe_join_suite.py::test_name_alias_on_multiple_join` is reduced from 1045B -> 799B, a 24% reduction. This fix was requested for SCOS; a sample query emulating a user workload similarly experiences a 15% reduction in query size in SCOS (2175B -> 1857B), and a 60% reduction for the Snowpark equivalent (328B -> 128B).

## Implementation Details

Though the original ask was specifically to implement this change for JOIN operations, this PR applies the optimization to all generated queries. It does so with changes in three locations:
1. `unary_expression_extractor` in `analyzer.py`: Avoids emitting SQL for an alias in locations when possible, when traversing a query plan. This is the simplest location to make this change, as it avoids the need to track down all call sites that generate an `Alias` node.
2. `derive_column_states_from_subquery` in `select_statement.py`: This method compares the analyzed query strings of expressions to see if their values have changed. Previously, aliases always fully resolved (`"A" AS "A"`), so aliased columns were always assigned the CHANGED_EXP state; now, since redundant aliases resolve to just the column name (`"A"`), this method assigns UNCHANGED_EXP instead. My understanding is that this behavior should be correct, but this produced bugs in nested joins where a top-level projection did not properly use an alias for an ambiguous column.
3. `_disambiguate` in `dataframe.py`: If two frames have no column names in common at all, then there is no need for disambiguation, and we can simply emit a `SELECT *`. This fix was proposed by @sfc-gh-aling in #4044.

I could not track down the root cause of the aliasing issue, as it appeared even with simpler fixes like modifying [`_alias_if_needed`](https://github.com/snowflakedb/snowpark-python/blob/266334bdcf49e701f090c8e6cf517297c20f3cfb/src/snowflake/snowpark/dataframe.py#L274) in `dataframe.py`, and replacing analyzer calls with `parse_local_name` as suggested by [comments](https://github.com/snowflakedb/snowpark-python/blob/266334bdcf49e701f090c8e6cf517297c20f3cfb/src/snowflake/snowpark/_internal/analyzer/select_statement.py#L2220) within `select_statement.py`. Based on running through the codebase (and asking Cursor), it is likely that a subquery alias mapping somewhere is not being populated somewhere during analysis, but it is not clear what step of the analysis process would be responsible for this.

The fix I chose provides the largest benefit (removing redundant aliasing for all queries, not just joins) while adhering as closely as possible to previous behavior when analyzing column change states.

## Why do we need to change both the analyzer and join disambiguation?

In cases where frames have no overlapping column names, the fix in #4044 to emit `SELECT *` instead of individual column names is sufficient. This can also lead to a reduction in issued DESCRIBE queries. Example:
```python
def without_common_columns():
    session.create_dataframe([(1,2,3,4)], schema=["a", "b", "c", "d"]).write.save_as_table(table_name="temptable1", table_type="temporary")
    session.create_dataframe([(5,6,7,8)], schema=["e", "f", "g", "h"]).write.save_as_table(table_name="temptable2", table_type="temporary")

    df1 = session.table("temptable1")
    df2 = session.table("temptable2")
    dfm = df1.join(df2)
    dfm.show()
```

However, if frames do have overlapping column names, explicit removal of aliasing is necessary to reduce query text size. This is a very realistic scenario, as JOIN operations frequently combine tables based on a common ID column or other key. Example:
```
def with_common_columns():
    df1 = session.create_dataframe([[0, 1, 3, 4, 5]], schema=["b", "a", "c", "d", "x"])
    df2 = session.create_dataframe([[0, 2, 3, 4, 5]], schema=["b", "a", "c", "e", "y"])
    df3 = session.create_dataframe([[0, 3, 3, 4, 5]], schema=["b", "a", "c", "f", "z"])
    result = (
        df1
            .join(df2, df1["b"] == df2["b"])
            .join(df3, df1["c"] == df3["c"])
            .filter(df1["a"] > 0)
            .select(df1["a"], df2["a"], df1["b"])
    )
    result.show()
```
The changes in this PR do not reduce the number of queries issued for this case, but can lead to a substantial reduction in text size when a joined DF is used repeatedly in a sub-query.

### DIFF
--- a/tests/integ/scala/test_dataframe_writer_suite.py
+++ b/tests/integ/scala/test_dataframe_writer_suite.py
@@ -258,7 +258,9 @@ def test_iceberg(session, local_testing_mode):
             ddl[0][0] == f"create or replace ICEBERG TABLE {table_name} (\n\t"
             f"A STRING,\n\tB LONG,\n\tTS TIMESTAMP_NTZ(9)\n)\n "
             f"PARTITION BY (A, BUCKET(5, B), TRUNCATE(3, A), DAY(TS))\n "
-            f"EXTERNAL_VOLUME = 'PYTHON_CONNECTOR_ICEBERG_EXVOL'\n CATALOG = 'SNOWFLAKE'\n "
+            f"EXTERNAL_VOLUME = 'PYTHON_CONNECTOR_ICEBERG_EXVOL'\n "
+            f"ICEBERG_VERSION = 3\n "
+            f"CATALOG = 'SNOWFLAKE'\n "
             f"BASE_LOCATION = 'snowpark_python_tests/';"
         )
 


src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -763,10 +763,18 @@ def unary_expression_extractor(
                     for k, v in df_alias_dict.items():
                         if v == expr.child.name:
                             df_alias_dict[k] = updated_due_to_inheritance  # type: ignore
+            origin = self.analyze(
+                expr.child, df_aliased_col_name_to_real_col_name, parse_local_name
+            )
+            if (
+                isinstance(expr.child, (Attribute, UnresolvedAttribute))
+                and origin == quoted_name
```
sfc-gh-aalam, line 771: why not move this check here?
https://github.com/snowflakedb/snowpark-python/blob/266334bdcf49e701f090c8e6cf517297c20f3cfb/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py#L364-L365

sfc-gh-joshi: I felt it would be clearer to put this check outside the call to a direct SQL generation function. IMO it would be very unintuitive if a call to `alias_expression` had control logic that could emit SQL that did not represent an actual alias expression.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
```diff
@@ -763,10 +763,18 @@ def unary_expression_extractor(
                     for k, v in df_alias_dict.items():
                         if v == expr.child.name:
                             df_alias_dict[k] = updated_due_to_inheritance  # type: ignore
+            origin = self.analyze(
+                expr.child, df_aliased_col_name_to_real_col_name, parse_local_name
+            )
+            if (
+                isinstance(expr.child, (Attribute, UnresolvedAttribute))
```
sfc-gh-aling, line 770: we are comparing origin against quoted_name, from analyzer code: https://github.com/snowflakedb/snowpark-python/blob/729fd3ecfa503a0b548eb6c1b2a09a5b0b771718/src/snowflake/snowpark/_internal/analyzer/analyzer.py#L416

I don't see `UnresolvedAttribute` is guaranteed to have quoted name

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -2244,11 +2244,14 @@ def derive_column_states_from_subquery(
...
+        result_name = analyzer.analyze(
+            c, from_.df_aliased_col_name_to_real_col_name, parse_local_name=True
+        ).strip(" ")
         if from_c_state and from_c_state.change_state != ColumnChangeState.DROPPED:
             # review later. should use parse_column_name
-            if c_name != analyzer.analyze(
-                c, from_.df_aliased_col_name_to_real_col_name, parse_local_name=True
-            ).strip(" "):
+            # SNOW-2895675: Always treat Aliases as "changed", even if it is an identity.
+            # The fact this check is needed may be a bug in column state analysis, and we should revisit it later.
```
sfc-gh-aling, line 2253: do you have a case that would fail if we don't treat aliases as changed?
and what's the generated sql

src/snowflake/snowpark/mock/_analyzer.py
```diff
@@ -654,12 +654,18 @@ def unary_expression_extractor(
...
-            alias_exp = alias_expression(
-                self.analyze(
-                    expr.child, df_aliased_col_name_to_real_col_name, parse_local_name
-                ),
-                quoted_name,
+            origin = self.analyze(
+                expr.child, df_aliased_col_name_to_real_col_name, parse_local_name
             )
+            if (
+                isinstance(expr.child, (Attribute, UnresolvedAttribute))
```
sfc-gh-aling, line 661: same question on quotes

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-joshi: My understanding is that all `UnresolvedAttribute` constructor calls are expected to quote the name of the attribute. You can verify this is the case with `rg 'UnresolvedAttribute\(' -A5 src`; every instance of the constructor call comes with quoting.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-joshi: `tests/integ/scala/test_dataframe_join_suite.py::test_name_alias_on_multiple_join` was previously failing without this, but it looks like it no longer fails after your disambiguation change. I think the underlying bug still exists though, so I'll see if I can find another reproducer.

sfc-gh-joshi: It looks like this is no longer necessary because of your changes to populate the alias map in `dataframe.py`. I've reverted this file.

sfc-gh-joshi: Never mind, CI revealed some failures:
```
FAILED tests/integ/test_cte.py::test_sql_simplifier - assert 2 == 1
FAILED tests/integ/scala/test_dataframe_suite.py::test_rename_join_dataframe - snowflake.snowpark.exceptions.SnowparkSQLException: (1304): 01c22c14-0813-e...
FAILED tests/integ/test_dataframe.py::test_dataframe_alias - snowflake.snowpark.exceptions.SnowparkSQLAmbiguousJoinException: (1303): Th...
```
I'll add a comment mentioning them.

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-aling: I find there's one place that doesn't use quoted name https://github.com/snowflakedb/snowpark-python/blob/d2e7450a2d02d5dc8976cb88278e43ed3f54bae2/src/snowflake/snowpark/column.py#L1494

src/snowflake/snowpark/_internal/analyzer/select_statement.py
```diff
@@ -2244,11 +2244,18 @@ def derive_column_states_from_subquery(
...
+            # The following tests fail without this check:
+            # - tests/integ/test_cte.py::test_sql_simplifier
+            # - tests/integ/scala/test_dataframe_suite.py::test_rename_join_dataframe
+            # - tests/integ/test_dataframe.py::test_dataframe_alias
+            if c_name != result_name or isinstance(c, Alias):
```
sfc-gh-aling, line 2258: should we be more tight on the condition here giving the improvement we made in the analyzer? or it's overkilling

```suggestion
            if c_name != result_name or (isinstance(c, Alias) and isinstance(c.child, (Attribute, UnresolvedAttribute)):
```

src/snowflake/snowpark/_internal/analyzer/analyzer.py
sfc-gh-joshi: It looks like this place is used for constructing raw SQL expressions (hence the `is_sql_text=True` flag), so it doesn't matter for the comparison against a column name here.

Either way, the worst case failure mode here is just a missing optimization, where we return an alias in scenarios where we would be able to return just an identifier.

src/snowflake/snowpark/_internal/analyzer/select_statement.py
sfc-gh-joshi: It shouldn't matter, since if the child expression isn't a simple Attribute/UnresolvedAttribute, then `result_name` will not be an identifier.

For example, in the following code:
```python
df = session.createDataFrame([[1, 2]], ["A", "B"])
df.select((col("A") * 2).as_("A"))
```

The analyzer will have `c_name = '"A"'` and `result_name = '("A" * 2) AS "A"'` when examining the alias.

sfc-gh-aalam (approved):

PR #4051, opened 2026-01-16
SNOW-2341888: preserve parameter names of user created functions (sprocs, UDFs, UDTFs, UDAFs)

sfc-gh-bkogan:

   

   Fixes SNOW-2341888

   This adds a `preserve_parameter_names` flag to the creation of stored procedures, UDFs, UDTFs, and UDAFs, defaults to False. When enabled, it will inspect the callable/file and extract the parameter names defined by the user, instead of arg1, arg2, etc. This enables users to invoke functions with named parameters with more readability, e.g. `CALL MY_SPROC(a => 3, b => 4)`.

### DIFF
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -559,6 +559,8 @@ def get_func_arg_names(
             arg_names = arg_names[1:]
 
         if len(arg_names) != num_args:
+            # This could happen for vectorized UDxFs, since there could be a single dataframe argument
+            # but potentially more than one column. We will do best-effort preservation but fallback if needed.
             return default_arg_names
 
         return arg_names


src/snowflake/snowpark/functions.py
```diff
@@ -9467,6 +9468,8 @@ def udf(
         resource_constraint: A dictionary containing a resource properties of a warehouse and then
             constraints needed to run this function. Eg ``{"architecture": "x86"}`` requires an x86
             warehouse be used for execution.
+        preserve_parameter_names: Whether to preserve the parameter names of the referenced function in the created UDF.
```
sfc-gh-aalam, line 9471: out of curiosity have you looked into why making this behavior `True` by default might be a breaking feature?

sfc-gh-bkogan: My understanding is that the call syntax for sproc and UDF supports named arguments, e.g. sproc(x => 2, y => 3). If users are creating UDFs/sprocs with Snowpark, and then have scripts that invoke them via named arguments, this would be breaking. I am not sure how common that is though.

@sfc-gh-yixie recommended guarding this as well, he may have another reason

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -493,6 +494,79 @@ def get_opt_arg_defaults_from_file():
...
+
+        arg_names = default_arg_names
+        if object_type == TempObjectType.TABLE_FUNCTION:
+            arg_names = retrieve_func_arg_names_from_source(
+                filename, TABLE_FUNCTION_PROCESS_METHOD, func_name
+            )
+        elif object_type == TempObjectType.AGGREGATE_FUNCTION:
+            arg_names = retrieve_func_arg_names_from_source(
+                filename, AGGREGATE_FUNCTION_ACCULUMATE_METHOD, func_name
+            )
```
sfc-gh-aling, line 535: just to clarify: for table and aggregation function, `func_name` is actually the class name?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -493,6 +494,79 @@ def get_opt_arg_defaults_from_file():
...
+        # 2. It's a table/aggregate function, ignore the "self" argument of the method
+        if object_type in (
+            TempObjectType.PROCEDURE,
+            TempObjectType.TABLE_FUNCTION,
+            TempObjectType.AGGREGATE_FUNCTION,
+        ):
+            arg_names = arg_names[1:]
+
+        if len(arg_names) != num_args:
+            return default_arg_names
```
sfc-gh-aling, line 564: do you know when arg_names != num_args would happend?
also should we raise error here or at least raise a warning?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -493,6 +494,79 @@ def get_opt_arg_defaults_from_file():
...
+        if object_type in (TempObjectType.PROCEDURE, TempObjectType.FUNCTION):
+            target_func = func
+
+        if target_func is None:
+            return default_arg_names
+        return inspect.getfullargspec(target_func).args
+
+    def get_arg_names_from_file() -> List[str]:
+        filename, func_name = func[0], func[1]
+        if not is_local_python_file(filename):
```
sfc-gh-aling, line 524: why we skip the remote file?

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bkogan: Yeah, I was following the naming of the function above (get_opt_arg_defaults) since this function has a very similar pattern. A better name would be handler_name or something like that.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bkogan: The scenario I was thinking about was for vectorized UDFs / UDTFs. In which case the number of args is the number of input column types found in the Pandas dataframe type, whereas the function will only have 1 argument for the dataframe. In this scenario, the flag doesn't really make sense to enable since there are no names. 

If they do enable it, I was thinking using 'df' as the arg name if it was a single column dataframe was fine, otherwise just falling back to the defaults. Do you think this warrants a warning log?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -493,6 +494,79 @@ def get_opt_arg_defaults_from_file():
...
+                filename, TABLE_FUNCTION_PROCESS_METHOD, func_name
+            )
+        elif object_type == TempObjectType.AGGREGATE_FUNCTION:
+            arg_names = retrieve_func_arg_names_from_source(
+                filename, AGGREGATE_FUNCTION_ACCULUMATE_METHOD, func_name
+            )
+        elif object_type in (TempObjectType.FUNCTION, TempObjectType.PROCEDURE):
+            arg_names = retrieve_func_arg_names_from_source(filename, func_name)
+
+        return arg_names
```
sfc-gh-aling, line 539: it seems like retrieve_func_arg_names_from_source can return None if func does not exist in the file

sfc-gh-bkogan: Good catch. Changed these to return optional lists and check for None on the caller side.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bkogan: I tried to stay consistent with `get_opt_arg_defaults`, which also returns defaults for non local files. I assume there is some limitation with inspecting remote files?

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aling: discussed offline, let's add a comment explaining why the mistmatch can happen due to vectorized udxf

PR #4055, opened 2026-01-22
Add support for DECFLOAT type

sfc-gh-bkogan:

   

   Fixes SNOW-2185194

   This adds support for the [DECFLOAT](https://docs.snowflake.com/en/sql-reference/data-types-numeric#label-data-type-decfloat) numeric data type, which is a decimal floating-point number. `decimal.Decimal` literals are still inferred as `NUMBER` type. Not sure if we can change that in any way without being backwards incompatible. Users can use `lit` with an explicit `datatype` argument for now.

	From reading the connector [PR](https://github.com/snowflakedb/snowflake-connector-python/pull/2167), it seems that the user is expected to set the `decimal` context to precision 38 in order to preserve the precision stored on the server. I kept that behavior here.

### DIFF
--- a/ci/snyk/requirements.txt
+++ b/ci/snyk/requirements.txt
@@ -1,5 +1,5 @@
 setuptools>=40.6.0
 wheel
 cloudpickle>=1.6.0,<=3.1.1,!=2.1.0,!=2.2.0
-snowflake-connector-python>=3.17.0,<5.0.0
+snowflake-connector-python>=4.0.0,<5.0.0
 typing-extensions>=4.1.0,<5.0.0
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -983,6 +983,9 @@ def return_all_datatypes(
     reason="session.sql not supported in local testing",
 )
 def test_register_sp_with_preserve_parameter_names(session, resources_path):
+    sp_pow_name = Utils.random_name_for_temp_object(TempObjectType.PROCEDURE)
+    mod5_name = Utils.random_name_for_temp_object(TempObjectType.PROCEDURE)
+
     def sp_pow(session_, x, y):
         return (
             session_.create_dataframe([[x, y]])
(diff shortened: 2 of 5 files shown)


src/snowflake/snowpark/_internal/type_utils.py
```diff
@@ -188,7 +189,7 @@ def convert_metadata_to_sp_type(
         return convert_sf_to_sp_type(
             column_type_name,
             metadata.precision or 0,
-            metadata.scale or 0,
+            metadata.scale,
```
sfc-gh-bkogan, line 192: I'm not sure if this is safe to do. From my experimentation, the connector will return type code 0 (FIXED) with precision = 38 and scale = None for DECFLOAT type. In order to detect the None, I made the argument nullable. Need someone with more context of this part to double check.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
       - `space`
       - `split_part`
 - Added `preserve_parameter_names` flag to sproc, UDF, UDTF, and UDAF creation
+- Added support for the `DECFLOAT` data type
```
sfc-gh-yuwang: can you add a brief description of this new data type in the change log? like:
Added support for data type `DECFLOAT` that ....

src/snowflake/snowpark/types.py
```diff
@@ -461,6 +461,13 @@ def _fill_ast(self, ast: proto.DataType) -> None:
         ast.double_type = True
 
 
+class DecFloatType(_FractionalType):
+    """DecFloat data type. This maps to the DECFLOAT data type in Snowflake."""
+
+    def _fill_ast(self, ast: proto.DataType) -> None:
```
sfc-gh-yuwang, line 467: I think we should implement other inherent function like simple_string(), json_value(), etc?

src/snowflake/snowpark/_internal/type_utils.py
sfc-gh-yuwang: I think this is safe as it does not seems to break any tests

src/snowflake/snowpark/types.py
sfc-gh-bkogan: The inherited behavior will result with "decfloat" as the output for both those, which I think is right. In the DecimalType type, it needs to be overwritten since there are parameters like precision and scale, but this type does not have any parameters.

CHANGELOG.md
sfc-gh-bkogan: Updated

PR #4068, opened 2026-02-03
SNOW-3034995: Add cloudpickle automatically when using artifact_repository with packages

sfc-gh-aling:

   

   Fixes SNOW-3034995

**Problem**
When using artifact_repository with custom packages, cloudpickle was not automatically added to the package list, causing ModuleNotFoundError: No module named 'cloudpickle' at runtime in UDF execution.
**Root Cause**
In src/snowflake/snowpark/_internal/udf_utils.py (lines 1232-1237), when both artifact_repository and packages were provided, the code directly assigned resolved_packages = packages without calling session._resolve_packages(), which meant cloudpickle was never added to the dependency list.
**Solution**
Added logic to automatically include cloudpickle when using artifact repositories with custom packages:

- Check for cloudpickle: After assigning the packages list, check if cloudpickle is already present using prefix match
- Auto-add if missing: If cloudpickle is not found, append cloudpickle=={version} to the package list
- Preserve user choice: If user already specified cloudpickle with any version constraint, respect it and don't duplicate

### DIFF
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -2331,7 +2331,7 @@ def test_sproc_artifact_repository(session):
                 )
     except SnowparkSQLException as ex:
         if "No matching distribution found for snowflake-snowpark-python" in str(ex):
-            pytest.mark.xfail(
+            pytest.xfail(
                 "Unreleased snowpark versions are unavailable in artifact repository."
             )
 
@@ -2366,7 +2366,7 @@ def test_sproc_artifact_repository_with_packages_includes_cloudpickle(session):
         assert test_cloudpickle_sproc(session=session) == "6"
     except SnowparkSQLException as ex:
         if "No matching distribution found for snowflake-snowpark-python" in str(ex):
-            pytest.mark.xfail(
+            pytest.xfail(
                 "Unreleased snowpark versions are unavailable in artifact repository."
             )
 
--- a/tests/unit/test_udf.py
+++ b/tests/unit/test_udf.py
@@ -116,7 +116,7 @@ def test_artifact_repository_adds_cloudpickle():
         udf_name="test_udf",
         stage_location=None,
         imports=None,
-        packages=["urllib3", "requests", "cloudpickle-non-existing"],
+        packages=["urllib3", "requests", "invalid package!!!"],
         artifact_repository="SNOWPARK_PYTHON_TEST_REPOSITORY",
     )
     _, _, _, all_packages, _, _ = result


src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1234,7 +1235,12 @@ def resolve_imports_and_packages(
                 raise TypeError(
                     "Artifact repository requires that all packages be passed as str."
                 )
-            resolved_packages = packages
+            resolved_packages = list(packages)
```
sfc-gh-joshi: nit: why do we need the `list` call here?

tests/integ/test_udf.py
```diff
@@ -2916,6 +2916,46 @@ def test_urllib() -> str:
...
 
 
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="artifact repository not supported in local testing",
+)
+@pytest.mark.skipif(IS_NOT_ON_GITHUB, reason="need resources")
+@pytest.mark.skipif(
+    sys.version_info < (3, 9), reason="artifact repository requires Python 3.9+"
+)
```
sfc-gh-joshi: We no longer support 3.8, is this still necessary?

sfc-gh-aling: good catch, let me drop it

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aling: added by cursor.. removed it

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1230,11 +1230,20 @@ def resolve_imports_and_packages(
...
-                )
+            has_cloudpickle = False
+            for pkg in packages:
+                if not isinstance(pkg, str):
+                    raise TypeError(
+                        "Artifact repository requires that all packages be passed as str."
+                    )
+                # Note: According to PyPI search (https://pypi.org/search/?q=cloudpickle), and Anaconda search (https://anaconda.org/search?q=cloudpickle),
+                # "cloudpickle" is the only package with this prefix, making startswith() check safe.
+                if pkg.startswith("cloudpickle"):
```
sfc-gh-joshi: Why not just do an equality check in case some other `cloudpickle-` package gets added in the future?

sfc-gh-aling: this is for backward compatibility.

initially I wanted to use `Requirement`, but it can raise `InvalidRequirement` -- today we don't validate user inputs, so it may potentially break users due to new error

then I wanted to use string match, but there are different forms cloudpickle >=, ~=, <=, !=, cloudpickle @ ..., and more, if I iterate all the case, it makes the implementation inefficient.

I wanted to use try Requirement check - except string check case, but I feel it's overkilling.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aling: actually this is legit, as per the auto code review tool, it avoid mutating the original input

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1230,11 +1230,20 @@ def resolve_imports_and_packages(
...
-                )
-            resolved_packages = packages
+            has_cloudpickle = False
+            for pkg in packages:
+                if not isinstance(pkg, str):
+                    raise TypeError(
+                        "Artifact repository requires that all packages be passed as str."
+                    )
+                # Note: According to PyPI search (https://pypi.org/search/?q=cloudpickle), and Anaconda search (https://anaconda.org/search?q=cloudpickle),
+                # "cloudpickle" is the only package with this prefix, making startswith() check safe.
```
sfc-gh-bkogan: The artifact repository, e.g. PyPI, could change in the future and add a new package called cloudpickle_*. A better approach would be using `Requirement(package)` built in and then checking `package.name.lower()`. See `_parse_packages`.

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1230,11 +1230,20 @@ def resolve_imports_and_packages(
...
+                if not has_cloudpickle and pkg.startswith("cloudpickle"):
+                    has_cloudpickle = True
```
sfc-gh-bkogan: nit: `has_cloudpickle = has_cloudpickle or pkg.startswith("cloudpickle")`.

sfc-gh-bkogan (commented):
Thanks

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bkogan: Just saw the other thread, perhaps its better to still use it and catch `InvalidRequirement`?

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1230,11 +1230,20 @@ def resolve_imports_and_packages(
...
+
+            resolved_packages = list(packages)
+            if not has_cloudpickle:
+                resolved_packages.append(f"cloudpickle=={cloudpickle.__version__}")
```
sfc-gh-jrose: nit: This could be more concise:
```suggestion
            if not all(isinstance(package, str) for package in packages):
                raise TypeError(
                    "Artifact repository requires that all packages be passed as str."
                )
            cloud_pickle = [] if any(pkg.startswith("cloudpickle") for pkg in packages) else [f"cloudpickle=={cloudpickle.__version__}"]
            resolved_packages = packages + cloudpickle
```

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-jrose: +1 to using Requirement

sfc-gh-aling: let me try making the code change since that's the preferred approach by most of us

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aling: updated to use Requirement as the default one, and fall back to string match in case of error

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-aling: updated to use Requirement as the default one, and fall back to string match in case of error

PR #4073, opened 2026-02-05
SNOW-3052213: Add support for the DEFAULT_PYTHON_ARTIFACT_REPOSITORY parameter

sfc-gh-bkogan:

   

   Fixes SNOW-3052213

   Reads the `SYSTEM$GET_DEFAULT_PYTHON_ARTIFACT_REPOSITORY` built in system function to determine the default artifact repository to use for function creation. SQL generation is unaffected, just have to change the places that assume `artifact_repository is None` means we are using conda.

	Some things to note:
- `session._packages` removed and merged with `session._artifact_repository_packages`. This is because we can no longer assume the default repo name, so we must tag each package with it's corresponding repo name.
- Added `session._default_artifact_repository_cache` to cache repeated invocations of getting the default repo in the same DB/schema context. The cache introduces a gap where users who manually alter the parameter value in same context will not see that value reflected if the cache is already set. This is a reasonable tradeoff for now, we can follow up if it becomes an issue.

### DIFF
--- a/tests/integ/test_udf.py
+++ b/tests/integ/test_udf.py
@@ -1971,20 +1971,36 @@ def test_basic_pandas_udf(session):
     Utils.check_answer(df.select(return1_pandas_udf()), [Row("1"), Row("1")])
     Utils.check_answer(df.select(add_series_pandas_udf("a", "b")), [Row(3), Row(7)])
     Utils.check_answer(df.select(add_one_df_pandas_udf("a", "b")), [Row(4), Row(8)])
-    Utils.check_answer(
-        df.select(get_type_str_series_udf("a")),
-        [
-            Row("<class 'pandas.core.series.Series'>"),
-            Row("<class 'pandas.core.series.Series'>"),
-        ],
-    )
-    Utils.check_answer(
-        df.select(get_type_str_df_udf("a")),
-        [
-            Row("<class 'pandas.core.frame.DataFrame'>"),
-            Row("<class 'pandas.core.frame.DataFrame'>"),
-        ],
-    )
+    try:
+        Utils.check_answer(
+            df.select(get_type_str_series_udf("a")),
+            [
+                Row("<class 'pandas.core.series.Series'>"),
+                Row("<class 'pandas.core.series.Series'>"),
+            ],
+        )
+        Utils.check_answer(
+            df.select(get_type_str_df_udf("a")),
+            [
+                Row("<class 'pandas.core.frame.DataFrame'>"),
+                Row("<class 'pandas.core.frame.DataFrame'>"),
+            ],
+        )
+    except AssertionError:
+        Utils.check_answer(
+            df.select(get_type_str_series_udf("a")),
+            [
+                Row("<class 'pandas.Series'>"),
+                Row("<class 'pandas.Series'>"),
+            ],
+        )
+        Utils.check_answer(
+            df.select(get_type_str_df_udf("a")),
+            [
+                Row("<class 'pandas.DataFrame'>"),
+                Row("<class 'pandas.DataFrame'>"),
+            ],
+        )
 
 
 @pytest.mark.skipif(
@@ -2100,20 +2116,29 @@ def test_pandas_udf_type_hints(session):
         (
             DateType,
             [[datetime.date(2021, 12, 20)]],
-            ("<class 'pandas._libs.tslibs.timestamps.Timestamp'>",),
+            (
+                "<class 'pandas._libs.tslibs.timestamps.Timestamp'>",
+                "<class 'pandas.Timestamp'>",
+            ),
             ("datetime64[s]", "datetime64[ns]", "datetime64[us]"),
         ),
         (ArrayType, [[[1]]], ("<class 'list'>",), ("object",)),
         (
             TimeType,
             [[datetime.time(1, 1, 1)]],
-            ("<class 'pandas._libs.tslibs.timedeltas.Timedelta'>",),
+            (
+                "<class 'pandas._libs.tslibs.timedeltas.Timedelta'>",
+                "<class 'pandas.Timedelta'>",
+            ),
             ("timedelta64[s]", "timedelta64[ns]", "timedelta64[us]"),
         ),
         (
             TimestampType,
             [[datetime.datetime(2016, 3, 13, 5, tzinfo=datetime.timezone.utc)]],
-            ("<class 'pandas._libs.tslibs.timestamps.Timestamp'>",),
+            (
+                "<class 'pandas._libs.tslibs.timestamps.Timestamp'>",
+                "<class 'pandas.Timestamp'>",
+            ),
             ("datetime64[s]", "datetime64[ns]", "datetime64[us]"),
         ),
     ],
--- a/src/snowflake/snowpark/_internal/type_utils.py
+++ b/src/snowflake/snowpark/_internal/type_utils.py
@@ -1468,18 +1468,25 @@ def format_year_month_interval_for_display(
     # Default initialization
     years = "0"
     months = "0"
-    is_negative = False
+    is_negative = cell.startswith("-")
+    # Remove the sign prefix and parse the remaining part
+    remaining = cell[1:]  # Remove the "+" or "-" prefix: "1-6"
 
     if has_internal_dash:
         # Format like "+1-03" or "-1-03" or "-1-6" (compound year-month)
-        is_negative = cell.startswith("-")
-
-        # Remove the sign prefix and parse the remaining "year-month" part
-        remaining = cell[1:]  # Remove the "+" or "-" prefix: "1-6"
         if "-" in remaining:
             parts = remaining.split("-", 1)  # Split only on first dash: ["1", "6"]
             years = str(int(parts[0]))
             months = str(int(parts[1]))
+    else:
+        # Format like "+2" or "-3"
+        if (
+            start_field == YearMonthIntervalType.YEAR
+            and end_field == YearMonthIntervalType.YEAR
+        ):
+            years = str(int(remaining))
+        else:
+            months = str(int(remaining))
 
     # Format based on start/end field
     sign_prefix = "-" if is_negative else ""
@@ -1502,9 +1509,7 @@ def format_year_month_interval_for_display(
     ):
         # Months only: MONTH - calculate total months
         total_months = int(years) * 12 + int(months)
-        if is_negative:
-            total_months = -total_months
-        return f"INTERVAL '{total_months}' MONTH"
+        return f"INTERVAL '{sign_prefix}{total_months}' MONTH"
 
 
 def format_day_time_interval_for_display(
--- a/src/snowflake/snowpark/dataframe_ai_functions.py
+++ b/src/snowflake/snowpark/dataframe_ai_functions.py
@@ -1212,7 +1212,7 @@ class DataFrameAIFunctions:
             ...     mode='LAYOUT',
             ... )
             >>> result = json.loads(result_df.collect()[0]["PARSED"])
-            >>> "| Customer Name |" in result["content"] and "| Country |" in result["content"]
+            >>> "| Customer Name |" in result["content"] and "| Country |" in result["content"] # doctest: +SKIP
             True
 
             >>> # Parse with page splitting for long documents (PDF only)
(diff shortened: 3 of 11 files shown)


src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1222,21 +1222,44 @@ def resolve_imports_and_packages(
     Optional[str],
     bool,
 ]:
-    if artifact_repository and artifact_repository != "conda":
-        # Artifact Repository packages are not resolved
+    from snowflake.snowpark.session import ANACONDA_SHARED_REPOSITORY
```
sfc-gh-bkogan: is there a better place to put this constant to avoid the import cycle?

src/snowflake/snowpark/session.py
```diff
@@ -2377,6 +2387,53 @@ def _upload_unsupported_packages(
...
+        fallback needs to be updated here.
+        """
+        if isinstance(self._conn, MockServerConnection):
+            return DEFAULT_ARTIFACT_REPOSITORY
+
+        cache_key = (self.get_current_database(), self.get_current_schema())
+
+        if (
+            self._default_artifact_repository_cache is not None
+            and self._default_artifact_repository_cache[0] == cache_key
```
sfc-gh-bkogan: I'm not sure exactly what safety guarantees Python gives you for reading and writing from a variable in threads. Is this a data race and should be wrapped in the package lock?

sfc-gh-bkogan:
Tests currently failing because the system function isn't available in the test environment yet

src/snowflake/snowpark/session.py
```diff
@@ -2377,6 +2387,53 @@ def _upload_unsupported_packages(
...
+        ):
+            return self._default_artifact_repository_cache[1]
```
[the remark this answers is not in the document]
sfc-gh-bkogan: yeahhhh you right

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1223,17 +1227,35 @@ def resolve_imports_and_packages(
...
+            if session
+            else DEFAULT_ARTIFACT_REPOSITORY
+        )
+
+    existing_packages_dict = (
+        session._artifact_repository_packages[artifact_repository] if session else {}
+    )
+
+    if artifact_repository != ANACONDA_SHARED_REPOSITORY:
+        # Non-conda artifact repository - skip conda-based package resolution
```
sfc-gh-joshi, line 1246: Should DEFAULT_ARTIFACT_REPOSITORY also trigger this code path?

src/snowflake/snowpark/session.py
```diff
@@ -321,6 +321,11 @@
 WRITE_PANDAS_CHUNK_SIZE: int = 100000 if is_in_stored_procedure() else None
 WRITE_ARROW_CHUNK_SIZE: int = 100000 if is_in_stored_procedure() else None
 
+# The fully qualified name of the Anaconda shared repository (conda channel).
+ANACONDA_SHARED_REPOSITORY = "snowflake.snowpark.anaconda_shared_repository"
+# In case of failures or the current default artifact repository is unset, we fallback to this
+DEFAULT_ARTIFACT_REPOSITORY = ANACONDA_SHARED_REPOSITORY
```
sfc-gh-joshi: Why does this need to be a separate global variable? If we want to allow users to configure this, we should be setting a session parameter instead.

src/snowflake/snowpark/session.py
```diff
@@ -599,10 +604,17 @@ def __init__(
...
         self._artifact_repository_packages: DefaultDict[
             str, Dict[str, str]
         ] = defaultdict(dict)
+        # Single-entry cache for the default artifact repository value.
+        # Stores a tuple of ((database, schema), cached_value).  Only one entry is
+        # kept at a time – switching to a different database/schema will evict the old
+        # value and trigger a fresh query on the next call.
+        self._default_artifact_repository_cache: Optional[
+            Tuple[Tuple[Optional[str], Optional[str]], str]
+        ] = None
```
sfc-gh-joshi, line 614: Would there be value to storing a dict of (database, schema) -> cached_value instead? That way we wouldn't need to rerun the query upon switching db/schema, though it may run the risk of the value becoming stale.

src/snowflake/snowpark/session.py
sfc-gh-mayliu: What was the context that made anaconda repository the default for Snowpark Python? What about `snowflake.snowpark.pypi_shared_repository`? https://docs.snowflake.com/en/developer-guide/udf/python/udf-python-packages#get-started

src/snowflake/snowpark/session.py
```diff
@@ -2377,6 +2387,50 @@ def _upload_unsupported_packages(
...
+                return self._default_artifact_repository_cache[1]
+
+            try:
+                python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+                result = self._run_query(
+                    f"SELECT SYSTEM$GET_DEFAULT_PYTHON_ARTIFACT_REPOSITORY('{python_version}')"
+                )
+                value = result[0][0] if result else None
+                resolved = value or DEFAULT_ARTIFACT_REPOSITORY
+            except Exception as e:
```
sfc-gh-mayliu, line 2422: a broad exception might hide issues, would catching specific exceptions and raising on certain error types help?

src/snowflake/snowpark/session.py
```diff
@@ -2377,6 +2387,50 @@ def _upload_unsupported_packages(
...
+            if (
+                self._default_artifact_repository_cache is not None
+                and self._default_artifact_repository_cache[0] == cache_key
+            ):
+                return self._default_artifact_repository_cache[1]
+
+            try:
+                python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+                result = self._run_query(
+                    f"SELECT SYSTEM$GET_DEFAULT_PYTHON_ARTIFACT_REPOSITORY('{python_version}')"
```
sfc-gh-mayliu, line 2418: Can you add a test that verifies graceful fallback when `SYSTEM$GET_DEFAULT_PYTHON_ARTIFACT_REPOSITORY` fails?

src/snowflake/snowpark/session.py
```diff
@@ -599,10 +604,17 @@ def __init__(
...
-        self._packages: Dict[str, str] = {}
+        # map of artifact repository name -> packages that should be added to functions under that repository
         self._artifact_repository_packages: DefaultDict[
             str, Dict[str, str]
         ] = defaultdict(dict)
+        # Single-entry cache for the default artifact repository value.
+        # Stores a tuple of ((database, schema), cached_value).  Only one entry is
+        # kept at a time – switching to a different database/schema will evict the old
+        # value and trigger a fresh query on the next call.
+        self._default_artifact_repository_cache: Optional[
```
sfc-gh-mayliu, line 612: what's the motivation for using a cache here, given that it's single-entry? If users commonly work across different DB/schemas, would a small LRU cache be more efficient?

src/snowflake/snowpark/session.py
sfc-gh-aling: are users supposed to read/write these two var / is this a request from the server team?
if not let's make them private first
```suggestion
# The fully qualified name of the Anaconda shared repository (conda channel).
_ANACONDA_SHARED_REPOSITORY = "snowflake.snowpark.anaconda_shared_repository"
# In case of failures or the current default artifact repository is unset, we fallback to this
_DEFAULT_ARTIFACT_REPOSITORY = ANACONDA_SHARED_REPOSITORY
```

src/snowflake/snowpark/_internal/udf_utils.py
```diff
@@ -1137,8 +1138,11 @@ def add_snowpark_package_to_sproc_packages(
             packages = [this_package]
         else:
             with session._package_lock:
-                if package_name not in session._packages:
-                    packages = list(session._packages.values()) + [this_package]
+                existing_packages = session._artifact_repository_packages[
```
sfc-gh-aling, line 1145: is `artifact_repository` guaranteed to exist in the `session._artifact_repository_packages`

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bkogan: Package resolution via the information schema will only happen with the conda repo. The server team mentioned they wont update this table; it will always be for conda.

src/snowflake/snowpark/session.py
sfc-gh-bkogan: There is a separate session parameter users can set. This default is used when the user never set a default via the parameter, or the system function fails. Maybe a better name is needed?

As to why it's separate, the server team mentioned to me that they will ask us to update the default (unset) value to pypi in the future. To make that change as minimal as possible, only this var needs to change, whereas we can keep ANACONDA_SHARED_REPOSITORY and all the code paths that explicitly check for it untouched. Does that make sense?

src/snowflake/snowpark/session.py
sfc-gh-bkogan: Yeah, it was the initial approach, but after discussion, we decided to keep it as a single value cache for the staleness reason you mentioned. This gives users a way to refresh the value in case of issues.

src/snowflake/snowpark/session.py
sfc-gh-bkogan: I'm not sure what these docs mean tbh. Currently the default if you don't specify any repo is conda, but the server team plans to move to pypi in the future.

sfc-gh-bkogan: They should be private. It seems we haven't been adding underscores for vars above it though. Will fix for these.

src/snowflake/snowpark/_internal/udf_utils.py
sfc-gh-bkogan: No, but `_artifact_repository_packages` is a `defaultdict` so there is no risk of error here.

src/snowflake/snowpark/session.py
sfc-gh-bkogan: We expect users to not switch contexts too often. With a single-entry we can evict old context values and refresh them if the user switches back, which helps address the staleness issue if we kept a dictionary/LRU.

src/snowflake/snowpark/session.py
sfc-gh-bkogan: The only expected error scenario here is if run_query fails, which we want to fallback. I am not sure of other error states, so for now I think catch all and log is fine.

src/snowflake/snowpark/session.py
sfc-gh-bkogan: Already added in test/unit/test_session

src/snowflake/snowpark/session.py
sfc-gh-joshi: Being separate makes sense.

You can set client-side session parameters with default values (example: https://github.com/snowflakedb/snowpark-python/blob/878fe0eca15ca9c67d89eb5246622486ddf1c586/src/snowflake/snowpark/session.py#L304-L306 and ctrl-F for other instances of this variable to see how it's used). IMO this is preferable to a global variable, especially if we run into a scenario where a user may have multiple session objects operating on different database/schema combinations.

sfc-gh-joshi (approved):

src/snowflake/snowpark/session.py
sfc-gh-bkogan: Resolved offline:

Currently there is no parameter available for this on the server side, and it's not really urgent to add it. Also the rollout plan from the server team to change the value will likely depend on the python version, i.e. only for 3.14, so I think we should wait a bit until we have a clearer plan from them.

tests/integ/test_udf.py
```diff
@@ -3070,3 +3070,49 @@ def test_urllib() -> str:
...
+    sys.version_info < (3, 9), reason="artifact repository requires Python 3.9+"
+)
+def test_use_default_artifact_repository(db_parameters):
+    with Session.builder.configs(db_parameters).create() as session:
+        session.use_schema("public")
+        session.sql(
+            "ALTER SESSION SET ENABLE_DEFAULT_PYTHON_ARTIFACT_REPOSITORY = true"
+        ).collect()
+        session.sql(
+            "ALTER schema set DEFAULT_PYTHON_ARTIFACT_REPOSITORY = testdb_snowpark_python.testschema_snowpark_python.SNOWPARK_PYTHON_TEST_REPOSITORY"
```
sfc-gh-aling, line 3118: 1. do we need to unset in the end?
2. do we need to create a new schema for testing, I'm wondering if this has the side effect affecting other tests using the same schema public

sfc-gh-bkogan: 1. Already unset at the end
2. Good point, will do

PR #4077, opened 2026-02-13
SNOW-3074563: Prepare for Python 3.14 Support

sfc-gh-bkogan:

   

   Fixes SNOW-3074563

   Add support for 3.14 runtime.

	- `copy.copy(super())` no longer works, changed to `new.__dict__.update(self.__dict__)` approach
	- `SnowflakePlan` has a `uuid` property that now interferes with the `uuid` type annotation (since in 3.14 it seems type annotations are now evaluated lazily). Fixed by aliasing the UUID library to `uuid_lib`.
	- `typing.Union[x,y]` now is represented as `x | y` in strings, so updated code and tests to check for either.
	- Pickle default protocol bumped to version 5, so some hardcoded hex strings in tests need to change.

	We will NOT update the changelog since it will be released as part of a BCR. We will update the changelog after that is released.

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -4981,7 +4981,7 @@ class Session:
                     handler_name,
                     return_type=return_type_map[return_type],
                     input_types=[StringType(), StringType()],
-                    packages=["snowflake-snowpark-python", "lxml<=6.0.2"],
+                    packages=["snowflake-snowpark-python", "lxml<7"],
                     replace=True,
                     _emit_ast=False,
                     _suppress_local_package_warnings=True,
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -1457,7 +1457,7 @@ class DataFrameReader:
             name=udtf_name,
             output_schema=output_schema,
             input_types=input_types,
-            packages=["snowflake-snowpark-python", "lxml<=6.0.2"],
+            packages=["snowflake-snowpark-python", "lxml<7"],
             if_not_exists=True,
             skip_upload_on_content_match=True,
             _suppress_local_package_warnings=True,


.github/workflows/precommit.yml
```diff
@@ -132,6 +132,18 @@ jobs:
...
+            cloud-provider: aws
+            os: ubuntu-latest-64-cores
+          # run py 3.14 doctests on azure/macos
+          - python-version: "3.14"
+            cloud-provider: azure
+            os: macos-latest
+          # # run py 3.14 tests on gcp on windows
+          - python-version: "3.14"
+            cloud-provider: gcp
+            os: windows-latest-64-cores
```
sfc-gh-bkogan, line 138: We should keep a subset of these? And/or remove an older version?

sfc-gh-joshi: It looks like 3.9 is already EOL, and Snowflake server-side 3.9 support is being decomissioned at the end of April: https://docs.snowflake.com/en/developer-guide/python-runtime-support-policy

src/snowflake/snowpark/_internal/analyzer/expression.py
```diff
@@ -158,7 +157,9 @@ def expr_id(self) -> uuid.UUID:
         return self._expr_id
 
     def __copy__(self):
-        new = copy.copy(super())
+        cls = self.__class__
+        new = cls.__new__(cls)
+        new.__dict__.update(self.__dict__)
```
sfc-gh-joshi, line 162: Just for clarity: this is necessary to ensure fields of classes that inherit `NamedExpression` also are copied, correct?

src/snowflake/snowpark/mock/_plan.py
```diff
@@ -590,7 +590,7 @@ def handle_function_expression(
             if param_name in exp.named_arguments:
                 type_hint = str(type_hints.get(param_name, ""))
                 keep_literal = "Column" not in type_hint
-                if type_hint == "typing.Optional[dict]":
+                if type_hint in ["typing.Optional[dict]", "dict | None"]:
```
sfc-gh-joshi: Do you know if Python internally normalizes this string, or if it can return variations like `typing.Dict | None`, `dict | types.NoneType`, etc.?

If it's normalized, would it be possible to instead compare `type_hint` against a normalized annotation (like calling `get_type_hints` on a function we declared to use `dict | None`) so we don't need to hard-code these string literals?

tests/integ/conftest.py
```diff
@@ -357,6 +357,7 @@ def session(
             "alter session set ENABLE_EXTRACTION_PUSHDOWN_EXTERNAL_PARQUET_FOR_COPY_PHASE_I='Track';"
         ).collect()
         session.sql("alter session set ENABLE_ROW_ACCESS_POLICY=true").collect()
+        session.sql("alter session set ENABLE_PYTHON_3_14=true").collect()
```
sfc-gh-joshi: Can we guard this against the locally-installed Python version?

src/snowflake/snowpark/_internal/analyzer/expression.py
sfc-gh-bkogan: Right, which is just the attribute classes IIRC. Seems it's called when we copy a dataframe plan.

src/snowflake/snowpark/mock/_plan.py
sfc-gh-bkogan: From experimenting, it seems it will not normalize, it will return either `typing.Optional[dict]` or `dict | None` depending on how the type annotation was authored.

.github/workflows/precommit.yml
sfc-gh-bkogan: Ok. But should I keep all 3 3.14 runs? Also not sure how we want to decommission 3.9 on the client side. Perhaps that should happen separately?

sfc-gh-bkogan: Seems like mac os tests are currently always skipped though

sfc-gh-joshi: When 3.8 reached EOL, we just removed it from the test matrix and updated README/CONTRIBUTING accordingly: https://github.com/snowflakedb/snowpark-python/pull/3189 (in fact, we did so a month before its official decomission).

I believe we should actually replace the 3.9 here with 3.10 since we're testing the min supported version. We could replace one of the 3.12/3.13 jobs with 3.14 to keep load low while still covering cloud/python version/OS permutations.

sfc-gh-bkogan: I think let's remove 3.9 in a separate PR to keep this one smaller. I removed one of the 3.12 tests in favor of keeping both 3.14 tests.

src/snowflake/snowpark/mock/_plan.py
sfc-gh-joshi: It looks like we can actually compare against the non-stringified version of the object. Tested on 3.11, but should work on any version:

```
>>> typing.Optional[dict] == typing.Union[None, dict]
True
>>> typing.Optional[dict] == (None | dict)
True
>>> (dict | None) == (None | dict)
True
```

sfc-gh-bkogan: Interesting find! Updated

.github/workflows/precommit.yml
sfc-gh-aling: +1 on remove 3.9 completely in one shot after it's officially deprecated

recipe/meta.yaml
```diff
@@ -37,7 +38,9 @@ requirements:
     - wheel
     # Snowpark IR
     - protobuf==3.20.1  # [py<=310]
-    - protobuf==4.25.3  # [py>310]
+    - protobuf==4.25.3  # [py>310 and py<314]
+    - protobuf==5.29.3  # [py>=314]
+    - libprotobuf >=5.29.3 # [py>=314]
```
sfc-gh-aling, line 43: @sfc-gh-heshah do we need any extra validation on python3.14 for ast generation using newer protobuf packages?

I see `Test AST enabled (phase0) py-ubuntu-latest-3.14 ` passed in the merge gate, is this enough?

scripts/conda_build.sh
```diff
@@ -4,3 +4,4 @@ conda build recipe/ -c sfe1ed40 --python=3.10 --numpy=1.21
 conda build recipe/ -c sfe1ed40 --python=3.11 --numpy=1.23
 conda build recipe/ -c sfe1ed40 --python=3.12 --numpy=1.26
 conda build recipe/ -c sfe1ed40 --python=3.13 --numpy=2.2.0
+conda build recipe/ -c sfe1ed40 --python=3.14 --numpy=2.4.2
```
sfc-gh-aling: where is the numpy version 2.4.2 from / why not other version?

I used to remember there is a link from conda/3rd party instructing the version, do you have the link?

src/snowflake/snowpark/session.py
```diff
@@ -4981,7 +4981,7 @@ def _get_or_register_xpath_udf(
                     handler_name,
                     return_type=return_type_map[return_type],
                     input_types=[StringType(), StringType()],
-                    packages=["snowflake-snowpark-python", "lxml<6"],
+                    packages=["snowflake-snowpark-python", "lxml<=6.0.2"],
```
sfc-gh-aling: is this for 3.14 support?
does lxml 6 introduce any BCR? / do we have the context why it was pinned to <6 before, it's probably good to check git blame

scripts/conda_build.sh
sfc-gh-bkogan: I just took the latest version at the time. I don't have a link.

src/snowflake/snowpark/session.py
sfc-gh-bkogan: lxml <6 does not support python 3.14, so I bumped it. The git blame points to @sfc-gh-jdu, do you have context?

scripts/conda_build.sh
sfc-gh-bkogan: Looks like it should be 2.3. @sfc-gh-aalam do you know if we need the patch version too, like 2.3.5? Since we did that for 3.13

recipe/meta.yaml
sfc-gh-heshah: Yes, that is sufficient.

src/snowflake/snowpark/session.py
sfc-gh-aling: we can check change log https://github.com/lxml/lxml/releases, identify if we use any removed/to be deprecated api. if the answer is no, we are safe to upgrade

scripts/conda_build.sh
sfc-gh-aling: based upon the sproc connector practice, we use the following link to decide the numpy version
https://github.com/AnacondaRecipes/aggregate/blob/master/conda_build_config.yaml#L111

src/snowflake/snowpark/session.py
sfc-gh-bkogan: Based on cursor analysis, we should be good

scripts/conda_build.sh
sfc-gh-bkogan: switched to use 2.3

PR #4088, opened 2026-02-24
SNOW-3052213: fix session package management

sfc-gh-bkogan:

   

   Fixes SNOW-3052213

   Due to the server side depending on private members of the `Session` class, some changes from [the previous PR](https://github.com/snowflakedb/snowpark-python/pull/4073) need to be reverted. We need to keep `_packages` around, it now represents all packages under the `_DEFAULT_ARTIFACT_REPOSITORY`. Also reverted changing the argument positions of `_resolve_packages`.

	There was another issue found where if the schema was not selected, the system function would error instead of propagating up. Fixed by explicitly setting the entity.

### DIFF
--- a/tests/unit/test_session.py
+++ b/tests/unit/test_session.py
@@ -348,6 +348,26 @@ def test_resolve_packages_non_conda_artifact_repository(mock_server_connection):
     assert_packages(packages)
 
 
+def test_resolve_packages_optional_artifact_repository(mock_server_connection):
+    session = Session(mock_server_connection)
+    session._get_default_artifact_repository = MagicMock(
+        return_value="snowflake.snowpark.pypi_shared_repository"
+    )
+    session._artifact_repository_packages = {
+        "snowflake.snowpark.pypi_shared_repository": {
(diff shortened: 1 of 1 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -2132,6 +2145,13 @@ def _resolve_packages(
         Returns:
             List[str]: List of package specifiers
         """
+        if artifact_repository is None:
```
sfc-gh-aling, line 2148: nit on styling: can this be written as `a = a or b`?

sfc-gh-bkogan: yeah, after formatting it will get turned into

```
artifact_repository = (
            artifact_repository or self._get_default_artifact_repository()
        )
```

which seems about the same tbh

sfc-gh-aling: actually I think you would have very subtle perf gain if you write it as `a = a or b` (you can test it by using `timeit`)
but since this is not on the hot path, and compared to what the rest code is doing, this overhead is negligible.

PR #4089, opened 2026-02-25
SNOW-3156942: Add support for the ARRAY_UNION_AGG function

sfc-gh-bkogan:

   

   Fixes SNOW-3156942

   Adds support for calling `ARRAY_UNION_AGG`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 #### Bug Fixes
 
 - Fixed a bug where `Session.udf.register_from_file` did not properly process the `strict` and `secure` parameters.
+- Fixed a bug when create dataframe with small data(< array binding threshold), and error is raised when have string value in a DecimalType column.
 
 #### Improvements
 
--- a/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
+++ b/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
@@ -420,7 +420,9 @@ def to_sql(
         else:
             return f"'{value}' :: FLOAT"
 
-    if isinstance(value, Decimal) and isinstance(datatype, DecimalType):
+    if (isinstance(value, Decimal) or isinstance(value, str)) and isinstance(
+        datatype, DecimalType
+    ):
         return f"{value} :: {analyzer_utils.number(datatype.precision, datatype.scale)}"
 
     if isinstance(datatype, DateType):
--- a/tests/integ/test_dataframe.py
+++ b/tests/integ/test_dataframe.py
@@ -3567,6 +3567,30 @@ def test_create_dataframe_from_none_data(session):
     assert session.create_dataframe([None] * 20000).collect() == [Row(None)] * 20000
 
 
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="This test exercises VALUES SQL generation and large local relation plan",
+    run=False,
+)
+def test_create_dataframe_decimal_string_value_inline_and_large(session):
+    from snowflake.snowpark._internal.analyzer import analyzer
+
+    schema = StructType([StructField("a", DecimalType(10, 2))])
+
+    # Small local data insert string with decimal type shall not fail.
+    df_small = session.create_dataframe([["1.23"]], schema=schema)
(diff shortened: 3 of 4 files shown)


src/snowflake/snowpark/functions.py
```diff
@@ -7450,6 +7450,38 @@ def array_to_string(
     return _call_function("array_to_string", a, s, _emit_ast=_emit_ast)
 
 
+@publicapi
+def array_union_agg(col: ColumnOrName, _emit_ast: bool = True) -> Column:
+    """Returns an ARRAY that contains the union of the distinct values from the input
+    arrays in the specified column.
+
+    The values in the returned ARRAY are in no particular order, and the order is not
+    deterministic. The function ignores NULL values in the column and in arrays in the
```
sfc-gh-joshi, line 7459: Given that the order is explicitly non-deterministic, we should either add a sort call to the output, or add a `# doctest: +SKIP`

tests/integ/test_function.py
```diff
@@ -2170,6 +2177,42 @@ def _result_str2lst(result):
...
 
 
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="FEAT: array_union_agg function not supported",
+)
+def test_array_union_agg(session):
+    def _result_str2lst(result):
+        col_str = result[0][0]
+        col_lst = [int(i) for i in re.findall(r"-?\d+", col_str)]
```
sfc-gh-joshi: nit: instead of parsing the result string, can we directly index into the column's elements instead?

src/snowflake/snowpark/functions.py
sfc-gh-bkogan: Good point, will sort

tests/integ/test_function.py
sfc-gh-bkogan: Was following the example for a similar function above it, i'll look into why we are doing this regex though. seems weird

sfc-gh-bkogan: So the connector returns these as strings, example:
`'[\n  1,\n  1,\n  2,\n  3\n]'`

Not sure why...

sfc-gh-joshi (approved):

tests/integ/test_function.py
```diff
@@ -2170,6 +2177,42 @@ def _result_str2lst(result):
...
+        col_lst.sort()
+        return col_lst
+
+    df1 = session.create_dataframe([[[1, 1, 2]], [[1, 2, 3]]], schema=["a"])
+    result_str = df1.select(array_union_agg("a").alias("result")).collect()
+    result_list = _result_str2lst(result_str)
+    expected_result = [1, 1, 2, 3]
+    assert (
+        result_list == expected_result
+    ), f"Unexpected result: {result_list}, expected: {expected_result}"
```
sfc-gh-mayliu: since `array_union_agg` does not guarantee order, would directly comparing result_list and expected_result introduce flakiness?

tests/integ/test_function.py
```diff
@@ -2170,6 +2177,42 @@ def _result_str2lst(result):
     ), f"Unexpected result: {result_list}, expected: {expected_result}"
 
 
+@pytest.mark.skipif(
```
sfc-gh-aling: these days we usually do not add integ tests for built-in functions -- the doc test already covers it

```python
        >>> import re
        >>> df = session.create_dataframe([[[1, 1, 2]], [[1, 2, 3]]], schema=["a"])
        >>> row = df.select(array_union_agg("a").alias("result")).collect()[0]
        >>> sorted(int(i) for i in re.findall(r"\d+", row[0]))
        [1, 1, 2, 3]
```

tests/integ/test_function.py
sfc-gh-bkogan: Yeah. However, the `_result_str2lst` helper sorts the results, so we should be good.

tests/integ/test_function.py
sfc-gh-bkogan: yeah, i'll remove

tests/integ/test_function.py
sfc-gh-bkogan: I'm going to remove the integ test anyways though based on Adam's comment

PR #4096, opened 2026-02-27
SNOW-3176017: Fix accidental removal of aliases in certain JOIN statements

sfc-gh-joshi:

   

   Fixes SNOW-3176017

Reverts #4095, restoring the optimizations made in SNOW-2895675.

The original optimization, which replaced a layer of `SELECT "A" AS "A", "B" AS "B"` with `SELECT *` in certain join operations, caused a test failure in SnowML's CI pipeline. A minimal version of the test is reproduced in this PR in `test_query_generator.py::test_disambiguate_skips_quoted_alias`, and is as follows:
```
df1 = session.read.parquet(stage_filename)
df2 = session.create_dataframe(data, schema=["ID", "A", "B"])
df_res = df1.join(df2, on=["ID"])[['"COL_0"', '"COL_1"']]
```
This would generate SQL like the following:
<details>
<summary>bad SQL</summary>
<pre>
SELECT 
    "COL_0", 
    "COL_1"
 FROM (
 SELECT  * 
 FROM (
(
 SELECT $1 AS "ID", $2 AS """COL_0""", $3 AS """COL_1""" FROM  VALUES (0 :: INT, 1 :: INT, 2 :: INT), (3 :: INT, 4 :: INT, 5 :: INT
)
) AS SNOWPARK_LEFT 
INNER JOIN 
(
 SELECT 
    "ID", 
    "A", 
    "B"
 FROM (
 SELECT $1 AS "ID", $2 AS "A", $3 AS "B" FROM  VALUES (0 :: INT, 1 :: INT, 2 :: INT), (3 :: INT, 4 :: INT, 5 :: INT)
)
) AS SNOWPARK_RIGHT
 USING (ID)
)
)
</pre>
</details>

The file read operation was producing aliases of triple-quoted column names (which are used in SQL to produce single-quoted column names), which were silently dropped by the join optimization, resulting in the top-level references to `COL_0` and `COL_1` becoming invalid.

When [_disambiguate](https://github.com/snowflakedb/snowpark-python/blob/fac826ea8a1db214359021fd94652c92cc63bd15/src/snowflake/snowpark/dataframe.py#L341-L349) is called during join operations, the new optimization skipped alias generation if the left and right side did not share any common column names. However, the `_alias_if_needed` actually serves the additional function of [stripping quotes](https://github.com/snowflakedb/snowpark-python/blob/fac826ea8a1db214359021fd94652c92cc63bd15/src/snowflake/snowpark/dataframe.py#L282) from identifiers it processes, including from triple-quoted identifiers. Even though there were no common column names in this case, aliasing was still necessary to rename `"""COL_0"""` to `"COL_0"`.

This error path is triggered only by `ReadFile` operations, which more or less directly [generate their own SQL SELECT/COPY statements with aliasing inside](https://github.com/snowflakedb/snowpark-python/blob/54539913e8c647867993e6ede80a5a7f0d5270c7/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L2025-L2040). The join operation then implicitly normalized triple-quoted names like `"""COL_0"""` to `"COL_0"` (whether this behavior is itself correct should be investigated separately).


src/snowflake/snowpark/dataframe.py
```diff
@@ -338,25 +378,37 @@ def _disambiguate(
...
+        lhs_prefix,
+        lsuffix,
+        [] if isinstance(join_type, (LeftSemi, LeftAnti)) else common_col_names,
+    )
+    rhs_aliases = _get_aliased_column_names(
+        rhs, rhs_names, rhs_prefix, rsuffix, common_col_names
+    )
+    if all(
+        l_name == l_aliased for l_name, l_aliased in zip(lhs_names, lhs_aliases)
+    ) and all(r_name == r_aliased for r_name, r_aliased in zip(rhs_names, rhs_aliases)):
```
sfc-gh-aling, line 399: just to clarify: the core change (compared to previous change) is this condition check that "ensure none of the column names have changed due to internal quote stripping".

the rest of code change is the same to the previous one, except minor refactor due to `_get_aliased_column_names`util.

sfc-gh-joshi: Yes, that's right. The minor refactor lets us check the name an aliased column would have without needing to create the aliased column.

PR #4108, opened 2026-03-05
SNOW-3198432: Fix incorrect type in JSON schema inference with USE_RELAXED_TYPES

sfc-gh-joshi:

   

   Fixes SNOW-3198432

When USE_RELAXED_TYPES is set, schema inference operations will use the most generic type possible to represent data (e.g. NUMBER values are up-casted to DOUBLE). This is useful to minimize the number of records that need to be read when performing I/O operations.

The current implementation of the Snowpark file reader does not use the computed relaxed type for JSON values, causing SQL compilation errors if the inferred type from the first few records is too narrow to fit all values in the file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fixed a bug in `ai_complete` where `model_parameters` and `response_format` values containing single quotes would generate malformed SQL.
 - Fixed a bug in `DataFrameReader.xml()` where reading XML with a custom schema whose field names contain colons (e.g., `px:name`) raised a `SnowparkColumnException`.
 - Fixed a bug in that caused SQL compilation errors in `Session.read.json` when `INFER_SCHEMA` was set to True, and the `USE_RELAXED_TYPES` field of `INFER_SCHEMA_OPTIONS` was also set to True.
+- Fixed a bug where passing a DataFrame created from a SQL `SET` command to Streamlit's `st.write` method would raise an exception.
 
 #### Improvements
 
--- a/src/snowflake/snowpark/_internal/analyzer/schema_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/schema_utils.py
@@ -77,7 +77,7 @@ def analyze_attributes(
     # SQL commands which cannot be prepared
     # https://docs.snowflake.com/en/user-guide/sql-prepare.html
     if lowercase.startswith(
-        ("alter", "drop", "use", "create", "grant", "revoke", "comment")
+        ("alter", "drop", "use", "create", "grant", "revoke", "comment", "set")
     ):
         return command_attributes()
     if lowercase.startswith(("ls", "list")):
--- a/tests/integ/scala/test_sql_suite.py
+++ b/tests/integ/scala/test_sql_suite.py
@@ -218,3 +218,14 @@ def test_sql_start(session):
         len(df2.queries["queries"]) == 2
     )  # convert to result_scan because sql is non-select
     assert "RESULT_SCAN" in df2.queries["queries"][1]
+
+
+def test_sql_set_variable_to_pandas(session):
+    # Covers SNOW-3198369. Passing the result of a SQL SET operation to Streamlit's
+    # st.write (which calls to_pandas on the result of a session.sql call) should succeed,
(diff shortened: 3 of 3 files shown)


src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1322,7 +1322,7 @@ def _infer_schema_for_file_format(
                 ):
                     identifier = expression
                 else:
-                    identifier = f"$1:{name}::{type}"
+                    identifier = f"$1:{name}::{convert_sp_to_sf_type(datatype)}"
```
sfc-gh-aling: the change is not put under the condition of `use_relaxed_types`

is `type` guaranteed to be equal to `convert_sp_to_sf_type(datatype)` in not use_relaxed_types cases?

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -2434,6 +2434,38 @@ def write_csv(data):
...
 
 
+@pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="relaxed_types not supported by local testing mode",
+)
+def test_use_relaxed_types_json(session):
+    # Covers SNOW-3198432, where the inferred type was not properly used with relaxed types.
+    nrows = 20
+    ndjson_blob = "\n".join([json.dumps({"x": i}) for i in range(nrows)])
```
sfc-gh-aling, line 2444: why this case would fail in the original implementation and now it is fixed by your PR?

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -2434,6 +2434,38 @@ def write_csv(data):
...
+            infer_schema_options = {
+                "MAX_RECORDS_PER_FILE": 10,
+                "USE_RELAXED_TYPES": True,
+            }
+            df = (
+                session.read.option("INFER_SCHEMA", True)
+                .option("INFER_SCHEMA_OPTIONS", infer_schema_options)
+                .json(f"@{stage_name}/{os.path.basename(file.name)}")
+            )
+            Utils.check_answer(df, [Row(i) for i in range(nrows)])
```
sfc-gh-aling, line 2465: can we also assert that double type is used

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-joshi: It should be. `datatype` is constructed from the parsed components of `type` right above this.

https://github.com/snowflakedb/snowpark-python/blob/35e49256e7fdafd9413b6d13ac03e113822148a2/src/snowflake/snowpark/dataframe_reader.py#L1292-L1304

If you're worried about the change, I can guard it with `use_relaxed_types` to be safe.

tests/integ/scala/test_dataframe_reader_suite.py
sfc-gh-joshi: Previously, schema inference used only the first 10 rows of the frame, which had values 0-9. The inferred type from this is `NUMBER(1, 0)`, so as soon as it attempts to load the next row with value 10, the value no longer fits and we hit an error.

`USE_RELAXED_TYPES` up-casts this inferred type to DOUBLE, the most permissive type possible. The type was properly getting replaced for other file format readers, but not JSON.

tests/integ/scala/test_dataframe_reader_suite.py
sfc-gh-joshi: added

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aling: yes I'm a little concerned about whether there is information loss/mismatch during the bidirectional conversion `convert_sf_to_sp_type` <-> `convert_sp_to_sf_type`

could you double-check the logic to see the transformation is truly lossless?

I'm also good with guarding the change under use_relaxed_types

sfc-gh-joshi: I've added a guard with `use_relaxed_types`.

sfc-gh-mayliu (approved):

PR #4109, opened 2026-03-06
[SNOW-3203938] Fix ai_parse_document_basic test

sfc-gh-mrek:

   

   Fixes SNOW-3203938

With latest changes related to the new error handling, basic parse document tests started failing because metadata key is mising in the response.
This is expected behaviour, more details can be found [here](https://docs.google.com/document/d/1eG_1tUYopEnBGUBl5AWn26wd-CbgvCEzf2JBfJS9vYA/edit?tab=t.0). 
I've modified tests to use session parameter, to test both legacy and new response without error details (the bcr process lasts ~3 months, so during this time both combinations will be available for clients).
It affects ai_parse_document only, as for other functions the value didn't change. 
New response type is {value:<old response>, error}, but when response error details are disabled, value is extracted.
In case of ai_parse_document, metadata was moved to the root level of the response {value,metadata,error} making it absent in value.
As a follow-up, it would be a good idea to cover new error handling with error details in these tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
 
 - Fixed a bug in `Session.client_telemetry` that trace does not have snowflake style trace id.
 - Fixed a bug in `ai_complete` where `model_parameters` and `response_format` values containing single quotes would generate malformed SQL.
-- Fixed a ai_parse_document test that failed with new error handling
 
 ## 1.47.0 (2026-03-05)
 


tests/integ/test_dataframe_ai.py
```diff
@@ -1070,8 +1070,49 @@ def test_dataframe_ai_transcribe_default_output_column(session, resources_path):
...
+            assert data["metadata"].get("pageCount", 0) >= 3
+    finally:
+        session.sql(
+            "ALTER SESSION SET AI_SQL_ERROR_HANDLING_USE_FAIL_ON_ERROR = TRUE"
+        ).collect()
+
+
+def test_dataframe_ai_parse_document_basic_new_eh(session, resources_path):
+    """Test DataFrame.ai.parse_document OCR on a PDF document.
+    This is a test that covers legacy behaviour (post error handling changes, metadata is absent)."""
```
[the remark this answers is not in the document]
sfc-gh-mrek: The response with return_error_details set to false should still cover legacy behavior, so there's a new error handling but user expects not detailed response ({content...} vs {value:{content...},metadata:{..}, error:null})

tests/integ/test_dataframe_ai.py
```diff
@@ -1070,8 +1070,49 @@ def test_dataframe_ai_transcribe_default_output_column(session, resources_path):
...
 
 
-def test_dataframe_ai_parse_document_basic(session, resources_path):
-    """Test DataFrame.ai.parse_document OCR on a PDF document."""
+def test_dataframe_ai_parse_document_basic_legacy(session, resources_path):
+    """Test DataFrame.ai.parse_document OCR on a PDF document.
+    This test covers legacy behavior (post error handling changes, metadata is present in response)."""
+    session.sql(
+        "ALTER SESSION SET AI_SQL_ERROR_HANDLING_USE_FAIL_ON_ERROR = FALSE"
+    ).collect()
```
sfc-gh-aling, line 1078: we normally do not do this because our tests run in parallel and it might pollute other test

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - Fixed a bug in `Session.client_telemetry` that trace does not have snowflake style trace id.
 - Fixed a bug in `ai_complete` where `model_parameters` and `response_format` values containing single quotes would generate malformed SQL.
+- Fixed a ai_parse_document test that failed with new error handling
```
sfc-gh-joshi:
```suggestion
```

Test-only fixes shouldn't have changelog entries.

tests/integ/test_dataframe_ai.py
sfc-gh-mrek: So how can I control the behaviour based on the parameter?
For a bit we'll live in a world where user can have `AI_SQL_ERROR_HANDLING_USE_FAIL_ON_ERROR` either enabled or disabled. It changes the response format for the default ai_parse_document behavior (as i explained in linked doc, the metadata is moved out of the value object for AI_SQL_ERROR_HANDLING_USE_FAIL_ON_ERROR=false).
Also, as this parameter is currently opt in, based on the account cofiguration tests will either succeed or fail if we leave the metadata check.

Should there be an assumption that moving forward metadata won't be at top level so the check should be simply removed (even though for some accounts it can still be here?).
This doesn't look like a good change, but at the same time is the safest one if we can't use session params to fully control functions behavior :)

CHANGELOG.md
sfc-gh-mrek: Got it, the github check was red due to changelog entry missing, that's why i added it :)

sfc-gh-joshi: I've added the NO-CHANGELOG-UPDATES tag to this PR, which bypasses the github check.

Issue #4102, opened 2026-03-03
SNOW-3198369: "Statement provided can not be prepared" error when using session.sql to set a SQL variable without .collect()


sfc-gh-vtimofeenko:
1. What version of Python are you using?

   `Python 3.11.6 (v3.11.6:8b6ee5ba3b, Oct  2 2023, 11:18:21) [Clang 13.0.0 (clang-1300.0.29.30)]`

2. What operating system and processor architecture are you using?

   `macOS-26.3-arm64-arm-64bit`

3. What are the component versions in the environment (`pip freeze`)?

   ```
   uv pip freeze | egrep 'streamlit|snowflake|snowpark'
   snowflake-connector-python==4.3.0
   snowflake-snowpark-python==1.45.0
   streamlit==1.54.0
   ```

4. What did you do?

   I was working with Streamlit in Snowflake on SPCS examining how SQL session variables behave. I found that if my query is `SET foo = 'bar'` and I just `st.write` it (without calling `collect()`), I get an odd error "Statement provided can not be prepared."

   If I do `.collect()` it, there are no errors and the variable works.

   The relevant code is:  `st.write(session.sql("SET foo = 'bar'"))`, full code of the application is provided at the end. I reproduced this in SPCS and locally.

5. What did you expect to see?

   `st.write` generally materializes the `session.sql` result without the need to explicitly `.collect()` it. I would expect that the statement `st.write(session.sql("SET foo = 'bar'"))` would work without an error.

6. Can you set logging to DEBUG and collect the logs?

Debug logs attached.

[debug-logs-session-collect.log](https://github.com/user-attachments/files/25728534/debug-logs-session-collect.log)

<details>
<summary>Streamlit code</summary>

```python
"""Testing materialization of st.write with differently constructed Streamlit sessions."""

import logging

import snowflake.snowpark
import streamlit as st

for logger_name in ("snowflake.snowpark", "snowflake.connector"):
    logger = logging.getLogger(logger_name)
    logger.setLevel(logging.DEBUG)
    ch = logging.StreamHandler()
    ch.setLevel(logging.DEBUG)
    ch.setFormatter(
        logging.Formatter(
            "%(asctime)s - %(threadName)s %(filename)s:%(lineno)d - %(funcName)s() - %(levelname)s - %(message)s"
        )
    )
    logger.addHandler(ch)

st.caption(f"snowpark {snowflake.snowpark.__version__}")

[...]

PR #4115, opened 2026-03-10
SNOW-3198369: Fix exception from passing SQL SET calls to Streamlit `st.write`

sfc-gh-joshi:

   

   Fixes SNOW-3198369 (#4102)

Streamlit internally attempts to call `to_pandas` on dataframe-like objects passed to `st.write` ([source](https://github.com/streamlit/streamlit/blob/093814974d2dba8813414cdbb31e300259dc367a/lib/streamlit/dataframe_util.py#L673-L680)). This previously caused issues if the dataframe was created from a SQL `SET` statement, as our internal analyzer inappropriately attempted to issue a DESCRIBE query rather than understanding that the output would always contain a STATUS column.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/schema_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/schema_utils.py
@@ -77,7 +77,7 @@ def analyze_attributes(
     # SQL commands which cannot be prepared
     # https://docs.snowflake.com/en/user-guide/sql-prepare.html
     if lowercase.startswith(
-        ("alter", "drop", "use", "create", "grant", "revoke", "comment")
+        ("alter", "drop", "use", "create", "grant", "revoke", "comment", "set")
     ):
         return command_attributes()
     if lowercase.startswith(("ls", "list")):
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/_internal/analyzer/schema_utils.py
```diff
@@ -77,7 +77,7 @@ def analyze_attributes(
     # SQL commands which cannot be prepared
     # https://docs.snowflake.com/en/user-guide/sql-prepare.html
     if lowercase.startswith(
-        ("alter", "drop", "use", "create", "grant", "revoke", "comment")
+        ("alter", "drop", "use", "create", "grant", "revoke", "comment", "set")
```
sfc-gh-aling, line 80: are there more such cases that we should add here?

sfc-gh-joshi: I'm not sure. Based on the list of SQL commands: https://docs.snowflake.com/en/sql-reference/sql-all we may also need `undrop` and `truncate`, but I'm not sure if there's any full enumeration of commands that return a single "status" column.

I'm also a little suspicious of the `startswith` matching in this function, but since it seems like there aren't any command keywords that are substrings of others, it should be fine to keep as-is.

PR #4118, opened 2026-03-12
SNOW-3002469: Support INCLUDE_METADATA in copy_into_table

sfc-gh-bkogan:

   

   Fixes SNOW-3002469

   Adds support for the `INCLUDE_METADATA` copy option for `copy_into_table`. This specifies a list of file metadata columns to also include with the copied rows. [Docs here](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#copy-options-copyoptions).

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -104,6 +104,7 @@ JOIN = " JOIN "
 NATURAL = " NATURAL "
 ASOF = " ASOF "
 MATCH_CONDITION = " MATCH_CONDITION "
+DIRECTED_JOIN = " DIRECTED JOIN "
 EXISTS = " EXISTS "
 CREATE = " CREATE "
 TABLE = " TABLE "
@@ -976,6 +977,7 @@ def snowflake_supported_join_statement(
     use_constant_subquery_alias: bool,
     left_uuid: Optional[str] = None,
     right_uuid: Optional[str] = None,
+    directed: bool = False,
 ) -> str:
     LEFT_UUID = format_uuid(left_uuid)
     RIGHT_UUID = format_uuid(right_uuid)
@@ -1020,6 +1022,8 @@ def snowflake_supported_join_statement(
         (MATCH_CONDITION + match_condition) if match_condition else EMPTY_STRING
     )
 
+    maybe_directed_sql = DIRECTED_JOIN if directed else JOIN
+
     source = (
         LEFT_PARENTHESIS
         + NEW_LINE
@@ -1033,7 +1037,7 @@ def snowflake_supported_join_statement(
         + SPACE
         + NEW_LINE
         + join_sql
-        + JOIN
+        + maybe_directed_sql
         + NEW_LINE
         + LEFT_PARENTHESIS
         + NEW_LINE
@@ -1062,6 +1066,7 @@ def join_statement(
     use_constant_subquery_alias: bool,
     left_uuid: Optional[str] = None,
     right_uuid: Optional[str] = None,
+    directed: bool = False,
 ) -> str:
     if isinstance(join_type, (LeftSemi, LeftAnti)):
         return left_semi_or_anti_join_statement(
@@ -1088,6 +1093,7 @@ def join_statement(
         use_constant_subquery_alias,
         left_uuid=left_uuid,
         right_uuid=right_uuid,
+        directed=directed,
     )
 
 
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -3449,6 +3449,8 @@ class DataFrame:
         self,
         right: "DataFrame",
         how: Optional[str] = None,
+        *,
+        directed: bool = False,
         _emit_ast: bool = True,
         **kwargs,
     ) -> "DataFrame":
@@ -3467,6 +3469,7 @@ class DataFrame:
                 You can also use ``join_type`` keyword to specify this condition.
                 Note that to avoid breaking changes, currently when ``join_type`` is specified,
                 it overrides ``how``.
+            directed: Whether the join is a directed join, which forces the left argument to be scanned before the right.
 
         Examples::
             >>> df1 = session.create_dataframe([[1, 2], [3, 4], [5, 6]], schema=["a", "b"])
@@ -3499,6 +3502,7 @@ class DataFrame:
             NaturalJoin(join_type),
             None,
             None,
+            directed,
         )
         stmt = None
         if _emit_ast:
@@ -3516,6 +3520,7 @@ class DataFrame:
                 ast.join_type.join_type__full_outer = True
             else:
                 raise ValueError(f"Unsupported join type {join_type}")
+            ast.directed.value = directed
 
         if self._select_statement:
             select_plan = self._session._analyzer.create_select_statement(
@@ -3588,6 +3593,7 @@ class DataFrame:
             lateral_join_type,
             on_expr,
             None,
+            False,
         )
 
         stmt = None
@@ -3625,6 +3631,7 @@ class DataFrame:
         lsuffix: str = "",
         rsuffix: str = "",
         match_condition: Optional[Column] = None,
+        directed: bool = False,
         _emit_ast: bool = True,
         **kwargs,
     ) -> "DataFrame":
@@ -3655,6 +3662,7 @@ class DataFrame:
             lsuffix: Suffix to add to the overlapping columns of the left DataFrame.
             rsuffix: Suffix to add to the overlapping columns of the right DataFrame.
             match_condition: The match condition for asof join.
+            directed: Whether the join is a directed join, which forces the left argument to be scanned before the right.
 
         Note:
             When both ``lsuffix`` and ``rsuffix`` are empty, the overlapping columns will have random column names in the resulting DataFrame.
@@ -3967,6 +3975,7 @@ class DataFrame:
                     ast.join_type.join_type__asof = True
                 else:
                     raise ValueError(f"Unsupported join type {join_type_arg}")
+                ast.directed.value = directed
 
                 join_cols = kwargs.get("using_columns", on)
                 if join_cols is not None:
@@ -3999,6 +4008,7 @@ class DataFrame:
                 lsuffix=lsuffix,
                 rsuffix=rsuffix,
                 match_condition=match_condition,
+                directed=directed,
                 _ast_stmt=stmt,
             )
 
@@ -4174,6 +4184,7 @@ class DataFrame:
         *,
         lsuffix: str = "",
         rsuffix: str = "",
+        directed: bool = False,
         _emit_ast: bool = True,
     ) -> "DataFrame":
         """Performs a cross join, which returns the Cartesian product of the current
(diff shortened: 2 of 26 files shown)


src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -1357,11 +1357,13 @@ def file_operation_statement(
...
 
 def convert_value_to_sql_option(
-    value: Optional[Union[str, bool, int, float, list, tuple]],
+    value: Optional[Union[str, bool, int, float, list, tuple, dict]],
     parse_none_as_string: bool = False,
 ) -> str:
     if value is None and parse_none_as_string:
         value = str(value)
+    if isinstance(value, dict):
+        return f"({', '.join(f'{k} = {v}' for k, v in value.items())})"
```
sfc-gh-joshi, line 1372: Will this break if the input contains special characters like commas/semicolons? I'm not sure if this is a concern for the other argument types as well.

src/snowflake/snowpark/dataframe.py
```diff
@@ -4864,6 +4869,25 @@ def copy_into_table(
             else None
         )
         copy_options = copy_options or reader_copy_options
+
+        if (
+            "INCLUDE_METADATA" in copy_options
+            and copy_options["INCLUDE_METADATA"] is not None
```
sfc-gh-joshi: nit: `copy_options.get("INCLUDE_METADATA", None) is not None`

src/snowflake/snowpark/dataframe.py
```diff
@@ -4864,6 +4869,25 @@ def copy_into_table(
...
+        ):
+            for metadata_col in copy_options["INCLUDE_METADATA"].values():
+                if quote_name(metadata_col) not in METADATA_COLUMN_TYPES:
+                    raise ValueError(
+                        f"Metadata column {metadata_col} is not supported. Supported {list(METADATA_COLUMN_TYPES.keys())}"
```
sfc-gh-joshi:
```suggestion
                        f"Metadata column {metadata_col} is not supported. Supported columns: {list(METADATA_COLUMN_TYPES.keys())}"
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -4864,6 +4869,25 @@ def copy_into_table(
...
+                    )
+
+            if "MATCH_BY_COLUMN_NAME" not in copy_options:
+                raise ValueError(
+                    "INCLUDE_METADATA can only be used with the MATCH_BY_COLUMN_NAME copy option."
+                )
+            if self._reader._file_type and self._reader._file_type.upper() == "CSV":
+                format_type_options = format_type_options or {}
```
sfc-gh-joshi: I'm not sure how consistent we are about this across the codebase, but `format_type_options` is provided by the user, so we should be careful about mutating it.
```suggestion
                format_type_options = {...format_type_options} if format_type_options else {}
```

sfc-gh-bkogan: good point

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-bkogan: It will, for example

```
df.copy_into_table(
    table_name,
    MATCH_BY_COLUMN_NAME="CASE_INSENSITIVE",
    INCLUDE_METADATA={
        "filename;_col": "METADATA$FILENAME",
        "ROW_NUM_COL": "METADATA$FILE_ROW_NUMBER",
    },
    force=True,
)
```

will return something like this:

```
syntax error line 1 at position 267 unexpected ';'
```

But is there something we can do about it? Users can wrap the columns in quotes to escape these, but if they don't, it seems like they should fix?

sfc-gh-joshi: I'm not sure whether forcing users to quote column names is intuitive (especially since the error message isn't very clear), but it shouldn't be the end of the world.

I noticed that you do check `quote_name(metadata_col)` against the list of valid metadata column names at the dataframe layer. Since that's the case, should we be quoting the column names here as well?

sfc-gh-bkogan: We can quote the the values, I just tested and `'METADATA$FILENAME'` and `'"METADATA$FILENAME"'` both work. However, the keys are the target table column names. If we wrap those in quotes, then they become case sensitive, which might break users who input them as lower case for example.

sfc-gh-joshi: Just so I'm understanding correctly, on line 4873-4878 of dataframe.py you have
```
        if copy_options.get("INCLUDE_METADATA", None) is not None:
            for metadata_col in copy_options["INCLUDE_METADATA"].values():
                if quote_name(metadata_col) not in METADATA_COLUMN_TYPES:
                    raise ValueError(
                        f"Metadata column {metadata_col} is not supported. Supported columns: {list(METADATA_COLUMN_TYPES.keys())}"
                    )
```

Doesn't this already make the function case-sensitive, since this check would raise if a user were to provide something like all-lowercase `'metadata$filename'`?

Either way, we should add a test case validating this behavior.

sfc-gh-bkogan: Turns out `quote_name` was already doing a .upper() call when it was not already wrapped in quotes. But I changed it to use .upper() on the caller side so it works in the already quoted case, i.e.:

```
if quote_name(metadata_col.upper()) not in METADATA_COLUMN_TYPES:
```

Modified one of the test columns to make sure it works case insensitively and with quotes.

sfc-gh-joshi (approved):

src/snowflake/snowpark/dataframe.py
```diff
@@ -4864,6 +4869,24 @@ def copy_into_table(
...
+            for metadata_col in copy_options["INCLUDE_METADATA"].values():
+                if quote_name(metadata_col.upper()) not in METADATA_COLUMN_TYPES:
+                    raise ValueError(
+                        f"Metadata column {metadata_col} is not supported. Supported columns: {list(METADATA_COLUMN_TYPES.keys())}"
+                    )
+
+            if "MATCH_BY_COLUMN_NAME" not in copy_options:
+                raise ValueError(
+                    "INCLUDE_METADATA can only be used with the MATCH_BY_COLUMN_NAME copy option."
+                )
```
sfc-gh-aling, line 4903: should we let the server side do the parameter validation? or this is an existing pattern in the library

src/snowflake/snowpark/dataframe.py
```diff
@@ -4864,6 +4869,24 @@ def copy_into_table(
...
+            if self._reader._file_type and self._reader._file_type.upper() == "CSV":
+                format_type_options = (
+                    format_type_options.copy() if format_type_options else {}
+                )
+                format_type_options["ERROR_ON_COLUMN_COUNT_MISMATCH"] = False
```
sfc-gh-aling, line 4912: why we set it to False here?

src/snowflake/snowpark/dataframe.py
```diff
@@ -4864,6 +4869,24 @@ def copy_into_table(
             else None
         )
         copy_options = copy_options or reader_copy_options
+
+        if copy_options.get("INCLUDE_METADATA", None) is not None:
```
sfc-gh-aling, line 4894: should the parameter name be lower cased to keep the styling consistent with other params?

src/snowflake/snowpark/dataframe.py
sfc-gh-bkogan: According to cursor, there are existing patterns in the lib (DataFrameReader._extract_time_travel_from_options, JDBC reader in dataframe_reader.py:1641–1644, DataFrameWriter.save_as_table validates overwrite_condition, etc)

My thought is it would be better to fail as quickly as possible and avoid any network delay

src/snowflake/snowpark/dataframe.py
sfc-gh-bkogan: In the documentation, it says:

```
When you load a CSV file with INCLUDE_METADATA, set the file format option ERROR_ON_COLUMN_COUNT_MISMATCH to FALSE.
```

Since adding metadata columns definitionally adds more columns than there are in the CSV.

src/snowflake/snowpark/dataframe.py
sfc-gh-bkogan: all the copy options are currently uppercased:

utils.py 173
```
COPY_INTO_TABLE_COPY_OPTIONS = {
    "ON_ERROR",
    "SIZE_LIMIT",
    "PURGE",
    "RETURN_FAILED_ONLY",
    "MATCH_BY_COLUMN_NAME",
    "ENFORCE_LENGTH",
    "TRUNCATECOLUMNS",
    "FORCE",
    "LOAD_UNCERTAIN_FILES",
    "INCLUDE_METADATA",
}
```

src/snowflake/snowpark/dataframe.py
sfc-gh-aling: if user accidentally sets it to True, should we raise error explicitly instead of silently overwriting the value internally?

src/snowflake/snowpark/dataframe.py
sfc-gh-aling: could you verify if our copy_options key lookups are exact matches (uppercase) or case-insensitive? Just want to confirm the logic in the current codebase.

src/snowflake/snowpark/dataframe.py
sfc-gh-bkogan: Hmm, not sure. Is there a best practice around this? Both approaches seem valid to me.

src/snowflake/snowpark/dataframe.py
sfc-gh-bkogan: This is the relevant helper:

```
def _get_options(
    options: Dict[str, Any], allowed_options: Set[str]
) -> Tuple[Dict[str, Any], Dict[str, Any]]:
    """Helper method that extracts common logic for getting options for
    COPY INTO TABLE and COPY INTO LOCATION command.
    """
    file_format_type_options = options.get("FORMAT_TYPE_OPTIONS", {})
    copy_options = options.get("COPY_OPTIONS", {})
    for k, v in options.items():
        if k in allowed_options:
            copy_options[k] = v
        elif k not in NON_FORMAT_TYPE_OPTIONS:
            file_format_type_options[k] = v
    return file_format_type_options, copy_options
```

allowed_options is a string set, and we check directly against it, so it seems it's case sensitive currently.

src/snowflake/snowpark/dataframe.py
sfc-gh-aling: my preference here is to raise error being explicit about the expected behavior

sfc-gh-bkogan: And if the user does not set it at all, should we silently set it? I feel like users will forget about that option so it would be better dev experience.

sfc-gh-aling: yes, I agree upon improving the default happy path experience for developers if they forget to set.

so the behavior would be like the following

- if not set by user, we set for user
- if set to False, do nothing
- if set to True, raise error

sfc-gh-bkogan: Got it

PR #4119, opened 2026-03-12
SNOW-3230493: Explicitly query CURRENT_ACCOUNT in default_artifact_repository retrieval path

sfc-gh-joshi:

   

   Fixes SNOW-3230493

A customer encountered this issue when creating a sproc, where a query like `SYSTEM$GET_DEFAULT_PYTHON_ARTIFACT_REPOSITORY('3.13', 'account', '"sfctest0"')` was raising an error because the account name was double-quoted and all lowercased. This would result in the exception being caught and falling back to the default conda repository path. Copying my analysis from the parent ticket:

- If the user explicitly passes artifact_repository=”snowflake.snowpark.pypi_shared_repository” to the sproc function, then SYSTEM$GET_DEFAULT_PYTHON_ARTIFACT_REPOSITORY should never get called.

- Setting an active database or schema on the Snowpark session should be a viable workaround: these also get double-quoted by the Snowpark Python client, but the system function is able to accept these properly. This is because of the artifact_repository parameter was not explicitly passed to sproc, the client will check the session’s active schema for a schema-level default artifact repository. If there is no active schema, then it will fall back to the active database; if there is no active database, it will fall back to the account (as seems to be the case here).

- ~~In addition to the double-quoting issue mentioned in the ticket, for some reason the Snowpark Python client quotes a lowercase version of the account name, while schema and database names are uppercased before they’re quoted. Passing an uppercase double-quoted value (whether it’s an account locator, schema, or database name) is fine, but making it lowercase causes the observed issue. Removing the double quotes makes the identifiers case-insensitive, which is the fix I have in-progress.~~

The issue occurs because Snowpark reads the connector's underlying `_account` property, which is set by user configuration file and not updated by queries (unlike database/schema). This field is treated as case-insensitive, but when Snowpark reads it in this case, quoting the field makes it case-sensitive, and breaks expected behavior. To remedy this, we now issue an explicit `SELECT CURRENT_ACCOUNT()` as a workaround.

Note that un-quoting the identifier is not an appropriate fix, because account names can contain hyphens, which are not allowed in unquoted identifiers and will raise a syntax error.

Account identifier rules: https://docs.snowflake.com/en/user-guide/admin-account-identifier
Unquoted object identifier rules: https://docs.snowflake.com/en/sql-reference/identifiers-syntax#label-identifier-casing

### DIFF
--- a/tests/integ/modin/frame/test_apply.py
+++ b/tests/integ/modin/frame/test_apply.py
@@ -245,7 +245,7 @@ def test_frame_with_timedelta_index():
 def test_axis_1_date_time_timestamp_type(data, func, expected_result):
     snow_df = pd.DataFrame(data)
     result = snow_df.apply(func, axis=1)
-    assert_snowpark_pandas_equal_to_pandas(result, expected_result)
+    assert_snowpark_pandas_equal_to_pandas(result, expected_result, check_dtype=False)
 
 
 @sql_count_checker(query_count=5, join_count=2, udtf_count=1)
@@ -330,7 +330,8 @@ def test_axis_1_return_not_json_serializable_label():
         ).to_pandas()
 
     with pytest.raises(
-        SnowparkSQLException, match="Object of type DataFrame is not serializable"
+        SnowparkSQLException,
+        match=r"Object of type DataFrame is not (JSON )?serializable",
     ):
         # return value
         snow_df.apply(lambda x: native_pd.DataFrame([1, 2]), axis=1).to_pandas()
--- a/tests/integ/modin/series/test_apply_and_map.py
+++ b/tests/integ/modin/series/test_apply_and_map.py
@@ -401,7 +401,7 @@ class TestApplyOrMapCallable:
         # In Python UDF, if the return type is variant, the return value must be
         # json serializable so it can become a variant in Snowflake.
         # type() returns a type object which is not json serializable.
-        with pytest.raises(SnowparkSQLException, match="is not serializable"):
+        with pytest.raises(SnowparkSQLException, match=r"is not (JSON )?serializable"):
             getattr(snow_series, method)(type).to_pandas()
 
     @pytest.mark.parametrize("func", [str, int, float, bytes, list, dict])
--- a/tests/integ/scala/test_udf_suite.py
+++ b/tests/integ/scala/test_udf_suite.py
@@ -567,7 +567,7 @@ def test_geometry_type(session):
     table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
     Utils.create_table(session, table_name, "g geometry", is_temporary=True)
     session._run_query(
-        f"insert into {table_name} values ('POINT(30 10)'), ('POINT(50 60)'), (null)"
+        f"insert into {table_name} values ('POINT(30.5 10.5)'), ('POINT(50.5 60.5)'), (null)"
     )
     df = session.table(table_name)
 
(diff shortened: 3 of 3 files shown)


src/snowflake/snowpark/session.py
```diff
@@ -2424,9 +2424,9 @@ def _get_default_artifact_repository(self) -> str:
             if isinstance(self._conn, MockServerConnection):
                 return _DEFAULT_ARTIFACT_REPOSITORY
 
-            account = self.get_current_account()
-            database = self.get_current_database()
-            schema = self.get_current_schema()
+            account = self._conn._get_current_parameter("account", quoted=False)
+            database = self._conn._get_current_parameter("database", quoted=False)
+            schema = self._conn._get_current_parameter("schema", quoted=False)
```
sfc-gh-aling: would the change break cases that we need quotes for the identifier (case sensitive)?
can we test identifiers w/, w/o double quotes that contains special characters as long as they are allowed by snowflake?

you can follow up with @sfc-gh-bkogan on it

tests/integ/test_session.py
```diff
@@ -1091,3 +1093,27 @@ def test_get_active_sessions_empty():
 
     with patch.object(session_module, "_active_sessions", return_value=set()):
         assert session_module._get_active_sessions(require_at_least_one=False) == set()
+
+
+def test_default_artifact_repository_with_no_db_schema(session, caplog):
```
sfc-gh-aling, line 1098: do we have tests for different forms (regarding double quotes, case sensitivity) of identifiers?

sfc-gh-joshi: After reading more closely about the distinction between quoted and unquoted identifiers, I've decided to revert the changes to database and schema (they now remain quoted as before). I've also decided to explicitly issue a query for the account name rather than reading the connector's `_account` property to ensure we get correct case-sensitivity behavior. The `sfctest0` account is actually sufficient to trigger the original bug, since the old implementation was producing quoted `"sfctest0"` rather than all-caps `"SFCTEST0"`, which is why I think this test case is sufficient.

I don't really think we can introduce meaningful test cases exercising different quoting/casing rules on the account name in particular: I don't think it's possible to change the account within a test runner, and if we were to mock the return value of `SELECT CURRENT_ACCOUNT()`, we would also have to either somehow create fake accounts with that name to ensure `SYSTEM$GET_DEFAULT_PYTHON_ARTIFACT_REPOSITORY` returns an appropriate value, or mock the response of the system function, at which point we aren't really testing something meaningful.

src/snowflake/snowpark/session.py
```diff
@@ -2424,7 +2427,11 @@ def _get_default_artifact_repository(self) -> str:
             if isinstance(self._conn, MockServerConnection):
                 return _DEFAULT_ARTIFACT_REPOSITORY
 
-            account = self.get_current_account()
+            # self.get_current_account uses a cached connector field that may not be properly cased, so we need to
+            # explicitly issue a query for it
+            account = quote_name_without_upper_casing(
+                self._conn._get_string_datum("SELECT CURRENT_ACCOUNT()")
+            )
```
sfc-gh-bkogan: What's the latency of this? Since we are doing it before the cache check, it's incurred on every call. Can we optimize this where if both a database and schema already exist, we skip calling this (since we won't need it).

sfc-gh-aling: why current_database and current_schema don't need the same change? are they not having the same issue as account?

if we are issuing a query, we can get all the info in one round-trip: `select current_account(), current_database(), current_schema()`

tests/integ/test_session.py
sfc-gh-aling: yeah account level test would be difficult.
I'm good if the current test is sufficient

src/snowflake/snowpark/session.py
sfc-gh-joshi: @sfc-gh-bkogan Good point, I've moved the check.

@sfc-gh-aling Our `Session.get_current_***` methods read from a connector property (_account, _database, _schema), and fall back to an explicit query if those are unset. _database and _schema apparently get cased correctly, while _account does not. Summarizing from Cursor:

> The `_account` field is set once from whatever the user provides in kwargs (line 1677 of connection.py, via setattr(self, "_" + name, value)) and only further processed by parse_account to strip region/domain suffixes (line 1829). It is never normalized to uppercase and never refreshed from the server.
> This contrasts with how `_database`, `_schema`, `_warehouse`, and `_role` are handled. Those fields are updated in two places:
> 1. At login time -- from the sessionInfo in the authentication response:
> 2. After every query -- from the final*Name fields in the query response:
> Notice that _account is conspicuously absent from both locations. The only place `_account` is ever written is during `__config`:
> ...which stores the raw user-provided string (e.g., "myaccount" lowercase).

I'm not sure if this should be considered a connector bug, since `_account` is an underscore property and not technically part of their public interface.

PR #4123, opened 2026-03-14
SNOW-3192256: Support XML infer schema

sfc-gh-mayliu:

   

   Fixes SNOW-3192256

Published PR to trigger merge gate for test results, but this PR is still a WIP. Will add google doc link to design doc.

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -10,6 +10,7 @@ import random
 import re
 import sys
 from collections import Counter
+from decimal import Decimal
 from functools import cached_property
 from logging import getLogger
 from types import ModuleType
@@ -5283,6 +5284,8 @@ class DataFrame:
                 start_field = getattr(datatype, "start_field", DayTimeIntervalType.DAY)
                 end_field = getattr(datatype, "end_field", DayTimeIntervalType.SECOND)
                 res = format_day_time_interval_for_display(cell, start_field, end_field)
+            elif isinstance(cell, Decimal):
+                res = format(cell, "f")
             else:
                 res = str(cell)
             return res.replace("\n", "\\n")
--- a/tests/integ/modin/frame/test_apply.py
+++ b/tests/integ/modin/frame/test_apply.py
@@ -245,7 +245,7 @@ def test_frame_with_timedelta_index():
 def test_axis_1_date_time_timestamp_type(data, func, expected_result):
     snow_df = pd.DataFrame(data)
     result = snow_df.apply(func, axis=1)
-    assert_snowpark_pandas_equal_to_pandas(result, expected_result)
+    assert_snowpark_pandas_equal_to_pandas(result, expected_result, check_dtype=False)
 
 
 @sql_count_checker(query_count=5, join_count=2, udtf_count=1)
@@ -330,7 +330,8 @@ def test_axis_1_return_not_json_serializable_label():
         ).to_pandas()
 
     with pytest.raises(
-        SnowparkSQLException, match="Object of type DataFrame is not serializable"
+        SnowparkSQLException,
+        match=r"Object of type DataFrame is not (JSON )?serializable",
     ):
         # return value
         snow_df.apply(lambda x: native_pd.DataFrame([1, 2]), axis=1).to_pandas()
--- a/tests/integ/modin/series/test_apply_and_map.py
+++ b/tests/integ/modin/series/test_apply_and_map.py
@@ -401,7 +401,7 @@ class TestApplyOrMapCallable:
         # In Python UDF, if the return type is variant, the return value must be
         # json serializable so it can become a variant in Snowflake.
         # type() returns a type object which is not json serializable.
-        with pytest.raises(SnowparkSQLException, match="is not serializable"):
+        with pytest.raises(SnowparkSQLException, match=r"is not (JSON )?serializable"):
             getattr(snow_series, method)(type).to_pandas()
 
     @pytest.mark.parametrize("func", [str, int, float, bytes, list, dict])
--- a/tests/integ/scala/test_udf_suite.py
+++ b/tests/integ/scala/test_udf_suite.py
@@ -567,7 +567,7 @@ def test_geometry_type(session):
     table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
     Utils.create_table(session, table_name, "g geometry", is_temporary=True)
     session._run_query(
-        f"insert into {table_name} values ('POINT(30 10)'), ('POINT(50 60)'), (null)"
+        f"insert into {table_name} values ('POINT(30.5 10.5)'), ('POINT(50.5 60.5)'), (null)"
     )
     df = session.table(table_name)
 
@@ -576,18 +576,18 @@ def test_geometry_type(session):
             return None
         else:
             g_str = str(g)
-            if "[50, 60]" in g_str and "Point" in g_str:
+            if "[50.5, 60.5]" in g_str and "Point" in g_str:
                 return g_str
             else:
-                return g_str.replace("0", "")
+                return g_str.replace(".5", ".2")
 
     geometry_udf = udf(geometry, return_type=StringType(), input_types=[GeometryType()])
 
     Utils.check_answer(
         df.select(geometry_udf(col("g"))),
         [
-            Row("{'coordinates': [3, 1], 'type': 'Point'}"),
-            Row("{'coordinates': [50, 60], 'type': 'Point'}"),
+            Row("{'coordinates': [30.2, 10.2], 'type': 'Point'}"),
+            Row("{'coordinates': [50.5, 60.5], 'type': 'Point'}"),
             Row(None),
         ],
     )
--- a/tests/integ/test_dataframe.py
+++ b/tests/integ/test_dataframe.py
@@ -3480,6 +3480,461 @@ def test_show_interval_formatting(session):
     )
 
 
+_SKIP_LOCAL_TESTING_DECIMAL = pytest.mark.skipif(
+    "config.getoption('local_testing_mode', default=False)",
+    reason="Local testing mock does not normalize Decimal scale to match DecimalType",
+    run=False,
+)
+
+
+@_SKIP_LOCAL_TESTING_DECIMAL
+@pytest.mark.parametrize(
+    "schema, data, expected",
+    [
+        pytest.param(
+            StructType([StructField("VAL", DecimalType(38, 15))]),
+            [
+                [Decimal("0.000000000000001")],
+                [Decimal("-0.000000000000001")],
+                [Decimal("-0.000000000000001")],
+            ],
+            """\
+            +------------------+
+            |"VAL"             |
+            +------------------+
+            |0.000000000000001 |
+            |-0.000000000000001|
+            |-0.000000000000001|
+            +------------------+
+            """,
+            id="positive_and_negative_small_p38_s15",
+        ),
+        pytest.param(
+            StructType([StructField("VAL", DecimalType(38, 30))]),
+            [
+                [Decimal("1E-30")],
+                [Decimal("9.999999999E-15")],
+                [Decimal("1.23456789012345E-10")],
+            ],
+            """\
+            +--------------------------------+
+            |"VAL"                           |
+            +--------------------------------+
+            |0.000000000000000000000000000001|
+            |0.000000000000009999999999000000|
+            |0.000000000123456789012345000000|
+            +--------------------------------+
+            """,
+            id="extreme_precision_p38_s30",
+        ),
+        pytest.param(
+            StructType([StructField("VAL", DecimalType(38, 0))]),
+            [
+                [Decimal("1000000000000000000000000000000")],
+                [Decimal("-9999999999000000")],
+                [Decimal("12345678901")],
+            ],
+            """\
+            +-------------------------------+
+            |"VAL"                          |
+            +-------------------------------+
+            |1000000000000000000000000000000|
+            |-9999999999000000              |
+            |12345678901                    |
+            +-------------------------------+
+            """,
+            id="large_whole_numbers_p38_s0",
+        ),
+        pytest.param(
+            StructType([StructField("VAL", DecimalType(38, 10))]),
+            [
+                [Decimal("123.456")],
+                [Decimal("-987.654321")],
+                [Decimal("0")],
+                [Decimal("0.0000000001")],
+                [Decimal("-0.0000000001")],
+            ],
+            """\
+            +---------------+
+            |"VAL"          |
+            +---------------+
+            |123.4560000000 |
+            |-987.6543210000|
+            |0.0000000000   |
+            |0.0000000001   |
+            |-0.0000000001  |
+            +---------------+
+            """,
+            id="mixed_positive_negative_zero_p38_s10",
+        ),
+        pytest.param(
+            StructType([StructField("VAL", DecimalType(38, 20))]),
+            [
+                [Decimal("0.1")],
+                [Decimal("0.00001")],
+                [Decimal("0.0000000001")],
+                [Decimal("0.000000000000001")],
+                [Decimal("10")],
+                [Decimal("100000")],
+                [Decimal("10000000000")],
+            ],
+            """\
+            +--------------------------------+
+            |"VAL"                           |
+            +--------------------------------+
+            |0.10000000000000000000          |
+            |0.00001000000000000000          |
+            |0.00000000010000000000          |
+            |0.00000000000000100000          |
+            |10.00000000000000000000         |
+            |100000.00000000000000000000     |
+            |10000000000.00000000000000000000|
+            +--------------------------------+
+            """,
+            id="wide_magnitude_range_p38_s20",
+        ),
+        pytest.param(
+            StructType([StructField("VAL", DecimalType(38, 0))]),
+            [
+                [Decimal("1E+5")],
+                [Decimal("2.5E+8")],
+                [Decimal("1.25E+15")],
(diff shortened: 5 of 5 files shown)


src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -102,13 +114,97 @@ def _restore_colons_in_template(template: Optional[dict]) -> Optional[dict]:
...
+    if isinstance(target_type, TimestampType):
+        try:
+            datetime.datetime.fromisoformat(value)
+            return True
+        except (ValueError, TypeError):
+            return False
+    return True
+
+
+def _validate_row_for_type_mismatch(
```
sfc-gh-yuwang, line 160: Is this function being called per row after each call of `element_to_dict_or_str` ? I am a little concerning that whether this is going to impact the performance. Do you think it is possible to put this logic into `element_to_dict_or_str` so that we don't have to traverse each element of the row again?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1070,19 +1093,53 @@ def xml(self, path: str, _emit_ast: bool = True) -> DataFrame:
...
+        mode = self._cur_options.get("MODE", "PERMISSIVE").upper()
+        corrupt_col_name = self._cur_options.get(
+            "COLUMNNAMEOFCORRUPTRECORD", "_corrupt_record"
+        )
+        if mode == "PERMISSIVE":
+            df_columns = {c.strip('"').strip("'") for c in df.columns}
+            if corrupt_col_name in df_columns:
+                corrupt_ref = df[single_quote(corrupt_col_name)]
+                cols.append(
+                    corrupt_ref.cast(StringType()).alias(
```
sfc-gh-aling, line 1119: what's the type for the column `_corrupt_record`? -- I'm wondering 1) if we need cast 2) if cast is always needed can the string conversion logic be moved into the UDTF?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1368,6 +1425,169 @@ def _get_schema_from_user_input(
         read_file_transformations = [t._expression.sql for t in transformations]
         return new_schema, schema_to_cast, read_file_transformations
 
+    def _resolve_xml_file_for_udtf(self, local_file_path: str) -> str:
+        """Return the UDTF file path, uploading to a temp stage in stored procedures."""
+        if is_in_stored_procedure():  # pragma: no cover
+            temp_stage = random_name_for_temp_object(TempObjectType.STAGE)
```
sfc-gh-aling: why we need an upload here? is it because the server doesn't have the latest implementation?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1368,6 +1425,169 @@ def _get_schema_from_user_input(
...
+            sql_create_temp_stage = (
+                f"create temp stage if not exists {temp_stage} {XML_READER_SQL_COMMENT}"
+            )
+            self._session.sql(sql_create_temp_stage, _emit_ast=False).collect(
```
sfc-gh-aling: snowpark session has a `get_session_stage` method which creates a temp stage for the session, are we  able to reuse that?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1368,6 +1425,169 @@ def _get_schema_from_user_input(
...
+        try:
+            file_size = int(
+                self._session.sql(f"ls {path}", _emit_ast=False).collect(
+                    _emit_ast=False
+                )[0]["size"]
+            )
+        except IndexError:
+            raise ValueError(f"{path} does not exist")
+
+        num_workers = min(16, file_size // DEFAULT_CHUNK_SIZE + 1)
```
sfc-gh-aling, line 1474: is the `num_workers` config inheriting from the ingestion logic?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1368,6 +1425,169 @@ def _get_schema_from_user_input(
...
+
+        for row in results:
+            schema_str = row[0]
+            if schema_str is None or schema_str == "":
+                continue
+            try:
+                partial_schema = type_string_to_type_object(schema_str)
+            except Exception:
+                continue
+            if not isinstance(partial_schema, StructType):
```
sfc-gh-aling, line 1524: what's the case that a partial_schema is returned?

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -102,13 +114,97 @@ def _restore_colons_in_template(template: Optional[dict]) -> Optional[dict]:
...
+        try:
+            float(value)
+            return True
+        except ValueError:
+            return False
+    if isinstance(target_type, BooleanType):
+        return value.lower() in ("true", "false", "1", "0")
+    if isinstance(target_type, DateType):
+        try:
+            datetime.date.fromisoformat(value)
```
sfc-gh-aling, line 147: I'm wondering if there are input of different format that doesn't follow isoformat but are valid input to spark system.

there is a `python-dateutil` lib which supports more format

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -102,13 +114,97 @@ def _restore_colons_in_template(template: Optional[dict]) -> Optional[dict]:
...
     schema = type_string_to_type_object(safe_string)
     template = struct_type_to_result_template(schema)
-    return _restore_colons_in_template(template)
+    return _restore_colons_in_template(template), schema
+
+
+def _can_cast_to_type(value: str, target_type: DataType) -> bool:
+    if isinstance(target_type, StringType):
+        return True
+    if isinstance(target_type, LongType):
```
sfc-gh-aling, line 131: does this cover all types -- I don't see decimal type here, so does it mean we don't need to handle the decimal type here

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -607,10 +716,22 @@ def process_xml_range(
...
+                    # Mode handling for type mismatch errors.
+                    row = _validate_row_for_type_mismatch(
+                        row,
+                        schema_type,
+                        mode,
+                        record_str,
+                        column_name_of_corrupt_record,
+                    )
+
+                if row is not None:
```
sfc-gh-aling, line 734: I want to confirm
1) the behavior in spark that a mismatch row is skipped instead of being presented as a null row?
2) does it change the current snowpark behavior?

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -690,5 +813,6 @@ def process(
             ignore_surrounding_whitespace,
             row_validation_xsd_path=row_validation_xsd_path,
             result_template=result_template,
+            schema_type=schema_type,
```
sfc-gh-aling, line 820: does this change the existing behavior in snowpark?
previously we don't do `_validate_row_for_type_mismatch`, is it always doing `_validate_row_for_type_mismatch` now?

src/snowflake/snowpark/_internal/xml_schema_inference.py
```diff
@@ -0,0 +1,711 @@
...
+    except (ValueError, OverflowError):
+        pass
+
+    # Boolean
+    if text.lower() in ("true", "false"):
+        return BooleanType()
+
+    # Date (Spark default pattern: yyyy-MM-dd via DateFormatter)
+    try:
+        date.fromisoformat(text)
```
sfc-gh-aling, line 110: same question should we use `python-dateutil` here

src/snowflake/snowpark/_internal/xml_schema_inference.py
```diff
@@ -0,0 +1,711 @@
...
+    """Normalize text by stripping whitespace if configured."""
+    if text is None:
+        return None
+    return text.strip() if ignore_surrounding_whitespace else text
+
+
+def _infer_primitive_type(text: str) -> DataType:
+    """
+    Infer the DataType from a single string value with below priority order:
+    - null/empty -> NullType
```
sfc-gh-aling, line 63: same question on decimal type support

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: `_corrupt_record` will always be stringType, and `xml_reader.py` UDTF always returns `VARIANT` for all columns including `_corrupt_record`. If user wants XML to return typed columns (guarded by `if effective_schema is not None`), the cast has to happen in `dataframe_reader.py`, can't be moved into UDTF.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: The upload is for sproc execution, which needs the UDTF file to be on stage for the UDTF `register_from_file` call to work later. Local execution does not need file upload.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: Yes, we can, I can reuse it in the next commit

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: No, it's XML-specific. [Previous perf benchmark](https://docs.google.com/document/d/11J2bL6QTpDs0DnMJZAq5EUNUCdipB9kfzojdlX_sCuU/edit?tab=t.0#heading=h.kvp48qund8z0) showed UDTF parallelism plateau at 16 workers, so this is inspired by the original Snowpark UDTF design here https://github.com/snowflakedb/snowpark-python/blob/1eb18b328e6c7caf2fbfae40ddb6d552cec9d097/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L1859

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: `results` will contain up to 16 rows, each mapping to the schema inferred from uniformly split byte range in file. `partial_schema` captures the schema from each of the 16 workers, and individually merge/widen the previously `merged_schema`.

If you were wondering when the exception happens -- realistically the exception should never fire, since `xml_schema_inference.py` handles types in a tight roundtrip (`_case_preserving_simple_string` and `type_string_to_type_object`). This is a defensive guard to ensure one bad worker result doesn't crash the entire schema inference flow.

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-mayliu: Spark's date and timestamp parser is strict, which looks for patterns like `yyyy-mm-dd` and the isoformat is the closest approximation to the manual date/timestamp format matching in Spark. `python-dateutil `is too permissive and would allow values that should remain as StringType to be misclassified as date/TS castable.

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-mayliu: Right, Spark by default doesn't infer decimal for XML. Spark's DecimalType inference is gated behind `options.prefersDecimal`, which defaults to false.
`case v if options.prefersDecimal && decimalTry.isDefined => decimalTry.get`

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-mayliu: 1. Only when mode=DROPMALFORMED, the row is skipped. When mode=PERMISSIVE, the bad fields are nullified and added to `_corrupt_record`. When mode=FAILFAST, raises an error. `_validate_row_for_type_mismatch` handles these behaviors upon checking `mode`
2. Original behavior without inferSchema or custom schema is preserved, since no VARIANT casting was needed. This does not change custom schema behavior, but prevent custom schema from raising SQL casting errors where it shouldn't.

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-mayliu: `_validate_row_for_type_mismatch` is guarded by `if schema_type is not None`, which only applies to custom schema and infer schema cases. It enhances custom schema behavior, and leaves the original default Variant-returning xml parser behaviors unchanged.

This fixes an issue that would previously throw exception for custom schema and now infer schema. If a schema that mismatches input data type is provided, original behavior would throw SQL exception failure to cast.

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-mayliu: Discussed offline, `_validate_row_for_type_mismatch` is O(n) where n is the number of schema fields/columns; whereas `element_to_dict_or_str` traverses every element recursively. Thus, `_validate_row_for_type_mismatch`'s performance impact is minimal compared it.

From contextual point of view, `_validate_row_for_type_mismatch` needs to validate against the resulting transformed dict, so it'd be better to semantically and sequentially separate these functions.

src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -102,13 +114,97 @@ def _restore_colons_in_template(template: Optional[dict]) -> Optional[dict]:
...
+            return False
+    if isinstance(target_type, BooleanType):
+        return value.lower() in ("true", "false", "1", "0")
+    if isinstance(target_type, DateType):
+        try:
+            datetime.date.fromisoformat(value)
+            return True
+        except (ValueError, TypeError):
+            return False
+    if isinstance(target_type, TimestampType):
```
sfc-gh-yuwang, line 151: currently, this function will return True when the target type is not in the check(not StringType, LongType, DoubleType ....)
For example, if the target type is TimeType, this function always return True, is this expected?

sfc-gh-mayliu: Good point, when the target type is not checked, we fall back to the original behavior to throw errors upon encountering corrupted data. Snce we will guard this type mismatch permissive mode behavior with `context._is_snowpark_connect_compatible`, this is a no-op to Snowpark users.

In SCOS, since the current defined scope of `PERMISSIVE` mode is to unblock `infer_schema` usage, this function checks all possible infer_schema types returned by Spark, and thus would be sufficient for SCOS `infer_schema` usage.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1368,6 +1413,163 @@ def _get_schema_from_user_input(
...
+        output_schema = StructType([StructField("SCHEMA_VALUE", StringType(), True)])
+        schema_udtf = self._session.udtf.register_from_file(
+            inference_file_path,
+            handler_name,
+            output_schema=output_schema,
+            input_types=input_types,
+            packages=["snowflake-snowpark-python", "lxml<6"],
+            replace=True,
+            _suppress_local_package_warnings=True,
+        )
```
sfc-gh-aling: would any of these input for the registration change for different user call in a snowpark session?
I'm wondering if this udtf can be cached to avoid re-registering per user call

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-aling: did we confirm that `options.prefersDecimal` is out of the scope for implementation at this stage?

tests/integ/test_xml_infer_schema.py
```diff
@@ -0,0 +1,1189 @@
...
+
+
+@pytest.fixture(autouse=True)
+def enable_scos_compatible_mode():
+    """XML inferSchema is gated behind _is_snowpark_connect_compatible_mode.
+    Enable it for every test in this module."""
+    original = context._is_snowpark_connect_compatible_mode
+    context._is_snowpark_connect_compatible_mode = True
+    yield
+    context._is_snowpark_connect_compatible_mode = original
```
sfc-gh-aling: our tests run in parallel, changing module config might impact other tests execution.
I normally use unittest.mock.patch to patch module variant

tests/integ/test_xml_reader_row_tag.py
```diff
@@ -32,6 +37,15 @@
 ]
 
 
+@pytest.fixture()
+def enable_scos_compatible_mode():
+    """Enable SCOS compatible mode so that type validation runs inside the UDTF."""
+    original = context._is_snowpark_connect_compatible_mode
+    context._is_snowpark_connect_compatible_mode = True
+    yield
+    context._is_snowpark_connect_compatible_mode = original
```
sfc-gh-aling: same comments for using unittest.mock.patch for module var modification

tests/integ/test_xml_infer_schema.py
sfc-gh-mayliu: Good catch, fixed in tests

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-mayliu: The UDTF registration input remains the same, so it does make sense to reuse UDTF instead of re-registering per use call. In the new commit, I leveraged existing params from `register_from_file` such as `name=udtf_name, replace=False, if_not_exists=True, skip_upload_on_content_match=True` for server to handle deduplication/caching natively per session/schema scope.

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-mayliu: I'd say it's out of scope for current stage. I've documented and will log a JIRA to follow up on `options.prefersDecimal` in inferSchema. This was not yet requested by customer, and ignoring Decimal here should not interfere with functional requirements in `inferSchema` and `permissive` mode -- it only affects minor type promotion logic and advance straight from Long to Double.

PR #4125, opened 2026-03-17
SNOW-3250094: Support DIRECTED JOIN

sfc-gh-joshi:

   

   Fixes SNOW-3250094

Adds support for DIRECTED JOIN operations (https://docs.snowflake.com/en/sql-reference/constructs/join).

### DIFF
--- a/tests/integ/modin/frame/test_apply.py
+++ b/tests/integ/modin/frame/test_apply.py
@@ -245,7 +245,7 @@ def test_frame_with_timedelta_index():
 def test_axis_1_date_time_timestamp_type(data, func, expected_result):
     snow_df = pd.DataFrame(data)
     result = snow_df.apply(func, axis=1)
-    assert_snowpark_pandas_equal_to_pandas(result, expected_result)
+    assert_snowpark_pandas_equal_to_pandas(result, expected_result, check_dtype=False)
 
 
 @sql_count_checker(query_count=5, join_count=2, udtf_count=1)
@@ -330,7 +330,8 @@ def test_axis_1_return_not_json_serializable_label():
         ).to_pandas()
 
     with pytest.raises(
-        SnowparkSQLException, match="Object of type DataFrame is not serializable"
+        SnowparkSQLException,
+        match=r"Object of type DataFrame is not (JSON )?serializable",
     ):
         # return value
         snow_df.apply(lambda x: native_pd.DataFrame([1, 2]), axis=1).to_pandas()
--- a/tests/integ/modin/series/test_apply_and_map.py
+++ b/tests/integ/modin/series/test_apply_and_map.py
@@ -401,7 +401,7 @@ class TestApplyOrMapCallable:
         # In Python UDF, if the return type is variant, the return value must be
         # json serializable so it can become a variant in Snowflake.
         # type() returns a type object which is not json serializable.
-        with pytest.raises(SnowparkSQLException, match="is not serializable"):
+        with pytest.raises(SnowparkSQLException, match=r"is not (JSON )?serializable"):
             getattr(snow_series, method)(type).to_pandas()
 
     @pytest.mark.parametrize("func", [str, int, float, bytes, list, dict])
--- a/tests/integ/scala/test_udf_suite.py
+++ b/tests/integ/scala/test_udf_suite.py
@@ -567,7 +567,7 @@ def test_geometry_type(session):
     table_name = Utils.random_name_for_temp_object(TempObjectType.TABLE)
     Utils.create_table(session, table_name, "g geometry", is_temporary=True)
     session._run_query(
-        f"insert into {table_name} values ('POINT(30 10)'), ('POINT(50 60)'), (null)"
+        f"insert into {table_name} values ('POINT(30.5 10.5)'), ('POINT(50.5 60.5)'), (null)"
     )
     df = session.table(table_name)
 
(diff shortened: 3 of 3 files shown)


src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -104,6 +104,7 @@
 NATURAL = " NATURAL "
 ASOF = " ASOF "
 MATCH_CONDITION = " MATCH_CONDITION "
+DIRECTED = " DIRECTED"
```
sfc-gh-mayliu:
```suggestion
DIRECTED = " DIRECTED "
```

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
```diff
@@ -189,15 +189,17 @@ def __init__(
...
         super().__init__(left, right)
         self.join_type = join_type
         self.join_condition = join_condition
         self.match_condition = match_condition
+        self.directed = directed
 
     @property
     def sql(self) -> str:
-        return self.join_type.sql
+        return self.join_type.sql + (" DIRECTED" if self.directed else "")
```
sfc-gh-mayliu, line 202:
```suggestion
        return self.join_type.sql + (" DIRECTED " if self.directed else "")
```

sfc-gh-mayliu (commented):
Overall LGTM, adding space before and after the SQL keyword is generally a safer approach

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
sfc-gh-joshi: This clause actually isn't used anywhere (the analyzer bypasses this and uses the join type directly). It also previously was not space-separated, since returning `self.join_type.sql` directly just returns "INNER", "OUTER", etc.

src/snowflake/snowpark/_internal/proto/ast.proto
```diff
@@ -1044,11 +1044,12 @@ message DataframeCreateOrReplaceView {
 
 // dataframe.ir:193
 message DataframeCrossJoin {
-  Expr lhs = 1;
-  google.protobuf.StringValue lsuffix = 2;
-  Expr rhs = 3;
-  google.protobuf.StringValue rsuffix = 4;
-  SrcPosition src = 5;
+  google.protobuf.BoolValue directed = 1;
```
sfc-gh-aling, line 1047: is this based upon the monorepo IR change?
I'm wondering why `directed = 1`(prepend to the beginning) instead of `directed = 6` (append to the last), would this break existing IR?

src/snowflake/snowpark/_internal/proto/ast.proto
```diff
@@ -1134,33 +1135,34 @@ message DataframeGroupByGroupingSets {
...
 message DataframeJoin {
-  Expr join_expr = 1;
-  JoinType join_type = 2;
-  Expr lhs = 3;
-  google.protobuf.StringValue lsuffix = 4;
-  Expr match_condition = 5;
-  Expr rhs = 6;
-  google.protobuf.StringValue rsuffix = 7;
-  SrcPosition src = 8;
+  google.protobuf.BoolValue directed = 1;
```
sfc-gh-aling, line 1147: same question on the position change caused by directed prepend

src/snowflake/snowpark/_internal/proto/ast.proto
```diff
@@ -1218,12 +1220,13 @@ message DataframeNaReplace {
...
 }
 
-// dataframe.ir:291
+// dataframe.ir:293
 message DataframeNaturalJoin {
-  JoinType join_type = 1;
-  Expr lhs = 2;
-  Expr rhs = 3;
-  SrcPosition src = 4;
+  google.protobuf.BoolValue directed = 1;
```
sfc-gh-aling, line 1225: +1 here on position change

src/snowflake/snowpark/dataframe.py
```diff
@@ -3444,6 +3444,8 @@ def natural_join(
         self,
         right: "DataFrame",
         how: Optional[str] = None,
+        *,
+        directed: bool = False,
```
sfc-gh-aling, line 3448: from the doc, it says:

> If the DIRECTED keyword is added, the join type — for example, INNER, LEFT, RIGHT, or FULL — is required.

do we need client side validation for this

src/snowflake/snowpark/_internal/analyzer/binary_plan_node.py
```diff
@@ -189,15 +189,17 @@ def __init__(
         join_type: JoinType,
         join_condition: Optional["Expression"],
         match_condition: Optional["Expression"],
+        directed: bool,
```
sfc-gh-aling, line 192: nit: I see other places default `directed` to False, do you want to apply the same style to the constructor here?

sfc-gh-joshi: Since this is an internal analyzer construct, I think the argument should be mandatory to ensure that the correct value is being passed in all cases and we don't accidentally forget to propagate it.

src/snowflake/snowpark/_internal/proto/ast.proto
sfc-gh-joshi: I checked with Hemit, and it seems that IR generation doesn't enforce any ordering.

> The protobuf is actually auto-generated from the .ir files that we edit, and I'm not sure if it has ever maintained ordering.
>
> From a protobuf message definition perspective, this is kind of bad for backwards compatibility, but the thing is we distribute the message protobuf definitions ourselves in the client library anyways (along with the generated Python stubs), so it will always be "up to date".

src/snowflake/snowpark/dataframe.py
sfc-gh-joshi: The client implicitly produces "INENR JOIN" if the user doesn't specify a type, so this shouldn't be necessary. If there ever is something incompatible, I think we should leave it up to the server to report the error to the user.

PR #4128, opened 2026-03-20
Fix leading slash in PATTERN→FILES path for INFER_SCHEMA

sfc-gh-dyadav:
# Fix leading slash in PATTERN-to-FILES path conversion for INFER_SCHEMA

## Problem

When `DataFrameReader` reads from an external stage with a `PATTERN` option, it converts `LIST @stage PATTERN='...'` results into explicit file paths for the `FILES` parameter of `INFER_SCHEMA`. This conversion strips the cloud prefix (e.g., `s3://my-bucket`) from the full file path returned by `LIST`.

However, when the stage location has **no trailing slash** (e.g., `s3://my-bucket` instead of `s3://my-bucket/`), the regex substitution leaves a **leading slash** on the file path:

```
LIST result:         s3://my-bucket/subdir/data/part-00000.parquet
Stage location:      s3://my-bucket        (no trailing slash)
After regex strip:   /subdir/data/part-00000.parquet   ← leading slash!
```

Snowflake's `INFER_SCHEMA` cannot resolve files with this leading slash, failing with **"Remote file not found"**. The existing code silently catches this and falls back to unfiltered `INFER_SCHEMA` (without `FILES`), which reads **all** files in the stage — including metadata files like `_common_metadata` that can corrupt schema inference (e.g., timestamps inferred as `VARIANT`/`StringType` instead of `TimestampType`).

## Root Cause

In `_infer_schema_for_file_format`, the prefix-stripping loop builds file paths from `LIST` results:

```python
for match in matches:
    prefix = match[0]
    for regex in regexes:
        prefix = regex.sub("", prefix)
    files.append(prefix)          # BUG: prefix may start with "/"
```

When the stage location lacks a trailing slash, `re.sub` removes `s3://my-bucket` but leaves the `/` that separates the bucket from the path.

## Fix

One-line fix — strip the leading slash after regex substitution:

```python
files.append(prefix.lstrip("/"))
```

This ensures the file path is always relative to the stage root, matching what Snowflake expects in the `FILES` parameter.

## Impact

Without this fix, any external stage whose location is configured without a trailing slash will silently fall back to unfiltered schema inference when `PATTERN` is used. This manifests as:

- **Incorrect type inference** when metadata files (e.g., `_common_metadata`, `_metadata`) are present in the stage alongside data files
- **Timestamps inferred as `VARIANT`** instead of `TimestampType` / `TimestampNTZType`
- **Silent degradation** — no error is raised; the fallback just produces wrong types

## Test

Added `test_pattern_with_infer_strips_leading_slash` which validates the prefix-stripping logic for both trailing-slash and no-trailing-slash stage locations, asserting that output file paths never start with `/`.

## Files Changed

| File | Change |
|------|--------|
| `src/snowflake/snowpark/dataframe_reader.py` | Add `.lstrip("/")` to prefix stripping in `_infer_schema_for_file_format` |
| `tests/integ/scala/test_dataframe_reader_suite.py` | Add regression test for leading-slash stripping |

   Fixes SNOW-2867055


tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -1913,6 +1913,44 @@ def base_reader(include_pattern=True):
             Utils.drop_stage(session, stage_name)
 
 
+def test_pattern_with_infer_strips_leading_slash():
```
sfc-gh-aling: the test shall be moved to the test/unit

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1272,7 +1272,7 @@ def _infer_schema_for_file_format(
                         prefix = match[0]
                         for regex in regexes:
                             prefix = regex.sub("", prefix)
-                        files.append(prefix)
+                        files.append(prefix.lstrip("/"))
```
sfc-gh-aling, line 1275: can we add a changelog for the fix?

tests/integ/scala/test_dataframe_reader_suite.py
sfc-gh-aling, line 1: can we add integration tests for the fix?
check out this test: https://github.com/snowflakedb/snowpark-python/blob/494aeb04daf171f5d24805f70cae5be52993e40c/tests/integ/test_dataframe.py#L5877-L5900

you can write df into the files in s3 bucket first
and then read them back to see if the types are correctly inferred with your fix

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-dyadav: Done

tests/integ/test_parquet_pattern_infer_schema.py
```diff
@@ -0,0 +1,96 @@
...
+            f"@{stage_name}/subdir"
+        )
+
+        schema = df.schema
+        ts_field = [
+            f for f in schema.fields if f.name.strip('"').upper() == "UPDATED_TIME"
+        ][0]
+        assert not isinstance(
+            ts_field.datatype, StringType
+        ), f"Expected timestamp-like type, got {ts_field.datatype}"
```
sfc-gh-aling: I think we should be clear on what type to expect, instead of what type it should not be

```suggestion
        assert not isinstance(
            ts_field.datatype, TimestampType
        ), f"Expected timestamp-like type, got {ts_field.datatype}"
```

tests/integ/test_parquet_pattern_infer_schema.py
sfc-gh-aling, line 1: nit: let's move this under test_dataframe_reader_file.py

tests/unit/test_infer_schema_pattern_leading_slash.py
sfc-gh-aling, line 1: actually do we need this unit test here? -- it's not testing any code change in snowpark
since we have the integ test now, I think we get enough coverage?

sfc-gh-dyadav:
@sfc-gh-mayliu no SAS test should require update for this one

PR #4132, opened 2026-03-23
SNOW-3266495: Defer HAVING, ORDER BY, LIMIT clauses in SCOS compatibility mode

sfc-gh-joshi:

   

   Fixes SNOW-3266495

Snowflake SQL requires HAVING, ORDER BY, and LIMIT clauses of a GROUP BY statement to appear in that particular order. For example:
```sql
SELECT 
    "DEPT", 
    count(1) AS "HEADCOUNT", 
    avg("SALARY") AS "AVG_SALARY"
 FROM (
 SELECT 
    "ID", 
    "DEPT", 
    "SALARY"
 FROM (
 SELECT $1 AS "ID", $2 AS "DEPT", $3 AS "SALARY" FROM  VALUES (1 :: INT, 'engineering' :: STRING, 80000 :: INT), (2 :: INT, 'engineering' :: STRING, 90000 :: INT), (3 :: INT, 'sales' :: STRING, 50000 :: INT), (4 :: INT, 'sales' :: STRING, 60000 :: INT), (5 :: INT, 'hr' :: STRING, 45000 :: INT), (6 :: INT, 'hr' :: STRING, 55000 :: INT), (7 :: INT, 'engineering' :: STRING, 85000 :: INT)
)
)
 GROUP BY 
    "DEPT"
 HAVING ("HEADCOUNT" > 1)
 ORDER BY "AVG_SALARY" DESC NULLS LAST
 LIMIT 2 OFFSET 0
```

Re-arranging any of these clauses produces a syntax error. Currently, when `_is_snowpark_connect_compatible_mode` is set, the order in which these clauses are added depends on the order in which they are specified by the user. That is, `df.groupBy(...).agg(...).sort(...).filter(...)` would emit `ORDER BY` before `HAVING`, resulting in a compilation error.

This PR defers the generation of these clauses for GROUP BY aggregations. Instead of appending each clause to the analyzer tree when a method is called, the new `DataFrame._build_post_agg_df` method ensures the clauses are added in the correct order.


tests/integ/test_df_aggregate.py
```diff
@@ -928,6 +928,152 @@ def test_filter_sort_limit_snowpark_connect_compatible(session, sql_simplifier_e
...
+            result10,
+            [
+                Row("sales", 2, 55000.0),
+                Row("hr", 2, 50000.0),
+                Row("engineering", 3, 85000.0),
+            ],
+        )
+
+        # Repeated filter: first filter uses the HAVING append path,
+        # second filter falls back to the regular (subquery WHERE) path.
```
sfc-gh-aling: can you help check the spark behavior?
I'm wondering if the second filter should be merged into having or a regular where

src/snowflake/snowpark/dataframe.py
```diff
@@ -689,7 +689,15 @@ def __init__(
...
         self.is_cached: bool = is_cached  #: Whether the dataframe is cached.
+        # Internal state variables used to construct flattened GROUP BY clauses in the correct order
+        # in SCOS compatibility mode.
+        # See comments on `_build_post_agg_df` for details.
         self._ops_after_agg = None
+        self._agg_base_plan = None
+        self._agg_base_select_statement = None
+        self._pending_having = None
+        self._pending_order_by = None
+        self._pending_limit = None
```
sfc-gh-aling: I'm curious how multiple filter would affect the plan generation

```python
df1 = df.groupBy(...).agg(...)
df2 = df1.filter(...).limit().filter(...)
df3 = df1.filter(...).filter(...).limit()
df4 = df1.limit().filter(...).filter(...)
```

1. do df2,3,4 output the same query?
2. what's the behavior in spark and do we align with spark behavior after your code change?

sfc-gh-joshi: That's a good point. It looks like in spark, `filter` is not commutative across a sequence of `df.filter(...).orderBy(...).limit(...).filter(...)` (the final call will see a deterministic subset of rows based on the prior order/limit). I'll need to do some more testing to see what this means for SQL generation, and whether the cases you mentioned have similar problems.

sfc-gh-joshi: @sfc-gh-aling I added some test cases covering this behavior, and checked the output against spark. The changes are:
1. Operations that occur after a LIMIT now always produce a new sub-query, since FILTER -> LIMIT and LIMIT -> FILTER are not equivalent.
2. Consecutive filter operations are now conjoined into a single HAVING clause. I don't think SQL has any short-circuiting evaluation behavior that imperative languages do, so I believe this should always be equivalent. The Spark `explain` plans I looked at did also combine filter clauses together into a single operator.
3. Consecutive ordering operations are now combined into a single ORDER BY, with the last ordering clause appearing first in the SQL.

Most of these cases were previously broken in SCOS, as the only sequence of operations that would have produced valid SQL was `df.groupby(...).agg(...).filter(...).orderBy(...).limit(...)`, where the operations appeared in the same order as that required by SQL.

sfc-gh-aling: nice, thanks for checking the behavior!
re1: does `orderBy` also produce a subquery like `limit`?

sfc-gh-joshi: ORDER BY does not create a new sub-query, since FILTER and ORDER BY are semantically commutative.

PR #4138, opened 2026-03-26
SNOW-3266242: Support TRY_CAST with user-provided schema in DataFrameReader

sfc-gh-yuwang:

   

   Fixes SNOW-3266242

   preiously try_cast is not supported when using user schema, this PR meant to fix this gap


src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -797,6 +797,7 @@ def csv(self, path: str, _emit_ast: bool = True) -> DataFrame:
         self._file_type = "CSV"
 
         schema_to_cast, transformations = None, None
+        use_user_schema = False
```
sfc-gh-aling: why we need to introduce this new var?
was there a bug before?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -833,7 +834,19 @@ def csv(self, path: str, _emit_ast: bool = True) -> DataFrame:
...
         else:
             self._cur_options["INFER_SCHEMA"] = False
-            schema = self._user_schema._to_attributes()
+            try_cast = self._cur_options.get("TRY_CAST", False)
+            if try_cast:
+                (
+                    schema,
+                    schema_to_cast,
+                    transformations,
+                ) = self._get_schema_from_csv_user_input(
```
sfc-gh-aling: can you check with @sfc-gh-skumbham to understand the scope? is this requirement just for csv?
do we need generalization for other formats?

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: technically there is a bug before,
 in the analyzer: https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L1960
we need this bool value to actually apply try_cast to the result, otherwise try_cast is ignored

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: I have confirmed with him that this PR's scope is limited to CSV

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1387,6 +1402,37 @@ def _infer_schema_for_file_format(
 
         return new_schema, schema_to_cast, read_file_transformations, None
 
+    def _get_schema_from_csv_user_input(
+        self, user_schema: StructType, try_cast: bool
```
sfc-gh-aling, line 1403: nit: do we need try_cast argument? this function is only called when cast is set to True

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1387,6 +1402,37 @@ def _infer_schema_for_file_format(
...
+            )
+            sf_type = convert_sp_to_sf_type(field.datatype)
+            source_column = f"${index}"
+            identifier = (
+                f"TRY_CAST({source_column} AS {sf_type})"
+                if try_cast
+                else f"{source_column}::{sf_type}"
+            )
+            schema_to_cast.append((identifier, field.name))
+            transformations.append(sql_expr(identifier))
```
sfc-gh-aling: would try cast affect output column name? we want the output column name unchanged
can we also test the output schema in the tests?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1387,6 +1402,37 @@ def _infer_schema_for_file_format(
...
+        for index, field in enumerate(user_schema.fields, start=1):
+            new_schema.append(
+                Attribute(
+                    field.column_identifier.quoted_name,
+                    field.datatype,
+                    field.nullable,
+                )
+            )
+            sf_type = convert_sp_to_sf_type(field.datatype)
+            source_column = f"${index}"
```
sfc-gh-aling: nit: I think after removing the try_cast bool check, this string can be consolidated into identifier

tests/integ/scala/test_dataframe_reader_suite.py
```diff
@@ -375,6 +375,30 @@ def test_read_csv(session, mode):
...
+            StructField("b", IntegerType()),
+            StructField("c", DoubleType()),
+        ]
+    )
+    df_try_cast = (
+        reader.schema(try_cast_schema).option("TRY_CAST", True).csv(test_file_on_stage)
+    )
+    try_cast_res = df_try_cast.collect()
+    try_cast_res.sort(key=lambda x: x[0])
+    assert try_cast_res == [Row(1, None, 1.2), Row(2, None, 2.2)]
```
sfc-gh-aling: what's the input data type from test_file_csv? what the result be different when testing w/ and w/o try_cast?

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: it will not affect output column name, will add test for column name too

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -834,7 +835,13 @@ def csv(self, path: str, _emit_ast: bool = True) -> DataFrame:
...
         else:
             self._cur_options["INFER_SCHEMA"] = False
-            schema = self._user_schema._to_attributes()
+            try_cast = self._cur_options.get("TRY_CAST", False)
+            (
+                schema,
+                schema_to_cast,
+                transformations,
+            ) = self._get_schema_from_csv_user_input(self._user_schema, try_cast)
+            use_user_schema = try_cast
```
sfc-gh-aling: why we do not always set use_user_schema to True here?
from the var name, it implies that the value should be determined by whether user schema is provided, not on whether `TRY_CAST` is used, did I miss any detail here

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@
 
 - Allow user input schema when reading Parquet file on stage.
 
+#### Bug Fixes
+
+- Fixed a bug that `TRY_CAST` reader option is ignored when calling `DataFrameReader.schema().csv()`.
+- Fixed a Snowflake platform compatibility issue (SNOW-3259059) where `concat(lit('"'), ...)` could lose the leading quote through some `EXCEPT` / chained set-operation plans by lowering that literal to `CHR(34)` in generated SQL.
+- Fixed a bug where chained `DataFrame.filter()` calls with raw SQL text containing `OR` produced incorrect results.
```
sfc-gh-aling: nit: do you need to rebase the main branch to fix the log drift?

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-aalam: this seems wrong to me too. we should `use_user_schema` if user provided us with a schema instead of giving the option of `INFER_SCHEMA`. If user provides us with a schema, we should try to cast the given csv column into the expected data.

sfc-gh-yuwang: I am trying to maintain the original behavior, in the older version, use_user_schema is always set to False in csv read even if user_schema exist

CHANGELOG.md
sfc-gh-yuwang: I'll fix this

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-yuwang: another reason is that the old behavior would not generate `schema_to_cast` and `transformations` under any condition when reading csv, thus it would raise an error here: https://github.com/snowflakedb/snowpark-python/blob/main/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py#L2030 if we set `use_user_schema` to True.

always generating `schema_to_cast` and `transformations` when using user schema would also create BCR

sfc-gh-aling (approved):
LGTM.

can you try making the scos change code on a dev branch to verify that this fix works using the scos regression test job?

I need on that dev branch you need to:
1. remove the temp try_cast selection code
2. bump the snowpark dependency

PR #4140, opened 2026-03-26
SNOW-3283826: retry upon cte error for snowpark connect mode

sfc-gh-aling:

   

   Fixes SNOW-3283826

## Description

When CTE optimization rewrites a query, the resulting SQL may occasionally fail at execution time with a `ProgrammingError` even though the equivalent unoptimized SQL would succeed. This change adds a transparent retry mechanism that catches these failures and re-executes with unoptimized SQL, preventing user-visible errors.

### How it works

When a `ProgrammingError` occurs during execution in `ServerConnection.get_result_set`:

1. **Guard checks** — Retry only triggers when all conditions are met: the query is synchronous (`block=True`), Snowpark Connect compatible mode is enabled, CTE optimization is enabled, and the failed query actually contains CTE syntax (starts with `WITH `).

2. **Retry with unoptimized SQL** — Calls `SnowflakePlan.get_execution_queries_without_cte()`, which recompiles the plan via `PlanCompiler.compile(skip_cte_optimization=True)` without mutating any session state. The unoptimized queries are then executed. If the retry also fails, the original error is re-raised.

3. **Auto-disable** — When the threshold is positive (`context._cte_error_threshold`, default 3), each successful retry increments a per-session counter (`session._cte_error_count`). Once the counter reaches the threshold, CTE optimization is automatically disabled for the remainder of the session with a warning log. Setting the threshold to 0 disables auto-disable entirely and skips counter increments.

4. **Telemetry** — Every successful retry sends a `cte_execution_retry` telemetry event containing the plan UUID, error message, `api_calls` chain, and both the failed CTE query ID and the successful retry query ID (real Snowflake sfqids, not raw SQL). When auto-disable triggers, a separate `cte_optimization_auto_disabled` event is sent.

### Files changed

| File | Change |
|------|--------|
| `server_connection.py` | Core retry logic in `get_result_set`, extracted `_execute_queries` helper |
| `snowflake_plan.py` | Added `get_execution_queries_without_cte()` |
| `plan_compiler.py` | Added `skip_cte_optimization` parameter to `compile()` |
| `context.py` | Added `_cte_error_threshold` module-level variable |
| `session.py` | Added `_cte_error_count` instance attribute |
| `telemetry.py` | Added `send_cte_execution_retry_telemetry` and `send_cte_optimization_auto_disabled_telemetry` |
| `telemetry_constants.py` | Added `CTE_QUERY_ID`, `RETRY_QUERY_ID`, `CTE_ERROR_COUNT` |
| `tests/integ/test_cte.py` | 10 integration tests covering retry success/failure, guard conditions, auto-disable, telemetry data, and threshold=0 |

### Test plan

- [x] `test_cte_retry_succeeds_increments_counter` — counter goes 0→1 on successful retry
- [x] `test_cte_retry_fails_reraises_original_error` — both CTE and unoptimized fail, counter stays 0
- [x] `test_cte_retry_no_retry_when_cte_disabled` — error propagates when CTE optimization is off
- [x] `test_cte_retry_no_retry_when_not_compatible_mode` — error propagates when compat mode is False
- [x] `test_cte_retry_no_retry_on_async_queries` — `block=False` skips retry
- [x] `test_cte_retry_auto_disable_at_threshold` — counter hits 3, CTE auto-disabled, telemetry sent
- [x] `test_cte_retry_counter_below_threshold_does_not_disable` — counter=1, CTE stays enabled
- [x] `test_cte_retry_telemetry_contains_reproduction_data` — real sfqids for both failed and retry queries, api_calls, plan_uuid, error_message
- [x] `test_cte_retry_threshold_zero_disables_auto_disable` — threshold=0 means no counter increment, no auto-disable
- [x] `test_cte_retry_no_retry_when_query_has_no_cte_prefix` — non-CTE query error propagates directly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 #### Bug Fixes
 
 - Fixed a bug where `Session.create_dataframe` raised `TypeError` when a `StringType` column was given a non-string Python value (e.g. `int`, `float`, `bool`, `Decimal`) for a small local relation (below the array bind threshold); `VALUES` SQL generation now coerces these types to string literals, consistent with the large-data bind-parameter path.
+- Fixed a bug where `DataFrame.approxQuantile` did not accept a `Column` for the `col` parameter.
 
 ### Snowpark Local Testing Updates
 
--- a/src/snowflake/snowpark/dataframe_stat_functions.py
+++ b/src/snowflake/snowpark/dataframe_stat_functions.py
@@ -23,7 +23,7 @@ from snowflake.snowpark._internal.telemetry import (
     adjust_api_subcalls,
 )
 from snowflake.snowpark._internal.type_utils import ColumnOrName, LiteralType
-from snowflake.snowpark._internal.utils import publicapi, warning
+from snowflake.snowpark._internal.utils import column_to_bool, publicapi, warning
 from snowflake.snowpark.functions import (
     _to_col_if_str,
     approx_percentile_accumulate,
@@ -91,7 +91,7 @@ class DataFrameStatFunctions:
              approximate percentile values if ``col`` is a list of column names.
         """
 
-        if not percentile or not col:
+        if not percentile or not column_to_bool(col):
             return []
 
         kwargs = {}
--- a/tests/integ/scala/test_dataframe_suite.py
+++ b/tests/integ/scala/test_dataframe_suite.py
@@ -608,36 +608,47 @@ def test_df_stat_cov(session):
     "config.getoption('local_testing_mode', default=False)",
     reason="FEAT: function approx_percentile_accumulate not supported",
 )
-def test_df_stat_approx_quantile(session):
-    assert TestData.approx_numbers(session).stat.approx_quantile("a", [0.5]) == [4.5]
+@pytest.mark.parametrize("use_col_object", [False, True], ids=["str", "Column"])
+def test_df_stat_approx_quantile(session, use_col_object):
+    def _col(name):
+        return col(name) if use_col_object else name
+
+    assert TestData.approx_numbers(session).stat.approx_quantile(_col("a"), [0.5]) == [
+        4.5
+    ]
     assert TestData.approx_numbers(session).stat.approx_quantile(
-        "a", [0.5], statement_params={"SF_PARTNER": "FAKE_PARTNER"}
+        _col("a"), [0.5], statement_params={"SF_PARTNER": "FAKE_PARTNER"}
     ) == [4.5]
 
     assert TestData.approx_numbers(session).stat.approx_quantile(
-        "a", [0, 0.1, 0.4, 0.6, 1]
+        _col("a"), [0, 0.1, 0.4, 0.6, 1]
     ) in (
         [-0.5, 0.5, 3.5, 5.5, 9.5],  # old behavior of Snowflake
         [0.0, 0.9, 3.6, 5.3999999999999995, 9.0],
     )  # new behavior of Snowflake
 
     with pytest.raises(SnowparkSQLException) as exec_info:
-        TestData.approx_numbers(session).stat.approx_quantile("a", [-1])
+        TestData.approx_numbers(session).stat.approx_quantile(_col("a"), [-1])
     assert "Invalid value [-1.0] for function 'APPROX_PERCENTILE_ESTIMATE'" in str(
         exec_info.value
     )
 
     with pytest.raises(SnowparkSQLException) as exec_info:
-        TestData.string1(session).stat.approx_quantile("a", [0.5])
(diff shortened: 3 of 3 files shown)


src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -696,12 +697,134 @@ def get_result_set(
...
+
+            logger.debug(
+                "CTE-optimized query failed with ProgrammingError: %s. "
+                "Retrying without CTE optimization.",
+                cte_error,
+            )
+
+            # Re-running _execute_queries is safe: setup queries preceding the
+            # main query are Snowpark-generated temp object DDL (e.g., CREATE
+            # TEMP FILE FORMAT IF NOT EXISTS) that are idempotent. CTE rewriting
```
sfc-gh-joshi: Are the post-action queries of the CTE-optimized attempt run on error? Also, is state like temporary identifier suffixes regenerated on the retry attempt?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -696,12 +697,134 @@ def get_result_set(
...
+                    error_message=str(cte_error),
+                    api_calls=plan.api_calls,
+                    cte_query_id=cte_query_id,
+                    retry_query_id=retry_query_id,
+                )
+
+                if context._cte_error_threshold > 0:
+                    with session._lock:
+                        session._cte_error_count += 1
+                        current_count = session._cte_error_count
```
sfc-gh-joshi: Can we make this lock more fine-grained to avoid potential contention with other code acquiring this lock to access other variables?

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -696,12 +697,134 @@ def get_result_set(
...
+                        logger.warning(
+                            "CTE optimization has caused %d execution failures "
+                            "(threshold=%d). Auto-disabling CTE optimization for "
+                            "the remainder of this session to avoid further "
+                            "performance impact.",
+                            current_count,
+                            context._cte_error_threshold,
+                        )
+                        with session._lock:
+                            session._cte_optimization_enabled = False
```
sfc-gh-joshi: There's a minor TOCTOU race condition with the code as-written. If another thread issues a CTE query that errors out between the two lock acquisitions in this function, it could issue an additional query with CTE enabled even though the threshold was already exceeded.

Checking `current_count` and setting `_cte_optimization_enabled = False` should occur in the previous lock acquisition block. The warning log can still be safely kept outside it.

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aling: 1) yes, they are run in the final block, and I added tests to confirm the behavior
2) yes, it's regenerated with fresh id

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aling: lock operation consolidated

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-aling: lock operation consolidated

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -696,12 +697,140 @@ def get_result_set(
...
+                    retry_query_id=retry_query_id,
+                )
+
+                if context._cte_error_threshold > 0:
+                    cte_disabled = False
+                    with session._lock:
+                        session._cte_error_count += 1
+                        if session._cte_error_count >= context._cte_error_threshold:
+                            session._cte_optimization_enabled = False
+                            cte_disabled = True
```
sfc-gh-joshi: This code should be race-free now, but my previous comment about making the lock more ["fine-grained" ](https://en.wikipedia.org/wiki/Lock_(computer_science)#Granularity)was about whether we could use a separate lock object to guard the CTE error variables instead of the global session lock. I'm not entirely certain where else this lock is used, but keeping locks specific to a single purpose helps reduce the possibility of contention, or in the worst case, deadlock.

What I have in mind is something like this:
```python
if context._cte_error_threshold > 0:
    with session._cte_error_lock:  # initialize this in session.py
        session._cte_error_count += 1
        if session._cte_error_count >= context._cte_error_threshold:
            session._cte_optimization_enabled = False
        cte_disabled = not session._cte_optimization_enabled
```
which can optionally be split into finer read/write operations if you think we need to be even more fine-grained.

sfc-gh-aling: this is a good idea, I have introduced the new lock to guard the _cte_error_count

Issue #4124, opened 2026-03-16
SNOW-3257427: DataFrame.approxQuantile was not accepting a single Column object when passed to the col parameter.


teja0909:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using? 
  `python 3.10`

2. What operating system and processor architecture are you using?

   Windows-10-10.0.26200-SP0

3. What are the component versions in the environment (`pip freeze`)?

 snowflake-snowpark-python -1.47.0

4. What did you do?

```
 data = [
    Row(id=1, dept="IT", salary=5000),
    Row(id=2, dept="IT", salary=7000),
    Row(id=3, dept="HR", salary=4000),
    Row(id=4, dept="HR", salary=4500),
]

df = session.create_dataframe(data)
approx_q_column_object = df.approxQuantile(
        col("salary"),
        [0.5, 0.9]
    )
```
TypeError: Cannot convert a Column object into bool: please use '&' for 'and', '|' for 'or', '~' for 'not' if you're building DataFrame filter expressions. For example, use df.filter((col1 > 1) & (col2 > 2)) instead of df.filter(col1 > 1 and col2 > 2).
5. What did you expect to see?

  Instead of error , I am expecting Dataframe with the results.

6. Can you set logging to DEBUG and collect the logs?

   ```
   import logging

   for logger_name in ('snowflake.snowpark', 'snowflake.connector'):
       logger = logging.getLogger(logger_name)
       logger.setLevel(logging.DEBUG)
       ch = logging.StreamHandler()
       ch.setLevel(logging.DEBUG)
       ch.setFormatter(logging.Formatter('%(asctime)s - %(threadName)s %(filename)s:%(lineno)d - %(funcName)s() - %(levelname)s - %(message)s'))
       logger.addHandler(ch)
   ```

sfc-gh-sghosh:
Hi @teja0909 ,

Thanks for the update.
We will work on the fix to eliminate the issue.
Meanwhile, please use the provided workaround if possible.

PR #4148, opened 2026-03-31
Fix/snow 3257427 columnorname inconsistency

NS601023:

   

   Fixes SNOW-3257427

I updated the API so that its runtime behavior is consistent with the type annotations and the documented behavior.
This patch PR is related to [this](https://github.com/snowflakedb/snowpark-python/issues/4124).

### DIFF
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -3677,7 +3677,7 @@ class DataFrame:
     def join(
         self,
         right: "DataFrame",
-        on: Optional[Union[ColumnOrName, Iterable[ColumnOrName]]] = None,
+        on: Optional[Union[ColumnOrName, Iterable[str]]] = None,
         how: Optional[str] = None,
         *,
         lsuffix: str = "",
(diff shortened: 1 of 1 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -3986,17 +3986,19 @@ def join(
...
+                    if isinstance(c, str):
+                        converted_using_columns.append(c)
+                    elif isinstance(c, Column) and isinstance(c._expression, NamedExpression):
+                        converted_using_columns.append(c._expression.name)
+                    else:
+                        raise TypeError(
+                            f"All list elements for 'on' or 'using_columns' must be a column name (str) or a Column referencing a named column. "
+                            f"Got: '{type(c)}' at index {i}"
+                        )
+                using_columns = converted_using_columns
```
sfc-gh-aalam: @NS601023 could you elaborate on why this change is required?

NS601023: Thank you for your comment.

[The issue](https://github.com/snowflakedb/snowpark-python/issues/4124) I submitted a PR for is about the inconsistency in `approx_quantile` between implementation and an argument type indicated in its docs and type annotations.
But `join` has a very similar inconsistency, which I noticed when using the API.
Therefore I tried to fix both and include them this PR.
This change is to fix the inconsistency in `join` function.
When `join` gets a `list[Column]` like `[col('col_1'), col('col_2')]` as using_columns, the code behaves like this:
```Python
            elif (
                isinstance(using_columns, Iterable)                              # ← True
                and len(using_columns) > 0                                       # ← True
                and not all([isinstance(col, str) for col in using_columns])     # ← When col is a list of Column, all() returns False because col is not str. Therefore this line gets True 
            ):
                bad_idx, bad_col = next(
                    (idx, col)
                    for idx, col in enumerate(using_columns)
                    if not isinstance(col, str)                                  # ← For instance, if col[0] is Column type, then bad_col = col[0]
                )
                raise TypeError(                                                 # ← Error is raised
                    f"All list elements for 'on' or 'using_columns' must be string type. "
                    f"Got: '{type(bad_col)}' at index {bad_idx}"
                )
```

To avoid this type error and to keep the behavior the same as when `using_columns` is a `list[str]`, I think this change should work and have a lower risk to break existing behaviors.
You probably have a much deeper understanding of the code base, so I would appreciate it if you could suggest a better way to solve the issue more efficiently.

sfc-gh-aalam: @NS601023 I will spend some more time here. This is critical part of the code so I would try not making changes here unless there is an existing bug.

sfc-gh-aalam: I don't see anything wrong with your implementation here @NS601023 but from a risk management perspective, I'd request you to revert this and keep the changes minimal.

NS601023: I appreciate your feedback.

I have reverted the two commits related to handling `ColumnOrName` in the `join` method, so this PR now focuses only on the `approx_quantile` issue described in the original issue (#4124).

Regarding [the docstring](https://github.com/snowflakedb/snowpark-python/blob/1542f265dfccc72c724bc44d8eb1810765cc2096/src/snowflake/snowpark/dataframe.py#L3691), it might be clearer if it stated that `on` can accept `Column`, `str`, and `Iterable[str]`, but not `Iterable[Column]`. 
Currently it reads as if `on` can accept `list[Column]`:
> on: A column name or a :class:`Column` object or a list of them to be used for the join.

Thanks.

sfc-gh-aalam:
recheck

PR #4155, opened 2026-04-02
SNOW-3409016: Support structured type schema string

sfc-gh-wshangguan:

SNOW-3409016

### Summary
`INFER_SCHEMA` can now return structured type strings (for example, `OBJECT(...)`, `MAP(...)`, `ARRAY(...)`) for parquet/json files with nested columns. The legacy parser in `_infer_schema_for_file_format` was based on simple string splitting and could fail on nested structured types.
This PR:
- Adds a recursive structured type parser for `OBJECT(...)`, `MAP(...)`, `ARRAY(...)` with nesting and `NOT NULL` handling.
- Adds safer fallback handling for bare structured keywords (`OBJECT`/`MAP`/`ARRAY`) by mapping them to `VariantType` and using `$1:{name}` (no cast) to preserve column access.

### Problem
When `INFER_SCHEMA` returns nested structured strings like `OBJECT(bio TEXT, social OBJECT(twitter TEXT, linkedin TEXT))`, the old parser path can break because it assumes parenthesized values are numeric precision/scale.
Even when parsing succeeds, cast generation for structured strings can be brittle (for example, `NOT NULL` annotations inside structured casts).

### Solution
New parser path (`_parse_structured_type_str` -> `_sf_type_to_type_object`):
- Recursively parses nested structured type strings into Snowpark `StructType` / `ArrayType` / `MapType` with structured semantics.
- Handles Snowflake metadata aliases (`TEXT` -> `StringType`, `REAL` -> `DoubleType`, `FIXED` -> `LongType`).
- Falls back to `convert_sf_to_sp_type` for simple and parameterized scalar types (backward compatible).
Structured infer-schema identifier handling:
- For structured types, preserves raw inferred type text while stripping `NOT NULL` from cast text.
- For bare structured keywords (older backend behavior), returns `VariantType` and uses `$1:{name}` (no cast) to avoid failures and preserve column names.

### Changes
Core implementation:
- `src/snowflake/snowpark/dataframe_reader.py`
  - Added `_extract_paren_content()`
  - Added `_sf_type_to_type_object()`
  - Added `_parse_structured_type_str()`
  - Updated `_infer_schema_for_file_format` to use structured parser path
  - Fixed `format.lower` -> `format.lower()`
  - Added bare-structured fallback behavior (`VariantType` + no-cast identifier)
Supporting wiring/tests:
- `src/snowflake/snowpark/session.py`
  - Added session parameter `_PYTHON_SNOWPARK_USE_STRUCTURED_TYPE_INFER_SCHEMA`
- `tests/unit/test_dataframe_reader_type_parsing.py`
  - Added parser and infer-schema coverage (nested structured, nullability, bare structured fallback, parquet/json path behavior)
  - 
### Backward Compatibility
- Simple scalar types (`NUMBER`, `VARCHAR`, `BOOLEAN`, etc.) continue to use existing conversion logic.
- Legacy behavior remains available when structured infer-schema flag is disabled.
- Non-structured infer-schema results are unaffected.

### DIFF
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -51,13 +51,9 @@ from snowflake.snowpark._internal.type_utils import (
     ColumnOrName,
     convert_sf_to_sp_type,
     convert_sp_to_sf_type,
-    DATA_TYPE_STRING_OBJECT_MAPPINGS,
-    extract_nullable_keyword,
-    get_number_precision_scale,
-    get_string_length,
     most_permissive_type,
-    split_top_level_comma_fields,
     type_string_to_type_object,
+    _parse_structured_type_str,
 )
 from snowflake.snowpark._internal.udf_utils import get_types_from_type_hints
 from snowflake.snowpark._internal.xml_reader import DEFAULT_CHUNK_SIZE
@@ -98,10 +94,6 @@ from snowflake.snowpark.mock._connection import MockServerConnection
 from snowflake.snowpark.table import Table
 from snowflake.snowpark.types import (
     ArrayType,
-    DataType,
-    DecimalType,
-    DoubleType,
-    LongType,
     MapType,
     StringType,
     StructType,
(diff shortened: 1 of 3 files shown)


src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -111,6 +120,156 @@
...
+    base = type_str[:paren_idx].strip()
+    depth = 0
+    for i in range(paren_idx, len(type_str)):
+        if type_str[i] == "(":
+            depth += 1
+        elif type_str[i] == ")":
+            depth -= 1
+            if depth == 0:
+                return base, type_str[paren_idx + 1 : i]
+    return None
```
sfc-gh-aling: in reality if this line of code gets executed and **depth != 0**, does it mean we have a broken type_str (imbalanced  paren) from backend?
in this case shall we error out explicitly?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -111,6 +120,156 @@
...
+    - On OBJECT fields: sets StructField.nullable = False
+    """
+    type_str = type_str.strip()
+    if not type_str:
+        raise ValueError("Empty type string")
+
+    # Strip NOT NULL before parsing — the caller is responsible for using
+    # the nullable info. For top-level calls from ARRAY/MAP/OBJECT handlers
+    # below, NOT NULL is handled before recursing.
+    type_str, _ = extract_nullable_keyword(type_str)
```
sfc-gh-aling: I'm wondering what would happen to the top level nullability, you mentioned some caller would handle it, do you know where that is?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -111,6 +120,156 @@
...
+    if result is None:
+        normalized = type_str.replace(" ", "").lower()
+        if normalized in DATA_TYPE_STRING_OBJECT_MAPPINGS:
+            return DATA_TYPE_STRING_OBJECT_MAPPINGS[normalized]()
+        if normalized in _SF_EXTRA_TYPE_MAPPINGS:
+            return _SF_EXTRA_TYPE_MAPPINGS[normalized]()
+        raise ValueError(f"'{type_str}' is not a supported type")
+
+    base, inner = result
+    base_upper = base.upper()
```
sfc-gh-aling, line 199: do we think we can do `type_str = type_str.lower()` at the beginning of the function such that we can avoid calling `lower()`/`upper()` in the following up operations?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -111,6 +120,156 @@
...
+            key_type, value_type, structured=True, value_contains_null=value_nullable
+        )
+
+    if base_upper == "OBJECT":
+        fields = split_top_level_comma_fields(inner)
+        struct_fields = []
+        for field_def in fields:
+            field_def = field_def.strip()
+            if not field_def:
+                continue
```
sfc-gh-aling: curious what is this case, is it something like ",    ,   ," -- is this considered valid?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -111,6 +120,156 @@
...
+
+def _parse_structured_type_str(type_str, max_string_size):
+    """Parse a Snowflake type string from INFER_SCHEMA into a Snowpark DataType.
+
+    For structured types (OBJECT, MAP, ARRAY), uses the recursive parser.
+    For simple types, delegates to convert_sf_to_sp_type for precision/scale.
+    """
+    type_str = type_str.strip()
+    if not type_str:
+        return VariantType()
```
sfc-gh-aling, line 263: is this possible?

src/snowflake/snowpark/session.py
```diff
@@ -710,6 +716,12 @@ def __init__(
             )
         )
 
+        self._use_structured_type_infer_schema: bool = (
+            self._conn._get_client_side_session_parameter(
```
sfc-gh-aling: is this feature available in prod? shall we have integ test for it?

src/snowflake/snowpark/session.py
```diff
@@ -710,6 +716,12 @@ def __init__(
...
+                _PYTHON_SNOWPARK_USE_STRUCTURED_TYPE_INFER_SCHEMA, False
```
sfc-gh-yuwang: do we enable this feature by:
`alter session set PYTHON_SNOWPARK_USE_STRUCTURED_TYPE_INFER_SCHEMA=True` ?
And like Adam said, can we add an integ test to test the end to end behavior, we can use mock if it is not available on prod

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -111,6 +120,156 @@
...
+_NOT_NULL_RE = re.compile(r"\s+NOT\s+NULL", re.IGNORECASE)
+
+_SF_EXTRA_TYPE_MAPPINGS = {
+    "text": StringType,
+    "real": DoubleType,
+    "fixed": LongType,
+}
+
+
+def _extract_paren_content(type_str: str) -> Optional[Tuple[str, str]]:
```
sfc-gh-yuwang, line 133: i feel like added helper functions should be moved to snowflake/snowpark/_internal/type_utils.py

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-wshangguan: Good suggestion. I'll raise an error for this case.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-wshangguan: There is actually another `nullable` column in the result of `INFER_SCHEMA`, which represents the top level nullability. And the existing snowpark python code (caller) handles the `nullable` column for top level. https://github.com/snowflakedb/snowpark-python/blob/c5362e4d5a753c7288065e6bc9fd45a976cfdf2c/src/snowflake/snowpark/dataframe_reader.py#L1345-L1346

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-wshangguan: I might want to keep the current shape: `type_str` is reused as part of the cast identifier we hand to SQL, and the two normalizations (`upper` for keyword dispatch, `lower+strip-spaces` for the `DATA_TYPE_STRING_OBJECT_MAPPINGS` lookup) are different.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-wshangguan: Changed to raise an exception for this case.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-wshangguan: Probably not, just a defensive check.

src/snowflake/snowpark/session.py
sfc-gh-wshangguan: It's currently disabled by now. QQ for the snowpark CI, does the test account have the permission to alter Snowflake internal parameters? If so, we can add integ tests.

src/snowflake/snowpark/session.py
sfc-gh-wshangguan: Actually no. The flag `PYTHON_SNOWPARK_USE_STRUCTURED_TYPE_INFER_SCHEMA` is just to guard the code change in snowpark python. There are other account level parameters to control whether the backend to return structured type schema string or not. So as mentioned https://github.com/snowflakedb/snowpark-python/pull/4155/changes/BASE..129d7ef3c2cc04f98811eb63d3ae522b3d412d3c#r3103032414, we can add integ test if test account used in snowpark is able to alter Snowflake non-public parameter.

sfc-gh-wshangguan: Updated to another way to avoid the confusion here.

src/snowflake/snowpark/session.py
sfc-gh-aling: if it's a parameter invisible to client, probably we can not test it.
qq: can the parameter be set at session level sql -- "alter session set <param>=true"?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -111,6 +120,174 @@
...
+        if type_str[i] == "(":
+            depth += 1
+        elif type_str[i] == ")":
+            depth -= 1
+            if depth == 0:
+                return base, type_str[paren_idx + 1 : i]
+    raise ValueError(f"Unbalanced parentheses in type string: '{type_str}'")
+
+
+def _sf_type_to_type_object(type_str: str) -> DataType:
```
sfc-gh-aling: +1 on @sfc-gh-yuwang 's comment, can we move those util functions to type_utils.py module?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -111,6 +120,174 @@
...
+    length = get_string_length(type_str)
+    if length:
+        return StringType(length)
+
+    normalized = base_upper.replace(" ", "").lower()
+    if normalized in DATA_TYPE_STRING_OBJECT_MAPPINGS:
+        return DATA_TYPE_STRING_OBJECT_MAPPINGS[normalized]()
+    if normalized in _SF_EXTRA_TYPE_MAPPINGS:
+        return _SF_EXTRA_TYPE_MAPPINGS[normalized]()
+    raise ValueError(f"'{type_str}' is not a supported type")
```
sfc-gh-aling: this part of logic repeats the one at the beginning of the function.
IIUC, this is for handling non-structured types right? do we need the logic twice in the same function?
and if we indeed want, shall we define a local util function to reuse the logic?

src/snowflake/snowpark/session.py
sfc-gh-wshangguan: There is a global system parameter `ENABLE_STRUCTURED_TYPE_INFER_SCHEMA_FOR_PARQUET`, which can be set at session level. However, it's internal parameter that's not visible to clients. Does it help here?

PR #4156, opened 2026-04-03
SNOW-3319055: Fix behavior when int is passed to StringType column fo…

sfc-gh-aalam:
…r small dataset

   Fixes SNOW-3319055

   We fixed `Session.create_dataframe` so small local VALUES plans no longer raise `TypeError` when a `StringType` column is filled with a `int`: `datatype_mapper.to_sql` now turns those ints into string literals (same quoting/casts as normal strings).

### DIFF
--- a/.github/workflows/daily_precommit.yml
+++ b/.github/workflows/daily_precommit.yml
@@ -110,6 +110,7 @@ jobs:
           - { os: {image_name: ubuntu-latest-64-cores, download_name: linux}, python-version: "3.11", cloud-provider: gcp }
           - { os: {image_name: ubuntu-latest-64-cores, download_name: linux}, python-version: "3.12", cloud-provider: aws }
           - { os: {image_name: ubuntu-latest-64-cores, download_name: linux}, python-version: "3.13", cloud-provider: azure }
+          - { os: {image_name: ubuntu-latest-64-cores, download_name: linux}, python-version: "3.14", cloud-provider: gcp }
 
           # macOS + rotating cloud providers
           - { os: {image_name: macos-latest, download_name: macos}, python-version: "3.9", cloud-provider: gcp }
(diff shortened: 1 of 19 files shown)


tests/unit/test_datatype_mapper.py
```diff
@@ -704,18 +706,6 @@ def test_numeric_to_sql_without_cast():
     assert numeric_to_sql_without_cast(Decimal(0.5), DecFloatType()) == "0.5"
 
 
-@pytest.mark.parametrize(
-    "value, datatype",
-    [
-        (123, StringType()),
-        (0.2, StringType()),
```
sfc-gh-joshi, line 711: Does this PR affect behavior with local floating point literals in string columns?

sfc-gh-aalam: I updated it to support boolean, float, decimal and int correctly. I skipped datetime types because their formatting may depend on local session variable values.

tests/integ/test_dataframe.py
```diff
@@ -4078,20 +4078,48 @@ def test_create_dataframe_from_none_data(session):
 def test_create_dataframe_decimal_string_value_inline_and_large(session):
     from snowflake.snowpark._internal.analyzer import analyzer
 
-    schema = StructType([StructField("a", DecimalType(10, 2))])
+    original_threshold = analyzer.ARRAY_BIND_THRESHOLD
+    try:
+        analyzer.ARRAY_BIND_THRESHOLD = 10
```
sfc-gh-aling, line 4083: I'm wondering if this has side effect to other tests
we can use mock patch to overwrite the value temporarily.

sfc-gh-aalam: We use the same pattern in several other tests. Each test is run in its own thread so it should be safe.

https://github.com/snowflakedb/snowpark-python/blob/213b1cacf77dec4aeaee5ddbd7172fc2c192039a/tests/integ/test_cte.py#L232-L237

https://github.com/snowflakedb/snowpark-python/blob/213b1cacf77dec4aeaee5ddbd7172fc2c192039a/tests/integ/scala/test_update_delete_merge_suite.py#L547-L564

https://github.com/snowflakedb/snowpark-python/blob/213b1cacf77dec4aeaee5ddbd7172fc2c192039a/tests/integ/compiler/test_query_generator.py#L502-L526

PR #4211, opened 2026-05-04
SNOW-3473261: Add Iceberg tag time travel via version_tag

sfc-gh-igarish:
## Summary

Add Iceberg tag time travel via a ``version_tag=`` parameter that emits
Snowflake's released `AT(VERSION_TAG => '<name>')` clause — the tag-only
form of Iceberg time travel (Spark Iceberg's
`VERSION AS OF '<tag_name>'` for tag reads).

**Scope: tag reads only.** Branch reads are deferred until Snowflake's
unified `VERSION_REF` syntax (which subsumes branch + tag) ships; the
Python kwarg `version_tag` and SQL `VERSION_TAG` keyword match
Snowflake's already-released grammar, so this PR can land independently.

This is the tag-only companion to the merged snapshot-id PR (#4231) and
mirrors that PR's surface decisions point-for-point:

* `version_tag` is **hidden behind `**kwargs`** on `Session.table()`,
  `DataFrameReader.table()`, and `Table.__init__()` — consumed by Snowpark
  Connect, **not** part of the advertised public Snowpark Python surface
  yet. Direct callers can still pass it; the doc page just doesn't list it.
* Reader option aliases `version_tag` and `version-tag` (matching the
  existing `snapshot-id` / `snapshot_id` style). Both auto-set
  `time_travel_mode='at'` since tag reads are positional (bound to a
  specific snapshot), not range-of-time.
* **No AST proto changes.** The field travels through `**kwargs` the
  same way `version=` does — keeps the AST surface minimal and lets the
  monorepo generate any AST coverage later if/when this becomes a
  first-class API.
* **No CHANGELOG entry** — internal kwargs surface, not customer-visible
  (matches PR #4231).

## Validation

* `time_travel_mode='before'` is rejected — tag reads are positional, not
  range-of-time.
* `version_tag` is mutually exclusive with `statement`, `offset`,
  `timestamp`, `stream`, and `version`.
* Non-string and empty `version_tag` values are rejected.
* When provided without a mode, defaults to `'at'` (same shape as
  `version=`).

## Test plan

- [x] Unit tests in `tests/unit/test_utils.py`:
      `test_time_travel_version_tag` (SQL emission, mutual exclusion,
      mode validation, type / empty validation) and
      `test_extract_time_travel_version_tag_option` (canonical +
      hyphen option aliases, before-mode rejection). 17/17 tests in
      `test_utils.py` pass.
- [x] Black 22.3.0 + flake8 + pre-commit clean.
- [ ] Integ tests added under `tests/integ/test_dataframe.py`:
      `test_iceberg_version_tag_time_travel_session_table` and
      `test_iceberg_version_tag_time_travel_dataframe_reader_option`.
      Both are `pytest.mark.skip`'d pending a CI account with a
      CLD-linked unmanaged Iceberg table that has named tags and
      `FEATURE_ICEBERG_TIME_TRAVEL` enabled — same posture as the
      merged snapshot-id integ tests. Manual reproducer lives in
      `snowflake-eng/sas` (`tests/sas_tests/test_iceberg_version_tag_sample.py`).

## Future work

* Add `version_ref=` (or equivalent) once Snowflake ships the unified
  `AT(VERSION_REF => '<name>')` clause that handles both branches and
  tags. At that point `version_tag` can either alias to `version_ref` or
  remain as a tag-specific spelling for clarity.
* Promote both `version=` and `version_tag=` to the public surface once
  the Snowpark Connect consumer is stable.

## Related

* JIRA: [SNOW-3473261](https://snowflakecomputing.atlassian.net/browse/SNOW-3473261)
* Companion PR (merged): #4231 (`version=` for snapshot ids)

[SNOW-3473261]: https://snowflakecomputing.atlassian.net/browse/SNOW-3473261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -2102,34 +2102,34 @@ class TimeTravelConfig(NamedTuple):
 
         Note on escaping: string-valued parameters (``statement``,
         ``stream``, ``version_tag``, ``timestamp``) are embedded inside
-        single-quoted SQL literals. We double any embedded ``'`` to the
-        Snowflake-standard ``''`` escape sequence before interpolation
-        so a value like ``release_'s`` produces ``'release_''s'``
-        rather than breaking SQL or opening an injection surface
-        (e.g. ``'); DROP TABLE foo; --``). Numeric parameters
+        single-quoted SQL literals via the existing ``str_to_sql``
+        helper in ``analyzer.datatype_mapper`` so embedded ``'``, ``\\``
+        and newline characters are properly escaped. This keeps the
+        emission consistent with the rest of Snowpark Python's SQL
+        generation and avoids both broken SQL and injection surface
+        (e.g. ``x'); DROP TABLE foo; --``). Numeric parameters
         (``offset``, ``version``) are not quoted and don't need
         escaping.
-        """
 
-        def _quote(value: str) -> str:
-            # Snowflake's SQL string-literal escape is doubled single
-            # quotes; backslashes have no special meaning in
-            # single-quoted literals here, so we don't need to escape
-            # those.
-            return "'" + str(value).replace("'", "''") + "'"
+        ``str_to_sql`` is imported lazily here because
+        ``analyzer.datatype_mapper`` imports from this module at top
+        level — same pattern used elsewhere in this file for analyzer
+        cross-references.
+        """
+        from snowflake.snowpark._internal.analyzer.datatype_mapper import str_to_sql
 
         clause = f" {self.time_travel_mode.upper()} "
 
         if self.statement is not None:
-            clause += f"(STATEMENT => {_quote(self.statement)})"
+            clause += f"(STATEMENT => {str_to_sql(self.statement)})"
         elif self.offset is not None:
             clause += f"(OFFSET => {self.offset})"
         elif self.stream is not None:
-            clause += f"(STREAM => {_quote(self.stream)})"
+            clause += f"(STREAM => {str_to_sql(self.stream)})"
         elif self.version is not None:
             clause += f"(VERSION => {self.version})"
         elif self.version_tag is not None:
-            clause += f"(VERSION_TAG => {_quote(self.version_tag)})"
+            clause += f"(VERSION_TAG => {str_to_sql(self.version_tag)})"
         elif self.timestamp is not None:
             if self.timestamp_type is not None:
                 timestamp_type = self.timestamp_type.upper()
@@ -2141,9 +2141,9 @@ class TimeTravelConfig(NamedTuple):
                     func_name = "TO_TIMESTAMP_TZ"
                 else:
                     func_name = "TO_TIMESTAMP"
-                clause += f"(TIMESTAMP => {func_name}({_quote(self.timestamp)}))"
+                clause += f"(TIMESTAMP => {func_name}({str_to_sql(self.timestamp)}))"
             else:
-                clause += f"(TIMESTAMP => {_quote(self.timestamp)})"
+                clause += f"(TIMESTAMP => {str_to_sql(self.timestamp)})"
 
         return clause
 
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -1956,6 +1956,7 @@ class TimeTravelConfig(NamedTuple):
     timestamp_type: Optional[str] = None
     stream: Optional[str] = None
     version: Optional[int] = None
+    version_tag: Optional[str] = None
 
     @staticmethod
     def validate_and_normalize_params(
@@ -1966,6 +1967,7 @@ class TimeTravelConfig(NamedTuple):
         timestamp_type: Optional[Union[str, "TimestampTimeZone"]] = None,
         stream: Optional[str] = None,
         version: Optional[int] = None,
+        version_tag: Optional[str] = None,
     ) -> Optional["TimeTravelConfig"]:
         """
         Validates and normalizes time travel parameters.
@@ -1988,7 +1990,8 @@ class TimeTravelConfig(NamedTuple):
             ValueError: If parameters are invalid.
         """
         time_travel_arg_count = sum(
-            arg is not None for arg in (statement, offset, timestamp, stream, version)
+            arg is not None
+            for arg in (statement, offset, timestamp, stream, version, version_tag)
         )
 
         # Validate mode
@@ -2023,10 +2026,32 @@ class TimeTravelConfig(NamedTuple):
                 f"'version' must be an int Iceberg snapshot id, got {type(version).__name__}."
             )
 
+        # version_tag (Iceberg tag name, mapped to Snowflake's
+        # ``AT(VERSION_TAG => '<name>')`` grammar) only works with 'at' mode —
+        # Iceberg tag reads are positional (bound to a specific snapshot),
+        # not range-of-time, so ``BEFORE`` has no meaning.
+        if version_tag is not None and time_travel_mode.lower() != "at":
+            raise ValueError(
+                "Iceberg version_tag time travel can only be used with "
+                "time_travel_mode='at', not 'before'."
+            )
+
+        # Validate version_tag type — Iceberg tag names are strings. Empty
+        # strings are invalid.
+        if version_tag is not None:
+            if not isinstance(version_tag, str):
+                raise ValueError(
+                    f"'version_tag' must be a string Iceberg tag name, "
+                    f"got {type(version_tag).__name__}."
+                )
+            if not version_tag:
+                raise ValueError("'version_tag' must be a non-empty Iceberg tag name.")
+
         # Validate exactly one parameter is provided
         if time_travel_arg_count != 1:
             raise ValueError(
-                "Exactly one of 'statement', 'offset', 'timestamp', 'stream', or 'version' must be provided."
+                "Exactly one of 'statement', 'offset', 'timestamp', 'stream', "
+                "'version', or 'version_tag' must be provided."
             )
 
         # Normalize timestamp
@@ -2061,6 +2086,7 @@ class TimeTravelConfig(NamedTuple):
             timestamp_type=timestamp_type,
             stream=stream,
             version=version,
+            version_tag=version_tag,
         )
 
     def generate_sql_clause(self) -> str:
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/_internal/proto/ast.proto
```diff
@@ -2267,6 +2267,7 @@ message ReadTable {
   google.protobuf.StringValue time_travel_mode = 7;
   Expr timestamp = 8;
   google.protobuf.StringValue timestamp_type = 9;
+  google.protobuf.StringValue iceberg_tag = 10;
```
sfc-gh-aling: Is the client change generated from the monorepo AST definition?

I remember a client AST change isn't strictly required right now; if it's needed, it should be derived from the monorepo updates -- there is a step by step doc for the AST mono repo updates.

@sfc-gh-heshah what's your recommendation here? can we just not add the ast in the PR and do that later?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -2058,6 +2069,8 @@ def generate_sql_clause(self) -> str:
             clause += f"(OFFSET => {self.offset})"
         elif self.stream is not None:
             clause += f"(STREAM => '{self.stream}')"
+        elif self.iceberg_tag is not None:
+            clause += f"(ICEBERG_TAG => '{self.iceberg_tag}')"
```
sfc-gh-aling: I'm unable to see this feature in public doc.
what release status of this iceberg tag feature?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -125,6 +125,8 @@
     "TIMESTAMP": "timestamp",
     "TIMESTAMP_TYPE": "timestamp_type",
     "STREAM": "stream",
+    "ICEBERG_TAG": "iceberg_tag",
+    "TAG": "iceberg_tag",
```
sfc-gh-aling: When using `tag` as an alias for `iceberg_tag`, is that mapping defined within the Snowflake Iceberg spec or it's a spark spec.

snowflake has its own "tag" [concept](https://docs.snowflake.com/en/sql-reference/sql/create-tag), will there be future conflict under the context of dataframe reader?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -2082,6 +2110,8 @@ def generate_sql_clause(self) -> str:
             clause += f"(STREAM => '{self.stream}')"
         elif self.version is not None:
             clause += f"(VERSION => {self.version})"
+        elif self.version_tag is not None:
```
sfc-gh-yuwang, line 2131: Medium – SQL literal is built with raw interpolation for version_tag
In src/snowflake/snowpark/_internal/utils.py, new SQL emission uses "(VERSION_TAG => '{self.version_tag}')" directly.
If tag names can contain ' (or if input is untrusted), this can break SQL text and is a potential injection surface.
Safer pattern: escape single quotes ('') or reuse existing SQL-option escaping utility before embedding.
this is from cursor but I think is a valid one,
i also noticed that the existing code seems to also have this issue

sfc-gh-igarish: Good catch — fixed in 193bde063.

Centralized the literal escaping in a single `_quote` helper inside `generate_sql_clause` that doubles embedded `'` to `''` (Snowflake's standard string-literal escape) and applied it to all four string-valued parameters at once: `version_tag` (new in this PR) plus the three pre-existing offenders you flagged — `statement`, `stream`, and the string form of `timestamp`. Numeric parameters (`offset`, `version`) need no escaping.

Added `test_time_travel_string_literal_escaping` in `tests/unit/test_utils.py` covering all four parameters with both benign embedded quotes (e.g. `release_'s` → `'release_''s'`) and an explicit injection payload (`x'); DROP TABLE foo; --` → `'x''); DROP TABLE foo; --'`), plus the typed `TO_TIMESTAMP_NTZ` variant.

src/snowflake/snowpark/_internal/proto/ast.proto
sfc-gh-igarish: Addressed — the AST proto change is no longer in this PR. The current diff is six files (no `.proto` files) and `version_tag` travels through `**kwargs` the same way the merged `version=` (snapshot-id) PR #4231 does, so the AST surface stays minimal. If/when `version_tag` is promoted to a first-class public API later, the AST coverage can be regenerated from the monorepo at that point (following the step-by-step doc you mentioned).

`git diff main...HEAD --stat`:
```
src/snowflake/snowpark/_internal/utils.py  |  ~70
src/snowflake/snowpark/dataframe_reader.py |  ~56
src/snowflake/snowpark/session.py          |  ~11
src/snowflake/snowpark/table.py            |  ~11
tests/integ/test_dataframe.py              |  ~67
tests/unit/test_utils.py                   | ~180
```

src/snowflake/snowpark/_internal/utils.py
sfc-gh-igarish: Addressed by renaming `iceberg_tag` → `version_tag` (Python kwarg) and emitting `AT(VERSION_TAG => '<name>')` on the Snowflake side. `VERSION_TAG` is the *already-released* Snowflake grammar for tag-only Iceberg time travel — the SQL syntax we're targeting here is GA, not preview. Public Snowflake docs:

- https://docs.snowflake.com/en/sql-reference/constructs/at-before (`VERSION_TAG`)
- `AT(VERSION_TAG => '<tag_name>')` for tag reads, `AT(VERSION => <snapshot_id>)` for snapshot-id reads.

The PR title/body and commit message were updated to reflect this — the scope is **tag reads only** (the released form). Branch reads are deferred until the unified `VERSION_REF` clause ships, at which point we can either alias `version_tag` to it or keep it as a tag-specific spelling for clarity.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-igarish: Addressed — the bare `"TAG"` alias is **gone**. The only `DataFrameReader.option(...)` aliases now are `version_tag` and `version-tag` (mirroring the existing `snapshot-id` / `snapshot_id` pair on the merged PR #4231). Current mapping in `dataframe_reader.py`:

```python
version_tag_value = options.get("VERSION_TAG")
version_tag_source = "version_tag"
if version_tag_value is None:
    version_tag_value = options.get("VERSION-TAG")
    version_tag_source = "version-tag"
```

That avoids any collision with Snowflake's own object-tag concept (https://docs.snowflake.com/en/user-guide/object-tagging) — both alias names are explicitly scoped to "version" so the intent is unambiguous at the reader-option layer.

A small note on the broader Spark surface: Spark Iceberg uses `option(SparkReadOptions.TAG, "...")` (the literal string `"tag"`) on the DataFrameReader. The Spark-Connect mapping for that lives in the **SCOS** layer (`snowflake-eng/sas`), which translates it to this `version_tag` kwarg server-side — see https://github.com/snowflake-eng/sas/pull/3882. So Snowpark Python itself doesn't need to expose a bare `tag` alias to honor the Spark API; the translation happens one layer up.

tests/integ/test_dataframe.py
```diff
@@ -8423,3 +8423,70 @@ def test_iceberg_snapshot_id_time_travel_dataframe_reader_option(session):
...
+
+
+# ----------------------------------------------------------------------
+# Iceberg tag (``version_tag=``) time travel.
+#
+# TODO(SNOW-3473261): Wire these up to a CI test account that has:
+#   * a Catalog-Linked Database (CLD) such as cldUnity / cldglue, AND
+#   * an unmanaged Iceberg table inside it that exposes a named tag
+#     (e.g. created via ``ALTER TABLE ... CREATE TAG <name>`` on the OSS
+#     Iceberg side, or via the catalog's tag API).
```
sfc-gh-joshi, line 8435: How much effort would it be to enable this for our sfctest0 account? Having these tests skipped seems brittle in case any future changes occur.

sfc-gh-igarish: Right now we are having daily jobs to run Iceberg tests. Snowpark Python run tests on different deployment like Preprod, QA6 etc. We will enable the test once we have iceberg setup available in all deployment. Until then we will rely on daily job for regression.

sfc-gh-joshi (approved):

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -2069,19 +2095,41 @@ def generate_sql_clause(self) -> str:
...
+        single-quoted SQL literals. We double any embedded ``'`` to the
+        Snowflake-standard ``''`` escape sequence before interpolation
+        so a value like ``release_'s`` produces ``'release_''s'``
+        rather than breaking SQL or opening an injection surface
+        (e.g. ``'); DROP TABLE foo; --``). Numeric parameters
+        (``offset``, ``version``) are not quoted and don't need
+        escaping.
         """
+
+        def _quote(value: str) -> str:
```
sfc-gh-yuwang: I think there are existing helper functions in snowflake/snowpark/_internal/analyzer/analyzer_utils.py or snowflake/snowpark/_internal/analyzer/datatype_mapper.py, I am not sure if it perfectly fit the current usage, but I think you can let cursor decide whether to use them

sfc-gh-yuwang (approved):
preapproval assuming the comment is addressed

src/snowflake/snowpark/_internal/utils.py
sfc-gh-igarish: Good suggestion — done in 660a3242e.

Found `str_to_sql` in `snowflake/snowpark/_internal/analyzer/datatype_mapper.py:72`:

```python
def str_to_sql(value: str) -> str:
    sql_str = str(value).replace("\\", "\\\\").replace("'", "''").replace("\n", "\\n")
    return f"'{sql_str}'"
```

That's exactly the right helper — it's already the canonical SQL string-literal emitter in Snowpark Python (used by `server_connection` and `session`). Switched all four time-travel sites (`statement`, `stream`, `version_tag`, string-form `timestamp`) to use it and removed the inline `_quote` helper I had added in the previous commit.

Picks up two extra escape behaviors for free on top of the single-quote doubling:
- `\` → `\\` (backslash doubling)
- embedded newline → literal `\n` two-char sequence

Imported lazily inside `generate_sql_clause` because `analyzer.datatype_mapper` imports from this module at top level — same lazy-import pattern already used elsewhere in `utils.py` for analyzer cross-references (see the existing call sites at L1527, L1937, L2179 etc.).

Extended `test_time_travel_string_literal_escaping` to pin the new backslash + newline behaviors in addition to the existing single-quote and injection-payload cases, so any future change to `str_to_sql` that affects time-travel emission fails this test instead of silently producing broken SQL. All 18 tests in `test_utils.py` still pass.

PR #4215, opened 2026-05-06
fix(types): handle connector >=4.3.0 interval year-month display format

sfc-gh-kgadomski:
## Summary

Fixes incorrect display of YEAR TO MONTH interval values when using `snowflake-connector-python>=4.3.0`.

### Root Cause

`snowflake-connector-python` v4.3.0 changed the string representation of INTERVAL YEAR and INTERVAL MONTH types (see [connector changelog](https://github.com/snowflakedb/snowflake-connector-python/blob/main/DESCRIPTION.md#v430february-122026): "Fix string representation of INTERVAL YEAR and INTERVAL MONTH types").

Previously, the connector always returned compound year-month format (e.g., `+4-0` for 4 years, 0 months). Starting with v4.3.0, when the months component is zero, the connector returns just `+4` (omitting the `-0` suffix).

`format_year_month_interval_for_display` in `type_utils.py` only interpreted a simple `+N` value as years when the type was exactly `INTERVAL YEAR` (`start_field=YEAR, end_field=YEAR`). For `INTERVAL YEAR TO MONTH` (`start_field=YEAR, end_field=MONTH`), the simple `+N` value was incorrectly treated as months, producing:

```
INTERVAL '0-4' YEAR TO MONTH   (wrong: 0 years, 4 months)
```

instead of:

```
INTERVAL '4-0' YEAR TO MONTH   (correct: 4 years, 0 months)
```

### Fix

When `start_field` is `YEAR`, always interpret a simple `+N` value (no dash) as years, regardless of `end_field`. This is correct for:

- **Old connector (<4.3.0)**: always uses compound format `+Y-M`, so the simple-value branch is never reached for YEAR TO MONTH types — no behavioral change.
- **New connector (>=4.3.0)**: uses `+N` when months=0 — now correctly interpreted as years.

### Reproduction

```python
from snowflake.snowpark._internal.type_utils import format_year_month_interval_for_display
from snowflake.snowpark.types import YearMonthIntervalType

# Bug: '+4' with YEAR TO MONTH type treated as 4 months instead of 4 years
result = format_year_month_interval_for_display('+4', YearMonthIntervalType.YEAR, YearMonthIntervalType.MONTH)
# Before: "INTERVAL '0-4' YEAR TO MONTH" (wrong)
# After:  "INTERVAL '4-0' YEAR TO MONTH" (correct)
```

This was discovered via Snowpark Connect (SCOS) CI where `try_multiply(4, INTERVAL '1' YEAR)` produced incorrect output after widening the connector version cap to `<5.0.0`.

## Test plan

- [x] Added parametrized unit tests covering both old compound format and new simple format for all interval subtypes (YEAR, MONTH, YEAR TO MONTH)
- [x] All existing `tests/unit/test_datatype_mapper.py` tests pass

### DIFF
--- a/tests/unit/test_datatype_mapper.py
+++ b/tests/unit/test_datatype_mapper.py
@@ -16,6 +16,9 @@ from snowflake.snowpark._internal.analyzer.datatype_mapper import (
     to_sql,
     to_sql_no_cast,
 )
+from snowflake.snowpark._internal.type_utils import (
+    format_year_month_interval_for_display,
+)
 from snowflake.snowpark._internal.udf_utils import generate_call_python_sp_sql
 from snowflake.snowpark.types import (
     ArrayType,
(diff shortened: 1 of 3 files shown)


sfc-gh-mayliu (approved):
LGTM. Can you update changelog.md for this bugfix? Please merge main to get the test infra fix for a required check

tests/unit/test_datatype_mapper.py
```diff
@@ -809,3 +809,104 @@ def test_schema_expression():
         schema_expression(DayTimeIntervalType(), False)
         == "INTERVAL '1 01:01:01.0001' DAY TO SECOND"
     )
+
+
+class TestFormatYearMonthIntervalForDisplay:
```
sfc-gh-yuwang: nint: can we make the test follow the existing style in this suite?

sfc-gh-yuwang (approved):
Approved with a nint, please take a look before merge, thanks!

PR #4217, opened 2026-05-07
SNOW-3484790: initialize aggregation functions list during SCOS init

sfc-gh-yuwang:

   

   Fixes SNOW-NNNNNNN

   _retrieve_aggregation_function_list fires two blocking Snowflake queries on the first filter() call. The function is added to fix a bug in SCOS in previous PR. This PR meant to file async query instead at the beginning of snowpark session init. 
   This change also remove the redundant select from infromation_schema.functions query as currently UDAF is not supported in scos and will not be used.


sfc-gh-yuwang:
the updated system function list is fetched with query:
`SHOW FUNCTIONS ->> SELECT LISTAGG('"' || LOWER("name") || '"', ',\n') WITHIN GROUP (ORDER BY LOWER("name")) AS result FROM $1 WHERE "is_aggregate" = 'Y';`

src/snowflake/snowpark/session.py
```diff
@@ -5073,22 +5073,7 @@ def _retrieve_aggregation_function_list(self) -> None:
             )
 
         # System built-in aggregation functions
-        try:
-            retrieved_set.update(
-                {
-                    r[0].lower()
-                    for r in self.sql(
-                        """show functions ->> select "name" from $1 where "is_aggregate" = 'Y'"""
```
sfc-gh-joshi, line 5081: My understanding is that queries like this ensure the server is always the source of truth, and we don't have to make client code changes that are aligned with whatever Snowflake version is being run on the server. Hard-coding it to a local change is something of a philosophical change, and may unexpectedly break workloads in some circumstances.

It looks to only ever get run once per workload anyway, so I think the benefits of this would be pretty minimal.

sfc-gh-yuwang: I had a discussion with Adam and Yijun this morning. One of the issue is that this is always run when there are potentially agg function in the query, which means user would still see this even if they just issued 1 query in SCOS.
This is also a pretty expensive operation that took 3~4 seconds, which is a big portion when user workflow is small.

To solve the problem that this list may diverge from server, I created a followup ticket: https://snowflakecomputing.atlassian.net/browse/SNOW-3489271

Currently my thoughts is adding a step to release pipeline to ensure that each snowpark release's system function list is up-to-date at the point it released. Thus user would only need to use latest snowpark to avoid the mis align issue.

But I think it also make sense to add parameter protection for this. Will add one

sfc-gh-joshi: Parameter-protecting makes sense. I'm slightly worried about server divergence because the same client release might get run against different version backends (I'm not sure if we the N-1 version policy still exists, but at the very least testing vs. prod environments, and customers on different release cadences might get different versions)

src/snowflake/snowpark/session.py
```diff
@@ -5055,43 +5057,89 @@ def _retrieve_aggregation_function_list(self) -> None:
...
+                    "Unable to use async aggregation function prefetch: %s",
+                    e,
+                )
+            finally:
+                self._agg_function_prefetch_job = None
+        else:
             _logger.debug(
-                "Unable to get user-defined aggregation functions: %s",
-                e,
+                "Async aggregation function prefetch job is unavailable; using sync fallback."
```
sfc-gh-joshi: Why is this branch triggered if the prefetch job is `None`? Isn't it possible to reach this if another thread initiated the async job?

src/snowflake/snowpark/session.py
```diff
@@ -5055,43 +5057,89 @@ def _retrieve_aggregation_function_list(self) -> None:
...
+
         try:
-            retrieved_set.update(
-                {
-                    r[0].lower()
-                    for r in self.sql(
-                        """show functions ->> select "name" from $1 where "is_aggregate" = 'Y'"""
-                    ).collect()
-                }
+            self._agg_function_prefetch_job = self._submit_internal_async_prefetch_query(
```
sfc-gh-joshi: I think we need to guard this assignment with something like a c[ondition variable](https://docs.python.org/3/library/threading.html#condition-objects) to ensure we catch the case where multiple threads are trying to initiate (or wait on) the result of this async query.

sfc-gh-yuwang: can you give me an example of multiple threads are trying to initiate it at the same time?
I am asking this because this function is strictly called once per session life time during session init, I think there should be no race condition?

src/snowflake/snowpark/session.py
```diff
@@ -856,8 +856,15 @@ def __init__(
         self._dataframe_profiler = DataframeProfiler(session=self)
         self._catalog = None
         self._client_telemetry = EventTableTelemetry(session=self)
+        self._agg_function_prefetch_job: Optional[AsyncJob] = None
+        # Guards the one-time atomic claim of _agg_function_prefetch_job.
+        self._agg_function_prefetch_lock = Lock()
```
sfc-gh-joshi: I don't think there's a way for the same thread to attempt to acquire this lock multiple times, but I think we should make this an `RLock` instead (which is already imported in this file) to be safe.

src/snowflake/snowpark/session.py
```diff
@@ -5055,43 +5062,139 @@ def _retrieve_aggregation_function_list(self) -> None:
...
-                "falling back to hardcoded list: %s",
+                "Unable to start async aggregation metadata prefetch: %s",
                 e,
             )
-            retrieved_set.update(context._KNOWN_AGGREGATION_FUNCTIONS)
+            self._agg_function_prefetch_job = None
 
-        with context._aggregation_function_set_lock:
-            context._aggregation_function_set.update(retrieved_set)
+    def _submit_internal_async_prefetch_query(self, query: str) -> Optional[AsyncJob]:
```
sfc-gh-joshi: nit: Can we inline this method since it's only called once, and pretty short?

tests/unit/test_session.py
```diff
@@ -808,89 +808,373 @@ def test_infer_is_return_table_uses_internal_describe():
...
+    session = Session(fake_conn)
+    session._agg_function_prefetch_job = None
+    session._agg_function_fetch_event = None
+    return session, fake_conn, ctx
+
+
+def _ctx_setup(ctx):
+    ctx._is_snowpark_connect_compatible_mode = True
+    ctx._snowpark_connect_flatten_select_after_sort = True
+    ctx._aggregation_function_set = set()
```
sfc-gh-joshi: Why do we need this instead of mocking these fields like in the other test?

tests/unit/test_session.py
```diff
@@ -808,89 +808,373 @@ def test_infer_is_return_table_uses_internal_describe():
...
+
+
+def _ctx_restore(ctx, orig):
+    ctx._is_snowpark_connect_compatible_mode = orig[0]
+    ctx._snowpark_connect_flatten_select_after_sort = orig[1]
+    ctx._aggregation_function_set = orig[2]
+
+
+def test_retrieve_agg_concurrent_waiters_see_result_not_sync_query():
```
sfc-gh-joshi: How is this different from `test_concurrent_retrieve_agg_waiters_no_sync_query` in the other test file?

sfc-gh-joshi (approved):

src/snowflake/snowpark/context.py
```diff
@@ -45,6 +45,11 @@
     set()
 )  # lower cased names of aggregation functions, used in sql simplification
 _aggregation_function_set_lock = threading.RLock()
+_aggregation_function_prefetch_state: dict[str, Any] = {
```
sfc-gh-joshi, line 48: Does this need to be a dict? Can we just use 3 variables or a singleton class instead?

sfc-gh-joshi (approved):

src/snowflake/snowpark/context.py
sfc-gh-yuwang: yes we definitely can, I thought using a dictionary make it more clear that this is for the agg function prefetch

src/snowflake/snowpark/session.py
```diff
@@ -5055,43 +5056,116 @@ def _retrieve_aggregation_function_list(self) -> None:
...
+            else:
+                fetch_event = None
+                wait_event = None
+
+        if wait_event is not None:
+            # The query typically finishes in ~5s; 20s gives ample headroom while
+            # bounding the hang in the rare case the winner thread dies before its
+            # finally block runs (e.g. os._exit, interpreter shutdown).
+            wait_event.wait(timeout=20)
+            if context._aggregation_function_set:
```
sfc-gh-bkogan: I think we need to synchronize the access of `_aggregation_function_set` using `_aggregation_function_set_lock` since otherwise it's a data race.

T1: just finished async job, running `set.update`
T2: starting `filter` call, reading with `if set:`

if unsynchronized, we have undefined behavior. same thing for other reads of this var

src/snowflake/snowpark/session.py
```diff
@@ -5055,43 +5056,116 @@ def _retrieve_aggregation_function_list(self) -> None:
...
+                    _logger.debug(
+                        "Unable to get aggregation functions via sync fallback query: %s",
+                        e,
+                    )
 
-        with context._aggregation_function_set_lock:
-            context._aggregation_function_set.update(retrieved_set)
+            # Fallback to the local hardcoded list only when metadata retrieval fails.
+            if not system_fetch_succeeded:
+                retrieved_set.update(context._KNOWN_AGGREGATION_FUNCTIONS)
```
sfc-gh-bkogan, line 5128: if for the first owner of the job, it fails both the async and the follow up sync query, it will still set to this hardcoded list and all the waiters will use this one. is that fine? do we ever want to try refetching in this scenario?

sfc-gh-bkogan: there's also a race condition to think about here:

T1: high latency, takes 25 seconds for both async and follow up sync to end up failing
T2: waiter stops waiting for job 1, kicks off sync query, finishes in 5 seconds

both T1 and T2 finish at the same time, we dont know which result we are going to keep. if T2 got a good sync query, we dont want `_KNOWN_AGGREGATION_FUNCTIONS` to override it.

src/snowflake/snowpark/session.py
sfc-gh-yuwang: Thanks for spotting this, will fix

sfc-gh-bkogan (approved):

sfc-gh-yuwang:
with companion change: https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkConnectRegressionRunnerV2/115/
without companion change: 
https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkConnectRegressionRunnerV2/116/

PR #4231, opened 2026-05-21
Add Iceberg snapshot-id time travel support (version=)

sfc-gh-igarish:
## Summary

Adds Snowflake `AT(VERSION => <snapshot_id>)` SQL emission to the
Snowpark Python time-travel layer so that Iceberg snapshot-id time
travel works on managed Iceberg tables.

* Internal `TimeTravelConfig` gains a `version` field; the
  `generate_sql_clause()` helper emits `AT(VERSION => N)` when set.
* For Spark Iceberg compatibility,
  `DataFrameReader.option("snapshot-id", N)` (and the `snapshot_id`
  variant) is accepted: the helper `_extract_time_travel_from_options`
  aliases it to the new internal `version` field and auto-sets
  `time_travel_mode="at"`. String snapshot ids are coerced to `int`.
* `version` is passed through `**kwargs` on the public surfaces
  (`Session.table`, `DataFrameReader.table`, `Table`) so it isn't
  surfaced as a first-class kwarg yet — keeping the public API
  intentionally narrow while the underlying server-side feature
  (`FEATURE_ICEBERG_TIME_TRAVEL`) is still rolling out across
  deployments.
* Removed the redundant license header at the top of
  `_internal/options.py` flagged by @sfc-gh-mlieu.

## Validation

* `version` requires `time_travel_mode='at'` (matches Snowflake
  grammar and Spark Iceberg "snapshot N" semantics, not "before N")
* `version` must be `int`; `bool` is explicitly rejected
* Exactly one of `statement` / `offset` / `timestamp` / `stream` /
  `version` may be set per read

Mirrors the existing `iceberg_tag` pattern. Tests in
`tests/unit/test_utils.py` cover SQL emission, validation errors,
and the Spark-compat option aliases.

## End-to-end verification

End-to-end behavior is verified through the downstream Snowpark
Connect (SCOS) PR [snowflake-eng/sas#4143][scos_pr], which wires
`spark.read.format("iceberg").option("snapshot-id", N).load(...)` →
`session.read.option("snapshot-id", N).table(...)` →
`AT(VERSION => N)` SQL emitted by this PR.

[scos_pr]: https://github.com/snowflake-eng/sas/pull/4143

Sample integration test (`tests/sas_tests/test_iceberg_snapshot_id_sample.py`,
opt-in via `SCOS_ICEBERG_TIME_TRAVEL_SAMPLE=1`) run against `sfctest0`
with `FEATURE_ICEBERG_TIME_TRAVEL` enabled, exercising both Unity and
Glue catalog-linked databases:

```
$ SCOS_ICEBERG_TIME_TRAVEL_SAMPLE=1 .venv/bin/pytest \
    tests/sas_tests/test_iceberg_snapshot_id_sample.py -s -v
...
[unity] cldUnity.scosschema.iceberg_snapshot_id_unity_<token> snapshot ids (oldest -> newest):
  [6640031493841049519, 6660085326931362766, 4700846704791762084]
[unity] Iceberg snapshot-id time travel sample PASSED for
  cldUnity.scosschema.iceberg_snapshot_id_unity_<token>:
    S1=6640031493841049519 (2 rows)
    S2=6660085326931362766 (4 rows)
    S3=4700846704791762084 (6 rows)

[glue]  cldglue.scosschema.iceberg_snapshot_id_glue_<token> snapshot ids (oldest -> newest):
  [3883438140153085040, 7571520968419837863, 1414581987682063595]
[glue]  Iceberg snapshot-id time travel sample PASSED for
  cldglue.scosschema.iceberg_snapshot_id_glue_<token>:
    S1=3883438140153085040 (2 rows)
    S2=7571520968419837863 (4 rows)
    S3=1414581987682063595 (6 rows)

======================== 2 passed, 2 warnings in 44.18s ========================
```

For each table the sample:
1. Creates an Iceberg table and commits three INSERT snapshots
2. Reads the current state via
   `spark.read.format("iceberg").load(fqn)` (6 rows)
3. Reads each historical snapshot via
   `spark.read.format("iceberg").option("snapshot-id", S<N>).load(fqn)`
   and asserts the row count matches the expected historical state
   (2 / 4 / 6 for S1 / S2 / S3 respectively)

   Fixes SNOW-NNNNNNN

   This change adds `version` as an internal `TimeTravelConfig`
   field and accepts the Spark-Iceberg `snapshot-id` reader option as
   an alias. Both surface SQL emission through the existing
   `generate_sql_clause()` path, producing `AT(VERSION => N)` for
   managed Iceberg tables that have the server-side
   `FEATURE_ICEBERG_TIME_TRAVEL` parameter enabled.


sfc-gh-mayliu (commented):
Is there a Snowflake doc explaining that the `AT(VERSION=> <id>)` syntax is officially supported? Is `VERSION`/`snapshot-id` specific to iceberg tables only? It'd be great to get more context for the review.

Also, could you add some integ tests in `test_dataframe.py` for E2E validation?

sfc-gh-igarish:
> Is there a Snowflake doc explaining that the `AT(VERSION=> <id>)` syntax is officially supported? Is `VERSION`/`snapshot-id` specific to iceberg tables only? It'd be great to get more context for the review.
> 
> Also, could you add some integ tests in `test_dataframe.py` for E2E validation?

Yes this is for unmanaged Iceberg table only.

https://docs.google.com/document/d/1Z7GlEWbqZRCYr4NDjh-7klgvJJAp0jF45p9GH1MbQ4M/edit?tab=t.0

sfc-gh-mayliu (commented):
As per offline discussions, please look into:
1. add parameter protection for this change to isolate downstream SCOS use and Snowpark-python usage
2. add integ tests to validate E2E behaviors; if Iceberg credentials cannot be resolved systematically, mark it with `pytest.skip` but reusable for manual testing

sfc-gh-igarish:
> As per offline discussions, please look into:
> 
> 1. add parameter protection for this change to isolate downstream SCOS use and Snowpark-python usage
> 2. add integ tests to validate E2E behaviors; if Iceberg credentials cannot be resolved systematically, mark it with `pytest.skip` but reusable for manual testing

@sfc-gh-mayliu thanks for the review. Pushed commit `696b37e5d` addressing both review points and the codecov patch-coverage drop.

**1. Snowflake doc / scope of `AT(VERSION => N)`**

Yes — this syntax is the supported way to do snapshot-id time travel on **unmanaged Iceberg tables** in Catalog-Linked Databases (Unity / Glue). Spec lives in [this design doc](https://docs.google.com/document/d/1Z7GlEWbqZRCYr4NDjh-7klgvJJAp0jF45p9GH1MbQ4M/edit?tab=t.0) and is gated server-side behind the `FEATURE_ICEBERG_TIME_TRAVEL` account parameter. Does not apply to Snowflake-managed Iceberg tables or to non-Iceberg tables.

**2. Gating the Snowpark surface**

Because the underlying Snowflake feature is itself account-gated, this PR now also gates the Snowpark API behind a session flag, default `False`:

\`\`\`python
session.iceberg_features_enabled = True   # opt-in
session.table("t", time_travel_mode="at", version=N)
session.read.option("snapshot-id", N).table("t")
\`\`\`

With the flag off, all three entry points (`Session.table`, `DataFrameReader.table` with either `version=` or `snapshot-id`/`snapshot_id` options, and direct `Table(...)` construction) raise `SnowparkClientException` pointing to the flag. SCOS / snowpark-connect flips the flag per-session via its `snowpark.connect.iceberg.timeTravel.enabled` config.

**3. Integ tests in `test_dataframe.py`**

Added three E2E tests (`@pytest.mark.skip` for now):

- `test_iceberg_snapshot_id_time_travel_session_table` — `Session.table(..., version=N)`
- `test_iceberg_snapshot_id_time_travel_dataframe_reader_option` — `session.read.option('snapshot-id', N).table(...)` matches the explicit kwarg path
- `test_iceberg_snapshot_id_flag_gates_version_kwarg` — flag-off raises before any SQL is issued

Skipped because they need a CLD (`cldUnity` / `cldglue`) with a multi-snapshot unmanaged Iceberg table AND `FEATURE_ICEBERG_TIME_TRAVEL` enabled on the account — neither of which the standard integ accounts have today. Run manually against `sfctest0`. A TODO above the block tracks unskipping once we wire up an Iceberg-capable integ account.

**4. Codecov patch coverage**

Added `tests/unit/test_iceberg_features_flag.py` (13 tests). Three of those patch `with_src_position` to return a stand-in with a `version` field, exercising the `hasattr(ast, "version")`-guarded `ast.version.value = version` lines in `session.py`, `dataframe_reader.py`, and `table.py` — the exact lines codecov flagged. Confirmed locally via `coverage json` that all three are now covered.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -568,6 +604,11 @@ def table(
             timestamp_type: Type of timestamp interpretation ('NTZ', 'LTZ', or 'TZ').
                 Can also be set via ``option("timestamp_type", "LTZ")``.
             stream: Stream name for time travel. Can also be set via ``option("stream", "stream_name")``.
+            version: Iceberg snapshot id (64-bit integer) for snapshot-based time
+                travel on Iceberg tables. Can also be set via
+                ``option("snapshot-id", 5129038471029384756)`` (Spark Iceberg
+                naming) or ``option("version", 5129038471029384756)``.
+                Automatically sets time_travel_mode='at'.
```
sfc-gh-mayliu: the explicit `option("version", ...)` path doesn't auto-set mode. If a user follows the current `option("version", ...)` doc literally, they still need to supply `time_travel_mode="at"`

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -605,6 +652,12 @@ def table(
             ...     .option("offset", -60)                 # This will be IGNORED
             ...     .table("my_table", time_travel_mode="at", offset=-3600))  # Only this is used
         """
+        if version is not None or any(
+            self._cur_options.get(k) is not None for k in ("SNAPSHOT-ID", "SNAPSHOT_ID")
```
sfc-gh-mayliu: shall we check `version` here as well?

tests/integ/test_dataframe.py
```diff
@@ -8326,3 +8326,95 @@ def test_time_travel_comprehensive_coverage(session):
     finally:
         Utils.drop_table(session, table1_name)
         Utils.drop_table(session, table2_name)
+
+
+# ----------------------------------------------------------------------
+# Iceberg snapshot id (``version=``) time travel.
+#
+# TODO(SNOW-NNNNNNN): Wire these up to a CI test account that has:
```
sfc-gh-mayliu: Do we have an existing JIRA to check this CLD dependency? Can we swap it with the actual JIRA for follow up?

src/snowflake/snowpark/table.py
```diff
@@ -320,6 +326,11 @@ def __init__(
                 ast.timestamp_type.value = str(timestamp_type)
             if stream is not None:
                 ast.stream.value = stream
+            # Guard with hasattr — the AST proto field is added in a separate
+            # proto change. Without it we still produce correct SQL; only AST
+            # replay/telemetry would lack the version value.
+            if version is not None and hasattr(ast, "version"):
+                ast.version.value = version
```
sfc-gh-mayliu: If version/snapshot-id is parameter protected and not exposed to Snowpark users, we might not need to modify AST. cc: @sfc-gh-heshah

tests/integ/test_dataframe.py
```diff
@@ -8326,3 +8326,95 @@ def test_time_travel_comprehensive_coverage(session):
...
+    assert via_kwarg == via_option
+
+
+@pytest.mark.skip(
+    reason=(
+        "Requires a regular Snowflake account; doesn't need a CLD. "
+        "Tested manually until we wire up an Iceberg-capable integ account."
+    )
+)
+def test_iceberg_snapshot_id_flag_gates_version_kwarg(session):
```
sfc-gh-mayliu: the flag-off path would raise and shouldn't need a CLD. can we unskip this one?

CHANGELOG.md
```diff
@@ -4,6 +4,19 @@
 
 ### Snowpark Python API Updates
 
+#### New Features
+
+- Added support for Iceberg snapshot id time travel via the `version` parameter
```
sfc-gh-mayliu: If `version` is parameter protected and only available to SCOS, we might not need to edit `changelog.md` as it'd add noise to snowpark users

sfc-gh-mayliu (commented):
Thanks for protecting the code with flag, it makes the change much safer. Have a few more thoughts below. Also, could you please attach a manual test result for the skipped integ test that requires CLD?

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-igarish: Good catch — fixed in ce019767d. `_extract_time_travel_from_options` now treats all three aliases (`snapshot-id` / `snapshot_id` / `version`) the same: auto-set `time_travel_mode='at'`, reject `'before'`, and reject mixing with each other. The docstring claim "Automatically sets time_travel_mode='at'" now holds for every documented path. Unit coverage in `test_extract_time_travel_snapshot_id_option` now exercises the `option("version", N)` auto-set and the `option("version", N) + before` rejection paths.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-igarish: Yes, you're right — fixed in ce019767d. The iceberg gate now also includes `VERSION`:

```python
if version is not None or any(
    self._cur_options.get(k) is not None
    for k in ("VERSION", "SNAPSHOT-ID", "SNAPSHOT_ID")
):
    self._session._require_iceberg_features_enabled(...)
```

The parametrize in `test_dataframe_reader_table_rejects_snapshot_id_option_when_flag_off` now runs over `["snapshot-id", "snapshot_id", "version"]` so the bypass can't regress.

tests/integ/test_dataframe.py
sfc-gh-igarish: Yes — swapped in ce019767d. The placeholder now points to [SNOW-3525585 — Support Iceberg Snapshot ID in the SCOS](https://snowflakecomputing.atlassian.net/browse/SNOW-3525585), which is the parent story for this work. I'll wire the integ-test plumbing under that JIRA.

src/snowflake/snowpark/table.py
sfc-gh-igarish: Agreed — removed in ce019767d. The Table / ReadTable protos don't carry a `version` field, the `hasattr(ast, "version")` guards were dead today, and the feature is parameter-protected (gated behind `iceberg_features_enabled`, only Snowpark Connect flips it), so the AST replay/telemetry case isn't there yet. Dropped the assignment in all three call sites (`Session.table`, `DataFrameReader.table`, `Table.__init__`) and the matching coverage tests in `test_iceberg_features_flag.py`. Each call site has a one-line comment pointing to the single `ast.version.value = version` restore when the proto is extended.

tests/integ/test_dataframe.py
sfc-gh-igarish: Yes, you're right — unskipped in ce019767d. The flag-off path raises client-side before any SQL is emitted, so the test only needs a session (any session). I also broadened it to cover the reader paths so the gate is exercised end-to-end: `session.table(version=...)`, `session.read.table(version=...)`, `session.read.option("snapshot-id", ...).table(...)`, and `session.read.option("version", ...).table(...)`. The original umbrella unit coverage in `test_iceberg_features_flag.py` still pins this with mocks too.

CHANGELOG.md
sfc-gh-igarish: Good point — removed in ce019767d. Since the feature defaults off and is only consumed by Snowpark Connect / SAS (which flips the flag per-session from its own config), there's no behavior visible to a plain snowpark-python user reading release notes. The CHANGELOG is back to the upstream `Documentation`-only entry from #4232. Happy to revisit and add a note when we eventually flip the default or expose a non-SCOS entry point.

tests/unit/test_iceberg_features_flag.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
+#
+
+#
+# Copyright (c) 2012-2026 Snowflake Computing Inc. All rights reserved.
+#
```
sfc-gh-mayliu: nit: 2 copyright headers

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -183,6 +192,42 @@ def _extract_time_travel_from_options(options: dict) -> dict:
...
         excluded_keys.add("TIMESTAMP")
 
+    # Handle Iceberg snapshot id (Spark `snapshot-id` / `snapshot_id` plus the
+    # Snowpark-native `version` option). All three aliases auto-set mode='at'
+    # since `AT(VERSION => N)` is the only valid Snowflake form.
+    snapshot_id_value = options.get("SNAPSHOT-ID")
+    snapshot_id_source = "snapshot-id"
+    if snapshot_id_value is None:
+        snapshot_id_value = options.get("SNAPSHOT_ID")
+        snapshot_id_source = "snapshot_id"
```
sfc-gh-mayliu, line 199: What if user sets both `SNAPSHOT-ID` and `SNAPSHOT_ID`? If one of the values will be silently dropped, shall we add another conflict check so that options cannot be mixed?

src/snowflake/snowpark/session.py
sfc-gh-aling, line 1: why we need a new property`iceberg_features_enabled`?
`version=`` / ``snapshot-id` are new options that snowpark users can also benefit from right?

I think we can expose them to snowpark users directly without the parameter protection which gives extra mental burden to users

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-igarish: it's like if user do option("snapshot-id",1234).option("snapshot-id",3456) one will drop. These are edage cases , in theory can be possible but we don't need to code it until we test user workload.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -625,15 +689,28 @@ def table(
...
+            # NOTE: ``version`` is intentionally NOT emitted to the AST. The
+            # ReadTable proto has no ``version`` field and the feature is
+            # parameter-protected (gated behind `iceberg_features_enabled`,
+            # consumed by Snowpark Connect only). When the proto is extended,
+            # restore a single ``ast.version.value = version`` line here.
+
+        if time_travel_mode is not None or version is not None:
+            # If version is provided without mode, default to 'at' (snapshot ids
+            # only make sense with AT — symmetric with iceberg_tag handling).
+            effective_mode = (
```
sfc-gh-aling, line 678: is there a chance default to `at` would conflict other parameter setting?

sfc-gh-aling: this issue will be gone if we move into kwargs

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -549,6 +594,7 @@ def table(
         timestamp: Optional[Union[str, datetime]] = None,
         timestamp_type: Optional[Union[str, TimestampTimeZone]] = None,
         stream: Optional[str] = None,
+        version: Optional[int] = None,
```
sfc-gh-aling: as per discussion, we can hide arguments from public api by moving them into `**kwargs` and in the code we can pop them out `version = kwargs.pop("version")`

src/snowflake/snowpark/session.py
sfc-gh-igarish: Removed in 7563e8244. After offline discussion we landed on the middle path you suggested: drop the `iceberg_features_enabled` umbrella flag entirely (and the `_require_iceberg_features_enabled` helper), and at the same time keep `version=` itself a soft-internal/experimental surface by hiding it behind `**kwargs` per Aling's follow-up suggestion below.

Net effect:
- No mental burden for Snowpark users — no opt-in flag, no `SnowparkClientException` when they happen to read a CLD Iceberg table.
- `version=` works for anyone who passes it (Snowpark Connect, and any direct user who learns about it), but it doesn't show up in the public signature, docstrings, doctest examples, or in `option("version", ...)`. The Spark Iceberg `option("snapshot-id", ...)` / `option("snapshot_id", ...)` paths remain (those are external Spark-compat and stay documented).

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-igarish: Applied in 7563e8244. All three entry points now use `**kwargs` and pop `version` inside the method:

```python
version = kwargs.pop("version", None)
if kwargs:
    raise TypeError(f"table() got unexpected keyword arguments: {sorted(kwargs)}")
```

`version` is gone from the public signatures of `Session.table` / `DataFrameReader.table` / `Table.__init__`, and from all their docstrings / doctest examples. The `option("version", ...)` alias is removed too (only `snapshot-id` / `snapshot_id` remain as Spark-compat options). Unit tests for `Session` / `DataFrameReader` / `Table` (88+1 skipped) and `test_utils.py` (15) all pass locally.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-igarish: Confirmed gone in 7563e8244 — `version` is now popped from `**kwargs`, so the `effective_mode` short-circuit in `DataFrameReader.table` only runs when the caller explicitly supplied it via the hidden kwarg, and there's no kwarg-default conflict with `time_travel_mode` (regular Snowpark users can't reach this branch from the public signature anymore).

tests/unit/test_iceberg_features_flag.py
sfc-gh-igarish: Good catch — fixed in 7563e8244 by deleting `tests/unit/test_iceberg_features_flag.py` entirely. The file existed only to test the `iceberg_features_enabled` umbrella flag and the per-entry-point gates, which were both removed in this commit per the offline discussion with Aling (see her thread above). So the duplicate header issue and the now-stale test coverage are both gone in one move.

sfc-gh-igarish:
Pushed 7563e8244 to address review feedback from @sfc-gh-aling and @sfc-gh-mayliu.

**1. Drop `iceberg_features_enabled` umbrella flag** (Aling)
The session property, setter, and `_require_iceberg_features_enabled` helper are removed, along with all per-entry-point gate calls in `Session.table` / `DataFrameReader.table` / `Table.__init__`. Net effect: no opt-in flag, no mental burden for Snowpark users, and no `SnowparkClientException` when they happen to read a CLD Iceberg table. The underlying `FEATURE_ICEBERG_TIME_TRAVEL` account parameter already provides server-side isolation, so a client-side gate was redundant.

**2. Move `version=` out of public signatures via `**kwargs`** (Aling)
All three entry points now use `version = kwargs.pop("version", None)` instead of an explicit kwarg. `version` is gone from public signatures, docstrings, doctest examples, and Sphinx-rendered API. The `option("version", ...)` alias is dropped too. The Spark Iceberg `option("snapshot-id", ...)` / `option("snapshot_id", ...)` paths remain (external Spark-compat surface, auto-set `time_travel_mode='at'` exactly as before).

**3. Duplicate copyright header in `test_iceberg_features_flag.py`** (May)
File removed entirely (only tested the now-removed flag).

**4. Tests**
- `tests/unit/test_iceberg_features_flag.py` deleted.
- `tests/unit/test_utils.py::test_extract_time_travel_snapshot_id_option` trimmed to cover only `snapshot-id` / `snapshot_id` cases.
- `tests/integ/test_dataframe.py::test_iceberg_snapshot_id_flag_gates_version_kwarg` removed (no flag to gate). The two skipped CLD-required E2E tests stay, minus the `iceberg_features_enabled = True` setup lines.

Local: `tests/unit/test_utils.py` (15), `tests/unit/test_dataframe.py` + `tests/unit/test_session.py` + `tests/unit/test_dataframe_reader_type_parsing.py` (237+1 skipped) all green.

Diff is now `+36 / −313` (mostly removal). PTAL when you have a moment, thanks!

sfc-gh-aling (approved):

sfc-gh-mayliu (approved):
LGTM, thanks for supporting the new time travel options for iceberg

sfc-gh-igarish:
**CI note — 6 failing tests are unrelated to this PR**

The three failing `Test py-ubuntu-latest-64-cores-3.{9,13,14}-aws` jobs in [run 26541665563](https://github.com/snowflakedb/snowpark-python/actions/runs/26541665563) all hit the same 6 failures in `tests/integ/test_dataframe_ai.py`:

- `test_dataframe_ai_split_text_markdown_header_basic`
- `test_dataframe_ai_split_text_markdown_header_default_output`
- `test_dataframe_ai_split_text_recursive_character_basic`
- `test_dataframe_ai_split_text_recursive_character_markdown_format`
- `test_dataframe_ai_split_text_recursive_character_custom_separators`
- `test_dataframe_ai_split_text_recursive_character_default_output`

All six raise the same SQL compilation error:

```
snowflake.snowpark.exceptions.SnowparkSQLException: (1304): ... 002001 (02000):
SQL compilation error:
Object 'snowflake.snowpark.anaconda_shared_repository' does not exist or not authorized.
```

This is an environmental issue with the CI Snowflake account (missing / unauthorized `anaconda_shared_repository` object that `DataFrameAIFunctions.split_text_*` looks up), not a code regression from this PR:

- This PR's diff (`+36 / −313`) touches only the Iceberg snapshot-id time travel path (`session.py`, `dataframe_reader.py`, `table.py`, `_internal/utils.py`) and their tests. None of `dataframe_ai_functions.py` / `test_dataframe_ai.py` / any Cortex code is modified.
- The previous commit on this branch (`ce019767d`, before today's push) passed the same `py-ubuntu-latest-64-cores-3.14-aws` job cleanly.
- `main` already has precedent for these AI tests being flaky — `dd6d6b3bb` ("SNOW-3201438: temporarily skipping failing ai function tests") landed an explicit skip-list for the same family of tests; the `split_text_*` cases appear to be a second wave that slipped past that earlier skip.

I've triggered a re-run of the three failed jobs in case the underlying object access was transient. If they fail the same way again, this is firmly a `dataframe_ai` ownership / CI-account fix, not something to gate this PR on.

PR #4234, opened 2026-05-21
SNOW-3081273: fix Creating a stored procedure locally results in an unsupported version of cloudpicker

sfc-gh-yuwang:

   

   Fixes SNOW-3081273

   This PR fixes a Python 3.13 local-registration issue for Snowpark UDFs/stored procedures by changing auto-injected cloudpickle dependency specs from strict pins (==) to compatible lower bounds (>=), so server-side resolution can choose a supported version instead of failing on an exact local version match.

This change is safe because cloudpickle is backward compatible, meaning higher version can deserialize encoding create by lower version

### DIFF
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -2133,7 +2133,7 @@ class Session:
             if isinstance(m, str) and m not in result_dict:
                 res.append(m)
             elif isinstance(m, ModuleType) and m.__name__ not in result_dict:
-                res.append(f"{m.__name__}=={m.__version__}")
+                res.append(f"{m.__name__}>={m.__version__}")
 
         return res
 
--- a/src/snowflake/snowpark/session.py
+++ b/src/snowflake/snowpark/session.py
@@ -2134,7 +2134,10 @@ class Session:
             if isinstance(m, str) and m not in result_dict:
                 res.append(m)
             elif isinstance(m, ModuleType) and m.__name__ not in result_dict:
-                res.append(f"{m.__name__}>={m.__version__}")
+                if m.__name__ == "cloudpickle":
+                    res.append(f"{m.__name__}>={m.__version__}")
+                else:
+                    res.append(f"{m.__name__}=={m.__version__}")
 
         return res
 
--- a/tests/integ/test_stored_procedure.py
+++ b/tests/integ/test_stored_procedure.py
@@ -2736,3 +2736,27 @@ def test_data_source_udtf_ingestion(db_parameters):
             ]
 
         assert normalize(df.collect()) == normalize(oracledb_real_data)
+
+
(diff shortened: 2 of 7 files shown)


sfc-gh-yuwang:
https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/PythonStoredProcBuildSnowfortTest/957/console
validation run have no extra failure

tests/integ/test_stored_procedure.py
```diff
@@ -2736,3 +2736,27 @@ def normalize(rows):
             ]
 
         assert normalize(df.collect()) == normalize(oracledb_real_data)
+
+
+@pytest.mark.skipif(IS_IN_STORED_PROC, reason="not supported in stored proc")
+def test_sproc_runtime_313_cloudpickle_ge_spec_compiles_and_executes(session):
```
sfc-gh-mayliu, line 2742: Not blocking: since this test specifies python==3.13, wondering whether other test matrices with 3.11, 3.12 etc would execute this? Would this be potentially flaky for other python versions, though the merge gate seems to be passing here?

src/snowflake/snowpark/session.py
```diff
@@ -2134,7 +2134,7 @@ def _get_req_identifiers_list(
             if isinstance(m, str) and m not in result_dict:
                 res.append(m)
             elif isinstance(m, ModuleType) and m.__name__ not in result_dict:
-                res.append(f"{m.__name__}=={m.__version__}")
+                res.append(f"{m.__name__}>={m.__version__}")
```
sfc-gh-mayliu: Seems like `cloudpickle` is the only `ModuleType` that goes through this check now, so we're good. but `>=` applies to any module a future caller might pass. Can we maybe add a one-line comment noting the intent so nobody fixes it back to `==`, or limit the `>=` logic to `cloudpickle` only?

sfc-gh-mayliu (approved):

src/snowflake/snowpark/session.py
sfc-gh-yuwang: great call out! I'll limit this to only cloudpickle

tests/integ/test_stored_procedure.py
sfc-gh-yuwang: I think this should be good even if 3.11/3.12 execute this test.
so this change is only making sure that cloudpickle can be correctly picked in stored proc. Say if 3.12 execute this with cloudpickle 2.2.1, it will try to pick cloudpickle >=2.2.1 for py313, as long as this cloudpickle pkg exist on server, the test would not fail. For the situation the test do fail, I think that is more of a packaging issue instead of flakiness

sfc-gh-jrose (approved):

PR #4236, opened 2026-05-22
SNOW-3556267: Fix AttributeError when write_pandas receives list name…

JonathanStefanov:
   Fixes SNOW-3556267

When Session.write_pandas is called with database= and schema= kwargs, _write_modin_pandas_helper builds a list like ["MY_DB",  "MY_SCHEMA", "MY_TABLE"] and passes it as name to to_snowflake. The below-threshold path re-routes to the Snowflake backend which unpacks that list correctly. The Parquet path passed the whole list to a function that called .upper() on it.

write_pandas already accepts table_name, schema, and database as separate string arguments. So the  #fix splits the list before calling it: last element is always the table name, second-to-last is the schema (if present), first is the database (if 3 parts). Each part is individually converted from Snowflake identifier format to the plain string write_pandas expects.

### DIFF
--- a/src/snowflake/snowpark/modin/plugin/extensions/utils.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/utils.py
@@ -838,7 +838,7 @@ def pandas_to_snowflake(
     )
     database_converted = (
         _convert_to_snowflake_table_name_to_write_pandas_table_name(name_parts[0])
-        if len(name_parts) >= 3
+        if len(name_parts) == 3
         else None
     )
 
--- a/src/snowflake/snowpark/modin/plugin/extensions/utils.py
+++ b/src/snowflake/snowpark/modin/plugin/extensions/utils.py
@@ -824,6 +824,10 @@ def pandas_to_snowflake(
         )
 
     name_parts = [name] if isinstance(name, str) else list(name)
+    if len(name_parts) > 3:
+        raise ValueError(
+            f"name must have at most 3 parts (database, schema, table), got {len(name_parts)}: {name_parts}"
+        )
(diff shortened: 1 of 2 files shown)


JonathanStefanov:

src/snowflake/snowpark/modin/plugin/extensions/utils.py
```diff
@@ -823,6 +823,21 @@ def pandas_to_snowflake(
...
+        name_parts[-1]
+    )
+    schema_converted = (
+        _convert_to_snowflake_table_name_to_write_pandas_table_name(name_parts[-2])
+        if len(name_parts) >= 2
+        else None
+    )
+    database_converted = (
+        _convert_to_snowflake_table_name_to_write_pandas_table_name(name_parts[0])
+        if len(name_parts) >= 3
```
Eagleseb: should we add a check for name_parts <= 3 and raise an error otherwise?

Eagleseb (approved):

src/snowflake/snowpark/modin/plugin/extensions/utils.py
JonathanStefanov: Agreed, thanks for the suggestion

src/snowflake/snowpark/modin/plugin/extensions/utils.py
```diff
@@ -823,6 +823,25 @@ def pandas_to_snowflake(
...
```
GaspardMerten: Should probably be equal here.

Eagleseb (approved):

PR #649, opened 2023-01-10
SNOW-697912: Fix SQL simplifier for renaming column names

sfc-gh-yixie:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #SNOW-697912

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am adding new credentials
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.
When sql simplifier is enabled, a DataFrame's and its previous level DataFrame (the subquery) share the same `Attribute` instances. This generates a wrong SQL in a specific case. The fix is to copy those `Attribute` instances.

PR #4245, opened 2026-05-29
SNOW-3485471: cte dedup for self-join case

sfc-gh-aling:

   

  In SCOS, spark.table() translates into SnowflakePlan nodes with non-empty expr_to_alias. The TPCDS_Q39 query aliases the same inventory DataFrame as inv1 and inv2 and self-joins them. The join path
   deep-copies the plan for each branch and calls add_aliases separately, resulting in both branches carrying the same alias values under different UUID keys. The old key-sort hashing produced
  different hashes for the two branches, so the CTE optimizer did not detect them as duplicates and inlined the full aggregation pipeline twice. The value-sort fix correctly identifies them as 
  the same computation and emits a single CTE. 

- hash node by expr_to_alias value instead expr_to_alias items
- conflicting alias detection to avoid over optimization

### DIFF
--- a/src/snowflake/snowpark/_internal/compiler/cte_utils.py
+++ b/src/snowflake/snowpark/_internal/compiler/cte_utils.py
@@ -307,7 +307,14 @@ def encode_query_id(node: "TreeNode") -> Optional[str]:
     if query_params:
         string = f"{string}#{query_params}"
     if hasattr(node, "expr_to_alias") and node.expr_to_alias:
-        string = f"{string}#{stringify(node.expr_to_alias)}"
+        # Sort by alias values (not UUID keys) so that two nodes representing
+        # the same computation hash identically even when the expression UUID
+        # keys differ (e.g. two deep-copied SelectStatement nodes for inv1/inv2
+        # in a self-join accumulate different UUID keys via add_aliases but carry
+        # the same set of alias values).  Different alias values (e.g. when
+        # _disambiguate adds a join suffix such as "_WITH_AD_GROUP") still
+        # produce different hashes, preserving the SNOW-2261400 fix.
+        string = f"{string}#{sorted(set(node.expr_to_alias.values()))}"
     if (
         hasattr(node, "df_aliased_col_name_to_real_col_name")
         and node.df_aliased_col_name_to_real_col_name
--- a/src/snowflake/snowpark/_internal/compiler/cte_utils.py
+++ b/src/snowflake/snowpark/_internal/compiler/cte_utils.py
@@ -307,13 +307,11 @@ def encode_query_id(node: "TreeNode") -> Optional[str]:
     if query_params:
         string = f"{string}#{query_params}"
     if hasattr(node, "expr_to_alias") and node.expr_to_alias:
-        # Sort by alias values (not UUID keys) so that two nodes representing
-        # the same computation hash identically even when the expression UUID
-        # keys differ (e.g. two deep-copied SelectStatement nodes for inv1/inv2
-        # in a self-join accumulate different UUID keys via add_aliases but carry
-        # the same set of alias values).  Different alias values (e.g. when
-        # _disambiguate adds a join suffix such as "_WITH_AD_GROUP") still
-        # produce different hashes, preserving the SNOW-2261400 fix.
+        # Hash by alias values only, not the UUID keys, since UUID keys are regenerated on every deep-copy/re-resolve (e.g. the two
+        # branches of a self-join). This lets nodes representing the same computation hash identically, enabling CTE dedup for self-joins.
+        # NOTE: since nodes with different UUID keys can now share a CTE, _replace_duplicate_node_with_cte must merge each duplicate's
+        # UUID→alias entries into the shared CTE so parent re-resolution can resolve any UUID variant (see companion comment there).
+        # Different alias values (e.g. a "_WITH_AD_GROUP" join suffix from _disambiguate) still hash differently, preserving SNOW-2261400.
         string = f"{string}#{sorted(set(node.expr_to_alias.values()))}"
     if (
         hasattr(node, "df_aliased_col_name_to_real_col_name")
--- a/src/snowflake/snowpark/_internal/compiler/repeated_subquery_elimination.py
+++ b/src/snowflake/snowpark/_internal/compiler/repeated_subquery_elimination.py
@@ -168,6 +168,15 @@ class RepeatedSubqueryElimination:
                     resolved_with_block = resolved_with_block_map[
                         node.encoded_node_id_with_query
                     ]
+                    # The CTE was built from an earlier duplicate node that may have
+                    # different expression UUID keys (though the same alias values).
+                    # Merge this node's UUID→alias entries into the shared CTE plan so
+                    # that parent nodes referencing this node's UUIDs can still resolve
+                    # column aliases correctly after substitution.
+                    if hasattr(node, "expr_to_alias") and node.expr_to_alias:
+                        for k, v in node.expr_to_alias.items():
+                            if k not in resolved_with_block.expr_to_alias:
+                                resolved_with_block.expr_to_alias[k] = v
                 else:
                     if (
                         self._query_generator.session.reduce_describe_query_enabled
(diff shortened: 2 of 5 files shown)


src/snowflake/snowpark/_internal/compiler/cte_utils.py
```diff
@@ -307,7 +307,12 @@ def stringify(d):
     if query_params:
         string = f"{string}#{query_params}"
     if hasattr(node, "expr_to_alias") and node.expr_to_alias:
-        string = f"{string}#{stringify(node.expr_to_alias)}"
+        # Hash by alias values only, not the UUID keys, since UUID keys are regenerated on every deep-copy/re-resolve (e.g. the two
+        # branches of a self-join). This lets nodes representing the same computation hash identically, enabling CTE dedup for self-joins.
+        # NOTE: since nodes with different UUID keys can now share a CTE, _replace_duplicate_node_with_cte must merge each duplicate's
+        # UUID→alias entries into the shared CTE so parent re-resolution can resolve any UUID variant (see companion comment there).
+        # Different alias values (e.g. a "_WITH_AD_GROUP" join suffix from _disambiguate) still hash differently, preserving SNOW-2261400.
+        string = f"{string}#{sorted(set(node.expr_to_alias.values()))}"
```
sfc-gh-aalam, line 315: what kind of cases will this produce same hash-keys. When it is wrong, what kind of risk are we dealing with

tests/integ/test_cte.py
```diff
@@ -874,12 +878,15 @@ def test_sql_simplifier(session):
...
+        # same alias values, different UUID keys). df2 and df3 are replaced with a
+        # shared CTE, but df1's left-join position remains inline. That gives 2
+        # CTEs (base VALUES + filtered df1) and the filter appears twice (once in
+        # the CTE body, once inline for the left-join position).
         assert (
-            count_number_of_ctes(Utils.normalize_sql(df6.queries["queries"][-1])) == 1
+            count_number_of_ctes(Utils.normalize_sql(df6.queries["queries"][-1])) == 2
         )
-        assert Utils.normalize_sql(df6.queries["queries"][-1]).count(filter_clause) == 3
+        assert Utils.normalize_sql(df6.queries["queries"][-1]).count(filter_clause) == 2
```
sfc-gh-aling, line 889: for clarity, more CTE is better for this case, see the below comparison between the generated sql:

  Before (1 CTE, filter appears 3 times):      
```                                                                                                                                     
  WITH CTE_values AS (
    SELECT $1 AS "A", $2 AS "B" FROM VALUES (1,2),(3,4)                                                                                                                             
  )                                                                                                                                                                                 
  SELECT * FROM (                                                                                                                                                                   
    (                                                                                                                                                                               
      SELECT "A_XXX", "B_XXX", "A" AS "A_YYY", "B" AS "B_YYY"                                                                                                                       
      FROM (                                                                                                                                                                        
        (                                                                                                                                                                           
          SELECT "A" AS "A_XXX", "B" AS "B_XXX"                                                                                                                                     
          FROM CTE_values WHERE ("A" = 1)       -- filter #1 (df1, inline)                                                                                                          
        ) AS SNOWPARK_LEFT                                                                                                                                                          
        INNER JOIN (                                                                                                                                                                
          SELECT "A", "B"                                                                                                                                                           
          FROM CTE_values WHERE ("A" = 1)       -- filter #2 (df2, inline)                                                                                                          
        ) AS SNOWPARK_RIGHT                                                                                                                                                         
      )                                                           
    ) AS SNOWPARK_LEFT
    INNER JOIN (      
      SELECT "A", "B"
      FROM CTE_values WHERE ("A" = 1)           -- filter #3 (df3, inline)
    ) AS SNOWPARK_RIGHT                                                   
  )                                                                                                                                                                                 
```                                                                                                                                                                                    
  After (2 CTEs, filter appears 2 times):        
```                                                                                                                                   
  WITH CTE_values AS (                                                                                                                                                              
    SELECT $1 AS "A", $2 AS "B" FROM VALUES (1,2),(3,4)                                                                                                                             
  ),                                                                                                                                                                                
  CTE_filtered AS (                              -- df2 and df3 deduplicated here                                                                                                   
    SELECT "A", "B"                                                                                                                                                                 
    FROM CTE_values WHERE ("A" = 1)             -- filter #1 (in CTE body)                                                                                                          
  )                                                                                                                                                                                 
  SELECT * FROM (                                                                                                                                                                   
    (                                                             
      SELECT "A_XXX", "B_XXX", "A" AS "A_YYY", "B" AS "B_YYY"
      FROM (                                                 
        (   
          SELECT "A" AS "A_XXX", "B" AS "B_XXX"
          FROM CTE_values WHERE ("A" = 1)       -- filter #2 (df1, still inline)
        ) AS SNOWPARK_LEFT                                                      
        INNER JOIN (SELECT * FROM CTE_filtered) AS SNOWPARK_RIGHT                                                                                                                   
      )                                                                                                                                                                             
    ) AS SNOWPARK_LEFT                                                                                                                                                              
    INNER JOIN (SELECT * FROM CTE_filtered) AS SNOWPARK_RIGHT                                                                                                                       
  )   
```

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-aling: 1) I assume you are asking the hash-keys for a node -- same hash-keys are produced by nodes with same sql, sql params, df aliases, and **expr_to_alias values** (before the change it's by expr_to_alias.items().

(hash-keys for expr_to_alias are generated from `uuid`, it can rarely happen that two uuid collide)


2) the only risk is parent column resolution as we relax the standard of repeated node identification. I introduced extra logic to detect conflicting expr_to_alias (same expr maps to two different alias) in nodes to prevent CTE in this case. thank you for bringing it up

sfc-gh-aalam (approved):

src/snowflake/snowpark/_internal/compiler/repeated_subquery_elimination.py
```diff
@@ -187,6 +209,12 @@ def _update_parents(
                         node.encoded_node_id_with_query
                     ] = resolved_with_block
                     self._total_number_ctes += 1
+                elif getattr(node, "expr_to_alias", None):
+                    # reuse the existing CTE block. expr_ids are regenerated on copy, so
+                    # this node's keys differ from the node the CTE was built from; merge
+                    # this node's entries so every expr_id variant stays resolvable during
+                    # parent re-resolution.
+                    resolved_with_block.expr_to_alias.update(node.expr_to_alias)
```
sfc-gh-joshi, line 217: I'm not very familiar with CTE design, so this is just for my own understanding: why are expr_ids regenerated on copy in the first place? Would keeping the UUIDs identical violate some invariant?

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-joshi: Is there any chance of duplicate value entries here that would silently get swallowed by the `sorted(set(...))` call?

src/snowflake/snowpark/_internal/compiler/repeated_subquery_elimination.py
sfc-gh-aling: > expr_ids regenerated on copy in the first place? Would keeping the UUIDs identical violate some invariant?

it was to address a sql simplifier issue. same expr_id generates wrong sql. more details in the [PR#649](https://github.com/snowflakedb/snowpark-python/pull/649)

src/snowflake/snowpark/_internal/compiler/cte_utils.py
sfc-gh-aling: Yes, duplicate alias can happen but consolidating them is harmless in our case, it's not reflected in the newly-generated CTE optimized node, and the generated sql in the hash is enough for dup node detection in the self-join case.

the `expr_to_alias` is mostly for internal column name resolution in the downstream compilation stage.

theoretically I think this info shall be excluded from the hash computation of a node, but right now I keep it as a defensive layer to distinguish nodes.

sfc-gh-joshi (approved):

sfc-gh-joshi (approved):

PR #4248, opened 2026-06-02
SNOW-3485482: Eliminate unnecessary SELECT * from joins

sfc-gh-joshi:

   

   Fixes SNOW-3485482

Removes unnecessary SELECT * layers from repeated join operations. Truncated example SQL from Python code `df1.join(df2.join(df3, on="key", how="inner"), on="key", how="inner")`, edited for readability:

Current main:
```
SELECT * FROM (
  SELECT * FROM (
    (
      SELECT * FROM (
        (
          SELECT "KEY", "VAL_BASE"
          FROM (
            SELECT $1 AS "KEY", $2 AS "VAL_BASE" FROM table_name
        ) AS SNOWPARK_LEFT
        INNER JOIN (
          SELECT "KEY", "VAL_1" FROM (
            SELECT $1 AS "KEY", $2 AS "VAL_1" FROM table_name
        ) AS SNOWPARK_RIGHT
          USING (key)
      )
    ) AS SNOWPARK_LEFT
    INNER JOIN (
      SELECT "KEY", "VAL_2" FROM (
        SELECT $1 AS "KEY", $2 AS "VAL_2" FROM table_name
    ) AS SNOWPARK_RIGHT
      USING (key)
  )
)
WHERE "VAL_BASE" > 100
ORDER BY "VAL_BASE" DESC NULLS LAST
LIMIT 5
```

New:
```
SELECT * FROM (
  SELECT * FROM (
    (
      SELECT "KEY", "VAL_BASE" FROM table_name
    ) AS SNOWPARK_LEFT
    INNER JOIN (
      SELECT "KEY", "VAL_1" FROM table_name
    ) AS SNOWPARK_RIGHT
      USING (key)
    INNER JOIN (
      SELECT "KEY", "VAL_2"FROM table_name
    ) AS SNOWPARK_TEMP_TABLE_Y7YO05WRW5
      USING (key)
  )
)
WHERE "VAL_BASE" > 100
ORDER BY "VAL_BASE" DESC NULLS LAST
LIMIT 5
```

The outer-most layers of SELECT * cannot be removed, but this PR prevents the addition of a new layer of SELECT * wrapping on every additional chained join operatio. For a sequence of 10 joins, this PR reduces the query text from 3752 characters to 3528 (-6%), but does not meaningfully affect SQL compilation time.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -989,24 +989,30 @@ def snowflake_supported_join_statement(
     left_uuid: Optional[str] = None,
     right_uuid: Optional[str] = None,
     directed: bool = False,
+    left_is_join: bool = False,
 ) -> str:
     LEFT_UUID = format_uuid(left_uuid)
     RIGHT_UUID = format_uuid(right_uuid)
 
-    # If left is a simple SELECT * FROM (\n<join_source>\n) wrapper from a
-    # previous join, flatten into a multi-way join by appending the new right
-    # operand directly to the existing join source. This avoids nested
+    # If left is the output of a previous join, flatten into a multi-way join
+    # by unwrapping the SELECT * FROM (...) envelope and appending the new
+    # right operand directly to the existing join source. This avoids nested
     # SELECT * layers that inflate query text without changing semantics.
     #
     # Though it is technically less efficient than constructing the join sub-queries
     # without the SELECT in the first place, the structure of our SQL processing code
-    # top-level projections to be wrapped by a select.
-    unwrapped_left = _unwrap_select_star_from(left)
+    # needs top-level projections to be wrapped by a select to be well-formed, so we
+    # must strip it here instead.
+    #
+    # We only unwrap the left side because it is simpler to deal with than unwrapping
+    # both left and right, and left-deep chains are more common, as they're produced
+    # by calls like df1.join(df2).join(df3) etc.
+    unwrapped_left = _unwrap_select_star_from(left) if left_is_join else None
     right_alias = (
         "SNOWPARK_RIGHT"
+        if use_constant_subquery_alias and unwrapped_left is None
         # Multi-way join: right alias must be unique to avoid collisions
         # with aliases already present in the flattened join source.
-        if use_constant_subquery_alias and unwrapped_left is None
         else random_name_for_temp_object(TempObjectType.TABLE)
     )
 
@@ -1098,6 +1104,7 @@ def join_statement(
     left_uuid: Optional[str] = None,
     right_uuid: Optional[str] = None,
     directed: bool = False,
+    left_is_join: bool = False,
 ) -> str:
     if isinstance(join_type, (LeftSemi, LeftAnti)):
         return left_semi_or_anti_join_statement(
@@ -1125,6 +1132,7 @@ def join_statement(
         left_uuid=left_uuid,
         right_uuid=right_uuid,
         directed=directed,
+        left_is_join=left_is_join,
     )
 
 
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -992,14 +992,21 @@ def snowflake_supported_join_statement(
 ) -> str:
     LEFT_UUID = format_uuid(left_uuid)
     RIGHT_UUID = format_uuid(right_uuid)
-    left_alias = (
-        "SNOWPARK_LEFT"
-        if use_constant_subquery_alias
-        else random_name_for_temp_object(TempObjectType.TABLE)
-    )
+
+    # If left is a simple SELECT * FROM (\n<join_source>\n) wrapper from a
+    # previous join, flatten into a multi-way join by appending the new right
+    # operand directly to the existing join source. This avoids nested
+    # SELECT * layers that inflate query text without changing semantics.
+    #
+    # Though it is technically less efficient than constructing the join sub-queries
+    # without the SELECT in the first place, the structure of our SQL processing code
+    # top-level projections to be wrapped by a select.
+    unwrapped_left = _unwrap_select_star_from(left)
     right_alias = (
         "SNOWPARK_RIGHT"
-        if use_constant_subquery_alias
+        # Multi-way join: right alias must be unique to avoid collisions
+        # with aliases already present in the flattened join source.
+        if use_constant_subquery_alias and unwrapped_left is None
         else random_name_for_temp_object(TempObjectType.TABLE)
     )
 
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -968,6 +968,24 @@ def lateral_join_statement(
...
+    output of project_statement([], join_source)), return <join_source>.
+    Only unwraps when the inner content starts with '(' which indicates
+    a parenthesized join operand rather than a wrapped SELECT statement."""
+    if sql.startswith(_SELECT_STAR_FROM_PREFIX) and sql.endswith(
+        _SELECT_STAR_FROM_SUFFIX
+    ):
+        inner = sql[len(_SELECT_STAR_FROM_PREFIX) : -len(_SELECT_STAR_FROM_SUFFIX)]
+        if inner.startswith(LEFT_PARENTHESIS):
+            return inner
+    return None
```
sfc-gh-aling, line 988: is it possible that the inner sql is not a join-produced select, and would unwrap cause regression in such case -- can we add such test cases?

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -981,14 +999,21 @@ def snowflake_supported_join_statement(
 ) -> str:
     LEFT_UUID = format_uuid(left_uuid)
     RIGHT_UUID = format_uuid(right_uuid)
-    left_alias = (
-        "SNOWPARK_LEFT"
```
sfc-gh-aling, line 978: can the alias `SNOWPARK_LEFT` be referenced in the parent plan `SNOWPARK_LEFT.<col>` through public apis? or this is completely internal

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -981,14 +999,21 @@ def snowflake_supported_join_statement(
...
+
+    # If left is a simple SELECT * FROM (\n<join_source>\n) wrapper from a
+    # previous join, flatten into a multi-way join by appending the new right
+    # operand directly to the existing join source. This avoids nested
+    # SELECT * layers that inflate query text without changing semantics.
+    #
+    # Though it is technically less efficient than constructing the join sub-queries
+    # without the SELECT in the first place, the structure of our SQL processing code
+    # top-level projections to be wrapped by a select.
+    unwrapped_left = _unwrap_select_star_from(left)
```
sfc-gh-aling: dumb question, why we do not also unwrap right node

sfc-gh-aling (commented):
> The outer-most layers of SELECT * cannot be removed
why we cannot -- where does the nested select come from?
I think the optimal is we just have one `select *` for this case

```sql
SELECT * FROM (
  (
    SELECT "KEY", "VAL_BASE" FROM table_name
  ) AS SNOWPARK_LEFT
  INNER JOIN (
    SELECT "KEY", "VAL_1" FROM table_name
  ) AS SNOWPARK_RIGHT
    USING ("KEY")
  INNER JOIN (
    SELECT "KEY", "VAL_2" FROM table_name
  ) AS SNOWPARK_TEMP_TABLE_Y7YO05WRW5
    USING ("KEY")
)
WHERE "VAL_BASE" > 100
ORDER BY "VAL_BASE" DESC NULLS LAST
LIMIT 5;
```

sfc-gh-joshi:
> > The outer-most layers of SELECT * cannot be removed
> why we cannot -- where does the nested select come from?
> I think the optimal is we just have one select * for this case

Here's the exact code I used for the example query:
```
base_data = [(1, 100), (2, 200), (3, 300), (4, 400), (5, 500), (6, 600), (7, 700), (8, 800)]

base = session.create_dataframe(base_data, schema=["key", "val_base"])

dims = []
for i in range(1, 10):
    dim = session.create_dataframe(base_data, schema=["key", f"val_{i}"])
    dims.append(dim)

df = base
for dim in dims:
    df = df.join(dim, on="key", how="inner")

df = df.filter(col("val_base") > 100)
df = df.order_by(col("val_base").desc())
df = df.limit(5)
```

One layer of SELECT * is from the final `join`, and one layer is added by the filter/order by/limit operations, independent on the depth of the join. I think it would be more complicated to get rid of all of these at once, and this PR focuses on ensuring that repeated joins don't result in repeated SELECT *s.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-joshi: It's completely internal. They're hidden from schema/columns, and I don't believe there's a way for users to access sub-queries from public APIs.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-joshi: A sequence of chained operations (`df1.join(df2).join(df3)...`) would only nest operations on the left side, while nesting on the right side would occur if you did `df1.join(df2.join(df3)...)`. Flattening only one side of the join is significantly simpler; here's Cursor's analysis:

The 2nd concern I raised was: "Merging two join expressions into one FROM clause is significantly more complex — you'd need to handle alias collisions from both sides and potentially break column resolution semantics."

Here's what that means concretely. Consider:

```
left_result = a.join(b, "key")     # produces: A JOIN B
right_result = c.join(d, "key")    # produces: C JOIN D
final = left_result.join(right_result, "key")
```

With left-only unwrapping, we produce:

```
SELECT * FROM (
  (A) AS L1 INNER JOIN (B) AS R1 USING (key)
  INNER JOIN (SELECT * FROM ((C) AS L2 INNER JOIN (D) AS R2 USING (key))) AS R3 USING (key)
)
```

The right operand keeps its SELECT * FROM (...) shell, acting as a self-contained derived table with its own alias namespace.

If we also unwrapped the right side, we'd need to merge both join expressions into a single flat FROM clause:

```
SELECT * FROM (
  (A) AS L1 INNER JOIN (B) AS R1 USING (key)
  INNER JOIN (C) AS L2 INNER JOIN (D) AS R2 USING (key)   -- ← problematic
  USING (key)  -- which USING clause binds to which JOIN?
)
```

This creates two problems:

1. Join precedence ambiguity — SQL evaluates joins left-to-right. Splicing in C JOIN D without a subquery boundary changes how the USING/ON clauses bind. The condition between left_result and right_result needs to apply to the combined result of C JOIN D, but without the subquery wrapper, it would syntactically bind only to D. You'd need explicit parenthesization of the right-side join group, which Snowflake doesn't support as a FROM item (the [TOK_JOIN] error we hit earlier).

2. Alias collisions — Both sides independently generate aliases (SNOWPARK_LEFT, SNOWPARK_RIGHT, or random names). If both join expressions are spliced into the same FROM clause, their aliases coexist in the same namespace. We'd need to rewrite aliases within the right-side join expression to avoid conflicts — which means parsing and modifying already-generated SQL strings, a fragile and error-prone operation.

By keeping the SELECT * FROM (...) wrapper on the right operand, it acts as an opaque derived table: its internal aliases are scoped, its join conditions bind correctly, and it integrates cleanly as a single table reference in the outer FROM clause.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-joshi: Good point, I added a flag in plan generation to ensure we only do this unwrapping if the prior operation was a join, and added a test to make sure it doesn't get tripped for manually-written SQL that matches the pattern.

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -961,6 +961,24 @@ def lateral_join_statement(
...
+def _unwrap_select_star_from(sql: str) -> Optional[str]:
+    """If sql is a join-produced `SELECT * FROM (\n<join_source>\n)` (the
+    output of project_statement([], join_source)), return <join_source>.
+    Only unwraps when the inner content starts with '(' which indicates
+    a parenthesized join operand rather than a wrapped SELECT statement."""
+    if sql.startswith(_SELECT_STAR_FROM_PREFIX) and sql.endswith(
+        _SELECT_STAR_FROM_SUFFIX
+    ):
+        inner = sql[len(_SELECT_STAR_FROM_PREFIX) : -len(_SELECT_STAR_FROM_SUFFIX)]
+        if inner.startswith(LEFT_PARENTHESIS):
```
sfc-gh-yuwang: cursor give me this:
Medium (behavioral/perf): flattening may stop in trace-SQL mode.
In analyzer_utils._unwrap_select_star_from, unwrapping requires the inner text to start with "(". But when UUID trace comments are injected, inner SQL may start with a comment instead, so chained joins might stop flattening after one level. This is likely only when SQL trace-to-dataframe context is enabled.
this sounds like a valid assumption, could this happen in real user workload?

sfc-gh-joshi: Good catch. I've adjusted the code to strip comments before checking for parentheses.

sfc-gh-aling (approved):
LGTM! please ensure the snowaprk connect test pass before merging

sfc-gh-yuwang (approved):

PR #4251, opened 2026-06-03
SNOW-3489633: remove redundant LIMIT 1 on each COUNT slot subquery

sfc-gh-yuwang:

   

   Fixes SNOW-NNNNNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
+++ b/src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
@@ -686,19 +686,20 @@ def aggregate_statement(
     aggregate_exprs: List[str],
     child: str,
     child_uuid: Optional[str] = None,
+    append_global_limit_one: bool = True,
 ) -> str:
     # add limit 1 because aggregate may be on non-aggregate function in a scalar aggregation
     # for example, df.agg(lit(1))
-    return project_statement(aggregate_exprs, child, child_uuid=child_uuid) + (
-        limit_expression(1)
-        if not grouping_exprs
-        else (
-            NEW_LINE
-            + GROUP_BY
-            + NEW_LINE
-            + TAB
-            + (COMMA + NEW_LINE + TAB).join(grouping_exprs)
+    if not grouping_exprs:
+        return project_statement(aggregate_exprs, child, child_uuid=child_uuid) + (
+            limit_expression(1) if append_global_limit_one else EMPTY_STRING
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
```diff
@@ -686,19 +686,20 @@ def aggregate_statement(
     aggregate_exprs: List[str],
     child: str,
     child_uuid: Optional[str] = None,
+    append_global_limit_one: bool = True,
 ) -> str:
     # add limit 1 because aggregate may be on non-aggregate function in a scalar aggregation
     # for example, df.agg(lit(1))
```
sfc-gh-aling, line 692: is this assumption legit in the context of SCOS? what cases will be broken if we always remove `limit_expression(1)` for SCOS?

given the complexity of the implementation (extra tree walk, code maintainability, etc.) and the minimal perf improvement it can help, I'm wondering if it's worth the efforts.

sfc-gh-yuwang:
https://snowpark-python-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/SnowparkConnectRegressionRunnerV2/259/
scos run to verify

src/snowflake/snowpark/_internal/analyzer/analyzer_utils.py
sfc-gh-yuwang: locally verified that the scenario mentioned in the comment would not break SCOS with current change in snowpark

sfc-gh-joshi: Do you have any metrics on the improvements from this change? I agree with Adam that this change seems risky for little benefit.

sfc-gh-aling: I'm curious about why the no-aggregation function in SCOS avoids the issue after we remove limit 1

However, if the underlying logic is overly complex or confusing, let's skip it. Without a clear understanding, we won't have high confidence against regressions.

sfc-gh-yuwang: I did a few re-test, it turns out that this change is also not safe for scos, so we will have to add a lot of logic to make it safe for scos like the first version of this change.
At this point, I think it does not worth it to have this change just to remove limit 1.

PR #4262, opened 2026-06-22
Add Iceberg incremental read via CHANGES AT(VERSION) END(VERSION)

sfc-gh-igarish:
Spark Iceberg incremental reads use start-snapshot-id / end-snapshot-id reader options. Snowflake expresses the same semantics as:

  SELECT * FROM t
    CHANGES (INFORMATION => APPEND_ONLY)
    AT (VERSION => start)
    [ END (VERSION => end) ]

Add IcebergChangesConfig and wire it through DataFrameReader.table(), Session.table(), Table, and SnowflakeTable plan generation. The Spark option aliases (dashed and underscored) are accepted and validated to be mutually exclusive with existing time-travel reader options.

Tests cover SQL clause generation and option extraction.

   Fixes SNOW-3471774


sfc-gh-yuwang:
Is this a public facing change? If so can you help add changelog in corresponding section?

src/snowflake/snowpark/table.py
sfc-gh-yuwang, line 405: looks like time_travel_config is the only thing being forwarded when __copy__() is called, should we also include the added parameter in this PR? Or is this a expected behavior?

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -2148,6 +2148,89 @@ def generate_sql_clause(self) -> str:
...
+    @staticmethod
+    def _coerce_snapshot_id(value: object, option_name: str) -> int:
+        try:
+            snapshot_id = int(value)  # type: ignore[arg-type]
+        except (TypeError, ValueError):
+            raise ValueError(
+                f"'{option_name}' must be a 64-bit integer Iceberg snapshot id, "
+                f"got {value!r}."
+            ) from None
+        if isinstance(snapshot_id, bool):
```
sfc-gh-yuwang: I have a question about this part of code:
It looks like after the first try block, it is impossible for `snapshot_id` to be a bool value here, can you tell me what is the scenario this check trying to guard?

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -248,6 +248,81 @@ def _extract_time_travel_from_options(options: dict) -> dict:
...
+    start = _get_reader_option(options, "start-snapshot-id", "start_snapshot_id")
+    end = _get_reader_option(options, "end-snapshot-id", "end_snapshot_id")
+    if start is None and end is None:
+        return {}
+    if start is None:
+        raise ValueError(
+            "Iceberg incremental read requires 'start-snapshot-id'; "
+            "'end-snapshot-id' cannot be used alone."
+        )
+    try:
```
sfc-gh-yuwang: this part looks to be the same logic as `_coerce_snapshot_id` introduced in utils, is it possible to reuse?

src/snowflake/snowpark/table.py
sfc-gh-igarish: Good catch — that was an oversight, not intentional. `_copy_without_ast()` already forwards `_iceberg_changes_config`; `__copy__()` should mirror that so shallow copies preserve incremental-read state when AST is enabled. I will update `__copy__()` to include the same `start_snapshot_id` / `end_snapshot_id` kwargs block.

sfc-gh-igarish:
Yes — this is a public-facing API change (`Session.table` / `DataFrameReader` incremental-read kwargs and the new `IcebergChangesConfig`). I will add a CHANGELOG entry in the Snowpark Python section describing the Iceberg `start-snapshot-id` / `end-snapshot-id` support and the emitted `CHANGES ... AT(VERSION => ...) END(VERSION => ...)` SQL.

src/snowflake/snowpark/_internal/utils.py
sfc-gh-igarish: This guards the Python `bool`-is-a-subclass-of-`int` footgun: `int(True)` / `int(False)` succeed and become `1` / `0`, which would silently accept boolean reader options as snapshot ids. The extra `isinstance(..., bool)` check rejects that after coercion (same pattern we use for numeric snapshot ids elsewhere).

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-igarish: Agreed — `_extract_iceberg_changes_from_options()` should delegate to `IcebergChangesConfig._coerce_snapshot_id()` instead of duplicating the try/int/validate logic. I will refactor that in the next push so both the reader-option path and `validate_and_normalize_params()` share one helper.

src/snowflake/snowpark/table.py
sfc-gh-igarish: Fixed in 05f8a9148 — `__copy__()` now forwards `_iceberg_changes_config` the same way as `_copy_without_ast()`.

src/snowflake/snowpark/_internal/utils.py
sfc-gh-igarish: Fixed in 05f8a9148 — the bool guard now checks the original value before `int()` coercion (`isinstance(value, bool)`), which is when Python would otherwise accept `True`/`False` as snapshot id `1`/`0`.

src/snowflake/snowpark/dataframe_reader.py
sfc-gh-igarish: Fixed in 05f8a9148 — `_extract_iceberg_changes_from_options()` now delegates to `IcebergChangesConfig._coerce_snapshot_id()`.

sfc-gh-igarish:
Added CHANGELOG entry under 1.53.0 New Features in 05f8a9148.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Snowpark Python API Updates
 
+#### New Features
+
+- Added Iceberg incremental read support via `start_snapshot_id` and `end_snapshot_id` on `Session.table()`, `DataFrameReader.table()`, and `Table`. Spark Iceberg reader options `start-snapshot-id` / `end-snapshot-id` (and underscore variants) are mapped to Snowflake `CHANGES (INFORMATION => APPEND_ONLY) AT (VERSION => ...) [ END (VERSION => ...) ]` SQL.
```
sfc-gh-yuwang: IMO we can remove "Spark Iceberg reader options `start-snapshot-id` / `end-snapshot-id` (and underscore variants) are mapped to Snowflake `CHANGES (INFORMATION => APPEND_ONLY) AT (VERSION => ...) [ END (VERSION => ...) ]` SQL." ?
I am thinking about this because I think we should not mention spark or implementation details in the changelog?

Would need another eye here, @sfc-gh-aling , what is your thought?

sfc-gh-yuwang (approved):
the changelog may need another eye to polish, other than that LGTM

CHANGELOG.md
sfc-gh-igarish: Done — removed the entire 1.53.0 New Features entry in c330956. This incremental-read surface is wired through Snowpark Connect (SAS) only, not as a standalone public Snowpark Python API, so it should not appear in the release notes.

src/snowflake/snowpark/_internal/utils.py
```diff
@@ -2201,6 +2201,89 @@ def generate_sql_clause(self) -> str:
...
+          [ END (VERSION => S2) ]
+
+    When ``end_version`` is omitted, Snowflake uses the current snapshot as
+    the end of the change interval (same semantics as omitting ``END`` on
+    generic ``CHANGES`` queries).
+    """
+
+    start_version: int
+    end_version: Optional[int] = None
+    information: str = "APPEND_ONLY"
```
sfc-gh-aling, line 2238: can the value of this information ever be changed by user? I don't see we extracting this key from user
I see it's always APPEND_ONLY`

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
```diff
@@ -97,10 +97,12 @@ def __init__(
         session: "Session",
         is_temp_table_for_cleanup: bool = False,
         time_travel_config: Optional[TimeTravelConfig] = None,
+        iceberg_changes_config: Optional[IcebergChangesConfig] = None,
     ) -> None:
         super().__init__()
         self.name = name
         self.time_travel_config = time_travel_config
+        self.iceberg_changes_config = iceberg_changes_config
```
sfc-gh-aling, line 105: it feels like the two configs can be consolidated into one

```python
    iceberg_config: Optional[TimeTravelConfig, IcebergChangesConfig] = None,
) -> None:
    ....
    self.iceberg_time_config = iceberg_time_config  # can be some other name
```

and then in the select_statement/snowflake_plan

we can just do
```python
if source_plan.iceberg_time_config:
    table_reference += source_plan.iceberg_time_config.generate_sql_clause()  # no if-elif logic needed
```

sfc-gh-aling: @sfc-gh-mayliu do you have any thoughts here?

src/snowflake/snowpark/_internal/utils.py
sfc-gh-igarish: Good question — today it is always `APPEND_ONLY`. Spark Iceberg incremental read only exposes `start-snapshot-id` / `end-snapshot-id`; there is no reader option (or Snowpark `Session.table` / `DataFrameReader.table` kwarg) that maps to Snowflake `CHANGES (INFORMATION => ...)`, so we never extract it from the user.

The field exists only as an internal default documenting the SQL we emit (Iceberg append-only incremental semantics). Happy to simplify in a follow-up by dropping the `information` parameter from `validate_and_normalize_params()` and hardcoding `APPEND_ONLY` in `generate_sql_clause()` if you prefer less surface area.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan_node.py
sfc-gh-igarish: Makes sense — both configs already share the same `generate_sql_clause()` hook, and the plan builders are only:

```python
if source_plan.iceberg_changes_config:
    table_reference += source_plan.iceberg_changes_config.generate_sql_clause()
elif source_plan.time_travel_config:
    table_reference += source_plan.time_travel_config.generate_sql_clause()
```

I kept them as two optional fields because they are different Snowflake table modifiers (`CHANGES ... AT/END` vs `AT (TIMESTAMP/VERSION/...)`) and we reject mixing them at validation time (`Table` / `DataFrameReader` kwargs). `TimeTravelConfig` is also pre-existing and used broadly outside this PR.

Open to a follow-up refactor (single optional slot + union/protocol) if we want to DRY the `SnowflakeTable` constructor further; I kept this PR scoped to the incremental-read wiring for SAS. Let me know if you would rather I fold that consolidation into this PR.

sfc-gh-aling (approved):

PR #4274, opened 2026-07-01
SNOW-3718333: escape backslashes and single quotes in stage/file path SQL generation

sfc-gh-aling:
### Summary
Stage and file paths passed to `COPY INTO` / `PUT` / `GET` were escaped for single quotes but not backslashes, so a path containing a backslash immediately followed by a single quote produced invalid SQL. `normalize_path` now escapes backslashes before single quotes so the path stays a single string literal.

### Changes
- `_internal/utils.py`: escape `\` before `'` in `normalize_path`.
- Unit tests (`tests/unit/test_internal_utils.py`) and integ tests for `DataFrame.write.csv` and Snowpark-pandas `DataFrame.to_csv`.
- A follow-up commit adjusts two of the new tests for platform / stage-storage behavior (Windows local-path normalization; a literal backslash is not preserved as a stage directory separator), so they assert the escaping guarantee without relying on a backslash round-trip.


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -429,7 +429,17 @@ def normalize_path(path: str, is_local: bool) -> str:
         return path
     if is_local and OPERATING_SYSTEM == "Windows":
         path = path.replace("\\", "/")
-    path = path.strip().replace("'", "\\'")
+    # Escape the backslash before the single quote so the path stays a single
+    # Snowflake string literal; the reverse order would let an escaped quote
+    # close the literal early and produce invalid SQL. Constants keep the
+    # replacements readable (no Python escape double-counting).
+    BACKSLASH = "\\"
```
sfc-gh-aling, line 488: @sfc-gh-yuwang I'm responding your question in this thread

> I have a dumb question here, looking at the old logic, it looks like \' would be written as \\\' because a single quote ' is replaced with \\', which is different from what the comment described?

In Python, `\` is the escape char, so `\\` is used represent a single  `\`.
I have updated the code to make it clear

sfc-gh-yuwang (approved):

sfc-gh-joshi (approved):

sfc-gh-helmeleegy (approved):

PR #4279, opened 2026-07-06
SNOW-3746330 Auto register query listener for Snowflake VS Code Extension

sfc-gh-zyao:
- Auto register a query listener to dump executed query UUIDs into target dir if the env var `SNOWFLAKE_SNOWPARK_VSC_QUERY_HISTORY_DIR` is set, for Snowflake VS Code Extension
  - More details in [this doc](https://docs.google.com/document/d/1BFbVQMfSTtqAkREDMR_XrUjHrS3-wkBfV9xp4N2T-1E/edit?tab=t.0#heading=h.dxyjvk3p71qh)

   Fixes SNOW-3746330

     - This change does not change Snowpark/DataFrame APIs at all

   This change auto registers a query listener to dump executed query UUIDs into target dir if the env var `SNOWFLAKE_SNOWPARK_VSC_QUERY_HISTORY_DIR` is set. [This](https://docs.google.com/document/d/1BFbVQMfSTtqAkREDMR_XrUjHrS3-wkBfV9xp4N2T-1E/edit?tab=t.0#heading=h.dxyjvk3p71qh) is the doc that describes the full flow.


src/snowflake/snowpark/query_history.py
```diff
@@ -141,3 +145,51 @@ def include_failures(self) -> bool:
     @property
     def base64_batches(self) -> List[str]:
         return self._ast_batches
+
+
+class _VscHistoryExporter(QueryListener):
+    """Exports executed query IDs to a directory watched by the Snowflake VS Code extension.
+
+    For each query, an empty file named after the query ID is created in the
+    target directory.
```
sfc-gh-aling, line 154: will it cause file number explosion which in the end hit some file system limitation?

I initially thought it would be a file (name being session id), and we keep appending query ids to the single file.

src/snowflake/snowpark/query_history.py
```diff
@@ -141,3 +145,51 @@ def include_failures(self) -> bool:
...
+        except OSError:
+            # Exporting query history is a best-effort feature for the VS Code
+            # extension; a failure here must not break the user's session.
+            _logger.debug(
+                "Failed to create query history directory %s",
+                query_history_dir,
+                exc_info=True,
+            )
+
+    def _notify(self, query_record: QueryRecord, **kwargs) -> None:
```
sfc-gh-aling, line 221: should we have sort of auto clean up mechanism? or we just let files accumulate in the user env

src/snowflake/snowpark/query_history.py
sfc-gh-zyao: It does not in normal flow. Because the auto cleanup happens in the VS Code Extension side. It will clean up the corresponding file when it finishes reading (and backup cleanup when the extension crashes etc). 

It would be a problem only if someone manually sets the `SNOWFLAKE_SNOWPARK_VSC_QUERY_HISTORY_DIR`. In this case they need to manage/clean up the dir they set. 

I could mitigate this situation by doing both of these
1. add a hard limit of file count in that code, such that if file count >= limit, log debug info and stop emitting
2. add a comment to advise against doing this manually, as it makes no sense to set it manually

src/snowflake/snowpark/query_history.py
sfc-gh-zyao: (also replied [here](https://github.com/snowflakedb/snowpark-python/pull/4279#discussion_r3539692956): we have cleanup mechanism, but not in snowpark-python)

src/snowflake/snowpark/query_history.py
sfc-gh-zyao: Another thing we can do, is that we accumulate query IDs in-memory. And after session closes, dump query IDs to one single file. Would that be preferred? (We didn't go with live append on snowpark side and then live read on the extension side, as the synchronization will be difficult to maintain)

sfc-gh-aling (approved):

src/snowflake/snowpark/query_history.py
sfc-gh-zyao: Discussed over slack, and I added documentation about _VscHistoryExporter's expected behavior and a sanity check in constructor.

sfc-gh-yuwang (approved):

sfc-gh-zyao:
<del>
Regarding the precommit failures,
- `precommit test / Doctest AST (phase0) py-macos-latest-3.10 (pull_request)` fails because in `snowpark.functions.ai_complete`, there is an error `SnowparkSQLException("512513 (P0000): Request failed for external function COMPLETE with remote service error: '400 'The model mistral-large has been deprecated on July 8, 2026, please use other models.`. That is irrelevant to this change.
- `precommit test / Test py-macos-latest-3.12-azure (pull_request)`, fails due to the same reason as above
</del>

PR #4288, opened 2026-07-20
SNOW-3472759: support for to_polars() for snowpark -> polars interoperability

sfc-gh-mayliu:

   

   Fixes SNOW-3472759

This PR adds `DataFrame.to_polars()` to enable direct conversion from a Snowpark DataFrame to a Polars DataFrame/LazyFrame

- Finalized benchmark evaluation: https://docs.google.com/document/d/1jABNlY-zNEMvQZKGg-wKdwu4r7-XpxIGnWE50Vjxiyg/edit?usp=sharing
- Complementary Design doc: https://docs.google.com/document/d/1FKeSB5K9HIHG-aQwcP_2oPRmT7Pd4hcj03l8tictEPM/edit?usp=sharing
- Following stacked PR for AST support: https://github.com/snowflakedb/snowpark-python/pull/4290
- Tests passed on sproc: https://buildkite.com/snowflake/snowflake--snowpark-python-sp-integ-tests/builds/160/tests?jid=019faa00-cf3e-4199-950e-0521dfec5c9b&tab=output


src/snowflake/snowpark/dataframe.py
```diff
@@ -1371,6 +1372,111 @@ def to_arrow_batches(
...
 
+    @publicapi
+    @overload
+    def to_polars(
+        self,
+        *,
+        lazy: bool = False,
+        statement_params: Optional[Dict[str, str]] = None,
+        _emit_ast: bool = True,
+        **kwargs: Dict[str, Any],
```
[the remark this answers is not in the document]
sfc-gh-mayliu: using `Dict[str, Any] `for consistency with how the rest of the functions in dataframe.py define

src/snowflake/snowpark/dataframe.py
```diff
@@ -1371,6 +1372,111 @@ def to_arrow_batches(
...
+            statement_params: Dictionary of statement level parameters to be set while executing this action.
+
+        Note:
+            Requires ``polars>=1.0``.
+        """
+        import polars as pl
+
+        if lazy:
+            schema = pl.from_arrow(
+                self.limit(1).to_arrow(
```
sfc-gh-mayliu: edge case: what if df is empty

src/snowflake/snowpark/_internal/polars_backend.py
```diff
@@ -0,0 +1,199 @@
...
+    import polars as pl
+
+    paths = _copy_df_to_stage(df, "to_polars_parquet_lazy", statement_params)
+    if not paths:
+        return pl.LazyFrame(
+            schema=_empty_frame_from_schema(df, statement_params).schema
+        )
+    # The stream objects are owned by the returned LazyFrame; Polars closes them
+    # when the scan is materialized (or if the LazyFrame is discarded).
+    streams = _open_stage_files_parallel(
```
sfc-gh-yuwang, line 166: i have a question about this ingestion path.
Since this path is ingesting with a stream of parquet file,
if I call collect() two times, will the second collect() still be able to fetch anything since the stream is already consumed at that time?

src/snowflake/snowpark/_internal/polars_backend.py
```diff
@@ -0,0 +1,199 @@
...
+
+    Uses ``DataFrame.write.copy_into_location`` so plan compilation and
+    statement params go through the same path as other write actions. No
+    pre-sort is applied — callers wanting row-group elimination should
+    ``.sort(...)`` before ``to_polars``.
+    """
+    rid = uuid.uuid4().hex[:12]
+    stage = df._session.get_session_stage().rstrip("/")
+    sub = f"{stage}/{sub_prefix}/{rid}/"
+    df.write.copy_into_location(
```
sfc-gh-yuwang, line 86: it appears like the uploaded parquet file is never removed, do we need a garbage cleaning mechanism to make sure the stage is not filled when user invoke this ingestion route multiple times?

src/snowflake/snowpark/_internal/polars_backend.py
```diff
@@ -0,0 +1,199 @@
...
+        sub,
+        file_format_type="parquet",
+        format_type_options={"COMPRESSION": "SNAPPY"},
+        header=True,
+        overwrite=True,
+        statement_params=statement_params,
+    )
+    rows = df._session.sql(f"LIST '{sub}'").collect(statement_params=statement_params)
```
sfc-gh-yuwang, line 94: I remember that List would not return fully qualified stage name, should we make it fully qualified before return?

sfc-gh-mayliu: Good point, LS returns stage-relative paths and we should make it fully qualified. We already have the fully-qualified `sub` prefix at hand, so the fix is to extract just the filename from the LIST result and prepend sub instead. Fix incoming.

src/snowflake/snowpark/_internal/polars_backend.py
sfc-gh-mayliu: `get_session_stage()` returns a temporary stage that Snowflake auto-drops on session close, so no need to recycle here

src/snowflake/snowpark/_internal/polars_backend.py
sfc-gh-mayliu: Both `session.file.get_stream` and `SnowflakeFile.open` eagerly downloads/buffers, so the multiple Polars `.collect()` calls are safe to get the predicate/projection pruned data on CPU without needing to re-download. However, it's a Snowflake infra-limitation that network I/O pruning costs can't be saved due to the lack of HTTP Range GET support in `session.file `and `SnowflakeFile`

Tested with 1M rows: first `.collect()` takes 0.04s, second `.collect()` takes 0.03s — effectively the same cpu-decode cost with no additional network I/O.

sfc-gh-yuwang: this sounds good!

src/snowflake/snowpark/_internal/polars_backend.py
sfc-gh-yuwang: I am thinking about a scenario in which user have a long running pipeline that call to_polars() in this ingestion route, in that scenario, the stage will keep accumulating uploaded parquet files. Do we need to worry about this scenario or is this not a expected usage?

sfc-gh-yuwang (approved):

src/snowflake/snowpark/_internal/polars_backend.py
sfc-gh-mayliu: Good question, discussed offline. TLDR: long-lived sessions exist but are rare, and there's no storage limit on temp stages. sproc also rejects `REMOVE {file}` calls, so a clean up would result in inconsistency between local and server environments. We will rely on temp stages' property to clean up after itself, similar to existing Snowpark patterns

src/snowflake/snowpark/_internal/polars_backend.py
```diff
@@ -0,0 +1,200 @@
...
+            finally:
+                try:
+                    f.close()
+                except Exception as e:
+                    _logger.warning("failed to close stage file %s: %s", p, e)
+
+    if not paths:
+        return []
+    with ThreadPoolExecutor(max_workers=max_workers) as ex:
+        return list(ex.map(_open_read, paths))
```
sfc-gh-aling: this is a styling issue -- I'm seeing repeated code structures here in 2 ways:

- within the function, two similar _open_read, one for sproc, one for non-sproc
- between _open_and_read_parquet_parallel and _open_stage_files_parallel, one is eager one is lazy

is it possible to improve reusability? There is a Python built-in `functools.partial` which might help.

I think the 4 (2x2) can be consolidated into one function without sacrificing too much readability, but would improve maintainability  

def _open_stage_files_parallel(session, path, is_sproc, is_lazy) -> list[pl.Dafarame] | list

src/snowflake/snowpark/dataframe.py
```diff
@@ -1371,6 +1373,163 @@ def to_arrow_batches(
...
+        *,
+        lazy: bool = True,
+        use_parquet: bool = False,
+        max_workers: Optional[int] = None,
+        statement_params: Optional[Dict[str, str]] = None,
+        _emit_ast: bool = False,
+    ) -> "polars.LazyFrame":
+        ...  # pragma: no cover
+
+    @experimental(version="1.54.0")
```
sfc-gh-aling: nit:
```suggestion
    @experimental(version="1.55.0")
```

src/snowflake/snowpark/dataframe.py
```diff
@@ -1371,6 +1373,163 @@ def to_arrow_batches(
...
+    def to_polars(
+        self,
+        *,
+        lazy: bool = True,
+        use_parquet: bool = False,
+        max_workers: Optional[int] = None,
+        statement_params: Optional[Dict[str, str]] = None,
+        _emit_ast: bool = False,
+    ) -> "polars.LazyFrame":
+        ...  # pragma: no cover
```
sfc-gh-aling, line 1401: nit: I remember overload function requires the parameter value to be defined in the following format

```
lazy: Literal[True]
lazy: Literal[False]
```

src/snowflake/snowpark/_internal/polars_backend.py
sfc-gh-mayliu: Sure, consolidated the 2 helpers into 1 for better maintainability

sfc-gh-aling (approved):

PR #4303, opened 2026-08-02
Fix ignored direct time travel parameters in DataFrameReader

fallintoplace:
## Summary

- recognize direct `statement`, `offset`, `timestamp`, and `stream` arguments when choosing the time-travel parameter path
- preserve the documented behavior that direct parameters override reader options
- add public API regression coverage for each affected selector

Previously, `DataFrameReader.table()` only selected the direct-parameter path when `time_travel_mode` or an Iceberg-specific argument was set. A selector such as `offset=-60` was therefore dropped before reaching `Session.table`, silently returning the current table instead of raising the existing missing-mode validation error.

## Testing

- `python -m pytest tests/unit/test_dataframe.py tests/unit/test_utils.py -q`
- `pre-commit run --files src/snowflake/snowpark/dataframe_reader.py tests/unit/test_dataframe.py`

### DIFF
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -833,12 +833,19 @@ class DataFrameReader:
                 start_snapshot_id=start_snapshot_id,
                 end_snapshot_id=end_snapshot_id,
             )
-        elif (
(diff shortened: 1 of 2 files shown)


fallintoplace:

PR #4304, opened 2026-08-04
Snow-3859269 migrate regress test

sfc-gh-jzeng:
---

Fixes SNOW-3859269

This PR only touches `.github/workflows/`, so it adds no tests, logging, telemetry, credentials, or dependencies.

Step one of SNOW-3859269: this only adds the GitHub Actions workflow. The Jenkins job `SnowparkPythonSnowflakePythonClientRegressRunner` is untouched and still running. Retiring it, and removing it from the release-validation fan-out in `jenkins_utils`, is a separate PR once this has run for a few nights.

That job runs the snowpark suite against the connector's unreleased `main` branch, so a connector change that breaks snowpark shows up before the connector ships. On Jenkins it connects to a preprod account whose parameters can differ from prod, so a red build was often an environment difference rather than a real regression, and its definition lived in a different repo from the code it tests.

The line to review closely: `daily_connector_regress.yml` runs `tox -e notdoctest-pandascap-pyarrowcap`, identical to what `scripts/jenkins_regress.sh` runs today. `jenkins_regress.sh` has no test list of its own and delegates everything to `tox.ini`, so that string is the whole guarantee that coverage matches the Jenkins job. Editing it changes coverage silently.

The workflow checks out the connector at `main`, builds a wheel, and sets `snowflake_path` to it, which is all `scripts/tox_install_cmd.sh` needs. Nothing under `src/` or `tests/` changes, and no new secrets are needed: it reuses `decrypt_parameters.sh` and `PARAMETER_PASSWORD`, which is also what lands the run on the same prod account as `daily_precommit.yml`. That account change is the actual fix for the drift.

It runs nightly at 06:00 UTC and on `workflow_dispatch`. `trigger_daily_tests_on_branch.yml` gains a fifth entry so release validation dispatches it alongside the four existing daily workflows.

The three assert steps exist because the bad outcome here is a green run that tested the released connector instead of `main`: the wheel has to carry a cp310 tag, contain the nanoarrow `.so`, and be the one the tests actually imported. Two things deliberately differ from the other daily workflows, both commented in the file: the env resolves to `-n logical` rather than `-n 6`, because a `daily*` env would change the env name and would also drop `--reruns 5` (`precommit.yml` already runs this env on the same runner size), and no junit artifact is uploaded, matching the other workflows here which upload coverage instead.

---

### DIFF
--- a/.github/workflows/daily_connector_regress.yml
+++ b/.github/workflows/daily_connector_regress.yml
@@ -1,11 +1,5 @@
-# Nightly regression of snowpark-python against the connector's UNRELEASED main branch.
-# Replaces the Jenkins job SnowparkPythonSnowflakePythonClientRegressRunner (SNOW-3859269).
-#
-# Two choices below are deliberate -- please do not "fix" them:
-#   * the connector is BUILT FROM MAIN, not installed from PyPI. Testing the released
-#     connector would just duplicate daily_precommit.yml and lose the early warning.
-#   * the tests run against the public/prod account, like every other workflow here,
-#     rather than the preprod account the Jenkins job used. That is the point of SNOW-3859269.
+# Nightly regression of snowpark-python against the connector's unreleased main branch,
+# replacing the Jenkins job SnowparkPythonSnowflakePythonClientRegressRunner (SNOW-3859269).
 
 name: Daily connector-main regress
 on:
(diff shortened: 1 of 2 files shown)


sfc-gh-jzeng:

sfc-gh-yuwang:
I have two questions:
1. is there a way to test the added action before merge?
2. how long does this job usually take?

sfc-gh-yuwang (approved):

sfc-gh-jzeng:
> I have two questions:
> 
> 1. is there a way to test the added action before merge?
> 2. how long does this job usually take?

1. The action/workflow can only be triggered after the definition .yml is merged. However, after merged, we can manually trigger it to see how it goes, and then evaluate how it performs at its expected triggering time.
2. Observed from the Jenkins job history, a typical one would take 1-2 hours. But in some cases, it would take over 10 hours with several false-positive failures. This is also the motivation for migrating from Jenkins to here.

PR #4312, opened 2026-08-11
SNOW-3923354 decouple pandas version from python connector and support pandas 3.x

sfc-gh-jzeng:

   

   Fixes SNOW-3923354

   The connector `[pandas]` extra still pins `pandas<3`. This PR makes Snowpark's `[pandas]` extra declare `pandas<4` and `pyarrow` itself, so a client can install pandas 3. pandas 2 stays supported. A Snowpark bump does not upgrade an existing pandas 2 install. `[modin]` stays on pandas 2 (`pandas<=2.4`).

   Customer BCR on pandas 3: `to_pandas()` / `to_pandas_batches()` use the `str` dtype for text columns (`VARCHAR`, `VARIANT`, `OBJECT`, `ARRAY`, `MAP`, geo). SQL NULL in those columns comes back as `nan`, so `value is None` misses. Use `pandas.isna`. `collect()` is unchanged. `future.infer_string = False` does not turn this off.

   `write_pandas` / `write_arrow` / `create_dataframe` convert duration columns to ns before write. Snowflake stores a unit-less integer; the contract is ns. pandas 3 defaults to `us`.

   Local testing rebuilds columns as Python lists (`dtype=object`) so SQL NULL stays `None` under pandas 3 `str` + Copy-on-Write. Precommit local-testing has two jobs: pandas 2 and pandas 3.


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -1937,6 +1937,12 @@ def get_sorted_key_for_version(version_str):
     )
 
 
+# Use this for pandas 2 vs 3 branches instead of parsing __version__ inline.
+pandas_major_version = (
```
sfc-gh-yuwang: it appears that this is not imported anywhere in this PR, do we really need it?

setup.py
```diff
@@ -43,8 +43,9 @@
     REQUIRED_PYTHON_VERSION = ">=3.10"
 
 PANDAS_REQUIREMENTS = [
-    f"snowflake-connector-python[pandas]{CONNECTOR_DEPENDENCY_VERSION}",
-    "pandas<3.0.0",
+    f"snowflake-connector-python{CONNECTOR_DEPENDENCY_VERSION}",
+    "pandas<4.0.0",
+    "pyarrow",
```
sfc-gh-yuwang: if I understand correctly, we add pyarrow here because we get rid of the [pandas] constraint?
If that is the case, do we need to align the pyarrow limit with connector? Like I remember they have a lower limit for pyarrow> xxx

src/snowflake/snowpark/_internal/utils.py
sfc-gh-jzeng: Thanks for the suggestion! This was imported but eventually fall back to the inline format after consideration. Already removed in the following commits.

setup.py
sfc-gh-jzeng: Yes, we previously relied on [connector side](https://github.com/snowflakedb/snowflake-connector-python/blob/b9a298b3eab417107f46188a2a9ef13195bd4ce9/setup.cfg#L111-L115) which has pyarrow imported in its `[pandas]` extra.

Applied the same constraint mirroring with connector's side now.

PR #4319, opened 2026-08-17
NO-SNOW: cast numeric prompt arg in AI complete test, drop expired date xfails

sfc-gh-bkogan:
## Summary

`test_dataframe_ai_complete_with_named_placeholders` has been failing in every daily precommit job (26/26, all clouds) since the blanket date-based `xfail`s in `src/conftest.py` and `tests/integ/test_dataframe_ai.py` expired on 2026-06-04. Deterministic, not a flake.

**Root cause:** the test passes an integer `rating` column into `PROMPT()`. Official docs say NUMBER is allowed ("any type coercible to a string (for example, VARCHAR, NUMBER, etc.), or FILE"), but the runtime stores the number verbatim and rejects it with `unsupported argument type`. With `return_error_details` off, that surfaces as a `NULL` completion.

Minimal repro:

```sql
SELECT AI_COMPLETE('llama3.1-8b', PROMPT('{0}', 1), {'return_error_details': true});
-- {"value": null, "error": "unsupported argument type"}
```

Sibling `llama3.1-8b` tests pass because they only pass string columns.

## Changes

- Cast `rating` to `StringType` in `input_columns` so this test covers named-placeholder rewriting instead of tripping over the server-side type bug.
- Remove the two expired date-based `xfail`s (doctest hook in `src/conftest.py`, module-level mark in `test_dataframe_ai.py`).

Server-side follow-up for the AI SQL team: either cast string-coercible args in `PROMPT()` as documented, or have the prompt parser stringify scalars. Separately, a type error collapsing to silent `NULL` is a bad failure mode.

## Test plan

- [ ] Daily precommit passes with `test_dataframe_ai_complete_with_named_placeholders` green.
- [ ] Other `dataframe_ai` tests and doctests now run unmasked; confirm no other latent failures.

Verified the casted SQL against a live deployment: all three rows return non-empty completions. Pytest integ run is covered by CI (no local snowpark integ credentials here).

### DIFF
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -2,7 +2,6 @@
 #
 # Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
 #
-import datetime
 import doctest
 import logging
 import os
@@ -101,16 +100,6 @@ def pytest_collection_modifyitems(config, items):
     This function is used to skip some doctests on certain conditions.
     For example, some functions in PrPr might not be available on Azure and GCP.
     """
-    # Mark flaky tests as xfail for a time
-    if datetime.date.today() <= datetime.date(2026, 6, 4):
-        xfail_ai = pytest.mark.xfail(
-            reason="dataframe_ai_functions doctests flaky due to infrastructure issues",
-            strict=False,
-        )
-        for item in items:
-            if isinstance(item, DoctestItem) and "dataframe_ai" in item.name:
-                item.add_marker(xfail_ai)
-
     host = (
         conn_params.get("host") or conn_params.get("HOST") or conn_params.get("account")
     )
--- a/tests/integ/test_dataframe_ai.py
+++ b/tests/integ/test_dataframe_ai.py
@@ -5,6 +5,7 @@
 import json
 import pytest
 from snowflake.snowpark.functions import ai_complete, ai_extract, col, lit, to_file
+from snowflake.snowpark.types import StringType
 from tests.utils import RUNNING_ON_JENKINS, TestFiles, Utils
 from snowflake.snowpark.exceptions import SnowparkSQLException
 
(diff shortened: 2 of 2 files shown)


tests/integ/test_dataframe_ai.py
```diff
@@ -14,14 +13,17 @@
...
 ]
 
 
+@pytest.mark.xfail(
+    reason=(
+        "AI_COMPLETE returns NULL for this named-placeholder prompt in daily CI "
+        "(26/26 jobs on 2026-08-17). Client SQL generation is correct; sibling "
+        "llama3.1-8b complete tests pass. Keep this test running so XPASS is visible."
+    ),
+    strict=False,
```
sfc-gh-bkogan: strict=False here means this test can never fail the suite for ANY reason, not just the documented server-side NULL. If the client SQL generation regresses (the exact path this test is meant to guard) it would be silently absorbed as XFAIL instead of caught. That's the inherent tradeoff of xfail vs. a narrowed assertion, and it's acceptable for a CI unblock, but it does mean the test provides no regression signal until the root cause is fixed and it flips to XPASS. Consider filing the follow-up SNOW ticket referenced in the PR body and putting its id in this reason string so the xfail is traceable and doesn't quietly outlive the underlying fix.

src/conftest.py
```diff
@@ -101,16 +100,6 @@ def pytest_collection_modifyitems(config, items):
     This function is used to skip some doctests on certain conditions.
     For example, some functions in PrPr might not be available on Azure and GCP.
     """
```
sfc-gh-bkogan, line 102: Good cleanup removing the expired date-gated block. Confirmed DoctestItem is still used just below (line 115) and the import is not orphaned, and datetime is no longer referenced anywhere in this file. No issues.

sfc-gh-bkogan (commented):
SHIP WITH NITS

Clean, well-scoped test-only change: expired blanket xfails removed with no dangling references, and the narrow xfail matches the file's existing convention. One tradeoff worth acknowledging inline about strict=False masking non-target failures.

tests/integ/test_dataframe_ai.py
sfc-gh-bkogan: Addressed by dropping the xfail entirely. Latest commit casts `rating` to `StringType` in `input_columns`, so the test runs and asserts again instead of swallowing failures under `strict=False`. The server-side NUMBER-arg bug is being raised separately with the AI SQL team (docs explicitly allow NUMBER).

src/conftest.py
sfc-gh-bkogan: Thanks — still holds on the latest commit; that cleanup is unchanged.

sfc-gh-jzeng (approved):

PR #4324, opened 2026-08-20
NO-SNOW: stop relying on connector threads in pandas CTE test

sfc-gh-bkogan:
## Summary
- Daily Snowpark pandas CI (`test_cte_optimization_for_snowpark_pandas`) has been failing on Linux/macOS since ~Aug 12 because `create_dataframe` no longer leaves a connector background thread around.
- The test only needed another live thread so `warn_session_config_update_in_multithreaded_mode` would fire; it now starts a dummy thread instead of depending on connector internals.
- Windows jobs were already passing because extra threads exist there anyway.

## Test plan
- [x] `pytest tests/integ/modin/test_session_cte_optimization.py::test_cte_optimization_for_snowpark_pandas` locally in the `snowflake` conda env (passed)
- [ ] Daily Snowpark pandas API job should go green for this test on Linux/macOS

### DIFF
--- a/tests/integ/modin/test_session_cte_optimization.py
+++ b/tests/integ/modin/test_session_cte_optimization.py
@@ -4,7 +4,6 @@
 #
 
 import logging
-import pandas as native_pd
 import threading
 from contextlib import contextmanager
 from typing import Any, Generator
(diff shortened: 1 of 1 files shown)


tests/integ/modin/test_session_cte_optimization.py
```diff
@@ -31,29 +30,32 @@ def session_parameter_override(
...
-        assert pd.session.cte_optimization_enabled is True
+            assert pd.session == session
+            assert pd.session.cte_optimization_enabled is True
 
-        # Verify that the warning is filtered out
-        assert (
-            "You might have more than one threads sharing the Session object trying to update"
-            not in caplog.text
-        )
+            assert (
```
sfc-gh-bkogan: This assertion passes both when the warning is suppressed and when it is never emitted at all. That means the test can go green vacuously if the setter ever stops calling warn_session_config_update_in_multithreaded_mode, or if the warning stops being triggered for any other reason. Consider adding a positive control so the test proves suppression rather than mere absence: for example, assert that this same warning DOES appear when the suppression filter is not active (probe thread alive, no _suppress_multithreading_warnings context). Otherwise a real regression in the trigger would silently keep this test passing.

sfc-gh-bkogan (commented):
SHIP WITH NITS

Correct, test-only change that decouples the multithreading-warning trigger from connector internals and is more reliable than the old create_dataframe dependency. The suppression path is still genuinely exercised because the probe thread guarantees active_count() > 1 at the moment cte_optimization_enabled flips to True. A few optional robustness nits are inline.

tests/integ/modin/test_session_cte_optimization.py
```diff
@@ -31,29 +30,32 @@ def session_parameter_override(
...
+                "You might have more than one threads sharing the Session object trying to update"
+                not in caplog.text
+            )
+    finally:
+        stop.set()
+        extra_thread.join()
```
sfc-gh-bkogan: join() has no timeout. It is safe as written since stop.set() runs immediately before it in the same finally and Event.wait() returns at once, but a small timeout (and optionally asserting not extra_thread.is_alive() afterwards) would prevent the whole suite from hanging indefinitely if the probe thread ever failed to observe the stop.

tests/integ/modin/test_session_cte_optimization.py
```diff
@@ -31,29 +30,32 @@ def session_parameter_override(
...
 
     caplog.set_level(logging.WARNING)
 
-    # Creating a dataframe ensures that another thread is running which triggers the warning
-    session.create_dataframe(
-        native_pd.DataFrame([[1, 11], [2, 12], [2, 13]], columns=["A", "B"])
+    # The multithreading warning is emitted only while another thread is alive.
+    stop = threading.Event()
+    extra_thread = threading.Thread(
+        target=stop.wait, name="cte-opt-warning-probe", daemon=True
```
sfc-gh-bkogan: daemon=True together with the explicit join() in the finally is slightly redundant since the join already makes shutdown deterministic. It is harmless as a safety net, just noting the daemon flag is not adding much given the guaranteed join.

tests/integ/modin/test_session_cte_optimization.py
sfc-gh-bkogan: Done. The test now assigns cte_optimization_enabled to itself first (no suppression filter) and asserts the warning is present, then checks it is absent when pd.session enables CTE under the filter.

tests/integ/modin/test_session_cte_optimization.py
sfc-gh-bkogan: Done. join() now uses timeout=5 and the test asserts the probe thread is dead afterward.

tests/integ/modin/test_session_cte_optimization.py
sfc-gh-bkogan: Done. Dropped daemon=True; shutdown is the join in finally.

sfc-gh-mayliu (approved):

PR #4331, opened 2026-09-01
feat(iceberg): emit SPARK_INCREMENTAL_READ for incremental reads

sfc-gh-igarish:
Replace the legacy CHANGES (INFORMATION => APPEND_ONLY) AT/END (VERSION => ...) table suffix with TABLE(SPARK_INCREMENTAL_READ('<table>', start [, end])). Spark-facing options (start-snapshot-id / end-snapshot-id) are unchanged.

   Fixes SNOW-NNNNNNN

### DIFF
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -2326,7 +2326,17 @@ class IcebergChangesConfig(NamedTuple):
         )
 
     def generate_table_reference(self, table_name: str) -> str:
-        """Return a ``TABLE(SPARK_INCREMENTAL_READ(...))`` source for *table_name*."""
+        """Return a ``TABLE(SPARK_INCREMENTAL_READ(...))`` source for *table_name*.
+
+        ``SPARK_INCREMENTAL_READ`` takes the table as its first **VARCHAR**
+        argument (not a SQL identifier), so we quote *table_name* with
+        :func:`~snowflake.snowpark._internal.analyzer.datatype_mapper.str_to_sql`.
+        The string is passed through unchanged from the Spark read path
+        (``session.table(...)`` / ``DataFrameReader.load(...)``); callers that
+        need a catalog-linked or multi-part name should supply the same FQN
+        they would use in a ``CHANGES`` read (e.g. ``db.schema.table``).
+        Snowflake resolves partial names inside the TVF using session context.
+        """
         from snowflake.snowpark._internal.analyzer.datatype_mapper import str_to_sql
 
         args = f"{str_to_sql(table_name)}, {self.start_version}"
--- a/src/snowflake/snowpark/_internal/utils.py
+++ b/src/snowflake/snowpark/_internal/utils.py
@@ -2255,7 +2255,7 @@ class TimeTravelConfig(NamedTuple):
 
 
 class IcebergChangesConfig(NamedTuple):
-    """Configuration for Iceberg incremental reads via Snowflake ``CHANGES``.
+    """Configuration for Iceberg incremental reads via ``SPARK_INCREMENTAL_READ``.
 
     Spark Iceberg exposes incremental reads through::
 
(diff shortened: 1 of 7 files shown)


src/snowflake/snowpark/_internal/utils.py
```diff
@@ -2327,14 +2325,14 @@ def validate_and_normalize_params(
...
-    def generate_sql_clause(self) -> str:
-        clause = (
-            f" CHANGES (INFORMATION => {self.information}) "
-            f"AT (VERSION => {self.start_version})"
-        )
+    def generate_table_reference(self, table_name: str) -> str:
+        """Return a ``TABLE(SPARK_INCREMENTAL_READ(...))`` source for *table_name*."""
+        from snowflake.snowpark._internal.analyzer.datatype_mapper import str_to_sql
+
+        args = f"{str_to_sql(table_name)}, {self.start_version}"
```
sfc-gh-yuwang, line 2342: str_to_sql would add single quotes around the table_name, which makes it a string literal. Do we need to worry about whether the table name is fully qualified?

sfc-gh-igarish: @yuyang — replying to your review on str_to_sql / FQN in generate_table_reference:

Yes — the single-quoted literal is intentional. SPARK_INCREMENTAL_READ takes the table as its first VARCHAR argument (not a SQL identifier), so the emitted shape is:

SELECT * FROM TABLE(SPARK_INCREMENTAL_READ('db.schema.table', <start> [, <end>]))
This matches the manual/integ SQL in test_iceberg_incremental_read_session_table_kwargs.

On FQN: we pass through entity.name unchanged from session.table(...) / DataFrameReader.load(...), same string the caller supplied on the Spark read path. For catalog-linked Iceberg the caller already passes an FQN (e.g. CLDUNITY.scosschema.snapshot_demo); for an unqualified name Snowflake resolves it inside the TVF using session context, analogous to the old ... CHANGES ... AT (VERSION => ...) suffix on a bare table identifier.

Pushed 5ec5fb4cd: expanded the generate_table_reference docstring and added test_iceberg_changes_config_generate_table_reference_unqualified_name.

sfc-gh-yuwang (approved):

sfc-gh-kmehta:
@sfc-gh-igarish heads-up: both merge-gate bugs that were blocking this PR are now fixed on `main` in #4335 (`fb4d77b`), so the workarounds on this branch are redundant and have become a conflict.

## What's on main now

- The integer optional-argument default fix — the same change as your cherry-picked `699c918e`.
- The `test_generator_table_function` `seq2(0)` relaxation — the same intent as your `c1ceb76d`.

## The conflict is small

I test-merged current `main` into `0a0247d` locally. Only one file conflicts:

```
CONFLICT (content): Merge conflict in tests/integ/test_dataframe.py
```

Three hunks, all inside `test_generator_table_function`, because your relaxation and the merged one differ in comment wording. The product files (`datatype_mapper.py`, `udf_utils.py`, `tests/unit/test_datatype_mapper.py`) merge cleanly, since your cherry-pick is byte-identical to what landed.

Suggested resolution: merge `main`, take main's version of that test, and drop `699c918e` and `c1ceb76d`. That should shrink this PR's diff back to just the Iceberg change and clear the `datatype_mapper` / `test_datatype_mapper` files from its file list.

For reference, the merged version relaxes only the `seq2(0)` + `timelimit` assertion to `<=` plus a `0 <= v <= 32767` range check, and keeps the three `seq1(1)` + `rowcount` assertions strictly increasing, since those really are distinct within one 256-value cycle.

## Your remaining red checks

- `Test modin-ubuntu-latest-64-cores-3.12-aws` and `-3.10-azure` fail on `tests/integ/modin/frame/test_apply.py::test_snowpandas_in_apply_negative`. That is unrelated to your change — a pre-existing bug in the Snowpark pandas `apply()` error translation, fixed in #4336. Not required checks, so they should not block you.
- `Test py-macos-latest-3.12-azure` shows as failed but was actually **cancelled** ("higher priority waiting request for `refs/pull/4331/merge`"). It just needs a re-run; there is nothing to fix.

Sorry for the churn — the two fixes deadlocked each other on the required GCP check, which is why they landed together in one PR rather than separately.

PR #4332, opened 2026-09-02
fix: allow duplicate SEQ values in test_generator_table_function

sfc-gh-kmehta:
## Summary
- `test_generator_table_function` failed merge-gate CI with `assert 0 < 0` after #4306 required strictly increasing SEQ values.
- Cause is **seq2 wrap-around** (2-byte width) when `generator(timelimit => 1)` emits enough rows that `ORDER BY seq2(0) LIMIT 3` returns `0, 0, 0`. SEQ docs: unique increasing integers with wrap-around. `seq8` does not wrap in 1s (see `Session.generator` Example 2).
- Fix: accept non-decreasing SEQ after `ORDER BY` (`<=`). Keep 3 rows and `uniform(..., seed=2) == 3`.
- Test-only CHANGELOG note under Other Changes so `check_change_log` can pass (this account cannot apply `NO-CHANGELOG-UPDATES`).

This unblocks same-repo precommit (e.g. #4331 after rebase).

## Test plan
- [x] Live `session.generator` reproduces `seq2` timelimit `0,0,0` (old `<` fails, new `<=` passes)
- [ ] Same-repo precommit: `test_generator_table_function` on `py-ubuntu-latest-64-cores-*-aws`

PR #4334, opened 2026-09-02
fix: allow duplicate SEQ values in test_generator_table_function

sfc-gh-kmehta:
## Summary
- Same-repo continuation of #4332 (fork PR). Merge-gate Test jobs cannot decrypt `parameters.py` on fork PRs.
- `test_generator_table_function` failed merge-gate CI with `assert 0 < 0` after #4306 required strictly increasing SEQ values.
- Cause is **seq2 wrap-around** (2-byte width) when `generator(timelimit => 1)` emits enough rows that `ORDER BY seq2(0) LIMIT 3` returns `0, 0, 0`. SEQ docs: unique increasing integers with wrap-around. `seq8` does not wrap in 1s (see `Session.generator` Example 2).
- Fix: accept non-decreasing SEQ after `ORDER BY` (`<=`). Keep 3 rows and `uniform(..., seed=2) == 3`.
- Test-only CHANGELOG note under Other Changes.

Closes #4332.

## Test plan
- [x] Live `session.generator` reproduces `seq2` timelimit `0,0,0` (old `<` fails, new `<=` passes)
- [ ] Same-repo precommit: `test_generator_table_function` on `py-ubuntu-latest-64-cores-*-aws`


sfc-gh-yuwang (approved):

sfc-gh-kmehta:
Closing in favor of #4335, which now carries this exact fix unchanged.

**Why:** this PR and #4335 deadlocked each other. `py-windows-latest-64-cores-3.14-gcp` is a required check, and on `main` it currently fails on two unrelated bugs:

| PR | fixes | was blocked by |
|---|---|---|
| this one | `test_generator_table_function` | the six `*_with_optional_args` tests |
| #4335 | the six `*_with_optional_args` tests | `test_generator_table_function` |

Neither branch could turn that check green on its own, so combining them was the only way to land either one without an admin override.

Two things worth recording:

1. #4335's run independently reproduced this failure on plain `main` with no involvement from #4331, confirming the diagnosis here — `assert 0 < 0` at `test_dataframe.py:576`, on the `seq2(0)` + `generator(timelimit => 1)` case.
2. It is no longer intermittent. On that run it failed on **six** jobs, including four AWS/Azure 64-core ones, not just GCP. The warehouse now comfortably exceeds one 32768-value `seq2` cycle in a second, so `ORDER BY seq2(0) ... LIMIT 3` returns `0, 0, 0` fairly reliably.

The approving review here (thanks @sfc-gh-yuwang) applies to the test change as ported; it is byte-identical, and the branch `kmehta-fix-generator-table-function-seq-duplicates` is retained if this needs to be reopened.

PR #4335, opened 2026-09-02
fix: integer optional-argument defaults + generator SEQ test, to clear merge gates

sfc-gh-kmehta:

   Related to SNOW-4013704 (server-side regression that surfaced this).

2. Pre-review checklist:

   - [ ] If adding any arguments to public Snowpark APIs or creating new public Snowpark APIs, I acknowledge that I have ensured my changes include AST support.

This PR carries two independent fixes, both of which are currently failing merge gates on every PR in the repo. They are together because each one fixes the check that blocks the other — see [Why both fixes are in one PR](#why-both-fixes-are-in-one-pr) at the bottom.

### Problem

Registering a UDF, UDTF, UDAF, or stored procedure with an **integer** optional argument generates:

```sql
CREATE FUNCTION ... (x BIGINT DEFAULT 0 :: INT, ...)
```

A parameter default only accepts a constant expression. Snowflake rejects the cast on some server versions:

```
000947 (42601): SQL compilation error:
invalid default argument expression CAST(0 AS NUMBER(38,0))
```

`None` defaults hit the same validation through `NULL :: INT`, reported as `invalid default argument expression SYSTEM$NULL_TO_FIXED(null)`.

This fails the `py-windows-latest-64-cores-3.14-gcp` merge gate (a required check) via these six tests:

- `tests/integ/test_udf.py::test_register_udf_with_optional_args[True|False]`
- `tests/integ/test_udtf.py::test_udtf_register_with_optional_args[True|False]`
- `tests/integ/test_stored_procedure.py::test_register_sp_with_optional_args[True|False]`

### Fix

The cast is redundant — the parameter already declares its own type — so integral defaults are now rendered as bare literals (`DEFAULT 0`, `DEFAULT NULL`) via a new `to_sql_for_default_arg`. It is scoped to the default-argument path only; `to_sql` is unchanged, so `VALUES` and expression rendering keep their casts.

I checked every default form Snowpark generates against an affected deployment. Only the integral one is rejected, so everything else keeps its existing rendering:

| default type | generated SQL | affected server |
|---|---|---|
| **int** | `0 :: INT` | **rejected** |
| float | `'1.0' :: FLOAT` | ok |
| str | `'one'` | ok |
| bool | `True :: BOOLEAN` | ok |
| decimal | `1 :: NUMBER (38, 0)` | ok |
| array / map | `PARSE_JSON(...) :: ARRAY` / `:: OBJECT` | ok |
| variant | `PARSE_JSON(...)` | ok |
| geometry / geography | `TO_GEOMETRY(...)` / `TO_GEOGRAPHY(...)` | ok |
| timestamp / date / time | `TIMESTAMP '...'` / `DATE '...'` / `TIME('...')` | ok |
| binary | `'313233' :: BINARY` | ok |

### Test plan

- New unit test `test_to_sql_for_default_arg` covering all integral types, negatives, `None`, `bool` normalization, and that all other types are unchanged relative to `to_sql`.
- Verified end-to-end against a live account on an affected version (10.30.50), which reproduces the failure before the change:

| case | before | after |
|---|---|---|
| `udf.register` (int defaults) | rejected | `add() = 0`, `add(3) = 3` |
| `udf.register_from_file` (int defaults) | rejected | `add2() = 12` |
| `udtf.register` (int default) | rejected | 2 rows |
| `sproc.register` (int default) | rejected | `bump() = 12` |
| `udf.register` (`None` default) | rejected | `maybe() = -1` |

- Regression check: re-registered the 14-type `return_all_datatypes` UDF from `test_register_udf_with_optional_args`, plus the mixed positional + array-default `return_arr`. Both register and evaluate their defaults correctly.
- Confirmed on CI: on the previous run of this branch, all six `optional_args` tests passed on `py-windows-latest-64-cores-3.14-gcp` with zero `invalid default argument` errors, and the job's pass count rose by exactly six.

## Fix 2 — `test_generator_table_function` (test-only)

Ported unchanged from #4334, which was approved by @sfc-gh-yuwang.

`test_generator_table_function` fails as `assert 0 < 0` on the `seq2(0)` + `generator(timelimit => 1)` case:

```
tests/integ/test_dataframe.py:576: in test_generator_table_function
    assert result[0][0] < result[1][0] < result[2][0]
E   assert 0 < 0
```

`seq2(0)` passes sign=0, so [per the SEQ docs](https://docs.snowflake.com/en/sql-reference/functions/seq1) it continues at 0 rather than going negative after 32767. A one-second `generator` run usually emits far more than one 32768-value cycle, so 0 repeats and `ORDER BY seq2(0) ... LIMIT 3` returns `0, 0, 0`. The strict `<` introduced in #4306 cannot hold in that case.

That one assertion now allows equal values, plus a `0 <= v <= 32767` range check to keep the sign=0 contract covered. The three `seq1(1)` + `rowcount` assertions stay strictly increasing, since 150 rows stay inside one 256-value cycle and really are distinct.

This is load-dependent, which is why it was intermittent at first. It is no longer intermittent: on the previous run of this branch it failed on **six** jobs, including four AWS/Azure 64-core ones, not just GCP.

## Why both fixes are in one PR

The two fixes deadlocked each other:

| PR | fixes | was blocked by |
|---|---|---|
| #4334 | the generator test | the six `optional_args` tests |
| #4335 (this) | the six `optional_args` tests | the generator test |

`py-windows-latest-64-cores-3.14-gcp` is a required check and it fails on both bugs, so neither branch could go green on its own. Combining them is the only way to land either without an admin override. #4334 is redundant once this merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 - Added interval type support for Python UDFs and stored procedures. Use `datetime.timedelta` as the type annotation for day-time interval (`DayTimeIntervalType`) parameters and return values, and `YearMonthInterval` (a type annotation sentinel from `snowflake.snowpark.types`) for year-month interval (`YearMonthIntervalType`) parameters and return values.
 
+#### Bug Fixes
+
+- Fixed registering a UDF, UDTF, or stored procedure with an integer optional argument failing with `SQL compilation error: invalid default argument expression`. Integer defaults were emitted as `DEFAULT <value> :: INT`, and a parameter default only accepts a constant expression. The redundant cast is no longer generated.
+
 ## 1.54.0 (2026-07-29)
 
 ### Snowpark Python API updates
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 - Fixed registering a UDF, UDTF, or stored procedure with an integer optional argument failing with `SQL compilation error: invalid default argument expression`. Integer defaults were emitted as `DEFAULT <value> :: INT`, and a parameter default only accepts a constant expression. The redundant cast is no longer generated.
 
+#### Other Changes
+
+- Test-only: fixed `test_generator_table_function` failing as `assert 0 < 0` on the `seq2(0)` + `generator(timelimit => 1)` case. `seq2(0)` passes sign=0, so it continues at 0 after 32767, and a one-second generator run usually emits far more than one 32768-value cycle, making `ORDER BY seq2(0) LIMIT 3` return `0, 0, 0`. That assertion now allows equal values; the `seq1(1)` + `rowcount` assertions stay strictly increasing.
+
 ## 1.54.0 (2026-07-29)
 
 ### Snowpark Python API updates
--- a/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
+++ b/src/snowflake/snowpark/_internal/analyzer/datatype_mapper.py
@@ -331,6 +331,25 @@ def to_sql_no_cast(
     return f"{value}"
 
 
+def to_sql_for_default_arg(value: Any, datatype: DataType) -> str:
+    """Convert a value with DataType to SQL usable as a parameter's DEFAULT.
+
+    Identical to :func:`to_sql` except that integral values are emitted as bare
+    literals rather than ``<value> :: INT``. A DEFAULT only accepts a constant
+    expression, and the parameter already declares its own type, so the cast is
+    redundant. Snowflake rejects it outright on some server versions with
+    ``invalid default argument expression`` (SNOW-4013704).
+    """
+    if isinstance(datatype, _IntegralType):
+        if value is None:
+            return "NULL"
+        if isinstance(value, int):
+            # bool is a subclass of int, so normalize it to 0/1 for an
+            # integral parameter rather than emitting True/False.
+            return str(int(value))
+    return to_sql(value, datatype)
+
+
 def to_sql(
     value: Any,
     datatype: DataType,
--- a/src/snowflake/snowpark/_internal/udf_utils.py
+++ b/src/snowflake/snowpark/_internal/udf_utils.py
@@ -32,7 +32,11 @@ from packaging.requirements import Requirement
 import snowflake.snowpark
 from snowflake.connector.options import installed_pandas, pandas
 from snowflake.snowpark._internal import code_generation, type_utils
-from snowflake.snowpark._internal.analyzer.datatype_mapper import to_sql, to_sql_no_cast
+from snowflake.snowpark._internal.analyzer.datatype_mapper import (
+    to_sql,
+    to_sql_for_default_arg,
+    to_sql_no_cast,
+)
 from snowflake.snowpark._internal.telemetry import TelemetryField
 from snowflake.snowpark._internal.type_utils import (
     NoneType,
@@ -448,7 +452,7 @@ def get_opt_arg_defaults(
 
         if num_optional_args != 0:
             default_values_to_sql_str = [
-                to_sql(value, datatype)
+                to_sql_for_default_arg(value, datatype)
                 for value, datatype in zip(default_values, input_types_for_default_args)
             ]
         else:
--- a/tests/integ/test_dataframe.py
+++ b/tests/integ/test_dataframe.py
@@ -551,8 +551,9 @@ def test_select_table_function(session):
 )
 def test_generator_table_function(session):
     # works with rowcount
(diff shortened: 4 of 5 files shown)


sfc-gh-kmehta:
## CI status: all 8 required checks green

Notably `Test py-windows-latest-64-cores-3.14-gcp`, the gate that both bugs were failing, is now clean:

```
= 6283 passed, 301 skipped, 25 xfailed, 34 xpassed, 1227 warnings in 1246.84s =
```

Zero occurrences of `invalid default argument` and no `assert 0 < 0`. The pass count moved 6276 → 6282 → 6283 across the three runs, which is exactly the six `optional_args` tests plus the one generator test, so both fixes are individually accounted for.

`py-windows-latest-64-cores-3.13-gcp` and the previously-red `py-ubuntu-64-cores-3.10-aws` / `3.14-aws` also went green, confirming the generator fix wasn't just papering over the GCP symptom.

## The two red jobs are pre-existing and unrelated

`modin-ubuntu-latest-64-cores-3.12-aws` and `modin-ubuntu-latest-64-cores-3.10-azure` both fail on one test only, `tests/integ/modin/frame/test_apply.py::test_snowpandas_in_apply_negative`, with 39208 and 39669 other tests passing.

Cause is the UDF sandbox failing to import `snowflake.snowpark`, so a `ModuleNotFoundError` is raised where the test expects the friendly "don't reference `modin.pandas` inside `apply()`" message. It's an environment/packaging problem, it predates this branch, and it's spreading across accounts on its own schedule (3.10-azure was green on the previous run of this same branch).

I specifically checked whether this change could be responsible, since modin's `apply()` does go through `udf_utils.py`. It isn't: neither log contains any reference to the default-argument path. Neither job is a required check.

## What's left

Just review. Fix 2 (the `test_generator_table_function` change) is byte-identical to the version @sfc-gh-yuwang already approved on #4334 — I verified with `git diff` between the branches — so the substantive review surface here is Fix 1, which is `to_sql_for_default_arg` in `datatype_mapper.py` plus its two call sites in `udf_utils.py`.

Worth flagging that the generator bug is currently failing merge gates on every PR in the repo and is no longer intermittent, so this is somewhat time-sensitive.

sfc-gh-yuwang (approved):
LGTM,
there is one pandas failure in the merge gate, can you verify if it is related?

sfc-gh-kmehta:
@sfc-gh-yuwang answering your question, and correcting myself: **it is not related to this PR, but it is a real bug in our code, not the environment issue I claimed above.**

Sorry for merging before replying — that was my miss.

## Not related to this PR

The failing test is `tests/integ/modin/frame/test_apply.py::test_snowpandas_in_apply_negative` on `modin-ubuntu-latest-64-cores-3.12-aws` and `-3.10-azure`.

- The failure is in `_apply_with_udtf_and_dynamic_pivot_along_axis_1` in `snowflake_query_compiler.py`, which this PR does not touch.
- The responsible code has been in place since dc4deee (2026-06-03), three months before this PR.
- Neither job's log references the default-argument path this PR changes. I checked specifically because modin's `apply()` does route through `udf_utils.py`.
- It was already failing on `3.12-aws` before this branch existed.

## What it actually is

My earlier comment on this PR called it "an environment/packaging problem... spreading across accounts." That was wrong, and I should not have written it off.

The test deliberately references `modin.pandas` inside `apply()`, so the generated UDTF handler imports a package the sandbox does not have. That server error is expected — it is the [documented limitation](https://docs.snowflake.com/developer-guide/snowpark/python/pandas-on-snowflake#limitations) being exercised. We catch it and raise a friendly message. The detection was:

```python
if "No module named 'snowflake'" in str(e) or "Modin is not installed" in str(e):
```

But the server said:

```
ModuleNotFoundError: No module named 'snowflake.snowpark' in function SNOWPARK_TEMP_TABLE_FUNCTION_GK4GB6JDMA with handler udf_py_480608373.compute
```

The character after `snowflake` is `.`, not `'`, so the guard misses, control falls to the `else` branch, `cache_result` is retried, fails identically, and the raw `ModuleNotFoundError` propagates. I confirmed against the verbatim message from the failing job: old guard `False`, new guard `True`.

Which name the sandbox reports depends on whether the image happens to carry some other `snowflake.*` distribution. That is what made it look account-dependent and flaky — it is neither, and the test was correctly reporting a real defect.

The user impact is that on those accounts, anyone hitting this limitation gets a `ModuleNotFoundError` naming an internal temp function instead of the guidance we wrote.

## Fix

#4336, with a unit test covering both message forms so the coverage no longer depends on which sandbox image an account has.

I also flagged there that the `else` branch blindly re-runs `cache_result(udtf_dataframe)` on a guard miss, so the whole UDTF-plus-cache path executes twice before failing. Left alone for a separate change.

PR #4336, opened 2026-09-03
fix: recognize missing snowpark submodule in Snowpark pandas apply() error

sfc-gh-kmehta:

   SNOW-1345395 area (Snowpark pandas `apply()`); no dedicated ticket yet.

2. Pre-review checklist:

   - [ ] If adding any arguments to public Snowpark APIs or creating new public Snowpark APIs, I acknowledge that I have ensured my changes include AST support.

## Problem

Referencing `modin.pandas` inside a Snowpark pandas `apply()` is a [documented limitation](https://docs.snowflake.com/developer-guide/snowpark/python/pandas-on-snowflake#limitations). We detect it and raise a message pointing the user at native pandas. That detection was a substring match on the server error:

```python
if "No module named 'snowflake'" in str(e) or "Modin is not installed" in str(e):
```

The generated UDTF handler imports a package the sandbox does not have, and **which name the sandbox reports depends on the image**. A sandbox carrying no `snowflake` package reports `snowflake`, matching the guard. One carrying some other `snowflake.*` distribution instead reports the missing submodule:

```
100357 (P0000): Python Interpreter Error:
ModuleNotFoundError: No module named 'snowflake.snowpark' in function SNOWPARK_TEMP_TABLE_FUNCTION_GK4GB6JDMA with handler udf_py_480608373.compute
```

The character after `snowflake` is `.`, not `'`, so the guard misses. Control falls to the `else` branch, which re-runs `cache_result`, fails identically, and propagates the raw `ModuleNotFoundError`.

So on those accounts a user hitting this limitation gets a `ModuleNotFoundError` naming an internal temp function, instead of the guidance we wrote for them.

This is also why `tests/integ/modin/frame/test_apply.py::test_snowpandas_in_apply_negative` fails on some accounts and not others — currently `modin-ubuntu-latest-64-cores-3.12-aws` and `modin-ubuntu-latest-64-cores-3.10-azure`. It looked like flakiness or a sandbox packaging problem; it is neither. The sandbox behavior is expected, and the test is correctly reporting that our error translation misses.

## Fix

Match on the prefix so both forms are recognized, extracted into `_is_missing_snowpark_or_modin_error` so it can be tested directly:

```python
return (
    "No module named 'snowflake" in message
    or "No module named 'modin" in message
    or "Modin is not installed" in message
)
```

`modin` is included for the same reason — the handler can fail on either import depending on the image.

Behavior is otherwise unchanged. Errors that are genuinely something else still take the existing path.

## Test plan

- New `test_is_missing_snowpark_or_modin_error` / `_negative` in `tests/unit/modin/test_snowflake_query_compiler.py`, covering the bare `snowflake` form, the `snowflake.snowpark` and `snowflake.snowpark.modin` submodule forms, `modin` / `modin.pandas`, `Modin is not installed`, and negatives including a different missing package (`scipy`) so the prefix match can't over-trigger.
- Verified the predicate against the **verbatim** message from the failing CI job: old guard `False`, new guard `True`.
- `tests/unit/modin` passes locally: 801 passed, 17 skipped, 4 xfailed. (`test_internal_frame.py` fails to collect in my local env both with and without this change — a pre-existing local artifact.)
- `test_snowpandas_in_apply_negative` should go green on the two modin jobs above, which is the real end-to-end confirmation.

## Note on scope

I deliberately left the `else` branch alone, but flagging it for a follow-up: on a guard miss it blindly re-runs `cache_result(udtf_dataframe)`, so the whole UDTF-plus-cache path executes twice before failing. That doubles the cost of any unrelated failure in this code path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@
 
 - Test-only: fixed `test_generator_table_function` failing as `assert 0 < 0` on the `seq2(0)` + `generator(timelimit => 1)` case. `seq2(0)` passes sign=0, so it continues at 0 after 32767, and a one-second generator run usually emits far more than one 32768-value cycle, making `ORDER BY seq2(0) LIMIT 3` return `0, 0, 0`. That assertion now allows equal values; the `seq1(1)` + `rowcount` assertions stay strictly increasing.
 
+### Snowpark pandas API Updates
+
+#### Bug Fixes
+
+- Fixed referencing `modin.pandas` inside a Snowpark pandas `apply()` raising a raw `ModuleNotFoundError` about an internal temporary function instead of the message explaining that only native pandas may be used inside `apply()`. The error was only recognized when the sandbox reported the missing package as `snowflake`, not when it reported a submodule such as `snowflake.snowpark`.
+
 ## 1.54.0 (2026-07-29)
 
 ### Snowpark Python API updates
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 
 #### Bug Fixes
 
-- Fixed referencing `modin.pandas` inside a Snowpark pandas `apply()` raising a raw `ModuleNotFoundError` about an internal temporary function instead of the message explaining that only native pandas may be used inside `apply()`. The error was only recognized when the sandbox reported the missing package as `snowflake`, not when it reported a submodule such as `snowflake.snowpark`.
+- Fixed a bug where referencing `modin.pandas` inside a Snowpark pandas `apply()` function raised a `ModuleNotFoundError`.
 
 ## 1.54.0 (2026-07-29)
 
--- a/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
+++ b/src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
@@ -787,30 +787,6 @@ def _apply_func_has_snowpark_function(func: Any) -> bool:
     )
 
 
-def _is_missing_snowpark_or_modin_error(message: str) -> bool:
-    """
-    Check whether a server error means the UDF sandbox could not import Snowpark or Modin.
-
-    Referencing ``modin.pandas`` inside ``apply()`` makes the generated handler
-    import a package the sandbox does not have. Which name appears in the error
-    depends on the image: sandboxes carrying no ``snowflake`` package at all
-    report ``snowflake``, while those carrying some other ``snowflake.*``
-    distribution report the missing submodule, e.g. ``snowflake.snowpark``.
-    Match on the prefix so both forms are recognized.
-
-    Args:
-        message: The server error message.
-
-    Returns:
-        True if the error is a missing Snowpark/Modin import, False otherwise.
-    """
-    return (
-        "No module named 'snowflake" in message
-        or "No module named 'modin" in message
-        or "Modin is not installed" in message
-    )
-
-
 @_propagate_attrs_on_methods
 class SnowflakeQueryCompiler(BaseQueryCompiler):
     """based on: https://modin.readthedocs.io/en/0.11.0/flow/modin/backends/base/query_compiler.html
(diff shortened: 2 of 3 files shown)


sfc-gh-kmehta:
Rebased onto `fb4d77b` (post-#4335 `main`) so the required `py-windows-latest-64-cores-3.14-gcp` check can pass — before the rebase this branch predated both fixes in #4335 and would have failed that gate on bugs unrelated to this change. Only the `CHANGELOG.md` entry needed resolving; the code change is unchanged.

Context: this answers @sfc-gh-yuwang's question on #4335 about the one pandas failure in the merge gate. It is not related to #4335 — the responsible code dates to dc4deee (2026-06-03) — but it is a real bug rather than the environment issue I first assumed. Full reasoning in [this comment](https://github.com/snowflakedb/snowpark-python/pull/4335#issuecomment-5518348277).

sfc-gh-kmehta:
## Verified end-to-end: the fix works

Both jobs that were failing this test now pass, with zero failures in either:

| job | `test_snowpandas_in_apply_negative` |
|---|---|
| `Test modin-ubuntu-latest-64-cores-3.12-aws` | pass |
| `Test modin-ubuntu-latest-64-cores-3.10-azure` | pass |

These are the two accounts whose sandbox reports the missing package as `snowflake.snowpark` rather than `snowflake`, so they are exactly the case the old guard missed. The friendly "only native pandas inside `apply()`" message is now raised as intended.

`Test py-windows-latest-64-cores-3.14-gcp` is also green after the rebase onto `fb4d77b`.

Required checks: 6 of 8 passing, the other 2 still running, nothing failing. Ready for review — @sfc-gh-yuwang, this closes out your question on #4335.

sfc-gh-jzeng (approved):

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
```diff
@@ -787,6 +787,30 @@ def _apply_func_has_snowpark_function(func: Any) -> bool:
     )
 
 
+def _is_missing_snowpark_or_modin_error(message: str) -> bool:
```
sfc-gh-yuwang: If the function in production code is only referred once, I think in this scenario we can put the if clause in place and not encapsulate it.
The test also need to be changed accordingly as well if we going this route

CHANGELOG.md
```diff
@@ -16,6 +16,12 @@
 
 - Test-only: fixed `test_generator_table_function` failing as `assert 0 < 0` on the `seq2(0)` + `generator(timelimit => 1)` case. `seq2(0)` passes sign=0, so it continues at 0 after 32767, and a one-second generator run usually emits far more than one 32768-value cycle, making `ORDER BY seq2(0) LIMIT 3` return `0, 0, 0`. That assertion now allows equal values; the `seq1(1)` + `rowcount` assertions stay strictly increasing.
 
+### Snowpark pandas API Updates
+
+#### Bug Fixes
+
+- Fixed referencing `modin.pandas` inside a Snowpark pandas `apply()` raising a raw `ModuleNotFoundError` about an internal temporary function instead of the message explaining that only native pandas may be used inside `apply()`. The error was only recognized when the sandbox reported the missing package as `snowflake`, not when it reported a submodule such as `snowflake.snowpark`.
```
sfc-gh-yuwang:
```suggestion
- Fixed a bug where referencing `modin.pandas` inside a Snowpark pandas `apply()` function raised a `ModuleNotFoundError`.
```

sfc-gh-yuwang (changes requested):
Hi Kunal, can you take a look at the comments before merging? Thank you!

src/snowflake/snowpark/modin/plugin/compiler/snowflake_query_compiler.py
sfc-gh-kmehta: Done in 7b6632a — condition is back inline in the `if`, helper removed, and the two unit tests that existed only to reach it are gone (that file is now identical to `main`).

One trade-off worth naming: the integration test `test_snowpandas_in_apply_negative` is now the only coverage, and it exercises the `snowflake.snowpark` form only on accounts whose sandbox happens to report that name — which is what made this bug invisible for so long. It does currently fail without the fix and pass with it on `modin-ubuntu-64-cores-3.12-aws` and `-3.10-azure`, so we do have real coverage today; it's just account-dependent rather than deterministic. Happy to leave it here.

CHANGELOG.md
sfc-gh-kmehta: Applied your wording in 7b6632a.

sfc-gh-kmehta:
@sfc-gh-yuwang both comments addressed in 7b6632a:

- **Inlined the guard.** The predicate had a single caller, so the condition is back in the `if` clause with a short comment on why it is a prefix match. The helper and the two unit tests that existed only to reach it are removed — `tests/unit/modin/test_snowflake_query_compiler.py` is now byte-identical to `main`.
- **Changelog** shortened to your suggested wording.

The behavior is unchanged. I checked that by parsing the shipped source and evaluating the actual condition from the method rather than a copy of it:

| message | matches |
|---|---|
| the verbatim CI failure (`No module named 'snowflake.snowpark'`) | yes |
| bare `No module named 'snowflake'` | yes |
| `No module named 'modin'` | yes |
| `Modin is not installed` | yes |
| unrelated `No module named 'scipy'` | no |
| unrelated SQL compilation error | no |

Diff is now 11 insertions and 64 deletions across two files. Lint and mypy clean; `tests/unit/modin` passes at 791 passed, 17 skipped, 4 xfailed.

Re-requesting review. Thanks for catching the over-encapsulation.

sfc-gh-yuwang (approved):

PR #4340, opened 2026-09-03
SNOW-3898656: XML perf optimization variant projection (Stacked 1)

sfc-gh-mayliu:

   

   Fixes SNOW-3898656

Adds `useVariantProjection` option for `DataFrameReader.xml()` to enable direct VARIANT-key projection, bypassing the expensive flatten and pivot operations.

### Technique

* **Current behavior:** Reading XML with `rowTag` set produces one VARIANT column (`ROW_DATA`) per record. Turning that into a table uses `flatten()` + a dynamic `pivot()`: flatten expands every record into one row per key, carrying the full data value each time (m×n rows of `{key, value}`), then pivot re-aggregates those m×n rows back into columns. This requires two passes over m×n rows of actual data due to the nature of dynamic pivot.
* **Optimized behavior (`useVariantProjection=True`):** Skips flatten+pivot entirely. It discovers the distinct key names with a single pass over `OBJECT_KEYS(ROW_DATA)` (m×n rows of key-name strings only, no data values), then projects each discovered key directly out of `ROW_DATA` via bracket notation (`ROW_DATA['key']`). This requires only one pass over key names instead of two passes over full values.

### Limitations & Defaults

* **Limitation:** Requires `cacheResult=True` (the default) or a known schema, since key discovery needs a materialized result to scan. With `cacheResult=False` and no schema, it falls back to the existing flatten+pivot behavior.
* **Default:** Opt-in (`False` by default). A plain `.xml()` call is byte-for-byte unchanged.

### Real-Workload Impact & Performance

Measured on the same warehouse:

| XML File | Before | After |
| :--- | :--- | :--- |
| **4GB / 805 cols** | 172.17s | 74.16s (−56.9%) |
| **50GB / 805 cols** | 1648.48s | 643.13s (−61.0%) |

* Row/column output verified identical before and after on both files.
* **SAS impact:** No regression. Verified against `snowpark-connect`'s XML expectation test suite: 12/12 pass, identical to main on every test. Also verified that SCOS XML expectation tests produced exact same DFs if this option is turned on. SCOS can consider opting in to optimize performance, but this is out of scope for this PR.

### DIFF
--- a/src/snowflake/snowpark/dataframe_reader.py
+++ b/src/snowflake/snowpark/dataframe_reader.py
@@ -1324,13 +1324,9 @@ class DataFrameReader:
                 This means the actual computation occurs before :meth:`DataFrame.collect` is called.
                 When set to ``False``, the DataFrame is computed lazily and the actual computation occurs when :meth:`DataFrame.collect` is called.
 
-              + ``useVariantProjection``: Whether to build the output by projecting each XML element or
-                attribute directly out of the reader's VARIANT output. The default value is ``False``, which
-                expands every record into one row per key with ``flatten()`` and reassembles it with a
-                dynamic ``pivot()``. Setting this to ``True`` removes that intermediate, which is
-                substantially faster on wide records, and makes the column order of a read without a schema
-                alphabetical. It has no effect when ``cacheResult`` is ``False`` and no schema is available,
-                since the keys cannot be discovered without materializing the result first.
+              + ``useVariantProjection``: A performance optimization that speeds up XML ingestion. The
+                default value is ``False``. Setting this to ``True`` requires ``cacheResult`` to also be
+                ``True`` (the default).
         """
         df = self._read_semi_structured_file(path, "XML")
         # AST.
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1321,6 +1323,10 @@ def xml(self, path: str, _emit_ast: bool = True) -> DataFrame:
                 When set to ``True`` (default), the result is cached and all subsequent operations on the DataFrame are performed on the cached data.
                 This means the actual computation occurs before :meth:`DataFrame.collect` is called.
                 When set to ``False``, the DataFrame is computed lazily and the actual computation occurs when :meth:`DataFrame.collect` is called.
+
+              + ``useVariantProjection``: A performance optimization that speeds up XML ingestion. The
```
sfc-gh-yixie, line 1327: Explain in what case this improves the performance? Users may not know when to use this parameter.

src/snowflake/snowpark/dataframe_reader.py
```diff
@@ -1365,19 +1371,49 @@ def _apply_xml_schema(self, df: DataFrame, schema: StructType) -> DataFrame:
...
             if corrupt_col_name in df_columns:
                 corrupt_ref = df[single_quote(corrupt_col_name)]
-                cols.append(
-                    corrupt_ref.cast(StringType()).alias(
-                        quote_name_without_upper_casing(corrupt_col_name)
-                    )
+                # Under variant projection the column is always present, so its presence
+                # alone doesn't mean a row was corrupt -- probe for a non-NULL value instead.
+                has_corrupt_record = (
+                    df.filter(corrupt_ref.is_not_null()).limit(1).count() > 0
```
sfc-gh-yixie, line 1377: Is `count` called here a must? So the DataFrame isn't lazy any more with the `count` call.

PR #4343, opened 2026-09-04
SNOW-3472759: support to_polars() eagerFrame interoperability

sfc-gh-mayliu:

   

   Fixes SNOW-3472759

This PR adds DataFrame.to_polars() to enable direct conversion from a Snowpark DataFrame to a Polars eager DataFrame. Cherry-picked from https://github.com/snowflakedb/snowpark-python/pull/4288 and removed LazyFrame implementation. LazyFrame will be a follow-up work upon remote data pruning becoming available.

- Finalized benchmark evaluation: https://docs.google.com/document/d/1jABNlY-zNEMvQZKGg-wKdwu4r7-XpxIGnWE50Vjxiyg/edit?usp=sharing
- Complementary Design doc: https://docs.google.com/document/d/1FKeSB5K9HIHG-aQwcP_2oPRmT7Pd4hcj03l8tictEPM/edit?usp=sharing
- Tests passed on sproc (params1 `test_df_`): https://buildkite.com/snowflake/snowflake--snowpark-python-sp-integ-tests/builds/246/tests?tags=new_failure%3A%21true%2Cinconclusive%3A%21true&query=test_python_sp_integ_dataframe

On an XL warehouse:
Env | Size | DF (default arrow) | DF (use_parquet)
-- | -- | -- | --
local | 100,000 | 4.01s | 2.50s
local | 1,000,000 | 7.60s | 6.81s
local | 10,000,000 | 21.13s | 10.86s
sproc | 100,000 | 2.23s | 1.56s
sproc | 1,000,000 | 7.12s | 4.54s
sproc | 10,000,000 | 12.63s | 4.20s

### DIFF
--- a/src/snowflake/snowpark/_internal/polars_backend.py
+++ b/src/snowflake/snowpark/_internal/polars_backend.py
@@ -1,0 +1,169 @@
+#
+# Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
+#
+from __future__ import annotations
+
+import uuid
+from concurrent.futures import ThreadPoolExecutor
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import polars as pl
+    from snowflake.snowpark.dataframe import DataFrame
+
+
+def _empty_frame_from_schema(
+    df: DataFrame, statement_params: dict[str, str] | None
+) -> pl.DataFrame:
+    """Return an empty polars DataFrame with the DataFrame's schema."""
+    import polars as pl
+
+    return pl.from_arrow(
+        df.limit(0).to_arrow(
+            statement_params=statement_params,
+            _emit_ast=False,
+        )
+    )
+
+
+def _open_stage_files_parallel(
+    session,
+    paths: list[str],
+    is_sproc: bool,
+    is_lazy: bool,
+    max_workers: int | None = None,
+) -> list:
+    """Open stage files concurrently.
+
+    * ``is_lazy=True``: return the raw file-like objects for
+      ``pl.scan_parquet`` to consume (Polars owns the handles from that point
+      on and closes them itself).
+    * ``is_lazy=False``: read each file with ``pl.read_parquet`` and return
+      the resulting ``pl.DataFrame`` (``with`` closes the handle after read).
+
+    Uses ``SnowflakeFile.open`` inside a stored procedure and
+    ``session.file.get_stream`` on the client.
+    """
+    if not paths:
+        return []
+
+    import polars as pl
+
+    if is_sproc:
+        from snowflake.snowpark.files import SnowflakeFile
+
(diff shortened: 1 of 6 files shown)


sfc-gh-yuwang (approved):

src/snowflake/snowpark/_internal/polars_backend.py
```diff
@@ -0,0 +1,135 @@
...
+    sub = f"{stage}/{sub_prefix}/{rid}/"
+    df.write.copy_into_location(
+        sub,
+        file_format_type="parquet",
+        format_type_options={"COMPRESSION": "SNAPPY"},
+        header=True,
+        overwrite=True,
+        statement_params=statement_params,
+    )
+    rows = df._session.sql(f"LIST '{sub}'").collect(statement_params=statement_params)
```
sfc-gh-yuwang, line 87: I remember LIST would return a not sorted result, will it be a issue here?

sfc-gh-yuwang (approved):
LGTM, left one comment

src/snowflake/snowpark/_internal/polars_backend.py
sfc-gh-mayliu: Thanks for bringing this up, I think it's fine. `COPY INTO` does not preserve parquet path order, so `LIST` being a downstream command won't have significant meanings to the originally unsorted results. Also Snowflake doesn't give row-order guarantee for queries without an explicit ORDER BY, so I think we can inherit this behavior in Snowpark

src/snowflake/snowpark/_internal/polars_backend.py
```diff
@@ -0,0 +1,135 @@
...
+
+    else:
+        opener = session.file.get_stream
+
+    def _work(p: str):
+        with opener(p) as f:
+            return pl.read_parquet(f)
+
+    with ThreadPoolExecutor(max_workers=max_workers) as ex:
+        return list(ex.map(_work, paths))
```
sfc-gh-aling, line 61: do we need to worry about partial failure (some file failed), and retry upon error?
or polar would handle for us?

you can take a look at connector to see if they have such logic

https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/file_transfer_agent.py

sfc-gh-mayliu: Good question, I don't think we need to handle this in Snowpark. Both client-path `get_stream` and sproc-path `SnowflakeFile.open` already have per-request retry and should handle vast majority of failures, so another retry within Snowpark would be redundant.

As for cross-file orchestration, the current behavior of one failure causing the entire call to fail is expected for this all-or-nothing design (if the above lower-level retry still does not help). There'd no meaningful partial result to return and could be worse to return a silently truncated DF.

PR #4348, opened 2026-09-08
SNOW-3898656: XML perf optimization to enable directory read and skip children (stacked 3)

sfc-gh-mayliu:

   

   Fixes SNOW-3898656

This PR adds `readDirectory`, `skipChildren`, and `includeSourcePos` options to `DataFrameReader.xml()`

Customer has around 50k smaller XML files placed in a directory with total size around 50GB. The previous stacked PR1 & PR2 optimizes performance for a single large 50GB XML file, and this PR3 & following PR4 solves the performance issue for 50k files totaling 50GB.

### Customer request
A common real-world XML shape wraps many child records inside one or more parent elements that carry file-level (or block-level) metadata as attributes:
```
<ANALYTICS>
  <INSTRUMENTS ASOF_DATE="11/29/2023" CREATE_DATE="11/30/2023" RECORDS="128553">
    <INSTRUMENT> ... 805 fields ... </INSTRUMENT>
    <INSTRUMENT> ... </INSTRUMENT>
    <!-- up to 128,553 more -->
  </INSTRUMENTS>
</ANALYTICS>
```
The customer wants ASOF_DATE, CREATE_DATE, and RECORDS from `INSTRUMENTS` attached to every flattened `INSTRUMENT` row. DataFrameReader.xml() already reads INSTRUMENT rows fine; the missing piece is getting the INSTRUMENTS attributes without re-reading or fully materializing the file a second time, and without guessing which parent a child belongs to when a file has more than one parent block.

### Technique

* **Previous Snowpark behavior:** `DataFrameReader.xml()` with `rowTag` set requires `path` to name one exact file. Passing a stage directory fails because the underlying UDTF calls `SnowflakeFile.open()` on that path, throwing *"File does not exist or not authorized"*. Reading many files today requires one `.xml()` call per file, unioned client-side by the caller.
* **Optimized behavior:**
  * `readDirectory=True`: Reads every file directly under a stage directory in one query plan — all files' byte ranges share a single worker-assignment table and are dispatched together, instead of $N$ separate plans.
  * `skipChildren=True`: Parses only the row tag's own opening-tag attributes and never descends into children — designed for fast header/parent-only extraction.
  * `includeSourcePos=True`: Emits `_source_byte_pos` and `_source_file_path` per record so child rows can be range-joined back to the parent block that encloses them — correct even with multiple parent blocks per file (unlike a `CROSS JOIN`).

### Limitations & Defaults

* Confirmed on SCOS, no regression or behavioral difference
* **Default:** All three options are opt-in (`False` by default). A plain `.xml(path)` call is unchanged.
* **Limitation:** `readDirectory` reads files directly under the given directory only (not subdirectories), and requires `rowTag` to be set.
* **Limitation:** `includeSourcePos` requires `useVariantProjection=True` and no schema provided — the extra typed columns cannot survive a fixed schema projection or the legacy pivot path.

---

#### Parent/Child Attribute Correlation (Single 50GB File)
*`skipChildren` + `includeSourcePos` + range JOIN vs. naive full-file-scan alternative:*

| Step | Time (s) |
| :--- | :--- |
| **Baseline** (no parent-attr correlation) | 477.71 |
| **Pass 1 — header** (`skipChildren=True`) | 67.93 |
| **Pass 2 — records** (`includeSourcePos=True`) | 483.65 |
| **Range JOIN** | 20.58 |
| **Total** | **573.32** (+20% vs. baseline) |
| *vs. full-file serial scan alternative* | *~1820 (+282%)* |

* Row/column output verified correct at every stage. The range-join is the only approach that stays correct when a file has more than one parent block (a `CROSS JOIN` alternative was measured at 128,553,000 rows / 2242s on a 1000-parent mock — confirmed wrong, not just slow).


src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -151,9 +152,73 @@
...
 
 # Default for numWorkers.
-# TODO SNOW-1983360: revisit once the UDTF scalability issue is resolved.
 DEFAULT_MAX_WORKERS: int = 16
 
+# Snowflake's own limit on expressions in a single VALUES list ("maximum number of
+# expressions in a list exceeded"), measured empirically. A directory of many large
+# files -- each needing multiple byte-range workers -- has no other ceiling on
+# worker-assignment row count.
+_XML_WORKER_ASSIGNMENT_MAX_ROWS_PER_VALUES: int = 200_000
```
sfc-gh-yuwang, line 161: is this limit document somewhere? If yes can you also put its link here?
One concern is that will this limit change during the time and make this parameter perf bottleneck silently?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -1850,6 +1915,36 @@ def process_list(list_property):
...
+        )
+
+    def _resolve_stage_file_size(self, file_path: str) -> int:
+        """Look up a single file's byte size via LIST, matched by basename since LIST's
+        reported name format is stage-type-dependent (e.g. a named external stage's LIST
+        reports the underlying cloud URL, unrelated to the stage's own name)."""
+        basename = _stage_listing_basename(file_path)
+        for row in self._list_stage_path(file_path):
+            if _stage_listing_basename(row["name"]) == basename:  # type: ignore
+                return int(row["size"])  # type: ignore
```
sfc-gh-yuwang, line 1930: i remember we used to determine the file size by get_file_size(), which is actually open the file with SnowflakeFile.open() and read bytes.
Now we are rely on the result of LIST stage_file.
Is it possible that LIST produce incorrect file size? Like when the file is compressed?(or maybe other scenarios)

sfc-gh-yuwang (approved):
pre approved assume all comments will be addressed

PR #4352, opened 2026-09-09
SNOW-4072353: preserve quoted column names for non-SELECT to_pandas() (Option B)

sfc-gh-fpawlowski:
1. Fixes SNOW-4072353

2. Pre-review checklist:
      - No new/changed public API signatures here, so no AST changes are needed.

## What
`DataFrame.to_pandas()` only relabels columns from `self._plan.attributes` when the driver can't return a real pandas DataFrame (Python Driver versions below v5 raise `NotSupportedError` for JSON-format non-SELECT results). The Python Driver on Universal Core (v5+) is Arrow-native and never hits that fallback for `SHOW`/`SET`/DDL, so it returns its own bare column labels for non-SELECT statements — diverging from the historically-quoted labels older driver versions produce (e.g. `'"status"'`).

**Updated approach (per review below):** rather than relabeling columns after the fact once v5+ has already returned a real DataFrame, `_to_data_or_iter` (`_internal/server_connection.py`) now forces v5+ to take the same JSON-fallback path pre-v5 drivers always took for non-Arrow result sets, via `force_json_fallback = to_pandas and IS_V5_DRIVER and results_cursor._query_result_format != "arrow"`. This makes `isinstance(result, pandas.DataFrame)` `False` for v5+ non-Arrow results too, so `dataframe.py`'s existing, unmodified `to_pandas()` logic already produces the correct quoted output — **no changes to `dataframe.py` are needed at all**. Also extracted the repeated `iter(results_cursor) if to_iter else results_cursor.fetchall()` into a shared `_json_fallback()` helper.

## Why
Draft PR (2 of 2) for SNOW-4072353 — see the Jira ticket for full root-cause analysis. This is the option that preserves today's quoted output as the intentional contract: no behavior change for pre-v5 driver users, and v5+ (Universal Core) driver output is brought into parity with it instead of the other way around.

See the sibling draft (`fix/snow-4072353-unquote-non-select`) for the alternative approach that instead unquotes non-SELECT results everywhere.

### Review history
@sfc-gh-pczajka requested changes on the original relabel-based approach, preferring to simulate the pre-v5 driver's behavior at the source rather than patch column labels downstream — this also restores other pre-v5 behavior (e.g. the "query result format is set to JSON" warning) that the relabel-only approach didn't. Implemented as requested; see commit history.

## Dependency — not yet runnable against a real v5 driver
`results_cursor._query_result_format` does not currently exist on the Universal Driver's Python bridge cursor. Confirmed:
- It **does** exist on the legacy connector (`cursor.py:410`/`1196`).
- The underlying value **is** already tracked in `sf_core` (`sf_core/src/rest/snowflake/query_response.rs:119`).
- It is **not** yet exposed on the v5 Python bridge cursor (checked `python/src/snowflake/connector/cursor/` and `_internal/cursor/` in the Universal Driver repo).

So this needs a small plumbing addition on the Universal Driver side before it can actually run against a real v5 connection — the data is already tracked, it just isn't exposed at the Python level yet.

## Test changes
None — this option is designed to leave existing test expectations (`test_to_pandas_non_select`, `test_sql_set_variable_to_pandas`) passing unchanged for pre-v5 drivers, and to make them pass under a v5+ driver too once the driver-side dependency above lands.

## Verification
Not yet run against a live account, and can't be meaningfully verified against v5 until `_query_result_format` is exposed there (see Dependency above).

Jira: https://snowflakecomputing.atlassian.net/browse/SNOW-4072353

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -25,7 +25,7 @@ from typing import (
 )
 
 from snowflake.connector import SnowflakeConnection, connect
-from snowflake.connector.constants import ENV_VAR_PARTNER, FIELD_ID_TO_NAME
+from snowflake.connector.constants import FIELD_ID_TO_NAME
 from snowflake.connector.cursor import ResultMetadata, SnowflakeCursor
 from snowflake.connector.errors import Error, NotSupportedError, ProgrammingError
 from snowflake.connector.network import ReauthenticationRequest
@@ -81,6 +81,9 @@ if TYPE_CHECKING:
         ResultMetadataV2 = ResultMetadata
 
 logger = getLogger(__name__)
+
+# backward compatibility constant
+ENV_VAR_PARTNER = "SF_PARTNER"
 
 # parameters needed for usage tracking
 PARAM_APPLICATION = "application"
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -467,7 +467,9 @@ class ServerConnection:
             )
             raise ex
 
-        notify_kwargs["requestId"] = str(results_cursor._request_id)
+        notify_kwargs["requestId"] = str(
+            results_cursor.request_id if IS_V5_DRIVER else results_cursor._request_id
+        )
         self.notify_query_listeners(
             QueryRecord(results_cursor.sfqid, results_cursor.query), **notify_kwargs
         )
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -28,8 +28,7 @@ from snowflake.connector import SnowflakeConnection, connect
 from snowflake.connector.constants import FIELD_ID_TO_NAME
 from snowflake.connector.cursor import ResultMetadata, SnowflakeCursor
 from snowflake.connector.errors import Error, NotSupportedError, ProgrammingError
-from snowflake.connector.network import ReauthenticationRequest
-from snowflake.connector.options import pandas
+from snowflake.snowpark._internal.options import pandas
 from snowflake.snowpark._internal.analyzer.analyzer_utils import (
     quote_name_without_upper_casing,
 )
@@ -54,6 +53,7 @@ from snowflake.snowpark._internal.telemetry import (
     get_plan_telemetry_metrics,
 )
 from snowflake.snowpark._internal.utils import (
+    IS_V5_DRIVER,
     create_rlock,
     create_thread_local,
     escape_quotes,
@@ -69,6 +69,11 @@ from snowflake.snowpark._internal.utils import (
     result_set_to_rows,
     unwrap_stage_location_single_quote,
 )
+
+if IS_V5_DRIVER:
+    from snowflake.connector.errors import ReauthenticationRequest
+else:
+    from snowflake.connector.network import ReauthenticationRequest
 from snowflake.snowpark import context
 from snowflake.snowpark.async_job import AsyncJob, _AsyncResultType
 from snowflake.snowpark.query_history import QueryListener, QueryRecord
--- a/src/snowflake/snowpark/dataframe.py
+++ b/src/snowflake/snowpark/dataframe.py
@@ -8,6 +8,7 @@ import datetime
 import itertools
 import random
 import re
+import typing
 from collections import Counter
 from decimal import Decimal
 from functools import cached_property, reduce
@@ -161,6 +162,7 @@ from snowflake.snowpark._internal.utils import (
     experimental,
     generate_random_alphanumeric,
     get_copy_into_table_options,
+    is_in_stored_procedure,
     is_snowflake_quoted_id_case_insensitive,
     is_snowflake_unquoted_suffix_case_insensitive,
     is_sql_select_statement,
@@ -238,6 +240,7 @@ from collections.abc import Iterable
 
 if TYPE_CHECKING:
     import modin.pandas  # pragma: no cover
+    import polars  # pragma: no cover
     from table import Table  # pragma: no cover
 
 _logger = getLogger(__name__)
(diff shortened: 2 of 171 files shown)


src/snowflake/snowpark/dataframe.py
```diff
@@ -1190,6 +1192,26 @@ def to_pandas(
                         for attr in self._plan.attributes
                     ],
                 )
+            elif not is_select_statement and IS_V5_DRIVER:
```
sfc-gh-pczajka: This gate is the part I'd like replaced — see the review summary for the shape.

Two problems with `not is_select_statement` as the trigger:

- `is_sql_select_statement` is `^(\s|\()*(select|with)` (`_internal/utils.py:134`), so a statement that *is* a SELECT but doesn't start with one lexically — `session.sql("/* tag */ select 1 as a")`, or a leading `--` comment — is classified non-SELECT. On legacy that query is Arrow-format, never reaches the fallback, and gets bare labels; here it would be relabeled to the quoted plan attributes. That's a new divergence from legacy in the opposite direction, created by the fix.
- The condition legacy actually uses is the result format, not the statement shape. Keying off the format makes the two drivers agree by construction rather than by approximation.

Independent of which shape we land: stripping leading comments in `is_sql_select_statement` is worth doing anyway, since the existing SELECT/non-SELECT unquote split above has the same exposure.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1190,6 +1192,26 @@ def to_pandas(
...
+                if len(attr_names) == len(result.columns):
+                    result.columns = attr_names
+                else:
+                    _logger.warning(
+                        "Could not relabel to_pandas() columns for a non-SELECT "
+                        "statement: expected %d columns from the query plan but "
+                        "the driver returned %d.",
+                        len(attr_names),
+                        len(result.columns),
+                    )
```
sfc-gh-pczajka: This is a positional rename guarded only on width, and `_plan.attributes` is not always the driver's column set.

`put_attributes()` (`_internal/analyzer/schema_utils.py:45-56`) lists 9 columns including `"encryption"`, but the Python driver returns 8: legacy omits it (`file_transfer_agent.py:831-840`) and the v5 driver deliberately mirrors that, emitting `encryption` only for the ODBC/JDBC result-set flavors. So `session.sql("put file://... @stage").to_pandas()` on v5 takes the `else` branch, logs a warning on a perfectly ordinary call, and keeps bare labels — no parity, plus log noise.

Routing through the existing fallback avoids the question: labels then come from the same `pandas.DataFrame(rows, columns=...)` construction legacy uses, and a genuine width mismatch surfaces the way it does on legacy instead of being swallowed into a warning.

src/snowflake/snowpark/dataframe.py
```diff
@@ -1170,16 +1171,17 @@ def to_pandas(
             )
 
         if block:
+            query = self._plan.queries[-1].sql.strip().lower()
```
sfc-gh-pczajka: With the fallback forced in `_to_data_or_iter`, this hoist and the `is_select_statement` computation can move back inside the `if not isinstance(...)` block below — `dataframe.py` then has no diff at all.

sfc-gh-pczajka (changes requested):
Requesting changes — I'd try to simulate old driver behavior as mych as possible - let's expose _query_result_format in connector (should be easy), and shape the logic like this: 

In `_to_data_or_iter` (`_internal/server_connection.py:584`):

```python
force_json_fallback = (
    IS_V5_DRIVER and results_cursor._query_result_format != "arrow"
)
if to_pandas and not force_json_fallback:
    try:
        data_or_iter = ...                                   # unchanged
    except NotSupportedError:
        data_or_iter = _json_fallback(results_cursor, to_iter)   # old Snowpark fallback
elif to_pandas:
    data_or_iter = _json_fallback(results_cursor, to_iter)    # old Snowpark fallback
```

sfc-gh-fpawlowski:
@sfc-gh-pczajka Implemented as suggested — pushed to this branch.

`_to_data_or_iter` now forces the same JSON-fallback path pre-v5 drivers always took for non-Arrow result sets on v5+ too (`force_json_fallback = to_pandas and IS_V5_DRIVER and results_cursor._query_result_format != "arrow"`), extracted the duplicated fallback expression into `_json_fallback()`, and reverted `dataframe.py` entirely — it needed zero changes once v5+ takes the same code path pre-v5 always did.

One thing worth flagging before this can actually be tested end-to-end: I checked the Universal Driver repo and `_query_result_format` isn't exposed on the v5 Python bridge's cursor yet (checked `python/src/snowflake/connector/cursor/` and `_internal/cursor/`). The underlying value is already tracked in `sf_core` (`query_response.rs:119`), and it already exists under that exact name on the legacy connector — so it looks like straightforward plumbing, matching your "should be easy." Do you know who owns adding that, or should I file something on the driver side? This PR can't be verified against a real v5 connection until it lands.

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -594,7 +594,20 @@ def _to_data_or_iter(
...
             results_cursor = new_cursor
 
-        if to_pandas:
+        # The Python Driver on Universal Core (v5+) is Arrow-native and
+        # succeeds at fetch_pandas_all() even for JSON-format result sets
+        # (e.g. SHOW/SET/DDL), where a pre-v5 driver always raised
+        # NotSupportedError and fell back to a plain row fetch. Force the
+        # same fallback on v5+ for non-Arrow result sets so to_pandas()
+        # behaves identically across driver generations instead of
+        # depending on which driver generation happens to be installed.
```
sfc-gh-pczajka: Can be shortened :D

```suggestion
        # Python Driver v5+ supports pandas conversion from JSON-format results
        # sets, but result slightly differs from output of Snowpark custom fallback.
        # Enforce Snowpark fallback for backwards compatibility.
```

sfc-gh-pczajka (approved):

src/snowflake/snowpark/_internal/server_connection.py
sfc-gh-fpawlowski: Applied — pushed. Thanks!

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -594,7 +594,16 @@ def _to_data_or_iter(
...
             results_cursor = new_cursor
 
-        if to_pandas:
+        # Python Driver v5+ supports pandas conversion from JSON-format result
+        # sets, but results slightly differ from output of Snowpark's custom
+        # fallback. Enforce Snowpark's fallback for backwards compatibility.
+        force_json_fallback = (
+            to_pandas
+            and IS_V5_DRIVER
+            and results_cursor._query_result_format != "arrow"
```
sfc-gh-yuwang, line 610: If I understand correctly, this will force result format to be arrow when using V5 driver ?

sfc-gh-pczajka: No, V5 driver converts json to arrow by design, but remembers it used to be json in `_query_result_format`. This gives us advantages, like to_pandas support for json for free, but caused problem in Snowpark:
- v4 driver just raised an error, so Snowpark implemented its own to_pandas logic
- this custom logic slightly differs from what python driver does for "arrow" (v4 or v5)
We did not want to cause result difference here, so this line means:

"When converting to_pandas, do not use v5 driver support for JSON result format - use Snowpark custom logic instead, replicating behavior for v4"

sfc-gh-yuwang (approved):

src/snowflake/snowpark/_internal/server_connection.py
```diff
@@ -587,14 +587,33 @@ def _to_data_or_iter(
...
+        # Python Driver v5+ supports pandas conversion from JSON-format result
+        # sets, but results slightly differ from output of Snowpark's custom
+        # fallback. Enforce Snowpark's fallback for backwards compatibility.
+        #
+        # A cursor loaded by get_results_from_sfqid is exempt. Pre-v5 drivers load
+        # such a result by re-running RESULT_SCAN, which is a SELECT, so their
+        # result format is "arrow" there whatever the original statement was and
+        # the fallback never ran. v5+ reads the stored result instead and reports
+        # the original statement's format, so the format is not a usable signal
+        # on this path.
```
sfc-gh-pczajka:
```suggestion
        # Python Driver v5+ supports pandas conversion from JSON-format result
        # sets, but results slightly differ from output of Snowpark's custom
        # fallback. Enforce Snowpark's fallback for backwards compatibility.
        #
        # from_query_id exception - pre-v5 driver loads such a result by re-running
        # RESULT_SCAN (default arrow format), v5+ uses REST API getting
        # the original statement's format.
```

PR #4355, opened 2026-09-09
SNOW-4081338: verify downloaded protoc against a pinned SHA-256 digest

sfc-gh-yuwang:

   Fixes SNOW-4081338

   - [x] I acknowledge that I have ensured my changes to be thread-safe (no runtime code changed — build tooling and a unit test only).

`.github/scripts/install_protoc.sh` downloaded the `protoc` release archive over the network, extracted it, put the binary on `PATH` and executed it — with no integrity check on the bytes. TLS authenticates GitHub's endpoint but cannot detect a release asset swapped through a compromised upstream account (CWE-494).

This is not CI-only. `.github/workflows/python-publish.yml` runs, on `release: published`, with `contents: write` + `id-token: write` + `PYPI_API_TOKEN`:

`install_protoc.sh` → `tox -e protoc` → `python -m build` → `sigstore sign` → publish to PyPI.

Code execution happens **before** signing, so a poisoned `protoc` yields a validly Sigstore-signed poisoned wheel shipped to customers.

### Fix

The installer now pins SHA-256 digests per platform and **verifies the downloaded ZIP before extracting it**, then re-verifies the extracted executable before it reaches `PATH`. Verifying the archive first also covers the `include/google/protobuf/*.proto` descriptors that feed codegen (not just the binary) and closes the zip-slip window. Verification is fail-closed: a mismatch, a missing file, an unpinned platform, or the absence of any SHA-256 utility all abort the job.

Download source and version (protoc 3.20.1) are unchanged, so build behaviour is identical for an authentic asset. Also hardened: `curl -L` → `curl -fL` (fail on HTTP error) and `mkdir` → `mkdir -p` (bare `mkdir` failed on re-run).

The install body is wrapped in a `[ "${BASH_SOURCE[0]:-$0}" = "$0" ]` guard so the verification helpers can be sourced by the test without downloading anything. Direct execution (how every workflow invokes it) is unaffected.

### Digest provenance

Every digest was verified by downloading the upstream asset and hashing it; the inner binary digests independently cross-confirm the archive digests.

| platform | archive | binary |
|---|---|---|
| linux-x86_64 | `3a0e900f…b265562` | `4231ac3f…d75257fe` |
| osx-x86_64 | `b4f36b18…be4cf33c` | `e2ad4603…d9afb526` |
| win64 | `897bf86b…e922cd46` | `7be0d716…dfd50030` |

### Bumping protoc

The digests are pinned constants, so a version bump has to update them. To keep that mechanical rather than manual, the installer gained a developer-only helper:

```
$ .github/scripts/install_protoc.sh --print-digests 3.20.1
# ---- paste over the digest constants in install_protoc.sh (protoc 3.20.1) ----
PROTOC_ZIP_SHA256_linux_x86_64="3a0e900f..."
PROTOC_ZIP_SHA256_osx_x86_64="b4f36b18..."
PROTOC_ZIP_SHA256_win64="897bf86b..."

PROTOC_BIN_SHA256_linux_x86_64="4231ac3f..."
PROTOC_BIN_SHA256_osx_x86_64="e2ad4603..."
PROTOC_BIN_SHA256_win64="7be0d716..."
```

It downloads every supported platform archive for the given version (default: the pinned `PROTOC_VERSION`), hashes both the ZIP and the executable inside it, and prints the block to paste. So bumping protoc is: run it with the new version, paste, review. All three platforms are covered from one machine, and the two halves of each pin cannot drift apart.

Progress goes to stderr and the paste block to stdout, so the output can be redirected cleanly. The helper installs nothing and touches neither `PATH` nor `GITHUB_PATH`.

`--help` documents both modes. CI continues to invoke the script with **no arguments**, which is unchanged; an unrecognised flag now aborts instead of silently installing.

**Self-check:** run against the currently pinned 3.20.1, the helper reproduces all six committed constants byte-for-byte — verified by grepping each emitted line back into the script. It was also exercised against a different upstream version (21.1) to confirm the bump path works end to end.

### Verification performed

- Ran the script end-to-end on macOS with an overridden `HOME` and stubbed dep installers: exit 0, both integrity checks pass, `libprotoc 3.20.1`.
- Fail-closed proven: served a tampered archive via a shim `curl` → exit 1, nothing extracted, `GITHUB_PATH` never appended, `protoc` never on `PATH`.
- `tests/unit/test_install_protoc_integrity.py`: 15 tests (sources the script and drives the verify helper: accepts a match; rejects tampered bytes, wrong digest, empty digest, missing file, absent digest tooling; asserts verify-zip → unzip → verify-binary ordering; plus four covering the new argument handling — `--help` and `--print-digests` must not enter the install path or write `GITHUB_PATH`, an unknown flag must fail closed, and the zip/binary pin tables must not disagree). **15 pass on this branch, 15 fail against `main`.** No network access in the tests.
- `bash -n` clean; `pre-commit run --files` all hooks pass.

### Scope / residual gap

Digest pinning must be updated deliberately on any protoc bump — a stale pin fails the build closed, which is the intended direction, and `--print-digests` makes the update mechanical. Other unpinned build tooling (the `mypy-protobuf` install in this same script) is SNOW-4081342's scope; that PR conflicts textually with this one on the final block of `install_protoc.sh`, and the resolution is trivial (keep this PR's guard indentation, keep that PR's version pin) — I have verified the merged result runs correctly and both test suites pass.

No CHANGELOG entry: this changes only CI/build tooling and nothing user-facing. Consistent with existing practice — no entry in the CHANGELOG history references workflows or the protoc installer.

### DIFF
--- a/.github/scripts/install_protoc.sh
+++ b/.github/scripts/install_protoc.sh
@@ -12,8 +12,6 @@ set -eu
 
 SCRIPT_NAME="$(basename "$0")"
 
-echo "${SCRIPT_NAME} is running... "
-
 PROTOC_VERSION=3.20.1
 PROTOC_OS_ARCH=""
 PROTOC_ZIP=""
@@ -39,8 +37,13 @@ EXPECTED_BIN_SHA256=""
 # extracted executable is then re-verified as a defence-in-depth check on the
 # unzip step itself.
 #
-# To bump PROTOC_VERSION, download each archive from the release URL below and
-# record `sha256` of the ZIP and of `bin/protoc` (`bin/protoc.exe` on Windows).
+# To bump PROTOC_VERSION, run
+#
+#     .github/scripts/install_protoc.sh --print-digests <new-version>
+#
+# and paste the emitted block over the constants below. The helper downloads
+# every supported platform archive and hashes both the ZIP and the executable
+# inside it, so the pins cannot drift out of sync with each other.
 PROTOC_ZIP_SHA256_linux_x86_64="3a0e900f9556fbcac4c3a913a00d07680f0fdf6b990a341462d822247b265562"
 PROTOC_ZIP_SHA256_osx_x86_64="b4f36b18202d54d343a66eebc9f8ae60809a2a96cc2d1b378137550bbe4cf33c"
 PROTOC_ZIP_SHA256_win64="897bf86b9c989f91c4171c7f99e3886fedfceb077a94dd150f1401cfe922cd46"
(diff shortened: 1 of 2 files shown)


sfc-gh-yuwang:
Closing: remediation for this batch is being scoped to SNOW-4081340 and SNOW-4081342 only.

Recording the residual risk so it is not lost. `install_protoc.sh` does run on the release path — it is step 5 of `python-publish.yml`, and it completes before the sigstore signing step, verified on the v1.54.0 release (run 30496704565: protoc installed 22:36:0x, signing 22:36:34–22:36:43, PyPI upload 22:36:52). So the CWE-494 exposure in SNOW-4081338 remains open, and SNOW-4081338 is not resolved by closing this PR.

The branch `SNOW-4081338-security-fix` is left in place if this is revisited. It contains the verified digest pins for all three platforms plus the `--print-digests` helper for protoc version bumps.

PR #4357, opened 2026-09-09
SNOW-4081340: pin release workflow actions to immutable commit SHAs

sfc-gh-yuwang:

   Fixes SNOW-4081340

   - [x] I acknowledge that I have ensured my changes to be thread-safe (no runtime code changed — a workflow and a unit test only).

`.github/workflows/python-publish.yml` referenced the PyPI publisher as `pypa/gh-action-pypi-publish@release/v1` — a **mutable branch ref** (CWE-829). An attacker with maintainer control of that public action repository can retarget the branch, and GitHub then resolves and executes it at release time inside the privileged `deploy` job (`contents: write`, `id-token: write`, `PYPI_API_TOKEN`).

Two consequences: the action can exfiltrate `PYPI_API_TOKEN` and retain it for future unauthorized uploads until manually revoked; and because the sigstore signing/verification happens at an **earlier** step with no digest re-check afterwards, it can substitute a different distribution in the shared workspace and upload that instead.

### Changes

- All three actions in this workflow pinned to full 40-hex commit SHAs with `# vX.Y.Z` comments. Leaving `checkout`/`setup-python` on mutable tags in the same privileged job would defeat the purpose.
- A digest guard: `Record artifact digests` snapshots `sha256sum dist/*` immediately after signing, then `sha256sum --check --strict` runs before publish and again after publish (`if: always()`).

| action | old ref | pinned SHA | version |
|---|---|---|---|
| `pypa/gh-action-pypi-publish` | `@release/v1` | `dc37677b2e1c63e2034f94d8a5b11f265b73ba33` | v1.14.2 |
| `actions/checkout` | `@v4` | `11d5960a326750d5838078e36cf38b85af677262` | v4.4.0 |
| `actions/setup-python` | `@v3` | `3542bca2639a428e1796aaa6a2ffef0c0f575566` | v3.1.4 |

Every SHA was verified against the GitHub API. `v1.14.2` is an annotated tag object (`a892a5a6…`) which dereferences to `dc37677b…`; that is also the current `release/v1` head, so **pinning is resolution-neutral today**. For `checkout`/`setup-python`, `git/matching-refs/tags` confirms both the major tag and the point release sit on the pinned commit — i.e. each pin equals exactly what the mutable tag resolved to. No major-version bumps were taken, so this is behaviour-preserving.

### On the other two suggested fixes

- **Trusted publishing (OIDC)** is the strongest control since it removes the long-lived token entirely, and this job already has `id-token: write`. It is **deliberately not done here**: it requires registering this repo/workflow as a trusted publisher on PyPI first, and removing the token before that exists breaks the next release. Recommended sequencing: register the trusted publisher on PyPI → merge a follow-up dropping the `with: user/password` block → revoke `PYPI_API_TOKEN`.
- **Splitting build/sign from a digest-checked publish job** would force an artifact round trip and a workflow rewrite. The in-place digest guard achieves the detection goal in three readable steps. Being precise: the post-publish check is **detection, not prevention** — the upload has already happened — but it converts a silent artifact swap into a failed run.

`permissions` are unchanged: `contents: write` is required by the three `gh release upload` steps and `id-token: write` by sigstore. Neither is removable without breaking the release, and there is only one job, so job-scoping would be cosmetic.

### Verification performed

- YAML parses; `bash -n` clean on both new shell bodies.
- `actionlint` 1.7.12: 9 shellcheck info-level findings, **byte-identical in count and location to the same run against `main`** (all pre-existing, on the `Signing` and `Copy files to release` steps). **Zero findings on the new steps.** Pre-existing ones left alone as out of scope.
- Behaviour preservation reviewed line by line: step order, `with:`/`env:` inputs, and the sigstore `--cert-identity …/python-publish.yml@${GITHUB_REF}` assertion all unchanged. `dist.sha256` is written to the repo root and does not match the `*.sigstore`/`*.sig`/`*.crt` upload globs.
- `tests/unit/test_release_workflow_action_pinning.py`: **2 pass on this branch, 2 fail against `main`**, correctly naming all three mutable refs.

### Not verified / follow-ups

- No real release was executed; the pins and digest steps are statically verified only. Worth watching on the first release: if `pypa/gh-action-pypi-publish` ever writes into `dist/`, the post-publish check would false-positive. It is `if: always()` and runs last, so that would fail the run *after* a successful upload rather than block publishing.
- SHA pinning stops automatic upstream security updates. A `.github/dependabot.yml` entry for `github-actions` should be added so pins are refreshed as reviewable PRs — intentionally not bundled here.
- Whether a PyPI trusted publisher is already registered is not visible to me.

No CHANGELOG entry: this changes only the release workflow and nothing user-facing, consistent with existing practice (no entry in the CHANGELOG history references workflows or GitHub Actions).

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -22,9 +22,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
     - name: Set up Python
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@3542bca2639a428e1796aaa6a2ffef0c0f575566 # v3.1.4
       with:
         python-version: '3.11'
     - name: Install dependencies
@@ -73,6 +73,11 @@ jobs:
         done
     - name: List artifacts after sign
       run: ls ./dist
+    - name: Record artifact digests
+      # Snapshot the digests of the signed, verified distributions so that any
+      # substitution of dist/* by a later step can be detected.
+      run: |
+        sha256sum dist/* | tee dist.sha256
     - name: Copy files to release
       run: |
         gh release upload ${{ github.event.release.tag_name }} *.sigstore
@@ -80,8 +85,19 @@ jobs:
         gh release upload ${{ github.event.release.tag_name }} *.crt
       env:
         GITHUB_TOKEN: ${{ github.TOKEN }}
+    - name: Verify artifact digests before publish
+      # Fail the release before anything is uploaded if the signed
+      # distributions no longer match what sigstore verified above.
+      run: |
+        sha256sum --check --strict dist.sha256
     - name: Publish package
-      uses: pypa/gh-action-pypi-publish@release/v1
+      uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
       with:
         user: __token__
         password: ${{ secrets.PYPI_API_TOKEN }}
+    - name: Verify artifact digests after publish
+      # Defence in depth: detect a third-party publish action that swapped a
+      # distribution in the shared workspace before uploading it to PyPI.
+      if: always()
+      run: |
+        sha256sum --check --strict dist.sha256
--- a/tests/unit/test_release_workflow_action_pinning.py
+++ b/tests/unit/test_release_workflow_action_pinning.py
@@ -1,0 +1,68 @@
+#
+# Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
+#
+"""Guards the release workflow against mutable third-party action references.
+
+``.github/workflows/python-publish.yml`` runs in a privileged job (``contents:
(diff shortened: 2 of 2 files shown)


tests/unit/test_release_workflow_action_pinning.py
sfc-gh-aling, line 1: I don't think we need a unit test for a publish yml change, this feels overkilling

.github/workflows/python-publish.yml
```diff
@@ -22,9 +22,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
     - name: Set up Python
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@3542bca2639a428e1796aaa6a2ffef0c0f575566 # v3.1.4
```
sfc-gh-aling: TBH I doubt the value of pinning the dep, we won't be able to receive latest fix automatically.

assuming gh action introduces vulnerability fix in 4.4.1, we need to manually update the yml now.

is it still the guidance from our security partner?

sfc-gh-aling (approved):

PR #4361, opened 2026-09-09
SNOW-4081342: pin build/release tooling to exact versions

sfc-gh-yuwang:

   Fixes SNOW-4081342

   - [x] I am adding a new dependency (no new dependency — existing build-time dependencies are pinned to exact versions)
   - [x] I acknowledge that I have ensured my changes to be thread-safe (no runtime code changed — build tooling and a unit test only).

The release/build pipeline resolved the `mypy-protobuf` build tool from PyPI with **no version pin**, then executed its console entry point `protoc-gen-mypy` while producing the officially signed distribution (CWE-829). An attacker publishing a newer `mypy-protobuf` could run arbitrary code in the release job — which holds `contents: write`, `id-token: write`, and `PYPI_API_TOKEN` — and ship a signed, poisoned SDK.

Release path: `install_protoc.sh` → `tox -e protoc` → `python -m build` (whose `setup.py` `build_py` invokes `protoc-gen-mypy` via `--mypy_out`) → `sigstore sign` → publish.

### Changes

- **`.github/scripts/install_protoc.sh`** — `mypy-protobuf==5.1.0` in both the `uv` and `pip` branches.
- **`tox.ini` `[testenv:protoc]`** — `mypy-protobuf==5.1.0`; `protobuf==6.33.6` (was unpinned). Invoked by `python-publish.yml` via `tox -e protoc`.
- **`pyproject.toml` `[build-system] requires`** — `mypy-protobuf==5.1.0` for py<3.14, `==3.6.0` for py>=3.14 (was `<=3.6.0`), `setuptools>=40.6.0,<85`.
- **`tests/unit/test_build_dependency_pins.py`** — asserts exact pinning across all three files.

### Important: 5.1.0, not 3.6.0

An obvious-looking "consistency" fix here is to pin `mypy-protobuf==3.6.0` everywhere, matching the py>=3.14 constraint. **That would be a silent regression.** `mypy-protobuf` latest is 5.1.0; 3.6.0 is from April 2024. The `python_version < '3.14'` branch is what the release job actually uses (`python-publish.yml` sets up Python **3.11**), so it resolves 5.1.0 today — pinning it to 3.6.0 would downgrade the stub generator two major versions.

Measured by running the real `[testenv:protoc]` codegen command in throwaway venvs on Python 3.11:

- **5.1.0 vs 3.6.0** — `ast_pb2.py` byte-identical (emitted by `protoc`, not `mypy-protobuf`), but `ast_pb2.pyi` differs **materially**: ~900 fewer lines and every annotation spelled differently. 3.6.0 emits plain unaliased imports and gates `sys.version_info >= (3, 10)` with `import typing as typing_extensions`; 5.1.0 emits `import builtins as _builtins` / `from collections import abc as _abc` / `import typing as _typing` and gates `>= (3, 11)`.
- **pinned (5.1.0 + protobuf 6.33.6) vs today's unpinned resolve (5.1.0 + protobuf 7.36.1)** — both generated files **byte-identical**. So this PR does not change what ships.

The py>=3.14 pin stays at 3.6.0; the existing comment documents why (mypy-protobuf 5.x's bundled `extensions_pb2.py` was generated with protobuf 6.x gencode and fails to import when the py3.14 build-isolation venv resolves protobuf 5.x). `<=3.6.0` already resolved to 3.6.0, so that branch is unchanged in effect.

### Why `protobuf==6.33.6` and not latest

The unpinned resolve pulls 7.36.1, which is outside the `protobuf>=3.20,<6.34` ceiling `setup.py` declares for the runtime. 6.33.6 is the highest release under that ceiling and codegen output is byte-identical either way, so the pin is free.

### Why `setuptools` is bounded, not exact-pinned

`pyproject.toml` ships inside the sdist, so it binds conda-forge and distro source builds on Pythons that do not exist yet; an exact pin would predictably break those, since new CPython releases generally need newer setuptools. This is an explicit **partial** mitigation — it still admits new `84.x` releases. Flagging rather than claiming closed.

### Verification performed

- Codegen equivalence measured as described above (pinned vs unpinned byte-identical; 3.6.0 materially different).
- `pyproject.toml` parses via `tomllib`; `tox.ini` parses; `bash -n` clean on the script.
- `tests/unit/test_build_dependency_pins.py`: **8 pass on this branch, 6 of 8 fail against `main`.**

### Limitations

Exact-version pinning does not stop a compromised **re-upload** of an existing version; only full transitive hash pinning (`--require-hashes`) does, which is platform/interpreter-specific and a reasonable follow-up. Transitive deps still float (`types-protobuf` via `mypy-protobuf`). Integrity of the `protoc` download itself is SNOW-4081338.

This PR conflicts textually with SNOW-4081338 on the final block of `install_protoc.sh`. Resolution is trivial (keep that PR's sourcing-guard indentation, keep this PR's version pin); I have verified the merged result runs correctly and all 19 tests across both suites pass.

No CHANGELOG entry: nothing user-visible changes. Codegen output is byte-identical to today's resolve, and existing CHANGELOG precedent for dependency entries covers runtime `install_requires`, not build-time tooling.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,13 +1,23 @@
 [build-system]
 requires = [
-    "setuptools",
+    # Bounded rather than exact-pinned: this file ships inside the sdist, so an
+    # exact setuptools pin would break source builds (conda-forge, distros) on
+    # newer Pythons that require a newer setuptools.
+    "setuptools>=40.6.0,<85",
     # py>=3.14: protoc-wheel-0 has no cp314 wheel on PyPI, so it cannot be installed
     # in pip's build isolation venv. setup.py falls back to shutil.which("protoc").
     "protoc-wheel-0==21.1; python_version < '3.14'", # Protocol buffer compiler for Snowpark IR
-    # py>=3.14: pin <=3.6.0 so the build isolation venv does not install mypy-protobuf 5.x,
+    # mypy-protobuf runs its protoc-gen-mypy entry point during the signed release
+    # build, so it must be exact-version pinned (CWE-829): an unpinned resolve lets a
+    # newly published release execute arbitrary code in the release job.
+    # py>=3.14: pin ==3.6.0 so the build isolation venv does not install mypy-protobuf 5.x,
     # whose extensions_pb2.py was generated with protobuf 6.x gencode and fails at import
     # time when the isolated venv resolves protobuf 5.x (Snowflake internal channel).
-    "mypy-protobuf<=3.6.0; python_version >= '3.14'",
-    "mypy-protobuf; python_version < '3.14'", # used in generating typed Python code from protobuf for Snowpark IR
+    "mypy-protobuf==3.6.0; python_version >= '3.14'",
+    # py<3.14: 5.1.0 is the version an unpinned resolve already installs today, so this
+    # pin is status-quo-preserving. Do NOT "align" it with the 3.6.0 pin above: 3.6.0
+    # emits materially different .pyi stubs (non-aliased imports, different
+    # typing_extensions handling, ~900 fewer lines) and would be a silent downgrade.
+    "mypy-protobuf==5.1.0; python_version < '3.14'", # used in generating typed Python code from protobuf for Snowpark IR
 ]
 build-backend = "setuptools.build_meta"
--- a/tests/unit/test_build_dependency_pins.py
+++ b/tests/unit/test_build_dependency_pins.py
@@ -1,0 +1,194 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
+#
+"""Guards against unpinned build/release tooling (SNOW-4081342, CWE-829).
+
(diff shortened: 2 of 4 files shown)


tests/unit/test_build_dependency_pins.py
sfc-gh-aling, line 1: same comment as the other PR that unit test for building script seems overkilling

pyproject.toml
```diff
@@ -1,13 +1,23 @@
 [build-system]
 requires = [
-    "setuptools",
```
sfc-gh-aling, line 3: do we need change in meta.yaml as well for anaconda build?

sfc-gh-yuwang: conda-forge has no mypy-protobuf 3.6.0
conda-forge's mypy-protobuf 5.1.0 requires protobuf >=6.32, which can't satisfy the recipe's existing pin.

so basically it cannot be aligned and already diverged. Thus I decided to not change meta.yml

sfc-gh-aling (approved):

PR #4365, opened 2026-09-09
Fix ai_extract FILE input routing

sfc-gh-mayliu:

   

   Fixes SNOW-NNNNNNN

This PR fixes a bug where `ai_extract` misrouted `FILE` columns as text when `scores` or `config` arguments were used.

`ai_extract` previously only recognized a `FILE` input when it was an inline `to_file()` call. A `FILE` column computed in an earlier step and referenced later by name got silently routed as text, failing with the error: `named arguments do not match any signature for function AI_EXTRACT`.
https://docs.snowflake.com/en/sql-reference/functions/ai_extract#syntax
* Keeps `input` and `response_format` as positional arguments. This lets Snowflake resolve `FILE` vs. text type, and appends `scores` and `config` as trailing named arguments if applicable
* Added a regression test with a `FILE` column built in an earlier step and referenced by name -- confirmed the test fails on main and passes with this fix.

### DIFF
--- a/src/snowflake/snowpark/functions.py
+++ b/src/snowflake/snowpark/functions.py
@@ -13173,40 +13173,27 @@ def ai_extract(
     else:
         ast = None
 
-    # Use named-argument form when scores or config is requested
-    if scores is not None or config is not None:
-        response_format_col = sql_expr(
-            _python_obj_to_sql_literal(response_format), is_constant=True
-        )
-        # Detect file vs text input: TO_FILE() calls produce a FunctionExpression named "to_file"
-        is_file = (
-            isinstance(input_col, Column)
-            and isinstance(input_col._expr1, FunctionExpression)
-            and input_col._expr1.name.upper() == "TO_FILE"
-        )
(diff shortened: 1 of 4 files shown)


src/snowflake/snowpark/functions.py
```diff
@@ -13173,40 +13173,27 @@ def ai_extract(
...
         _python_obj_to_sql_literal(response_format), is_constant=True
     )
+    call_args = [input_col, response_format_col]
+    if scores is not None:
+        call_args.append(
+            sql_expr(
+                f"scores => {_python_obj_to_sql_literal(scores)}", is_constant=True
+            )
+        )
+    if config is not None:
```
sfc-gh-yuwang, line 13188: the config route seems have no test coverage, can we also add some test for it?

sfc-gh-mayliu: Good point, added more unit/integ tests

sfc-gh-yuwang (approved):

PR #4367, opened 2026-09-10
SNOW-4072353: force JSON fallback for non-Arrow results on v5+ (implements review on #4352)

sfc-gh-fpawlowski:
## What
Implements @sfc-gh-pczajka's review on #4352, stacked on top of it so the diff shows exactly what changes on top of that draft.

Instead of relabeling `to_pandas()` columns after the fact once the v5+ driver has already returned a real `pandas.DataFrame`, this simulates the pre-v5 driver's behavior at the source: `_to_data_or_iter` (`server_connection.py`) now forces the same JSON-fallback path pre-v5 drivers always took for non-Arrow result sets (SHOW/SET/DDL) on v5+ too, via `force_json_fallback = to_pandas and IS_V5_DRIVER and results_cursor._query_result_format != "arrow"`.

Net effect: `dataframe.py`'s `to_pandas()` needs **zero changes** — `isinstance(result, pandas.DataFrame)` is `False` for v5+ non-Arrow results too, same as pre-v5, so the existing unmodified ternary (`unquote_if_quoted(...) if is_select_statement else attr.name`) already produces the correct quoted output. This commit removes #4352's relabel branch as no longer necessary.

Also extracted the repeated `iter(results_cursor) if to_iter else results_cursor.fetchall()` into a shared `_json_fallback()` helper (was duplicated twice already, pre-existing).

## Dependency — this is not yet runnable on a real v5 driver
`results_cursor._query_result_format` does not currently exist on the Universal Driver's Python bridge cursor. I confirmed:
- It **does** exist on the legacy connector: `cursor.py:410`/`1196` (`self._query_result_format = data.get("queryResultFormat", "json")`).
- The underlying value **is** already tracked deep in `sf_core`: `sf_core/src/rest/snowflake/query_response.rs:119` (`pub query_result_format: Option<String>`).
- It is **not** yet exposed on the v5 Python bridge cursor — no matches under `python/src/snowflake/connector/cursor/` or `_internal/cursor/` in the Universal Driver repo.

So the data already exists internally; this needs a small plumbing addition on the Universal Driver side (a `@property` exposing what `sf_core` already tracks) before this branch can actually run against a real v5 connection. Opening this as a draft specifically to make that dependency concrete and visible, and to get eyes on whether the `server_connection.py` shape itself looks right before pursuing the driver-side change.

## Pros vs. #4352's current approach
- Single source of truth: reuses `dataframe.py`'s existing, already-tested logic verbatim for both driver generations instead of adding a second column-labeling code path.
- True behavioral parity (not just labels) — e.g. the "query result format is set to JSON" warning at `dataframe.py:1177-1181` also now fires correctly for v5+ non-SELECT statements, matching legacy, which #4352's relabel-only approach did not restore.

## Cons
- Real cross-repo dependency on the Universal Driver exposing `_query_result_format` (see above) — blocks this from being mergeable/testable against v5 until that lands.
- Touches the shared `_to_data_or_iter` path used by every `to_pandas()`/`to_arrow()` call, not just the narrow non-SELECT case, though the added condition is narrowly gated.

Jira: https://snowflakecomputing.atlassian.net/browse/SNOW-4072353

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/src/snowflake/snowpark/_internal/server_connection.py
+++ b/src/snowflake/snowpark/_internal/server_connection.py
@@ -594,7 +594,20 @@ class ServerConnection:
             new_cursor.get_results_from_sfqid(qid)
             results_cursor = new_cursor
 
-        if to_pandas:
+        # The Python Driver on Universal Core (v5+) is Arrow-native and
+        # succeeds at fetch_pandas_all() even for JSON-format result sets
+        # (e.g. SHOW/SET/DDL), where a pre-v5 driver always raised
+        # NotSupportedError and fell back to a plain row fetch. Force the
+        # same fallback on v5+ for non-Arrow result sets so to_pandas()
+        # behaves identically across driver generations instead of
+        # depending on which driver generation happens to be installed.
+        force_json_fallback = (
+            to_pandas
+            and IS_V5_DRIVER
+            and results_cursor._query_result_format != "arrow"
+        )
+
+        if to_pandas and not force_json_fallback:
             try:
                 data_or_iter = (
                     map(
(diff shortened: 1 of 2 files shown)


sfc-gh-fpawlowski:
Closing — cherry-picked this commit directly onto fix/snow-4072353-preserve-quoted-non-select (#4352) instead of keeping it as a separate stacked PR.

PR #4369, opened 2026-09-11
SNOW-3898656: XML perf optimization for multi-file batching (stacked 4)

sfc-gh-mayliu:

   

   Fixes SNOW-3898656

This PR adds small-file batching and intra-batch concurrency to `DataFrameReader.xml()` directory reads. 

This fixes the batching-eligibility threshold to trigger automatically at realistic file sizes, removing the need for manual workarounds. It eliminates the performance bottleneck of reading thousands of small files and scales correctly across the entire size spectrum: many small files, mixed-size directories, and single huge files.

### Technique

* **Batching:** Small files are grouped into fewer UDTF calls, packing up to `batchTargetBytes` (default 50MB) per row. This collapses 50k 1MB-files into ~1,000 invocations.
* **Concurrency:** A thread pool reads a batch's files concurrently to hide I/O latency. Overhead is paid once per batch, not once per file.
* **Attribution:** Length-prefixed encoding prevents path corruption. Every output record retains its original file path and byte offset, preserving parent/child range-joins.
* **Eligibility Threshold (New):** A new, independent `minWorkerBytes` (~4MB) determines batching eligibility. This decouples batching from `numWorkers` and the internal 1KB read-buffer size.

### Behavior by Workload

* **Single large file (e.g., 50GB):** Splits across up to `numWorkers` workers. Unaffected by batching.
* **Many small files (e.g., 50,000 × 1MB):** Every file is a batching candidate by default. `numWorkers` and `batchTargetBytes` tune independently without interfering with each other.
* **Mixed-size directories (1KB–10GB):** Small files batch together while large files split across workers. Both happen in the same read (verified by integration tests).

### Limitations & Activation

* **Defaults:** `batchTargetBytes=50MB` (exposed option); `minWorkerBytes≈4MB` (internal).
* **Scope:** Only files under `minWorkerBytes` are batched. Larger files always split.
* **Activation:** Automatic. No option needs to be set to trigger batching on realistically-sized small files.

#### Full-Workflow Benchmark (50,000 files / 53.5GB)
*End-to-end pipeline on an XL warehouse:*

| Pass 1 (header) | Pass 2 (children) | Range JOIN | Total |
| :--- | :--- | :--- | :--- |
| 72.86s | 173.23s | 31.54s | **277.63s** |

* **Efficiency:** Generated only 1,036 worker-assignment rows (down from 50,000 files).
* **Integrity:** 100% accurate. Exact row match (1,602,033 rows) with 0 duplicates and all 50,000 source files represented.

#### Single-File Benchmark (50GB)
*`numWorkers=64`, Medium warehouse:*

* **Execution:** 555.66s. Confirms single-large-file read performance is completely unaffected by this change.

#### Concurrency Speedup (intra-batch threadpoolexecutor)
*Isolated test reading 100 files (~101MB) in a single UDTF invocation:*

* **Concurrent:** ~93-95s
* **Sequential:** ~118-121s
* **Result:** Intra-batch concurrency yields a consistent **20-23% speedup** across warehouse sizes.

### DIFF
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -160,12 +160,9 @@ DEFAULT_MAX_WORKERS: int = 16
 # worker-assignment row count.
 _XML_WORKER_ASSIGNMENT_MAX_ROWS_PER_VALUES: int = 200_000
 
-# Limits on how many small files are packed into one worker-assignment row. The byte target
-# keeps a row's total work near one worker's share so batched rows do not become stragglers;
-# the file cap bounds a row independently of size, since batched files are read concurrently
-# and each one costs memory in the UDTF sandbox.
+# How many bytes of small files to pack into one worker-assignment row, keeping a row's
+# total work near one worker's share so batched rows do not become stragglers.
 XML_BATCH_TARGET_BYTES: int = 50 * 1024 * 1024
-XML_BATCH_MAX_FILES: int = 500
 
 
 def _xml_worker_assignments(
@@ -217,10 +214,7 @@ def _pack_xml_assignments(
             flush()
             rows.extend(ranges)
             continue
-        if pending and (
-            pending_bytes + file_size > target_bytes
-            or len(pending) >= XML_BATCH_MAX_FILES
-        ):
+        if pending and pending_bytes + file_size > target_bytes:
             flush()
         pending.append(ranges[0])
         pending_bytes += file_size
--- a/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
+++ b/src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
@@ -160,6 +160,13 @@ DEFAULT_MAX_WORKERS: int = 16
 # worker-assignment row count.
 _XML_WORKER_ASSIGNMENT_MAX_ROWS_PER_VALUES: int = 200_000
 
+# Limits on how many small files are packed into one worker-assignment row. The byte target
+# keeps a row's total work near one worker's share so batched rows do not become stragglers;
+# the file cap bounds a row independently of size, since batched files are read concurrently
+# and each one costs memory in the UDTF sandbox.
+XML_BATCH_TARGET_BYTES: int = 50 * 1024 * 1024
+XML_BATCH_MAX_FILES: int = 500
+
 
 def _xml_worker_assignments(
     file_path: str, file_size: int, max_workers: int, chunk_size: int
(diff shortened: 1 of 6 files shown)


src/snowflake/snowpark/_internal/xml_reader.py
```diff
@@ -904,6 +953,37 @@ def _iter_xml_records(
...
+
+def _process_batch_concurrently(batch, row_tag, **read_kwargs):
+    """Read a batched row's files on separate threads, yielding (path, record, offset).
+    Overlapping the reads is what makes packing files into one invocation pay off, since
+    each spends most of its time waiting on stage I/O. ``result_template``/``schema_type``
+    are safe to share across threads since neither is mutated."""
+
+    def read_one(entry):
+        path, start, end = entry
+        return path, list(process_xml_range(path, row_tag, start, end, **read_kwargs))
```
sfc-gh-yuwang, line 977: if I understand correctly, this threadpool is happening on the client env to kickoff UDTF in parallel?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -160,13 +160,26 @@
 # worker-assignment row count.
 _XML_WORKER_ASSIGNMENT_MAX_ROWS_PER_VALUES: int = 200_000
 
+# How many bytes of small files to pack into one worker-assignment row, keeping a row's
+# total work near one worker's share so batched rows do not become stragglers.
+XML_BATCH_TARGET_BYTES: int = 50 * 1024 * 1024
```
sfc-gh-yuwang, line 165: discussed offline, we should have some document to state how these numbers are decided

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
```diff
@@ -160,13 +160,26 @@
...
+
```
sfc-gh-yuwang, line 166: Suppose file size are very small, ~1kb for each file, if I understand correctly, there will be 50k files in each batch with the 50MB limit, in this case, the length of query can easily exceed the limit of 1MB(according to cursor) and fail. Is this true?

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-yuwang, line 977: If one file in the batch failed reading, it would fail the whole batch right? How does the retry work in this scenario?

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-mayliu: Good point that we need to honor cell size limit when packing files into a batch. Just pushed a commit to cap multi-file batch by both `batchTargetBytes` and `VARIANT_COLUMN_SIZE_LIMIT`.

However, I'm not sure where the 1MB came from. `VARIANT` limit is around 128MB, but standard table's cell size limit is 16MB and the default value for `VARIANT_COLUMN_SIZE_LIMIT` is also 16MB, so I adhered to it for encoded path + start/end byte pos for safety's sake.

src/snowflake/snowpark/_internal/xml_reader.py
sfc-gh-mayliu: Another great point, it's right that the previous design would have one bad file fail the entire batch. Instead of retrying in our UDTF, I added try/except to catch `snowflakeFile` related errors and hook them with `mode`. Then permissive mode has an corrupted-column entry describing the failure, and failfast mode would have the one bad file read stop the pipeline to align single file vs directory read behaviors.

src/snowflake/snowpark/_internal/analyzer/snowflake_plan.py
sfc-gh-yuwang: oh, the 1 MB I mean the snowflake recommend the query being executed have a text size smaller than 1 MB. according to glean:
 Queries larger than 1 MB may still execute, but Snowflake can truncate them before storing them in metadata, so rerun/retry behavior and query-history text may be incomplete.
we should test this behavior

sfc-gh-yuwang (approved):
pre approve assume all comments will be addressed.
please also check the merge gate to make sure they are green!

PR #4380, opened 2026-09-21
Rewrite Anaconda package pins to catalog version strings

sfc-gh-aling:
## Summary
- Follow-up to the doctest unblock (#4381). This is the **client** fix: Snowpark validates Anaconda availability with PEP 440 but `CREATE FUNCTION PACKAGES` matches catalog strings exactly, so pins like `python-dateutil==2.9.0.post0` fail against `2.9.0post0+snowflake1`.
- When an exact `==` pin is PEP 440-equivalent to a catalog version but is not a literal catalog string, rewrite it to the catalog form. Unpinned names and prefix/range specifiers such as `pandas==2.3.*` are unchanged.
- Does **not** change Python 3.14 / PyPI artifact-repository resolution (that path never consults the Anaconda catalog). Dedicated tests cover Anaconda rewrite, PyPI keep-pin, and 3.14 defaults.

## Test plan
- [x] Unit tests on 3.13 and 3.14 (`test_anaconda_package_spec`, `test_resolve_packages_rewrites_pep440_pin_to_catalog_string`, `test_resolve_packages_pypi_keeps_pep440_pin`, `test_py314_*`, `test_pypi_udf_and_sproc_keep_pep440_package_pin`)
- [ ] After merge of #4381 or this PR, confirm daily precommit macOS AWS/Azure doctests pass
- [ ] Optional live: `session.add_packages(dateutil)` on AWS/Azure and check `PACKAGES` uses `2.9.0post0+snowflake1`


sfc-gh-aling:
Split out of the original combined change. Merge **#4381** first to unblock daily precommit (doctest-only). This PR is the product fix plus tests and does not depend on #4381 (hunks do not overlap).

